PR #4946, opened 2022-12-12
Add new app config connection string

praveenkuttappan:
Add new app configuration to test server

### DIFF
--- a/eng/pipelines/templates/steps/apiview-ui-tests.yml
+++ b/eng/pipelines/templates/steps/apiview-ui-tests.yml
@@ -53,6 +53,7 @@ steps:
     APIVIEW_AUTHENTICATIONSCHEME: "Test"
     APIVIEW_SENDGRID__KEY: "Test"
     APIVIEW_APIKEY: "$(azuresdk-apiview-apikey)"
+    APIVIEW_CONNECTIONSTRINGS__APPCONFIG: $(apiview-appconfig-connectionstring)
 
 - task: Powershell@2
   inputs:


eng/pipelines/templates/steps/apiview-ui-tests.yml
```diff
@@ -53,6 +53,7 @@ steps:
     APIVIEW_AUTHENTICATIONSCHEME: "Test"
     APIVIEW_SENDGRID__KEY: "Test"
     APIVIEW_APIKEY: "$(azuresdk-apiview-apikey)"
+    APIVIEW_CONNECTIONSTRINGS__APPCONFIG: $(apiview-appconfig-connectionstring)
```
benbp, line 56: Not directly relevant here, but could we pair a single keyvault with each apiview instance to load config secrets on startup, that way secrets don't have to be passed in? Or is the env method easier to work with in an asp environment?

benbp (approved):

eng/pipelines/templates/steps/apiview-ui-tests.yml
praveenkuttappan: this is only for test instance we run on pipeline. Production and deployed instances use KV and App configuration

src/dotnet/APIView/APIViewWeb/Pages/Assemblies/Index.cshtml
```diff
@@ -237,9 +243,14 @@
                                 </div>
                                 <div class="tab-pane" id="cadl-help">
                                     <ol>
-                                        <li>
-                                            Provide URL to cadl project root path.
-                                        </li>                                        
+                                        
+                                       Provide URL to cadl project root path.
```
maririos, line 247: The changes in this file are for the description box thing I showed you in the morning, right?

praveenkuttappan: correct

.azure-pipelines/apiview.yml
```diff
@@ -141,11 +141,8 @@ stages:
 
       - job: 'Test'
```
weshaggard, line 142: I would really like to eventually see this under eng/pipelines or under the apiview directory as ci.yml there. I know this is only your backlog but every time I see it I think how inconsistent our tools repo is.

PR #4956, opened 2022-12-13
Pylint Updates

l0lawrence:
Checkers being Enabled (currently only will be on ApiView) and pylint version will be bumped to 2.15.8.

- Distributed Tracing -- was previously disabled due to false positives, edited the checker and re-enabled
- TypePropertyNameTooLong - checks that the naming of everything is less than 40 characters
- DeleteOperationReturnStatement - For delete* or begin_delete* operations checks that they return None or an LROPoller
- AsyncMethodsReturnAsyncIterables - checks that async methods if they are an ast.Call of another function, are returning async functions. (i.e AsyncLROPoller)
- NoAzureCoreTracebackUseRaiseFrom - checks that we are not using azure.core traceback import for exceptions by checking the imports of each file

Did test on main SDKs and will get a better view of the impact once they are on APiView, but DistributedTracing and TypePropertyNameTooLong looks like they will have the most impact.


tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1985,6 +1960,185 @@ def visit_importfrom(self, node):
...
+    __implements__ = IAstroidChecker
+
+    """Rule to check that the character length of type and property names are not over X characters."""
+    name = "name-too-long"
+    priority = -1
+    msgs = {
+        "C4751": (
+            "Name is over standard character length of 40.",
+            "name-too-long",
+            "Only use names that are less than 40 characters."
```
tjprescott, line 1974: Is this actually in the guidelines?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -357,6 +356,10 @@ class ClientMethodsHaveTracingDecorators(BaseChecker):
         ),
     )
     ignore_clients = ["PipelineClient", "AsyncPipelineClient", "ARMPipelineClient", "AsyncARMPipelineClient"]
+    ignore_functions = ["send_request"]
+    ignore_decorators = ["typing.overload", "builtins.classmethod", "azure.core.tracing.decorator.distributed_trace", 
+        "azure.core.tracing.decorator_async.distributed_trace_async"]
+    ignore_internal = False
```
tjprescott, line 362: Is it not possible to simply annotate the relevant azure-core things with suppression annotations instead of special-casing these to never flag in the first place?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1068,56 +1087,6 @@ def visit_classdef(self, node):
             pass
 
 
-class PackageNameDoesNotUseUnderscoreOrPeriod(BaseChecker):
```
tjprescott, line 1071: Did this get removed? If so, why?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
kristapratico: Related issue: https://github.com/Azure/azure-sdk-for-python/issues/26640

Was 40 chars the number that proved to have best signal:noise?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1985,6 +1960,185 @@ def visit_importfrom(self, node):
...
+                            node=i,
+                            confidence=None,
+                        )        
+                except:
+                    # Gets the names of ast.Assign statements
+                    for j in i.targets:
+                        if len(j.name) > self.STANDARD_CHARACTER_LENGTH:
+                            self.add_message(
+                                msgid=f"name-too-long",
+                                node=i,
```
kristapratico, line 2021: should be `j`?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1985,6 +1960,185 @@ def visit_importfrom(self, node):
...
+
+    def visit_functiondef(self,node):
+        """Visits all delete functions and checks that their return types
+        are LROPoller or None. """
+        try:
+            if node.name.startswith("delete") or node.name.startswith("begin_delete"):
+                inferred = node.infer_call_result()
+                for x in inferred:
+                    if str(x)=="Uninferable" or x == None or x.value == None or \
+                        x.name == "LROPoller" or x.name == "AsyncLROPoller":
```
kristapratico, line 2052: will this check if it's LROPoller[None] or LROPoller[Any]?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1985,6 +1960,185 @@ def visit_importfrom(self, node):
...
+
+class AsyncMethodsReturnAsyncIterables(BaseChecker):
+    __implements__ = IAstroidChecker
+
+    """Rule to check that an async method returns an async iterable."""
+    name = "async-return-async-iterable"
+    priority = -1
+    msgs = {
+        "C4753": (
+            "async method returns sync iterable.",
```
kristapratico, line 2073: nit: maybe include in the error message what it should return.

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1985,6 +1960,185 @@ def visit_importfrom(self, node):
...
+            "async-return-async-iterable",
+            "async methods should return async iterables."
+        ),
+    }
+
+    def visit_asyncfunctiondef(self, node):
+        """Checks all async functions, if there is a return statement, checks that it is preceded by an await. 
+        If it is a call to a function, it checks that the function is async. i.e AsyncLROPoller"""
```
kristapratico, line 2081: AsyncItemPaged?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1985,6 +1960,185 @@ def visit_importfrom(self, node):
...
+        try:
+            if "Client" in node.parent.name:
```
kristapratico, line 2083: Not sure I understand what this one should be checking. From the description it seems like we should we be looking for an ItemPaged returned on an async method?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1985,6 +1960,185 @@ def visit_importfrom(self, node):
...
+
+class NoAzureCoreTracebackUseRaiseFrom(BaseChecker):
+    __implements__ = IAstroidChecker
+
+    """Rule to check that we don't use raise_with_traceback in azure core."""
+    name = "no-raise-with-traceback"
+    priority = -1
+    msgs = {
+        "C4754": (
+            "Don't use raise_with_traceback, use python 3 raise from syntax.",
```
kristapratico, line 2107: nit

```suggestion
            "Don't use raise_with_traceback, use python 3 'raise from' syntax.",
```

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1985,6 +1960,185 @@ def visit_importfrom(self, node):
...
+            "no-raise-with-traceback",
+            "Don't use raise_with_traceback instead use python 3 raise from syntax."
+        ),
+    }
+    AZURE_CORE= "azure.core.exceptions"
```
kristapratico, line 2112: looks unused... should this be used to check if it's being imported from that module?

tools/pylint-extensions/pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
```diff
@@ -3115,3 +3115,168 @@ def test_allowed_imports(self):
...
+                    line=2,
+                    node=class_node,
+                    col_offset=0, 
+                    end_line=2, 
+                    end_col_offset=52
+                )
+        ):
+            self.checker.visit_classdef(class_node)
+
+    def test_variable_name_too_long(self):
```
kristapratico, line 3142: test_function_name_too_long?

tools/pylint-extensions/pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
```diff
@@ -3115,3 +3115,168 @@ def test_allowed_imports(self):
...
+                    line=3,
+                    node=function_node,
+                    col_offset=4, 
+                    end_line=3, 
+                    end_col_offset=54
+                )
+        ):
+            self.checker.visit_functiondef(function_node)
+
+    def test_function_name_too_long(self):
```
kristapratico, line 3165: test_variable_name_too_long?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -371,15 +374,23 @@ def visit_functiondef(self, node):
...
+            split_path = path.split(".")
+            new_path = ".".join(split_path[:len(split_path)-1])
+            if new_path.count("_") == 0:
+                if node.parent.name.endswith("Client") and node.is_method() and not node.name.startswith("_") and \
+                        node.parent.name not in self.ignore_clients:
+                    if node.args.kwarg and node.name not in self.ignore_functions and not node.name.endswith("client"):
+                        if self.ignore_decorators[0] not in node.decoratornames() \
+                            and self.ignore_decorators[1] not in node.decoratornames() \
+                                and self.ignore_decorators[2] not in node.decoratornames():
+
```
tjprescott, line 389: This syntax isn't maintainable. You probably want to do something like take the union of `decoratornames` and `ignore_decorators` and verify that union is empty. That way if you add to the list you don't need to modify this.

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -389,18 +400,25 @@ def visit_asyncfunctiondef(self, node):
...
+            path = node.root().name
+            split_path = path.split(".")
+            new_path = ".".join(split_path[:len(split_path)-1])
+            if new_path.count("_") == 0:
+                if node.parent.name.endswith("Client") and node.is_method() and not node.name.startswith("_") and \
+                        node.parent.name not in self.ignore_clients:
+                    if node.args.kwarg and node.name not in self.ignore_functions and not node.name.endswith("client") \
+                        and self.ignore_decorators[0] not in node.decoratornames() \
+                            and self.ignore_decorators[1] not in node.decoratornames() \
+                                and self.ignore_decorators[3] not in node.decoratornames():
```
tjprescott, line 416: Same comment here. This should be simplified so it doesn't need to be changed if `ignore_decorators` changes.

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
l0lawrence: Our pylintrc file has us skipping setup.py for pylint checks, so this rule never actually is used here. I made an [issue](https://github.com/Azure/azure-sdk-for-python/issues/28202)to investigate if this is a check we want in another part of the pipeline instead.

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
l0lawrence: 40 was enough that it wasn't flagging everything, but it still catches unnecessarily long names

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -371,15 +373,22 @@ def visit_functiondef(self, node):
...
-                    )
-        except AttributeError:
+            path = node.root().name
+            split_path = path.split(".")
+            new_path = ".".join(split_path[:len(split_path)-1])
+            if new_path.count("_") == 0:
+                if node.parent.name.endswith("Client") and node.is_method() and not node.name.startswith("_") and \
+                        node.parent.name not in self.ignore_clients:
+                    if node.args.kwarg and node.name not in self.ignore_functions and not node.name.endswith("client") \
+                        and len(self.ignore_decorators.intersection(node.decoratornames())) == 0 and \
```
tjprescott, line 385: You could probably simplify this to
`and not self.ignore_decorators.intersection(node.decoratornames())` since an empty set would be falsey.

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -888,22 +908,39 @@ def visit_return(self, node):
...
-                        if any(v for v in node.value.infer() if "def by_page" in v.as_string()):
-                            paging_class = True
-                    except (astroid.exceptions.InferenceError, AttributeError, TypeError): # astroid can't always infer the return
+                        for value in node.value.infer():
+                            if str(value) == "Uninferable":
+                                logger.debug("Pylint custom checker failed to check if client list method uses core paging.")
+                                return
+                            elif value.name == "ItemPaged" and isinstance(node.parent, astroid.FunctionDef):
+                                paging_class = True
+                            elif value.name == "AsyncItemPaged" and isinstance(node.parent, astroid.AsyncFunctionDef):
```
kristapratico, line 919: I'm not sure an async paging method will be captured as an astroid.AsyncFunctionDef since it's not defined with `async def`

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1985,6 +1978,165 @@ def visit_importfrom(self, node):
...
+    STANDARD_CHARACTER_LENGTH = 40
+
+    def visit_classdef(self,node):
+        """Visit every class and check that the
+         class name is within the character length limit."""
+
+        try:
+            if len(node.name) > self.STANDARD_CHARACTER_LENGTH:
+                self.add_message(
+                    msgid=f"name-too-long",
```
kristapratico, line 2004: nit: doesn't look like this needs to be an f-string

tools/pylint-extensions/pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
```diff
@@ -1909,7 +1950,7 @@ class SomeClient(): #@
             async def list_thing(self): #@
                 return AsyncItemPaged()
             @distributed_trace
-            def list_thing2(self): #@
+            async def list_thing2(self): #@
```
kristapratico, line 1953: We can't rely on `async def` for our async paging methods. We'll need to think of another way to detect this :thinking:

tools/pylint-extensions/pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
```diff
@@ -3115,3 +3150,125 @@ def test_allowed_imports(self):
...
+    """Test that we are checking the return type of delete functions is correct"""
+    CHECKER_CLASS = checker.DeleteOperationReturnStatement
+
+    def test_delete_operation(self):
+        node = astroid.extract_node(
+        """
+            from azure.core.polling import LROPoller 
+            from typing import Any
+            class MyClient():
+                def delete_some_function(self, **kwargs): #@
```
kristapratico, line 3236: should be begin_delete if returning an LROPoller unless this is meant to test that the delete function is not returning None?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1985,6 +1977,165 @@ def visit_importfrom(self, node):
...
+                                node=j,
+                                confidence=None,
+                            )          
+        except:
+            pass      
+
+    visit_asyncfunctiondef = visit_functiondef
+
+
+class DeleteOperationReturnStatement(BaseChecker):
```
kristapratico, line 2047: Thinking about this one... the delete* seems pretty straightforward because we can sniff out whether it's returning a non-None value. The begin_delete* is a bit tougher because I don't think it's straightforward to tell what the final result of the LROPoller is in code. In general, we want to make sure this isn't something like LROPoller[JobStatus]. I wonder if it would make sense to check the type annotation for the method instead.

l0lawrence: It is now checking node.returns which can be populated by the typehints

l0lawrence:
broke this down and merged in separate prs into repro for clarity

PR #4959, opened 2022-12-13
Add script for removing worm storage accounts

benbp:


eng/scripts/Remove-WormStorageAccounts.ps1
```diff
@@ -0,0 +1,38 @@
...
+            }
+            $ctx = New-AzStorageContext -StorageAccountName $account.StorageAccountName
+            $ctx | Get-AzStorageContainer | Get-AzStorageBlob | Remove-AzStorageBlob -Force
+            # Use AzRm cmdlet as deletion will only work through ARM with the immutability policies defined on the blobs
+            $ctx | Get-AzStorageContainer | % { Remove-AzRmStorageContainer -Name $_.Name -StorageAccountName $ctx.StorageAccountName -ResourceGroupName $group.ResourceGroupName -Force }
+            Remove-AzStorageAccount -StorageAccountName $account.StorageAccountName -ResourceGroupName $account.ResourceGroupName -Force
+        }
+    }
+    if ($PSCmdlet.ShouldProcess($group.ResourceGroupName)) {
+        Remove-AzResourceGroup -ResourceGroupName $group.ResourceGroupName -Force -AsJob
```
benbp, line 34: I'm wrapping this instead of just using the native WhatIf because even with WhatIf it still farms it out to a job which can be misleading and make the user think the group is actually being deleted.

eng/scripts/Remove-WormStorageAccounts.ps1
```diff
@@ -0,0 +1,38 @@
...
+$groups = Get-AzResourceGroup | ? { $_.ResourceGroupName.StartsWith($GroupPrefix) } | ? { $_.ProvisioningState -ne 'Deleting' }
+
+foreach ($group in $groups) {
+    $accounts = Get-AzStorageAccount -ResourceGroupName $group.ResourceGroupName
+    if ($accounts) {
+        foreach ($account in $accounts) {
+            if ($PSCmdlet.ShouldProcess($group.ResourceGroupName)) {
+                Write-Host "Removing $($account.StorageAccountName) in $($account.ResourceGroupName)"
+            } else {
+                Write-Host "What if: Removing $($account.StorageAccountName) in $($account.ResourceGroupName)"
```
weshaggard, line 24: I thought the whatif logic in shouddprocess would already output this

weshaggard: Did you want to ship if whatif is set? I see you are still calling remove below.

benbp: I fixed it up to just check `$WhatIfPreference`. The extra stuff from the `ShouldProcess` log makes the output hard to parse in my opinion.

I still call Remove below because the parent `-WhatIf` will still pass through to the Az calls and I want those to be part of the whatif logging since it shows each container that would be deleted.

weshaggard (approved):

Issue #4478, opened 2022-10-20
Implement an existing fabricbot rules using github actions.


JimSuplizio:
We're looking at moving from FabricBot into GitHub Actions. This issue is for tracking the work to do this.

PR #4974, opened 2022-12-15
Delete CreateRuleFabricBot tool

konrad-jamrozik:
@weshaggard @benbp, per my chat with @jsquire and @JimSuplizio, this tool is obsolete and no longer used. Currently, the `.github/fabricbot.json` files, like the one for [azure-sdk-for-net](https://github.com/Azure/azure-sdk-for-net/blob/main/.github/fabricbot.json) are considered source of truth. Plus, we will be moving away from these files soon, per 
- #4478

Once this PR is merged, I will also remove corresponding internal and public pipelines:  
https://dev.azure.com/azure-sdk/internal/_build?definitionId=1602  
https://dev.azure.com/azure-sdk/public/_build?definitionId=1601


weshaggard (commented):
I will defer to @jsquire and @JimSuplizio about whether or not we use this tool any longer. If we don't I'm good with deleting it but it might be worth keeping around until we finish our full fabricbot transition.

konrad-jamrozik:
Per discussion with @weshaggard, closing this PR until #4478 is done. We can reopen it then.

PR #4979, opened 2022-12-15
Add github token when cloning cadl repo

praveenkuttappan:
Set GitHub PAT as env variable to clone private repo

PR #4980, opened 2022-12-15
Feature/block auto review revision updates

praveenkuttappan:
Hide button to add new revision for automatic and manual reviews


maririos (approved):
Changes in [Revisions.cshtml](https://github.com/Azure/azure-sdk-tools/pull/4980/files#diff-653650f300d7e3b87c4121a2ff7e6427a088f357574df05d91b0c2927d4ff2cb) LGTM.
 I imagine the other changes are from another work that you based your branch from.

maririos:
Could you  make sure to close https://github.com/Azure/azure-sdk-tools/pull/4814 ?

eng/scripts/Create-Apiview-Token-Cadl.ps1
```diff
@@ -78,7 +79,13 @@ if ($revs)
...
         if ($GitRepoName -ne $prevRepo)
         {
-            git clone --no-checkout --filter=tree:0 "https://github.com/$GitRepoName"
+            $gitUrl = "https://github.com"
+            if ($GitPat)
+            {
+                $gitUrl = "https://$GitPat@github.com"
+            }
+
+            git clone --no-checkout --filter=tree:0 "$gitUrl/$GitRepoName"
```
weshaggard, line 88: Are we doing clones on top of each other if there is multiple different repos? i.e. one from private and one from public? We might need to make sure we clone into unique directories.

weshaggard (approved):
One question related to multiple reviews but otherwise looks good assuming the pat has the necessary permissions to the repo.

eng/scripts/Create-Apiview-Token-Cadl.ps1
weshaggard: Looks like this is not intended for this PR but instead for https://github.com/Azure/azure-sdk-tools/pull/4979.

PR #4937, opened 2022-12-10
Set .NET used by our tools to `net6.0` + minor updates to .sln files

konrad-jamrozik:
This PR addresses issues #4934 and #4935 for the following tools and other projects, by migrating their target frameworks to `net6.0`, and ensuring that their `ci.yml` sets `DotNetCoreVersion: 6.0.403`:
- `version-guard`;
- `CreateRuleFabricBot` and its tests;
- `code-owners-parser` and its tests. This required disabling execution of `get-codeowners.ps1` script as part of the test suite. Follow-up work to re-enable it: #4938;
- `notification-configuration`;
- `identity-resolution`;
- `generator` (i.e. `Stress.Generator`) and its tests;
- `Azure.ClientSdk.Analyzers` and its tests;
- `SwaggerApiParser`. This only required setting `ci.yml` to point to .NET 6, as it was already targeting `net6.0`;
- `snippet-generator` and its tests;
- `tools\agent-time-extractor\AgentTimeExtractor\AgentTimeExtractor.csproj`;
- `src\dotnet\APIView\TestLibrary\TestLibrary.csproj`, which was using `netcoreapp2.1` which is out of support since [August 2021](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core), but [the APIView pipeline](https://dev.azure.com/azure-sdk/internal/_build?definitionId=1136&_a=summary) and the pipeline seems to be failing due to unrelated reasons, so usage of `netcoreapp2.1` seems to be non-blocking. Nevertheless, I updated it to `net6.0` anyway;
- `Azure.Sdk.Tools.HttpFaultInjector`;
- `Azure.Sdk.Tools.HttpFaultInjector.StorageBlobsSample`;
- `Azure.Sdk.Tools.HttpFaultInjector.HttpClientSample`;
- Sample projects of `TestProxy`, but not the `TestProxy` itself, as it was already on `net6.0`, albeit it also targets `net5.0`, which I left unchanged;
- `PixelServer`, which required adding an import to `Program.cs`. Note I couldn't find a corresponding `ci.yml` or pipeline for it.

## Tools not updated to `net6.0` in this PR

`<TargetFramework>` occurrences **not** migrated as part of this PR, and why:

- Any tools targetting `netstandard` as [it is compatible with .NET 6.0](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1);
[...]

Issue #4934, opened 2022-12-10
Some of our internal pipelines fail to build due to using out-of-support .NET SDKs


konrad-jamrozik:
Our projects target out of support .NET versions, causing pipelines to fail, because they treat ".NET SDK version out of support" warnings as errors. For examples, see below. 

I believe this problem occurs when a newest version of SDK is used to build a project targeting older version of SDK. Thus building project targeting `net5.0` with `net7.0` (i.e. currently latest) will fail the build (see `ConfigureFabricBot` example below), but building it with `net5.0` won't (see e.g. [this notification-configuration build](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2036958&view=logs&j=6d7966cd-0423-5022-78e0-a1ae4dd99f62&t=f8b5ff87-1d8d-5ec2-1299-b1e6d411c139)). We end up building many projects with newest SDK, per #4935.

Notably, we use [.NET Core 3.1.405 in globals.yml](https://github.com/Azure/azure-sdk-tools/blob/fb47586166d62344b0849705808957b857b4bd7b/eng/pipelines/templates/variables/globals.yml#L4) which will [go out of support on December 13, 2022](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core), presumably breaking most of our pipelines. I believe the breakage will happen only if `net6.0` is used to build a project targeting obsolete .NET SDK, which is another reason to set .NET SDK instead of using latest one, per #4935.

## Proposed fix

The fix to this problem is composed of following parts:

- Review all our `.csproj` files `<TargetFrameowork>` and retarget them to `net6.0`, per https://github.com/Azure/azure-sdk-tools/issues/4888#issue-1479651387. 
  See PRs:
  -  #4915 
  - and #4937.
- Set the SDK versions of the projects by properly declaring them in `ci.yml` files; see #4935 and PRs #4915 and #4937.
- Once the projects are migrated to `.net6.0` and imminent threat of failure is mitigated, get rid of the SDK versions set per project, and instead set them globally, as done PR #4916.

## Example failures

Below are some examples of failures:

## tools - version-guard failure

Observe [recent build failure of internal "tools - version guard" pipeline](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2047073&view=logs&j=6d7966cd-0423-5022-78e0-a1ae4dd99f62&t=f8b5ff87-1d8d-5ec2-1299-b1e6d411c139) caused by:

[...]

PR #4984, opened 2022-12-15
Migrate Tests.CI.BuildTasks.csproj and Tests.CI.Common.csproj to net6.0

konrad-jamrozik:
This PR is a follow-up to:
- #4937

it contributes to addressing:
- #4934

Here I migrate some more projects from `netcoreapp2.1` to `net6.0`.

I also quenched the error (escalated from warning):

> MSB3270: There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference (...) AMD64"

by forcing the relevant tests project to be win10-x64 too.


sima-zhu (approved):
LGTM. Thanks for taking quick action on this.

PR #4987, opened 2022-12-16
Show EBN attribute and hidden APIs by default

JoshLove-msft:
EBN attribute:
image

Show Hidden APIs selected by default:
image


KrzysztofCwalina:
Can we add an option to hide the attribute? It breaks the flow of overloads and makes the API listing harder to read.

JoshLove-msft:
> Can we add an option to hide the attribute? It breaks the flow of overloads and makes the API listing harder to read.

Couldn't you just Unselect Show Hidden APIs? Then both the attribute and the EBN'ed method are not shown.

Issue #2770, opened 2022-02-16
Add support for wildcard handling of file matching in codeowners library


weshaggard:
See comment https://github.com/Azure/azure-sdk-for-js/pull/20354#discussion_r806177441.

It would be helpful to support some basic wildcard matching in our code owners file library. Maybe we cannot easily support all wildcard patterns but we could probably support `**` for something like `/sdk/**/ci.mgmt.yml`.

konrad-jamrozik:
@weshaggard I analyzed the relevant code and here is my plan of action. Sharing with you in case you have any concerns or suggestions.

## What code needs to change

Principally, there are two places that need to be modified:

1. [CodeOwnersFile.FindOwnersForClosestMatch](https://github.com/Azure/azure-sdk-tools/blob/main/tools/code-owners-parser/CodeOwnersParser/CodeOwnersFile.cs#LL79-L90C14):

``` csharp
// We want to find the match closest to the bottom of the codeowners file.
// CODEOWNERS sorts the paths in order of 'RepoPath', 'ServicePath' and then 'PackagePath'.
for (int i = codeOwnerEntries.Count - 1; i >= 0; i--)
{
    string codeOwnerPath = codeOwnerEntries[i].PathExpression.Trim('/');
    // Note that this only matches on paths without glob patterns which is good enough
    // for our current scenarios but in the future might need to support globs
    if (targetPath.StartsWith(codeOwnerPath, StringComparison.OrdinalIgnoreCase))
    {
        return codeOwnerEntries[i];
    }
}
```

2. [PullRequestLabelFolderCapability.ReadConfigurationFromFile](https://github.com/Azure/azure-sdk-tools/blob/23703d23e922257be2723625c8af8b16dc66a116/tools/CreateRuleFabricBot/CreateRuleFabricBot/Rules/PullRequestLabel/PullRequestLabelFolderCapability.cs#LL64-L75C18):

``` csharp
// Entries with wildcards are not yet supported
if (entries[i].ContainsWildcard)
{
    // log a warning there

    if (entries[i].PRLabels.Any())
    {
        Colorizer.WriteLine("[Yellow!Warning]: The path '[Cyan!{0}]' contains a wildcard and a label '[Magenta!{1}]' which is not supported!", entries[i].PathExpression, string.Join(',', entries[i].PRLabels));
    }

    continue; //TODO: regex expressions are not yet supported
}
```

## Proposed modifications

[...]

Issue #4859, opened 2022-12-02
Implement CODEOWNERS file changes validation (gating), with baselining


konrad-jamrozik:
## Background

Our Azure SDK repositories leverage `CODEOWNERS` files (e.g. [azure-sdk-for-python](https://github.com/Azure/azure-sdk-for-python/blob/main/.github/CODEOWNERS), [azure-sdk-for-net](https://github.com/Azure/azure-sdk-for-net/blob/main/.github/CODEOWNERS)) in two ways.

First, to determine pull request reviewers, via GitHub interpreter of code owners file, [as documented on GitHub docs "About code owners"](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners).

Second, to apply various additional augmentations, like automatically label PRs, notify about new issues being filed, or notify about relevant build failures. This is done by our own custom logic, as documented [in azure-sdk/docs/pipelines/opensource.md](https://github.com/Azure/azure-sdk/blob/main/docs/policies/opensource.md#codeowners). 

## Problem statement

The problem on the functionality relying on code owners contents, is that we do not have code owners file contents validation, thus resulting in ill-defined files, resulting in problems like notifications silently failing to be sent.

**The GitHub-based logic, responsible for adding reviewers to PRs, has a GitHub-provided validator** that results in a warnings banner if the file validation fails, e.g.:

![image](https://user-images.githubusercontent.com/4429827/205234737-cd1094f7-9a06-43d4-acb7-c7a6fd8aedfc.png)

![image](https://user-images.githubusercontent.com/4429827/205234774-60e03674-8e1a-487d-b480-4ce15d749a5e.png)

The main issue here is these are just warnings, while we would like to prevent merging any changes to code owners file unless they pass validation (i.e. to _gate_ the changes). For example, we would like to prevent folks from adding groups to the file, if they don't have necessary write permissions, as pointed out by [this comment](https://github.com/Azure/azure-sdk-tools/issues/4487#issuecomment-1289598162). We might also want to restrict the teams to a pre-vetted list, as seen [here (Microsoft-internal)](https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/49/Request-Access-to-Azure-SDK-repos).

**More importantly, our own logic does not have any validation in place**, hence if the code owners file is invalid, our augmentations will just silently be ignored, with no easy way to debug the issue.

## Proposed solution 

[...]

PR #5030, opened 2022-12-29
Add glob-based support for wildcards in `CODEOWNERS` file parser (disabled, behind a feature flag)

konrad-jamrozik:
This is a first PR to address:

- #2770 

## Changes made

Changes in [CodeOwnersFile.cs](https://github.com/Azure/azure-sdk-tools/pull/5030/files#diff-ee9b4b68edf43f257d84470fb4a37ffebc78b42e9d6fa0eedaabeb1416628427):

- In this PR I add a new implementation for matching entries in our CODEOWNERS file. **This implementation is disabled**, hidden behind a method feature flag parameter. 
- The method targeted by this PR is `CodeOwnersFile.FindOwnersForClosestMatch`. It now either calls its previous, unchanged, legacy implementation, now located in [CodeOwnersFile.FindOwnersForClosestMatchLegacyImpl](https://github.com/Azure/azure-sdk-tools/pull/5030/files#diff-ee9b4b68edf43f257d84470fb4a37ffebc78b42e9d6fa0eedaabeb1416628427R81), or the newly introduced implementation. The decision is [made here](https://github.com/Azure/azure-sdk-tools/pull/5030/files#diff-ee9b4b68edf43f257d84470fb4a37ffebc78b42e9d6fa0eedaabeb1416628427R76-R78) and is based on newly introduced method feature flag parameter, `useNewFindOwnersForClosestMatchImpl`. 
  - The `useNewFindOwnersForClosestMatchImpl` is by default `false`, meaning the code uses the legacy implementation. It is only ever set to `true` in the added tests.
- If `useNewFindOwnersForClosestMatchImpl` is `true`, then new implementation is used. That implementation's entry point is the newly-introduced [MatchedCodeOwnerEntry ctor](https://github.com/Azure/azure-sdk-tools/pull/5030/files#diff-30c0d1efa23b7260289c736ed51b407a59fafac23dece7348c08dddb3c64ec1bR19).

Newly added [MatchedCodeOwnerEntry.cs](https://github.com/Azure/azure-sdk-tools/pull/5030/files#diff-30c0d1efa23b7260289c736ed51b407a59fafac23dece7348c08dddb3c64ec1b) class:

- This class contains the logic for the new matching logic, as explained above.
- This class leverages [Microsoft.Extensions.FileSystemGlobbing](https://www.nuget.org/packages/Microsoft.Extensions.FileSystemGlobbing) ([namespace](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.filesystemglobbing?view=dotnet-plat-ext-7.0)) to reuse support for glob matching of `*` and `**` during path matching. See the class implementation and comments for details.

Newly added [Azure.Sdk.Tools.CodeOwnersParser.Tests/CodeOwnersFileTests.cs](https://github.com/Azure/azure-sdk-tools/pull/5030/files#diff-b396d224b17b2aac1b9c9db9d514ccf28255dcf37e41336f2302e24c563c78c9):

- This is a new project and a test class for testing the implementations of `FindOwnersForClosestMatch` directly. 
- It includes [a battery of over 50 test cases][battery] that achieve 100% code coverage both on the legacy and new parsing logic.
- The test battery also compares the behavior of the legacy and new matching logic, to be discussed w.r.t. backward compatibility issues and pre-existing bugs in the legacy logic.

Changes to [Azure.Sdk.Tools.RetrieveCodeOwners.Tests.csproj](https://github.com/Azure/azure-sdk-tools/pull/5030/files#diff-5272f6d935a874880addc7b67c8a12406291ae0d15c4e57cdf4c282b9570cd57):

- I updated and added some missing dependencies, based on the dependencies created when creating a new NUnit project in visual studio. 
    - This includes linter for NUnit, which resulted in some warnings around assertions, which broke PR build. I fixed them.
    - I also enabled strict nullability checking.

## Testing done

- Added the [battery] of tests with 100% coverage both of legacy and new implementation.
- There are also existing tests in [Azure.Sdk.Tools.RetrieveCodeOwners.Tests/MainTests.cs](https://github.com/Azure/azure-sdk-tools/pull/5030/files#diff-3cc91b617fdab1f05f05db30158b0867fa404612e1bc0953354d9e36fed07b13).

[battery]: https://github.com/Azure/azure-sdk-tools/pull/5030/files#diff-b396d224b17b2aac1b9c9db9d514ccf28255dcf37e41336f2302e24c563c78c9R24

## Next steps

- We need to discuss the discrepancies between legacy and new matching logic, as captured by the [battery] of tests. **Note that this is not a blocker to merging this PR**, as the new matching logic is currently unused beyond this test suite.
  - Notably, when the input path is e.g. `foo` the parser doesn't know if it is a file or a directory. This is a fundamental limitation, because CODEOWNERS entry of form `foo/` should match the path only if it is a directory, but not a file. We could work around this problem in two ways:
      - assume all `targetPaths` are files, always;
      - or assume that a `targetPath` is a directory if and only if it ends with `/`.
- As mentioned in the issue #2770, we also need to decide what to do with wildcard paths in our language repos that start to match once we enable the new matching logic. For example [these paths in azure-sdk-for-net CODEOWNERS](https://github.com/Azure/azure-sdk-for-net/blob/b8b0121fdbf309cefc08bae449c2fe768ef82100/.github/CODEOWNERS#L724-L725).
- Discovered a bug in the used library, have to find a workaround:  
    - https://github.com/dotnet/runtime/issues/80076
- In a follow-up PR, I am going to flip the feature flag to enable the new matching logic.
  - Once we are confident it works as expected, I am going to make another PR deleting the legacy logic.

## Reference

- https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax  
- https://git-scm.com/docs/gitignore#_pattern_format  
- https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.filesystemglobbing.matcher?view=dotnet-plat-ext-7.0
- https://man7.org/linux/man-pages/man7/glob.7.html

### DIFF
--- a/tools/code-owners-parser/CodeOwnersParser/CodeOwnersFile.cs
+++ b/tools/code-owners-parser/CodeOwnersParser/CodeOwnersFile.cs
@@ -59,13 +59,27 @@ namespace Azure.Sdk.Tools.CodeOwnersParser
             return entries;
         }
 
-        public static CodeOwnerEntry ParseAndFindOwnersForClosestMatch(string codeOwnersFilePathOrUrl, string targetPath)
+        public static CodeOwnerEntry ParseAndFindOwnersForClosestMatch(
+            string codeOwnersFilePathOrUrl,
+            string targetPath,
+            bool useNewFindOwnersForClosestMatchImpl = false)
         {
             var codeOwnerEntries = ParseFile(codeOwnersFilePathOrUrl);
-            return FindOwnersForClosestMatch(codeOwnerEntries, targetPath);
+            return FindOwnersForClosestMatch(codeOwnerEntries, targetPath, useNewFindOwnersForClosestMatchImpl);
         }
 
-        public static CodeOwnerEntry FindOwnersForClosestMatch(List<CodeOwnerEntry> codeOwnerEntries, string targetPath)
+        public static CodeOwnerEntry FindOwnersForClosestMatch(
+            List<CodeOwnerEntry> codeOwnerEntries,
+            string targetPath,
+            bool useNewFindOwnersForClosestMatchImpl = false)
+        {
+            return useNewFindOwnersForClosestMatchImpl
+                ? new MatchedCodeOwnerEntry(codeOwnerEntries, targetPath).Value
+                : FindOwnersForClosestMatchLegacyImpl(codeOwnerEntries, targetPath);
+        }
+
+        private static CodeOwnerEntry FindOwnersForClosestMatchLegacyImpl(List<CodeOwnerEntry> codeOwnerEntries,
+            string targetPath)
         {
             // Normalize the start and end of the paths by trimming slash
             targetPath = targetPath.Trim('/');
--- a/tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
+++ b/tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
@@ -1,0 +1,184 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.FileSystemGlobbing;
+
+namespace Azure.Sdk.Tools.CodeOwnersParser
+{
+    /// <summary>
+    /// Represents a CODEOWNERS file entry that matched to targetPath from
+    /// the list of entries, assumed to have been parsed from CODEOWNERS file.
+    ///
+    /// To obtain the value of the matched entry, reference "Value" member.
+    /// </summary>
+    internal class MatchedCodeOwnerEntry
+    {
+        public readonly CodeOwnerEntry Value;
+
+        private static readonly char[] unsupportedChars = { '[', ']', '!', '?' };
+
+        public MatchedCodeOwnerEntry(List<CodeOwnerEntry> entries, string targetPath)
+        {
+            this.Value = FindOwnersForClosestMatch(entries, targetPath);
+        }
+
+        /// <summary>
+        /// Returns a CodeOwnerEntry from codeOwnerEntries that matches targetPath
+        /// per algorithm described in:
+        /// https://git-scm.com/docs/gitignore#_pattern_format
+        /// and
+        /// https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax
+        ///
+        /// If there is no match, returns "new CodeOwnerEntry()".
+        /// </summary>
+        private static CodeOwnerEntry FindOwnersForClosestMatch(
+            List<CodeOwnerEntry> codeownersEntries,
+            string targetPath)
+        {
+            // targetPath is assumed to be absolute w.r.t. repository root, hence we ensure
+            // it starts with "/" to denote that.
+            if (!targetPath.StartsWith("/"))
+                targetPath = "/" + targetPath;
+
+            // We do not trim or add the slash ("/") at the end of the targetPath because its
+            // presence influences the matching algorithm:
+            // Slash at the end denotes the target path is a directory, not a file, so it might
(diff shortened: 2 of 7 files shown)


konrad-jamrozik:
@weshaggard while the new parser implementation in this PR requires a rewrite, I would like to merge this PR to keep it as a reference, e.g. for the bug I reported here:
- https://github.com/dotnet/runtime/issues/80076

Is it OK if we merge this? Note the new implementation is not used. I will make a follow-up PR that changes the implementation as we discussed in:
- https://github.com/Azure/azure-sdk-tools/issues/4859#issuecomment-1370360622

tools/code-owners-parser/CodeOwnersParser/CodeOwnersFile.cs
```diff
@@ -59,13 +59,27 @@ public static List<CodeOwnerEntry> ParseContent(string fileContent)
...
 
-        public static CodeOwnerEntry FindOwnersForClosestMatch(List<CodeOwnerEntry> codeOwnerEntries, string targetPath)
+        public static CodeOwnerEntry FindOwnersForClosestMatch(
+            List<CodeOwnerEntry> codeOwnerEntries,
+            string targetPath,
+            bool useNewFindOwnersForClosestMatchImpl = false)
+        {
+            return useNewFindOwnersForClosestMatchImpl
+                ? new MatchedCodeOwnerEntry(codeOwnerEntries, targetPath).Value
+                : FindOwnersForClosestMatchLegacyImpl(codeOwnerEntries, targetPath);
```
weshaggard, line 78: Is there a reason you are leaving the old implementation? Or is that just temporary while you test out your new?

tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
```diff
@@ -0,0 +1,184 @@
...
+            // Entries below take precedence, hence we read the file from the bottom up.
+            // By convention, entries in CODEOWNERS should be sorted top-down in the order of:
+            // - 'RepoPath',
+            // - 'ServicePath'
+            // - and then 'PackagePath'.
+            // However, due to lack of validation, as of 12/29/2022 this is not always the case.
+            for (int i = codeownersEntries.Count - 1; i >= 0; i--)
+            {
+                string codeownersPath = codeownersEntries[i].PathExpression;
+                if (ContainsUnsupportedCharacters(codeownersPath))
```
weshaggard, line 57: From a diagnostic standpoint it is probably interesting to know we encountered an unsupported char.

tools/code-owners-parser/CodeOwnersParser/CodeOwnersFile.cs
weshaggard: OK I now understand your intention after reading your description. I guess I should have started there but I jumped directly into the code.

weshaggard (approved):
Given the new implementation is currently disabled and you want it for the glob reference I'm good with you merging this and then pulling in your change to switch to regex.

tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
konrad-jamrozik: I'll add diagnostic output in the follow-up PR:

- #5063

tools/code-owners-parser/CodeOwnersParser/CodeOwnersFile.cs
konrad-jamrozik: @weshaggard I was also thinking to keep the old implementation at least for some time, in case the new implementation behavior leads to undesired effects. Then it will be easier to revert or reuse the old implementation. But probably after a week, or so, I would get rid of it.

PR #5035, opened 2023-01-03
CADL api review is not generated when emitter throws warnings

praveenkuttappan:
Update cadl emitter version and add additional logging

### DIFF
--- a/eng/pipelines/apiview-review-gen-cadl.yml
+++ b/eng/pipelines/apiview-review-gen-cadl.yml
@@ -45,7 +45,7 @@ jobs:
 
   - pwsh: |
       npm install -g @cadl-lang/compiler@0.37.0
-      npm install -g @azure-tools/cadl-apiview@latest
+      npm install -g @azure-tools/cadl-apiview@0.3.0
       npm list -g
     displayName: "Install npm packages"
 
--- a/eng/pipelines/apiview-review-gen-cadl.yml
+++ b/eng/pipelines/apiview-review-gen-cadl.yml
@@ -45,7 +45,7 @@ jobs:
 
   - pwsh: |
       npm install -g @cadl-lang/compiler@0.37.0
-      npm install -g @azure-tools/cadl-apiview@0.3.0
+      npm install -g @azure-tools/cadl-apiview@latest
       npm list -g
     displayName: "Install npm packages"
 
--- a/eng/scripts/Create-Apiview-Token-Cadl.ps1
+++ b/eng/scripts/Create-Apiview-Token-Cadl.ps1
@@ -44,6 +44,8 @@ function Stage-Apiview-File($packagePath, $reviewId, $revisionId)
     $tokenFilePath = Join-Path $packagePath "cadl-output"
     $stagingReviewPath = Join-Path $OutputDir $reviewId
     $stagingPath = Join-Path $stagingReviewPath $revisionId
+    Write-Host "Out files in cadl-output"
+    Get-ChildItem -Path $tokenFilePath
     Write-Host "Copying APIView file from '$($tokenFilePath)' to '$($stagingPath)'"
     New-Item $stagingPath -ItemType Directory -Force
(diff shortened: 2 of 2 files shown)


eng/scripts/Create-Apiview-Token-Cadl.ps1
```diff
@@ -44,6 +44,8 @@ function Stage-Apiview-File($packagePath, $reviewId, $revisionId)
     $tokenFilePath = Join-Path $packagePath "cadl-output"
     $stagingReviewPath = Join-Path $OutputDir $reviewId
     $stagingPath = Join-Path $stagingReviewPath $revisionId
+    Write-Host "Out files in cadl-output"
+    Get-ChildItem -Path $tokenFilePath
```
scbedd, line 48: Do you want recursion on this?

eng/pipelines/apiview-review-gen-cadl.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
   - pwsh: |
       npm install -g @cadl-lang/compiler@0.37.0
-      npm install -g @azure-tools/cadl-apiview@0.3.0
```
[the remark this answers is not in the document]
praveenkuttappan: Yup. It should only be published when a new emitter is ready to be consumed. @tjprescott has mentioned to use latest available on npm. My personal preference is to use a pinned version.

eng/scripts/Create-Apiview-Token-Cadl.ps1
praveenkuttappan: Nope. I only want to list first level output under `cadl-output` directory

eng/pipelines/apiview-review-gen-cadl.yml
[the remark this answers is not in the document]
scbedd: I see. Ok I'll step out of the way.

eng/pipelines/apiview-review-gen-cadl.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
   - pwsh: |
       npm install -g @cadl-lang/compiler@0.37.0
-      npm install -g @azure-tools/cadl-apiview@0.3.0
+      npm install -g @azure-tools/cadl-apiview@latest
```
tjprescott, line 48: This will probably fail since the latest version of cadl-apiview uses a different version of cadl compiler. It might be sufficient to just install cadl-apiview, which should pull in the supported version of the compiler.

praveenkuttappan:
Fixes https://github.com/Azure/azure-sdk-tools/issues/5038

eng/pipelines/apiview-review-gen-cadl.yml
weshaggard: Just as an FYI @ruowan is working on putting this version as a global setting inside of the specs repo. Once that gets worked out we should try to figure out how to use that same version in this pipeline.

eng/scripts/Create-Apiview-Token-Cadl.ps1
```diff
@@ -31,7 +31,7 @@ function Generate-Apiview-File($packagePath)
     try
     {
         npm install
-        cadl compile . --emit=@azure-tools/cadl-apiview
+        cadl compile . --emit=@azure-tools/cadl-apiview --warn-as-error=false
```
weshaggard, line 34: Can you help me understand more why warnings cause issues? Or are they errors?

tjprescott: They are definitely warnings and output is generated despite the warnings when run locally. But on CI it seems no output is generated, almost as if the CI were treating warnings as errors.

eng/pipelines/apiview-review-gen-cadl.yml
ruowan: Current in PR pipeline, I just use the latest version. Once the global config is done in spec repo, I will make some changes to pick up that global version.

Issue #4975, opened 2022-12-15
CADL review cannot be created manually from pr repo


praveenkuttappan:
Sparse checkout of private repo is failing when creating API review manually for CADL.

maririos:
@praveenkuttappan  is this fixed?

PR #5037, opened 2023-01-03
Support uploading API review token file for CADL review

praveenkuttappan:
CADL review can only be created using a link to CADL repo. This is not sufficient in all cases if CADL source repo has different folder structure or when a user wants to emit token file locally and upload as a revision using different cadl emitter version. This PR is to add an option so user can choose Json token file as source to create a revision.

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
+++ b/src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
@@ -204,11 +204,12 @@ namespace APIViewWeb.Managers
             string name,
             string label,
             Stream fileStream,
+            string language = "",
             bool awaitComputeDiff = false)
         {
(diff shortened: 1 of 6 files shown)


src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
```diff
@@ -204,11 +204,12 @@ public async Task<ReviewModel> GetReviewAsync(ClaimsPrincipal user, string id)
             string name,
             string label,
             Stream fileStream,
+            string language = "",
```
praveenkuttappan, line 207: Input file type will be used when it's empty to find language processor to process input file.

praveenkuttappan:
Fixes https://github.com/Azure/azure-sdk-tools/issues/4975

tjprescott (approved):

PR #5047, opened 2023-01-04
Disable auto gpg signing

### LINKED ISSUE #4720: [Asset-Sync] Git tagging hangs if tag signing enabled
For those of us with GPG signing enabled locally for all commits/tags, the process to push a new tag hangs because a signed tag requires a tag message. So `git tag` will automatically open the user's editor. This causes the outer process to hang indefinitely.

After killing the process, I get:
```
Unhandled exception. Azure.Sdk.Tools.TestProxy.Common.Exceptions.HttpException: Invocation of "git tag python/monitor/azure-monitor-query_6cab4adc27" had a non-zero exit code -1.
StdOut: Vim: Error reading input, exiting...
Vim: Finished.

StdErr: Vim: Warning: Output is not to a terminal

error: There was a problem with the editor 'vim'.
Please supply the message using either -m or -F option.
```

Some workarounds/solutions could be to:

- Include a default `-m` message with the `git tag` command so the editor does not open.
- Include the `-c tag.gpgsign=false` parameter with the `git tag` command. To get around this issue, I had to `git config --global tag.gpgsign false`

scbedd:
Resolves #4720


hallipr (approved):
looks good

PR #5029, opened 2022-12-28
Update `<TargetFramework>netstandard2.0` to `net6.0` in `Azure.ClientSdk.Analyzers` and `CodeOwnersParser`

konrad-jamrozik:
Contributes to:

- #4888 

By updating all projects whose `TargetFramework` was `nestandard2.0` to `net6.0`.

Notably it does leftover update to `CodeOwnersParser`, which was missed by this PR:

- #4937

Note that this PR might require a follow-up package update, same as it was done b this PR:

- #4988

For a list of all PRs doing migration to `net6.0`, see this comment:
- https://github.com/Azure/azure-sdk-tools/issues/4888#issuecomment-1356639653

## Tests done

- `APIView` build, the only tool iun `azure-sdk-tools` that depends on `Azure.ClientSdk.Analyzers`:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2083442&view=results

PR #5048, opened 2023-01-04
Revert `Azure.ClientSdk.Analyzers` library to target `netstandard2.0` instead of `net6.0`.

konrad-jamrozik:
**UPDATE** 1/13/2023: this PR is required as it broke things. See:
- https://github.com/Azure/azure-sdk-tools/pull/5048#issuecomment-1382298227
- https://github.com/Azure/azure-sdk-tools/pull/5029/files#r1069929465

**UPDATE** 1/4/2023: I decided to abandon this PR. See [my comment below](https://github.com/Azure/azure-sdk-tools/pull/5048#issuecomment-1371768721) why.

This PR reverts changes in:

- #5029

plus adds a comment why.

### DIFF
--- a/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.csproj
+++ b/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.csproj
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <!-- This library is fixed to netstandard2.0 as we target the lowest common framework for shared libraries 
-    to allow them to be used in the widest possible set of applications. -->
+    <!-- This library is fixed to netstandard2.0 due to:
+    https://github.com/Azure/azure-sdk-tools/pull/5048#issuecomment-1382298227
+    -->
     <TargetFramework>netstandard2.0</TargetFramework>
     <BuildOutputTargetFolder>analyzers/dotnet/cs/</BuildOutputTargetFolder>
     <VersionPrefix>0.1.1</VersionPrefix>
--- a/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.csproj
+++ b/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.csproj
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <!-- This library is fixed to netstandard2.0 due to:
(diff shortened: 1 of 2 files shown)


konrad-jamrozik:
@weshaggard on second thought, I think we should keep it to `net6.0`. I've read more about it.

`netstandard` is obsolete, replaced by .net6.0: https://learn.microsoft.com/en-us/dotnet/standard/net-standard

Per this [blog post](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/):

> .NET 5 and all future versions will always support .NET Standard 2.1 and earlier. The only reason to retarget from .NET Standard to .NET 5 is to gain access to more runtime features, language features, or APIs. So, you can think of .NET 5 as .NET Standard vNext.

weshaggard:
In general, if we have all the APIs we need in .NET Standard 2.0 it is generally recommended to target that for libraries. I agree for these libraries we likely don't need to worry about this wider reach but all things being equal I still tend to recommend targeting .NET Standard 2.0 unless there are APIs in later versions you need to consume.

If we haven't released a package yet with the new .NET 6.0 target then I'd still move these back to .NET Standard 2.0 but if we have already released and started consuming a new package with .NET 6.0 then I can accept sticking with that as long as we don't find a use case where we need to target older frameworks.

konrad-jamrozik:
> In general, if we have all the APIs we need in .NET Standard 2.0 it is generally recommended to target that for libraries. I agree for these libraries we likely don't need to worry about this wider reach but all things being equal I still tend to recommend targeting .NET Standard 2.0 unless there are APIs in later versions you need to consume.
> 
> If we haven't released a package yet with the new .NET 6.0 target then I'd still move these back to .NET Standard 2.0 but if we have already released and started consuming a new package with .NET 6.0 then I can accept sticking with that as long as we don't find a use case where we need to target older frameworks.

@weshaggard both of the packages have been published already:

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2083477&view=results

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2083478&view=results

I found one case where move to `net6.0` makes things a little bit cleaner:

``` csharp
.FirstOrDefault(
    entry => Matches(targetPath, entry), 
    // assert: none of the codeownersEntries matched targetPath
    new CodeOwnerEntry());
```

When reverted to `netstandard2.0` I get:

image

So I will keep it to `net6.0` per you being OK with it given the packages were published. However, of course I will remember your general guidance going forward. Sorry again for the too-aggressive update!

heaths:
This should probably be reopened, per internal thread, since it's breaking track 2 SDK builds. Our SDKs will "always" target netstandard2.0 at a minimum. We have no plans at this time to upgrade as there's really no need, especially now that we'll start multi-targeting as needed, starting with `netstandard2.0;net461` by default for build-time benefits. Some libraries like Key Vault will be taking advantage of this to work around the few "lies" that net461 supports netstandard2.0, of which the majority of those few APIs not in netstandard2.0 are in `System.Security.Cryptography` that the KV libraries use.

weshaggard (approved):

heaths (approved):

PR #5050, opened 2023-01-04
Resolve Paths within Linux Executable

scbedd:
### Problem

The `test-proxy-standalone-linux-*` binaries are of the form:

```
<unzip root>/
  tools/
    test-proxy/
       linux-*/
         <executable>
```    

When `mac` and `windows` look like:

```
<unzip root>/
  <executable>
```    

We should make certain these are identical.

### Discussion

This was happening for all packages that were building using `produce-standalone-executables.yml`.

Effectively, when you tar a target file that is BELOW your current working directory, it'll include the relative path in the tar file itself.

`tar -czvf blah.tar.gz tools/test-proxy/linux-x64`

Will result in a zip folder with the actual relpath `tools/test-proxy/linux-x64` still present within. This is obviously not super helpful for our linux users.

The traditional way I'd solve this is to make use of the `-C <change directory>` option of tar. However, that ALSO doesn't work because of  bugs handling how to move the newly created tar.gz to the target folder. Instead, just cd into the directory we need to tar. 

This is what using the -C argument looked like.

```
semick@mars:/repo/azure-sdk-tools/tools/test-proxy$ tar -cvzf -C "tools/test-proxy/linux-x64" "/repo/artifacts/test-prox
y-standalone-linux-x64.tar.gz"
tools/test-proxy/linux-x64/
tools/test-proxy/linux-x64/test-proxy.xml
tools/test-proxy/linux-x64/appsettings.json
tools/test-proxy/linux-x64/test-proxy
tools/test-proxy/linux-x64/test-proxy.pdb
tools/test-proxy/linux-x64/appsettings.Development.json
tar: Removing leading `/' from member names
tar: /repo/artifacts/test-proxy-standalone-linux-x64.tar.gz: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
```

I got the same _type_ of error when I tried:
- `tar -cvzf -C "tools/test-proxy/linux-x64" "/repo/artifacts/test-prox
y-standalone-linux-x64.tar.gz"`
- `tar -cvzf -C "tools/test-proxy/linux-x64" "/repo/artifacts/test-prox
y-standalone-linux-x64.tar.gz ."`

However, when I CD into the target directory, and zip `.`, that same target tar.gz works just fine. Not burning any more time into this.


hallipr (approved):
looks good

danieljurek (approved):

PR #5063, opened 2023-01-05
Replace the glob-based CODEOWNERS parser with regex-based

konrad-jamrozik:
This PR is a follow-up to:

- #5030

It rewrites the new parser logic to match the rules as outlined in:
- https://github.com/Azure/azure-sdk-tools/issues/4859#issuecomment-1370360622

The new parser now supports diagnostic logging via `Microsoft.Extensions.Logging.Console`. To make this work, I took a dependency on this package, version `7.0.0`, and updated that package and `Microsoft.Extensions.Logging` to `7.0.0` in the code that uses `CodeOwnersParser` as dependency, to avoid the "package version downgrade detected" build failure.

### DIFF
--- a/tools/code-owners-parser/CodeOwnersParser/Azure.Sdk.Tools.CodeOwnersParser.csproj
+++ b/tools/code-owners-parser/CodeOwnersParser/Azure.Sdk.Tools.CodeOwnersParser.csproj
@@ -7,7 +7,6 @@
 
   <ItemGroup>
     <PackageReference Include="CommandLine.Net" Version="2.3.0" />
-    <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" Version="7.0.0" />
   </ItemGroup>
 
 </Project>
--- a/tools/code-owners-parser/CodeOwnersParser/Azure.Sdk.Tools.CodeOwnersParser.csproj
+++ b/tools/code-owners-parser/CodeOwnersParser/Azure.Sdk.Tools.CodeOwnersParser.csproj
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="CommandLine.Net" Version="2.3.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
   </ItemGroup>
 
 </Project>
--- a/tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
+++ b/tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
@@ -1,7 +1,6 @@
-using System;
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.Extensions.FileSystemGlobbing;
+using System.Text.RegularExpressions;
 
 namespace Azure.Sdk.Tools.CodeOwnersParser
 {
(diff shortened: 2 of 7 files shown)


tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
```diff
@@ -1,20 +1,46 @@
-using System;
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.Extensions.FileSystemGlobbing;
```
weshaggard, line 4: As part of this change I assume you can remove the package reference as well https://github.com/Azure/azure-sdk-tools/pull/5030/files#diff-c80850aeaec1dd2efd7977c09fd3a8393b8482a1ecade588e30a23eab4349190R10?

tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
```diff
@@ -40,145 +64,104 @@ public MatchedCodeOwnerEntry(List<CodeOwnerEntry> entries, string targetPath)
...
-            var dir = new InMemoryDirectoryInfo(
-                // This 'rootDir: "/"' is used here only because the globMatcher API requires it.
-                rootDir: "/", 
-                files: new List<string> { targetPath });
-
-            var patternMatchingResult = globMatcher.Execute(dir);
-            return patternMatchingResult;
+            pattern = pattern.Replace("_DOUBLE_STAR_/", "(.*)");
+            pattern = pattern.Replace("/_DOUBLE_STAR_", "(.*)");
+            pattern = pattern.Replace("_DOUBLE_STAR_", "(.*)");
```
weshaggard, line 162: What about `/_DOUBLE_STAR_/`? Do we need to preserve any `/` in this double star case?

tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
```diff
@@ -40,145 +64,104 @@ public MatchedCodeOwnerEntry(List<CodeOwnerEntry> entries, string targetPath)
...
+
+            // Kind of hoping here that the CODEOWNERS path will never have
+            // "_DOUBLE_STAR_" or "_SINGLE_STAR_" strings in it.
+            pattern = pattern.Replace("**", "_DOUBLE_STAR_");
+            pattern = pattern.Replace("*", "_SINGLE_STAR_");
+
+            pattern = Regex.Escape(pattern);
+
+            // Denote that all paths are absolute by prepending "beginning of string" symbol.
+            pattern = "^" + pattern;
```
weshaggard, line 146: Do we need to similarly mark an end `$`?

weshaggard: Never mind I see you did below.

tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
```diff
@@ -40,145 +64,104 @@ public MatchedCodeOwnerEntry(List<CodeOwnerEntry> entries, string targetPath)
...
-                patternsToInclude.Add(ConvertDirectorySuffix(codeownersPath));
-            }
-            else
-            {
-                patternsToInclude.Add(ConvertDirectorySuffix(codeownersPath + "/"));
-                patternsToInclude.Add(codeownersPath);
-            }
-
-            return patternsToInclude;
+            // If the considered CODEOWNERS path ends with "/", it means we can
```
weshaggard, line 105: I'm not sure I'm following this logic. If we do want to allow cases here would it be better handled with the regex pattern as opposed to altering the target path we are checking?

weshaggard (approved):
Except for removing the PackageReference I think we can move my other questions to your main PR. So after removing that reference go ahead and merge this.

tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
konrad-jamrozik: We will preserve one `/` because `/_DOUBLE_STAR_/` case will be replaced by `_DOUBLE_STAR_/` case, preserving the prefix `/`. I [added comments explaining that](https://github.com/Azure/azure-sdk-tools/blob/5aaa156e0eaaecdda673ae7663262cf7a921ad2f/tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs#L166-L172). 

Your remark also made me realize [I was missing a test case](https://github.com/Azure/azure-sdk-tools/blob/5aaa156e0eaaecdda673ae7663262cf7a921ad2f/tools/code-owners-parser/Azure.Sdk.Tools.CodeOwnersParser.Tests/CodeOwnersFileTests.cs#L99) for the `_DOUBLE_START` case. Thanks!

tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs
konrad-jamrozik: Replaced with [regex manipulation](https://github.com/Azure/azure-sdk-tools/blob/2fd99b4ba2f6ceb30cff343ebb61344931b980d4/tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs#L137-L159) instead.

tools/code-owners-parser/CodeOwnersParser/Azure.Sdk.Tools.CodeOwnersParser.csproj
```diff
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="CommandLine.Net" Version="2.3.0" />
-    <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
```
weshaggard, line 10: What do we gain by depending on this logging package? Is it worth having this extra package dependency?

konrad-jamrozik: @weshaggard if I remove it, [I cannot use the logger, here](https://github.com/Azure/azure-sdk-tools/blob/121207b09156a63f2274b7cd0d0a10ba238688c2/tools/code-owners-parser/CodeOwnersParser/MatchedCodeOwnerEntry.cs#L55-L67):

image

I used the logger instead of doing `Console.Out.WriteLine`, because I decided to follow the convention established by existing code, as seen in [notification-configuration Program.cs](https://github.com/Azure/azure-sdk-tools/blob/121207b09156a63f2274b7cd0d0a10ba238688c2/tools/notification-configuration/notification-creator/Program.cs#L42-L46), and thus have uniform logging approach.

Does that seem reasonable?

weshaggard: I don't know anything about this particular package so I'm not sure what benefits we gain but in general the benefits should be enough to warrant taking an extra dependency (even if that is a MS extension package).

konrad-jamrozik: @weshaggard I removed this dependency in:
- #5134

PR #5083, opened 2023-01-07
[Perf] Refactor tests.yml into one file per package per language

mikeharder:
## TODO
- Improve LanguageVersion
  - How are defaults selected
  - Report requested and actual versions, verify versions match

## Testing
- Manually run storage tests in all languages
- Run pipelines in all languages


tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Models/TestInfo.cs
```diff
@@ -5,7 +5,7 @@ namespace Azure.Sdk.Tools.PerfAutomation.Models
     public class TestInfo
     {
         public string Test { get; set; }
+        public string Class { get; set; }
```
heaths, line 8: This is misleading. How about "fixture"? A common term across numerous languages' common test runners/frameworks.

heaths:
I'm concerned about the burden - not just time, but being prone to errors of omission - this puts on developers. Across all languages, we already made a decision of what scenarios to add. Why can't you just run the perf tests matching some repo-defined pattern? There's zero coordination across repos at that point that hasn't been done already.

What is the point of this extra file? To unify scenario names to compare? If so, certainly there's a better way that should work across languages, like emitting a specific pattern of test name or whatever SDKs decide to unify on e.g., `get-secret`. I know NUnit, at least, lets us control our test names and I imagine other languages should have something similar, or at least a way to transform output before being consumed by your reporting tool.

This extra file means that any time we change scenarios - adds, edits, or removes - we also have to remember to update this file. It seems like another hurdle to onboarding developers.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Models/TestInfo.cs
mikeharder: I don't think `Class` is particularly misleading, since a perf test is represented by a class in each language.  But I'm fine with `Fixture` as well.

@AlexanderSher, @g2vinay, @HarshaNalluru, @timovv, @annatisch, @pvaneck, @gearama: Any opinions?

jsquire:
To Heath's point - are we not allowing some kind of wildcard to say "run all the tests" and using that as a default?   I like the option to narrow things down as an opt-in, if we need it for special cases, but I agree that it would be burdensome to have to remember to keep another asset in sync for each change that someone makes.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Models/TestInfo.cs
HarshaNalluru: `Fixture` sounds good, don't really mind either since this won't flow into the language repos anyway.

mikeharder: I chatted with @heaths, he is OK with `Class` since each perf test is represented by a class.  He was thinking each perf test was represented by a method (like unit tests).

mikeharder: > `Fixture` sounds good, don't really mind either since this won't flow into the language repos anyway.

This change would flow into the language repos, since it would also rename from `Class` to `Fixture` in all the `perf-tests.yml` files.

HarshaNalluru: woah, ok

mikeharder:
@heaths, @jsquire, @christothes:

This set of PRs is a huge improvement over the status quo, where the metadata for all tests across all languages lives in a single file in the tools repo:

https://github.com/Azure/azure-sdk-tools/blob/de045afa1901dd6b6a3d43a01bec1f3d9fb4f4f0/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/tests.yml
 
These PRs move this metadata to the language+service directories, which allows updating the perf tests in a single PR to the language repo, rather than also requiring a PR to the tools repo.  It also allows languages to additional tests or arguments, beyond the standard set we keep consistent across all languages.

The perf v-team has been waiting months for this refactoring.  So I would like to land these PRs with the current design, and we can discuss further improvements after these are merged.

The `perf-tests.yml` file contains all the metadata needed for the perf pipeline (using the PerfAutomation app) to know which tests to run with which arguments.  Specifically:

- The language-agnostic name for the service (e.g. `storage-blob`)
- The language-specific project containing the perf tests for this service (e.g. `sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Azure.Storage.Blobs.Perf.csproj`)
- The sets of package versions to run the perf tests against (e.g. "latest GA" and "sources")
- The language-agnostic name for each test (e.g. `download`)
- The language-specific class representing this test (e.g. `DownloadBlob`)
- The arguments to pass to each test (e.g. list of blob sizes)

It may be possible to move some or most of this information into the perf test source code itself.  For example, in .NET, the list of arguments could be expressed as a set of custom attributes on the perf test class.  And I can see some advantages to this.

However, one of the principles we have been using for our performance infrastructure, is to keep the code specific to each language as simple as possible, and move the complexity into the language-agnostic PerfAutomation app.  Anything language-specific needs to be implemented 5 times (.NET, Java, JS, Python, C++), and kept consistent across the languages.  So we have tried to keep the language-specific perf frameworks as simple and lightweight as possible, and encoding this metadata in `perf-tests.yml` rather than in the language-specific perf source code allows this to happen.

If .NET would like to prototype an end-to-end solution that moves this metatadata from `perf-tests.yml` into the perf source code, I am happy to review.  The next question would be if all the other languages want to implement something similar, or continue using `perf-tests.yml` (so .NET would be an outlier).

My personal opinion is the current `perf-tests.yml` is the best solution overall, since it minimizes the amount of work we need to do in each language.  But if the language teams feel differently, and are willing to fund the language-specific work to change this, I am open to the idea.  However, I will also say if the language teams want to spend more time on perf, I think there are higher-value activities than moving this metadata to a different place.

jsquire:
I think the main concern here for me is the number of things that a performance test author has to remember to update and the disconnect for understanding why the new test that you wrote wasn't executed.   I'd prefer to see us remove the need to have metadata present to run the test (run the mall by default) and, instead, let the dashboard show results that don't align if it's not set.   

That way, it's clear that a test is running and that "something" needs to happen for it to be grouped with its peers from other languages which can help drive the necessary follow-up conversations to discover what steps are needed.

g2vinay (approved):

g2vinay:
The feedback provided from .NET folks above around the design is helpful and can be factored in next iteration of perf infra.

heaths:
I've set up a meeting with @jsquire and @pallavit to discuss next steps, if any. I'm thinking cross-language and appreciate @mikeharder's design ideas we discussed offline. I think there's relatively low-cost ways to achieve the same thing without complicating the onboarding process for new contributors to maintain yet another document, since they already have to write code to add tests. 

The value of this PR over the existing process is not in question: clearly this is a vast improvement. But the less maintenance hassles in the long run the better. Ideally, I think the runners can effectively multiplex the different permutations into separate processes (this is how .NET test runners generally work, and even multi-targeting builds that we already use for testing). That may be harder on some languages than others, granted.

But I also empathize with Mike that this is something we do relatively infrequently. Still, if we make it a docs issue...well, I've already seen more than enough cases where we have docs - somewhere - and people still go outside those norms like we've seen recently with tests/samples/snippets for .NET. More automation and less manual effort tends to help.

PR #5085, opened 2023-01-09
Update versions of packages `RetrieveCodeOwners`, `NotificationConfiguration` and `PipelineOwnersExtractor` to most recent as of 1/8/2023

konrad-jamrozik:
code-owners-parser:
- https://dev.azure.com/azure-sdk/internal/_build?definitionId=3188
- https://dev.azure.com/azure-sdk/public/_artifacts/feed/azure-sdk-for-net/NuGet/Azure.Sdk.Tools.RetrieveCodeOwners/overview/1.0.0-dev.20230108.6

notification-configurator:
- https://dev.azure.com/azure-sdk/internal/_build?definitionId=1815
- https://dev.azure.com/azure-sdk/public/_artifacts/feed/azure-sdk-for-net/NuGet/Azure.Sdk.Tools.NotificationConfiguration/overview/1.0.0-dev.20230108.1

pipeline-owners-extractor:
- https://dev.azure.com/azure-sdk/internal/_build?definitionId=5109
- https://dev.azure.com/azure-sdk/public/_artifacts/feed/azure-sdk-for-net/NuGet/Azure.Sdk.Tools.PipelineOwnersExtractor/overview/1.0.0-dev.20230105.1

This PR should unblock currently failing pipeline of [automation - build-failure-notification-subscriptions](https://dev.azure.com/azure-sdk/internal/_build?definitionId=679&_a=summary).

PR #5088, opened 2023-01-09
Enable the new, regex-based, wildcard-supporting CODEOWNERS matcher

konrad-jamrozik:
This PR enables the parser introduced in:

- #5063

And as such, this PR contributes to:

- #2770

Before a package published out of this PR is depended upon, a review of affected `CODEOWNERS` files is necessary, as explained in `Behavior change of existing CODEOWNERS paths with wildcards` section of this comment:
- https://github.com/Azure/azure-sdk-tools/issues/2770#issuecomment-1338923752

Full behavior difference between the old and new parser can be determined by comparing the two rightmost columns of this [test battery](https://github.com/Azure/azure-sdk-tools/blob/61a1d763b704d3d787c667a968fbbd6bb089c9fb/tools/code-owners-parser/Azure.Sdk.Tools.CodeOwnersParser.Tests/CodeOwnersFileTests.cs#L29-L132).

This parser will be executed by the [`automation - build-failure-notification-subscriptions`](https://dev.azure.com/azure-sdk/internal/_build?definitionId=679) pipeline, once the package it depends on is published and updated, similar as it is done in this PR:
- #5085

As a result, the following `CODEOWNERS` files are expected to be affected, and thus require a review:

- https://github.com/Azure/azure-dev/blob/main/.github/CODEOWNERS
- https://github.com/Azure/azure-sdk-for-c/blob/main/.github/CODEOWNERS
- https://github.com/Azure/azure-sdk-for-cpp/blob/main/.github/CODEOWNERS
- https://github.com/Azure/azure-sdk-for-go/blob/main/.github/CODEOWNERS
- https://github.com/Azure/azure-sdk-for-java/blob/main/.github/CODEOWNERS
- https://github.com/Azure/azure-sdk-for-js/blob/main/.github/CODEOWNERS
- https://github.com/Azure/azure-sdk-for-net/blob/main/.github/CODEOWNERS
- https://github.com/Azure/azure-sdk-for-python/blob/main/.github/CODEOWNERS

Following files won't be affected, as they are not triggered by the pipeline, whose source is [`notifications.yml`](https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/notifications.yml):

- https://github.com/Azure/azure-sdk-for-android/blob/main/.github/CODEOWNERS
- https://github.com/Azure/azure-sdk-for-ios/blob/main/.github/CODEOWNERS
- https://github.com/Azure/azure-sdk-tools/blob/main/.github/CODEOWNERS

In addition, following work needs to be implemented and applied to the CODEOWNERS files before we can enable support for wildcards:

- #5181

### Related work

- #4859

### DIFF
--- a/tools/code-owners-parser/CodeOwnersParser/CodeownersFile.cs
+++ b/tools/code-owners-parser/CodeOwnersParser/CodeownersFile.cs
@@ -8,7 +8,7 @@ namespace Azure.Sdk.Tools.CodeOwnersParser
 {
     public static class CodeownersFile
     {
-        private const bool UseRegexMatcherDefault = false;
+        private const bool UseRegexMatcherDefault = true;
 
         public static List<CodeownersEntry> GetCodeownersEntriesFromFileOrUrl(
             string codeownersFilePathOrUrl)


konrad-jamrozik:
@weshaggard as I mentioned in this PR description - a review of CODEOWNERS file is necessary, to see how support of wildcards changes the owners. Could you provide me with guidance what algorithm I should apply to updating the files? 

Should I reach out first to all the folks who will now start getting ADO notifications for build failures because of wildcards now matching, and ask them if they want to opt out before I deploy this change?

Or maybe I should comment out all wildcard-supporting paths and ask the assigned owners to let me know if I should uncomment them? 

Due to wildcards now matching possibly some set of people will stop getting notifications, but I think figuring this set out would be a lot of work, so I am not sure if I should try to do it.

Thanks!

weshaggard:
My biggest concern is the **/ci.yml and **/test.yml patterns that are setup at the bottom of the files which point to engsys folks. That is going to cause all pipeline notifications to come to those folks instead of the service teams which is not what we want to happen. 

I would suggest doing a test where you look at the owners of all the *.yml files in the repo before and after your changes and compare the results to see what changed. 

Might consider that for any other patterns you see in the codeowners where ** patterns exist.

weshaggard (changes requested):
I don't want this merged until we verify the yml file mappings as I believe this will break all our notification mappings.

konrad-jamrozik:
> My biggest concern is the **/ci.yml and **/test.yml patterns that are setup at the bottom of the files which point to engsys folks. That is going to cause all pipeline notifications to come to those folks instead of the service teams which is not what we want to happen. I suggest we simply remove those patterns from the Codeowners files.
> 
> I would suggest doing a test where you look at the owners of all the *.yml files in the repo before and after your changes and compare the results to see what changed.
> 
> Might consider that for any other patterns you see in the codeowners where ** patterns exist.

@weshaggard OK, cool. I am going to make all the tests as you proposed. 

My plan is to augment the [`retrieve-codeowners` tool](https://github.com/Azure/azure-sdk-tools/blob/main/tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners/Program.cs) (which is wrapped by [`get-codeowners.ps1` script](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/get-codeowners.ps1)) to allow it to not only target specific directory to check its ownership, but to be able to target a path with wildcards, find all concrete paths matching that path-with-wildcards, and return ownership information for all of them.

I will then run this tool with the old and new parser and report here the ownership diff here.

konrad-jamrozik:
I have confirmed that commenting out the following lines in `CODEOWNERS` for `azure-sdk-for-net`:

```
#/**/*Management*/                               @fengzhou-msft @archerzz @ArcturusZhang @Yao725 @ArthurMa1978
#/**/Azure.ResourceManager*/                     @fengzhou-msft @archerzz @ArcturusZhang @Yao725 @ArthurMa1978

#/**/tests.yml   @hallipr @benbp
#/**/ci.yml      @hallipr @benbp
```

results in exactly the same owners being matched using the new regex matcher as in the obsolete prefix-based matcher.

I'll add more details here and link to the tool I used a bit later.

weshaggard:
I think it is fine to remove the **/*.yml entries for engsys but we should talk with the mgmt folks about those before we remove them.

weshaggard:
With our new approach we might also want to consider removing the /sdk/ pattern which was all about having the leads be the owners so they can push to get any missing owners figured out but once we get our pipeline setup we can have the pipeline identify new code paths that don't have owners.

konrad-jamrozik:
> Looks like that diff still has the **/ci.yml and **/tests.yml paths in the codeowners file going to EngSys I know that is definitely going to be wrong so I would like to see a diff with those patterns removed.

@weshaggard I added the diff you requested [to the Excel diff]((https://microsoft-my.sharepoint.com/:x:/p/kojamroz/EXDbCuNkC3RGu86BtoqcR9wBHHQ1nikCWb2QPukAG44E9A?e=VV9iPd)), in the tab named `azure-sdk-for-net_owners_diff_2`.

> With our new approach we might also want to consider removing the /sdk/ pattern

You mean only the the line with exactly `/sdk/`, yes? But not paths like `/sdk/objectanchors/`?

Here is my thinking:
- If that diff looks OK to you, I will remove the `**/ci.yml` and `**/tests.yml` paths from the `azure-sdk-for-net` and then enable the the matcher for `azure-sdk-for-net` repo **only**, keeping the old matcher for other repos (I will hardcode some rules in it).
- After some cooking time I will then repeat this process for each of our other language repos.

Re the removal of `/sdk/` I see following rules matching to it in the diff:

image

So, in a separate PR, I could remove `/sdk/` and add a bunch of rules like `/sdk/maps/`, `/sdk/quantum/` etc., with the same owners assigned. Then let the two owners, `pallavit` and `jsquire` know, to determine exact owners for these paths.

Thoughts?

konrad-jamrozik:
@weshaggard I have updated all the `CODEOWNERS` files in scope of the new matcher. You can see what I did by reviewing the PRs I linked to this issue. There are still few PRs open, as can be seen above. But as soon as I get your approval on them, I will merge them, and then we can proceed with merging this PR.

benbp (approved):

weshaggard (approved):

PR #5096, opened 2023-01-09
[pylint] decorated tracing

l0lawrence:
`    @distributed_trace
    async def create_or_update_alias(self, alias, **kwargs):

    @distributed_trace
    async def create_alias(self, alias, **kwargs):
        # type: (SearchIndex, **Any) -> SearchAlias`


tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -372,14 +374,19 @@ def visit_functiondef(self, node):
...
         try:
-            if node.parent.name.endswith("Client") and node.is_method() and not node.name.startswith("_") and \
-                    node.parent.name not in self.ignore_clients:
-                if node.args.kwarg and "azure.core.tracing.decorator.distributed_trace" not in node.decoratornames() \
-                        and "builtins.classmethod" not in node.decoratornames():
-                    self.add_message(
-                        msgid="client-method-missing-tracing-decorator", node=node, confidence=None
-                    )
-        except AttributeError:
+            path = node.root().name
```
kristapratico, line 377: (un-related line) Do you think maybe in the description of the error we should say something about how this only applies to methods that make network calls? The scenario I'm thinking of is when we have a false positive and someone reads the error and adds it because the rule says so

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2060,7 +2075,7 @@ def register(linter):
     # Rules are disabled until false positive rate improved
     # linter.register_checker(CheckForPolicyUse(linter))
     # linter.register_checker(ClientHasApprovedMethodNamePrefix(linter))
-    # linter.register_checker(ClientMethodsHaveTracingDecorators(linter))
+
```
kristapratico, line 2078: reminder for tests! :)

l0lawrence: there are tests already created (line 17 of test custom plugins)

kristapratico (approved):

PR #5097, opened 2023-01-09
[pylint] type property name too long

l0lawrence:
Names that would be flagged at current length of 40:

return_headers_and_deserialized_path_list
get_deleted_path_properties_from_generated_code
build_get_incidents_by_anomaly_detection_configuration_request
build_delete_datasource_credential_request
obtain_token_by_refresh_token_on_behalf_of
build_get_anomalies_by_anomaly_detection_configuration_request
class AnalyzeHealthcareEntitiesLROPollingMethod


kristapratico:
> Names that would be flagged at current length of 40:
> 
> return_headers_and_deserialized_path_list get_deleted_path_properties_from_generated_code _list_metric_enriched_series_data_requests _list_anomalies_for_detection_configuration _list_anomalies_for_detection_configuration_requests build_get_incidents_by_anomaly_detection_configuration_request build_delete_datasource_credential_request obtain_token_by_refresh_token_on_behalf_of _get_accounts_having_matching_refresh_tokens _infer_max_workers_greater_than_one_if_needed _validate_topic_subscription_and_rule_types

I think we can probably ignore any names that are prefixed with `_`

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2028,6 +2028,71 @@ def _check_import(self, name, node):
                 msgid="no-raise-with-traceback", node=node, confidence=None
             )
 
+class TypePropertyNameTooLong(BaseChecker):
```
kristapratico, line 2031: Will this also capture if a class var or instance attr is too long? I couldn't tell from the code so asking :)

l0lawrence: Yes it will! I can change the checker name to make it clearer?

tools/pylint-extensions/pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
```diff
@@ -3136,4 +3136,84 @@ def test_raise_traceback(self):
                     end_col_offset=96
                 )
         ):
-            self.checker.visit_importfrom(node)
\ No newline at end of file
+            self.checker.visit_importfrom(node)
+
+class TestTypePropertyNameLength(pylint.testutils.CheckerTestCase):
```
kristapratico, line 3141: can we add two more tests:

test_instance_attr_name_too_long and test_class_var_name_too_long?

tools/pylint-extensions/pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2028,6 +2028,79 @@ def _check_import(self, name, node):
...
+
+    def visit_functiondef(self, node):
+        """Visit every function and check that the function and 
+        its variable names are within the character length limit."""
+        try:
+            self.iterate_through_names(node, False)    
+        except:
+            pass      
+
+    def iterate_through_names(self, node, ignore_function):
```
l0lawrence, line 2065: helper function called by both visit_function and visit_class, so we check that if it is a classdef call to the helper function that we don't double check functions.

kristapratico: what is ignore_function? I can't find where it's defined

l0lawrence: Added it~ `:param ignore_function: Whether the function is being called 
             from `visit_classdef` or not. If it is called from `visit_classdef`
             ignore_function should be `True` to avoid repeat warnings on functions. `

kristapratico (approved):

PR #5106, opened 2023-01-10
Improved internal type detection; Added support for TemplateTemplate arguments.

LarryOsterman:


RickWinter (approved):

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -452,6 +452,22 @@ class AstNamedCastExpr : public AstExpr {
   }
 };
 
+template <typename T, class Function>
+void DumpList(T start, T end, AstDumper* dumper, Function function)
```
antkmsft, line 456: I'd declare `function` as `std::function<void(AstDumper*, decltype(*start)&>)>`:
```cpp
#include <iostream>
#include <vector>

#include <functional>

template<typename T>
void f(T iter, std::function<void(decltype(*iter)&)> fn)
{
    fn(*iter);
}

int main()
{
    int i = 1;
    int* p = &i;
    f(p, [&](auto& val) { val = 2; });
    std::cout << *p << "\n";

    std::vector<int> vec = { 3 };
    f(vec.begin(), [&](auto& val) { val = 4; });
    std::cout << vec[0];
}
```

antkmsft (approved):

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
LarryOsterman: Hmmm. I thought about that, but wasn't quite sure if that gave me enough flexibility. Let me poke around and see what comes out.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/ProcessorImpl.cpp
```diff
@@ -99,9 +115,21 @@ ApiViewProcessorImpl::ApiViewProcessorImpl(
...
-  if (configurationJson.contains("includeInternal"))
+
+  // Let's make sure we understand all the values passed in.
+  for (auto nodeName : configurationJson.items())
+  {
+    if (std::find(KnownSettings.begin(), KnownSettings.end(), nodeName.key())
+        == KnownSettings.end())
+    {
+      std::cerr << "Unknown setting found in ApiViewSettings.json file: " << nodeName.key()
+                << std::endl;
```
RickWinter, line 126: Well clang formatted this one in a special way :)

RickWinter: nvm, that was the browser (facepalm)

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -452,6 +452,34 @@ class AstNamedCastExpr : public AstExpr {
...
 
+template <typename T>
+void DumpList(
+    T start,
+    T end,
+    AstDumper* dumper,
+    DumpNodeOptions dumpOptions,
+    std::function<void(AstDumper*, decltype(*start)&)> function)
+{
+  bool firstArg{true};
```
antkmsft, line 463: I see you exclusively use this (`T t{...}`) notation now. Are there benefits compared to `auto t = ...` or `T t = ...` when used for a generic type T (not talking specifically about boolean; Is it a case similar to `++x` being better than `x++` when it is used for custom iterator types?), or is it just a personal preference? Perf or syntactical benefits? Just asking out of curiosity.

antkmsft (approved):

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
LarryOsterman: Actually the reason I use the initializer syntax (`T t{...}`) (as opposed to assignment syntax (`T t = ...`)) is because of [C++ Core Guidelines C.49](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c49-prefer-initialization-to-assignment-in-constructors), which says we should prefer initialization over assignment in constructors.

The reason I don't like the function call syntax (`T t(...)`) is because it is easily confused with a function call (this can be especially true if you're creating a temporary. As to why that can be a challenge, consider this case: `Foo(x, y, Bar(...));` does the `Bar(...)` code call a function named `Bar` or does it construct a `Bar` type?)

LarryOsterman: Oh, and I also forgot about: [ES.23](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax).

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -452,6 +452,34 @@ class AstNamedCastExpr : public AstExpr {
   }
 };
 
+template <typename T>
+void DumpList(
+    T start,
+    T end,
+    AstDumper* dumper,
+    DumpNodeOptions dumpOptions,
```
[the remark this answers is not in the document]
LarryOsterman: Hmm. I've been annoyed at passing this by value for a while (not 100% sure why I did that, tbh). It was unpleasant but IMHO the code is better passed by const reference.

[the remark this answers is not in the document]
LarryOsterman: Also I found a bunch of missing `overload` directives in the process, so this was a good catch regardless.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
antkmsft: I am not suggesting anything here, just talking.

I think it is ES.23 more than C.49.

I read C.49 as it is about using initializer lists whenever possible, which holds true regardless of variable initialization syntax, i.e.: `C(int i) : m_i(i) {}` is better than `C(int i) { m_i = i; }`.

For variable declarations, `T t = 1` is a syntactic sugar that results in calling `T(int)` and no `operator=`.
But besides that, ES.23 has many good points. What I only don't like is that it is hard to tell whether `size_t i = 1; std::vector<size_t> v{i}` is a call to `vector(size_t count)` constructor, or a call to `vector(initializer_list<size_t> init,
 Allocator const & alloc = Allocator())`. But besides that, in many other places `{}` initialization is better. But still, I agree, if I wanted to pick one initialization syntax that I use universally, I'd also pick `{}` as you do.

LarryOsterman: Yeah - it's more ES.23 than C.49 - C.49 was just the first hit. And the discussion on ES.23 discusses the std::vector case. And as I understand it, the `T t=1` case isn't always syntactic sugar, which can be a problem (it can call `operator=` instead of `T()`  which may have greater overhead (`operator=` could execute nop code to free an existing value, it's also possible that you'll have an unnecessary call to `T()`). 

The reality is that most compilers will absolutely optimize this combination away because it's  such a common pattern, but it's still additional work and additional ambiguity.

antkmsft: It is not up to compiler do decide, when you write `T t = 1`, it is always `T T(1)`, but it won't compile if that constructor is explicit. For the explicit constructor, you can write `T t = T(1)`, but that again will be effectively calling only `explicit T(int)`, and nothing else like `T(T const&)` after that.
https://en.cppreference.com/w/cpp/language/copy_initialization : 
_"The equals sign, =, in copy-initialization of a named variable is not related to the assignment operator. Assignment operator overloads have no effect on copy-initialization."_

That being said, list initialization (`{}`) is more universal.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
antkmsft: I just thought, `std::function<void(AstDumper* dumper, decltype(*start)& element)> doDumpElement` could possibly be a better name if you still have that function around (I can't find it in the latest PR iteration.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -452,6 +452,34 @@ class AstNamedCastExpr : public AstExpr {
   }
 };
 
+template <typename T>
+void DumpList(
+    T start,
+    T end,
+    AstDumper* dumper,
+    DumpNodeOptions const& dumpOptions,
+    std::function<void(AstDumper*, decltype(*start)&)> function)
```
antkmsft, line 461: Is this a better name?
```suggestion
    std::function<void(AstDumper* dumper, decltype(*start)& element)> doDump(List)Element)
```

LarryOsterman: Hmm. I didn't think you could have type names in the std::function template parameter. But it does seem to work.

tools/apiview/parsers/cpp-api-parser/README.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+The `ParseAzureSdkCpp` tool has the following command line switches:
+
+- `-r`, `--review` - Specifies the name of the API review for the API Review tool.
+- `--packageVersion` - Specifies the version of the package for the API Review.
+- `-o`, `--output` - Specifies the output file for the API review.
+- `--version` - Prints the version of the ParseAzureSdkCpp tool.
+- `-h`, `--help` - Prints help text about the tool.
+- `-c`, `--console` - Prints the ApiView output to the console as well as the output JSON file.
```
RickWinter, line 29: Future improvement:
Display this information from tool via '-?" option.

LarryOsterman: I took this information from the output of the tool :).

RickWinter (approved):

PR #5109, opened 2023-01-10
[pylint] async client list methods should return AsyncItemPaged

l0lawrence:


l0lawrence:
Holding off on this check for this version of the checker

PR #5113, opened 2023-01-11
Update on existing metadata json file.

sima-zhu:
The PR is to update the existing metadata json instead of check in a new one.

The updated metadata json: https://github.com/Azure/azure-docs-sdk-dotnet/commit/2af2828fb2694080520f1e6b6e0fd465d226985e#diff-afa38ac4d7e2700dd78b4ca6547bd0cab8fbff7794dd803d18600eb8783bd1dc

Testing pipeline: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2115645&view=logs&j=72a39880-3efd-52c2-15be-d04db16925c9

### DIFF
--- a/eng/common/scripts/Update-DocsMsMetadata.ps1
+++ b/eng/common/scripts/Update-DocsMsMetadata.ps1
@@ -49,10 +49,10 @@ param(
   [Parameter(Mandatory = $true)]
   [string]$DocRepoLocation,
 
-  [Parameter(Mandatory = $true)]
+  [Parameter(Mandatory = $false)]
   [string]$Language,
 
-  [Parameter(Mandatory = $true)]
+  [Parameter(Mandatory = $false)]
   [string]$RepoId,
 
   [Parameter(Mandatory = $false)]
@@ -192,8 +192,20 @@ function UpdateDocsMsMetadataForPackage($packageInfoJsonLocation) {
   }
   $packageMetadataName = Split-Path $packageInfoJsonLocation -Leaf
   $packageInfoLocation = Join-Path $DocRepoLocation "metadata/$metadataMoniker"
+  if (Test-Path "$packageInfoLocation/$packageMetadataName") {
+    Write-Host "The docs metadata json $packageMetadataName exists, updating..."
+    $docsMetadata = Get-Content "$packageInfoLocation/$packageMetadataName" -Raw | ConvertFrom-Json
+    foreach ($property in $docsMetadata.PSObject.Properties) {
+      if ($packageInfo.PSObject.Properties.Name -notcontains $property.Name) {
+        $packageInfo | Add-Member -MemberType $property.MemberType -Name $property.Name -Value $property.Value -Force
+      }
+    }
+  }
+  else {
+    Write-Host "The docs metadata json $packageMetadataName does not exist, creating a new one to docs repo..."
+    New-Item -ItemType Directory -Path $packageInfoLocation -Force
+  }
   $packageInfoJson = ConvertTo-Json $packageInfo
-  New-Item -ItemType Directory -Path $packageInfoLocation -Force
   Set-Content `
     -Path $packageInfoLocation/$packageMetadataName `
     -Value $packageInfoJson


eng/common/scripts/Update-DocsMsMetadata.ps1
```diff
@@ -192,8 +192,17 @@ function UpdateDocsMsMetadataForPackage($packageInfoJsonLocation) {
   }
   $packageMetadataName = Split-Path $packageInfoJsonLocation -Leaf
   $packageInfoLocation = Join-Path $DocRepoLocation "metadata/$metadataMoniker"
-  $packageInfoJson = ConvertTo-Json $packageInfo
-  New-Item -ItemType Directory -Path $packageInfoLocation -Force
+  $docsMetadata = $packageInfo
+  if (Test-Path $packageInfoLocation) {
+    $docsMetadata = Get-Content "$packageInfoLocation/$packageMetadataName" -Raw | ConvertFrom-Json
+    $docsMetadata.PSObject.Properties | ForEach-Object {
+      $packageInfo | Add-Member -MemberType $_.MemberType -Name $_.Name -Value $_.Value -Force
```
weshaggard, line 199: Does Add-Member work if we are updating a property?

weshaggard: It looks like we are overriding the values with the docsMetadata from the docs repo I would have expected the other direction.

weshaggard, line 199:
```suggestion
    $docsMetadata.PSObject.Properties | ForEach-Object {
      // If Property doesn't exist on packageInfo otherwise leave it alone
      $packageInfo | Add-Member -MemberType $_.MemberType -Name $_.Name -Value $_.Value -Force
```

sima-zhu: Now logic goes: (Docs metadata means the json file from docs repo, package json means the json file from release pipeline.)
1. For each docs metadata property:
  a. If package json has the property, then do nothing .
  b. If package json doesn't have the property, then update to package json.

2. Then write package json back to docs repo.

eng/common/pipelines/templates/steps/update-docsms-metadata.yml
```diff
@@ -108,10 +109,11 @@ steps:
 
     - template: /eng/common/pipelines/templates/steps/git-push-changes.yml
       parameters:
-        BaseRepoBranch: $(TargetBranchName)
+        BaseRepoBranch: test_metadta
```
weshaggard, line 112: I assume this will be reverted before merging?

eng/common/scripts/Update-DocsMsMetadata.ps1
```diff
@@ -49,10 +49,10 @@ param(
   [Parameter(Mandatory = $true)]
   [string]$DocRepoLocation,
 
-  [Parameter(Mandatory = $true)]
+  [Parameter(Mandatory = $false)]
```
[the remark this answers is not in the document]
sima-zhu: I tested in my local which Language and RepoId can read from Language-Settings.ps1
Don't want to make it required. We can have fewer parameters when use the script.

eng/common/pipelines/templates/steps/update-docsms-metadata.yml
```diff
@@ -108,10 +109,11 @@ steps:
...
         BaseRepoOwner: ${{ parameters.TargetDocRepoOwner }}
         CommitMsg: "Update docs metadata"
         TargetRepoName: ${{ parameters.TargetDocRepoName }}
         TargetRepoOwner: ${{ parameters.TargetDocRepoOwner }}
         WorkingDirectory: $(DocRepoLocation)
         ScriptDirectory: ${{ parameters.WorkingDirectory }}/${{ parameters.ScriptDirectory }}
+        PushArgs: -f
```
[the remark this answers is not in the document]
sima-zhu: Not necessary as this step finished really quick we don't want to override blindly. I am having this for testing purpose. Reverted.

eng/common/scripts/Update-DocsMsMetadata.ps1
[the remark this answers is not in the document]
weshaggard: I see where we set Language but I don't see RepoId set anywhere. So if it isn't passed what will happen?

weshaggard (approved):
On question about the RepoId but otherwise looks good.

eng/common/scripts/Update-DocsMsMetadata.ps1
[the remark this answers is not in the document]
sima-zhu: I plan to add one in parallel with `$Language`. I did not remove the parameter for the backward combability. 
Here is my plan:
1. Mark these two parameters as optional.
2. Go to each language-settings.ps1 and add `RepoId` as needed. 
3. Remove the `Language`, `RepoId` parameters in yml files which are calling this script. This step is the same PR with Step 2.
4. (Optional)Then, remove these two parameters from the script.

Issue #5119, opened 2023-01-11
API review - Package name is empty for Go API reviews


praveenkuttappan:
Package name is empty for Go API reviews, and this causes an issue when checking API approval status for a package from pipeline.

PR #5120, opened 2023-01-11
Set package name using module path if source is extracted package onl…

praveenkuttappan:
Fixes https://github.com/Azure/azure-sdk-tools/issues/5119

### DIFF
--- a/src/go/cmd/module.go
+++ b/src/go/cmd/module.go
@@ -54,6 +54,20 @@ func NewModule(dir string) (*Module, error) {
 		}
 		fmt.Printf("Package Name: %s\n", packageName)
 	}
+
+	//Package name can still be empty when generating API review using uploaded zip folder of a specific package in which case parent directory will not be sdk
+	if packageName == "" {
+		modulePath := mf.Module.Mod.Path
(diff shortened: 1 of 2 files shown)


src/go/cmd/module.go
```diff
@@ -54,6 +54,20 @@ func NewModule(dir string) (*Module, error) {
...
 		fmt.Printf("Package Name: %s\n", packageName)
 	}
+
+	//Package name can still be empty when generating API review using uploaded zip folder of a specific package in which case parent directory will not be sdk
+	if packageName == "" {
+		modulePath := mf.Module.Mod.Path
+		packageName := path.Base(modulePath)
+		fmt.Printf("Module path: %s\n", modulePath)
+		// Set relative path as package name for internal package to avoid collision
+		if packageName == "internal" {
```
jhendrixMSFT, line 64: This is special casing for modules like `sdk/internal`, `sdk/keyvault/internal` etc yes?

praveenkuttappan: correct. Otherwise, all internal packages will get same "internal" as name.

jhendrixMSFT (approved):

PR #5103, opened 2023-01-10
Update `get-codeowners.ps1` to work with the updated `RetrieveCodeOwners` executable + add some tests; make assorted refactorings

konrad-jamrozik:
This PR is part of work required in preparation to merge:
- #5088

Specifically, to enable review of ownership changes due to upcoming wildcards support, as explained in:
- https://github.com/Azure/azure-sdk-tools/pull/5088#issuecomment-1376542202

As such, this PR contributes to:
- #2770

This PR is a prerequisite for following PRs:
- https://github.com/Azure/azure-sdk-tools/pull/5431

### Merging prerequisite

This PR can be merged only after the following PRs are merged and relevant NuGet package published
- https://github.com/Azure/azure-sdk-tools/pull/5437
- #5104

This is because this PR depends on that NuGet package.

PR #5134, opened 2023-01-13
Add to the `retrieve-codeowners` tool support for returning owners for all paths matching given glob path.

konrad-jamrozik:
This PR implements for the `retrieve-codeowners` tool the ability to return not only owners for a single path, but a list of all owners for all paths resolved when matching a glob path given as input.

As such, this PR contributes to:
- #5135

This work is necessary to be able to compare and diff the owners of all files in repository before and after the regex matcher is turned on, per this comment:
- https://github.com/Azure/azure-sdk-tools/pull/5088#pullrequestreview-1241376684

In other words, this PR contributes to unblocking to the following PR:
- #5088

And as such, also contributes to:
- #2770

This PR will require some upstream changes, which are captured by:
- #5103

### Additional changes

- Removed logger from `CodeOwnersParser`; replaced with `Console.Out` and `Console.Error`. This addresses the following comment:
    - https://github.com/Azure/azure-sdk-tools/pull/5063#discussion_r1066019443
- Prevented the new regex matcher from matching paths that have `*` in them - such paths are malformed anyway (at least on Windows).
- A lot of assorted changes to surrounding production & test code - please see the file diff for details.

### Implementation notes

This PR leverages [`Microsoft.Extensions.FileSystemGlobbing`](https://www.nuget.org/packages/Microsoft.Extensions.FileSystemGlobbing).
Doc: https://learn.microsoft.com/en-us/dotnet/core/extensions/file-globbing


tools/code-owners-parser/CodeOwnersParser/CodeownersEntry.cs
```diff
@@ -157,5 +167,24 @@ private static bool IsGitHubUserAlias(string alias)
...
+            // @formatter:off
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj.GetType() != this.GetType()) return false;
+            return Equals((CodeownersEntry)obj);
+            // @formatter:on
+        }
+
+        public override int GetHashCode()
+            => HashCode.Combine(PathExpression, Owners, PRLabels, ServiceLabels);
```
konrad-jamrozik, line 188: @weshaggard I am using `SequenceEqual` in `Equals` above, so I do believe this is not exactly right approach to computing hash, per [this Stack Overflow question](https://stackoverflow.com/questions/28546239/handling-collections-in-gethashcode-implementation). However, I am not sure if this is big enough problem to try to get the hash code computed properly.

tools/code-owners-parser/CodeOwnersParser/Azure.Sdk.Tools.CodeOwnersParser.csproj
```diff
@@ -3,11 +3,14 @@
...
     <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>10.0</LangVersion>
+    <Nullable>enable</Nullable>
+    <WarningsAsErrors>Nullable</WarningsAsErrors>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="CommandLine.Net" Version="2.3.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" Version="7.0.0" />
```
konrad-jamrozik, line 13: FYI this is used in `GlobFilePath`.

tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners.Tests/Azure.Sdk.Tools.RetrieveCodeOwners.Tests.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>10.0</LangVersion>
```
weshaggard, line 5: Is this necessary to specify? What is the default language feature?

tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners/Azure.Sdk.Tools.RetrieveCodeOwners.csproj
```diff
@@ -3,6 +3,9 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>10.0</LangVersion>
```
weshaggard, line 6: Same question about these are they necessary?

tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners/Program.cs
```diff
@@ -1,54 +1,118 @@
...
-        /// </returns>
-        public static int Main(
-            string targetPath,
-            string codeownersFilePathOrUrl,
-            bool excludeNonUserAliases = false)
+        targetPath = targetPath.Trim();
+        targetDir = targetDir?.Trim();
+        codeownersFilePathOrUrl = codeownersFilePathOrUrl.Trim();
+
+        Debug.Assert(!string.IsNullOrWhiteSpace(targetPath));
```
weshaggard, line 48: Why not make these actual runtime checks?

tools/code-owners-parser/CodeOwnersParser/CodeownersEntry.cs
weshaggard: It isn't that critical unless it is used in a lookup table and there is risk of collision.

tools/code-owners-parser/CodeOwnersParser/GlobFilePath.cs
```diff
@@ -0,0 +1,36 @@
...
+    private readonly string filePath;
+
+    public GlobFilePath(string globFilePath)
+    {
+        Debug.Assert(globFilePath.IsGlobFilePath());
+        this.filePath = globFilePath;
+    }
+
+    public static bool IsGlobFilePath(string path)
+        => path.Contains('*');
```
weshaggard, line 21: Is that the only special char?

tools/code-owners-parser/CodeOwnersParser/MatchedCodeownersEntry.cs
```diff
@@ -52,20 +52,11 @@ internal class MatchedCodeownersEntry
         /// </summary>
         private static readonly char[] unsupportedChars = { '[', ']', '!', '?' };
 
-        private readonly ILogger<MatchedCodeownersEntry> log;
```
weshaggard, line 55: Is the ILogger interface from the other package you removed? If we ever want to redirect this logging to a file or some other service we might need to reintroduce a logging abstraction of some sort.

weshaggard (approved):
A few questions/comments but otherwise seems reasonable.

tools/code-owners-parser/CodeOwnersParser/MatchedCodeownersEntry.cs
konrad-jamrozik: @weshaggard yes, this is the logger I added recently, and now removed per [our discussion](https://github.com/Azure/azure-sdk-tools/pull/5063#discussion_r1066019443). Basically I am following YAGNI / You Aren't Gonna Need It principle here. If we need more advanced logging, then I'll re-add back the logger.

tools/code-owners-parser/CodeOwnersParser/GlobFilePath.cs
konrad-jamrozik: Yes, per these docs:
- https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.filesystemglobbing.matcher?view=dotnet-plat-ext-7.0#remarks
- https://learn.microsoft.com/en-us/dotnet/core/extensions/file-globbing#pattern-formats

I will add these links to the comments to eliminate the guesswork.

tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners/Program.cs
konrad-jamrozik: @weshaggard good point; I am going to make it throw always, but I will catch it, to print out to STDERR and return exit code of 1. 

This is because currently, only [the get-codeowners.ps1, uses this tool](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/get-codeowners.ps1#L41-L57), like so:

``` ps1
  if($includeNonUserAliases) {
    $codeOwnersString = & $command --target-directory $targetDirectory --code-owner-file-path $codeOwnerFileLocation 2>&1
  }
  else {
    $codeOwnersString = & $command --target-directory $targetDirectory --code-owner-file-path $codeOwnerFileLocation --filter-out-non-user-aliases 2>&1
  }
  # Failed at the command of fetching code owners.
  if ($LASTEXITCODE -ne 0) {
    Write-Host $codeOwnersString
    return ,@()
  }
  
  $codeOwnersJson = $codeOwnersString | ConvertFrom-Json
  if (!$codeOwnersJson) {
    Write-Host "No code owners returned from the path: $targetDirectory"
    return ,@()
  }
```

As you can see, it doesn't throw upon failure, only prints out the message and returns empty array. Hence, at least in this PR, I want to stay compatible with that usage.

tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners.Tests/Azure.Sdk.Tools.RetrieveCodeOwners.Tests.csproj
konrad-jamrozik: @weshaggard Removing this makes the projects revert to C# 8, which is unclear to me why, because [they should default to C# 10](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults). Thus I force it to C# 10 to be able to use bunch of newer features from it. On C# 8 I end up with bunch of compilation errors (unsurprisingly):

image

If you have an idea why the default is 8 and not 10, let me know.

tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners/Azure.Sdk.Tools.RetrieveCodeOwners.csproj
konrad-jamrozik: Replied [here](https://github.com/Azure/azure-sdk-tools/pull/5134#discussion_r1070172769).

tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners.Tests/Azure.Sdk.Tools.RetrieveCodeOwners.Tests.csproj
konrad-jamrozik: @weshaggard A ha! I think it is coming from [Directory.Build.props](https://github.com/Azure/azure-sdk-tools/blob/c124d9b0eec9927f28613747254d01f51dc2fb84/eng/Directory.Build.props#L7) ! I will make a PR to update this to `10.0`, now that we have moved everything to `net6.0`.

konrad-jamrozik: Submitted PR:
- #5175

tools/code-owners-parser/CodeOwnersParser/CodeownersEntry.cs
konrad-jamrozik: @weshaggard https://github.com/Azure/azure-sdk-tools/pull/5636

PR #5136, opened 2023-01-13
[Cadl APIView] Release for latest Cadl compiler

tjprescott:
Updating dependencies. No other changes.


tools/apiview/emitters/cadl-apiview/package.json
```diff
@@ -54,14 +54,14 @@
     "!dist/test/**"
   ],
   "devDependencies": {
```
weshaggard, line 56: @tjprescott should these actually be dependencies vs devDependencies? Do you need these at runtime when you are running the emitter?

tjprescott: Is there a way to tell npm to only install dependencies (not devDependencies)? But yeah, some of these are key dependencies and others are just for dev work.

weshaggard: https://stackoverflow.com/questions/18875674/whats-the-difference-between-dependencies-devdependencies-and-peerdependencies

Yes I believe so. In general I want us to list any cadl specific dependencies under dependencies so we are sure things will work if we only reference the emitter package which is what we plan to do in our automation.

johanste (approved):

Issue #5148, opened 2023-01-16
Change the growup logic in docker image to download latest .Net sdk configured in .Net SDK repo


dw511214992:
Not only .Net sdk, but also the version of other language.

PR #5149, opened 2023-01-16
Upgrade .Net sdk version (Temp Solution)

dw511214992:
It's a temp solution. I will change it to download .Net sdk by fetch configuration from .Net sdk repository automatically when user start the docker container. (tracking issue: https://github.com/Azure/azure-sdk-tools/issues/5148)

### DIFF
--- a/tools/sdk-generation-pipeline/Dockerfile
+++ b/tools/sdk-generation-pipeline/Dockerfile
@@ -17,7 +17,7 @@ ENV M2_HOME /usr/maven
 ENV MAVEN_HOME=/usr/maven
 ENV PATH $M2_HOME/bin:$PATH
 RUN mkdir -p "$MAVEN_HOME"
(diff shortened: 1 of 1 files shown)


TonyHCIR (commented):
Mp4

tools/sdk-generation-pipeline/Dockerfile
```diff
@@ -40,9 +40,9 @@ RUN rm packages-microsoft-prod.deb
 RUN wget https://dotnet.microsoft.com/download/dotnet/scripts/v1/dotnet-install.sh && chmod 777 ./dotnet-install.sh
 RUN bash ./dotnet-install.sh
 RUN bash ./dotnet-install.sh -c 3.1
-ENV DOTNET_ROOT=/root/.dotnet PATH=$PATH:/root/.dotnet
```
weshaggard, line 43: Should we be updating to a newer dotnet Docker image instead of installing it ourselves?

weshaggard: Actually never mind of this Docker needs to contain the tools for other all languages currently so it cannot start with one language Docker.

PR #5154, opened 2023-01-17
Test Roslyn analyzer changes against net472

heaths:
From conversations with .NET regarding #5048, make sure we test against
.NET Framework TFMs to help make sure we don't break analyzers again.

### DIFF
--- a/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/Azure.ClientSdk.Analyzers.Tests.csproj
+++ b/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/Azure.ClientSdk.Analyzers.Tests.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


jsquire (approved):

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/Azure.ClientSdk.Analyzers.Tests.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
```
weshaggard, line 5: Does this work the way you expect? In other words, does this target end up in the package? Or does that only happen if we build on windows?

heaths: Oh, that's a good point. This is only the test project so there's no packaging, but if we never run it on Windows it won't matter.

heaths: You're right: we only build and test on Ubuntu. I'll open a separate track issue.

Issue #5155, opened 2023-01-17
APIView - Swagger API review step failed with stack overfkow


praveenkuttappan:
Swagger API review gen step in swagger PR pipeline failed with stack overflow error when it was attempting to generate APIView token file from swagger file.

Stack overflow.at System.Collections.Generic.List`1[[System.__Canon, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].Add(System.__Canon)at SwaggerApiParser.BaseSchema.GetKeywords()at SwaggerApiParser.BaseSchema.GetPropertyKeywordsFromBaseSchema(SwaggerApiParser.BaseSchema, System.String, SwaggerApiParser.BaseSchema)at SwaggerApiParser.BaseSchema.TokenSerializeProperties(SwaggerApiParser.SerializeContext, SwaggerApiParser.BaseSchema, System.Collections.Generic.Dictionary`2<System.String,SwaggerApiParser.BaseSchema>, System.Collections.Generic.List`1<APIView.CodeFileToken>, System.Collections.Generic.List`1<SwaggerApiParser.SchemaTableItem> ByRef, Boolean)at SwaggerApiParser.BaseSchema.TokenSerializeInternal(SwaggerApiParser.SerializeContext, SwaggerApiParser.BaseSchema, System.Collections.Generic.List`1<SwaggerApiParser.SchemaTableItem> ByRef, Boolean)at SwaggerApiParser.BaseSchema.TokenSerializeArray(SwaggerApiParser.SerializeContext,

https://github.com/Azure/azure-rest-api-specs/pull/19539/checks?check_run_id=10491916725

PR #5160, opened 2023-01-18
fix root schema circular reference issue 

ruowan:
Fix root schema circular reference issue by adding root schema ref in refChain. 

This PR is to address https://github.com/Azure/azure-sdk-tools/issues/5155

### DIFF
--- a/tools/apiview/parsers/swagger-api-parser/SwaggerApiParser/SwaggerAPIViewGenerator.cs
+++ b/tools/apiview/parsers/swagger-api-parser/SwaggerApiParser/SwaggerAPIViewGenerator.cs
@@ -124,7 +124,15 @@ namespace SwaggerApiParser
                         
                         if (response.schema != null)
                         {
-                            schema = schemaCache.GetResolvedSchema(schema, currentSwaggerFilePath);
+                            // The initial refChain is the root level schema.
+                            // There is some scenario that the property of the root level schema is a ref to the root level itself (circular reference). Like "errorDetail" schema
+                            LinkedList<string> refChain = new LinkedList<string>();
+                            if (schema.Ref != null)
+                            {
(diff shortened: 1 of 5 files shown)


tools/apiview/parsers/swagger-api-parser/SwaggerApiParser/SwaggerAPIViewGenerator.cs
```diff
@@ -124,7 +124,16 @@ public static SwaggerApiViewSpec GenerateSwaggerApiView(SwaggerSpec swaggerSpec,
...
+                            // The initial refChain is the root level schema.
+                            // There are some scenarios that the property of the root level schema is a ref to the root level itself (circular reference).
+                            // Like "errorDetail" schema in common types.
+                            LinkedList<string> refChain = new LinkedList<string>();
+                            if (schema.Ref != null)
+                            {
+                                refChain.AddFirst(schema.Ref);
+                            }
+                            
+                            schema = schemaCache.GetResolvedSchema(schema, currentSwaggerFilePath, refChain);
```
praveenkuttappan, line 136: Do we need this change here as well? https://github.com/Azure/azure-sdk-tools/blob/ffb60fcabb8ea12b22c8940d3273b87d45b29a59/tools/apiview/parsers/swagger-api-parser/SwaggerApiParser/SwaggerAPIViewGenerator.cs#L94

praveenkuttappan (approved):

tools/apiview/parsers/swagger-api-parser/SwaggerApiParser/SwaggerAPIViewGenerator.cs
ruowan: I think maybe for now, we don't need added here. Since I don't see any parameters has circular reference in spec repo.

PR #5164, opened 2023-01-18
fix failing test-proxy livetests -- the run-pester-tests template was renamed

scbedd:
We really should be doing a find/replace when we delete/rename an existing template :|


benbp (approved):
FYI @sima-zhu

PR #5169, opened 2023-01-19
Handle service principal creation quota errors

benbp:
The warning message in this scenario is not the best, so I added some diagnostics for the common failure case here to try to be a bit more helpful for the user.

The specific error in question is for "The directory object quota limit for the Principal has been exceeded"

CC @kashifkhan

### DIFF
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -139,7 +139,7 @@ function Retry([scriptblock] $Action, [int] $Attempts = 5)
                 Write-Warning "Attempt $attempt failed: $_. Trying again in $sleep seconds..."
                 Start-Sleep -Seconds $sleep
             } else {
-                Write-Error -ErrorRecord $_
+                throw
             }
         }
     }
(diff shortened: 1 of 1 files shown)


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -158,9 +159,28 @@ function NewServicePrincipalWrapper([string]$subscription, [string]$resourceGrou
...
+        $servicePrincipal = Retry {
+            New-AzADServicePrincipal -Role "Owner" -Scope "/subscriptions/$SubscriptionId/resourceGroups/$ResourceGroupName" -DisplayName $displayName
+        }
+    } catch {
+        # The underlying error "The directory object quota limit for the Principal has been exceeded" gets overwritten by the module trying
+        # to call New-AzADApplication with a null object instead of stopping execution, which makes this case hard to diagnose because it prints the following:
+        #      "Cannot bind argument to parameter 'ObjectId' because it is an empty string."
+        # Provide a more helpful diagnostic prompt to the user if appropriate:
+        $apps = Get-AzADApplication -OwnedApplication
+        if ($apps.Length -ge 135) {
```
weshaggard, line 173: Where is this magic number coming from?

benbp: I couldn't find any documentation on per-user quotas, but it seems to be the app registration number that I start getting quota errors for, and I can only assume it's a tenant-level setting. I can try to get some others to corroborate to see if this is consistent across our tenant, or we could just always print this warning along with the current user's app count.

weshaggard: I would probably start with the later and aways print the count and give perhaps give folks a warning if the number is high enough that this might be the problem.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -140,6 +140,7 @@ function Retry([scriptblock] $Action, [int] $Attempts = 5)
                 Start-Sleep -Seconds $sleep
             } else {
                 Write-Error -ErrorRecord $_
+                throw
```
heaths, line 143: This ends up adding two records to the `$Error` stream. Better to just `throw $_`, which will also effectively `write-error` anyway.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -158,9 +159,26 @@ function NewServicePrincipalWrapper([string]$subscription, [string]$resourceGrou
...
+        #      "Cannot bind argument to parameter 'ObjectId' because it is an empty string."
+        # Provide a more helpful diagnostic prompt to the user if appropriate:
+        $totalApps = (Get-AzADApplication -OwnedApplication).Length
+        $msg = "App Registrations owned by you total $totalApps and may exceed the max quota (likely around 135)." + `
+               "`nTry removing some at https://ms.portal.azure.com/#view/Microsoft_AAD_IAM/ActiveDirectoryMenuBlade/~/RegisteredApps" + `
+               " or by running the following command to remove apps created by this script:" + `
+               "`n    Get-AzADApplication -DisplayNameStartsWith '$baseName' | Remove-AzADApplication" + `
+               "`nNOTE: You may need to wait for the quota number to be updated after removing unused applications."
+        Write-Warning $msg
+        throw
```
heaths, line 179: Same as above, but also confusing because it's terminal, yes you write it as a warning. I would just `throw $msg`.

heaths (approved):

eng/common/TestResources/New-TestResources.ps1
benbp: Wouldn't `throw $msg` swallow the underlying error? Granted in this case it's already a misleading one so perhaps it's better.

weshaggard (approved):
Looks reasonable but also unclear on the error handling that Heath is mentioning.

eng/common/TestResources/New-TestResources.ps1
benbp: The warning here is a hint as to why the error might have occurred, but not guaranteed to be the error itself.

PR #5175, opened 2023-01-19
Unhardcode C# lang versions, making them use TFM default (currently `10.0` due to TFM of `net6.0`).

konrad-jamrozik:
As in subject. Because we use `net6.0`, [currently the default lang version is `10.0`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults).

Previously the lang version was forced by [`eng/Directory.Build.props`](https://github.com/Azure/azure-sdk-tools/pull/5175/files#diff-81138f4f0bdd24607d3d5979bdc08f91e9cfe44255de13ce958719925f1ae88b), hence many projects were overriding it. With this PR this is no longer necessary.

The need for this work was discovered during the following:
- https://github.com/Azure/azure-sdk-tools/pull/5134#discussion_r1070172769

### DIFF
--- a/eng/Directory.Build.props
+++ b/eng/Directory.Build.props
@@ -4,7 +4,6 @@
     <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
     <Platform Condition="'$(Platform)' == ''">AnyCPU</Platform>
     <PlatformName Condition="'$(PlatformName)' == ''">$(Platform)</PlatformName>
-    <LangVersion>8.0</LangVersion>
     <SignAssembly>true</SignAssembly>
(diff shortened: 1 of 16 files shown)


weshaggard:
The description is a little misleading as it doesn't default to latest but it is set depending on what TFM you are using. As you mention in your description nert6.0 uses version 10. At any rate I like the idea of letting the tooling default the language version unless there is a strong reason to override it.

weshaggard:
Based on CI looks like there might still be some need to override some versions.

weshaggard (approved):
General change looks good but please make sure CI is passing first and only override LangVersion in cases that require it.

PR #5195, opened 2023-01-24
Fix Performance Pipeline Profiling when Using Java 17

alzimmermsft:
Fixes/adds support for running performance pipeline profiling when running with both Java 8 and Java 17.


mikeharder (approved):
Approved, but should be merged after https://github.com/Azure/azure-sdk-tools/pull/5205

PR #5196, opened 2023-01-24
Install packages from cadl package and list installed packages

praveenkuttappan:
CADL package should be installed from package.json and also added a debug line to show installed packages.

### DIFF
--- a/eng/scripts/Create-Apiview-Token-Cadl.ps1
+++ b/eng/scripts/Create-Apiview-Token-Cadl.ps1
@@ -31,6 +31,7 @@ function Generate-Apiview-File($packagePath)
     try
     {
         npm install
+        npm list
         cadl compile . --emit=@azure-tools/cadl-apiview
     }
     finally
--- a/eng/scripts/Create-Apiview-Token-Cadl.ps1
+++ b/eng/scripts/Create-Apiview-Token-Cadl.ps1
@@ -11,6 +11,7 @@ param (
 
 Set-StrictMode -Version 3
 $sparseCheckoutFile = ".git/info/sparse-checkout"
+$globalPackageJsonPath = "specification/package.json"
 
 
 function Sparse-Checkout($branchName, $packagePath)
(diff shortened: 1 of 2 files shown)


praveenkuttappan:
This PR will resolve issue of API review not getting created for CADL even though token file is generated.

tjprescott (changes requested):
You will still need to install cadl-apiview (soon to be typespec-apiview) if the package.json does not list it in order for APIView generation to work.

praveenkuttappan:
> You will still need to install cadl-apiview (soon to be typespec-apiview) if the package.json does not list it in order for APIView generation to work.

cadl-apiview will get installed when installing packages using global package.json and that will cover the scenario of not having this dependency in project specific package.json.

tjprescott (approved):

eng/scripts/Create-Apiview-Token-Cadl.ps1
```diff
@@ -19,18 +20,29 @@ function Sparse-Checkout($branchName, $packagePath)
...
     Write-Host "Generating API review token file from path '$($packagePath)'"
     Push-Location $packagePath
     try
     {
+
+        # Install required packages using project specific package.json          
+        npm install
+        npm list
+        # Install additional/missing packages using gloal package.json
+        Rename-Item package.json package-orig.json
```
weshaggard, line 40: Is this just for testing or temporary until we get the correct global settings?

eng/scripts/Create-Apiview-Token-Cadl.ps1
```diff
@@ -11,6 +11,7 @@ param (
 
 Set-StrictMode -Version 3
 $sparseCheckoutFile = ".git/info/sparse-checkout"
+$globalPackageJsonPath = "specification/package.json"
```
weshaggard, line 14: Note this is going to change to the root of the repo soon in https://github.com/Azure/azure-rest-api-specs/pull/22735.

eng/scripts/Create-Apiview-Token-Cadl.ps1
weshaggard: Or are you trying to enable both sets of package installs?

praveenkuttappan: Done.

eng/scripts/Create-Apiview-Token-Cadl.ps1
```diff
@@ -19,24 +20,19 @@ function Sparse-Checkout($branchName, $packagePath)
...
-        npm install
-        cadl compile . --emit=@azure-tools/cadl-apiview
-    }
-    finally
-    {
-        Pop-Location
-    }
+    npm install
+    npm list
+    cadl compile . --emit=@azure-tools/cadl-apiview
```
weshaggard, line 34: Doesn't this also require that you add this package into that global package.json to get it to work?

weshaggard: Nevermind it is already added https://github.com/Azure/azure-rest-api-specs/blob/main/package.json#L17

PR #5207, opened 2023-01-24
Check for empty snippets

### LINKED ISSUE #5206: SnippetGenerator should validate regions are non-empty
Because SnippetGenerator is already part of the workflow (indirectly via CodeChecks.ps1), it should validate that any markdown sample region is non-empty after supposed replacement. It already checks that all snippets are referenced and none are undefined, but there may be reasons - such as preproc conditions at one point in https://github.com/Azure/azure-sdk-for-net/pull/33619 - where regions end up empty.

For example, consider a test like this:

```c#
class Tests
{
#if NETSTANDARD2_1_OR_NEWER
  [Test]
  public void Test1()
  {
    #region Snippet:Tests.Test1
    Console.WriteLine("It works!");
    #endregion
  }
#endif
}
```

Because SnippetGenerator currently doesn't define any other symbols than `SNIPPET`, that `#region` isn't even found at compilation time. Thus, any reference like below will be empty:

````markdown
```C# Snippet:Tests.Test1
```
````

### PR DESCRIPTION
**Pull request by heaths**
Resolves #5206

### DIFF
--- a/tools/snippet-generator/Azure.Sdk.Tools.SnippetGenerator.Tests/TestData/Snippets.cs
+++ b/tools/snippet-generator/Azure.Sdk.Tools.SnippetGenerator.Tests/TestData/Snippets.cs
@@ -1,0 +1,17 @@
+using NUnit.Framework;
+
+namespace Azure.Sdk.Tools.SnippetGenerator.TestData
+{
+    public class Snippets
+    {
+        [Test]
+        public void EmptySnippet()
+        {
+#if SHOULD_NEVER_BE_DEFINED
+            #region Snippet:EmptySnippet
+            throw new NotImplementedException();
+            #endregion
+#endif
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by christothes — APPROVED** [review]
Looks good!

**Comment by heaths** [comment]
Everything but check-enforcer passed.

File: tools/snippet-generator/Azure.Sdk.Tools.SnippetGenerator.Tests/TestData/Snippets.cs [hunk 1]
```diff
@@ -0,0 +1,17 @@
...
+namespace Azure.Sdk.Tools.SnippetGenerator.TestData
+{
+    public class Snippets
+    {
+        [Test]
+        public void EmptySnippet()
+        {
+#if SHOULD_NEVER_BE_DEFINED
+            #region Snippet:EmptySnippet
+            throw new NotImplementedException();
```
annelo-msft, line 12: Curious: why is this single line considered empty?

heaths: Because the symbol `SHOULD_NEVER_BE_DEFINED` wasn't defined, so there ends up being no content. This is effectively what happened in my case that prompted this feature, so I figured it would make a good test case.

Somehow (I didn't dig in too much), the region still gets defined. Presumably it's because of the visitor we have and how it's called by Roslyn, that all preprocessor instructions are considered "in".

Issue #5204, opened 2023-01-24
Test-Proxy common usage updates


scbedd:
Discovered as part of a discussion with @LarryOsterman about usability improvements

- [x] Upload test-proxy.log artifact by default
- [x] Ensure that we dump the exception to console in all cases (regardless of verbosity level)evel)

scbedd:
Resolved with the two mentioned PRs.

PR #5218, opened 2023-01-25
test-proxy always logs exceptions on console

scbedd:
Related to #5204

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/HttpExceptionMiddleware.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/HttpExceptionMiddleware.cs
@@ -1,5 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Text.Json;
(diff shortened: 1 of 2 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/HttpExceptionMiddleware.cs
```diff
@@ -79,6 +80,8 @@ public async Task Invoke(HttpContext context)
                 response.Headers.Add("x-request-exception", "true");
                 response.Headers.Add("x-request-exception-error", Convert.ToBase64String(Encoding.UTF8.GetBytes(e.Message)));
 
+                DebugLogger.LogError(unexpectedStatusCode.ToString() + Environment.NewLine + e.Message + Environment.NewLine + e.StackTrace);
```
benbp, line 83: Would it make more sense to put this logic into `LogError`, e.g.:

```
DebugLogger.LogError(unexpectedStatusCode.ToString(), e);
```

and

```
public static void LogError(string details, Exception? err)
{
    if (null != logger)
    {
        logger.LogError(err, details);
    }
    else if (null != Exception)
    {
        System.Console.WriteLine(details + Environment.NewLine + err.Message + Environment.NewLine + err.StackTrace);
    }
    else
    {
        System.Console.WriteLine(details);
    }
}
```

benbp (approved):
Minor suggestion but LGTM either way.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/HttpExceptionMiddleware.cs
scbedd: Fair point.

PR #5225, opened 2023-01-25
[testmodeler] Update dependency semver ranges

### LINKED ISSUE #5219: [testmodeler] Update dependency semver ranges
In package `@autorest/testmodeler`, I recommend updating dependency semver ranges as follows:

1. Most dependencies should float using `^`.
2. Pre-release dependencies should be pinned, since they can have breaking changes at any time.
3. `typescript` should float using `~`, since `typescript` allows breaking changes in patch releases.

I noticed some dependencies also have newer major versions available, but these should be updated separately and checked carefully for breaking changes.

mikeharder:
- Upgrade @types/node to ^14, since Node 14 is oldest supported version
- Downgrade pre-release versions to pinned, since breaking changes are allowed at any time
- Downgrade typescript from ^ to ~, since breaking changes are allowed in minor releases
- Upgrade oav to 3.2.2, since 3.2.3 has a (likely accidental) breaking change
- Upgrade all other packages from ~ to ^
- Fixes #5219


changlong-liu:
Hi @mikeharder , I continue this PR at https://github.com/Azure/azure-sdk-tools/pull/5263/commits/c2ba6e24467ee930895b585d4ffda1ef041d9178, we may cherry-pick it into this PR to make CI pass.


And during update the tests, I noticed oav@3.2.0 added some new properties (comparing to 0av@3.0.3) in the output and this will be a breaking change for autorest.csharp (which cosumes the output of autorest.testmodeler).
Since the issue https://github.com/Azure/azure-sdk-tools/issues/5183 has been closed, shall we still use oav@3.0.3 in this PR?

changlong-liu:
> Hi @mikeharder , I continue this PR at [c2ba6e2](https://github.com/Azure/azure-sdk-tools/commit/c2ba6e24467ee930895b585d4ffda1ef041d9178), we may cherry-pick it into this PR to make CI pass.
> 
> And during update the tests, I noticed oav@3.2.0 added some new properties (comparing to 0av@3.0.3) in the output and this will be a breaking change for autorest.csharp (which cosumes the output of autorest.testmodeler). Since the issue #5183 has been closed, shall we still use oav@3.0.3 in this PR?

oh, I found oav@3.2.0 updated lots of old dependencies, so I tend to think updating to oav@3.2.0 is necessary. Thanks Mike.

mikeharder:
> And during update the tests, I noticed oav@3.2.0 added some new properties (comparing to 0av@3.0.3) in the output and this will be a breaking change for autorest.csharp (which cosumes the output of autorest.testmodeler). Since the issue #5183 has been closed, shall we still use oav@3.0.3 in this PR?

If `oav` is following semver, there should not be breaking changes in minor or patch versions, and we should be able to float the dependency using `^`.  If `oav` is not following semver, then we need to understand what is their policy for breaking changes, are they only allowed in minor versions, or even in patch versions?  And float using `~` or pin a specific version as required.

The goal of upgrading `oav` is to get onto the latest version `3.2.4` when it is released, which fixes issues like https://github.com/Azure/oav/issues/950.  We can wait until `oav@3.2.4` is released, or upgrade now to the latest version still compatible with testmodeler, then upgrade again to `oav@3.2.4` when it's released.

changlong-liu:
> > And during update the tests, I noticed oav@3.2.0 added some new properties (comparing to 0av@3.0.3) in the output and this will be a breaking change for autorest.csharp (which cosumes the output of autorest.testmodeler). Since the issue #5183 has been closed, shall we still use oav@3.0.3 in this PR?
> 
> If `oav` is following semver, there should not be breaking changes in minor or patch versions, and we should be able to float the dependency using `^`. If `oav` is not following semver, then we need to understand what is their policy for breaking changes, are they only allowed in minor versions, or even in patch versions? And float using `~` or pin a specific version as required.
> 
> The goal of upgrading `oav` is to get onto the latest version `3.2.4` when it is released, which fixes issues like [Azure/oav#950](https://github.com/Azure/oav/issues/950). We can wait until `oav@3.2.4` is released, or upgrade now to the latest version still compatible with testmodeler, then upgrade again to `oav@3.2.4` when it's released.
> 
> @raych1: Do you know what versioning policy is used by `oav`? When are breaking changes allowed?

Yes, absolutely agree that we can wait until `oav@3.2.4` released.

Adding background for the breaking change discussion: 
The changes in `oav@3.2.0` is `adding some optional properties in apiscenario`, it may be not considered as breaking for JS projects like oav. Meanwhile this may be a breaking for autoreset.csharp because it's written in C# and need to update yaml schema. So I guess we can use `~` or pin a specific version of oav in autorest.testmodeler.

mikeharder:
> Yes, absolutely agree that we can wait until `oav@3.2.4` released.
> 
> Adding background for the breaking change discussion: The changes in `oav@3.2.0` is `adding some optional properties in apiscenario`, it may be not considered as breaking for JS projects like oav. Meanwhile this may be a breaking for autoreset.csharp because it's written in C# and need to update yaml schema. So I guess we can use `~` or pin a specific version of oav in autorest.testmodeler.

`oav@3.2.4` was released last night.  If you think `@autorest/testmodeler` (or downstream dependencies like `@autorest/csharp`) are sensitive to changes in `oav`, then pinning a specific version is the best option.  I agree that "breaking changes" can be tricky to define in the context of tooling like autorest.

PR #5226, opened 2023-01-25
Resolve HResult Failure for Mac Binaries

### LINKED ISSUE #5051: Signed `test-proxy` mac binary fails to run
Targeting [this release](https://github.com/Azure/azure-sdk-tools/releases/tag/test-proxy_1.0.0-dev.20221212.5) from [this build](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2051615&view=artifacts&pathAsName=false&type=publishedArtifacts) as an example.

```
matthewp@Matthews-MacBook-Air-2  ~/Downloads/test-proxy-standalone-osx-arm64  ./test-proxy
Failed to create CoreCLR, HRESULT: 0x80004005
```

There is no output when running the unsigned binary, probably due to the untrusted executable. Need to repro on the mac machine we have in the office.

scbedd:
@benbp this is the result of all the fighting with the mac signing. Primarily, shifts the `produce executables` to its own job matrix. I do this so that mac builds can be produced on mac, the others will be produced on a ubuntu machine.

We **have** to produce the `mac` binaries on mac because we need to codesign it with a set of entitlements, and this tool isn't available off of mac. I wouldn't expect _any_ of the produced dotnet binaries to work on mac by default without this.

Known todo:

- [x] Need to update azure-sdk-build-tools to consume each executable as an individual artifact download
- [x] move the entitlements file to a general use-case one, rather than existing in test-proxy.

Resolves #5051

This PR will be closed out in favor of the auto-pr from azure-sdk-build-tools. Leaving it open for validation in the meantime.


eng/pipelines/templates/scripts/assemble-dotnet-standalone-exe.ps1
```diff
@@ -0,0 +1,92 @@
...
+# resolves to tools/test-proxy/win-x64
+$outputPath = Join-Path $($Target) $($Rid)
+
+# ensure the destination artifact directory exists
+if (!(Test-Path $destinationArtifactFolder)){
+   New-Item -Force -Path $destinationArtifactFolder -ItemType directory
+}
+
+Write-Host "dotnet publish -f $($Framework) -c Release -r $($Rid) -p:PublishSingleFile=true --self-contained --output $outputPath $Target"
+dotnet publish -f $($Framework) -c Release -r $($Rid) -p:PublishSingleFile=true --self-contained --output $outputPath $Target
```
benbp, line 48: A tad more readable without the unnecessary string wrappers since you aren't referencing any properties:

```suggestion
$destinationArtifactFolder = Join-Path $ArtifactStagingDirectory $Rid

# resolves to <artifactfolder>/win-x64/test-proxy-standalone-win-x64 (.zip or .tar.gz will be added as appropriate for platform)
$destinationPathSegment = Join-Path $destinationArtifactFolder "$(Split-Path -Leaf "$Target")-standalone-$Rid"

# resolves to tools/test-proxy/win-x64
$outputPath = Join-Path $Target $Rid

# ensure the destination artifact directory exists
if (!(Test-Path $destinationArtifactFolder)){
   New-Item -Force -Path $destinationArtifactFolder -ItemType directory
}

Write-Host "dotnet publish -f $Framework -c Release -r $Rid -p:PublishSingleFile=true --self-contained --output $outputPath $Target"
dotnet publish -f $Framework -c Release -r $Rid -p:PublishSingleFile=true --self-contained --output $outputPath $Target
```

eng/pipelines/templates/scripts/assemble-dotnet-standalone-exe.ps1
```diff
@@ -0,0 +1,92 @@
...
+   $binaryFile = (Get-ChildItem -Path . | Where-Object { !([System.IO.Path]::HasExtension($_)) } | Select-Object -First 1).ToString().Replace("`\","/")
+   bash -c "chmod +x $binaryFile"
+   tar -cvzf "$($destinationPathSegment).tar.gz" .
+   Pop-Location
+}
+elseif("$($Rid)".Contains("osx")){
+   # need to codesign the binary with an entitlements file such that the signed and notarized binary will properly invoke on
+   # a mac system. However, the `codesign` command is only available on a MacOS agent. With that being the case, we simply special case
+   # this function here to ensure that the script does not fail outside of a MacOS agent.
+   if ($IsMacOS) {
```
benbp, line 72: Missing some context, why do we not want it to fail outside of macos if the target platform is osx? Is it because it gets called anyway for the whole matrix, and we know that there are separate jobs handling the macos part?

eng/pipelines/templates/steps/produce-net-standalone-packs.yml
```diff
@@ -18,47 +18,53 @@ parameters:
...
-      env:
-        DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
-        DOTNET_CLI_TELEMETRY_OPTOUT: 1
-        DOTNET_MULTILEVEL_LOOKUP: 0
+      - template: /eng/common/pipelines/templates/steps/publish-artifact.yml
+        parameters:
+          ArtifactName: "standalone-${{ target.rid }}"
+          ArtifactPath: "$(Build.ArtifactStagingDirectory)/${{ target.rid }}"
 
+    - ${{ if and(not(startswith(target.rid, 'osx')), not(eq(variables['Agent.OS'], 'Darwin'))) }}:
```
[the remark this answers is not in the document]
scbedd: Will reduce dup here.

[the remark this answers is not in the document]
scbedd: Actually yes this duplication has to be here. Mapping it out:

```
Mac Job:
  osx-x64
    run -> publish
  osx-arm64
    run -> publish
  win-64
    skip
  linux-x64
    skip
  linux-arm64
    skip
Ubuntu Job:
  osx-x64
   skip osx
  osx-arm64
   skip osx
  win-64
    build->publish
  linux-x64
    build->publish
  linux-arm64
    build->publish
```

If we place the publish at the surface level, we will always run `publish-artifact` even when we have skipped the build. Make sense?

eng/pipelines/templates/steps/produce-net-standalone-packs.yml
```diff
@@ -17,48 +17,65 @@ parameters:
...
-          $binaryFile = (Get-ChildItem -Path . | Where-Object { !([System.IO.Path]::HasExtension($_)) } | Select-Object -First 1).ToString().Replace("`\","/")
-          bash -c "chmod +x $binaryFile"
-          tar -cvzf "$($destinationPathSegment).tar.gz" .
-          Pop-Location
-        }
-        else {
-          Compress-Archive -Path "$($sourcePath)/*" -DestinationPath "$($destinationPathSegment).zip"
-        }
+    # ensure that we only build the mac standalone executables on a mac agent, everything else on not mac
+    - ${{ if startsWith(target.rid, 'osx') }}:
```
weshaggard, line 34: Am I missing something or are these 2 code blocks exactly the same except for the condition on OS agent? Should we just validate OS agent and rid inside of the script?

weshaggard: Well not necessarily validate but filter them to only run the osx rid entries on Mac.

weshaggard (approved):
Looks reasonable other then my one comment where it would be nice to avoid duplicating the set of steps and instead just filter down the matrix to the correct sets for mac and non-mac.

eng/pipelines/templates/steps/produce-net-standalone-packs.yml
scbedd: This is the only way to make that happen. `Agent.OS` is a _runtime_ variable, so cannot be used in template expression syntax. If you look at the actual steps that are executed, everything lines up.

scbedd: My original version of this was exactly a 


`if target.rid starts with osx and agent.os == darwin` -> run steps
`if target.rid doesn't start with osx and agent.os != darwin` -> run steps.

This combines it such that we only run the appropriate step on the right platform.

weshaggard: I guess I was thinking we could stop looking in the template and do it in the script and just filter based on AgentOS and Rid. However, I agree this approach as well but we could do it dynamically.

scbedd: The artifact publish itself remains complex and wonky if we do it that way. I am confident this will work, and I have burned **far** too much time into the PR already.

I really appreciate you signed off while having this discussion, and we definitely can revisit. I'm certain that I will have another opportunity in here.

scbedd:
Closing in favor of merging the changes into this PR: https://github.com/Azure/azure-sdk-tools/pull/5226

weshaggard:
I believe he intended to link https://github.com/Azure/azure-sdk-tools/pull/5174.

PR #5230, opened 2023-01-25
[Cadl APIView] Switch to peerDependencies

tjprescott:
Change to hopefully simplify automatic generation of APIViews.

### DIFF
--- a/tools/apiview/emitters/cadl-apiview/package.json
+++ b/tools/apiview/emitters/cadl-apiview/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@azure-tools/cadl-apiview",
-  "version": "0.3.4",
+  "version": "0.3.5",
   "author": "Microsoft Corporation",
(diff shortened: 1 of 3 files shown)


tools/apiview/emitters/cadl-apiview/package.json
```diff
@@ -53,11 +53,8 @@
     "dist/**",
     "!dist/test/**"
   ],
-  "dependencies": {
-    "@azure-tools/cadl-azure-core": "~0.25.0",
-    "@azure-tools/cadl-dpg": "~0.25.0",
+  "peerDependencies": {
```
weshaggard, line 56: How does peerDependencies actually help with our problems?

mikeharder: We should try to avoid using `peerDependencies` unless absolutely necessary, since many package managers do not handle them well.  So I agree with Wes, I'd like to know more about why we think we need to use `peerDependencies` and what problem we are trying to solve.

PR #5231, opened 2023-01-25
Move cadl project scripts to common

m-nash:


eng/common/scripts/Cadl-Project-Generate.ps1
```diff
@@ -0,0 +1,67 @@
...
+    try {
+        $currentDur = Resolve-Path "."
+        Write-Host "Generating from $currentDur"
+        if (Test-Path "package.json") {
+            Remove-Item -Path "package.json" -Force
+        }
+        if (Test-Path ".npmrc") {
+            Remove-Item -Path ".npmrc" -Force
+        }
+        $replacementPackageJson = "$PSScriptRoot/../../emitter-package.json"
```
weshaggard, line 24: We might want to allow the overriding of this package path as well but I'm find with this as a default.

eng/common/scripts/Cadl-Project-Generate.ps1
```diff
@@ -0,0 +1,67 @@
...
+
+function NpmInstallForProject([string]$workingDirectory) {
+    Push-Location $workingDirectory
+    try {
+        $currentDur = Resolve-Path "."
+        Write-Host "Generating from $currentDur"
+        if (Test-Path "package.json") {
+            Remove-Item -Path "package.json" -Force
+        }
+        if (Test-Path ".npmrc") {
```
weshaggard, line 21: Do we also want to potentially remove any lock files and node_modules directory? I've found if I try to do this work with 2 different emitters back to back then those cause issues as well.

eng/common/scripts/Cadl-Project-Generate.ps1
```diff
@@ -0,0 +1,67 @@
...
+$emitterName = &$GetEmitterNameFn
+$cadlConfigurationFile = Resolve-Path "$ProjectDirectory/cadl-location.yaml"
+
+Write-Host "Reading configuration from $cadlConfigurationFile"
+$configuration = Get-Content -Path $cadlConfigurationFile -Raw | ConvertFrom-Yaml
+
+$specSubDirectory = $configuration["directory"]
+$innerFolder = Split-Path $specSubDirectory -Leaf
+
+$tempFolder = "$ProjectDirectory/TempCadlFiles"
```
weshaggard, line 44: If there isn't a good reason to have the Temp folder in the same directory it would be nice come up with some other convention, maybe even the global temp folder. 

If we keep it under the repo we will need to get it added to .gitignore in the other repos as well.

eng/common/scripts/Cadl-Project-Generate.ps1
```diff
@@ -0,0 +1,67 @@
...
+        if (Test-Path "package.json") {
+            Remove-Item -Path "package.json" -Force
+        }
+        if (Test-Path ".npmrc") {
+            Remove-Item -Path ".npmrc" -Force
+        }
+        $replacementPackageJson = "$PSScriptRoot/../../emitter-package.json"
+        Write-Host("Copying package.json from $replacementPackageJson")
+        Copy-Item -Path $replacementPackageJson -Destination "package.json" -Force
+        npm install
```
weshaggard, line 27: I think it might be worth adding the `--no-lock-file` option here to avoid leaving behind a lock file in the source tree as well.

eng/common/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,138 @@
...
+    if (!(Test-Path $sparseCheckoutFile) -or !((Get-Content $sparseCheckoutFile).Contains($subDirectory))) {
+        Write-Output $subDirectory >> .git/info/sparse-checkout
+    }
+}
+
+function CopySpecToProjectIfNeeded([string]$specCloneRoot, [string]$mainSpecDir, [string]$dest, [string[]]$specAdditionalSubDirectories) {
+    $source = "$specCloneRoot/$mainSpecDir"
+    Write-Host "Copying spec from $source"
+    # $mainSpecDir is the PR folder, we just need to copy its subfolders which include the cadl project folder
+    Get-ChildItem –Path "$source" -Exclude @("data-plane", "resource-manager")|
```
weshaggard, line 23: I'm not sure it is worth filtering these directories. I might be worth keeping this in all cases as it shouldn't cause any real issues. These will definitely be needed if we end up using these scripts to run the autorest emitter as well.

eng/common/scripts/Cadl-Project-Generate.ps1
m-nash: It is added to .gitignore and the advantage here is if you want to do some local changes / what if having it right next to your project is convenient

weshaggard (approved):
This looks like a good starting point. One thing we should try to figure out is how we test this when we make changes. We should try to write some test scripts that cover the scenarios so we can be sure we don't break things.

Issue #4723, opened 2022-11-16
APIView - Batch process API review update background task


praveenkuttappan:
APIView automatically upgrades all existing reviews when a parser version is modified. Currently it's run as a background task that fetches all reviews and updates individual revisions. This approach has few limitations.

1. If update fails for one language, then it impacts and stops this background task for all reviews.
2. We don't have a mechanism to enable and disable update of a specific language.
3. We should be able to batch process and update more reviews on each pipeline run when sandboxing is enabled.

PR #5238, opened 2023-01-26
Batch background review update when parser version changes

praveenkuttappan:
API review is updated when parser version changes as part of background task during server startup. This causes too many pipelines running in parallel when running one pipeline for each review to be updated. This PR has changes to:
1. Batch process reviews to be updated with an option to configure batch size
2. Also add an option to disable a specific language from background update task instead of disabling the task for all languages due to an issue in one language.

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/HostedServices/ReviewBackgroundHostedService.cs
+++ b/src/dotnet/APIView/APIViewWeb/HostedServices/ReviewBackgroundHostedService.cs
@@ -1,6 +1,7 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using APIViewWeb.Managers;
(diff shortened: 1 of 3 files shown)


praveenkuttappan:
Fixes https://github.com/Azure/azure-sdk-tools/issues/4723

weshaggard (approved):
Looks reasonable

src/dotnet/APIView/APIViewWeb/HostedServices/ReviewBackgroundHostedService.cs
```diff
@@ -33,6 +36,18 @@ public ReviewBackgroundHostedService(IReviewManager reviewManager, IConfiguratio
             {
                 _autoArchiveInactiveGracePeriodMonths = 4;
             }
+            var backgroundTaskDisabledLangs = configuration["ReviewUpdateDisabledLanguages"];
+            if(!string.IsNullOrEmpty(backgroundTaskDisabledLangs))
```
konrad-jamrozik, line 40:
```suggestion
            if (!string.IsNullOrEmpty(backgroundTaskDisabledLangs))
```

praveenkuttappan: I had missed this comment. I will pull in this change in my next PR. thanks @konrad-jamrozik

PR #5241, opened 2023-01-26
Enable regex-based wildcard-supporting CODEOWNERS matcher for `azure-sdk-for-net` repo

konrad-jamrozik:
This PR is part of an incremental roll-out of:
- #5088
- #2770

It rolls out the new regex-based wildcard-supporting matcher to the `azure-sdk-for-net` repository.

Related preliminary work done to make this PR happen:
- https://github.com/Azure/azure-sdk-for-net/pull/33584
- https://github.com/Azure/azure-sdk-for-net/pull/33595
- https://github.com/Azure/azure-sdk-for-net/pull/33597

### DIFF
--- a/tools/notification-configuration/notification-creator/Contacts.cs
+++ b/tools/notification-configuration/notification-creator/Contacts.cs
@@ -75,7 +75,10 @@ internal class Contacts
             "Searching CODEOWNERS for matching path for '{buildDefinitionFilePath}'",
             buildDefinitionFilePath);
 
-        CodeownersEntry matchingCodeownersEntry = GetMatchingCodeownersEntry(yamlProcess, codeownersEntries);
+        CodeownersEntry matchingCodeownersEntry = GetMatchingCodeownersEntry(
+            yamlProcess,
+            codeownersEntries,
+            repoUrl.ToString());
         List<string> contacts = matchingCodeownersEntry.Owners;
 
         this.log.LogInformation(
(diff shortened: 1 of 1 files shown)


tools/notification-configuration/notification-creator/Contacts.cs
```diff
@@ -109,14 +112,45 @@ private Uri GetCodeownersRepoUrl(BuildDefinition buildDefinition)
...
+    /// Note that enabling the new matcher will not change the reviewers assigned to
+    /// PRs, because this is handled by built-in GitHub matcher. 
+    /// But it will change who receives build failure notifications.
+    ///
+    /// [1] https://github.com/Azure/azure-sdk-tools/pull/5088
+    /// </summary>
+    private bool UseRegexMatcher(string repoUrl)
+    {
+        // Expected repoUrl: https://github.com/Azure/azure-sdk-for-net
+        if (repoUrl.Contains("azure-sdk-for-net"))
```
weshaggard, line 145: Given this is only supposed to be temporary until we transition all the existing codeowners files I'm OK with this here but if we need to keep this longer I'm not sure I like hardcoding this here and we should consider some sort of config or option that gets passed through instead.

konrad-jamrozik: @weshaggard got it. I will keep in mind the preferred way in general is to have proper options. As you noticed, this is temporary, so I will remove this code within few weeks max.

PR #5242, opened 2023-01-26
[test resources] Consolidate naming logic and generate short hash names for local use

### LINKED ISSUE #4931: New-TestResources should warn when overwriting existing env variables with the subscription config
This will make it easier to diagnose and debug implicit issues where someone defines the env variables twice, once in the yaml or local shell, and again from the sub config+arm template outputs.

CC @christothes

benbp:
Follow up fix related to discussion on https://github.com/Azure/azure-sdk-for-net/pull/33262

- Make name generation simpler and more consistent across test resource scripts (found a bug where remove-testresources.ps1 would not have worked for nested service directories like data/aztables)
- Use a short hash for resource names deployed locally, to avoid edge cases where a long service directory or username causes resource names to be too long for validation (e.g. keyvault).
- Fix issue with `EnvironmentVariables` ref getting updated from the parameters when splatting in CI mode and polluting the env variable outputs on subsequent runs in the same shell.
- Add more warning logging if we're overwriting environment variables, for #4931

Fixes #4931

### DIFF
--- a/eng/common/TestResources/SubConfig-Helpers.ps1
+++ b/eng/common/TestResources/SubConfig-Helpers.ps1
@@ -16,10 +16,40 @@ function GetUserName() {
     return $UserName
 }
 
-function GetBaseName([string]$user, [string]$serviceDirectoryName) {
+function GetBaseAndResourceGroupNames(
+    [string]$baseNameDefault,
+    [string]$resourceGroupNameDefault,
+    [string]$user,
+    [string]$serviceDirectoryName,
+    [bool]$CI
+) {
+    if ($CI) {
+        $base = 't' + (New-Guid).ToString('n').Substring(0, 16)
+        # Format the resource group name based on resource group naming recommendations and limitations.
+        $generatedGroup = "rg-{0}-$BaseName" -f ($serviceName -replace '[\.\\\/:]', '-').ToLowerInvariant().Substring(0, [Math]::Min($serviceName.Length, 90 - $BaseName.Length - 4)).Trim('-')
+        $group = $resourceGroupNameDefault ? $resourceGroupNameDefault : $generatedGroup
+
+        Log "Generated resource base name '$base' and resource group name '$group' for CI build"
+
+        return $base, $group
+    }
+
+    if ($baseNameDefault) {
+        $base = $baseNameDefault.ToLowerInvariant()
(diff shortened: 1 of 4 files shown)


eng/common/TestResources/SubConfig-Helpers.ps1
```diff
@@ -16,10 +16,40 @@ function GetUserName() {
...
+        $group = $resourceGroupNameDefault ? $resourceGroupNameDefault : ("rg-$baseNameDefault".ToLowerInvariant())
+        return $base, $group
+    }
+
     # Handle service directories in nested directories, e.g. `data/aztables`
     $serviceDirectorySafeName = $serviceDirectoryName -replace '[\./\\]', ''
-    return "$user$serviceDirectorySafeName".ToLowerInvariant()
+    # Hash to keep resource names short enough to not break naming restrictions (e.g. keyvault name length)
+    $baseNameStream = [IO.MemoryStream]::new([Text.Encoding]::UTF8.GetBytes("${user}${serviceDirectorySafeName}"))
+    $base = 't' + (Get-FileHash -InputStream $baseNameStream -Algorithm SHA1).Hash.Substring(0, 16).ToLowerInvariant()
```
heaths, line 47: For longevity since SHA1 is about to get flagged, you might use `[SHA256]` now instead. Either way, we're cutting entropy (even for MD5) so it doesn't really matter, but scanners will flag use of SHA1 sooner or later no matter the reason (we saw this *years* ago in WiX using MD5 to generate seeded GUIDs).

eng/common/TestResources/build-test-resource-config.yml
```diff
@@ -36,3 +36,4 @@ steps:
           UpdateSubscriptionConfiguration $configBase $config
 
         displayName: Merge Test Resource Configurations
+)
```
heaths, line 39: Bad yaml. I mean, badder than normal yaml.

heaths (approved):

eng/common/TestResources/SubConfig-Helpers.ps1
benbp: Ah good point. Updated.

benbp:
@heaths @weshaggard I marked this ready for review now. If you look at the commits I'm rolling in a couple bug fixes and quality of life updates here.

weshaggard (approved):
Looks reasonable to me

PR #5283, opened 2023-01-31
Revert naming `test-proxy`

scbedd:
Save ourselves the downstream pain of:
- [x] not aligning with the namespace (which plays hell with the .net usage of this thing)
- [x] needing to worry about name squatters (generic `test-proxy` being the name :|)
- [x] fixing the numerous other places I have probably broken stuff by the rename

~~Just...just save myself the pain.~~ 

**spoiler alert**

I did not save myself anything unfortunately. I have discovered all of these issues with the latest eng/common update that involves this. Reverting, updating that eng/common one last time, and hopefully being done with this.


scbedd:
Real talk though I think this was an ill-advised change to make the standalone executable a bit nicer to use. At the end of the day I don't think that anyone will really care about the difference between `test-proxy` and `azure.sdk.tools.testproxy` when it's the same binary.

scbedd:
This is going to require another update of `azure-sdk-build-tools` to fix, due to the fact that the method of detecting the binary is flawed.

It's my code so I can't be sad about it.

benbp (approved):

eng/pipelines/templates/scripts/assemble-dotnet-standalone-exe.ps1
```diff
@@ -86,7 +89,7 @@ else {
 }
 
 # clean up the uncompressed artifact directory
-Remove-Item -Recurse -Force -Path $outputPath
+# Remove-Item -Recurse -Force -Path $outputPath
```
scbedd, line 92:
```suggestion
Remove-Item -Recurse -Force -Path $outputPath
```

scbedd:
This is going to come through the same avenue as a previous build that required `azure-sdk-build-tools` update. Will kludge these changes into the automatically submitted sdk-build-tools reference update. Closing.

scbedd:
This PR is now reflected [in the sdk-build-tools update PR.](https://github.com/Azure/azure-sdk-tools/pull/5283/commits)

PR #5291, opened 2023-01-31
[test proxy] Move migration script to eng/common

timovv:
For ease of use when migrating over to the new assets repo.

Cc @scbedd @HarshaNalluru


scbedd (approved):
@timovv this looks great. I didn't see any places where you didn't update the location of the script.

Thank you for moving this, now approve the first step of the eng/common process, which will create the sync PRs.

On those same PRs, we will want to add little commits to update documentation referencing the old location of this script. I will help with that, but let me know when those PRs are created.

HarshaNalluru (approved):

PR #5298, opened 2023-02-01
[testmodeler] upgrade dependencies

changlong-liu:


changlong-liu:
Hi @tadelesh , I find there are some new properties added in the apiscenario after upgrading to oav@3.2.4, would you help to confirm whether it breaks Golang testGen, I guess it doesn't?

tadelesh:
> Hi @tadelesh , I find there are some new properties added in the apiscenario after upgrading to oav@3.2.4, would you help to confirm whether it breaks Golang testGen, I guess it doesn't?

Additional change will not break Go test generator. Thanks for reminder. I'll do some test when upgrading testmodeler.

mikeharder (approved):
The changes in `package.json` look good to me.  I can't speak to the changes in `model.ts` or the `*.yaml` files.

PR #5304, opened 2023-02-01
[Test-Proxy] During recording, re-use existing request body

scbedd:
Context:

We are seeing an _extremely_ odd behavior during recording of monitor ingestion tests. Specifically the upload of logs.

When running from within the .NET test framework, some of the requests that look this:

```
[18:45:08] dbug: DebugLogging[0]
      Request URI: [https://<monitorUrl>/dataCollectionRules/dcr-3241746bf36d4042b01e60adc3982153/streams/Custom-MyTableRawData?api-version=2021-11-01-preview]
      Request method: [POST]
      Request headers: [[{"Key":"Accept","Value":["application/json"]},{"Key":"User-Agent","Value":["azsdk-net-Monitor.Ingestion/1.0.0-alpha.20230131.1","(.NET 6.0.13; Microsoft Windows 10.0.22621)"]},{"Key":"Authorization","Value":["Bearer <snip>"]},{"Key":"traceparent","Value":[""]},{"Key":"x-ms-client-request-id","Value":["ec222bca52f1be17a31425a0bb4c9514"]},{"Key":"x-ms-return-client-request-id","Value":["true"]}]]
```

**Randomly**   400 with `badrequest`

```
[18:45:08] dbug: DebugLogging[0]
      Response Statuscode: [ BadRequest]
      Response Headers: [[{"Key":"Server","Value":["Microsoft-HTTPAPI/2.0"]},{"Key":"Date","Value":["Wed, 01 Feb 2023 02:45:09 GMT"]},{"Key":"Connection","Value":["close"]}]]
```

@JoshLove-msft is suspicious that gzip may be non-deterministic when compressing/recompressing. Not that the content itself is an issue, but the `Content-Length` headers WOULD be an issue if this were the case. If the header doesn't match by a couple bytes, its very possible that the service could throw back a 400 bad request at us.

What we'll do here is bypass that issue by merely keeping the original request bytes around and essentially paste them into the upstream request.

I'm somewhat concerned that if it can happen there, why wouldn't it happen during playback as well (where we HAVE to recompress as we need to decompress to sanitize). Worth a try though.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
@@ -194,10 +194,9 @@ namespace Azure.Sdk.Tools.TestProxy
                 throw new HttpException(HttpStatusCode.BadRequest, $"There is no active recording session under id {recordingId}.");
             }
 
-            (RecordEntry entry, byte[] bytes) = await CreateEntryAsync(incomingRequest).ConfigureAwait(false);
-            var entry = entryTuple.Item1;
+            (RecordEntry entry, byte[] requestBody) = await CreateEntryAsync(incomingRequest).ConfigureAwait(false);
 
-            var upstreamRequest = CreateUpstreamRequest(incomingRequest, entryTuple.Item2);
+            var upstreamRequest = CreateUpstreamRequest(incomingRequest, requestBody);
 
             HttpResponseMessage upstreamResponse = null;
 
@@ -471,7 +470,7 @@ namespace Azure.Sdk.Tools.TestProxy
             }
         }
 
-        public static async Task<Tuple<RecordEntry, byte[]>> CreateEntryAsync(HttpRequest request)
+        public static async Task<(RecordEntry, byte[])> CreateEntryAsync(HttpRequest request)
         {
             var entry = new RecordEntry();
             entry.RequestUri = GetRequestUri(request).AbsoluteUri;
@@ -488,7 +487,7 @@ namespace Azure.Sdk.Tools.TestProxy
             byte[] bytes = await ReadAllBytes(request.Body).ConfigureAwait(false);
 
             entry.Request.Body = CompressionUtilities.DecompressBody(bytes, request.Headers);
-            return new Tuple<RecordEntry, byte[]>(entry, bytes);
+            return (entry, bytes);
         }
 
         #endregion
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
@@ -194,9 +194,10 @@ namespace Azure.Sdk.Tools.TestProxy
                 throw new HttpException(HttpStatusCode.BadRequest, $"There is no active recording session under id {recordingId}.");
             }
 
-            var entry = await CreateEntryAsync(incomingRequest).ConfigureAwait(false);
+            var entryTuple = await CreateEntryAsync(incomingRequest).ConfigureAwait(false);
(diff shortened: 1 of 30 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -194,9 +194,10 @@ public async Task HandleRecordRequestAsync(string recordingId, HttpRequest incom
                 throw new HttpException(HttpStatusCode.BadRequest, $"There is no active recording session under id {recordingId}.");
             }
 
-            var entry = await CreateEntryAsync(incomingRequest).ConfigureAwait(false);
+            var entryTuple = await CreateEntryAsync(incomingRequest).ConfigureAwait(false);
```
JoshLove-msft, line 197: Can do something like this:
```suggestion
            (RecordEntry entry, byte[] bytes) = await CreateEntryAsync(incomingRequest).ConfigureAwait(false);
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -470,7 +471,7 @@ public async Task HandlePlaybackRequest(string recordingId, HttpRequest incoming
             }
         }
 
-        public static async Task<RecordEntry> CreateEntryAsync(HttpRequest request)
+        public static async Task<Tuple<RecordEntry, byte[]>> CreateEntryAsync(HttpRequest request)
```
JoshLove-msft, line 474: This will use a ValueTuple rather than a Tuple to avoid allocating
```suggestion
        public static async Task<(RecordEntry, byte[])> CreateEntryAsync(HttpRequest request)
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -487,7 +488,7 @@ public static async Task<RecordEntry> CreateEntryAsync(HttpRequest request)
             byte[] bytes = await ReadAllBytes(request.Body).ConfigureAwait(false);
 
             entry.Request.Body = CompressionUtilities.DecompressBody(bytes, request.Headers);
-            return entry;
+            return new Tuple<RecordEntry, byte[]>(entry, bytes);
```
JoshLove-msft, line 491:
```suggestion
            return (entry, bytes);
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: committed basically exactly this

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: Yeah my commit included this one :)

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -487,7 +487,7 @@ public static async Task<RecordEntry> CreateEntryAsync(HttpRequest request)
             byte[] bytes = await ReadAllBytes(request.Body).ConfigureAwait(false);
 
             entry.Request.Body = CompressionUtilities.DecompressBody(bytes, request.Headers);
```
JoshLove-msft, line 489: The bytes are not mutated by DecompressBody, right?

scbedd: Sure.

JoshLove-msft (approved):

PR #5308, opened 2023-02-01
Fix suffix `*` handling in our `CODEOWNERS` path matcher

konrad-jamrozik:
As in subject. Prompted by this comment thread:

- https://github.com/Azure/azure-sdk-for-python/pull/28539#discussion_r1093762875

There were some other issues with the matcher being inconsistent with GitHub's matcher. For details, please see this comment:
- https://github.com/Azure/azure-sdk-tools/pull/5308#issuecomment-1413228282

This PR contributes to:

- #2770


tools/code-owners-parser/Azure.Sdk.Tools.CodeOwnersParser.Tests/CodeownersFileTests.cs
```diff
@@ -83,8 +83,8 @@ public class CodeownersFileTests
         new(        "/a" , "*"           ,  false , false ),
         new(        "/*" , "*"           ,   true , false ),
         new(        "/*" , "a"           ,  false , true  ),
-        new(        "/*" , "a/"          ,  false , false ),
-        new(        "/*" , "a/b"         ,  false , false ),
+        new(        "/*" , "a/"          ,  false , true  ),
+        new(        "/*" , "a/b"         ,  false , true  ),
```
konrad-jamrozik, line 87: Note to self: double check GH behavior on this case before merging

weshaggard (approved):

konrad-jamrozik:
@weshaggard I made some significant changes once you approved; turns out, GitHubs CODEOWNERS matching logic is a bit inconsistent. I did manual tests using [my own scratchpad repo](https://github.com/konrad-jamrozik/scratchpad) to confirm its behavior. This is explained [here](https://github.com/Azure/azure-sdk-tools/blob/2b7d12d767e2dee5fdd3d86ac14ab3dbde5cbbd1/tools/code-owners-parser/CodeOwnersParser/MatchedCodeownersEntry.cs#L250-L280).

In a nutshell:

- For top level dir, `/` doesn't work. One has to use `/**`.
- But for nested dirs, it works. I.e. one can write `/foo/` and it is equivalent to `/foo/**`.
- `*` has different interpretations. If used with preceding slash, like `/*` or `/foo/*` it means "things only in this dir. But when used as a suffix, it means "anything". So `/foo*` is effectively `/foo*/** OR /foo*`.

tools/code-owners-parser/CodeOwnersParser/MatchedCodeownersEntry.cs
```diff
@@ -136,21 +140,41 @@ private static bool ContainsUnsupportedCharacters(string codeownersPath)
...
-                {
-                    Console.Error.WriteLine(
-                        $"CODEOWNERS path \"{codeownersPath}\" ends with " +
-                        $"unsupported sequence of \"{sequence}\". Replace it with \"/\". " +
-                        "Until then this path will never match.");
-                    return true;
-                }                
+                Console.Error.WriteLine(
+                    $"CODEOWNERS path \"{codeownersPath}\" ends with " +
+                    "unsupported sequence of \"/**/\". Replace it with \"/**\".");
```
weshaggard, line 176:
```suggestion
                    "unsupported sequence of \"/**/\". Replace it with \"/\".");
```
Why not just suggest to end with `/`

weshaggard (approved):

tools/code-owners-parser/CodeOwnersParser/MatchedCodeownersEntry.cs
konrad-jamrozik: @weshaggard per [this comment](https://github.com/Azure/azure-sdk-tools/pull/5308#issuecomment-1413228282), this is a special case:

image

weshaggard: I agree for that one edge case but what about a path like `/foo/bar/**/` if they change to `/foo/bar/**` then your other rule above will trigger on that and tell them to change to `/foo/bar`.

konrad-jamrozik: > /foo/bar/**/

Yeah, I also realized that, and have since fixed it in:

- https://github.com/Azure/azure-sdk-tools/pull/5325

Thanks!

PR #5309, opened 2023-02-01
[stress testing] Enable nodepool update via bicep and update node SKUs

### LINKED ISSUE #4830: [stress testing] Update node image versions
https://learn.microsoft.com/en-us/azure/aks/node-image-upgrade

benbp:
- Updating stress cluster node SKUs - use d4v4 everywhere, separate system from user pools, and add stress-watcher to the system pool
- Enable updating stress cluster nodepools via bicep deployment via an `-UpdateNodes` switch.
- Remove warnings from bicep build
- Update documentation for pool SKU label selectors
- Minor provision script fixes: missing object id and custom environment parameter validation

Fixes #4830


weshaggard (approved):
looks reasonable

Issue #4641, opened 2022-11-07
APIView - Convert .Net API review parser to a standalone tool


praveenkuttappan:
Currently API review parser for .Net is included as part of core APIView.dll. We need to create a command line tool which takes path to .Nuget or .dll .Net package library as input and generate api review token file in a given output path. This is required to enable sandboxing for .Net API review gen. Sandboxing will allow APIView to run Nuget to token file conversion using pipeline instead of running within APIView server.

PR #5310, opened 2023-02-01
[.NET] Updating ApiView to handle input/output files

kinelski:
Fixes https://github.com/Azure/azure-sdk-tools/issues/4641.


kinelski:
Hello @praveenkuttappan,

These changes should address issue https://github.com/Azure/azure-sdk-tools/issues/4641. Could you check if the generated output file is what we want?

This is the actual JSON file generated by the console app (not formatted): https://github.com/kinelski/azure-sdk-tools/blob/av-output/src/dotnet/APIView/APIViewUnitTests/SampleTestFiles/apiview.json

This is the JSON output after formatting (formatting done manually): https://github.com/kinelski/azure-sdk-tools/blob/av-output/src/dotnet/APIView/APIViewUnitTests/SampleTestFiles/apiview_format.json

pallavit:
@praveenkuttappan , @maririos Could you please take a look at the change when you get a chance. We want to ensure this solves the sandboxing issue for .NET.

src/dotnet/APIView/APIView/App.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace ApiView
+{
+    public static class App
```
JoshLove-msft, line 7: Why not just use Program.cs?

src/dotnet/APIView/APIViewUnitTests/AppTests.cs
```diff
@@ -0,0 +1,77 @@
...
+                Path.Combine("missing_folder", "output.json")
+            };
+
+            Assert.ThrowsAsync<FolderNotFoundException>(async () => await App.RunAsync(args));
+        }
+
+        [Fact]
+        public async Task RunAsyncGeneratesOutputFile()
+        {
+            var args = new string[] { InputPath, OutputPath };
```
JoshLove-msft, line 60: Do we need to clean up the output file?

JoshLove-msft (approved):

src/dotnet/APIView/APIViewUnitTests/AppTests.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Threading.Tasks;
+using ApiView;
+using Microsoft.TeamFoundation.Build.WebApi;
+using Xunit;
+
+namespace APIViewUnitTests
+{
+    public class AppTests
+    {
+        private string InputPath => Path.Combine("SampleTestFiles", "APIView.dll");
```
praveenkuttappan, line 13: What is this APIVIew.dll? Why do we need to check in this binary?

praveenkuttappan: Can't we directly run using local project path?

kinelski: The `ApiView.dll` is just a dll I'm using for testing (we could use any binary). I added a test where we use this dll as an input to the console app and check if it produces the result we expect.

JoshLove-msft: I think the point is that it would be better to have a human-readable artifact like a csharp class rather than the dll. The test could handle compiling the class using the CompilationFactory.

praveenkuttappan: @kinelski Can you please make the change as Josh suggested for test file instead of dll?

src/dotnet/APIView/APIView/App.cs
kinelski: I swear it made sense at some point. Moved `RunAsync` to `Program`.

src/dotnet/APIView/APIViewUnitTests/AppTests.cs
kinelski: Done. I'm consuming the asset files in the `ExactFormatting` folder now since they already cover multiple scenarios.

JoshLove-msft (approved):

PR #5313, opened 2023-02-02
Handle multi-line string concatenation with Write-Warning

benbp:


kristapratico (approved):

weshaggard (approved):

heaths (approved):
Interesting. Didn't think that wouldn't work.

benbp:
Me neither! It gives the following error:

```
Write-Warning: A positional parameter cannot be found that accepts argument '+'.
```

PR #5316, opened 2023-02-02
Temporarily disable error handling for Get-AzKeyVault -InRemovedState

benbp:
Follow-up tracked in https://github.com/Azure/azure-sdk-tools/pull/5316

### DIFF
--- a/eng/common/scripts/Helpers/Resource-Helpers.ps1
+++ b/eng/common/scripts/Helpers/Resource-Helpers.ps1
@@ -75,15 +75,23 @@ function Get-PurgeableResources {
 
   Write-Verbose "Retrieving deleted Key Vaults from subscription $subscriptionId"
 
-  # Get deleted Key Vaults for the current subscription.
-  $deletedKeyVaults = @(Get-AzKeyVault -InRemovedState `
-    | Add-Member -MemberType NoteProperty -Name AzsdkResourceType -Value 'Key Vault' -PassThru `
-    | Add-Member -MemberType AliasProperty -Name AzsdkName -Value VaultName -PassThru)
+  # TODO: Remove try/catch handler for Get-AzKeyVault - https://github.com/Azure/azure-sdk-tools/issues/5315
+  # This is a temporary workaround since Az module >= 9.2.0 uses a more recent API
+  # version than is supported in the dogfood cloud environment:
+  #
+  #   | The resource type 'deletedVaults' could not be found in the namespace 'Microsoft.KeyVault' for api version '2022-07-01'. The supported api-versions are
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Helpers/Resource-Helpers.ps1
```diff
@@ -75,14 +75,24 @@ function Get-PurgeableResources {
...
 
-  if ($deletedKeyVaults) {
-    Write-Verbose "Found $($deletedKeyVaults.Count) deleted Key Vaults to potentially purge."
-    $purgeableResources += $deletedKeyVaults
+    if ($deletedKeyVaults) {
+      Write-Verbose "Found $($deletedKeyVaults.Count) deleted Key Vaults to potentially purge."
+      $purgeableResources += $deletedKeyVaults
+    }
+  } catch {
+    return $purgeableResources
```
mikeharder, line 95: Do we need an extra return here?  If the catch block is a no-op, won't execution just fall through to the return below?

mikeharder (changes requested):

mikeharder (approved):

eng/common/scripts/Helpers/Resource-Helpers.ps1
benbp: I can make an empty catch block, that's cleaner.

eng/common/scripts/Helpers/Resource-Helpers.ps1
```diff
@@ -75,14 +75,24 @@ function Get-PurgeableResources {
...
-    | Add-Member -MemberType AliasProperty -Name AzsdkName -Value VaultName -PassThru)
+  # TODO: Remove try/catch handler for Get-AzKeyVault - https://github.com/Azure/azure-sdk-tools/issues/5315
+  # This is a temporary workaround since Az module >= 9.2.0 uses a more recent API
+  # version than is supported in the dogfood cloud environment:
+  #
+  #   | The resource type 'deletedVaults' could not be found in the namespace 'Microsoft.KeyVault' for api version '2022-07-01'. The supported api-versions are
+  #   | '2016-10-01,2018-02-14-preview,2018-02-14,2019-09-01,2021-04-01-preview,2021-06-01-preview,2021-10-01,2021-11-01-preview'.
+  try {
+    # Get deleted Key Vaults for the current subscription.
+    $deletedKeyVaults = @(Get-AzKeyVault -InRemovedState `
```
weshaggard, line 86: Is there any way we can query to see if the api version exists or not?

benbp: We can but at that point we may as well pin the Az module version to an older version (which we should discuss tomorrow as an approach regardless), because knowing which api version will be queried under the hood requires knowing the Az.KeyVault module version we're using.

PR #5329, opened 2023-02-03
Register preview features in stress cluster

benbp:
Adds a step to install preview features in the cluster (not available via bicep yet). First feature being added is the [image cleaner](https://learn.microsoft.com/en-us/azure/aks/image-cleaner?tabs=azure-cli) which removes stale/vulnerable images from the container image cache on cluster nodes.


weshaggard (approved):

benbp:
Part of a few fixes for #4827

PR #5330, opened 2023-02-03
Fix regex-based matcher to be a strict subset of GitHub-based `CODEOWNERS` matcher.

konrad-jamrozik:
As in subject. This PR is a follow-up to:
- #5308

This PR contributes to:
- #2770 

Most notably, now the matcher is a strict subset of GitHub CODEOWNERS matcher. That is, for each CODEOWNERS path our matcher should behave exactly the same as GitHub CODEOWNERS path matcher, except when that path violates additional validation imposed by us (see the `MatchedCodeownersEntry` class comments for details). In such case, given CODEOWNERS path is skipped (ignored) by our matcher, never matching any path.

Note that to make the matcher be a subset of the GitHub matcher, it can no longer normalize paths missing prefix `/` by adding it. This change was made in this PR.


tools/code-owners-parser/CodeOwnersParser/MatchedCodeownersEntry.cs
```diff
@@ -282,9 +282,16 @@ private static string SetPatternSuffix(string pattern)
                 return pattern;
 
             // If the pattern doesn't end with "/" nor "*", it is a path to a file,
-            // hence we must append "$", to avoid treating the regex pattern as a
-            // prefix match.
-            return pattern + "$";
+            // or to a directory with exact match, hence the suffix must be:
```
weshaggard, line 285: While github may allow this I thought we were trying to be stricter and always require a trailing slash to match a directory and all its children.

konrad-jamrozik: That is correct, but!

- I do not have any automated validation for this restriction yet; the matcher would need to know if there is a file or directory at given path, but currently it doesn't read repo contents except the CODEOWNERS file.
- Even once I implement the validation, I believe I should keep it separate from the matcher, to retain high cohesion and separation of responsibilities. This way it will be easier to reason about the implementation behavior: one can say "our matcher behaves exactly like GitHub matcher plus there is a validation layer that restricts valid paths, thus pre-filtering the inputs before they reach the matcher". Otherwise it would be something like "the matcher behaves like GitHub matcher but with some behavior differences, because it assumes its inputs have been pre-validated, but it cannot check it, as it would have to query the repo contents, which it doesn't right now".

konrad-jamrozik: @weshaggard you are correct. I was going a bit back and forth how I wan t to handle invalid paths, but you prompted me to clarify the matcher approach - I have updated the code, as well as the PR title and description. The matcher now behaves the way you said. In general, it is now a subset of GitHub CODEOWNERS path matcher.

weshaggard (approved):

Issue #4734, opened 2022-11-17
In Go APIView, when click the parameter definition, it is unable to jump to the correct location.


lirenhe:
In https://apiview.dev/Assemblies/Review/518fb41eeb5447ea850c25d28cf9e7dc

For func (client *Artifact) [DeleteTag](ctx context.Context, tag string, options *DeleteTagOptions) error

When click DeleteTagOptions, we are unable to jump to its definition 

type [DeleteTagOptions]generated.ContainerRegistryClientDeleteTagOptions

Expect: It should jump to the correct location.

PR #5346, opened 2023-02-06
Add type link for Go API view

tadelesh:
Main process:
1. When parsing the AST, try to figure out all the types import path:
- type in the same package or module, add link prefix <navigator> to the type string
- type in different module or system type, do nothing
2. When generating token list, add link for all types' token

Resolve: https://github.com/Azure/azure-sdk-tools/issues/4734


tadelesh:
@jhendrixMSFT @chlowell Is there any best practice to do the testing for Go API view?

jhendrixMSFT:
Can you please share a link to an APIView generated with these changes?

tadelesh:
> Can you please share a link to an APIView generated with these changes?

I don't know how to deploy the updated code to the [staging env](https://apiviewstaging.azurewebsites.net). I could only set up a local test env and you could refer this generated page: [azcore - apiview.dev.zip](https://github.com/Azure/azure-sdk-tools/files/10682261/azcore.-.apiview.dev.zip) from my local env.

jhendrixMSFT:
No need to update staging, just create the JSON file locally and upload it as part of "Create + Review" process.

tadelesh:
> No need to update staging, just create the JSON file locally and upload it as part of "Create + Review" process.

New one: https://apiviewstaging.azurewebsites.net/Assemblies/Review/6f329f3801d544179f8508217726d44e
Old one: https://apiviewstaging.azurewebsites.net/Assemblies/Review/fa8d33b6ab2e4df19b19b1ba421fa4c5

chlowell:
I like the navigation links but I'm unsure more qualification of type names helps readers. For example, when I'm looking at `azcore`, if I didn't know that `runtime.PipelineOptions` is in the module, the navigation link alone would tell me that. I'm particularly doubtful about absolute import paths. Will this feature replace every `azcore.TokenCredential` with `github.com/Azure/azure-sdk-for-go/sdk/azcore.TokenCredential`?

tadelesh:
My original idea is to avoid ambiguous type name like two `policy.ClientOptions` (`azcore/policy.ClientOptions`, `azcore/arm/policy.ClientOptions`) and replace import alias with original name. But I agree current way is not good. How about we add all import statements after the package declaration.

jhendrixMSFT:
The type navigation is great (I assume this impacts only the links on the left column).

I'm unclear on your other change about alias in the same module.  Things look the same to me.  Can you please clarify?

tadelesh:
> The type navigation is great (I assume this impacts only the links on the left column).
> 
> I'm unclear on your other change about alias in the same module. Things look the same to me. Can you please clarify?

Maybe link is more clear than navigator. This PR is to add links for all the types of struct fields, method/function param/return value and type alias. With the link, you could easily jump to the type definition when you want to know the detail. For example, you could click `azruntime.PipelineOptions` to jump to the struct definition in `azcore/runtime` subpackage as follows.
image
image
I'll only add link for the type in the same module. Types in other modules or types in go language modules will not have links.

jhendrixMSFT:
@JeffreyRichter if you want to see the new change

https://apiviewstaging.azurewebsites.net/Assemblies/Review/409a829c9c394ba292eef8a2465a9ad0

src/go/cmd/pkg.go
```diff
@@ -27,6 +29,7 @@ var ErrNoPackages = errors.New("no packages found")
 
 // Pkg represents a Go package.
 type Pkg struct {
+	ModulePath  string
```
jhendrixMSFT, line 32: Nit: Is there a reason why this is exported compared to the other fields?  If not, can you please unexport it for consistency.

src/go/cmd/pkg.go
```diff
@@ -46,8 +49,9 @@ type Pkg struct {
 
 // NewPkg loads the package in the specified directory.
 // It's required there is only one package in the directory.
-func NewPkg(dir, moduleName string) (*Pkg, error) {
+func NewPkg(dir, moduleName, modulePath string) (*Pkg, error) {
```
jhendrixMSFT, line 52: Doesn't `moduleName` become duplicate now that we're specifying the full module path (which I assume is really the module identity)?

src/go/cmd/content.go
```diff
@@ -57,14 +57,14 @@ func (c content) isEmpty() bool {
 }
 
 // adds const and var declaration to the exports list
-func (c *content) addGenDecl(pkg Pkg, tok token.Token, vs *ast.ValueSpec) Declaration {
+func (c *content) addGenDecl(pkg Pkg, tok token.Token, vs *ast.ValueSpec, imports map[string]string) Declaration {
```
jhendrixMSFT, line 60: The `imports` param appears to be optional.  If this is the case, it would be good to add a comment explaining when it's needed (applies to all affected `content` methods).

src/go/cmd/pkg.go
tadelesh: Removed. You also remind me to fix the possible problem for v2+ version packages.

tadelesh:
@jhendrixMSFT @JeffreyRichter do you have any further concern about this feature?

JeffreyRichter:
I have no context on this. You are adding a feature to APIView that allows someone to clink on a type name and go to the definition of that type, right?
If so, this sounds great to me!

tadelesh:
> I have no context on this. You are adding a feature to APIView that allows someone to clink on a type name and go to the definition of that type, right? If so, this sounds great to me!

Yes, exactly. Thanks for taking a look!

jhendrixMSFT (approved):

PR #5347, opened 2023-02-06
[testmodeler] All devDependencies after webpack

changlong-liu:
1. move all deps to devDependencies.
2. add 'source-map-support' for easy debugging, which is also used in autorest core: https://github.com/Azure/autorest/blob/main/packages/libs/codegen/package.json#L53
3. make webpack.config.js options same with autorest
4. replace `jsonpath` with `jsonpath-plus` since find `jasonpath` has problem working with webpack, and the `jsonpath-plus` is also used in autorest lib https://github.com/Azure/autorest/blob/main/packages/libs/datastore/package.json#L63


mikeharder (approved):
LGTM

tadelesh (approved):

PR #5366, opened 2023-02-08
Bugfix: when looking for a match in `CODEOWNERS`, do not emit error message for commented out paths

konrad-jamrozik:
Before this change, commented out paths in `CODEOWNERS` violated the rule that the paths should start with `/`. As a result, the `Console.Err` was spammed with error messages, resulting in `Console.Out` truncation for `azure-sdk-for-js` and `azure-sdk-for-java` repos, making it impossible to deserialize the result written to `Console.Out` into valid json. This broke `CodeownersManualAnalysisTests` with `OutOfMemoryException` upon getting value from `Console.Err`.

This bug was introduced by:
- #5330

as that PR stopped prepending missing `/` and instead started outputting message to `Console.Err` if they were missing.

## Secondary fix

This PR also fixes a bug related to validating paths.

### DIFF
--- a/tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners.Tests/CodeownersManualAnalysisTests.cs
+++ b/tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners.Tests/CodeownersManualAnalysisTests.cs
@@ -37,7 +37,7 @@ namespace Azure.Sdk.Tools.RetrieveCodeOwners.Tests;
 /// Enable the new, regex-based, wildcard-supporting CODEOWNERS matcher
 /// https://github.com/Azure/azure-sdk-tools/pull/5088
 /// </summary>
-[TestFixture]//(Ignore = "Tools to be used manually")]
+[TestFixture(Ignore = "Tools to be used manually")]
 public class CodeownersManualAnalysisTests
 {
     private const string OwnersDiffOutputPathSuffix = "_owners_diff.csv";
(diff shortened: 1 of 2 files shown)


tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners.Tests/CodeownersManualAnalysisTests.cs
```diff
@@ -37,7 +37,7 @@ namespace Azure.Sdk.Tools.RetrieveCodeOwners.Tests;
 /// Enable the new, regex-based, wildcard-supporting CODEOWNERS matcher
 /// https://github.com/Azure/azure-sdk-tools/pull/5088
 /// </summary>
-[TestFixture(Ignore = "Tools to be used manually")]
+[TestFixture]//(Ignore = "Tools to be used manually")]
```
konrad-jamrozik, line 40:
```suggestion
[TestFixture(Ignore = "Tools to be used manually")]
```

tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners.Tests/CodeownersManualAnalysisTests.cs
```diff
@@ -352,6 +352,7 @@ private void WriteLangRepoOwnersDiffToCsv(string langName)
         outputLines.AddRange(PathsWithMissingPrefixSlash(entries));
         outputLines.AddRange(PathsWithMissingSuffixSlash(targetDir, entries, paths));
         outputLines.AddRange(InvalidPaths(entries));
+        // TODO: add a check here for CODEOWNERS paths that do not match any dir or file.
```
benbp, line 355: Do we need an issue tracking this?

benbp (approved):

tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners.Tests/CodeownersManualAnalysisTests.cs
konrad-jamrozik: @benbp not really. I put it here more as a reminder that this validation is currently missing, in the unlikely scenario I (or somebody else) will have to use it again. I forgot at first to include it. But I kind of expect for it to be actually picked up when working on:
- https://github.com/Azure/azure-sdk-tools/issues/4859

PR #5369, opened 2023-02-08
[JS] Capture profiling artifacts

HarshaNalluru:
This PR 
- allows capturing the profiling artifacts generated by the perf framework in the JS perf pipelines.
- provides the names of the artifacts.

The counterpart in the JS repo https://github.com/Azure/azure-sdk-for-js/pull/24240

Validation: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2201082&view=results


tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/JavaScript.cs
```diff
@@ -140,6 +142,11 @@ public class JavaScript : LanguageBase
 
             var projectDirectory = Path.Combine(WorkingDirectory, project);
 
+            if(profile){
+                var stripPackageName = primaryPackage.Split(new char[] { '/' })[1];
```
HarshaNalluru, line 146:
```suggestion
                var stripPackageName = primaryPackage.Split(new char[] { '/' })[1]; // "@azure/storage-blob" -> "storage-blob"
```

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/ILanguage.cs
```diff
@@ -26,6 +26,7 @@ public interface ILanguage
             string testName,
             string arguments,
             bool profile,
-            object context);
+            object context,
+            int iteration);
```
mikeharder, line 30:
```suggestion
            object context);
```

Unfortunately, I don't think passing the `iteration` is actually sufficient to disambiguate, because the same test can be run with different arguments.  So the profiles for `download(10KB)` would conflict with `download(10MB)`.  If you want a counter, I would increment a static integer like `Java.cs`.  Another option would be to try to include the arguments in the filename, but this might get very long (e.g. `download-size-10240-parallel-64-1`).

In practice, people usually collect profiles for one test and one set of arguments at a time, but I wouldn't want conflicts if someone did try to collect profiles for multiple arguments in a single run.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/JavaScript.cs
```diff
@@ -140,6 +142,11 @@ public class JavaScript : LanguageBase
 
             var projectDirectory = Path.Combine(WorkingDirectory, project);
 
+            if(profile){
```
mikeharder, line 145:
```suggestion
            if (profile)
            {
```

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/JavaScript.cs
```diff
@@ -140,6 +142,11 @@ public class JavaScript : LanguageBase
...
+                var stripPackageName = primaryPackage.Split(new char[] { '/' })[1]; // "@azure/storage-blob" -> "storage-blob"
```
mikeharder, line 146:
```suggestion
                var stripPackageName = primaryPackage.Substring(primaryPackage.LastIndexOf('/') + 1); // "@azure/storage-blob" -> "storage-blob"
```

This code is slightly better, because it also handles packages without a namespace (e.g. `mocha`).  `"mocha".LastIndexOf('/')` will return `-1`, so this will keep the whole name.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/JavaScript.cs
```diff
@@ -140,6 +142,11 @@ public class JavaScript : LanguageBase
...
+                var profileOutputPath = Path.GetFullPath(Path.Combine(Util.GetProfileDirectory(WorkingDirectory),stripPackageName, $"{packageVersions[primaryPackage]}_{testName}_{iteration}.cpuprofile"));
```
mikeharder, line 147:
```suggestion
                var profileOutputPath = Path.GetFullPath(Path.Combine(Util.GetProfileDirectory(WorkingDirectory), $"{packageVersions[primaryPackage]}_{testName}_{iteration}.cpuprofile"));
```

I would put all the profiles in a flat directory, rather than putting in a subdirectory based on the primary package name, because you might be interested in all the loaded packages, not just the primary one.  If you do this, you don't need to compute `stripPackageName` either.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/JavaScript.cs
```diff
@@ -140,6 +142,11 @@ public class JavaScript : LanguageBase
...
+                arguments = arguments + $" --profile true --profile-filepath {profileOutputPath}";
```
mikeharder, line 148:
```suggestion
                arguments +=  $" --profile true --profile-filepath {profileOutputPath}";
```

mikeharder (changes requested):

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/ILanguage.cs
HarshaNalluru: I was thinking of this long name, but it'll be too long. Will do.
Let's revert at a later point if we feel like it is too long.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/JavaScript.cs
```diff
@@ -140,6 +142,11 @@ public class JavaScript : LanguageBase
 
             var projectDirectory = Path.Combine(WorkingDirectory, project);
 
+            if(profile){
+                var stripPackageName = primaryPackage.Split(new char[] { '/' })[1]; // "@azure/storage-blob" -> "storage-blob"
+                var profileOutputPath = Path.GetFullPath(Path.Combine(Util.GetProfileDirectory(WorkingDirectory),stripPackageName, $"{packageVersions[primaryPackage]}_{testName}_{iteration}.cpuprofile"));
+                arguments +=  $" --profile true --profile-filepath {profileOutputPath}";
```
HarshaNalluru, line 148:
```suggestion
                arguments +=  $" --profile --profile-filepath {profileOutputPath}";
```

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/JavaScript.cs
HarshaNalluru: Got this from stackoverflow 😋

Updated

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/JavaScript.cs
HarshaNalluru: I don't think I understand fully.
Profiles generated for the storage-blob tests will be stored inside the "storage-blob" subfolder.

> because you might be interested in all the loaded packages, not just the primary one

the same profile would contain the core-http and whichever dependency stack storage-blob(the primary package) depends on

mikeharder (approved):

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/JavaScript.cs
mikeharder: I don't think creating a storage-blob subfolder is necessary, the test name is enough.  However I'm fine keeping the subfolder if you want.

mikeharder:
Verification build: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2201082&view=results

PR #5328, opened 2023-02-02
stress test docker baseimage updates

ckairen:
updating the base images for stress test examples and stress watcher, already triggered the test-resource-deployer docker build [here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2159549&view=results)

PR #5379, opened 2023-02-08
Update `.NET` from `6.0.403` to `7.0.102` + `DOTNET_ROLL_FORWARD`: `Major` + fix wrong quotes in `install-net.yml`

konrad-jamrozik:
This .NET version is available on the Ubuntu 22.04 pools we use per:

- https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#net-tools
- https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/1es-hosted-azure-devops-pools/createpoolimage

This PR also sets `DOTNET_ROLL_FORWARD` to `Major` so that tools targetting .NET `6.0` can run on .NET `7.0`.

For context, please see [this teams chat](https://teams.microsoft.com/l/message/19:59dbfadafb5e41c4890e2cd3d74cc7ba@thread.skype/1675890941633?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1675885183657&teamName=Azure%20SDK&channelName=Engineering%20System%20%F0%9F%9B%A0%EF%B8%8F&createdTime=1675890941633) and this comment:
- https://github.com/Azure/azure-sdk-tools/pull/5328#pullrequestreview-1290001058

Related:
- https://github.com/Azure/azure-sdk-tools/pull/5380

### DIFF
--- a/eng/pipelines/templates/steps/install-dotnet.yml
+++ b/eng/pipelines/templates/steps/install-dotnet.yml
@@ -5,12 +5,22 @@ parameters:
     default: ''
 
 steps:
+  # We set DOTNET_ROLL_FORWARD so that .NET assemblies targeting older versions of .NET can run on newer ones.
+  # See also:
+  # https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet#options-for-running-an-application
+  # https://learn.microsoft.com/en-us/dotnet/core/runtime-discovery/troubleshoot-app-launch?pivots=os-windows#required-framework-not-found
+  # https://aka.ms/dotnet/app-launch-failed
+  # https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=powershell
+  # https://learn.microsoft.com/en-us/azure/devops/pipelines/process/set-variables-scripts?view=azure-devops&tabs=powershell
+  - pwsh: |
+      echo "##vso[task.setvariable variable=DOTNET_ROLL_FORWARD;]Major"
+    displayName: "Set DOTNET_ROLL_FORWARD to Major"
   # https://learn.microsoft.com/azure/devops/pipelines/tasks/reference/use-dotnet-v2?view=azure-pipelines
   - task: UseDotNet@2
     displayName: "Use .NET SDK ${{ coalesce( parameters.DotNetCoreVersion, 'from global.json') }}"
     retryCountOnTaskFailure: 3
     inputs:
-      ${{ if eq( 'parameters.DotNetCoreVersion', '') }}:
+      ${{ if eq( parameters.DotNetCoreVersion, '') }}:
         useGlobalJson: true
       ${{ else }}:
         version: ${{ parameters.DotNetCoreVersion }}
--- a/global.json
+++ b/global.json
@@ -3,7 +3,7 @@
     "Microsoft.Build.Traversal": "3.2.0"
   },
   "sdk": {
-    "version": "6.0.403",
+    "version": "7.0.102",
     "rollForward": "feature"
   }
 }


global.json
```diff
@@ -3,7 +3,7 @@
     "Microsoft.Build.Traversal": "3.2.0"
   },
   "sdk": {
-    "version": "6.0.403",
+    "version": "7.0.102",
```
benbp, line 6: Should we just use 7.0.x here?

weshaggard: No we need to avoid 101 as there was a very bad bug that was fixed in 102

benbp (approved):

benbp:
@konrad-jamrozik do you want to roll in the fix to the quotation evaluation here as well?

konrad-jamrozik:
> @konrad-jamrozik do you want to roll in the fix to the quotation evaluation here as well? https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/templates/steps/install-dotnet.yml#L13

yes :D

konrad-jamrozik:
OK so even after getting rid of this second global.json, apparently .net 7 cannot handle running code targeting .net 6.0.

 

Build:

https://dev.azure.com/azure-sdk/public/_build/results?buildId=2173632&view=logs&j=2d72bf5f-9a0c-5aa3-6e85-aec77a4486f7&t=71826648-8fa0-5027-ec86-1fe5798108f0

 

Code (just 1 commit on top of main):

https://github.com/Azure/azure-sdk-tools/commit/06f2689af99c4b5b4b68dd4b279f4b3b323eb278#diff-8df3cd354bc584349d04ad5675b33c042d8b99b741b8b95af394c55e0f5001bf

konrad-jamrozik:
OK, I setting `DOTNET_ROLL_FORWARD` to `Major` fixed the issue.

heaths:
> OK, I setting `DOTNET_ROLL_FORWARD` to `Major` fixed the issue.

You and @mikeharder are conflating build time with runtime. You always need to build with net7.0 because we may use newer features net6.0 can't build, in part because we now, in the azure-sdk-for-net repo, use LangVersion=11.0. But you need to:

1. Install any SDKs you want to target and run on (technically, you only need the runtimes to run on, but you have to have the SDK to target when building).
2. Either individually build a project with TargetFramework=<version> or TargetFramework*s*=<semicolon-delimited versions>.

That's how tests can target and run on different framework versions. You shouldn't have to mess with _global.json_.

eng/pipelines/templates/steps/install-dotnet.yml
```diff
@@ -5,12 +5,22 @@ parameters:
     default: ''
 
 steps:
+  # We set DOTNET_ROLL_FORWARD so that .NET assemblies targeting older versions of .NET can run on newer ones.
```
weshaggard, line 8: What prompted the need for this? We should try setting multilevellookup before doing this forced roll forward https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/use-dotnet-v2?view=azure-pipelines

konrad-jamrozik: @weshaggard this:
- https://github.com/Azure/azure-sdk-tools/pull/5328#pullrequestreview-1290001058
- 
Basically @ckairen wants to target .NET 7. At first the idea was to use [`DotNetCoreVersion: 7.0.x`](https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/templates/steps/install-dotnet.yml#L26) but it didn't work, as can be seen here:
- https://github.com/Azure/azure-sdk-tools/pull/5379#issuecomment-1423312514

So I proposed an alternative approach where we install .NET 7 instead of .NET 6 and make everything use it, even if it targets .NET 6.

But I understand you would prefer to keep `global.json` to `.NET 6.0.x`, but for @ckairen's tool use `DotNetCoreVersion: 7.0.x` and just install both .NETs ? I understand the `performMultiLevelLookup` would help in finding the right framework.

konrad-jamrozik:
> > OK, I setting `DOTNET_ROLL_FORWARD` to `Major` fixed the issue.
> 
> You and @mikeharder are conflating build time with runtime. You always need to build with net7.0 because we may use newer features net6.0 can't build, in part because we now, in the azure-sdk-for-net repo, use LangVersion=11.0. But you need to:
> 
> 1. Install any SDKs you want to target and run on (technically, you only need the runtimes to run on, but you have to have the SDK to target when building).
> 2. Either individually build a project with TargetFramework={version} or TargetFramework**s**={semicolon-delimited versions}.
> 
> That's how tests can target and run on different framework versions. You shouldn't have to mess with _global.json_.

@heaths I am not sure what I am misunderstanding. The situation is as follows:

- Vast majority of our tools was built with `.NET 6.0.x` and is targeting `.NET 6.0.x`. Hence `global.json` has `.NET 6.0.x`.
- But @ckairen wants to build one tool with `.NET 7.0.x`, and then run it. But you cannot build with `.NET 7.0.x` if `global.json` has only `.NET 6.0.x`. Hence I did bump `global.json` to `.NET 7.0.x`. So @ckairen can now both build & run his `.NET 7.0.x` tool. But now our other tools, that target `.NET 6.0.x`, won't run, because we get an error "You are targeting .NET 6 but you have only .NET 7". Hence I added the `DOTNET_ROLL_FORWARD`.

What is incorrect in what I wrote above?

weshaggard:
@konrad-jamrozik you should be able to update global.json to move to .NET 7 but that will just update the SDK/tools, but we should still be able to target and run on .NET 6 if needed. When you install .NET 7 if you specify performMultiLevelLookup I believe it should allow the tools to fallback and look for older versions of the runtime as needed without forcing those to run on the latest runtime. That is what we do the .NET repo and I believe it should work here as well.

konrad-jamrozik:
> @konrad-jamrozik you should be able to update global.json to move to .NET 7 but that will just update the SDK/tools, but we should still be able to target and run on .NET 6 if needed. When you install .NET 7 if you specify performMultiLevelLookup I believe it should allow the tools to fallback and look for older versions of the runtime as needed without forcing those to run on the latest runtime. That is what we do the .NET repo and I believe it should work here as well. 

I understand this would mean that we would have to also install .NET 6. Right now we have only .NET 7 and everything appears to work - including building and running tools that target 6. Won't it be simpler and faster to have only .NET 7 and use it for everything?

benbp:
@konrad-jamrozik but the tools we build and run actually get moved to 7 as opposed to 6 under the hood right, if I'm understanding correctly. I think the scenario we would want is:

- Dotnet tooling is always latest
- Target SDKs for each C# project moves in the direction of latest, but may need to be upgraded one by one so that one project doesn't block other projects being upgraded.
- We should install all target SDKs that are specified by our tools (right now 6 and 7) in the pipeline.

weshaggard:
Yes exactly we want to make sure our tools that aren't targeting .NET 7 actually run on their intended frameworks as opposed to being forced to run on .NET 7. For example, projects that currently target .NET 6 are being forced to run on .NET 7 in our pipelines here which isn't ensuring that they run on .NET 6 which we would like to ensure. In general, we should try to run our tools on the target frameworks to ensure they work there otherwise we risk missing things.

weshaggard:
>We should install all target SDKs that are specified by our tools (right now 6 and 7) in the pipeline.

I don't think we need to install all the SDK/tools but we do need to make sure the runtimes are there so we can run the tools against those targets. I think we can build with the newer SDK but we need to still run against the older runtimes to ensure we continue to work there.

konrad-jamrozik:
@benbp @weshaggard 

> @konrad-jamrozik but the tools we build and run actually get moved to 7 as opposed to 6 under the hood right
> Yes exactly

I don't this is the case. Here is proof, where I build using .NET 7 but end up having app targeting .NET 6.

``` pwsh
azure-sdk-tools\tools [main ≡]> dotnet --version
7.0.102
azure-sdk-tools\tools\code-owners-parser [main ≡]> $env:DOTNET_ROLL_FORWARD="Major"
azure-sdk-tools\tools\code-owners-parser [main ≡]> dotnet build
MSBuild version 17.4.1+9a89d02ff for .NET
  Determining projects to restore...
  All projects are up-to-date for restore.
  Azure.Sdk.Tools.CodeOwnersParser -> azure-sdk-tools\artifacts\bin\Azure.Sdk.Tools.CodeOwnersPars
  er\Debug\net6.0\Azure.Sdk.Tools.CodeOwnersParser.dll
  Azure.Sdk.Tools.RetrieveCodeOwners -> azure-sdk-tools\artifacts\bin\Azure.Sdk.Tools.RetrieveCode
  Owners\Debug\net6.0\Azure.Sdk.Tools.RetrieveCodeOwners.dll
  Azure.Sdk.Tools.CodeOwnersParser.Tests -> azure-sdk-tools\artifacts\bin\Azure.Sdk.Tools.CodeOwne
  rsParser.Tests\Debug\net6.0\Azure.Sdk.Tools.CodeOwnersParser.Tests.dll
  Azure.Sdk.Tools.RetrieveCodeOwners.Tests -> azure-sdk-tools\artifacts\bin\Azure.Sdk.Tools.Retrie
  veCodeOwners.Tests\Debug\net6.0\Azure.Sdk.Tools.RetrieveCodeOwners.Tests.dll
  ```

And looking at the DLL we can see `[assembly: TargetFramework(".NETCoreApp,Version=v6.0", FrameworkDisplayName = ".NET 6.0")]`:

image

@weshaggard 

> I don't think we need to install all the SDK/tools but we do need to make sure the runtimes are there so we can run the tools against those targets. I think we can build with the newer SDK but we need to still run against the older runtimes to ensure we continue to work there.

OK I'll make a PR tomorrow to fix it.

konrad-jamrozik:
@weshaggard 

> I don't think we need to install all the SDK/tools but we do need to make sure the runtimes are there so we can run the tools against those targets. I think we can build with the newer SDK but we need to still run against the older runtimes to ensure we continue to work there.

OK, I can fix it, but I have a question: does this include `dotnet test`, `dotnet run` or just executions of the executables pulled from the NuGet feed? In other words, should the `ci.yml` definitions that publish to NuGet exercise the target runtime? 

I think I will revert the `DOTNET_ROLL_FORWARD` change and instead use `performMultiLevelLookup` with both `.NET 6` an `.NET 7` SDKs installed. I am not sure how much it would complicate our pipeline invocations of `dotnet` to have only one SDK but two runtimes, and figure out when to use the older runtime.

heaths:
So we're not having to change our tools every year, targeting them at net6.0 makes more sense (then we only _have_ to change them every 3 years). But we should always build with the latest - currently net7.0, which has 1 year of support. We may use new language features for productivity. But as long as we target net6.0, I see no problem with rolling forward for *tools*. It makes them easier to use.

In general, though, on our build machines we should install any runtimes at least that we need, like we do in the azure-sdk-for-net, so that we actually test *running* on those runtimes instead of automatically rolling forward. Again, for tools, I think this is less critical, but is critical that we run on what we say we should run on e.g., net461 (baseline; technically it still runs on .NET Framework 4.8 but with a compat layer for net461), netcoreapp3.1 (for some track 1s yet), net6.0, and net7.0.

For perf, I think we should always target and run on latest across all languages as perf is generally a focus of any new version.

konrad-jamrozik:
@heaths that makes sense, thx. I have a question about this:

> In general, though, on our build machines we should install any runtimes at least that we need, like we do in the azure-sdk-for-net, so that we actually test _running_ on those runtimes instead of automatically rolling forward. 

Does this mean that if I built a project targeting .NET 6 with .NET 7 SDK and the output assembly states it is targeting .NET 6, then I must confirm it works when run on .NET 6 runtime, and just running it on .NET 7 runtime is not enough? I just don't understand why I have to do that. Like, is it possible that an assembly is targeting .NET 6 but it doesn't work on .NET 6 runtime, only .NET 7? How is it possible? Wouldn't compiler prevent me in the first place from emitting such assembly?

heaths:
For tools, though, I think roll-forward behavior still makes sense. We don't generally use reflection (shouldn't be), and the likelihood of breakage is low.

More so what I meant about my statement you cited above is that I meant, especially for testing, for bits we ship to customers e.g., SDKs, we want to make sure that it works on supported runtimes. Since our SDKs target netstandard2.0, technically we support everything from net461 up and netcoreapp2.0 and up. But we only test net461 as a baseline (.NET Framework has strict no-breaking change policies - most often can't even fix runtime behaviors) and net6.0 and net7.0.

Frankly, I think you should leave the roll-forward for *tools*, but still need to install the net6.0 SDK so that you can target it. Keep global.json with net7.0, though, so that we can use new languages and so you can, when needed, run/build for net7.0 (which seems like the original case, right?).

weshaggard:
@heaths while we do produce a lot of tools here we also produce libraries that get used in other context and I want to keep us in a consistent state of trying to test on the frameworks we claim we want to support. I'm not a fan of force roll-forward as a repo wide setting and I think testing on the other frameworks is easy enough and we should do that by default.

PR #5384, opened 2023-02-09
`Info/Active/` now offers data for individual sessions

### LINKED ISSUE #5362: Update `Info/Active` to reveal `recordingId-specific` details
Right now, they only show session-level sanitizers.

Need to enhance to honor

`Info/Active/<recordingId/` to also optionally retrieve the configuration for those.

scbedd:
Resolves #5362 

@samvaity is working on transition appconfiguration to test-proxy for java. As she was doing that, she wanted to see the `order of application` of the sanitizers _after_ they are registered. This is because the java framework does a lot of behind-the-scenes work to pregister more than a few sanitizers.

It is a much easier experience to debug if folks can _actually see_ the session level customizations when they view the `Info/Active` page.

Of course, the proviso is that you can't get active information for a recording or playback session that has ended, but the test-proxy should enable this regardless.

How to use it? Before your recording stops, and while you have that recordingId on hand, hit a browser with URL:

`<proxyurl>/Info/Active?id=<recordingId>`

TODO:
- [x] Honor argument with optional specific recordingId
- [x] Confirm via visual inspection that content is rendering properly
- [x] We are patching out the odd `BackingField` that is showing up on `RecordedTestSanitizer`, by excluding any of the fields that show up on the base classes `RecordedTestSanitizer`, `ResponseTransform`, and `RecordMatcher`. These don't offer any customizations, so we can safely exclude their fields.
- [x] Add error page for unrecognized recording Id

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Models/ActiveMetadataModel.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Models/ActiveMetadataModel.cs
@@ -8,11 +8,17 @@ using System.IO;
 using Azure.Sdk.Tools.TestProxy.Common;
 using System.Collections.Concurrent;
 using Microsoft.CodeAnalysis.Operations;
+using Azure.Sdk.Tools.TestProxy.Common.Exceptions;
 
 namespace Azure.Sdk.Tools.TestProxy.Models
 {
     public class ActiveMetadataModel : RunTimeMetaDataModel
     {
+        public ActiveMetadataModel(string recordingId)
+        {
+            RecordingId = recordingId;
+        }
+
         public ActiveMetadataModel(RecordingHandler pageRecordingHandler)
         {
             Descriptions = _populateFromHandler(pageRecordingHandler, "");
@@ -39,6 +45,7 @@ namespace Azure.Sdk.Tools.TestProxy.Models
                 handler.InMemorySessions
             };
 
+            var recordingFound = false;
             if (!string.IsNullOrWhiteSpace(recordingId)){
                 foreach (var sessionDict in searchCollections)
                 { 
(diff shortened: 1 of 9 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/launchSettings.json
```diff
@@ -1,20 +1,21 @@
...
-  },
   "profiles": {
     "Azure.Sdk.Tools.TestProxy": {
       "commandName": "Project",
+      "workingDirectory": "C:/repo/sdk-for-java",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
         "Logging__LogLevel__Microsoft": "Information"
       },
       "applicationUrl": "https://localhost:5001;http://localhost:5000"
```
scbedd, line 10: Will undo.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Views/Info/ActiveExtensions.cshtml
```diff
@@ -17,9 +17,20 @@
...
+    @if (!string.IsNullOrWhiteSpace(Model.RecordingId))
+    {
+        <p>
+            The below extensions are configured for recording <b>@Model.RecordingId</b>.
+        </p>
+    }
+    else
+    {
+        <p>
+            The below extensions are currently configured.
```
scbedd, line 30:
```suggestion
            The below extensions are currently configured for all sessions.
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Models/ActiveMetadataModel.cs
```diff
@@ -1,26 +1,79 @@
...
+                        }
+
+                        recordingFound = true;
+                        break;
+                    }
+                }
+
+                if (!recordingFound)
+                {
+                    throw new HttpException(System.Net.HttpStatusCode.BadRequest, $"{recordingId} is not found in any Playback, Recording, or In-Memory sessions.");
```
hallipr, line 69: It seems odd to throw api layer exceptions from the model when you're looking for this specific throw in the controller.  i,e. The BadRequest status code and message are unused.

hallipr (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Models/ActiveMetadataModel.cs
scbedd: Yeah you're right. Converted to `SessionNotActiveException`

PR #5395, opened 2023-02-10
Install  `.NET 6.0.x` runtime. Remove `DotNetCoreVersion` param. Undo `DOTNET_ROLL_FORWARD`.

konrad-jamrozik:
Changes made per this comment:
- https://github.com/Azure/azure-sdk-tools/pull/5379#issuecomment-1424766311

and my discussion with @weshaggard 

We were already using the newest .NET SDK to build things; it comes from [`global.json`](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json). But with this PR, we also install `.NET 6.0.x` runtime, so that our tooling targeting it runs against it.

### A note on parameter support

I have removed `DotNetCoreVersion` because:

- Setting `global.json` `.NET` to `6.0.x` and trying to override with `DotNetCoreVersion 7.0.x` does not work (<- this may not be true, but we want to have newest .NET in `global.json` anyway).
- Setting `global.json` `.NET` to `7.0.x` and overriding with `DotNetCoreVersion 6.0.x` would work, but then the parameter would have to be passed in many places, as most of our tools target `6.0.x`.
- We could change the `DotNetCoreVersion` to something like `SkipRuntimeInstall`; in such case, the `.NET 6.0.x` would be installed by default, but then for tools migrated to `7.0.x` one could skip it by passing `SkipRuntimeInstall`. Once all tools are migrated to `7.0.x` we could update this file to no longer install `6.0.x` runtime and get rid of all such `SkipRuntimeInstall` arguments. However, this seems to me as some amount of work not necessarily worth the effort.

Having said that, once we need more flexibility for the runtimes installed, we could add the param back.

### Testing done

This build confirms that an assembly targeting `7.0.x` works (I made `Azure.Sdk.Tools.RetrieveCodeOwners` and `Azure.Sdk.Tools.CodeOwnsersParser` target `7.0.0` for this build):

- https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2177144&view=results

This build confirms that an assembly targeting `6.0.x` works:
- https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2177181&view=logs&s=c0a22914-42ba-51eb-e68d-84c64519cae1

### DIFF
--- a/eng/pipelines/templates/steps/install-dotnet.yml
+++ b/eng/pipelines/templates/steps/install-dotnet.yml
@@ -1,29 +1,25 @@
-parameters:
-  # Use this parameter if you want to override the .NET SDK set by global.json
-  - name: DotNetCoreVersion
-    type: string
-    default: ''
-
 steps:
-  # We set DOTNET_ROLL_FORWARD so that .NET assemblies targeting older versions of .NET can run on newer ones.
-  # See also:
-  # https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet#options-for-running-an-application
-  # https://learn.microsoft.com/en-us/dotnet/core/runtime-discovery/troubleshoot-app-launch?pivots=os-windows#required-framework-not-found
-  # https://aka.ms/dotnet/app-launch-failed
-  # https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=powershell
-  # https://learn.microsoft.com/en-us/azure/devops/pipelines/process/set-variables-scripts?view=azure-devops&tabs=powershell
-  - pwsh: |
-      echo "##vso[task.setvariable variable=DOTNET_ROLL_FORWARD;]Major"
-    displayName: "Set DOTNET_ROLL_FORWARD to Major"
-  # https://learn.microsoft.com/azure/devops/pipelines/tasks/reference/use-dotnet-v2?view=azure-pipelines
-  - task: UseDotNet@2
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/steps/install-dotnet.yml
```diff
@@ -1,29 +1,25 @@
...
     inputs:
-      ${{ if eq( parameters.DotNetCoreVersion, '') }}:
-        useGlobalJson: true
-      ${{ else }}:
-        version: ${{ parameters.DotNetCoreVersion }}
+      packageType: runtime
+      version: 6.0.x
+      # performMultiLevelLookup comes into play when given .NET executable target runtime is different
+      # than the installed .NET SDK. Without this, such runtime would not be found.
+      performMultiLevelLookup: true
```
weshaggard, line 22: I was expecting to see this on the SDK install above. Did you try to verify if we need to explicitly have this step for the 6 runtime?

konrad-jamrozik: Per our chat, these builds shows the behavior on `performMultiLevelLookup` for SDK:

([source](https://github.com/Azure/azure-sdk-tools/commit/6b2a4efe490eee3027b5ef9ce5a1c179b047b66d))

For 6.0 target (fails):
- https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2177201&view=logs&j=951e43bb-1e6d-59fb-d232-79b6dbbdcd74&t=3c49b6ca-e856-54cd-c01f-bfd23f6f6d30&l=22

For 7.0 target (works):
- https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2177200&view=results

weshaggard (approved):

PR #5399, opened 2023-02-10
SwaggerAPIView upgrade to dotnet7

ruowan:


ruowan:
@konrad-jamrozik Help to review this PR? I think I need to upgrade dotnet7

konrad-jamrozik (approved):
@ruowan this should just work. I am just curious: why you need to update to .NET 7?

ruowan:
The dotnet version in global.json is 7. So to keep it consistent, we need to update the dotnet version in swagger-api-parser tools.

konrad-jamrozik:
> The dotnet version in global.json is 7. So to keep it consistent, we need to update the dotnet version in swagger-api-parser tools.

@ruowan Just to be clear: it will also work with .NET 6. We had a long discussion about it, and a series of PRs, here:
- https://github.com/Azure/azure-sdk-tools/pull/5379

PR #5403, opened 2023-02-10
Add `test-proxy` contributing.md

### LINKED ISSUE #5392: Create `test-proxy` CONTRIBUTING.md

scbedd:
@mario-guerra when you have a couple minutes please take a look.

Resolves #5392

### DIFF
--- a/tools/test-proxy/CONTRIBUTING.md
+++ b/tools/test-proxy/CONTRIBUTING.md
@@ -115,7 +115,7 @@ dotnet --list-sdks
 
 Find a version starting with a `6`, if you don't have one on your system, [install it!](https://dotnet.microsoft.com/en-us/download/dotnet/6.0)
 
-Once you have that, find the `global.json` at root of this repo.
+Once you have that, find the [`global.json`](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json) at root of this repo.
 
 ```json
 {
--- a/tools/test-proxy/CONTRIBUTING.md
+++ b/tools/test-proxy/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Within this folder are all the components that make up the the tool `Azure.Sdk.Tools.TestProxy`, henceforth referred to as the `test-proxy`.
 
-This product is currently developed on Visual Studio 2022, but CLI build/test also work fine. Currently, this project is built against `.net 6.0.403`.
+This product is currently developed on Visual Studio 2022, but CLI build/test also work fine. Currently, this project only supports `.net 6`, though `.net 7` support is coming.
 
 To contribute a change:
 
--- a/tools/test-proxy/CONTRIBUTING.md
+++ b/tools/test-proxy/CONTRIBUTING.md
@@ -1,0 +1,129 @@
+# Azure SDK Test-Proxy Contribution guide
+
+Within this folder are all the components that make up the the tool `Azure.Sdk.Tools.TestProxy`, henceforth referred to as the `test-proxy`.
+
+This product is currently developed on Visual Studio 2022, but CLI build/test also work fine. Currently, this project is built against `.net 6.0.403`.
+
+To contribute a change:
+
+- Fork.
+- Create a branch on your fork with your change.
+- Locally test your change, add a unit test, commit.
+- Submit a PR to this repo from your forked branch!
+  - Relevant azure-sdk devs will review your PR.
+
+## Directory Guide
+
+| Folder | Purpose |
+|---|---|
+| `Azure.Sdk.Tools.TestProxy/` | The actual recording server. Standard `csproj`, minimal external dependencies, packed as .NET tool.  |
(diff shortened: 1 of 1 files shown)


mario-guerra:
@scbedd I took a quick pass through the doc and it looks good. I'd like to delve into it a bit deeper and follow along with the instructions to see if I can successfully make a change and build the project before we commit this doc. I will do that early next week.

kurtzeborn (approved):
This looks great!

tools/test-proxy/CONTRIBUTING.md
```diff
@@ -0,0 +1,132 @@
+# Azure SDK Test-Proxy Contribution guide
+
+Within this folder are all the components that make up the the tool `Azure.Sdk.Tools.TestProxy`, henceforth referred to as the `test-proxy`.
+
+This product is currently developed on Visual Studio 2022, but CLI build/test also work fine. Currently, this project is built against `.net 6.0.403`.
```
benbp, line 5: Will the version get out of date quickly, given we pin to the major version only in the csproj?

tools/test-proxy/CONTRIBUTING.md
```diff
@@ -0,0 +1,132 @@
...
+    "Microsoft.Build.Traversal": "3.2.0"
+  },
+  "sdk": {
+    "version": "7.0.102",
+    "rollForward": "feature"
+  }
+}
+```
+
+Change the `version` to one that shows up in the list when you call `dotnet --list-sdks`.
```
benbp, line 132: We're trying to configure this such that you only need to make sure to have the targeted sdk/runtime installed for the csproj, and global.json will always point to the newest version of the tooling. As such, this last step may not be needed FYI @konrad-jamrozik

konrad-jamrozik: To elaborate on what @benbp said, we settled on installing both v. 6 and 7 via `install-dotnet.yml` with the idea that everything is built with v. 7 from global.json (7 as in "currently the latest") but runs on the runtime it targets. (#5405).

So there are few ways to solve it: 
- coerce everything to .net 6 as @scbedd is proposing;
- keep global.json unchanged, don't install .net 6, and ".net roll forward" as in https://github.com/Azure/azure-sdk-tools/pull/5379;
- keep global.json unchanged and install .net 6 sdk or runtime. But not sure if VS will be happy with that. It might still prioritize .net 7 coming from global.json and complain one is trying to build project targeting 6 with .net 7 sdk.

tools/test-proxy/CONTRIBUTING.md
```diff
@@ -0,0 +1,132 @@
...
+6.0.113 [C:\Program Files\dotnet\sdk]
+6.0.308 [C:\Program Files\dotnet\sdk]
+6.0.401 [C:\Program Files\dotnet\sdk]
+6.0.405 [C:\Program Files\dotnet\sdk]
+7.0.101 [C:\Program Files\dotnet\sdk]
+```
+
+Find a version starting with a `6`, if you don't have one on your system, [install it!](https://dotnet.microsoft.com/en-us/download/dotnet/6.0)
+
+Once you have that, find the `global.json` at root of this repo.
```
konrad-jamrozik, line 118:
```suggestion
Once you have that, find the [`global.json`](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json) at root of this repo.
```

tools/test-proxy/CONTRIBUTING.md
scbedd: It's pinned in our `globals.json`. The framework is only partially pinned in the csproj. I highlight how to fix that further on in the doc.

tools/test-proxy/CONTRIBUTING.md
scbedd: Per my understanding `global.json` essentially wins.

tools/test-proxy/CONTRIBUTING.md
scbedd: I'm thinking we just support .NET 7 too. Skip the issues :P

tools/test-proxy/CONTRIBUTING.md
benbp: @scbedd global.json wins for the dotnet cli/tooling version required, but if you just add net 6 in the csproj, everything will still work provided you have the net 6 sdk+runtime installed.

If I'm understanding correctly, our plan is to always have the latest version of tooling, but pin our tools to the LTS, which is 6.

@konrad-jamrozik all this said, I've also had trouble making sure my local installation works seamlessly, so we should have some docs here around how to set this up because it can get pretty confusing.

scbedd: Yeah. I'll update the doc to >= .NET 6, but take action in another PR to add .NET 7 support too.

tools/test-proxy/CONTRIBUTING.md
scbedd, line 5:
```suggestion
This product is currently developed on Visual Studio 2022, but CLI build/test also work fine. Currently, this project only supports `.net 6`, though `.net 7` support is coming.
```

PR #5404, opened 2023-02-10
Update Profile Directory

alzimmermsft:
Updates profiling in performance pipeline to use new profiling directory. Performance test runs will now only run one language so there is no need to prefix the profiling directory with the language name and it can now just be called `profile`.

### DIFF
--- a/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Program.cs
+++ b/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Program.cs
@@ -13,7 +13,6 @@ using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using YamlDotNet.Core;
 using YamlDotNet.Serialization;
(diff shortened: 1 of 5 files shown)


tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Program.cs
```diff
@@ -13,6 +13,7 @@
 using System.Threading.Tasks;
 using YamlDotNet.Core;
 using YamlDotNet.Serialization;
+using static Azure.Sdk.Tools.PerfAutomation.Program;
```
mikeharder, line 16: Is this needed?

alzimmermsft: Doesn't appear so, I need to properly set up my .NET dev environment to clean this up automatically.

mikeharder (approved):

PR #5406, opened 2023-02-10
[Perf] Set "Profile" parameter to true by default

mikeharder:
- Increases test coverage of profiling code


alzimmermsft (approved):

eng/common/pipelines/templates/jobs/perf.yml
```diff
@@ -171,6 +171,7 @@ jobs:
       targetPath: $(System.DefaultWorkingDirectory)/${{ parameters.Language }}-profile.zip
       artifactName: ${{ parameters.Language }}-profile.zip
     condition: ${{ parameters.Profile }}
+    continueOnError: true
```
weshaggard, line 174: Is this failing on retry? if so you might want to consider using our [publish-artifact](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/steps/publish-artifact.yml) template

mikeharder: No, it's failing because sometimes the artifact doesn't exist.  What we want is "PublishPipelineArtifactIfFileExists".

mikeharder: I'm not sure if `continueOnError` will do what we want, but I'm testing it.  It's OK if this step fails, we just don't want it to turn the pipeline red.

weshaggard: Ok yeah we don't have a great solution for only do it if the file exists. In some other scenarios we have uploaded a directory and we just aways ensure the directory exists.

mikeharder: OK, I should be able to do something similar for perf.

Issue #4670, opened 2022-11-09
Enable email sender service


maririos:
Instead of continuing the use of SendGrid, this [doc](https://dev.azure.com/devdiv/Engineering/_wiki/wikis/CNEKB/24555/How-to-send-email-from-Azure) about how to use MS service to send email.

richardpark-msft:
Just FYI, I just realized I've been filtering all of these emails because it met two of my criteria - [EXTERNAL] and it has an unsubscribe link, which is typical for spam.

PR #5408, opened 2023-02-10
APIView changes to send emails using logic app service

praveenkuttappan:
APIView was using SendGrid to send email notifications and this has a limitation in number of emails we can send. This PR has a change to send emails using Azure logic app service added in Azure subscription instead of sendgrid.


src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -4,52 +4,61 @@
 using APIView.Identity;
 using APIViewWeb.Models;
 using Microsoft.Extensions.Configuration;
-using SendGrid;
```
weshaggard, line 7: Can we also remove the package reference to sendgrid now as well?

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -4,52 +4,61 @@
...
+            ICosmosReviewRepository reviewRepository,
+            ICosmosUserProfileRepository userProfileRepository)
         {
-            _sendGridClient = sendGridClient ?? new SendGridClient(configuration[SENDGRID_KEY_SETTING]);
-            _endpoint = configuration.GetValue<string>(ENDPOINT_SETTING);
+            _apiviewEndpoint = configuration.GetValue<string>(ENDPOINT_SETTING);
             _reviewRepository = reviewRepository;
             _userProfileRepository = userProfileRepository;
+            _configuration = configuration;
+            _testEmailToAddress = configuration["apiview-email-test-address"] ?? "";
```
weshaggard, line 43: What does test address mean in this context?

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -60,20 +69,17 @@ public async Task NotifyApproversOfReview(ClaimsPrincipal user, string reviewId,
             {
                 var reviewerProfile = await _userProfileRepository.TryGetUserProfileAsync(reviewer);
                 await SendUserEmailsAsync(review, reviewerProfile,
-                    GetApproverReviewContentHeading(userProfile, false),
```
weshaggard, line 63: Are we also removing some notification emails as part of this change?

weshaggard: Or do we no longer need the plain text with the new approach?

src/dotnet/APIView/APIViewWeb/Models/EmailModel.cs
```diff
@@ -0,0 +1,26 @@
+using System.Text.Json.Serialization;
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
weshaggard, line 2: Move header to the top

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -158,94 +164,84 @@ private string GetHtmlContent(CommentModel comment, ReviewModel review)
...
-                var backupEmail = GetUserEmail(userBackup);
-                if (!backupEmail.IsNullOrEmpty())
-                {
-                    e = new EmailAddress(backupEmail, user.UserName);
-                }
-                else
-                {
-                    return;
-                }
+                _telemetryClient.TrackTrace($"Email address is not available for user {user.UserName}. Not sending email.");
```
weshaggard, line 175: Do we want to someone watch this? Do we have any upfront validation to ensure that if someone requests a review or mentioned someone that tells folks that they don't have an email configured? Or do we expect this to be an exceptional case because we are mapping to AAD accounts now?

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
praveenkuttappan: This is to enable testing on local and on even deployed instances if required to a specific email address instead of using actual email address on test instance.

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
praveenkuttappan: We no longer need that. SendGrid, for some reasons, require plain text.

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
praveenkuttappan: We use the email address in APIView profile. Notification system has not been changed yet to pick email address from AAD.

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
weshaggard: Ok so if that value is set we have to restart the server to unset it?

weshaggard: Opps that question was intended for the test email address configuration

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -158,94 +164,87 @@ private string GetHtmlContent(CommentModel comment, ReviewModel review)
...
+            HashSet<string> notifiedEmails = new HashSet<string>();
+            if (notifiedUsers != null)
             {
-                var backupEmail = GetUserEmail(userBackup);
-                if (!backupEmail.IsNullOrEmpty())
-                {
-                    e = new EmailAddress(backupEmail, user.UserName);
-                }
-                else
+                foreach (var usename in notifiedUsers)
```
weshaggard, line 188:
```suggestion
                foreach (var username in notifiedUsers)
```

weshaggard (approved):
Looks reasonable. Lets make sure we get some good testing on this change.

praveenkuttappan:
Fixes https://github.com/Azure/azure-sdk-tools/issues/4670

PR #5426, opened 2023-02-13
Java ApiView: Add context to annotation id

srnagar:
Currently, annotation ids include line numbers which can change when the source code changes and results in incorrectly showing a diff in API view.
image

This PR changes the annotation id to use the context (i.e the node on which the annotation is defined) to create the id.

Before:

```json
{
    "DefinitionId" : "com.azure.data.appconfiguration.ConfigurationAsyncClient.ServiceClient-L51",
    "Kind" : 6,
    "Value" : "@ServiceClient"
  },
```

After
```json
{
    "DefinitionId" : "com.azure.data.appconfiguration.ConfigurationAsyncClient.ServiceClient-ConfigurationAsyncClient",
    "Kind" : 6,
    "Value" : "@ServiceClient"
  },
```

### DIFF
--- a/src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/util/ASTUtils.java
+++ b/src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/util/ASTUtils.java
@@ -10,6 +10,8 @@ import com.github.javaparser.ast.body.BodyDeclaration;
 import com.github.javaparser.ast.body.CallableDeclaration;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.ConstructorDeclaration;
+import com.github.javaparser.ast.body.EnumConstantDeclaration;
+import com.github.javaparser.ast.body.EnumDeclaration;
 import com.github.javaparser.ast.body.FieldDeclaration;
 import com.github.javaparser.ast.body.MethodDeclaration;
 import com.github.javaparser.ast.body.TypeDeclaration;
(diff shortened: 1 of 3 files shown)


src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/util/ASTUtils.java
```diff
@@ -197,9 +200,38 @@ public static String makeId(String fullPath) {
         return MAKE_ID.matcher(fullPath).replaceAll("-");
     }
 
-    public static String makeId(AnnotationExpr annotation) {
-        int line = annotation.getBegin().orElseThrow(RuntimeException::new).line;
-        return makeId(getNodeFullyQualifiedName(annotation.getParentNode()) + "." + annotation.getNameAsString() + "-L" + line);
+    public static String makeId(AnnotationExpr annotation, NodeWithAnnotations<?> nodeWithAnnotations) {
+        String annotationContext = getAnnotationContext(nodeWithAnnotations);
+
+        String idSuffix = "-" + annotationContext;
```
JonathanGiles, line 206: We probably don't want to append 'null', maybe move this append into an 'else' clause?

JonathanGiles (approved):

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/util/ASTUtils.java
srnagar: If `annotationContext` is `null` or empty, `idSuffix` gets overwritten with `-L<int>` in the `if` block below. So, `-null` should never be used but if it improves readability, I'll move it to `else` clause.

JonathanGiles: Great! Reviewing on my phone I missed that!

PR #5427, opened 2023-02-13
fix(build): cadl build error when repo url does not end with `.git`

archerzz:
`[.git]` is a char set which means the url should ends with either `.`, or `g`, or `i`, or `t`. So if the url doesn't end with `.git` then there is no match which results in error.

The fix is to remove `[.git]` char set. We don't need `(.git)*` either, because `\S+` will suppress it.

### DIFF
--- a/eng/common/scripts/Cadl-Project-Sync.ps1
+++ b/eng/common/scripts/Cadl-Project-Sync.ps1
@@ -44,10 +44,10 @@ function GetGitRemoteValue([string]$repo) {
         $gitRemotes = (git remote -v)
         foreach ($remote in $gitRemotes) {
             if ($remote.StartsWith("origin")) {
-                if ($remote -match 'https://github.com/\S+[\.git]') {
+                if ($remote -match 'https://github.com/\S+') {
                     $result = "https://github.com/$repo.git"
                     break
-                } elseif ($remote -match "git@github.com:\S+[\.git]"){
+                } elseif ($remote -match "git@github.com:\S+"){
                     $result = "git@github.com:$repo.git"
                     break
                 } else {


archerzz:
## Current Behavior
image

## After Fix
image

m-nash (approved):
I believe this will work.
Please verify locally both ssh and uri matching.

@weshaggard do you mind double checking this change as well?

eng/common/scripts/Cadl-Project-Sync.ps1
```diff
@@ -44,10 +44,10 @@ function GetGitRemoteValue([string]$repo) {
         $gitRemotes = (git remote -v)
         foreach ($remote in $gitRemotes) {
             if ($remote.StartsWith("origin")) {
-                if ($remote -match 'https://github.com/\S+[\.git]') {
+                if ($remote -match 'https://github.com/\S+') {
```
weshaggard, line 47: What scenarios do we not have the ".git" at the end? I know it is supported but the majority of workflows folks use should have it. Instead of removing perhaps we make it optional `(\.git)?`. I'm actually not sure the older expression was doing what we wanted anyway as it is just matching a single char in the set `[\.git]`.

m-nash: I think the older expression was intended to be optional, but I used `[ ]` mistakenly instead of `( )` and due to the nature of of the difference it still passed testing

weshaggard: Actually investigating further I think what you have probably works for the goal of this function which is really only to determine if you are using https or git protocol. We probably don't even need to use regex for that and could just do `$remote.StartsWith("...")` however what is here also works.

weshaggard (approved):

haolingdong-msft:
Overall LGTM. nit: We have [single quote](https://github.com/Azure/azure-sdk-tools/blob/e70fd62b7cc72774148310002c019e833645d314/eng/common/scripts/Cadl-Project-Sync.ps1#L47) and [double quote](https://github.com/Azure/azure-sdk-tools/blob/e70fd62b7cc72774148310002c019e833645d314/eng/common/scripts/Cadl-Project-Sync.ps1#L50) format regex, shall we all use single quote?

haolingdong-msft:
I checked in Java sdk repo, the change works now.

haolingdong-msft (approved):

PR #5428, opened 2023-02-13
Comment update: clarify we need `sdk` in `install-dotnet.yml` due to `asp.net core` runtime

konrad-jamrozik:
For context, please see:
- https://github.com/Azure/azure-sdk-tools/pull/5405#discussion_r1105006774

### DIFF
--- a/eng/pipelines/templates/steps/install-dotnet.yml
+++ b/eng/pipelines/templates/steps/install-dotnet.yml
@@ -19,9 +19,6 @@ steps:
       # Specifically, test-proxy requires asp.net core runtime, which is installed only when sdk option
       # is selected, per: https://github.com/microsoft/azure-pipelines-tasks/issues/14405
       #
-      # Build showing the lack of asp.net core runtime when 'runtime' is used instead of 'sdk':
-      #    https://dev.azure.com/azure-sdk/public/_build/results?buildId=2177474&view=logs&j=f285c022-fd09-5cd2-85b4-c0252bc68dbc&t=1e6dcdea-6b31-5539-122f-c6b9e4b7da81
-      #
       # For additional context, see:
       # https://github.com/Azure/azure-sdk-tools/pull/5405#discussion_r1105006774
       # https://github.com/Azure/azure-sdk-tools/pull/5405
--- a/eng/pipelines/templates/steps/install-dotnet.yml
+++ b/eng/pipelines/templates/steps/install-dotnet.yml
@@ -16,7 +16,15 @@ steps:
     retryCountOnTaskFailure: 3
     inputs:
       # We must install sdk, not just runtime, as it is required by some of our tools, like test-proxy.
-      # For additional context, see PR: https://github.com/Azure/azure-sdk-tools/pull/5405
+      # Specifically, test-proxy requires asp.net core runtime, which is installed only when sdk option
+      # is selected, per: https://github.com/microsoft/azure-pipelines-tasks/issues/14405
+      #
+      # Build showing the lack of runtime when 'runtime' is used instead of 'sdk':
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/steps/install-dotnet.yml
```diff
@@ -16,7 +16,15 @@ steps:
     retryCountOnTaskFailure: 3
     inputs:
       # We must install sdk, not just runtime, as it is required by some of our tools, like test-proxy.
-      # For additional context, see PR: https://github.com/Azure/azure-sdk-tools/pull/5405
+      # Specifically, test-proxy requires asp.net core runtime, which is installed only when sdk option
+      # is selected, per: https://github.com/microsoft/azure-pipelines-tasks/issues/14405
+      #
+      # Build showing the lack of runtime when 'runtime' is used instead of 'sdk':
```
konrad-jamrozik, line 22:
```suggestion
      # Build showing the lack of asp.net core runtime when 'runtime' is used instead of 'sdk':
```

eng/pipelines/templates/steps/install-dotnet.yml
```diff
@@ -16,7 +16,15 @@ steps:
     retryCountOnTaskFailure: 3
     inputs:
       # We must install sdk, not just runtime, as it is required by some of our tools, like test-proxy.
-      # For additional context, see PR: https://github.com/Azure/azure-sdk-tools/pull/5405
+      # Specifically, test-proxy requires asp.net core runtime, which is installed only when sdk option
+      # is selected, per: https://github.com/microsoft/azure-pipelines-tasks/issues/14405
+      #
+      # Build showing the lack of asp.net core runtime when 'runtime' is used instead of 'sdk':
```
weshaggard, line 22: This link is probably not helpful as the build will go ahead after some time. I think the note above is more important.

weshaggard (approved):
remove the build link otherwise looks good.

eng/pipelines/templates/steps/install-dotnet.yml
```diff
@@ -16,7 +16,15 @@ steps:
     retryCountOnTaskFailure: 3
     inputs:
       # We must install sdk, not just runtime, as it is required by some of our tools, like test-proxy.
-      # For additional context, see PR: https://github.com/Azure/azure-sdk-tools/pull/5405
+      # Specifically, test-proxy requires asp.net core runtime, which is installed only when sdk option
```
scbedd, line 19: Ahhhhhh so it's not that there was anything with the runtime we were using, it just didn't have the `asp.net core` runtime in the standard. 👍

eng/pipelines/templates/steps/install-dotnet.yml
scbedd: > as the build will go ahead

You mean `expire` here right? Agreed.

scbedd (approved):

eng/pipelines/templates/steps/install-dotnet.yml
weshaggard: yes the build will expire and the link will be broken.

konrad-jamrozik:
@weshaggard should I remove the build link even with this?

image

I always manually retain builds I link to.

eng/pipelines/templates/steps/install-dotnet.yml
konrad-jamrozik: Replied [here](https://github.com/Azure/azure-sdk-tools/pull/5428#issuecomment-1428808128)

weshaggard:
>@weshaggard should I remove the build link even with this?_

I'd say it still isn't worth it. If you want log messages to be maintained I'd suggest copying the part that is interesting into this PR so folks could find it later if needed.

PR #5454, opened 2023-02-15
Add Delayed Response

### LINKED ISSUE #5393: [Prioritize] Support delayed responses during playback
Context: one of our services works by establishing a connection and then streaming multiple "chunks" of content as they are ready.

The test-proxy works just fine with these services during `recording`, but during `playback` _all_ of the chunks come back at the same time. This is due to the fact that the test-proxy is merely matching the request and writing the entire content to the response stream.

[Here is a PR that contains context on .NET side](https://github.com/Azure/azure-sdk-for-net/pull/33990) -- `StreamingCompletionsTest` is the target.

To properly resolve this, we need to either A) offer a transport customization or B) recognize _something_ in the requests that activates a "delay" behavior in the writing of the response stream.

Use the above as an acceptance test. Add new feature

`Admin/SetRecordingOptions` 

with 

```
"Transport": {
  "PlaybackResponseTime": "<integer in ms>"
}
```

And the test-proxy will "smear" the response bytes over that input time. By default we'll set `0`, which means no delay at all.

scbedd:
Resolves #5393

@mikeharder this is my first go at this. Of course, it requires a `batchCount` to properly smear the bytes. I just used `10` to start with.

Gotta actually try this thing now to confirm.👍 

Confirmation from the sdk-for-net folks that this works as expected!


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/TransportCustomizations.cs
```diff
@@ -31,5 +31,11 @@ public class TransportCustomizations
         /// Each certificate pair contained within this list should be added to the clientHandler for the server or an individual recording.
         /// </summary>
         public List<PemPair> Certificates { get; set; }
+
+        /// <summary>
+        /// During playack, a response is normally returned all at once. By offering this response time, we can
```
scbedd, line 36:
```suggestion
        /// During playback, a response is normally returned all at once. By offering this response time, we can
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -464,8 +467,67 @@ public async Task HandlePlaybackRequest(string recordingId, HttpRequest incoming
...
+                if (match.Response.Headers.ContainsKey("Content-Length"))
+                {
+                    outgoingResponse.ContentLength = bodyData.Length;
+                }
+
+                await WriteBodyBytes(bodyData, session.PlaybackResponseTime, outgoingResponse);
+            }
+        }
+
+        public byte[][] GetBatches(byte[] bodyData, int batchCount)
```
mikeharder, line 479: Perf could be improved to avoid copying bytes by returning something like `IEnumerable<ReadOnlyMemory<byte>>` instead of `byte[][]`, but I don't think many requests will use the `playbackResponseTime` feature, and certainly not perf-sensitive users, so the current implementation is fine.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -464,8 +467,67 @@ public async Task HandlePlaybackRequest(string recordingId, HttpRequest incoming
...
+                for(int i = 0; i < chunks.Length; i++)
+                {
+                    var chunk = chunks[i];
+
+                    byteCountSent+= chunk.Length;
+                    await outgoingResponse.Body.WriteAsync(chunk).ConfigureAwait(false);
+
+                    if (i != chunks.Length - 1)
+                    {
+                        Thread.Sleep(sleepLength);
```
mikeharder, line 524:
```suggestion
                        await Task.Delay(sleepLength);
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -464,8 +467,67 @@ public async Task HandlePlaybackRequest(string recordingId, HttpRequest incoming
...
+                return result;
+            }
+
+            int chunkLength = bodyData.Length / batchCount;
+            int remainder = (bodyData.Length % batchCount);
+            var batches = new byte[batchCount + (remainder > 0 ? 1 : 0)][];
+
+            for(int i = 0; i < batches.Length; i++)
+            {
+                var calculatedChunkLength = i == batches.Length - 1 && batches.Length > 1 && remainder > 0 ? remainder : chunkLength;
```
mikeharder, line 495:
```suggestion
                var calculatedChunkLength = ((i == batches.Length - 1) && (batches.Length > 1) && (remainder > 0)) ? remainder : chunkLength;
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -464,8 +467,67 @@ public async Task HandlePlaybackRequest(string recordingId, HttpRequest incoming
...
+
+        public async Task WriteBodyBytes(byte[] bodyData, int playbackResponseTime, HttpResponse outgoingResponse)
+        {
+            if (playbackResponseTime > 0)
+            {
+                int batchCount = 10;
+                int sleepLength = playbackResponseTime / batchCount;
+
+                byte[][] chunks = GetBatches(bodyData, batchCount);
+                int byteCountSent = 0;
```
mikeharder, line 513:
```suggestion
```

Looks like this is never used?

mikeharder (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: Damnit. Missed during cleanup.

PR #5456, opened 2023-02-15
[Perf] Add parameter TimeoutInMinutes, update pools

mikeharder:


eng/common/pipelines/templates/jobs/perf.yml
```diff
@@ -106,10 +100,6 @@ jobs:
         Commitish: ${{ parameters.ToolsRepoCommitish }}
         WorkingDirectory: $(System.DefaultWorkingDirectory)/azure-sdk-tools
 
-  - template: /eng/common/pipelines/templates/steps/verify-agent-os.yml
```
weshaggard, line 109: Do you think this verification is no longer interesting here? While talking with @konrad-jamrozik about whether or not to keep the vmImage name or not my suggestion is to keep it but in a standardized from (i.e. windows-2022 instead of MM2022) so we could still keep these OS checks in our general pipelines.

mikeharder: I will create another PR to add the standardized image names back.

PR #5473, opened 2023-02-16
change file owner when docker start

dw511214992:


tools/sdk-generation-pipeline/scripts/change-owner.sh
```diff
@@ -11,14 +11,12 @@ done
 if [ -d "${SPEC_REPO}" ]; then
   while true
```
zzvswxy, line 12: how exit the while loop?

tools/sdk-generation-pipeline/scripts/change-owner.sh
```diff
@@ -11,14 +11,12 @@ done
 if [ -d "${SPEC_REPO}" ]; then
```
zzvswxy, line 11: how to exit the script, for line 7-9 while loop

tools/sdk-generation-pipeline/scripts/change-owner.sh
dw511214992: No need to exit while running. You can check the entrypoint.sh, and you will find it runs in backend.

zzvswxy (approved):
Checked with Wei:
1 SPEC_REPO will be cloned out of docker
2 docker run this script to change the generated sdk file's owner and group to pipeline's owner and group
3 the script is a resident process, will work until the docker exit

PR #5479, opened 2023-02-16
[powershell] mirror files

JackTn:


eng/pipelines/mirror-files.yml
```diff
@@ -0,0 +1,25 @@
...
+        GH_TOKEN: $(azuresdk-github-pat)
+        Repos:
+
+          JackTn/TestRepo-One:
+            Branch: "main"
+            Path: "/specification/common-types/"
+            TargetRepos:
+              JackTn/TestRepo-Two:
+                Branch: "main"
+                Path: "/specification/common-types/"
```
raych1, line 22: How are we going to handle multiple paths in same repo?

benbp:
I don't have much context here, can you explain what the purpose of this pipeline is? And how is it different from https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/templates/steps/sync-repo-branch.yml ? We should endeavor to re-use our yaml and github scripts as much as possible, and isolate changes into independent scripts rather than nest them in the yaml, so they can be tested locally more easily.

Also, are these changes being currently tested with the azure-sdk PAT? If so, this is dangerous and we've had issues with this before. Please test with a PAT from your own credentials created with a more limited scope.

A couple other things that stood out to me:
- There are many calls to `git push --force` and I think this is one misconfiguration away from us making damaging changes to the wrong repository and/or branch. Why is a force push necessary?
- Let's use the PAT directly via the variable instead of passing it around via a github parameter. See https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/steps/create-pull-request.yml#L51 for example. It seems like the existing pipeline that you are copying from is doing this too, but ideally we don't continue with this method.

konrad-jamrozik:
If we decide to keep the added `.yml` files, I would appreciate some comments in them explaining what they do, what are their inputs/output, and how to use them. Plus some PR description providing the broader context why this change is made.

eng/pipelines/templates/steps/sync-repo-branch.yml
```diff
@@ -79,9 +79,10 @@ steps:
           if (!$TargetBranch) {
             $TargetBranch = $defaultBranch
           }
+          Write-Host ${{ target.value.Rebase }}
           if (-not '${{ target.value.Rebase }}') {
             git checkout -B target_branch source_branch
-            git push Target "target_branch:refs/heads/${TargetBranch}"
+            git push --force Target "target_branch:refs/heads/${TargetBranch}"
```
weshaggard, line 85: Why are we force pushing? We should never have any scenario in our syncing that requires a force push.

JackTn:
Work Changes, pending for this change and  close this pr

JackTn:
tips: this code is for extract powershell script from yml

weshaggard:
@raych1 @JackTn did we have an issue tracking this work that we can link to this initial work once we pick this up again?

PR #5515, opened 2023-02-17
Handle resource group deprovision race condition

benbp:
Preventing spurious pipeline warning emails when we try to tag resource groups that have been marked for deletion after we queried them. For example https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2199954&view=logs&j=a460d732-23aa-5493-a411-cc406067acf8&t=2ae92968-c003-547a-10d9-00fa5e34d21d&l=323

### DIFF
--- a/eng/scripts/live-test-resource-cleanup.ps1
+++ b/eng/scripts/live-test-resource-cleanup.ps1
@@ -285,7 +285,18 @@ function FindOrCreateDeleteAfterTag {
     $deleteAfter = [datetime]::UtcNow.AddHours($DeleteAfterHours)
     if ($Force -or $PSCmdlet.ShouldProcess("$($ResourceGroup.ResourceGroupName) [DeleteAfter (UTC): $deleteAfter]", "Adding DeleteAfter Tag to Group")) {
       Write-Host "Adding DeleteAfter tag with value '$deleteAfter' to group '$($ResourceGroup.ResourceGroupName)'"
-      $ResourceGroup | Update-AzTag -Operation Merge -Tag @{ DeleteAfter = $deleteAfter }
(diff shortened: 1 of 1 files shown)


eng/scripts/live-test-resource-cleanup.ps1
```diff
@@ -285,7 +285,18 @@ function FindOrCreateDeleteAfterTag {
     $deleteAfter = [datetime]::UtcNow.AddHours($DeleteAfterHours)
     if ($Force -or $PSCmdlet.ShouldProcess("$($ResourceGroup.ResourceGroupName) [DeleteAfter (UTC): $deleteAfter]", "Adding DeleteAfter Tag to Group")) {
       Write-Host "Adding DeleteAfter tag with value '$deleteAfter' to group '$($ResourceGroup.ResourceGroupName)'"
-      $ResourceGroup | Update-AzTag -Operation Merge -Tag @{ DeleteAfter = $deleteAfter }
+      $result = ($ResourceGroup | Update-AzTag -Operation Merge -Tag @{ DeleteAfter = $deleteAfter }) 2>&1
+      if ("Exception" -in $result.PSObject.Properties.Name) {
+          # Handle race conditions where the group starts deleting after we get its info, in order to avoid pipeline warning/failure emails
+          # "The resource group '<group name>' is in deprovisioning state and cannot perform this operation"
+          if ($result.Exception.Message -like '*is in deprovisioning state*') {
```
mikeharder, line 292: I might be misunderstanding, but is this logic backwards?

mikeharder (changes requested):

eng/scripts/live-test-resource-cleanup.ps1
benbp: It is! I was testing the inverse condition and forgot to change it back (facepalm)

PR #5431, opened 2023-02-13
Remove the obsolete, prefix-based `CODEOWNERS` matcher

konrad-jamrozik:
A follow-up to:
- https://github.com/Azure/azure-sdk-tools/pull/5088

With this PR, now only the new, regex-based, wildcard-supporting `CODEOWNERS` matcher is present in the codebase.

This PR also:
- removes the obsolete `ProgramSimplePathTests` class (restored and upgraded in #5518);
- renames `ProgramGlobPathTests` to `RetrieveCodeOwnersProgramTests`;
- updates comments to remove mentions of the old, prefix-based matcher;
- makes few comment improvements to classes testing the new matcher;
- removes obsolete tests from `get-codeowners.ps1`.
Related PR:
- https://github.com/Azure/azure-sdk-tools/pull/5103

This PR contributes to:
- https://github.com/Azure/azure-sdk-tools/issues/2770

### Merging prerequisite

This PR can be merged only after the following PRs are merged and relevant NuGet package published

- https://github.com/Azure/azure-sdk-tools/pull/5103
- https://github.com/Azure/azure-sdk-tools/pull/5431

PR #5518, opened 2023-02-19
Restore and overhaul simple target path tests into `RetrieveCodeOwnersProgramTests`

konrad-jamrozik:
This PR is a follow-up to:
- #5431

Here I recreate tests that exercise the "simple target path" (as opposed to "glob target path") logic of the `RetrieveCodeOwners` tool.

Related work:
- https://github.com/Azure/azure-sdk-tools/issues/2770


weshaggard (approved):
looks reasonable

Issue #5530, opened 2023-02-21
Remove ability to set local spec root in cadl-location.yaml


m-nash:
https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Cadl-Project-Sync.ps1#L117

https://github.com/Azure/azure-sdk-tools/issues/5288#issuecomment-1439043756

konrad-jamrozik:
@weshaggard I am removing this from `Spec PR Tools` as it looks to be `SDK Automation` instead to me. FYI in case I am mistaken.

Issue #5288, opened 2023-01-31
[FEATURE REQ] Add support for a local repo in the cadl scripts


m-nash:
There are two ways to refer to the cadl files. One is a remote git url, the other is the local file path. In the cadl-location.yaml, you can provide (repo, commit) to indicate an remote URL of the cadl, then cadl-project-sync.ps1 will sparse-checkout the cadl folder. Also you can provide a local swagger repo folder via spec-root-dir, cadl-project-sync.ps1 just copy cadl filed to the TempCadlFiles.

previous comments for context
https://github.com/Azure/azure-sdk-tools/pull/5272#discussion_r1091364117
https://github.com/Azure/azure-sdk-tools/pull/5272#discussion_r1092234754

tg-msft:
+1 for this feature.  It's a showstopper for iterative Cadl development if I can't see what the C# API will look like without committing and pushing a draft spec every time I make a small adjustment.

m-nash:
@tg-msft  you can do this today (although one check in is needed).  If you add `cleanup: false` to your cadl-location.yaml it will leave the `TempCadlFiles` in place in your project directory.  You can then cd into there and make any local changes to the cadl directly that you want and then simply rerun the cadl compile command from within that directory

mikekistler:
Our understanding is that the request is to use a local specs repo that is already present rather than sparse cloning the relevant path from the specs repo.

raych1:
Talked to @chunyu3, this is already supported. Can you link the code location here?

chunyu3:
> Our understanding is that the request is to use a local specs repo that is already present rather than sparse cloning the relevant path from the specs repo.

If you want to use a local specs repo that is already present, you can specify `spec-root-dir` to provide the path of the root directory of the specs repo, e.g. /home/azure-rest-api-specs, and don't provide the `commit` and `repo`.

weshaggard:
@chunyu3 unless I'm missing something wouldn't someone need to manually change their cadl-location.yaml file to set that directory? https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Cadl-Project-Sync.ps1#L117 Ideally, we would support such flags with whatever root script we expect folks to use.

m-nash:
It looks like I accidentally left some of the old code that pulled this from cadl-location.yaml which should be removed.  This feature should be a temporary flag passed in as we would never expect a cadl-location.yaml to be checked in with this setting.

[...]

PR #5539, opened 2023-02-22
update the properties of cadl-location.yaml

chunyu3:
- add `spec-root-dir` property in the doc


doc/common/Cadl-Project-Scripts.md
```diff
@@ -91,8 +91,9 @@ This file should live under the project directory for each service and has the f
...
-| <a id="repo-anchor"></a> repo | The repo this spec lives in.  This should be either `Azure/azure-rest-api-specs` or `Azure/azure-rest-api-specs-pr`.  Note that pr will work locally but not in CI until we add another change to handle token based auth. | true |
+| <a id="repo-anchor"></a> repo | The repo this spec lives in.  This should be either `Azure/azure-rest-api-specs` or `Azure/azure-rest-api-specs-pr`.  Note that `Azure/azure-rest-api-specs-pr` repo will work locally but not in CI until we add another change to handle token based auth. | true |
 | <a id="cleanup-anchor"></a> cleanup | This will remove the TempCadlFiles directory after generation is complete if true otherwise this directory will be left to support local changes to the files to see how different changes would affect the generation. | false: default = true |
+|<a id="specRootDirectory-anchor"></a> spec-root-dir | When you want to use the local specs repo  that is already present, you can specify the root directory of the repo. e.g. /home/azure-rest-api-specs| false|
```
weshaggard, line 96: I would suggest we not document this as I don't think we want it long term. See https://github.com/Azure/azure-sdk-tools/issues/5288 and https://github.com/Azure/azure-sdk-tools/issues/5530

cc @m-nash

m-nash (changes requested):
This property creates some scenarios we want to avoid lets solve the issues associated with this concept instead then update the documentation accordingly.  For now this is an unsupported scenario.

Issue #2453, opened 2021-12-15
Test proxy doesn't normalize line endings


pakrym:
![image](https://user-images.githubusercontent.com/1697911/146276401-deea4098-ac13-4ee4-8910-5a203359dd80.png)

@JoshLove-msft @scbedd

scbedd:
Doing some digging into this. Found [this issue](https://github.com/dotnet/runtime/issues/54715) which is basically exactly our scenario.

So I think we need to create a custom `Utf8JsonWriter` and override the `WriteIndent` call to insider `\n` instead of `Environment.NewLine`.

We should also update our we _load_ the recording to ignore `crlf` when deserializing from the recording file.

JoshLove-msft:
> Doing some digging into this. Found [this issue](https://github.com/dotnet/runtime/issues/54715) which is basically exactly our scenario.
> 
> So I think we need to create a custom `Utf8JsonWriter` and override the `WriteIndent` call to insider `\n` instead of `Environment.NewLine`.
> 
> We should also update our we _load_ the recording to ignore `crlf` when deserializing from the recording file.

I think the issue is with the actual content, not the formatted JSON. When matching we renormalize the JSON anyways so it should handle this for us. But the issue happens when someone adds line breaks based on Environment.NewLine from their own test.

scbedd:
> When matching we renormalize the JSON anyways so it should handle this for us.

Yeah I'm seeing this now.

> But the issue happens when someone adds line breaks based on Environment.NewLine from their own test into the request content

Hmmm. So add a hardcoded replace on `\r\n` if the request content is a `text` content type? 

> when someone adds line breaks based on Environment.NewLine from their own test into the request content

I can see how this might be the proxy's place to do this, but I'm super leery towards it. Bright side is that this is _not_  a breaking change. Existing recordings would just have `crlf` content silently replaced when matching.

JoshLove-msft:
Agreed that this shouldn't be a breaking change. I think essentially normalizing the line breaks when matching is the way to go.

PR #5548, opened 2023-02-22
Add Sanitizer that Normalizes Text Bodies

scbedd:
related to #2453

...and discover that the NormalizeJsonBodies already applies in the testcase I created.

@JoshLove-msft how were you getting `\r\n` injected into the request in a way that didn't auto normalize? I'd assume it's because it's NOT a json body and still somehow includes `\r\n`. multipart/mixed?

I'm not certain how the management folks ran into this. What does `normalize` even mean in this case? The only way it wouldn't be auto-normalized is if the user deliberately sent a request with this material within it, in which case I don't want to replace it. That's what was recorded.


JoshLove-msft:
> @JoshLove-msft how were you getting `\r\n` injected into the request in a way that didn't auto normalize? I'd assume it's because it's NOT a json body and still somehow includes `\r\n`. multipart/mixed?

It looks like this could happen for any non-JSON text content types:
https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordEntry.cs#L129

PR #5552, opened 2023-02-23
Creating `config` verb and adding `locate/show` sub-verbs

### LINKED ISSUE #4652: [Prioritized] Add CLI verb `config`
In short, we want an "official" way to interact with the assets.json only involving the test-proxy and now pshell to do things like create it.

- [ ] test-proxy config initialize (pending, I don't have a good way to discover `language` in an automated way
- [ ] test-proxy config show (shows details about config, relevant paths, etc)
- [ ] test-proxy config locate (shows location of assets repo for config file)
- [ ] test-proxy config diff (generates a git link that diffs current and previous tags)

scbedd:
Resolves #4652 

@JimSuplizio wanted to get your opinion on this. It appears that `CommandLineParser` doesn't treat `sub-verbs`/`sub-commands` as a thing. [Example of what I'm looking for](https://learn.microsoft.com/en-us/dotnet/standard/commandline/define-commands#define-subcommands). 

[Here is where I was inspired in my approach to skip and pass to different parser](https://github.com/commandlineparser/commandline/issues/13#issuecomment-429634063).

I think that outside of doing what I do here, we are going to need to rewrite to `System.CommandLine` to get the sub-command experience we want.

Honestly this feels pretty hacky, though I think that the results for the users is what we want/expect.

Example Usage:

```bash
test-proxy.exe config locate -a sdk/keyvault/azure-keyvault-keys/assets.json --storage-location=C:/repo/sdk-for-python
C:/repo/sdk-for-python/.assets/Y0iKQSfTwa/python
```

todo:

- [ ] clean up hackiness a bit
- [ ] Add integration tests for our expected scenarios

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
@@ -507,7 +507,7 @@ namespace Azure.Sdk.Tools.TestProxy
                 path = Path.Join(contextDirectory, pathToAssetsJson.Replace("\\", "/"));
             }
 
-            return path;
+            return path.Replace("\\", "/");
         }
 
         public async Task Restore(string pathToAssetsJson)
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Startup.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Startup.cs
@@ -27,6 +27,8 @@ using System.Collections.Generic;
 using System.Linq;
 using Azure.Sdk.Tools.TestProxy.CommandParserOptions;
 using Azure.Sdk.Tools.TestProxy.CommandParserOptions.ConfigVerbs;
+using CommandLine.Text;
+using Microsoft.AspNetCore.Mvc;
 
 namespace Azure.Sdk.Tools.TestProxy
 {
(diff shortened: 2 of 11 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -507,7 +507,7 @@ public static string GetAssetsJsonLocation(string pathToAssetsJson, string conte
                 path = Path.Join(contextDirectory, pathToAssetsJson.Replace("\\", "/"));
             }
 
-            return path;
+            return path.Replace("\\", "/");
```
scbedd, line 510: Removing.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
```diff
@@ -66,16 +66,16 @@ public GitStore(IConsoleWrapper consoleWrapper)
...
             var config = await ParseConfigurationFile(pathToAssetsJson);
 
             if (!string.IsNullOrWhiteSpace(config.AssetsRepoPrefixPath.ToString()))
             {
-                return Path.Combine(config.AssetsRepoLocation.ToString(), config.AssetsRepoPrefixPath.ToString());
+                return new NormalizedString(Path.Combine(config.AssetsRepoLocation.ToString(), config.AssetsRepoPrefixPath.ToString()));
             }
 
-            return config.AssetsRepoLocation.ToString();
+            return new NormalizedString(config.AssetsRepoLocation.ToString());
```
hallipr, line 78: What was the output like before the change to NormalizedString?

scbedd: Path1 + Path2 = "blah\\blah\\blah\\blah/python"

^ I like my paths to be entirely consistent, and display with `/` not `\`.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Startup.cs
```diff
@@ -56,22 +60,83 @@ private static string resolveRepoLocation(string storageLocation = null)
...
+            else
+            {
+                await parser.ParseArguments<StartOptions, PushOptions, ResetOptions, RestoreOptions, ConfigOptions>(args)
+                    .WithNotParsed(ExitWithError)
+                    .WithParsedAsync(Run);
+            }
         }
 
-        static void ExitWithError(IEnumerable<Error> errors)
+        static void SubVerbExitWithError<T>(ParserResult<T> result, IEnumerable<Error> errors, string defaultHelpText)
```
hallipr, line 95: Can this be encapsulated in ConfigOptions class as all the subverbs related to "config"?

scbedd: Maybe. I'll try.

PR #5572, opened 2023-02-25
Add dotnet-trace as profiler for Net

AlexanderSher:
Validation build: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2593387&view=results

### DIFF
--- a/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Net.cs
+++ b/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Net.cs
@@ -100,7 +100,7 @@ namespace Azure.Sdk.Tools.PerfAutomation
             var dllPath = Path.Combine(PublishDirectory, dllName);
 
             var result = profile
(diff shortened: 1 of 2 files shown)


mikeharder (approved):

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Net.cs
```diff
@@ -121,6 +121,12 @@ public class Net : LanguageBase
             };
         }
 
+        private string GetProfileOutputFile(string testName, string arguments)
+        {
+            var fileName = $"{testName}_{arguments.Replace(" --", "_").Replace("--", "_").Replace(" ", "-")}_{DateTime.Now.ToString("yyyyMMdd_hhmmss")}.nettrace";
```
mikeharder, line 126: You may also want to add the version of the primaryPackage to determine if the profile is for `source` or a GA version like `12.14.1`.

PR #5577, opened 2023-02-27
Use --tags and --name for rotation config filtering

hallipr:
The new command syntax is:

```
> secrets list --tags test, ado-pat --config ../path/to/config/root/

name: pipeline-pat-test - tags: test, ado-pat, pipeline-witness
name: generator-pat-test - tags: test, ado-pat, pipeline-generator

> secrets rotate --tags test, ado-pat --config ../path/to/config/root/
```

this allows us to nest configs in subfolders and target them by name or tags


benbp:
To my eyes this over-complicates the CLI UX a bit and makes it hard to use intuitively, because it moves from a simple `-c` flag to having a `-p` flag for the root path and a positional value without a flag that's structured as a csv.

Do we anticipate having use cases where calling a few back to back rotate commands with inclusive glob paths is significantly more cumbersome than adding a negative glob? That is to say, is it enough to have inclusive globs added to the existing `--configuration` flag only?


```
secrets rotate -c ../path/to/config/root/foo/**/rotate/this.json
secrets rotate -c ../path/to/config/root/bar/**/rotate/this.json

# or

secrets rotate -c ../path/to/config/root/**/rotate/this.json --exclude ../path/to/config/root/not/**/this.json
```

weshaggard:
I tend to agree with @benbp I don't know that we need that complex set of pattern matching. We can always move configs around to make the patterns easier.

benbp:
@weshaggard we talked offline and came up with the following approach that uses tags in the json instead of directory structure (below is copy/pasted from @hallipr):

- schema to a nice place in the repo so urls are shorter
- tags in the config so we can positive match tags in cli
- schema url in the configs so we can skip json files that aren't rotation configs and error on invalid files that are
- add List command so we can see the configs matching our tags before we do anything.   like an even do-nothinger form of whatif

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Configuration/PlanConfiguration.cs
```diff
@@ -1,37 +1,55 @@
...
+        Converters =
+        {
+            new JsonStringEnumConverter()
+        },
+        PropertyNameCaseInsensitive = true,
+        PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+        ReadCommentHandling = JsonCommentHandling.Skip,
+    };
+
+    private static readonly Regex schemaPattern = new (@"https\://raw\.githubusercontent\.com/azure/azure-sdk-tools/(?<branch>.*?)/schemas/secretrotation/(?<version>.*?)/schema\.json", RegexOptions.IgnoreCase);
```
benbp, line 21: The version regex segment matches optional/any number, but I think we would want to require it?

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Configuration/RotationConfiguration.cs
```diff
@@ -18,60 +18,52 @@ public class RotationConfiguration
...
-    public RotationPlan? GetRotationPlan(string name, ILogger logger, TimeProvider timeProvider)
-    {
-        PlanConfiguration? planConfiguration =
-            PlanConfigurations.FirstOrDefault(configuration => configuration.Name == name);
-
-        return planConfiguration != null
-            ? ResolveRotationPlan(planConfiguration, logger, timeProvider)
-            : null;
-    }
+        if (names.Any())
```
benbp, line 45: I think if the name and tags arguments are passed, we should fail to the user as those are mutually exclusive arguments in my mind?

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Tests/CoreTests/PlanConfigurationTests.cs
```diff
@@ -5,29 +5,42 @@ namespace Azure.Sdk.Tools.SecretRotation.Tests.CoreTests;
...
+    [Test]
+    public void TryLoadFromFile_ValidPath_ReturnConfiguration()
     {
         string configurationPath = TestFiles.ResolvePath("TestConfigurations/Valid/random-string.json");
 
         // Act
-        PlanConfiguration configuration = PlanConfiguration.FromFile(configurationPath);
+        bool success = PlanConfiguration.TryLoadFromFile(configurationPath, out var configuration);
 
+        Assert.True(success);
```
benbp, line 43: Is there any value in adding tag tests as well?

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Configuration/PlanConfiguration.cs
hallipr: True.   I'll update to `(?<branch>.+?)`  and `(?<version>.+?)`.   I don't want to bake semver checking into the regex match.   I'll leave it up to version comparison later in the file if we want to restrict to specific versions or branches.   For now, I just want to separate is-our-file from not-our-file.

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Tests/CoreTests/PlanConfigurationTests.cs
hallipr: There is, but I'd want to mock the file system.  I wouldn't want the name/tag filtering test to be affected by test files being added or removed from the folder.

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Configuration/RotationConfiguration.cs
benbp: Ah ok, I was thinking name was the actual file name, not a possibly ambiguous identifier.

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Tests/CoreTests/PlanConfigurationTests.cs
benbp: If you fed an array of de-serialized objects to the tag filter method, and checked the output, that wouldn't have to touch the fs.

hallipr: I went ahead with adding another test config folder named "TagMatching". any changes in the folder could cause issues with tag related tests.  We'll catch them in PR if it happens

hallipr: Tag and name filtering happens as 2 linq statements in the large config reader method:

```cs
        if (names.Any())
        {
            planConfigurations = planConfigurations
                .Where(plan => names.Contains(plan.Name, StringComparer.OrdinalIgnoreCase));
        }

        if (tags.Any())
        {
            planConfigurations = planConfigurations
                .Where(plan => tags.All(tag => plan.Tags.Contains(tag, StringComparer.OrdinalIgnoreCase)));
        }
```

I could extract those lines to a separate testable method, but at that point, we'd mostly be testing LINQ.

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Configuration/RotationConfiguration.cs
hallipr: with folders names not being a component of the plan name, file name based plan names could still be ambiguous if the plans are in separate folders

benbp: So will this scenario no longer work, or still work?

> secrets list --config ../path/to/config/allthingsinhere/

benbp (approved):

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Configuration/RotationConfiguration.cs
hallipr: Will work, won't apply any filtering to the plans found in the folder, recursive search

PR #5587, opened 2023-03-01
Add script to download latest contract

pshao25:
This is a part of https://github.com/Azure/autorest.csharp/issues/2887. We need to download the latest GA version code. Workflow of this script.
image


weshaggard:
@pshao25 do you have a linked issue that describes more of what we are trying to do here? Also I have to wonder if what the performance of this will end up being if we have to clone more places down every time we regenerate.

eng/common/scripts/Download-Latest-Contract.ps1
```diff
@@ -0,0 +1,71 @@
+[CmdletBinding()]
+param (
+    [Parameter(Position=0)]
+    [ValidateNotNullOrEmpty()]
+    [string] $ProjectDirectory,
+    [ValidateNotNullOrEmpty()]
+    [String] $TagName
```
m-nash, line 7: Can we determine this from the tag name?  I think for each language it might be a different algo but we can use the delegate pattern [like this](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Cadl-Project-Generate.ps1#L57).

pshao25:
> @pshao25 do you have a linked issue that describes more of what we are trying to do here? Also I have to wonder if what the performance of this will end up being if we have to clone more places down every time we regenerate.

We are going to do https://github.com/Azure/autorest.csharp/issues/2887, in which the generated API depends on what the existing GAed API looks like. So we want a script to download the GA code. And we have other similar ongoing features need this "download code" script.

eng/common/scripts/Cadl-Project-Sync.ps1
```diff
@@ -9,14 +9,8 @@ param (
 
 $ErrorActionPreference = "Stop"
 . $PSScriptRoot/Helpers/PSModule-Helpers.ps1
+. $PSScriptRoot/Helpers/Sparse-Clone-Helpers.ps1
```
m-nash, line 12: no changes should be in this file as its deprecated and will be removed shortly

eng/common/scripts/Download-Latest-Contract.ps1
```diff
@@ -0,0 +1,78 @@
...
+
+    Push-Location $sdkCloneDir.Path
+    try {        
+        $projectRelativePath = GetProjectRelativePath
+        if (!(Test-Path ".git")) {
+            InitializeSparseGitClone $sdkGitRemote
+            AddSparseCheckoutPath $projectRelativePath
+        }
+        git checkout $latestCommit
+        return "Latest contract found: $(Join-Path $sdkCloneDir.Path $projectRelativePath)"
```
m-nash, line 71: I assume the caller of this script is supposed to parse out the directory?

Can we simply return the directory here and remove the extra words?

eng/common/scripts/Download-Latest-Contract.ps1
```diff
@@ -0,0 +1,78 @@
+[CmdletBinding()]
```
m-nash, line 1: Can we add documentation on how language teams are supposed to integrate with this file?

Touch on areas like
- what if you have no ga releases yet
- which functions in do you need to override in your language-settings.ps1
- what return value you are supposed to expect and what to do with empty string
- where the latest contract will end up
- etc

eng/common/scripts/Download-Latest-Contract.ps1
```diff
@@ -0,0 +1,78 @@
...
+    
+    $sdkCloneDir = GetSparseCloneDir $ProjectDirectory $projectName "sdk"
+
+    Write-Host "Setting up sparse clone for $projectName at $sdkCloneDir"
+
+    Push-Location $sdkCloneDir.Path
+    try {        
+        $projectRelativePath = GetProjectRelativePath
+        if (!(Test-Path ".git")) {
+            InitializeSparseGitClone $sdkGitRemote
```
m-nash, line 67: Would it be helpful if different languages could give different sub directories that they want?

For instance dotnet would only want `./src` but others will want other things?

eng/common/scripts/Helpers/Sparse-Clone-Helpers.ps1
```diff
@@ -0,0 +1,34 @@
...
+    if ($LASTEXITCODE) { exit $LASTEXITCODE }
+    git sparse-checkout init
+    if ($LASTEXITCODE) { exit $LASTEXITCODE }
+    Remove-Item $sparseCheckoutFile -Force
+}
+
+function AddSparseCheckoutPath([string]$subDirectory) {
+    if (!(Test-Path $sparseCheckoutFile) -or !((Get-Content $sparseCheckoutFile).Contains($subDirectory))) {
+        Write-Output $subDirectory >> .git/info/sparse-checkout
+    }
```
benbp, line 33: I would let git handle all this for you.

```suggestion
    git sparse-checkout add $subDirectory
    if ($LASTEXITCODE) { exit $LASTEXITCODE }
```

For the sparse checkout file exists check, it would be better to try to add the path anyway, because you would want the script to fail quickly if you were trying to add a path before initializing the sparse checkout. Otherwise you might think a behavior is happening when it's not.

For the sparseCheckoutFile contains check, the ordering of the sparse checkout file matters. So you could have a path specified for inclusion on line 1, but then another path pattern on line 2 that excludes the path on line 1. So in this case, if a user of this script was building a set of patterns, you wouldn't want to make ordering decisions for them. I think this is important because if we're writing a centralized helper script we'll probably start using it in more places.

eng/common/scripts/Download-Latest-Contract.ps1
pshao25: Three facts here:
1. This is not an actual return, it just prints the return value to stdout.
2. There are other strings printed to stdout like:
`Write-Host "Setting up sparse clone for $projectName at $sdkCloneDir"`
3. To pass data between two processes, we redirect the stdout to the pipe created between them.

When we start this script from another process (in this case it is generator), we should have some indicator to let that process know from where we could start to take as the actual information we want. I don't have better ideas than adding a "Latest contract found:" here, do you?

eng/common/scripts/Helpers/Sparse-Clone-Helpers.ps1
pshao25: > if you were trying to add a path before initializing the sparse checkout. 

I'm afraid we are always doing this. After running "git sparse-checkout init", the autogenerated sparse-checkout file contains `/*` and `!/*/`, which are not what we want. So we remove the sparse-checkout file in function `InitializeSparseGitClone`.

> For the sparseCheckoutFile contains check, the ordering of the sparse checkout file matters.

I'm not sure I follows your concern. We don't add any excludes pattern. We don't even add pattern. We only add literal paths.

eng/common/scripts/Download-Latest-Contract.ps1
```diff
@@ -0,0 +1,84 @@
...
+    return $result
+}
+
+function GetProjectRelativePath() {
+    $rootPath = GetProjectRootPath $ProjectDirectory
+    $subDirectory = &$GetSourceCodeSubDirectory
+    return [System.IO.Path]::GetRelativePath($rootPath, (Join-Path $ProjectDirectory $subDirectory)).Replace("\","/")
+}
+
+if (!(Test-Path "Function:$GetLatestTagFn")) {
```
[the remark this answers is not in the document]
weshaggard: All the languages use release tags and to determine if one is a GA or not you look at the versions. You can see an example of where I did that at https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Update-Release-Versions.ps1#L201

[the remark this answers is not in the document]
pshao25: Seems these functions are retrieving data from https://github.com/Azure/azure-sdk/tree/main/_data/releases/latest. And it seems still need specific logic to each language to get the package name. For example, .net uses the folder name, while js uses the name indicated in package.json.

How about we just parse the `CHANGELOG.md` to get the tag? Then we don't need any cross-repo effort.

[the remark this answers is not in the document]
weshaggard: That isn't use the csv data this process is about updating that csv data. The release tags are gathered from github in https://github.com/Azure/azure-sdk/blob/main/eng/scripts/PackageVersion-Helpers.ps1. Once you gather them you can parse and sort them to get the latest version for a given package. If getting the package name is the challenge there is already code in each language to the package information (example https://github.com/Azure/azure-sdk-for-net/blob/main/eng/scripts/Language-Settings.ps1#L13). You cannot depend on the folder name as it doesn't always match the package name. 

We also have helpers that can parse the changelog but you still need to find the right one and depending on what versions you are looking at branch it may or may not have all the correct versions in there. For example, if you are working in a feature branch that hasn't taken the latest merge from may where the last GA shipped you will miss that GA version.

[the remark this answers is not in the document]
pshao25: So, to get package name we still need a language specific function, correct?

And the script `PackageVersion-Helpers` calls github grapql api. That needs pat, which is not expected. We don't want our users to have such additional steps. We hope user don't need to know there is a download step.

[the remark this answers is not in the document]
weshaggard: That example uses a PAT and calls the graphql but if you have a local clone you can also get the list of tags from the cloned repo as well.

PR #5591, opened 2023-03-01
Add devops-pat and az website stores to secret rotation

hallipr:
- Add Azure DevOps PAT store for service accounts
  - This uses service account credentials defined in the store's parameters to authenticate into ADO
  - PATs are issued against the service account, not the tool runner
- Add Azure App Service store for restarting websites after rotation
  - The store is configured with UpdateAfterPrimary so the website restart happens after the secret is stored in the vault.
- Update rotation logic to support post-primary stores and rotation-complete annotation
- To accommodate the post-primary website restart, the new rotation flow is:
    1. Originate in origin
    2. Write to Secondaries where UpdateAfterPrimary == false
    3. Write to Primary if Primary != Origin
    4. Write to Secondaries where UpdateAfterPrimary == true
    5. Annotate Primary with "rotation-complete: true"

### DIFF
--- a/tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Core/RotationPlan.cs
+++ b/tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Core/RotationPlan.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.ObjectModel;
+using System.Collections.ObjectModel;
 using Azure.Sdk.Tools.SecretRotation.Core;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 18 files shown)


tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Core/RotationPlan.cs
```diff
@@ -134,51 +134,67 @@ private async Task RotateAsync(string operationId, SecretState currentState, boo
         DateTimeOffset invocationTime = this.timeProvider.GetCurrentDateTimeOffset();
 
         SecretValue newValue = await OriginateNewValueAsync(operationId, invocationTime, currentState, whatIf);
+        // TODO: some providers will issue secrets for longer than we requested. Should we propagate the real expiration date, or the desired expiration date?
```
benbp, line 137: Do we have a difference between expiration date and desired rotation date?

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Core/RotationPlan.cs
```diff
@@ -134,51 +134,67 @@ private async Task RotateAsync(string operationId, SecretState currentState, boo
...
-        await WriteValueToSecondaryStoresAsync(newValue, currentState, whatIf);
+        if (OriginStore != PrimaryStore)
+        {
+            await WriteValueToPrimaryAsync(newValue, currentState, invocationTime, whatIf);
+        }
 
-        await WriteValueToPrimaryAndOriginAsync(newValue, invocationTime, currentState, whatIf);
+        await WriteValueToPostPrimaryStoresAsync(newValue, currentState, whatIf);
+
+        await MarkRotationCompleteAsync(newValue, currentState, invocationTime, whatIf);
```
benbp, line 148: Unrelated to this PR, but would it be easier to set whatIf on the class vs. having to allow for it in all method signatures and remember to propagate?

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Stores.AzureAppService/AzureWebsiteStore.cs
```diff
@@ -0,0 +1,114 @@
...
+                credential,
+                logger);
+        };
+    }
+
+    public override async Task WriteSecretAsync(SecretValue secretValue, SecretState currentState, DateTimeOffset? revokeAfterDate, bool whatIf)
+    {
+        ResourceIdentifier resourceId = WebSiteResource.CreateResourceIdentifier(this.subscriptionId, this.resourceGroupName, this.websiteName);
+        ArmClient client = new ArmClient(this.credential);
+        WebSiteResource website = client.GetWebSiteResource(resourceId);
```
benbp, line 71: Do we want to wrap all of this in the whatIf? Or is whatIf used in the tool to still perform reads.

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Stores.AzureDevOps/ServiceAccountPersonalAccessTokenStore.cs
```diff
@@ -0,0 +1,262 @@
...
+                parameters.RevocationAction,
+                credential,
+                secretProvider);
+        };
+    }
+
+    public override async Task<SecretValue> OriginateValueAsync(SecretState currentState, DateTimeOffset expirationDate, bool whatIf)
+    {
+        VssConnection connection = await GetConnectionAsync();
+        TokensHttpClient client = connection.GetClient<TokensHttpClient>();
```
benbp, line 120: Same comment above re: whatIf scope.

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Stores.AzureDevOps/ServiceAccountPersonalAccessTokenStore.cs
```diff
@@ -0,0 +1,262 @@
...
+                        $"https://dev.azure.com/{this.organization}");
+                    
+                    await client.RevokeAsync(authorizationId);
+                }
+                catch (SessionTokenNotFoundException)
+                {
+                    // ignore "not found" exception on delete
+                }
+            }
+        };
```
benbp, line 200:
```suggestion
            if (whatIf)
            {
                this.logger.LogInformation(
                    "WHAT IF: Remove PAT with authorization id '{AuthorizationId}' from '{OrganizationUrl}'",
                    authorizationId,
                    $"https://dev.azure.com/{this.organization}");

                return;
            }

            try
            {
                this.logger.LogInformation(
                    "Removing PAT with authorization id '{AuthorizationId}' from '{OrganizationUrl}'",
                    authorizationId,
                    $"https://dev.azure.com/{this.organization}");

                await client.RevokeAsync(authorizationId);
            }
            catch (SessionTokenNotFoundException)
            {
                // ignore "not found" exception on delete
            }
        };
```

benbp:
Some questions/comments but LGTM

benbp (approved):

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Core/RotationPlan.cs
hallipr: could do.   I was intentionally avoiding putting state on the class, but that doesn't really make sense as it's functionally a stateful rotation context.

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Stores.AzureAppService/AzureWebsiteStore.cs
hallipr: I want whatif to do everything except for volatile writes.   I'd like to catch access and configuration issues in --what-if if possible. i.e.  --what-if shouldn't tell me "WHATIF: Restart Azure Website foo" if the current principal doesn't have access to the subscription.

hallipr: I want the user to feel safe doing a --what-if whenever.   I also want --what-if to find as many configuration and access issues as possible.

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Stores.AzureDevOps/ServiceAccountPersonalAccessTokenStore.cs
hallipr: same here.  whatif should do everything but the write.

tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Core/RotationPlan.cs
hallipr: we do have instances where an origin will issue a secret with an expiration date other than what was requested

PR #5598, opened 2023-03-02
Fix PS optional arg issue to not pass empty args

weshaggard:
Another follow-up of https://github.com/Azure/azure-sdk-tools/pull/4746 when moving to PS 7.3.

Follow-up to https://github.com/Azure/azure-sdk-tools/pull/4746

### DIFF
--- a/eng/common/scripts/git-branch-push.ps1
+++ b/eng/common/scripts/git-branch-push.ps1
@@ -24,8 +24,9 @@ param(
     [Parameter(Mandatory = $true)]
     [string] $GitUrl,
 
+    # Explicitly set this to null so that PS command line parser doesn't try to parse and pass it as ""
     [Parameter(Mandatory = $false)]
-    [string] $PushArgs = "",
+    [string] $PushArgs = $null,
 
     [Parameter(Mandatory = $false)]
     [string] $RemoteName = "azure-sdk-fork",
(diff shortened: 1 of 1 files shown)


benbp (approved):

konrad-jamrozik (approved):

konrad-jamrozik:
For reference, I suspect this is the change:
https://learn.microsoft.com/en-us/powershell/scripting/whats-new/what-s-new-in-powershell-73?view=powershell-7.3#experimental-features

> [PSNativeCommandArgumentPassing](https://learn.microsoft.com/en-us/powershell/scripting/learn/experimental-features?view=powershell-7.3#psnativecommandargumentpassing) - PowerShell now uses the ArgumentList property of the StartProcessInfo object rather than the old mechanism of reconstructing a string when invoking a native executable.

https://learn.microsoft.com/en-us/powershell/scripting/learn/experimental-features?view=powershell-7.3#psnativecommandargumentpassing

image

@weshaggard FYI

weshaggard:
We aren't currently blocked since 1ES rolled back their image update. The more I think about and test this I think for at least this script where we call git alot I want to set `$PSNativeCommandArgumentPassing = "Legacy"` just to avoid any other weird behaviors here.

PR #5600, opened 2023-03-02
Update swagger manual review gen pipeline to use a variable for swagg…

praveenkuttappan:
Update swagger manual review gen pipeline to use a variable we can configure at pipeline level to pass parser version to make it easier to update parser version when a new parser is released.

### DIFF
--- a/eng/pipelines/apiview-review-gen-swagger.yml
+++ b/eng/pipelines/apiview-review-gen-swagger.yml
@@ -27,7 +27,7 @@ jobs:
   - script: >
       dotnet tool install
       Azure.Sdk.Tools.SwaggerApiParser
-      --version 1.0.7-dev.20230227.3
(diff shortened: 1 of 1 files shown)


maririos (approved):

eng/pipelines/apiview-review-gen-swagger.yml
```diff
@@ -27,7 +27,7 @@ jobs:
   - script: >
       dotnet tool install
       Azure.Sdk.Tools.SwaggerApiParser
-      --version 1.0.7-dev.20230227.3
+      --version $(SWAGGER_API_PARSER_VERSION)
```
weshaggard, line 30: As we chatted while I'm ok with this for now we should have a larger conversation at some point about the right place to pull this version from. I would like to have it be shared with all the places that consume it and not duplicated in variables across different pipelines.

Issue #5434, opened 2023-02-13
Move the `test` flag from `get-codeowners.ps1` to the common PS1 test infra, `common-tests`


konrad-jamrozik:
Per this comment:
- https://github.com/Azure/azure-sdk-tools/pull/5103#discussion_r1068680818

Related work:
- https://github.com/Azure/azure-sdk-tools/pull/5040
- https://github.com/Azure/azure-sdk-tools/pull/5095
- https://github.com/Azure/azure-sdk-tools/issues/4882

PR #5608, opened 2023-03-02
Move `get-codeowners.ps1` test mode to `common-tests` + related improvements & cleanup

konrad-jamrozik:
This PR moves the `Test` flag away from `get-codeowners.ps1` to the established `eng/common-tests` infrastructure, using PowerShell Pester testing framework. As such, it solves:

- https://github.com/Azure/azure-sdk-tools/issues/5434

This PR also:
- adds a README for the `eng/common-tests`;
- extracts the `get-codeowners.lib.ps1`, so that the functions can be reused between normal `get-codewners.ps1` usage and through Pester unit test invocations coming from `get-codeowners.tests.ps1`;
- changes some paths for related scripts and test scripts;
- moves some default params to `get-codeowners.lib.ps1` and hides some unused params; this was necessary to avoid default param value duplication between calls from `get-codeowners.ps1` and from `get-codeowners.tests.ps1`;
- applies some refactorings.

Associated changes made manually once this PR got merged:

- deleted internal pipeline [tools - eng-common-tests](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2605155&view=results); instead, created the public equivalent: [tools - eng-common-tests - ci](https://dev.azure.com/azure-sdk/public/_build?definitionId=5985).
  - For context, see [this comment](https://github.com/Azure/azure-sdk-tools/pull/5608#discussion_r1124827938).

Follow-up work:
- https://github.com/Azure/azure-sdk-tools/issues/5878


tools/code-owners-parser/ci.yml
```diff
@@ -23,14 +23,7 @@ pr:
...
-    TestPostSteps:
-      - task: PowerShell@2
-        displayName: Test on code owner script
-        inputs:
-          pwsh: true
-          filePath: eng/common/scripts/get-codeowners.ps1
-          arguments: >
-            -Test
\ No newline at end of file
+    TargetDirectory: eng/common-tests/
```
konrad-jamrozik, line 28: TODO: need to point to `eng/common-tests/codeowners`; but first need to move the .ps1 files to appropriate subdirs (both in common and common-tests)

eng/common-tests/ci.yml
```diff
@@ -8,7 +8,6 @@ trigger:
       - hotfix/*
   paths:
     include:
-      - eng/common/scripts/*
```
weshaggard, line 11: Why are you removing this trigger?

eng/common/pipelines/templates/stages/archetype-sdk-tool-pwsh.yml
```diff
@@ -15,20 +15,20 @@ variables:
   - template: /eng/pipelines/templates/variables/globals.yml
 
 stages:
-  - stage: 'eng_script_tests'
+  - stage: Run_PowerShell_Tests
```
weshaggard, line 18: You can set `displayName: Run PowerShell Tests` and leave the id name empty. That allows for better names without the need for reserved chars like _

eng/common-tests/ci.yml
konrad-jamrozik: Sorry, my bad. For a moment I thought it should be removed because there should be no tests in this dir, but obviously it should be included because the test scripts should kick-off upon any change to this dir.

konrad-jamrozik: @weshaggard IIUC this entire code is currently being ignored anyway due to ADO UI override. I am going to remove the override:

image

eng/common-tests/README.md
```diff
@@ -0,0 +1,25 @@
+# About common-tests
+
+Every PowerShell source that is complex enough to have unit tests should have these tests be written in the [Pester](https://pester.dev/)
+framework and placed in `eng/common-tests/**`, i.e. this directory or one of its descendants.  
+By design, unlike `eng/common`, `eng/common-tests` is not subject to
+[mirroring to repositories](https://github.com/Azure/azure-sdk-tools/blob/main/doc/common/common_engsys.md).
+
+## When tests in this directory are executed
+
+The PowerShell Pester unit tests will be executed by the [`tools - eng-common-tests`](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=220791) pipeline
```
weshaggard, line 10: Does this need to be an internal pipeline? If it doesn't require secrets which I don't believe it does we should move it to public. That way we can have it automatically trigger for any PR.

eng/common/scripts/get-codeowners/get-codeowners.ps1
```diff
@@ -0,0 +1,84 @@
+<#
```
weshaggard, line 1: I assume you moved this to try to align the directory structure with the tests. So far we haven't been splitting the scripts up unless there is a group of them. So if there isn't a strong reason I would leave it in the root of the scripts directory.

weshaggard: Oh I see you also added the helper functions script, for some reason I thought you put that under Helpers. I would suggest you rename that to codeowners-helpers.ps1 and put that under Helpers and then move the get-codeowners.ps1 back to the root of the scripts folder.

konrad-jamrozik: @weshaggard I was following the job-matrix convention:

image

I also have a strong dislike for the name `helper`, as a concept that is too vague and hence lacking cohesion. It is akin to creating a C# class named `Methods` that has all the methods used anywhere in the code. Or a class called `Constants` that has all the constants.

Now the `get-codeowners` dir contains two files that form a part of cohesive whole: the entry point of `get-codeowners.ps1` and the internal functions it leverages, `get-codeowners-functions.ps1`. The fact `get-codeowners.ps1` is is its own dir it also means I can point to that dir from the `codeowners-parser` `ci.yml` and run only tests relevant to it, instead of all tests.

weshaggard: I'm not hung up on the word helpers but the goal is to eventually have the entry point scripts live in the root and have some set of helper functions be separated from the entry point scripts so that they can be like a set of shared code that could be imported by any of the scripts. I would like to make it clear which set of scripts you call and which are used by other scripts. 

If we want to use a "-functions" convention I'd be OK with that as well, but I'm not sure it is worth having a subdirectory for every entry-point script. For larger areas I think it could be worth it. 

@benbp I'd be interested in your thoughts here as well.

konrad-jamrozik: I see your point on "scripts dir has all the scripts that are valid entry points" but I want to clarify that `get-codeowners-functions.ps1` is not be used by anything else except these two use cases:
- to be called by `get-codeowners.ps1`
- to be called by `common-tests/get-codeowners/get-codeowenrs.tests.ps1` Pester unit tests 

Basically the `get-codeowners.ps1` functions is equivalent to C# `internal` class methods that have the `[InternalsVisibleTo]` attribute to enable the unit test project (here: Pester) to call them directly to simplify the UT logic, as calling `get-codeowners.ps1` directly from the Pester unit tests would be cumbersome.

konrad-jamrozik: This is the pester usage:

``` powershell
. $PSScriptRoot/../../common/scripts/get-codeowners/get-codeowners-functions.ps1

function TestGetCodeowners([string]$targetPath, [string]$codeownersFileLocation, [bool]$includeNonUserAliases = $false, [string[]]$expectedOwners) {
  Write-Host "Test: Owners for path '$targetPath' in CODEOWNERS file at path '$codeownersFileLocation' should be '$expectedOwners'"
  
  $actualOwners = Get-Codeowners `
    -ToolVersion $ToolVersion `
    -ToolPath $ToolPath `
    -DevOpsFeed $DevOpsFeed `
    -VsoVariable $VsoVariable `
    -targetPath $targetPath `
    -codeownersFileLocation $codeownersFileLocation `
    -includeNonUserAliases $IncludeNonUserAliases

  $actualOwners.Count | Should -Be $expectedOwners.Count
  for ($i = 0; $i -lt $expectedOwners.Count; $i++) {
    $expectedOwners[$i] | Should -Be $actualOwners[$i]
  }
}
```

and then:
``` powershell
      $resolvedCodeownersPath = (Resolve-Path $codeownersPath)
      TestGetCodeowners `
        -targetPath $targetPath `
        -codeownersFileLocation $resolvedCodeownersPath `
        -includeNonUserAliases $true `
        -expectedOwners $expectedOwners
```
`

eng/common-tests/README.md
konrad-jamrozik: I created a public equivalent:
https://dev.azure.com/azure-sdk/public/_build?definitionId=5985&_a=summary

Once this PR is merged, I will delete the internal version:
https://dev.azure.com/azure-sdk/internal/_build?definitionId=5809

and also the internal `tools eng-script-tests`, which appears redundant to me:

https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=5823

eng/common/scripts/get-codeowners/get-codeowners.ps1
weshaggard: Does the script need to be split up to be tested in pester? As long as the script is split into functions you should be able to `.` source the entry point script as well.

eng/common-tests/README.md
weshaggard: Yes we can combine them if they are doing similar things. 

For the public pipeline we generally add the "- ci" postfix. So `tools - eng-common-tests - ci` or `tools - script-tests - ci` if we make it more generic.

eng/common/scripts/get-codeowners/get-codeowners.ps1
konrad-jamrozik: @weshaggard from what I see dot-sourcing is executing the top-level logic [1]. To my understanding we want for the scripts in `eng/scripts` dir be executable by just invoking them (instead of dot-sourcing and then calling a function) hence they need to have top-level logic, hence we cannot follow the "Dot-source in BeforeAll" pattern in Pester tests, unless we are dot-sourcing a ps1 file that has no top-level logic, like the `-functions.ps1` file I introduced.

[1]

Consider `test.ps1`:

``` powershell
function Foo()
{
  Write-Host "Hello from Foo!"
}

Write-Host "Hello from root!"
```

And following session snippets:
``` powershell
. ./test.ps1
Hello from root!
```

konrad-jamrozik: @weshaggard I guess I could scrap dot-sourcing in `BeforeAll` and invoke the `get-codeowners.ps1` directly from the test case. It is effectively a pass-through delegate to the dot-sourced `Get-Codeowners` function anyway (see snippet above). This way I could avoid introducing the `-functions.ps1` file. But I have a suspicion this won't always be the case, i.e. sometimes we really would prefer skipping the top-level logic and testing one of the functions directly. If so, then we would end up with inconsistent approach in our tests: sometimes dot-sourcing in BeforeAll (to skip the top-level logic), sometimes calling directly. I feel we should avoid such inconsistency.

I am also going to ask @benbp what is his take on this :)

eng/common-tests/README.md
konrad-jamrozik: Renamed to `tools - eng-common-tests - ci`. The `tools - script-tests` looks to be totally redundant to me, just a subset of `eng-common-tests`, so I will outright delete it.

konrad-jamrozik: OK correction: the `script-tests` was running tests in `eng/scripts` but we have since moved all such tests to `eng/common-tests`

weshaggard: OK I'm supportive of having just the one script testing pipeline.

konrad-jamrozik: @weshaggard ungh, actually, there are tests, in:

`eng\scripts\contentreplacements\tests\Repo-File-Content-Replacements\Repo-File-Content-Replacements.Tests.ps1`

I was just running pester against `scripts` instead of `eng/scripts`. Should I move everything to `eng/scripts` ?

In any case, I cannot delete that pipeline yet. But I am thinking to take all the tests  in `eng/scripts` (which is just the one file above) and move them to `eng/common-tests` and rename `eng/common-tests` to e.g. `eng/powershell-tests`.

Does that sound reasonable?

If so, I'll make a separate PR for that, and for now hold off with any changes to the `tools - script-tests` pipeline.

eng/common/scripts/get-codeowners/get-codeowners.ps1
benbp: I started introducing the `-functions.ps1` or `-lib.ps1` pattern to function as a sort of internal class as @konrad-jamrozik described, so that we could easily call things from test, dot source functions locally for development and bug repros in the shell, and nudge towards better code organization (many of our scripts which haven't done this have grown to hundreds of lines of untestable top-level spaghetti over the years). I also agree that dot sourcing and invoking functions directly is a better pattern for unit tests.

To me this is distinctly different from `-Helpers.ps1` files for which inner functions are intended to be shared more broadly, and the script itself isn't a cohesive whole but rather a grab bag of related pieces of functionality.

There is a way to avoid the whole dot source/direct invocation difficulties via [this directive](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/stress-testing/generate-scenario-matrix.ps1#L90):

```
if ($MyInvocation.InvocationName -ne ".") {
    # Call entrypoint function
}
```

However we ran into annoyances in some devops contexts, for example if you are calling your script within the azure cli task to make use of login credentials then it actually ends up dot sourcing the script, so nothing is invoked. Due to this and the code organization considerations I listed above, I have tried to use the lib/functions file and invocation file convention instead.

I agree with @weshaggard that if we just have just two files it's probably not worth its own directory.

benbp: Re test convention around dot sourcing and calling directly, I think this boils down to two fundamental approaches to powershell development:

1. Use files as the de-facto function boundary, defining one cmdlet per file (with optional micro-helper functions inside). I would say this is a little bit more in line with the powershell language, in that it's easier to consume and define behavior for built-ins like global/script scopes, common variables like `$ErrorActionPreference` and common cmdlet parameters like `-WhatIf`.
2. Use functions as the function boundary. It can be a little more annoying to keep various globally scoped settings and values from colliding, but I think it creates a better development experience (not having to traverse dozens of files, not having to maintain dot source hierarchies).

Like I said in my above comment, I try to trend towards approach 2, because with approach 1 it's hard to keep people from adding too much code over time to what are intended to be small files. Whereas if one is editing a file where all code is in functions, naturally they are going to add their own code into a function.

Back to the testing perspective, I think it's ok to have tests that run the whole file, and tests that run functions from within the file. However my ideal pattern is that the lib/functions file will have a main/entrypoint function defined, and if the test wants to call the whole flow, it dot sources the file and calls main, rather than invoking the script file. This gives us some flexibility with how we allow users to invoke scripts without the unit tests having to follow the same format.

konrad-jamrozik: @benbp @weshaggard OK so I think we mostly agree that `-functions.ps1` is a good idea.

Re this:

> I agree with @weshaggard that if we just have just two files it's probably not worth its own directory.

I want to point out that [job-matrix](https://github.com/Azure/azure-sdk-tools/tree/main/eng/common/scripts/job-matrix) has its own directory. Plus if we move `-functions.ps1` to `/eng/common/scripts/`, then we will break the rule that "all top-level `.ps1` files are runnable". Hence I would still keep `/eng/common/scripts/get-codeowners/`.

weshaggard: I'm good with most of is being said here but we had a little further discussion in the teams channel. Mostly whatever we decide we should document in our powershell.md file and start to move towards that convention.

konrad-jamrozik: @weshaggard OK! How about I merge this PR, then make a PR updating the conventions doc, and also updating the layout to match the conventions? I want to get this PR in as it also has some other changes besides file layout, including changes to `eng/common`, so I don't want to make it linger.

weshaggard: Sounds good.

eng/common-tests/get-codeowners/get-codeowners.tests.ps1
```diff
@@ -0,0 +1,50 @@
+Import-Module Pester
+
+BeforeAll {
+  . $PSScriptRoot/../../common/scripts/get-codeowners/get-codeowners-functions.ps1
+
+    $ToolVersion = "1.0.0-dev.20230214.3"
```
weshaggard, line 6: Do we want to duplicate the version number here?

tools/code-owners-parser/ci.yml
```diff
@@ -20,17 +20,10 @@ pr:
   paths:
     include:
       - tools/code-owners-parser
-      - eng/common/scripts/get-codeowners.ps1
+      - eng/common/scripts/get-codeowners
 
 extends:
-  template: /eng/pipelines/templates/stages/archetype-sdk-tool-dotnet.yml
+  template: /eng/common/pipelines/templates/stages/archetype-sdk-tool-pwsh.yml
```
weshaggard, line 26: Why did your remove the building and production of the .NET tool with this change?

weshaggard (approved):
Looks good once you fix the code-owner-parser\ci.yml to add back the building of the .NET project.

tools/code-owners-parser/ci.yml
konrad-jamrozik: My bad, thanks for catching this!

eng/common-tests/get-codeowners/get-codeowners.tests.ps1
konrad-jamrozik: Deduped!

eng/common/scripts/get-codeowners/get-codeowners.ps1
```diff
@@ -0,0 +1,16 @@
...
+  [string]$TargetPath = "",
+  [string]$TargetDirectory = ""
+)
+
+. $PSScriptRoot/get-codeowners-functions.ps1
+
+return Get-Codeowners `
+  -TargetPath $TargetPath `
+  -TargetDirectory $TargetDirectory
+# kja need to restore param: https://github.com/Azure/azure-sdk-for-cpp/blob/2850c5d32c8a86491b49e801433b8f186fa81745/eng/pipelines/templates/stages/archetype-cpp-release.yml#L196
```
konrad-jamrozik, line 15: TODO: fix these before merging

eng/common-tests/README.md
konrad-jamrozik: Two things.

### 1. PowerShell test pipeline setup

I had a chat with @weshaggard. We should keep multiple pipelines
- one pipeline for running PowerShell tests in `/eng/common-tests/`;
- one pipeline per each language repo for running PowerShell tests in its `/eng/scripts/` dir.

This is because:
- There are two separate sets of PowerShell scripts: `/eng/common/scripts/` and `/eng/scripts/` and both these sets need to have tests written for them.
- The dir `/eng/scripts/` in each language repo has language-repo-specific scripts; currently they hardly have any tests, but in the future they should have more tests.

The pipelines can be sourced from different `ci.yml` files - the `ci.yml` source would differ only by parameters passed to the underlying abstraction for running Pester tests. Most importantly, the `TargetDirectory` would point to appropriate dir.

### 2. PowerShell file naming conventions

Per chat in this PR and over Teams between me, @weshaggard and @benbp we want to adopt following convention for naming files:

- `/eng/scripts/foo.ps1` and `/eng/common/scripts/foo.ps1` for directly-invocable entry points;
- `/eng/scripts/foo.lib.ps1` and `/eng/common/scripts/foo.lib.ps1` for dot-sourceable functions; the `foo.ps1` is expected to be in practice just a delegate calling the "entry point" function from `foo.lib.ps1`;
- `/eng/scripts/foo.tests.ps1` and `/eng/common-tests/foo.tests.ps1` for Pester tests for `foo.lib.ps1`.

konrad-jamrozik:
Due to release week not going to merge this PR until 3/15/2023. This PR poses risk of breaking docs pipeline.

eng/common-tests/README.md
konrad-jamrozik: Created an issue to track the work above:
- https://github.com/Azure/azure-sdk-tools/issues/5878

PR #5613, opened 2023-03-03
True Version Override in Proxy Tooling

scbedd:
In the event that I actually break folks with a proxy version upgrade in `target_version.txt`, I want to be certain that our templates have true override that we can easily set when calling our template.

Even if they set it so that it doesn't start the proxy, they can invoke the `test-proxy-tool` yaml with `runProxy` set to false, so it still overrides the version but doesn't actually run it for them. Then if their tooling looks at `target_version.txt` that will be properly overwritten.

Before this change, _our_ tooling would boot up the proper proxy override version, but not update the actual file at rest in the repo. Any dependent tooling would therefor be looking for the wrong version of the proxy.

This PR is in reaction to a situation that came up with @billwert and @samvaity . They don't honor the `target_version.txt` yet, but if they _did_, before this PR an override wouldn't have worked for them. I'd like to avoid that when we eventually really need this to work.

`eng/common` changes won't be created until after release week, no reason to rock the boat.

### DIFF
--- a/eng/common/testproxy/override-proxy-version.ps1
+++ b/eng/common/testproxy/override-proxy-version.ps1
@@ -1,0 +1,25 @@
+<#
+.SYNOPSIS
+Replaces target test-proxy version present in target_version.txt.
+
+.PARAMETER TargetVersion
+The replacement version. Used in its entirety, so don't exclude parts of the version definition.
+#>
+[cmdletbinding(SupportsShouldProcess=$True)]
+param(
+   [Parameter(mandatory=$true)] [string] $TargetVersion
+)
+
+$versionFile = Join-Path $PSScriptRoot "target_version.txt"
+$existingVersionText = Get-Content -Raw -Path $versionFile
+$existingVersion = $existingVersionText.Trim()
+
+if ($PSCmdLet.ShouldProcess($versionFile)){
+   Write-Host "Replacing version `"$existingVersion`" with version `"$TargetVersion`" in $versionFile."
+   Set-Content -Path $versionFile -Value "$TargetVersion`n"   
+}
+else {
+   Write-Host "Would replace version `"$existingVersion`" with version `"$TargetVersion`" in $versionFile."
+}
+
+
--- a/eng/common/testproxy/test-proxy-docker.yml
+++ b/eng/common/testproxy/test-proxy-docker.yml
@@ -10,6 +10,15 @@ steps:
     displayName: 'Language Specific Certificate Trust'
     condition: and(succeeded(), ${{ parameters.condition }})
 
+  - task: PowerShell@2
+    displayName: 'Override proxy version if necessary'
+    condition: and(succeeded(), ${{ parameters.condition }}, ne('${{ parameters.targetVersion }}', ''))
+    inputs:
+      targetType: filePath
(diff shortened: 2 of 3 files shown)


konrad-jamrozik:
@scbedd just few n00b questions for my learning.

Re:

> Even if they set it so that it doesn't start the proxy, they can invoke the test-proxy-tool yaml with runProxy set to false, so it still overrides the version but doesn't actually run it for them. Then if their tooling looks at target_version.txt that will be properly overwritten.

Why would they (their tooling) care about test proxy version if they are not actually running it?

> Before this change, our tooling would boot up the proper proxy override version, but not update the actual file at rest in the repo. Any dependent tooling would therefor be looking for the wrong version of the proxy.

I am not familiar with the flow here. Is the idea here something like that? Our tooling is using some test proxy override to determine which version to install on the agent. But then the user/consumer repo is looking at its `target_version.txt` and sees the different non-overridden version, and hence is unable to find the test proxy. So a bit like installed .NET runtime mismatch ?

> This PR is in reaction to a situation that came up with @billwert and @samvaity . They don't honor the target_version.txt yet, but if they did, before this PR an override wouldn't have worked for them. I'd like to avoid that when we eventually really need this to work.

So I guess for now they decide themselves in some other way what test proxy version to use? I guess they hardcoded it? And I guess for now the fix was to ask them to point to different test proxy version, the one we actually install?

scbedd:
> Why would they (their tooling) care about test proxy version if they are not actually running it?

If their tooling is doing things right, they should be honoring the version targeted in `eng/common/testproxy/target_version.txt`.

> I am not familiar with the flow here. Is the idea here something like that? Our tooling is using some test proxy override to determine which version to install on the agent. But then the user/consumer repo is looking at its target_version.txt and sees the different non-overridden version, and hence is unable to find the test proxy. So a bit like installed .NET runtime mismatch ?

The test-proxy is commonly versioned, I roll a new version out when I can get one tested and new features are being waited for. Let's say I roll out a version of the proxy that _actually breaks_ a couple of unit tests for, say, `javascript`. It's release week. They don't want to re-record tests or otherwise investigate why the new version of the proxy broke them. 

Instead, they go to where [the test-proxy is installed](https://github.com/Azure/azure-sdk-for-js/blob/0d7c05401f7ae6c901a78e3f8f9b679271d49bdf/eng/pipelines/templates/steps/test.yml#L34), and set `targetVersion`: `<whatever the old version their code worked with>`. Now either case of A) running the version of the proxy that is already installed or B) looking at `target_version.txt` to download a standalone version, everything will align to the version they expect. Tests pass. Release is unblocked.

This is a very unlikely scenario, as I am _extremely_ paranoid about how I roll out proxy updates, but I'd like our break glass to work.

> So I guess for now they decide themselves in some other way what test proxy version to use? I guess they hardcoded it? And I guess for now the fix was to ask them to point to different test proxy version, the one we actually install?

Most do **not** hardcode it, they trust me to roll out a new common proxy version as new features are added. I ensure that all existing builds (I trigger a couple important on each repo when eng/common/proxy/testproxy folder has updated) are green. However, I'm not perfect, and I'd like folks to be able to force reversion to older proxy version in cases of duress.

konrad-jamrozik (approved):

PR #5615, opened 2023-03-03
Release `typespec-apiview` 

tjprescott:
Renames `cadl-apiview` to `typespec-apiview`.


konrad-jamrozik (approved):

tjprescott:
@praveenkuttappan @chidozieononiwu any ideas why this is not running the CI pipeline? The tests all pass when run locally. :thinking:

tjprescott:
That is correct. This is only compatible with `@typespec/compiler 0.41`

Issue #5056, opened 2023-01-05
Refactor repo root dir layout


konrad-jamrozik:
Per following @weshaggard's comments:

- https://github.com/Azure/azure-sdk-tools/pull/5044#discussion_r1061908902
- https://github.com/Azure/azure-sdk-tools/pull/5044#discussion_r1061909338
- https://github.com/Azure/azure-sdk-tools/pull/5044#discussion_r1061909589
- https://github.com/Azure/azure-sdk-tools/pull/5044#discussion_r1061909921

we should restructure our repo directory layout. Namely, we should:

- Move `/.azure-pipelines/` under `/eng/pipelines/`
- Move `/config/` under `/eng/`
- Move `/scripts/` under `/eng/`
- Move `/src/` and `/web/` under appropriate subdirs of `/tools/`

konrad-jamrozik:
@weshaggard I have few question about files under `/.azure-pipelines/`. There are pipelines defined over the `.yml` files in that dir:

| .yml | pipeline |
|-|-|
`.azure-pipelines/apiview-sync-staging.yml` | [APIView Sync Production to Staging](https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=2002)
`.azure-pipelines/apiview.yml` | [APIView - Deploy - Staging](https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=1960)
`.azure-pipelines/apiview.yml` | [APIView](https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=1136)
`.azure-pipelines/warden.yml` | [azure-sdk-tools - warden](https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=106)

My questions:

- is there an easy, reliable way for me to ensure the list above is exhaustive? I am fairly sure I should be able to find relevant ADO API call to return me json with all the pipelines and so search for given `.yml`, (I mean, our `notification-configuration` does something similar) but I was just wondering if we already have some automation for that.
[...]

PR #1359, opened 2021-02-01
Pool provisioning config

mitchdenny:
This PR contains a copy of the pre-provisioning setup for each 1ES pool that we operate. This PR represents a proposal for some initial optimizations around our pool sizes. Basically I'm saying that after 9pm each day we scale the pool down to zero pre-allocated and then increase the pre-allocated in the morning up to its current level.

The long term goal for putting this information under version control is that we can make changes and review them, and then automatically apply the configuration to our pools.

NOTE: The times are in UTC.

PR #5620, opened 2023-03-04
Delete `/config/`

konrad-jamrozik:
Update 3/6/2023: deleting the files, per [Mike's comment](https://github.com/Azure/azure-sdk-tools/pull/5620#pullrequestreview-1325135455)

-----

Contributes to:
- https://github.com/Azure/azure-sdk-tools/issues/5056

I believe the moved files are just a copy not being depended on by anything, per this PR which introduced them:
- https://github.com/Azure/azure-sdk-tools/pull/1359

@weshaggard @mikeharder given these files appear to be untouched and unused since 3 years, should they be deleted?


mikeharder (changes requested):
These files should be deleted.  Our pools now use "auto pre-provisioning" instead of a manual config like these files.

PR #5634, opened 2023-03-06
Move `/scripts/python/` to `/eng/python/`

konrad-jamrozik:
Contributes to:

- #5056


eng/python/docs-automation/README.md
```diff
@@ -17,7 +17,7 @@ This is intended to be run _after_ a docs.MS CI run. For more details on the how
 This is intended to be run against an [intermediate doc repo.](https://github.com/MicrosoftDocs/azure-docs-sdk-node)
 
 ```
-  python .\scripts\python\docs-automation\sync_tocs.py 
+  python .\eng\python\docs-automation\sync_tocs.py
```
benbp, line 20:
```suggestion
  python ./eng/python/docs-automation/sync_tocs.py 
```

while you're in there

benbp (approved):

weshaggard:
Why did we put these under /eng/python? As opposed to /eng/scripts or /eng/scripts/python?

konrad-jamrozik:
@weshaggard I mean, python programs are scripts "by design" - i.e. they are interpreted, not compiled, and so can be run directly. So I would expect that if somebody sees `python` dir they will expect for it to have executable python programs/scripts. But maybe I am wrong. @scbedd what do you think? Would it be more "pythonic" to put it in `/eng/scripts/python/` ?

konrad-jamrozik:
> It isn't really about correctness but about convention and whether or not we want to mix script types in the same folder. I was hoping to keep all the scripts no matter the type under the eng/scripts folder.

@weshaggard addressed by:
- https://github.com/Azure/azure-sdk-tools/pull/5877

Issue #5281, opened 2023-01-31
Deduplicate reusable code into shared library: logging, file utils, etc.


konrad-jamrozik:
A lot of our tools implement one-off abstractions, like logging or file system utilities. This leads to duplication and lack of consistency.

This work item is about building a case for deduplicating all such code into one reusable library per language, with initial focus on .NET. The benefits include less code to maintain and established conventions to follow.

# .NET deduplication candidates

Below follows a list of candidate code to be deduplicated; if you find more, please add a comment in this issue, and I will incorporate it in this list!

TODO

weshaggard:
FYI @benbp

benbp:
On the logging topic I've had a good experience with [Serilog](https://github.com/serilog/serilog). It makes it easy to [configure output formats](https://github.com/Azure/azure-sdk-tools/blob/main/tools/stress-cluster/services/Stress.Watcher/src/PodEventHandler.cs#L59-L67) and create [nested property contexts](https://github.com/Azure/azure-sdk-tools/blob/main/tools/stress-cluster/services/Stress.Watcher/src/PodEventHandler.cs#L126-L127).

benbp:
Also to add to this list: we will benefit from standardizing our argument parsing library, whether that's CommandLineParser or the new System.CommandLine.

PR #5636, opened 2023-03-06
Make `CodeownersEntry.GetHashCode` support Enumerables

konrad-jamrozik:
As in subject.

Candidate to be added to:
- https://github.com/Azure/azure-sdk-tools/issues/5281

Note I didn't change the type of `CodeownersEntry` to `record` as it would not help, per this doc:
- https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/records

> Two variables of a class type are equal if the objects referred to are the same class type and the variables refer to the same object.


weshaggard (approved):
looks reasonable.

PR #5653, opened 2023-03-07
Python APIView Fixes

### LINKED ISSUE #3759: [Python APIView] make sure we support best practices for accessing annotations
See here: https://docs.python.org/3.10/howto/annotations.html

There is still a debate whether https://peps.python.org/pep-0563/ will become the default behavior in Python or if https://peps.python.org/pep-0649/ will supersede it. If we follow the best practices to access annotations, I believe we should be okay in either case.

### LINKED ISSUE #4860: [Python APIView] display decorators on models
From @annatisch 
>  it seems that APIview currently only displays decorators for methods, is it possible to add support for displaying decorators on models as well to prevent this being missed in future?

### LINKED ISSUE #5197: [Python APIview] ivar annotations per PEP526 showing up as cvars
Context: https://apiview.dev/Assemblies/Review/f654d0b58676400a9fe225b5df013d73?diffRevisionId=c550c432e8eb485eac666b722bfcdcd4&diffOnly=True&revisionId=c5256ddacaff4b25b6f6eae95676fca5&doc=False#24bfe5a892a5434bb01aeb2d039909bd

PEP526: https://peps.python.org/pep-0526/#class-and-instance-variable-annotations

```python
class BasicStarship:
    captain: str = 'Picard'                  # instance variable with default
    damage: int                                 # instance variable without default
    stats: ClassVar[Dict[str, int]] = {}  # class variable
```

`captain` and `damage` should be `ivar` in APIView, while `stats` should be `cvar` since it is explicitly marked.

tjprescott:
Fixes #5197. Fixes #4860. Fixes #3759.


kristapratico (approved):
🚀

PR #5654, opened 2023-03-08
Update sparse checkout to target source branch instead of source version

benbp:


eng/common/pipelines/templates/steps/sparse-checkout.yml
```diff
@@ -6,7 +6,7 @@ parameters:
     type: object
     default:
       - Name: $(Build.Repository.Name)
-        Commitish: $(Build.SourceVersion)
+        Commitish: $(Build.SourceBranch)
```
weshaggard, line 9: Does SourceBranch work? What does it get set to in the various cases? i.e. CI/Scheduled vs PR and PR from fork somethings is different as well.

benbp: It doesn't work as I had expected. We end up getting a branch like `refs/heads/sync-eng/common-benbp/sparse-branch-5654-ForTestPipeline` which has the same issue as the commit id: it doesn't test on the merge commit with main. I'm trying to figure out a simple way to get the pull request id so I can calculate the merge ref and use `ls-remote` to get the correct sha.

benbp:
active

benbp:
Active

Issue #5673, opened 2023-03-09
Requests failing with `Reading the request body timed out due to data arriving too slowly. See MinRequestBodyDataRate.`


scbedd:
`azure-keyvault-administration` [tests](https://dev.azure.com/azure-sdk/public/_build/results?buildId=2610713&view=ms.vss-test-web.build-test-results-tab&runId=39717417&resultId=100034&paneView=debug) are failing unexpectedly with some very odd behavior. 

When [the failing tests](https://dev.azure.com/azure-sdk/public/_build/results?buildId=2610713&view=ms.vss-test-web.build-test-results-tab) are run individually, this behavior does not repro. Otherwise, it only repros when all tests are run for the package. (like in CI)

I personally think that there is something going a bit wonky with the keyvault-administration client. Caching or the like? Sending wrong content-length? But we need to dig in and find out.

In the meantime going to dash out a PR that adjust `MinRequestBodyDataRate`.

Stacktrace:
```
at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1ContentLengthMessageBody.ReadAsyncInternal(CancellationToken cancellationToken)
at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder\u00601.StateMachineBox\u00601.System.Threading.Tasks.Sources.IValueTaskSource\u003CTResult\u003E.GetResult(Int16 token)
at System.IO.Pipelines.PipeReader.CopyToAsyncCore[TStream](TStream destination, Func\u00604 writeAsync, CancellationToken cancellationToken)
at Azure.Sdk.Tools.TestProxy.RecordingHandler.ReadAllBytes(Stream stream) in /mnt/vss/_work/1/s/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs:line 1081
at Azure.Sdk.Tools.TestProxy.RecordingHandler.CreateEntryAsync(HttpRequest request) in /mnt/vss/_work/1/s/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs:line 547
at Azure.Sdk.Tools.TestProxy.RecordingHandler.HandlePlaybackRequest(String recordingId, HttpRequest incomingRequest, HttpResponse outgoingResponse) in /mnt/vss/_work/1/s/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs:line 438
[...]

PR #5675, opened 2023-03-09
Lower MinRequestBodyDataRate

scbedd:
This is contributing to #5673


mikeharder (approved):
Worth a try

scbedd:
This turned out to be un-necessary. The failure was because the request the python tests were sending had a Content-Length, but never actually sent a body. 

Reason was that the test was caching an empty credential body erroneously. [Ensuring the cache was properly](https://github.com/Azure/azure-sdk-for-python/pull/29124/commits/2ebe1a8c6480637a03332543ff78eabd5093e4a3) cleared means that the body would be sent with actual value, which resolved the problem.

PR #5676, opened 2023-03-09
Select childmost deployment template only when deploying resources

benbp:
I am not aware of any place in our repos that is actually utilizing the "search for all templates" behavior. So I propose we change the new-testresources.ps1 behavior to stop searching for templates after we find the first one in a child directory. If people want to have >1 templates, then they should use bicep instead and reference the parent template(s) via module in their child template. If bicep had been around when this was all originally written, I think this may have been the right decision anyway.

The hierarchy is:

Find most nested template from within the service directory:
- service directory `foo/bar/` -> search in `foo/bar/` for a template, then search in `foo/` for a template
- If both arm and bicep templates are found in the directory, only pick the bicep one (this also should have been the existing behavior)

This came about because a service team is trying to use a common template for N-1 sub-directories under their service umbrella, and a separate template for the remaining sub-directory, and it's impossible to achieve this without duplicating the common template and changing a bunch of pipeline yaml to update references (since ServiceDirectory is overloaded in a few ways).

### DIFF
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -400,11 +400,9 @@ try {
         Get-ChildItem -Path $root -Filter "$_" -Recurse | ForEach-Object {
             Write-Verbose "Found template '$($_.FullName)'"
             if ($_.Extension -eq '.bicep') {
-                $templateFile = @{originalFilePath = $_.FullName; jsonFilePath = (BuildBicepFile $_)}
-                $templateFiles += $templateFile
+                $templateFiles += @{originalFilePath = $_.FullName; jsonFilePath = (BuildBicepFile $_)}
             } else {
-                $templateFile = @{originalFilePath = $_.FullName; jsonFilePath = $_.FullName}
-                $templateFiles += $templateFile
+                $templateFiles += @{originalFilePath = $_.FullName; jsonFilePath = $_.FullName}
             }
         }
     }
@@ -413,6 +411,19 @@ try {
         Write-Warning -Message "No template files found under '$root'"
         exit
     }
+
+    $templateFile = ""
+    foreach ($file in $templateFiles) {
+        $templateFileDepth = ($templateFile.OriginalFilePath -split "[\\/]").Length
+        $fileDepth = ($file.OriginalFilePath -split "[\\/]").Length
+        if ($templateFileDepth -lt $fileDepth) {
+            $templateFile = $file
+        } elseif ($templateFileDepth -eq $fileDepth) {
+            throw ("Found multiple deploy template matches '$($templateFile.OriginalFilePath)' and '$($file.OriginalFilePath)'. " + `
+                  "Remove all but one file or pass a more specific -ServiceDirectory parameter.")
+        }
+    }
+
 
     $serviceName = GetServiceLeafDirectoryName $ServiceDirectory
     $BaseName, $ResourceGroupName = GetBaseAndResourceGroupNames `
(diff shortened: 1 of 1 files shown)


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -735,76 +746,73 @@ try {
         'PowerShell', '${{env:{0}}} = ''{1}'''
     }
 
-    # Deploy the templates
-    foreach ($templateFile in $templateFiles) {
```
benbp, line 739: All the changes from here down are just me removing the `foreach` loop and removing one level of indentation. The diff doesn't make this clear.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -414,6 +412,19 @@ try {
...
+        $fileDepth = ($file.OriginalFilePath -split "[\/]").Length
+        if ($templateFileDepth -lt $fileDepth) {
+            $templateFile = $file
+        } elseif ($templateFileDepth -eq $fileDepth) {
+            # Favor .bicep files over ARM .json files
+            $templateFile = $templateFile.OriginalFilePath.Length -gt $file.OriginalFilePath.Length ? $templateFile : $file
+        }
+    }
+
+
```
benbp, line 427: This is the actual meat of the functionality proposed.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -414,6 +412,19 @@ try {
         exit
     }
 
+    $templateFile = ""
+    foreach ($file in $templateFiles) {
+        $templateFileDepth = ($templateFile.OriginalFilePath -split "[\\/]").Length
+        $fileDepth = ($file.OriginalFilePath -split "[\/]").Length
```
mikeharder, line 418: Should the arguments here be the same, or is it intentional one is `[\\/]` and the other is `[\/]`?

mikeharder (approved):

eng/common/TestResources/New-TestResources.ps1
benbp: Good catch.

eng/common/TestResources/New-TestResources.ps1
heaths: Tip: it does if you hide whitespace in the PR options. 😉

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -414,6 +412,19 @@ try {
         exit
     }
 
+    $templateFile = ""
```
heaths, line 415: Instead of this, why don't we just use an ancestor search like [this](https://github.com/heaths/profile/blob/0ac9475b577218076bc3e792564edf553f23b02a/Modules/MyPrompt/MyPrompt.psm1#L91)? Seems much less cumbersome than splitting on directory paths. I mean, in this case, isn't the `$root` above potentially a package (to support _sdk/{serviceDirectory}/{package}_ so searching up would be more accurate than searching down to find one that matches?

eng/common/TestResources/New-TestResources.ps1
benbp: Mind blown

eng/common/TestResources/New-TestResources.ps1
benbp: @heaths It's a fair question. I was in the mindset of adapting to the outputs from the existing code but should have considered this approach.

A challenge is that we don't know the sub directory, just the root, so we're searching from the parent anyways. I could write a depth first traversal that just returns on the first template file. I guess a downside of either implementation is in the following layout we would not know which template to pick, but I think supporting said layout isn't really a convention we should support anyways.

```
ServiceDirectory: foo

foo/bar/test-resources.json
foo/qux/test-resources.json
```

heaths: In this scenario, going top-down makes sense then, but then how do we know which ones to pick up for a package? You said we still start from the service directory, but how are we picking the right package directory (if any) test-resources.jsons now?

benbp: Updated to throw

heaths (approved):

benbp:
Closing as this still has some gaps that I don't think this change can completely address.

PR #5685, opened 2023-03-10
Test Specific HeaderRegexSanitizer scenario

scbedd:
@samvaity just a sanity check on proxy side. As I expected it's working fine.

Wondering if there is another encoding issue, but I'm moving on to testing your branch directly.


scbedd:
Working as intended. Closing PR!

PR #5690, opened 2023-03-13
Install same emitter only once when generating SDK

pshao25:
Current behavior: we do `npm install` for exactly same packages in all the cadl projects. Same `npm install` would do `n` times.

After this change: Same emitter and its dependencies would install only once at the root folder. Reduce time cost from `O(n)` to `O(1)`.


archerzz (approved):

eng/common/scripts/Cadl-Project-Generate.ps1
```diff
@@ -14,6 +14,78 @@ $ErrorActionPreference = "Stop"
...
+        npm install --no-lock-file
+
+        if (Test-Path "package.json") {
+            Remove-Item -Path "package.json" -Force
+        }
+        if (Test-Path "package-lock.json") {
+            Remove-Item -Path "package-lock.json" -Force
+        }
+
+        if ($LASTEXITCODE) { exit $LASTEXITCODE }
```
benbp, line 82:
```suggestion
        if ($LASTEXITCODE) { Pop-Location; exit $LASTEXITCODE }
```

weshaggard:
@mikeharder is this a case where we should consider npm ci over npm install?

pshao25:
> @mikeharder is this a case where we should consider npm ci over npm install?

We shouldn't have any package-lock.json there. Could you put more details why we should consider npm ci?

mikeharder:
@pshao25: In general, any scenario where we are installing a JavaScript-based tool, we should be installing from a lockfile to ensure we are pinning the versions of all transitive dependencies.  Otherwise, our tool can be broken at any time by a dependency publishing a new minor or patch version with a breaking change.

pshao25:
> @pshao25: In general, any scenario where we are installing a JavaScript-based tool, we should be installing from a lockfile to ensure we are pinning the versions of all transitive dependencies. Otherwise, our tool can be broken at any time by a dependency publishing a new minor or patch version with a breaking change.

In development scenarios, I totally agree. 

But this script is for user scenarios. We published our emitter packages and this script could help customer install that package (download from public registry). In this case, we don't even pack the lockfile into that package. So IMO we should still use `npm install`.

mikeharder:
> > @pshao25: In general, any scenario where we are installing a JavaScript-based tool, we should be installing from a lockfile to ensure we are pinning the versions of all transitive dependencies. Otherwise, our tool can be broken at any time by a dependency publishing a new minor or patch version with a breaking change.
> 
> In development scenarios, I totally agree.
> 
> But this script is for user scenarios. We published our emitter packages and this script could help customer install that package (download from public registry). In this case, we don't even pack the lockfile into that package. So IMO we should still use `npm install`.

I think we should take a deeper look at the customer scenario as well.  It's important to understand the difference between JS **packages** and JS **tools**.

Packages are intended to be used in applications, where the application developer can (and should) use their own lockfile to pin all transitive dependencies.  Packages should float their dependencies to the widest range possible, for maximum interoperability with other packages and to reduce the number of dependency versions that need to be installed.

Tools are intended to be run directly from the command line, and the user will not have their own lockfile.  Tools should do as much as they can to ensure they are not broken by dependency updates, since tools want to guarantee that if users install a particular version of the tool, it will always work exactly the same regardless of what new dependency versions.  This means tools either need to ship their own lockfile or bundle all their dependencies.  The easiest way is to include an `npm-shrinkwrap.json` file in the package itself.

https://docs.npmjs.com/cli/v9/configuring-npm/npm-shrinkwrap-json

pshao25:
@mikeharder Ah I got your point now. I believe our customers use it as a JS tool. @m-nash Is there any possibility that our customers would use emitter as JS package?

There is a quote from the [link](https://docs.npmjs.com/cli/v9/configuring-npm/npm-shrinkwrap-json) that discourage us from using shrinkwrap:

> It's strongly discouraged for library authors to publish this file, since that would prevent end users from having control over transitive dependency updates.

@m-nash there are 2 solutions:
1. We include an npm-shrinkwrap.json file in the package itself as Mike suggested.
2. We could add a `emitter-package-lock.json` right near `emitter-package.json`, so that when we install emitter from sdk repo, we could lock the dependency version.

I think other language repos don't have above files. They might need to do the same thing.

m-nash:
I will defer to @mikeharder expertise here if you think our best approach is for all emitters to include a npm-shrinkwrap.json and @pshao25 you don't see any blocker reasons to not do that lets go with option 1.

mikeharder:
> It's strongly discouraged for library authors to publish this file, since that would prevent end users from having control over transitive dependency updates.

If you want to publish something as both a **library** and a **tool**, you should publish two packages.  For example, the tool package could be named `foo`, and the library package something like `foo-lib`.  The tool package would depend on the library package and include an `npm-shrinkwrap.json`, and the library package would not include `npm-shrinkwrap.json`.

mikeharder:
@pshao25: I think this conversation has diverged from the original purpose of this PR, which I believe is to install the emitter only once.  Do all packages in the repo use the same emitter version, and do we think this will be true going forward?  Do we ever see a world where different packages use different versions of the emitter?  Might this be necessary as the emitter evolves and we may not want to regenerate all packages at once?

If we are sure all packages will use the same emitter, should we just install it globally using `npm install -g`?

lmazuel:
FWIW, I'm in totally in sync with @mikeharder in spirit here, our emitters are applications, not library. While I'm not JS expert, if option 1 is the right eng approach to pin dependencies of our emitters here then let's do that.

lmazuel:
> @pshao25 Pan Shao FTE: I think this conversation has diverged from the original purpose of this PR, which I believe is to install the emitter only once. Do all packages in the repo use the same emitter version, and do we think this will be true going forward? Do we ever see a world where different packages use different versions of the emitter? Might this be necessary as the emitter evolves and we may not want to regenerate all packages at once?
> 
> If we are sure all packages will use the same emitter, should we just install it globally using `npm install -g`?

I think it should be an unique version for the repo. It doesn't mean you're forced to regenerate each time there is a release (it's interesting debate, but not here :)), but if you regenerate, you need to follow the current recommendation hardcoded in the repo. It's been the way SDK Automation has been doing it by the way before even it was close like that (when it was [SwaggerToSdk](https://github.com/Azure/swagger-to-sdk) in a previous life)

pshao25:
@mikeharder I created a new issue to track npm-shrinkwrap.json change. For the PR itself,

> Do all packages in the repo use the same emitter version, and do we think this will be true going forward?

Yes.

> Do we ever see a world where different packages use different versions of the emitter?

No.

> Might this be necessary as the emitter evolves and we may not want to regenerate all packages at once?

No, AFAIK.

> If we are sure all packages will use the same emitter, should we just install it globally using npm install -g?

My concern is if we install emitter globally, the dependecies (typespec compiler and others) would also be updated when we run this script in our dev machine. I don't think it is a good practice, because developers might want to install specific version of typespec compiler. I really don't want to contaminate the global place, and really don't want the place where I'm fetching the emitter to be contaminated (I mean it is possible we could install globally at any other place, right?)

mikeharder:
@pshao25

> My concern is if we install emitter globally, the dependecies (typespec compiler and others) would also be updated when we run this script in our dev machine.

Is this script intended for use both in pipelines and on dev machines?  If so, I agree this adds a lot of complications.  Installing things globally in pipelines is usually fine, as long as it works for the duration of the pipeline.  But we do want to be more careful on dev machines.

I think it's fairly unusual for scripts under `tools/eng/common` to be used in both pipelines and dev machines.

@weshaggard: Are you aware of any other scripts under `tools/eng/common` used on dev machines?

weshaggard:
>@weshaggard: Are you aware of any other scripts under tools/eng/common used on dev machines?

Yes there are other scripts under eng/common (e.g. prepare-release.ps1) and if we need these scripts to be in all the repos I think it should be under eng/common for either local or pipelines.

weshaggard:
For my earlier question about npm ci I agree while we don't have a lock file we cannot use that and I agree we should try to use shrinkwrap instead to help get these tools installed.

eng/common/scripts/Cadl-Project-Generate.ps1
pshao25: `Pop-Location` is alreay in the `finally` block.

pshao25:
> Is this script intended for use both in pipelines and on dev machines?

Yes.

@mikeharder @weshaggard @benbp @lmazuel Thank you all for the suggestion. I will track the npm-shrinkwrap.json in another issue and notify other languages.

If there is no further concern, I would have this merged after "cadl rename" finished.

m-nash:
Definitely should not globally install it would mess with all other development / testing constantly.

eng/common/scripts/Cadl-Project-Generate.ps1
```diff
@@ -14,6 +14,78 @@ $ErrorActionPreference = "Stop"
 . $PSScriptRoot/common.ps1
 Install-ModuleIfNotInstalled "powershell-yaml" "0.4.1" | Import-Module
 
+function GetNpmPackageVersion([string]$packagePath) {
```
m-nash, line 17: this should be moved to the new `Typespec-Project-Generate.ps1` as this file will be deleted after all languages swap to the new one.

pshao25: Yes, will move this change to `Typespec-Project-Generate.ps1` after "cadl rename" finished.

chunyu3: > Typespec-Project-Generate.ps1

It should be `TypeSpec-Project-Generate.ps1`.

pshao25:
> For example, could [workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) or even manual `npm link` commands work? 

No. We will still face the same problem when two projects start to install packages in parellel. For each project (which are independent and don't know each other), it calls "TypeSpec-Project-Generate.ps1" to 1) install npm packages and 2) generate SDK by this package. The problem here is:  the package is not installed in advance, it is installed when we call "TypeSpec-Project-Generate.ps1". Therefore, no matter we are mono repo or not, as long as we install the pacakge in the same place, we are facing the process race issue.

> Are the child directory `package.json` files different

No, they are the same.

> and if not, could we instead only keep a single `package.json` in the relevant parent directory (this way `npm install` would bubble up)?

Yes, that is exactly what we are doing in this PR. We try to install the package in the parent directory and hope it is only installed once. Then it comes to the problem we are facing "two processes might install together". And we want to hear the suggestion from you if it is OK to apply a file lock implemented in this PR.

eng/common/scripts/Helpers/Lock-Helpers.ps1
```diff
@@ -0,0 +1,17 @@
+function FileLockEnter([string]$FilePath) {
```
weshaggard, line 1: I don't like this approach as it has race conditions itself so if the goal is to prevent stomping on each other then we need a higher level step. Perhaps we only do one npm install and then disable the install on all the other parallel jobs.

benbp: I'm also curious whether the race condition is an actual problem that will occur as a result of other automation, or rather a potential problem we anticipate if users run multiple npm installs in parallel?

m-nash: If we did that @pshao25 would option 1 install it locally inside the project folder?  Or is it up to the local user to know not to do things in parallel?

Would they even know what parallel options are magically happening under the covers if they ran
`dotnet build ./eng/service.proj /t:GenerateCode`

weshaggard: Couldn't we do the install once up front and then have our scripts have some opt-out, perhaps an env variable, that tells it not to do install any longer and we simply skip all those parallel installs?

pshao25:
@m-nash @weshaggard @benbp I removed the lock and add an `$env:SKIPINSTALL` to control the installation. The extra steps we need to take is: each pipeline should add this variable and a step to install the package itself. Please let me know if this solution is OK to go.

eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -14,6 +14,78 @@ $ErrorActionPreference = "Stop"
...
+            $replacementPackageJson = &$GetEmitterPackageJsonPathFn
+        }
+
+        $emitterName = &$GetEmitterNameFn
+        $installedPath = Join-Path $root "node_modules" $emitterName "package.json"
+        if (Test-Path $installedPath) {
+            $installedVersion = GetNpmPackageVersion $installedPath
+            Write-Host "Installed emitter version: $installedVersion"
+
+            $replacementVersion = GetEmitterDependencyVersion $replacementPackageJson
```
weshaggard, line 56: Why not pass in the emitterName here instead of having the function called again?

eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -36,17 +108,9 @@ function NpmInstallForProject([string]$workingDirectory) {
             Remove-Item -Path "package-lock.json" -Force
```
weshaggard, line 108: Don't we also want to remove the places we are deleting the files here? I don't think we want to delete those unless we actually do an npm install.

eng/common/scripts/TypeSpec-Project-Generate.ps1
weshaggard, line 56: Why not pass in the emitterName here instead of having the function called again?

eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -14,6 +14,78 @@ $ErrorActionPreference = "Stop"
...
+    }
+}
+
+function NpmInstallAtRoot() {
+    $root = git rev-parse --show-toplevel
+    Push-Location $root
+    try {
+        #default to root/eng/emitter-package.json but you can override by writing
+        #Get-${Language}-EmitterPackageJsonPath in your Language-Settings.ps1
+        $replacementPackageJson = "$PSScriptRoot/../../emitter-package.json"
```
weshaggard, line 45: We should probably pick either `$root` or `$PSScriptRoot` and be consistent with that throughout.

eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -14,6 +14,78 @@ $ErrorActionPreference = "Stop"
...
+
+        if (Test-Path "node_modules") {
+            Remove-Item -Path "node_modules" -Force -Recurse
+        }
+
+        Write-Host("Copying package.json from $replacementPackageJson")
+        Copy-Item -Path $replacementPackageJson -Destination "package.json" -Force
+        npm install --no-lock-file
+
+        if (Test-Path "package.json") {
```
weshaggard, line 75: Why are we removing these after install? Also there shouldn't be a lock file if you are passing --no-lock-file

eng/common/scripts/Helpers/Lock-Helpers.ps1
m-nash: I prefer command line parameter vs environment.  @weshaggard  is there a specific advantage you are after for using env?

weshaggard: It ends up depending on how deep this needs to be passed. Env variables make it easier to set on the outside and not need to worry about the layers. Also I suspect this is going to be a challenge for other languages as well which doesn't have msbuild type properties so we would need some way to do this for them as well.

eng/common/scripts/TypeSpec-Project-Generate.ps1
pshao25: Which function do you think I'm calling twice? Here I want to get two versions of the emitter package. One is the already installed one under root folder, the other is the going-to-be installed one in `emitter-package.json`.

eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -14,6 +14,78 @@ $ErrorActionPreference = "Stop"
...
+        }
+
+        Write-Host "Installing package at $root"
+
+        if (Test-Path "node_modules") {
+            Remove-Item -Path "node_modules" -Force -Recurse
+        }
+
+        Write-Host("Copying package.json from $replacementPackageJson")
+        Copy-Item -Path $replacementPackageJson -Destination "package.json" -Force
```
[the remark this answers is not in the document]
pshao25: It shoudn't be. We don't have the `package.json` under root.

eng/common/scripts/TypeSpec-Project-Generate.ps1
pshao25: Because we shouldn't have it under the root. It is more like a temp file just helping us install emitter under root.

eng/common/scripts/TypeSpec-Project-Generate.ps1
pshao25: We need to do that in case user installs emitter under the project folder. If we don't do that, node will not load the emitter installed under root. It will load the emitter installed under the projedct folder.

eng/common/scripts/Helpers/Lock-Helpers.ps1
pshao25: It's not deep at all. Other languages don't have msbuild type properties, but they still could pass the parameter from where they call the script. (It's just we .net call the script from msbuild, so we pass the parameter from msbuild type properties)

eng/common/scripts/TypeSpec-Project-Generate.ps1
weshaggard: `$emitterName = &$GetEmitterNameFn` You are calling this on line 50 and then again inside of the GetEmitterDeependecyVersion function.

eng/common/scripts/TypeSpec-Project-Generate.ps1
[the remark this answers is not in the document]
weshaggard: We don't for .NET and maybe other languages but we do for the azure-rest-api-specs repo and other autorest repos. So I just want to make sure that we are careful here if these scripts get used in those contexts.

eng/common/scripts/TypeSpec-Project-Generate.ps1
weshaggard: If files like package.json are present under this folder do we think that folks have it committed to the repo? If so then deleting it will cause a change in the files. I think we might actually want to warn people instead of blindly deleting these.

eng/common/scripts/TypeSpec-Project-Generate.ps1
weshaggard: I don't see you setting emitterName anywhere. Also you will need to pass it as a parameter here.

eng/common/scripts/TypeSpec-Project-Generate.ps1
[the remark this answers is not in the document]
pshao25: Yes, I agree this is a problem. But I don't want to fix so many problems in a single PR. So I was just following the previous logic "we copy the package.json to where we want to install the package, run npm install, delete the package.json and package-lock.json".

Actually I don't think we should have the package.json, as Laurent mentioned the emitter is a tool, not a library. So in the latest iteration, I change the above logic to "we just run `npm install "$emitterName@$replacementVersion"`", which is the correct behavior to a tool.

But one thing to mention: even we use it as a tool, a package.json will still be generated.

eng/common/scripts/TypeSpec-Project-Generate.ps1
pshao25: This place `$workingDirectory` is in the gitignore. It is actually `{root}\sdk\anomalydetector\Azure.AI.AnomalyDetector\TempCadlFiles\AnomalyDetector` for example. 

The scenario I'm thinking is: before this change, there will be `package.json` and `npm_modules` in the `$workingDirectory`. Then users apply this change, and they should load packages from root folder. So I want to remove the  `package.json` and `npm_modules` existing before.

eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -46,27 +46,40 @@ function NpmInstallAtRoot() {
...
+            Write-Host "Remove existing node_modules at"(Get-Location)
             Remove-Item -Path "node_modules" -Force -Recurse
         }
 
-        npm install "$emitterName@$replacementVersion" --no-lock-file
+        $hasPackageFile = Test-Path "package.json"
+
+        $npmInstallCommand = "npm install --prefix $root $emitterName@$replacementVersion --no-package-lock --omit=dev"
+        Write-Host($npmInstallCommand)
+        Invoke-Expression $npmInstallCommand
```
benbp, line 75:
```suggestion
        Invoke-Expression $npmInstallCommand
        if ($LASTEXITCODE) { exit $LASTEXITCODE }
```

eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -14,6 +14,80 @@ $ErrorActionPreference = "Stop"
...
+        $packageVersion = node -p -e "require('./$packageName').version"
+        return $packageVersion
+    }
+    finally {
+        Pop-Location
+    }
+}
+
+function GetEmitterDependencyVersion([string]$packagePath) {
+    $matchString = "`"$emitterName`": `"(.*?)`""
```
weshaggard, line 32: `$emitterName` is not set anywhere or passed in here.

pshao25: It is set globally in Line 121.

eng/common/scripts/TypeSpec-Project-Generate.ps1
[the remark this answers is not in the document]
weshaggard: I've added more comments to where you added that install step as I think we need to think about that change a little more.

eng/common/scripts/TypeSpec-Project-Generate.ps1
weshaggard: If this is under the temp working directory why would there ever be a package.json and node_modules files under that directory?

weshaggard: If that is the problem we are worried about we should just have folks remove the entire temp working directory (i.e. `git clean -xdf`). I don't think we should try and special case this one upgrade scenario.

eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -14,6 +14,81 @@ $ErrorActionPreference = "Stop"
...
+        Write-Host "Installing package at"(Get-Location)
+
+        if (Test-Path "node_modules") {
+            Write-Host "Remove existing node_modules at"(Get-Location)
+            Remove-Item -Path "node_modules" -Force -Recurse
+        }
+
+        $hasPackageFile = Test-Path "package.json"
+
+        $npmInstallCommand = "npm install --prefix $root $emitterName@$replacementVersion --no-package-lock --omit=dev"
```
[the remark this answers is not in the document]
weshaggard: We should make an explicit decision to make this change as it has larger implications on other things like other dependencies. I don't want to change to this unless we have thought it through and switched to that strategy everywhere.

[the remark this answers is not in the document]
raych1: @weshaggard , it's not aligned for all language emitters. I would suggest to use package.json for installation rather than install emitter individually. we could drop package.json to `sdk` folder to avoid conflicts with package.json in root folder.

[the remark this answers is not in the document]
MaryGao: @weshaggard This is the case the extra dependency is NOT needed for TS emitter but required for TypeSpec files. 

Not only for package `@typespec/versioning`, but also for below packages if TypeSpec project referred as dependencies.

```
@azure-tools/typespec-autorest
@typespec/openapi
@typespec/openapi2
...
```
Adding these dependencies into emitter is okay but may introduce [phantom dependency issue](https://rushjs.io/pages/advanced/phantom_deps/). Not sure this is best practice or not.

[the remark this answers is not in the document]
weshaggard: I tend to agree we should stick with a package.json file and I believe we can put it under the tempTypeSpec directory it just needs to live under the directory structure of wherever we put the TypeSpec project.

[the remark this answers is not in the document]
raych1: @weshaggard , I think @pshao25 's purpose is to avoid multiple times installation for emitter. Put package.json under tempTypeSpec directory doesn't work and it should be put under upper parent folder like `sdk` folder.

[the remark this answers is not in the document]
m-nash: @weshaggard the design was that we didn't have any additional dependencies in the emitter-package.json.

m-nash:
> @m-nash @weshaggard @benbp I removed the lock and add an `$env:SKIPINSTALL` to control the installation. The extra steps we need to take is: each pipeline should add this variable and a step to install the package itself. Please let me know if this solution is OK to go.

Can you detail out the dev experience when running things locally?

It sounds like we would need to do the following.

- copy the emitter-package.json to the root of the lang repo
- run `npm install` on this package.json
- Set this env variable `SKIPINSTALL`
- reload the env variables for all tools / shells that might use that env variable but cache these
- run our sync / generate commands

This seems overly cumbersome for a local dev experience.  Or are we assuming that locally you will never get this optimization to take effect?

Instead of an env variable have we considered an optional command line argument?

weshaggard:
> > @m-nash @weshaggard @benbp I removed the lock and add an `$env:SKIPINSTALL` to control the installation. The extra steps we need to take is: each pipeline should add this variable and a step to install the package itself. Please let me know if this solution is OK to go.
> 
> Can you detail out the dev experience when running things locally?
> 
> It sounds like we would need to do the following.
> 
> * copy the emitter-package.json to the root of the lang repo
> * run `npm install` on this package.json
> * Set this env variable `SKIPINSTALL`
> * reload the env variables for all tools / shells that might use that env variable but cache these
> * run our sync / generate commands
> 
> This seems overly cumbersome for a local dev experience. Or are we assuming that locally you will never get this optimization to take effect?
> 
> Instead of an env variable have we considered an optional command line argument?

I don't think we are depending on any environment variables any longer. @pshao25 has switched to a model of simply looking to see if it is already installed and if so skip install otherwise install. I believe that should work just fine for the local dev scenarios as well.

eng/common/scripts/TypeSpec-Project-Generate.ps1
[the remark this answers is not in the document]
hallipr: I don't think this will accommodate the devDependency pinning and emitter-package-lock.json work we're doing. We would prefer to `npm ci` using `emitter-package.json` and `emitter-package-lock.json` rather than explicit version installation.

PR #5691, opened 2023-03-13
typespec renaming

chunyu3:
- renaming cadl to typespec


eng/common/scripts/Typespec-Project-Generate.ps1
```diff
@@ -55,17 +55,17 @@ function NpmInstallForProject([string]$workingDirectory) {
 
 $resolvedProjectDirectory = Resolve-Path $ProjectDirectory
 $emitterName = &$GetEmitterNameFn
-$cadlConfigurationFile = Resolve-Path "$ProjectDirectory/cadl-location.yaml"
+$typespecConfigurationFile = Resolve-Path "$ProjectDirectory/typespec-location.yaml"
```
weshaggard, line 58: Changes such as this will be dependent on getting all the repo's updated to use this new file.

mikeharder: I'm not sure about the details of this tool, but I know for some tools like "Cadl Validation", we are planning to support both "cadl" and "typespec" names and file extensions until everything has been converted.

weshaggard: That might be an interesting approach here as well.

chunyu3: Yes. we can do this.

benbp:
The convention that I have seen is either `TypeSpec` when capitalized, or `typespec` when lowercased. So anywhere you are using `Typespec` I think that should be changed to `TypeSpec.`

doc/common/Typespec-Project-Scripts.md
```diff
@@ -68,31 +68,31 @@ function Get-dotnet-EmitterPackageJsonPath() {
...
 }
 ```
 
 ## Per project setup
 
-Each project will need to have a configuration file that will tell the scripts where to find the cadl spec.
+Each project will need to have a configuration file that will tell the scripts where to find the typespec spec.
 
-### cadl-location.yaml
+### typespec-location.yaml
```
iscai-msft, line 85: i think we should do `tsp-location.yaml` instead. I think it corresponds better with the reaming we've seen in our cadl files, i.e. `cadl-config.yml` is now `tspconfig.yml` etc

chunyu3: Thanks @iscai-msft Good suggestion. I updated.

chunyu3:
> The convention that I have seen is either `TypeSpec` when capitalized, or `typespec` when lowercased. So anywhere you are using `Typespec` I think that should be changed to `TypeSpec`.

Thanks @benbp I updated.

weshaggard (approved):
Given we are doing this side-by-side with the cadl files until we can convert everything I'm good with this change. Lets make sure we have a tracking issue to clean these up once things are converted.

eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -0,0 +1,101 @@
...
+    if ($LASTEXITCODE) { exit $LASTEXITCODE }
+}
+finally {
+    Pop-Location
+}
+
+$shouldCleanUp = $configuration["cleanup"] ?? $true
+if ($shouldCleanUp) {
+    Remove-Item $tempFolder -Recurse -Force
+}
```
mikeharder, line 101: NIT: Missing newline at end-of-file.

chunyu3: updated

PR #5715, opened 2023-03-15
Dump Proxy Version on every command

scbedd:
Per feedback from @benbp


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Startup.cs
```diff
@@ -132,6 +132,10 @@ private static async Task Run(object commandObj)
             new GitProcessHandler().VerifyGitMinVersion();
             DefaultOptions defaultOptions = (DefaultOptions)commandObj;
 
+            var assembly = System.Reflection.Assembly.GetExecutingAssembly();
```
scbedd, line 135:
```suggestion
```

benbp (approved):

PR #5726, opened 2023-03-16
[stress] stress test rerun failed jobs feature

### LINKED ISSUE #5361: [stress] Allow for re-running only failed stress tests
Most stress testers are now running > 7ish tests. 

If a couple of these tests fail to run (for instance, deployment fails) you end up with a partially completed deployment. Today, you can run the entire job again but this will also destroy and start a brand new set of pods for each stress test.

What would be nice is there was something simple that I can run that will relaunch the pods that have failed in a way that doesn't force all the jobs to get redeployed so, as I fix things, I can get my stress tests to 100%. This is particularly valuable if a pod fails when the others have been running for a couple of hours or days, where I don't want to lose my progress.

@benbp and I discussed a few ideas around this. The one that we want to go with is to just craft a separate helm deployment for the "failed" run, but still run it in the same namespace. When we create the new helm deployment we should filter down to only the tests that have failed (either in init, or in the pod itself). This is similar to what a lot of test runners offer where you can "rerun only failed tests".

The ability to change the release name could be useful for other things as well but for now "rerun with failed" is the first use case.

ckairen:
closes #5361


eng/common/scripts/stress-testing/deploy-stress-tests.ps1
```diff
@@ -25,6 +25,8 @@ param(
     # Renders chart templates locally without deployment
     [Parameter(Mandatory=$False)][switch]$Template,
 
+    [Parameter(Mandatory=$False)][switch]$RerunFailedJobs,
```
benbp, line 28: I know all the wording in the original issue says "Rerun" but I'm thinking "Retry" is a better choice here?

benbp: Also it would be retrying failed pods, not jobs. Should we call it something like `RetryFailedTests`?

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -215,9 +216,56 @@ function DeployStressPackage(
...
-    
+
     $genValFile = Join-Path $pkg.Directory "generatedValues.yaml"
     $genVal = Get-Content $genValFile -Raw | ConvertFrom-Yaml -Ordered
+    $releaseName = $pkg.ReleaseName
+    if ($RerunFailedJobs) {
+        $pods = kubectl get pods -n $pkg.namespace -o json | ConvertFrom-Json
+
+        # Get all jobs within this helm release
+        $helmResources = helm status -n $pkg.Namespace $pkg.ReleaseName --show-resources
```
benbp, line 227: A comment here with an example of the output would be nice.

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -215,9 +216,56 @@ function DeployStressPackage(
...
+        $discoveredJob = $False
+        $jobs = @()
+        foreach ($helmResource in $helmResources) {
+            if ($discoveredJob -and $helmResource -match "==>") {break}
+            if ($discoveredJob) {
+                $jobs += ($helmResource -split '\s+')[0] | where{($_ -ne "NAME") -and ($_)}
```
benbp, line 233:
```suggestion
                $jobs += ($helmResource -split '\s+')[0] | Where-Object {($_ -ne "NAME") -and ($_)}
```

tools/stress-cluster/chaos/examples/network-stress-example/Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM mcr.microsoft.com/cbl-mariner/base/core:2.0
 # Included packages: https://github.com/microsoft/CBL-Mariner/blob/1.0/SPECS/core-packages/core-packages.spec
 
 ADD ./poll.sh /poll.sh
+RUN tdnf -y install wget
```
benbp, line 5: Ooh yay thank you

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -215,9 +216,56 @@ function DeployStressPackage(
...
+        $releaseName = "$($pkg.ReleaseName)-$revision-retry"
+
+        $genValRerun = @{"scenarios"=@()}
+        foreach ($failedScenario in $failedJobsScenario) {
+            $failedScenarioObject = $genVal.scenarios | Where {$_.Scenario -eq $failedScenario}
+            $genValRerun.scenarios += $failedScenarioObject
+        }
+
+        if (!$genValRerun.scenarios.length) {
+            Write-Host "There are no failed jobs to rerun."
```
benbp, line 263: Same comment here re: naming

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -215,9 +216,56 @@ function DeployStressPackage(
     if ($LASTEXITCODE) {exit $LASTEXITCODE}
 
     $dockerBuildConfigs = @()
-    
+
     $genValFile = Join-Path $pkg.Directory "generatedValues.yaml"
     $genVal = Get-Content $genValFile -Raw | ConvertFrom-Yaml -Ordered
+    $releaseName = $pkg.ReleaseName
+    if ($RerunFailedJobs) {
+        $pods = kubectl get pods -n $pkg.namespace -o json | ConvertFrom-Json
```
benbp, line 224: Can we move this into its own helper function? That will make the outer function smaller and it will be easier to tell which variables are set/updated from this scope that are important for the remaining function code.

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
ckairen: what's the advantage of doing so?

ckairen: "where" seems to only work on arrays & will return empty array instead of null when no matches were found

benbp: We favor using full cmdlet names to reduce confusion (see [explanation](https://github.com/Azure/azure-sdk-tools/blob/main/doc/development/powershell.md#:~:text=Use%20full%20CmdLet%20names%20for%20better%20scripts)). For example, `where` as it is used here is an alias for `Where-Object` and is specifically different from `.Where()` which I think you are describing?

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -375,3 +381,68 @@ function CheckDependencies()
...
+    $jobs = @()
+    foreach ($helmResource in $helmResources) {
+        if ($discoveredJob -and $helmResource -match "==>") {break}
+        if ($discoveredJob) {
+            $jobs += ($helmResource -split '\s+')[0] | Where-Object {($_ -ne "NAME") -and ($_)}
+        }
+        if ($helmResource -match "==> v1/Job") {
+            $discoveredJob = $True
+        }
+    }
```
benbp, line 419: Another suggestion for clearer variable names:

```suggestion
    foreach ($line in $helmStatusOutput) {
        if ($discoveredJob -and $line -match "==>") {break}
        if ($discoveredJob) {
            $jobs += ($line -split '\s+')[0] | Where-Object {($_ -ne "NAME") -and ($_)}
        }
        if ($line -match "==> v1/Job") {
            $discoveredJob = $True
        }
    }
```

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -375,3 +381,68 @@ function CheckDependencies()
...
+
+    $failedJobsScenario = @()
+    $revision = 0
+    foreach ($job in $jobs) {
+        $jobRevision = [int]$job.split('-')[-1]
+        if ($jobRevision -gt $revision) {
+            $revision = $jobRevision
+        }
+
+        $podPhase = kubectl describe jobs -n $pkg.Namespace $job | Select-String "0 Failed"
```
benbp, line 429:
```suggestion
        $jobOutput = RunOrExitOnFailure kubectl describe jobs -n $pkg.Namespace $job
        $podPhase = $jobOutput | Select-String "0 Failed"
```

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -375,3 +381,68 @@ function CheckDependencies()
     }
 
 }
+
+function runRetryTests ($pkg, [ref]$releaseName, [ref]$genVal) {
+    $pods = kubectl get pods -n $pkg.namespace -o json | ConvertFrom-Json
```
benbp, line 386:
```suggestion
    $podOutput = RunOrExitOnFailure kubectl get pods -n $pkg.namespace -o json
    $pods = $podOutput | ConvertFrom-Json
```

benbp:
Just some exit handling and naming comments otherwise this is looking good. I think because the variables are untyped and getting used across so much code we need to be extra careful that the variable names describe their underlying values accurately.

eng/common/scripts/stress-testing/test.ps1
```diff
@@ -0,0 +1,16 @@
+
+function determineRetryTests ([ref] $val, [ref]$val2) {
```
benbp, line 2: Did you mean to check this in? If so let's make it a pester test: https://github.com/Azure/azure-sdk-tools/blob/main/doc/development/powershell.md#unit-testing

ckairen: oh no, accident, it's my testing ps script

benbp (approved):

Issue #5660, opened 2023-03-08
[Prioritized] Retry all git commands (aside from internal mechanisms)


scbedd:
Specifically in the case of a **timeout**, not other command errors. Is there any way to verify that exit code?

scbedd:
What this looks like in log:

```bash
2023-03-08T20:08:36.6127360Z ========================== Starting Command Output ===========================
2023-03-08T20:08:36.6165040Z [command]/usr/local/bin/pwsh -NoLogo -NoProfile -NonInteractive -Command . '/Users/runner/work/_temp/eb6b21ee-5314-44bf-9a9b-f90564d6af6a.ps1'
2023-03-08T20:08:37.1478840Z git --version
2023-03-08T20:08:37.2786400Z git remote -v
2023-03-08T20:08:37.2979630Z git clone -c core.longpaths=true --no-checkout --filter=tree:0 [https://github.com/Azure/azure-sdk-assets](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FAzure%2Fazure-sdk-assets&data=05%7C01%7Cscbedd%40microsoft.com%7C6fbdcd3242ef4ee4e2e608db201d51ae%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638139082529505728%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=D2L5lcciGUtHx7HmcIRrkLdyzbTGSZJL2rY78TmeUvo%3D&reserved=0) .
[...]

PR #5747, opened 2023-03-18
Add retry on specific git output

scbedd:
Attempts to address #5660

Given the fact that we're shelling out and the exit code isn't different for every error, this looks like the way to go.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitProcessHandler.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitProcessHandler.cs
@@ -8,6 +8,8 @@ using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Azure.Sdk.Tools.TestProxy.Common;
 using Azure.Sdk.Tools.TestProxy.Common.Exceptions;
+using Microsoft.Extensions.Internal;
+using NuGet.Protocol.Core.Types;
 
 namespace Azure.Sdk.Tools.TestProxy.Store
 {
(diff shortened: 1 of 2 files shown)


benbp (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitProcessHandler.cs
```diff
@@ -127,54 +127,71 @@ public virtual CommandResult Run(string arguments, string workingDirectory)
...
+                                continueToAttempt = IsRetriableGitError(result);
 
-                        if (result.ExitCode != 0)
-                        {
-                            throw new GitProcessException(result);
+                                if(!continueToAttempt)
+                                {
+                                    throw new GitProcessException(result);
+                                }
+                            }
```
benbp, line 188: This could be simplified by just removing the whole else block.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitProcessHandler.cs
```diff
@@ -214,52 +277,69 @@ public virtual bool TryRun(string arguments, string workingDirectory, out Comman
...
-                            Arguments = arguments
-                        };
+                            if (commandResult.ExitCode == 0)
+                            {
+                                break;
+                            }
+                            else if (commandResult.ExitCode != 0)
+                            {
+                                continueToAttempt = IsRetriableGitError(commandResult);
+                            }
```
benbp, line 336:
```suggestion
                            continueToAttempt = IsRetriableGitError(commandResult);
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitProcessHandler.cs
scbedd: I think I got what you were asking for in my latest commit.

Issue #5712, opened 2023-03-15
Creation of identical `assets` tag should continue without error


scbedd:
Currently, the generated tag for a commit to the assets rep is:

`<TagPrefix>_<SHA>`

Where `SHA`  is the commit sha of the assets repo for the new tag.

We are running into an issue where the proxy doesn't properly handle a situation where we've generated a new tag that _already exists_ in the assets repo. This causes the migration script to set an empty tag within the assets.json.

@witemple-msft and the JS folks discovered this while implementing their [migration tool](https://github.com/Azure/azure-sdk-for-js/pull/25229).

Solution: if a tag we're attempting to push already exists, we don't have an error and can simply return the tag as-is.

Repro of this problem:
1. use the migration script against an existing package. It generates an assets.json, pushes new tag, and deletes recordings in place.
2. `git reset --hard` (still within the language repo that has the new assets.json + deleted recordings)
3. Run migration script again for same target. Assets is generated with no target tag.

Implementation suggestion:

> I think we can pretty much do a `git ls-remote origin ${generatedTagName}` and see if it spits out the commit sha of the current branch.

scbedd:
Addressed in #5748

PR #5748, opened 2023-03-18
Prevent duplicate tag push

scbedd:
...and ensure that ending up with a duplicate tag offers a seamless path forward.

Addresses #5712 

- [x] add integration test that actually exercises this scenario

This PR also updates some of the TestHelper infrastructure to:
- [x] Reduce clone time when cloning to a temp dir for pushing tag cleanup
- [x] Properly swap the `assets` config to a pushed integration tag (from C# integration tests). We were leaking tags from the C# integration tests. (Note the multiple additional `assets` cleanups)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
```diff
@@ -124,7 +124,20 @@ public async Task<string> GetPath(string pathToAssetsJson)
...
+
+                    if (remoteResult.ExitCode != 0)
+                    {
+                        if (string.IsNullOrWhiteSpace(remoteResult.StdOut))
+                        {
+                            GitHandler.Run($"push origin {generatedTagName}", config);
+                        }
+                        else
+                        {
+                            _consoleWrapper.WriteLine("Not attempting to push new tag, as we have found it already within the assets repo");
```
benbp, line 138:
```suggestion
                            _consoleWrapper.WriteLine($"Not attempting to push tag '{generatedTagName}', as it already exists within the assets repo");
```

benbp (approved):

scbedd:
Unfortunately I broke the integration tests somehow so this won't be going in as-is.

benbp (approved):

PR #5802, opened 2023-03-24
Render response models breadth-first

### LINKED ISSUE #5795: Properties of models should not be interspersed with other models
In a recent [review](https://apiview.dev/Assemblies/Review/f24ce6901b6748c4a8567986cb56852d?diffRevisionId=a5dad105338141f68f79a382fd1af392&diffOnly=False&revisionId=b6a2d1fb4f774d1d81c295876c90ba7c&doc=False), a particular response was a bit confusing because it showed a `nextLink` under a `RetentionPolicy` model. I would've thought nothing of it except for the `GET` operation having declared `x-ms-pageable: nextLink`. This indicated the `nextLink` actually should've been "grouped" with the response model instead of appearing under `RetentionPolicy`. Almost seems like property models are getting expanded "inline" with the models that include them, which makes the rendered view difficult to interpret.

heaths:
Swagger response models were being rendered depth-first which made results very misleading, such that response models' properties were listed and seemingly nested after nested property model types, making those properties look like part of another model.

Fixes #5795

### DIFF
--- a/tools/apiview/parsers/swagger-api-parser/SwaggerApiParser/SwaggerSpec/SwaggerTypes.cs
+++ b/tools/apiview/parsers/swagger-api-parser/SwaggerApiParser/SwaggerSpec/SwaggerTypes.cs
@@ -85,6 +85,7 @@ public class BaseSchema : ITokenSerializable
     [JsonPropertyName("$ref")] public string Ref { get; set; }
 
     private List<SchemaTableItem> tableItems;
+    private Queue<(BaseSchema, SerializeContext)> propertyQueue = new();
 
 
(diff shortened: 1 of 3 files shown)


praveenkuttappan:
Thanks @heaths  for this change. I will review and it and we can get this included in next APIView release.

praveenkuttappan:
@chidozieononiwu just FYI.

praveenkuttappan (approved):

heaths:
@praveenkuttappan when might this go into production, and is it possible to regenerate all swagger views? Or will that potentially break comments?

PR #5816, opened 2023-03-24
Update the scheduled event governor based upon the actions rate limit

JimSuplizio:
Previously, I was ulta-conservative on the governor for scheduled events. The limit was set to 100 items per scheduled task run which is still correct for non-enterprise repository actions. For enterprise, however, the actions rate limit is 15,000 and, in that case, we can allow the scheduled tasks to process up to 1000, which is the number of results you can get back from search calls in a given time without hitting the secondary rate limit.

I also put the scheduled event into a try/finally since, we can hit the secondary rate limit on subsequent calls to get the next page of results. If that happens, we still need to process any pending changes made from previous pages of search results, if there are any.

### DIFF
--- a/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/MockGitHubEventClient.cs
+++ b/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/MockGitHubEventClient.cs
@@ -458,14 +458,14 @@ namespace Azure.Sdk.Tools.GitHubEventProcessor.Tests
 
         /// <summary>
         /// Override for the GitHubEventClient funcion which computes this based upon the repository's core rate limit.
-        /// Since this isn't necessary for the tests, just return the ActionRateLimitNonEnterprise/10.
+        /// Since this isn't necessary for the tests, just return the 100 which is the size of one page.
         /// </summary>
         /// <returns>int</returns>
 #pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
         public override async Task<int> ComputeScheduledTaskUpdateLimit()
 #pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
         {
-            return RateLimitConstants.ActionRateLimitNonEnterprise/10;
+            return 100;
         }
     }
 }
--- a/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/MockGitHubEventClient.cs
+++ b/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/MockGitHubEventClient.cs
@@ -455,5 +455,17 @@ namespace Azure.Sdk.Tools.GitHubEventProcessor.Tests
         {
             return _gitHubIssuesToLock;
         }
+
+        /// <summary>
+        /// Override for the GitHubEventClient funcion which computes this based upon the repository's core rate limit.
+        /// Since this isn't necessary for the tests, just return the ActionRateLimitNonEnterprise/10.
+        /// </summary>
+        /// <returns>int</returns>
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+        public override async Task<int> ComputeScheduledTaskUpdateLimit()
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
+        {
+            return RateLimitConstants.ActionRateLimitNonEnterprise/10;
+        }
     }
 }
--- a/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/Constants/RateLimitConstants.cs
+++ b/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/Constants/RateLimitConstants.cs
@@ -1,0 +1,21 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Azure.Sdk.Tools.GitHubEventProcessor.Constants
+{
+    public class RateLimitConstants
+    {
+        // https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-github-actions
+        // When using the GITHUB_TOKEN the number of Actions per hour for an Enterprise repository is 15000/hour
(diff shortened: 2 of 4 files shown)


tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/MockGitHubEventClient.cs
```diff
@@ -455,5 +455,17 @@ public List<GitHubIssueToLock> GetGitHubIssuesToLock()
...
+        /// <summary>
+        /// Override for the GitHubEventClient funcion which computes this based upon the repository's core rate limit.
+        /// Since this isn't necessary for the tests, just return the ActionRateLimitNonEnterprise/10.
+        /// </summary>
+        /// <returns>int</returns>
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+        public override async Task<int> ComputeScheduledTaskUpdateLimit()
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
+        {
+            return RateLimitConstants.ActionRateLimitNonEnterprise/10;
```
benbp, line 468: Can you return this value wrapped in a Task so you don't have to set these warnings?

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/Constants/RateLimitConstants.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Azure.Sdk.Tools.GitHubEventProcessor.Constants
+{
+    public class RateLimitConstants
+    {
+        // https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-github-actions
+        // When using the GITHUB_TOKEN the number of Actions per hour for an Enterprise repository is 15000/hour
+        public const int ActionRateLimitEnterprise = 15000;
```
benbp, line 13: Given that we call the rate limit API at the start of this workflow anyway, do we want to just set these values dynamically based on the result?

benbp (approved):

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/EventProcessing/ScheduledEventProcessing.cs
```diff
@@ -89,6 +94,7 @@ public static async Task CloseAddressedIssues(GitHubEventClient gitHubEventClien
         {
             if (gitHubEventClient.RulesConfiguration.RuleEnabled(RulesConstants.CloseAddressedIssues))
             {
+                int ScheduledTaskUpdateLimit = await gitHubEventClient.ComputeScheduledTaskUpdateLimit();
```
weshaggard, line 97: Are we logging this value anywhere? If not I'd suggest we do.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/GitHubEventClient.cs
```diff
@@ -279,10 +282,37 @@ public async Task WriteRateLimits(string prependMessage = null)
...
+        public virtual async Task<int> ComputeScheduledTaskUpdateLimit()
+        {
+            // CoreRateLimit will be set in WriteRateLimits but if that hasn't been called yet, call it now.
+            if (CoreRateLimit == 0)
+            {
+                var miscRateLimit = await GetRateLimits();
+                CoreRateLimit = miscRateLimit.Resources.Core.Limit;
+            }
+            // If the repository is an enterprise repository allow the max number search results, 1000, to be processed.
+            if (CoreRateLimit == RateLimitConstants.ActionRateLimitEnterprise)
```
weshaggard, line 300: This exact check makes me a little uneasy, but I guess we don't have any other great way to figure this out. 

Should we instead always do `CoreRateLimit/10` and max it at 1000?

benbp: I was thinking if we wanted to do a greater than/less than vs. exact and choose it dynamically, but at the same time I think their rate limits are rather predictable.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/MockGitHubEventClient.cs
JimSuplizio: I don't see one being better than the other. If this wasn't a mock, I'd probably just do but I don't see a problem with disabling the warning and just returning the value in this case. Also, this isn't the only function that's disabling the warning because the mock doesn't make any async calls but the virtual does. I'm going to leave this as is.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/EventProcessing/ScheduledEventProcessing.cs
JimSuplizio: @weshaggard I'll log this in the ComputeScheduledTaskUpdateLimit function.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/GitHubEventClient.cs
JimSuplizio: Sure can, it'll be updated with the next commit.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/Constants/RateLimitConstants.cs
weshaggard, line 13: After the update you don't use these constants anymore so they could be removed.

weshaggard (approved):
Some unused constants but otherwise looks good.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/Constants/RateLimitConstants.cs
JimSuplizio: They were removed with the last commit.

PR #5830, opened 2023-03-27
Tweak to InitialIssueTriage and PullRequestTriage rules

JimSuplizio:
There are several changes here:

1. The InitialIssueTriage interim rule wasn't initially set to add the customer-reported and question labels. When looking at this @jsquire and I determined that the PullRequestTriage rule also needed to be updated.
2. The Author Feedback Needed and Author Feedback rules, in the README were swapped in which one should add the comment which meant the rules also had to be updated. The README and rules function comments reflect this.
3. The way Actions updated an issue or pull_request had to be changed. The short version is that the IssueUpdate can't really be used for Actions. This was causing things that the user did (updating labels, or changing things like the title and description) while the actions were being processed, to get reverted or, in some cases, completely overwritten. For labels, the IssueUpdate is a complete replace. Adding/Removing labels is now done through Octokit's IssueLabelsClient which is additive when adding labels. The only thing the IssueUpdate is used for, in Actions, is to set the state. Everything else is nulled out to prevent it updating. This fixes: #5835


jsquire (approved):

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/GitHubEventClient.cs
```diff
@@ -320,7 +408,19 @@ public IssueUpdate GetIssueUpdate(Issue issue, bool isProcessingAction = true)
...
-                    _issueUpdate = issue.ToUpdate();
+                    // For Actions, the IssueUpdate should only be used to set the state.
+                    // Everything else should be null so it doesn't touch those other fields
+                    // except for the Milestone which, if null, would clear it out if one was
+                    // set. That's the only field to pull from the payload.
+                    _issueUpdate = new IssueUpdate
+                    {
+                        Milestone = issue.Milestone == null
+                                    ? new int?()
+                                    : issue.Milestone.Number,
```
jsquire, line 419:
```suggestion
                        Milestone = issue.Milestone?.Number,
```

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/GitHubEventClient.cs
```diff
@@ -344,7 +444,20 @@ public IssueUpdate GetIssueUpdate(PullRequest pullRequest, bool isProcessingActi
...
-                    _issueUpdate = CreateIssueUpdateForPR(pullRequest);
+                    // For Actions, the IssueUpdate should only be used to set the state.
+                    // Everything else should be null so it doesn't touch those other fields
+                    // except for the Milestone which, if null, would clear it out if one was
+                    // set. That's the only field to pull from the payload.
+                    _issueUpdate = new IssueUpdate
+                    {
+                        Milestone = pullRequest.Milestone == null
+                                    ? new int?()
+                                    : pullRequest.Milestone.Number,
```
jsquire, line 455:
```suggestion
                        Milestone = pullRequest.Milestone?.Number,
```

jsquire (approved):

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/GitHubEventClient.cs
JimSuplizio: Ditto from previous.

PR #5845, opened 2023-03-28
Prepare `pylint-guidelines-checker` for pypi release

scbedd:
Hey @l0lawrence please check this out.

I figured it we're going to upload this thing to pypi we should at least have the readme properly showing up, etc. No reason to have had it before!


l0lawrence:
Looks good to me, but should we also add in a changelog now?

scbedd:
> Looks good to me, but should we also add in a changelog now?

Answered offline, but for the public:

> only if we need to release another version

l0lawrence (approved):

Issue #5833, opened 2023-03-27
Work related to the rename of Cadl -> TypeSpec


praveenkuttappan:
Create new pipeline to generate TypeSpec API review manually.

PR #5848, opened 2023-03-28
Create new pipeline to generate TypeSpec API review manually

praveenkuttappan:
Pipeline and script to generate TypeSpec API review manually. This is a side by side support in the same way CADL API review is generated.

### DIFF
--- a/eng/scripts/Create-Apiview-Token-TypeSpec.ps1
+++ b/eng/scripts/Create-Apiview-Token-TypeSpec.ps1
@@ -10,15 +10,10 @@ param (
 )
 
 Set-StrictMode -Version 3
-$sparseCheckoutFile = ".git/info/sparse-checkout"
 
 
 function Sparse-Checkout($branchName, $packagePath)
 {
-    if (Test-Path $sparseCheckoutFile)
-    {
-        Remove-Item $sparseCheckoutFile -Force
-    }    
     git sparse-checkout init
     git sparse-checkout set $packagePath
     git checkout $branchName
--- a/eng/scripts/Create-Apiview-Token-Cadl.ps1
+++ b/eng/scripts/Create-Apiview-Token-Cadl.ps1
@@ -10,15 +10,9 @@ param (
 )
 
 Set-StrictMode -Version 3
-$sparseCheckoutFile = ".git/info/sparse-checkout"
-
 
 function Sparse-Checkout($branchName, $packagePath)
 {
-    if (Test-Path $sparseCheckoutFile)
-    {
-        Remove-Item $sparseCheckoutFile -Force
-    }    
     git sparse-checkout init
     git sparse-checkout set $packagePath
     git checkout $branchName


praveenkuttappan:
Successfully generated review using following pipeline: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2661472&view=logs&j=aa63fed7-7abe-58e5-e274-079c1c8d88b3&t=4dc27163-a637-528f-7162-0ab582d58cb2

praveenkuttappan:
@weshaggard @maririos Can you please approve this PR? This is TypeSpec sandboxing pipeline to generate review manually

eng/pipelines/apiview-review-gen-typespec.yml
```diff
@@ -0,0 +1,70 @@
...
+
+pool:
+  name: azsdk-pool-mms-ubuntu-2204-general
+  vmImage: MMSUbuntu22.04
+
+variables:
+  NodeVersion: '18.x'
+
+jobs:
+- job: CreateSwaggerReviewCodeFile
```
maririos, line 21: is Job something referenced somewhere else? or we can change the name? asking as it contains `Swagger` on it

maririos (approved):

eng/scripts/Create-Apiview-Token-TypeSpec.ps1
```diff
@@ -0,0 +1,110 @@
...
+
+
+function Sparse-Checkout($branchName, $packagePath)
+{
+    if (Test-Path $sparseCheckoutFile)
+    {
+        Remove-Item $sparseCheckoutFile -Force
+    }    
+    git sparse-checkout init --no-cone
+    git sparse-checkout set $packagePath $globalPackageJson
```
weshaggard, line 24: Is this actually needed? Doesn't sparse-checkout automatically include the files in the root?

eng/scripts/Create-Apiview-Token-TypeSpec.ps1
```diff
@@ -0,0 +1,110 @@
...
+    [ValidateNotNullOrEmpty()]
+    [string] $Reviews,
+    [ValidateNotNullOrEmpty()]
+    [string] $WorkingDir,
+    [ValidateNotNullOrEmpty()]
+    [string] $OutputDir,
+    [string] $GitPat = ""
+)
+
+Set-StrictMode -Version 3
```
weshaggard, line 12: Doesn't have to be now but can you file an issue to figure out how to use https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/TypeSpec-Project-Sync.ps1 so we can share this syncing process with the other emitters?

praveenkuttappan: Done. https://github.com/Azure/azure-sdk-tools/issues/5856

eng/scripts/Create-Apiview-Token-TypeSpec.ps1
praveenkuttappan: I will check this again.

eng/pipelines/apiview-review-gen-typespec.yml
praveenkuttappan: Thanks. I will update it.

eng/scripts/Create-Apiview-Token-TypeSpec.ps1
benbp: 1. Sparse checkout by default WILL get files in the root. Also if you're not doing negative globs, I recommend not doing `--no-cone` as it's not necessary.
2. I think it would be better to put the `git clone` command into this function.
3. You don't need the conditional sparse checkout removal. `init` is idempotent and `set` will overwrite any previous contents of the sparse config file.

praveenkuttappan: Thanks @weshaggard  and @benbp. This pipeline will process API review gen for multiple reviews when we add support for review upgrade and it's possible to have each review generated from different forks of specs repo or different branches within same repo. `git clone` is skipped when iterating over the reviews if it's same repo. I have removed `--no-clone` and sparse checkout config file removal.

benbp (approved):

praveenkuttappan:
This PR will fix https://github.com/Azure/azure-sdk-tools/issues/5833

PR #5884, opened 2023-03-31
Remove project-to-label action

weshaggard:
Now that github has added direct support for label to project in the project workflows we can remove this from the repo. I'm going to keep this PR open until I go and configure all the existing labels we are using.


weshaggard:
I updated all the existing configuration directly in the projects. @maririos if you see any gaps let me know.

PR #5901, opened 2023-04-04
Updated ApiView to use clang-15 to work around MSVC STL changes; fixed warning in keyvault-admin ApiView

LarryOsterman:
Use a custom vcpkg port for clang-15 because vcpkg currently doesn't have clang-15 available.

clang-15 port files liberally stolen from https://github.com/microsoft/vcpkg/pull/26902.

### DIFF
--- a/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
+++ b/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
@@ -2759,6 +2759,8 @@ std::unique_ptr<AstNode> AstNode::Create(
       //    azureClassesDatabase));
     case Decl::Kind::Using:
       return nullptr;
+    case Decl::Kind::UsingDirective:
+      return nullptr;
       //   return std::make_unique<AstUsing>(cast<UsingDecl>(decl));
     default: {
       llvm::errs() << raw_ostream::Colors::RED << "Unknown DECL node "
--- a/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
+++ b/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
@@ -2148,6 +2148,39 @@ public:
   }
 };
 
+class AstUsingDirective : public AstNode {
+  std::string m_namedNamespace;
+
+public:
+  AstUsingDirective(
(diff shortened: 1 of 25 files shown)


LarryOsterman:
Ping :).

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -2748,6 +2781,12 @@ std::unique_ptr<AstNode> AstNode::Create(
     case Decl::Kind::Friend:
       return std::make_unique<AstFriend>(cast<FriendDecl>(decl), azureClassesDatabase, parentNode);
 
+    case Decl::Kind::UsingDirective:
+      // A "UsingDirective" is a "using namespace" directive. We consider this to be an error
+      // condition, so we add an AstNode so the error will appear in the ApiView.
```
RickWinter, line 2786: Nit
```suggestion
      // A "UsingDirective" is a "using namespace" directive. We consider this an error
      // condition, add an AstNode so the error appears in the ApiView.
```

tools/apiview/parsers/cpp-api-parser/ParseTests/tests.cpp
```diff
@@ -529,7 +528,42 @@ TEST_F(TestParser, Templates)
...
+  {
+    if (msg.DiagnosticId == "CPA0009")
+    {
+      usingNamespaces += 1;
+    }
+  }
+  EXPECT_EQ(usingNamespaces, 1ul);
 }
 
 #if 0
```
RickWinter, line 569: Are these two tests every going to be valid?  We should enable or remove.

RickWinter (approved):
Can you add notes into a README explaining why we need to do the port overlay.

tools/apiview/parsers/cpp-api-parser/ParseTests/tests.cpp
LarryOsterman: Could you help me understand how this would stop being valid in the future?

RickWinter: I was referring to the #if 0 tests further below.  Unfortunately, the PR mechanisms didn't let me highlight those lines.  Its an orthogonal problem unrelated to your PR, but since you're in here I noticed/mentioned it.

PR #5911, opened 2023-04-04
stress test error out for breaking helm versions

ckairen:
Stress test deployment now requires helm version 3.11.0+, now throwing a more descriptive error

### DIFF
--- a/eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
+++ b/eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
@@ -386,8 +386,15 @@ function generateRetryTestsHelmValues ($pkg, $releaseName, $generatedHelmValues)
     $podOutput = RunOrExitOnFailure kubectl get pods -n $pkg.namespace -o json
     $pods = $podOutput | ConvertFrom-Json
 
+    # helm version example: v3.11.2+g912ebc1
+    $helmVer = (helm version --short).substring(1) -replace '\+.*',''
+    $helmVer = [AzureEngSemanticVersion]::new($helmVer)
+    $minHelmVer =[AzureEngSemanticVersion]::new("3.11.0")
+    Write-Host $helmVer.CompareTo($minHelmVer)
+    if ($helmVer.CompareTo($minHelmVer) -le 0) {
+        throw "Please update helm to version >= 3.11.0`nAdditional information for updating helm version can be found here: https://helm.sh/docs/intro/install/"
+    }
     # Get all jobs within this helm release
-    
     $helmStatusOutput = RunOrExitOnFailure helm status -n $pkg.Namespace $pkg.ReleaseName --show-resources
     # -----Example output-----
     # NAME: <Release Name>
--- a/eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
+++ b/eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
@@ -4,6 +4,7 @@ $ErrorActionPreference = 'Stop'
 $FailedCommands = New-Object Collections.Generic.List[hashtable]
 
 . (Join-Path $PSScriptRoot "../Helpers" PSModule-Helpers.ps1)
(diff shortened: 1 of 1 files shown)


eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -386,8 +387,14 @@ function generateRetryTestsHelmValues ($pkg, $releaseName, $generatedHelmValues)
     $podOutput = RunOrExitOnFailure kubectl get pods -n $pkg.namespace -o json
     $pods = $podOutput | ConvertFrom-Json
 
+    # helm version example: v3.11.2+g912ebc1
+    $helmVer = (helm version --short).substring(1) -replace '\+.*',''
+    $helmVer = [AzureEngSemanticVersion]::new($helmVer)
+    $minHelmVer = [AzureEngSemanticVersion]::new("3.11.0")
+    if ($helmVer.CompareTo($minHelmVer) -le 0) {
+        throw "Please update helm to version >= 3.11.0`nAdditional information for updating helm version can be found here: https://helm.sh/docs/intro/install/"
+    }
```
benbp, line 396: The `AzureEngSemanticVersion` class implements `IComparable` so you can just do `-lt`. Also I suggest you rename the variables to be more readable, no good reason to shorten them here, as well as adding a `$MIN_HELM_VERSION` constant up at the top of the file so it's easier to update.

```suggestion
    # helm version example: v3.11.2+g912ebc1
    $helmVersionString = (helm version --short).substring(1) -replace '\+.*',''
    $helmVersion = [AzureEngSemanticVersion]::new($helmVersionString)
    $minHelmVersion = [AzureEngSemanticVersion]::new($MIN_HELM_VERSION)
    if ($helmVersion -lt $minHelmVersion) {
        throw "Please update helm to version >= $MIN_HELM_VERSION (current version: $helmVersionString)`nAdditional information for updating helm can be found here: https://helm.sh/docs/intro/install/"
    }
```

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -386,8 +387,14 @@ function generateRetryTestsHelmValues ($pkg, $releaseName, $generatedHelmValues)
     $podOutput = RunOrExitOnFailure kubectl get pods -n $pkg.namespace -o json
     $pods = $podOutput | ConvertFrom-Json
 
+    # helm version example: v3.11.2+g912ebc1
```
benbp, line 390: Let's do this check in the `CheckDependencies` function rather than in the retry function. I think it's ok to require the latest version even if it's only used for a subset of features that not everyone will touch.

ckairen: Yeah that was my concern, is to not affect other's who are not using this feature But I guess it's fine to encourage people to use the latest tools

benbp: Yeah otherwise you could get halfway through an installation and then fail out. Better to alert up front if anything is missing. One option we could consider is using `CheckDependencies` but checking if `-RetryFailedTests` is set, that way you fail fast but still only if you're using the feature.

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -376,18 +379,27 @@ function CheckDependencies()
...
+    }
+
     if ($shouldError) {
         exit 1
     }
 
 }
 
 function generateRetryTestsHelmValues ($pkg, $releaseName, $generatedHelmValues) {
+    CheckDependencies
```
benbp, line 397: This gets run on entry, so we don't need it here: https://github.com/Azure/azure-sdk-tools/blob/7554cbcd100d828bb9ba75cd4993551199660f76/eng/common/scripts/stress-testing/deploy-stress-tests.ps1#L41

```suggestion
```

benbp (approved):

PR #5914, opened 2023-04-04
Fix NullReferenceException happening in ApiPath

AaronPatterson:
Ensure ApiPath does not get a NullReferenceExcepton when copying common parameters to operation method parameters.

Previously, when the operation method does not have any parameters a NullRefferenceException occurs when attempting to copy the common parameters to the operation method.

I encountered this when working on DevCenter RP and it looks like others have hit this issue as well: #5810


heaths (approved):
Assuming `this.parameters` are the parameters declared just under `"paths"` in swagger, which it seems is the case by what I see here, this LGTM.

AaronPatterson:
> Assuming `this.parameters` are the parameters declared just under `"paths"` in swagger, which it seems is the case by what I see here, this LGTM.

@heaths You are correct.  `this.parameters` are the ones declared under the path.

PR #5926, opened 2023-04-05
Update KeyVaultSecretStore.cs

weshaggard:
Fix logging name from certificate to secret.


tools/secret-rotation/Azure.Sdk.Tools.SecretRotation.Stores.KeyVault/KeyVaultSecretStore.cs
```diff
@@ -275,7 +275,7 @@ public override async Task MarkRotationCompleteAsync(SecretValue secretValue, Da
                 "The PrimaryState value passed to KeyVaultSecretStore was not of type KeyVaultSecret");
         }
 
-        this.logger.LogInformation("Adding tag 'rotation-complete' to certificate '{CertificateName}' in vault '{Vault}'",
+        this.logger.LogInformation("Adding tag 'rotation-complete' to secret '{CertificateName}' in vault '{Vault}'",
```
weshaggard, line 278: @hallipr I didn't dig deep but why is CertificateName the right value in the secret store?

PR #5935, opened 2023-04-07
Add diagnostic rules for ExpandableStringEnum

srnagar:
### Missing `fromString()` method

image

### Missing `values()` method

image

### DIFF
--- a/src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/rules/ExpandableStringEnumDiagnosticRule.java
+++ b/src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/rules/ExpandableStringEnumDiagnosticRule.java
@@ -42,10 +42,10 @@ public class ExpandableStringEnumDiagnosticRule implements DiagnosticRule {
             AtomicBoolean hasValuesMethod = new AtomicBoolean(false);
             getPublicOrProtectedMethods(cu).forEach(method -> {
                 final String methodName = method.getNameAsString();
-                if (methodName.equals("fromString") && method.isPublic()) {
+                if (methodName.equals("fromString") && method.isPublic() && method.isStatic()) {
                     hasFromStringMethod.set(true);
                 }
-                if (methodName.equals(("values")) && method.isPublic()) {
+                if (methodName.equals(("values")) && method.isPublic() && method.isStatic()) {
                     hasValuesMethod.set(true);
                 }
             });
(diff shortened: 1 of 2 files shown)


src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/rules/ExpandableStringEnumDiagnosticRule.java
```diff
@@ -0,0 +1,69 @@
...
+            if (typeDeclaration instanceof ClassOrInterfaceDeclaration) {
+                ClassOrInterfaceDeclaration classDeclaration = (ClassOrInterfaceDeclaration) typeDeclaration;
+                // check if the type extends ExpandableStringEnum
+                if (classDeclaration.getExtendedTypes() != null
+                        && classDeclaration.getExtendedTypes().stream()
+                        .anyMatch(extendedType -> extendedType.getNameAsString().equals("ExpandableStringEnum"))) {
+                    checkRequiredMethodsExist(cu, listing);
+                }
+            }
+        });
```
JonathanGiles, line 36: At some point it might be cool to have a `SubTypeDiagnosticRule` where the constructor just takes `ExpandableStringEnum` (or the appropriate parent type name), and then calls into the abstract method to process when it matches. Not saying we should do it now, but it would be nice to avoid this code block every time.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/rules/ExpandableStringEnumDiagnosticRule.java
```diff
@@ -0,0 +1,69 @@
...
+        getClassName(cu).ifPresent(className -> {
+            AtomicBoolean hasFromStringMethod = new AtomicBoolean(false);
+            AtomicBoolean hasValuesMethod = new AtomicBoolean(false);
+            getPublicOrProtectedMethods(cu).forEach(method -> {
+                final String methodName = method.getNameAsString();
+                if (methodName.equals("fromString") && method.isPublic()) {
+                    hasFromStringMethod.set(true);
+                }
+                if (methodName.equals(("values")) && method.isPublic()) {
+                    hasValuesMethod.set(true);
```
JonathanGiles, line 49: We should also test that it is static.

JonathanGiles:
Thanks for doing this - every diagnostic in APIView is one less opportunity for bad API to ship!

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/rules/ExpandableStringEnumDiagnosticRule.java
srnagar: Yeah, I was considering this but didn't have any other subtype rule now, so, didn't add it but will be a good refactoring to do when we have more than one diagnostic rule requiring this.

JonathanGiles (approved):

PR #5951, opened 2023-04-11
Switch to using standard PAT tokens instead of base 64

weshaggard:
For most of these we can use the standard System.AccessToken given to the build instead of maintaining a specific token. However that token isn't base 64 encoded so we need to encode it.

With this we can stop explicitly passing PAT's unless we need to access another DevOps org and we also don't have to remember to keep the PAT's in KV base 64 encoded.

### DIFF
--- a/eng/common/scripts/Invoke-DevOpsAPI.ps1
+++ b/eng/common/scripts/Invoke-DevOpsAPI.ps1
@@ -8,12 +8,8 @@ function Get-Base64EncodedToken([string]$AuthToken)
   $unencodedAuthTokenBytes = [System.Text.Encoding]::UTF8.GetBytes($unencodedAuthToken)
   $encodedAuthToken = [System.Convert]::ToBase64String($unencodedAuthTokenBytes)
 
-  if ($isDevOpsRun) {
-    # We are doing this here so that there is zero chance that this token is emitted in Azure Pipelines
-    # build logs. Azure Pipelines will see this text and register the secret as a value it should *** out
-    # before being transmitted to the server (and shown in logs). It means if the value is accidentally
-    # leaked anywhere else that it won't be visible. The downside is that when the script is executed
-    # on a local development box, it will be visible.
+  if (Test-SupportsDevOpsLogging) {
+    # Mark the encoded value as a secret so that DevOps will star any references to it that might end up in the logs
     Write-Host "##vso[task.setvariable variable=_throwawayencodedaccesstoken;issecret=true;]$($encodedAuthToken)"
   }
 
--- a/eng/common/scripts/Invoke-DevOpsAPI.ps1
+++ b/eng/common/scripts/Invoke-DevOpsAPI.ps1
@@ -2,6 +2,24 @@
 
 $DevOpsAPIBaseURI = "https://dev.azure.com/{0}/{1}/_apis/{2}/{3}?{4}api-version=6.0"
 
+function Get-Base64EncodedToken([string]$AuthToken)
+{
+  $unencodedAuthToken = "nobody:$AuthToken"
+  $unencodedAuthTokenBytes = [System.Text.Encoding]::UTF8.GetBytes($unencodedAuthToken)
+  $encodedAuthToken = [System.Convert]::ToBase64String($unencodedAuthTokenBytes)
+
+  if ($isDevOpsRun) {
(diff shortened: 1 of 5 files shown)


eng/common/scripts/Invoke-DevOpsAPI.ps1
```diff
@@ -2,6 +2,24 @@
...
+{
+  $unencodedAuthToken = "nobody:$AuthToken"
+  $unencodedAuthTokenBytes = [System.Text.Encoding]::UTF8.GetBytes($unencodedAuthToken)
+  $encodedAuthToken = [System.Convert]::ToBase64String($unencodedAuthTokenBytes)
+
+  if ($isDevOpsRun) {
+    # We are doing this here so that there is zero chance that this token is emitted in Azure Pipelines
+    # build logs. Azure Pipelines will see this text and register the secret as a value it should *** out
+    # before being transmitted to the server (and shown in logs). It means if the value is accidentally
+    # leaked anywhere else that it won't be visible. The downside is that when the script is executed
```
konrad-jamrozik, line 15: You wrote:

>  It means if the value is accidentally leaked anywhere else that it won't be visible.

So if I am reading it right, ADO works like that:
- somebody tries to output text "foo"
- ADO checks and sees that "foo" was being previously output (which is here) as a secret value, so it added it to its "secret values map", so it stars it out.

?

eng/common/scripts/Invoke-DevOpsAPI.ps1
```diff
@@ -2,6 +2,24 @@
...
+    # on a local development box, it will be visible.
```
konrad-jamrozik, line 16: > on a local development box, it will be visible.

Maybe we could add some additional check to see if we run on local devbox, and if so, do not apply this protection? Not sure if worth the complexity, though.

eng/common/scripts/Invoke-DevOpsAPI.ps1
```diff
@@ -2,6 +2,24 @@
 
 $DevOpsAPIBaseURI = "https://dev.azure.com/{0}/{1}/_apis/{2}/{3}?{4}api-version=6.0"
 
+function Get-Base64EncodedToken([string]$AuthToken)
+{
+  $unencodedAuthToken = "nobody:$AuthToken"
+  $unencodedAuthTokenBytes = [System.Text.Encoding]::UTF8.GetBytes($unencodedAuthToken)
+  $encodedAuthToken = [System.Convert]::ToBase64String($unencodedAuthTokenBytes)
+
+  if ($isDevOpsRun) {
```
konrad-jamrozik, line 11: Are we sure this variable will be recognized? I know that scoping in PowerShell is funky, so not sure if a body of function definition will see a variable from dotsourced file.

konrad-jamrozik (approved):

eng/common/scripts/Invoke-DevOpsAPI.ps1
weshaggard: That is what the isDevOpsRun check is about. That shouldn't be set on local dev box.

eng/common/scripts/Invoke-DevOpsAPI.ps1
weshaggard: We define that globally in our logging.ps1 at https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/logging.ps1#L1. And that is typically how we determine if we can call the DevOps specific logging commands. This variable will be defined anytime our logging functions are also defined otherwise, we will have errors.

eng/common/scripts/Invoke-DevOpsAPI.ps1
konrad-jamrozik: > That is what the isDevOpsRun check is about. That shouldn't be set on local dev box.

OK then in such case this comment:

> The downside is that when the script is executed on a local development box, it will be visible.

Is a bit confusing to me. Because I understand you are saying that it is extremely unlikely that `$isDevOpsRun` will be set in local run. Then why mention this risk of outputting things locally? OTOH if it is real risk, then we could have something like `if ($isDevOpsRun && <running_locally>) then { throw }`

weshaggard: Fair let me clean up that comment. I honestly didn't give it much thought as I just copied it from the other place but I should clean it up.

PR #5996, opened 2023-04-18
[TypeSpec APIView] ﻿Unrestrict dependencies

tjprescott:
Should allow APIView to "just work" with future TypeSpec releases without the need for a release.

### DIFF
--- a/tools/apiview/emitters/typespec-apiview/package.json
+++ b/tools/apiview/emitters/typespec-apiview/package.json
@@ -54,14 +54,13 @@
     "!dist/test/**"
   ],
   "peerDependencies": {
-    "@azure-tools/typespec-azure-core": ">=0.26 <1.0",
-    "@azure-tools/typespec-autorest": ">=0.26 <1.0",
     "@typespec/compiler": ">=0.40 <1.0",
-    "@typespec/http": ">=0.40 <1.0",
-    "@typespec/rest": ">=0.40 <1.0",
     "@typespec/versioning": ">=0.40 <1.0"
   },
   "devDependencies": {
+    "@azure-tools/typespec-azure-core": ">=0.26 <1.0",
+    "@typespec/http": ">=0.40 <1.0",
+    "@typespec/rest": ">=0.40 <1.0",
     "@typespec/eslint-plugin": ">=0.40 <1.0",
     "@typespec/library-linter": ">=0.40 <1.0",
     "@typespec/prettier-plugin-typespec": ">=0.40 <1.0",
--- a/tools/apiview/emitters/typespec-apiview/package.json
+++ b/tools/apiview/emitters/typespec-apiview/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@azure-tools/typespec-apiview",
-  "version": "0.4.3",
+  "version": "0.4.4",
   "author": "Microsoft Corporation",
   "description": "Library for emitting APIView token files from TypeSpec",
   "homepage": "https://github.com/Azure/azure-sdk-tools",
@@ -54,18 +54,18 @@
     "!dist/test/**"
   ],
   "dependencies": {
-    "@azure-tools/typespec-azure-core": "0.29.0",
-    "@azure-tools/typespec-autorest": "0.29.0",
-    "@typespec/compiler": "0.43.0",
-    "@typespec/http": "0.43.1",
-    "@typespec/rest": "0.43.0",
-    "@typespec/versioning": "0.43.0"
+    "@azure-tools/typespec-azure-core": ">=0.26 <1.0",
+    "@azure-tools/typespec-autorest": ">=0.26 <1.0",
+    "@typespec/compiler": ">=0.40 <1.0",
+    "@typespec/http": ">=0.40 <1.0",
+    "@typespec/rest": ">=0.40 <1.0",
+    "@typespec/versioning": ">=0.40 <1.0"
   },
   "devDependencies": {
-    "@typespec/eslint-plugin": "~0.43.0",
-    "@typespec/library-linter": "~0.43.0",
-    "@typespec/prettier-plugin-typespec": "^0.43.0",
-    "@typespec/eslint-config-typespec": "~0.6.0",
+    "@typespec/eslint-plugin": ">=0.40 <1.0",
+    "@typespec/library-linter": ">=0.40 <1.0",
+    "@typespec/prettier-plugin-typespec": ">=0.40 <1.0",
+    "@typespec/eslint-config-typespec": "0.x",
     "@types/mocha": "~9.1.0",
     "@types/node": "~16.0.3",
     "@changesets/cli": "^2.24.4",
--- a/tools/apiview/emitters/typespec-apiview/package.json
+++ b/tools/apiview/emitters/typespec-apiview/package.json
@@ -58,9 +58,9 @@
     "@typespec/versioning": ">=0.40 <1.0"
   },
   "devDependencies": {
-    "@azure-tools/typespec-azure-core": ">=0.26 <1.0",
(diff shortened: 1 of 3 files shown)


tools/apiview/emitters/typespec-apiview/package.json
```diff
@@ -54,18 +54,18 @@
     "!dist/test/**"
   ],
   "dependencies": {
-    "@azure-tools/typespec-azure-core": "0.29.0",
-    "@azure-tools/typespec-autorest": "0.29.0",
-    "@typespec/compiler": "0.43.0",
-    "@typespec/http": "0.43.1",
-    "@typespec/rest": "0.43.0",
-    "@typespec/versioning": "0.43.0"
+    "@azure-tools/typespec-azure-core": ">=0.26 <1.0",
```
weshaggard, line 57: Are we sure we want to do make this change? We should think about this more holistically for all the emitters and not just do it for this one. I know there is an email thread on what prompted you to update this but I haven't fully understood everything there yet. I'd suggest we hold off on this change for right now.

mikeharder: I would also like to better understand what we are trying to accomplish and how to think about compatibility across different packages and versions.

tjprescott: Well, this is the package.json for just this emitter, so I'm not sure why it would be relevant to any others. The majority of TypeSpec APIView releases do nothing more than bump the package dependencies. This would eliminate the need to publish a new version of this package just to bump the TypeSpec dependency. The only time we should have to actually bump the minimum version here is if a TypeSpec release necessitated changes in TypeSpec APIVIew. This change would help prevent the situation where a new version of TypeSpec is released and the CI blows up because there's not a compatible version of APIView released (the exact situation @lmazuel) ran into.

The only reason these dependencies were pinned in the first place is because I mistakenly thought we could not accept version ranges for pre-released packages, but that is not the case.

tools/apiview/emitters/typespec-apiview/package.json
```diff
@@ -53,19 +53,19 @@
...
   },
   "devDependencies": {
-    "@typespec/eslint-plugin": "~0.43.0",
-    "@typespec/library-linter": "~0.43.0",
-    "@typespec/prettier-plugin-typespec": "^0.43.0",
-    "@typespec/eslint-config-typespec": "~0.6.0",
+    "@typespec/eslint-plugin": ">=0.40 <1.0",
+    "@typespec/library-linter": ">=0.40 <1.0",
+    "@typespec/prettier-plugin-typespec": ">=0.40 <1.0",
+    "@typespec/eslint-config-typespec": "0.x",
```
mikeharder, line 68:
```suggestion
    "@typespec/eslint-config-typespec": ">=0.6.0 <1.0",
```

Use this to align with other deps?

mikeharder (changes requested):

tools/apiview/emitters/typespec-apiview/package.json
```diff
@@ -53,19 +53,19 @@
...
+    "@typespec/http": ">=0.40 <1.0",
+    "@typespec/rest": ">=0.40 <1.0",
+    "@typespec/versioning": ">=0.40 <1.0"
   },
   "devDependencies": {
-    "@typespec/eslint-plugin": "~0.43.0",
-    "@typespec/library-linter": "~0.43.0",
-    "@typespec/prettier-plugin-typespec": "^0.43.0",
-    "@typespec/eslint-config-typespec": "~0.6.0",
+    "@typespec/eslint-plugin": ">=0.40 <1.0",
```
mikeharder, line 65: devDependencies can stay pinned

mikeharder (changes requested):

tools/apiview/emitters/typespec-apiview/package.json
tjprescott: I feel like it makes sense to let these typespec ones float even though they are dev dependencies. Presumably if the eslint config or prettier plugin get updated we'd want to move to them, and we often don't know they've been released since they are announced like the core packages are.

tools/apiview/emitters/typespec-apiview/package.json
```diff
@@ -53,19 +53,19 @@
...
-    "@typespec/compiler": "0.43.0",
-    "@typespec/http": "0.43.1",
-    "@typespec/rest": "0.43.0",
-    "@typespec/versioning": "0.43.0"
+  "peerDependencies": {
+    "@azure-tools/typespec-azure-core": ">=0.26 <1.0",
+    "@azure-tools/typespec-autorest": ">=0.26 <1.0",
+    "@typespec/compiler": ">=0.40 <1.0",
+    "@typespec/http": ">=0.40 <1.0",
+    "@typespec/rest": ">=0.40 <1.0",
```
mikeharder, line 61: I believe `@typespec/http` and `@typespec/rest` can be moved to `devDependency` since they are only used in test code

tools/apiview/emitters/typespec-apiview/package.json
```diff
@@ -53,19 +53,19 @@
...
-  "dependencies": {
-    "@azure-tools/typespec-azure-core": "0.29.0",
-    "@azure-tools/typespec-autorest": "0.29.0",
-    "@typespec/compiler": "0.43.0",
-    "@typespec/http": "0.43.1",
-    "@typespec/rest": "0.43.0",
-    "@typespec/versioning": "0.43.0"
+  "peerDependencies": {
+    "@azure-tools/typespec-azure-core": ">=0.26 <1.0",
+    "@azure-tools/typespec-autorest": ">=0.26 <1.0",
```
mikeharder, line 58: I believe `@azure-tools/typespec-autorest` can be removed, I don't see it used anywhere in the codebase.

mikeharder (changes requested):

timotheeguerin (approved):

mikeharder (approved):

tools/apiview/emitters/typespec-apiview/package.json
```diff
@@ -53,19 +53,18 @@
...
+  "peerDependencies": {
+    "@typespec/compiler": ">=0.40 <1.0",
+    "@typespec/versioning": ">=0.40 <1.0"
   },
   "devDependencies": {
-    "@typespec/eslint-plugin": "~0.43.0",
-    "@typespec/library-linter": "~0.43.0",
-    "@typespec/prettier-plugin-typespec": "^0.43.0",
-    "@typespec/eslint-config-typespec": "~0.6.0",
+    "@azure-tools/typespec-azure-core": ">=0.26 <1.0",
```
timotheeguerin, line 61: The dev dependencies can be pinned(`~`) to the last you tested as this only matters to local developement

tjprescott: I left the linter ones open but pinned the libraries.

tools/apiview/emitters/typespec-apiview/package.json
```diff
@@ -53,19 +53,18 @@
     "dist/**",
     "!dist/test/**"
   ],
-  "dependencies": {
+  "peerDependencies": {
+    "@typespec/compiler": ">=0.40 <1.0",
+    "@typespec/versioning": ">=0.40 <1.0"
+  },
+  "devDependencies": {
     "@azure-tools/typespec-azure-core": "0.29.0",
```
weshaggard, line 61: @tjprescott is devDependency correct for these? You no longer have any runtime dependencies.

tjprescott: Based on feedback from @mikeharder he suggested making this dev dependencies since these packages are only used in the APIView tests. The hard dependencies are really just the ones listed as peer dependencies: compiler and verisoning.

PR #6003, opened 2023-04-18
Add note about sharing between WSL / windows

scbedd:
@benbp I had a 3rd person today reach out with proxy issues started with

> so I'm doing this on WSL

and I see `/mnt/c/blah/` in the problem description and immediately know what happened. Adding some documentation 👍

### DIFF
--- a/tools/test-proxy/documentation/asset-sync/README.md
+++ b/tools/test-proxy/documentation/asset-sync/README.md
@@ -67,7 +67,7 @@ Each of these CLI Commands takes an `assets.json` argument that provides the _co
 
 ### A note about using on Windows + WSL
 
-When using a Windows machine, it is technically possible to invoke tests from WSL against a windows clone. That path would appear under `/mnt/c/path/to/your/repo`. This is _not_ a supported scenario, as the `test-proxy` shells out to git for the push/restore actions. Running a `git restore` from _linux_ against a repo that was cloned down using a _windows_ git client can have unexpected results. Better to avoid the situation entirely and use an entirely separate clone for work on WSL.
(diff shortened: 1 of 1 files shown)


tools/test-proxy/documentation/asset-sync/README.md
```diff
@@ -65,6 +65,10 @@ Interactions with the external assets repository are accessible when the proxy i
 
 Each of these CLI Commands takes an `assets.json` argument that provides the _context_ that should be pushed to the external repository.
 
+### A note about using on Windows + WSL
+
+When using a Windows machine, it is technically possible to invoke tests from WSL against a windows clone. That path would appear under `/mnt/c/path/to/your/repo`. This is _not_ a supported scenario, as the `test-proxy` shells out to git for the push/restore actions. Running a `git restore` from _linux_ against a repo that was cloned down using a _windows_ git client can have unexpected results. Better to avoid the situation entirely and use an entirely separate clone for work on WSL.
```
scbedd, line 70:
```suggestion
When using a Windows machine, it is technically possible to invoke tests from WSL against a windows clone. That path would appear under `/mnt/c/path/to/your/repo`. This is _not_ a supported scenario, as the `test-proxy` shells out to git for the push/restore actions. Running a `git push/pull` from _linux_ against a repo that was cloned down using a _windows_ git client can have unexpected results. Better to avoid the situation entirely and use an entirely separate clone for work on WSL.
```

benbp:
@scbedd docs look good. I think we should also have the proxy fail to start/run if it detects a path that starts with `/mnt/[A-z]`.

benbp (approved):

scbedd:
> @scbedd docs look good. I think we should also have the proxy fail to start/run if it detects a path that starts with `/mnt/[A-z]`.

Great suggestion, will file an issue for it. Not hard at all to add.

PR #6030, opened 2023-04-21
Have event processor get configs from raw githubusercontent instead of needing a sparse checkout

JimSuplizio:
Instead if requiring a sparse-checkout in the action, pull the config files (CODEOWNERS and rules config) files from the raw githubusercontent.

One thing of note here: I realize it's odd that I'm using the FileHelpers from the CodeOwnersParser to load the rules config file but there's no common utilities project and I really didn't want to copy the code and have to add the retry logic in multiple places.

The other change was when rules are loaded. Previously rules were loaded when the GitHubEventClient was constructed but now, like the CODEOWNERS, which is loaded when it's first needed.

The DirectoryUtil is going to remain until we know we can absolutely remove it.

### DIFF
--- a/tools/code-owners-parser/CodeOwnersParser/FileHelpers.cs
+++ b/tools/code-owners-parser/CodeOwnersParser/FileHelpers.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Net;
 using System.Net.Http;
 
 namespace Azure.Sdk.Tools.CodeOwnersParser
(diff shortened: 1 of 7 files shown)


tools/code-owners-parser/CodeOwnersParser/FileHelpers.cs
```diff
@@ -22,10 +23,42 @@ public static string GetFileOrUrlContents(string fileOrUrl)
...
+            int delayTimeInMs = 1000;
             using HttpClient client = new HttpClient();
-            HttpResponseMessage response =
-                client.GetAsync(url).ConfigureAwait(false).GetAwaiter().GetResult();
-            return response.Content.ReadAsStringAsync().ConfigureAwait(false).GetAwaiter().GetResult();
+            while (attempts <= maxRetries)
+            {
+                try
+                {
+                    HttpResponseMessage response = client.GetAsync(url).ConfigureAwait(false).GetAwaiter().GetResult();
```
benbp, line 34: I don't know how many changes it would require in the codeowners tool, but in the future it would be nice to just make this whole function async. Not sure why it isn't.

```
                    HttpResponseMessage response = await client.GetAsync(url);
```

benbp (approved):

tools/code-owners-parser/CodeOwnersParser/FileHelpers.cs
JimSuplizio: I agree. It's most likely the fact that making this one method async would cascade through the chain of calling functions that would require things using Codeownes to be async that wouldn't normally be.

tools/code-owners-parser/CodeOwnersParser/FileHelpers.cs
```diff
@@ -22,10 +23,42 @@ public static string GetFileOrUrlContents(string fileOrUrl)
...
         private static string GetUrlContents(string url)
         {
+            int maxRetries = 3;
+            int attempts = 1;
+            int delayTimeInMs = 1000;
             using HttpClient client = new HttpClient();
-            HttpResponseMessage response =
-                client.GetAsync(url).ConfigureAwait(false).GetAwaiter().GetResult();
-            return response.Content.ReadAsStringAsync().ConfigureAwait(false).GetAwaiter().GetResult();
+            while (attempts <= maxRetries)
```
weshaggard, line 30: Might not be necessary in this case but there are some more standard retry patterns for HttpClient, see https://stackoverflow.com/questions/19260060/retrying-httpclient-unsuccessful-requests for examples.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/GitHubEventClient.cs
```diff
@@ -940,5 +945,32 @@ public virtual async Task<List<string>> QueryAILabelService(IssueEventGitHubPayl
...
+        }
+
+        /// <summary>
+        /// Set the config file overrides for codeowners and rulesconfig which will cause them to get pulled
+        /// from the URL instead of requiring a sparse checkout of the configuration directory in order to run.
+        /// </summary>
+        /// <param name="repository">Octkit.Repository from the event payload</param>
+        public void SetConfigEntryOverrides(Repository repository)
+        {
+            CodeOwnerUtils.codeOwnersFilePathOverride = CreateRawGitHubURLForFile(repository, CodeOwnerUtils.CodeownersSubDirectory, CodeOwnerUtils.CodeownersFileName);
```
weshaggard, line 971: Just something to keep in mind this potentially adds more race conditions since these are global static values.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/Utils/RulesConfiguration.cs
```diff
@@ -47,9 +49,17 @@ public RulesConfiguration(string configurationFile = null)
...
+                if (null != rulesConfigFilePathOverride)
+                {
+                    // If the user overrode the location
+                    configLoc = rulesConfigFilePathOverride;
+                }
+                else
+                {
+                    // Load the config from the well known location, somewhere under the .github directory
+                    // which is in the root of the repository
+                    configLoc = DirectoryUtils.FindFileInRepository(RulesConfigFileName, RulesConfigSubDirectory);
```
weshaggard, line 61: Is it expect that folks would still use local files if they want to test new configuration files before updating? I was trying to figure out how we would validate any changes to these files in a language repo. I don't think there is any great way to validate them in the CI but it would be nice to have a way to at least run them through the tool before blindly taking changes to them in the language repo.

PR #6031, opened 2023-04-21
Adjust `Java` asset transition script

scbedd:
@billwert @alzimmermsft 

- [x] Don't include recordingless packages in transition reporting

`408 packages / 1% transitioned` --> `236 packages / 6% transitioned`

[Summary](https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/834/Test-Proxy-Transition-Overview)
[Details](https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/936/Java-Details)

### DIFF
--- a/eng/scripts/python/assets-automation/generate_assets_report.py
+++ b/eng/scripts/python/assets-automation/generate_assets_report.py
@@ -90,6 +90,15 @@ def get_repo(language: str) -> str:
             target_folder,
         ]
         run(command, cwd=generated_folder)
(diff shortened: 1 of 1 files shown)


billwert (approved):

alzimmermsft:
I believe there is still something slightly off as `*-perf` libraries are still being included but they don't contain any `session-records`. But this still makes the view much cleaner than before!

PR #6034, opened 2023-04-24
Create sync template for azure-sdk-tools

JimSuplizio:
eng-common-sync.yml pushes changes out from azure-sdk-tools eng/common directory to the list of repositories defined in the Repos variable of the yml. Because there's going to need to be a similar pipeline to for event processor yml files with a different set of repositories, moving the common code into a template was the way to prevent duplication.

Initially, the pipeline had the default **pool** settings defined at the pipeline level which were used in both CreateSync stage jobs with CreateSyncPRs and VerifyAndMerge setting their own **pool**. Because we can't define the pipeline level pool in the template, define it in the CreateSync jobs which are the only ones that were using it and didn't have pools defined.


weshaggard (approved):
Looks fine assuming that what you copied over exactly matches (except the one pool move which is fine).

JimSuplizio:
> Looks fine assuming that what you copied over exactly matches (except the one pool move which is fine).

Everything is exactly the same with the exception of the pipeline level pool declaration and having to change the sync-directory.yml path from a relative path to an absolute one. The latter being necessary because the template was in a different location.

JimSuplizio:
Since there's no pipeline testing this to appease check-enforcer I'm going to override and merge.

PR #6048, opened 2023-04-25
secrets tool - fix NoWarn for strong name signing

benbp:


benbp:
I don't know why these aren't associating: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2728163&view=results

JimSuplizio (approved):

tools/secret-management/Directory.Build.props
```diff
@@ -6,7 +6,7 @@
         <IsPackable>false</IsPackable>
         <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-        <WarningsNotAsErrors>8002</WarningsNotAsErrors>
+        <NoWarn>CS8002</NoWarn>
```
weshaggard, line 9:
```suggestion
        <NoWarn>$(NoWarn);CS8002</NoWarn>
```

However as you pinged offline this will still cause signing of the library to fail so we should see if we can find another way to do this with a strong name signed assembly.

PR #6086, opened 2023-05-03
Parametrize `ArtifactName` in `detect-pending-changes.yml`

scbedd:
This will be utilized later in [this PR](https://github.com/Azure/azure-sdk-for-python/pull/30173).

We need to honor a different artifact name `packages-extended` instead of `packages`  when submitting APIReview.


praveenkuttappan (approved):
Look good

PR #6088, opened 2023-05-03
Fix logging message in secret tool

weshaggard:


weshaggard:
@benbp looks like you missed updating the public ci pipeline after the rename. I fixed it.

weshaggard:
Well I renamed it the pipeline is still broken because of the strong name issue.

benbp:
> @benbp looks like you missed updating the public ci pipeline after the rename. I fixed it.

Ah good catch, thanks. I only saw the internal one.

benbp:
> Well I renamed it the pipeline is still broken because of the strong name issue.

If we merge in https://github.com/Azure/azure-sdk-tools/pull/6048 it would probably fix the public pipeline at least?

weshaggard:
Fixed in https://github.com/Azure/azure-sdk-tools/pull/6095

PR #6095, opened 2023-05-04
Add additional documentation to Secrets Rotator

hallipr:


tools/secret-management/docs/stores/aad-application-secret.md
```diff
@@ -0,0 +1,18 @@
+# AAD Application Secret
+
+## Implementing Class
```
konrad-jamrozik, line 3: It is easier for me to find quickly what is in the document if if there is an empty line after the heading:

https://github.com/DavidAnson/markdownlint/blob/main/doc/md022.md

weshaggard (approved):

tools/secret-management/docs/stores/aad-application-secret.md
hallipr: I'll fix this with a docs generator in a following PR.

PR #6096, opened 2023-05-04
Stop using the auto-merge label as it no longer does anything.

JimSuplizio:


konrad-jamrozik (approved):

eng/pipelines/templates/stages/archetype-sdk-tool-repo-sync.yml
```diff
@@ -165,20 +165,6 @@ stages:
                         -AuthToken "$(azuresdk-github-pat)"
                       pwsh: true
 
-                  - task: PowerShell@2
```
weshaggard, line 168: At some point if we want to use the real github automerge feature we can replace this with https://github.com/Azure/azure-sdk/blob/main/eng/pipelines/version-updater.yml#L91-L97

JimSuplizio:
I'm going to override check enforcer since there is no pipeline to test this.

PR #6149, opened 2023-05-11
Add exclusions for internal packages

RickWinter:
Exclude internal packages which do not have test recordings
Exclude azcore which does not use test recordings or proxy


scbedd (approved):
Feel free to run + apply changes to existing reports if you wish! Otherwise when I update this on Monday I'll ensure I get your changes 👍

PR #2317, opened 2021-11-18
[Test-Proxy] Header Exclusion Updates

scbedd:
@HarshaNalluru This should address both issues you encountered yesterday afternoon.

And:
example

PR #2185, opened 2021-10-28
[Test-Proxy] Don't match on `x-ms-useragent`

scbedd:
We actually already treat the header `User-Agent` as volatile, so this isn't even a departure from what the test-proxy already does. Just an extension of it.

@HarshaNalluru

PR #6151, opened 2023-05-11
[Test Proxy] Fix typo - Update RecordMatcher.cs

HarshaNalluru:
## History
In order to skip matching the browser's dynamic headers for our recordings, we skipped the headers listed in https://github.com/Azure/azure-sdk-tools/pull/2185#discussion_r751718764

And this was Scott's reaction and the typo crept in
[[Test-Proxy] Header Exclusion Updates by scbedd · Pull Request #2317 · Azure/azure-sdk-tools (github.com)](https://github.com/Azure/azure-sdk-tools/pull/2317)

## Context on how we found it
@jeremymeng was upgrading the puppeteer version and the `"sec-ch-ua"` header mismatch occurred, and @timovv pointed out the list. 
We could've used a sanitizer, but fixing the source feels righteous.


scbedd (approved):
LGTM

[What this header is](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA) for those who are curious.

PR #6152, opened 2023-05-11
[Test Proxy] Add "Accept-Language" header to the ignore list - Update RecordMatcher.cs

HarshaNalluru:
Ignoring the `"Accept-Language"` header

@timovv shared the discomfort about the number of times he had to re-record because his locale is en-CA instead of en-US.

And @jeremymeng 's puppeteer upgrade may unveil a pain later on, example below

> `RecorderError: Unable to find a record for the request POST https://keyvault_name.vault.azure.net/keys/listKeyName-cangettheversionsofakeypaged-/create?api-version=7.4 Header differences: <sec-ch-ua> values differ, request <"HeadlessChrome";v="113", "Chromium";v="113", "Not-A.Brand";v="24">, record <> <Accept-Language> is absent in request, value <en-US> Body differences:`

`sec-ch-ua` is just fixed in #6151

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordMatcher.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordMatcher.cs
@@ -63,7 +63,8 @@ namespace Azure.Sdk.Tools.TestProxy.Common
             "Referrer",
             "Referer",
             "Origin",
-            "Content-Length"
+            "Content-Length",
+            "Accept-Language"
         };
 
         /// <summary>


scbedd (approved):
👍 

Going to talk to Mike about this one.

HarshaNalluru:
Upon discussing more with @scbedd.. 

The issue posted above won't be fixed by adding to the ignore headers, we'll have to exclude it since ignore would mean that, we'll check for its presence.

And excluding feels wrong as there might be real utility.

cc @jeremymeng

scbedd (commented):
Rescinding approval till I chat with Mike.

scbedd (changes requested):
Holding off.

PR #6193, opened 2023-05-17
Add tool link checks in nightly link checks

weshaggard:
We need to add a place to help catch broken links in things like eng/common so that we can fix them before they break more pipelines in the other languages. This will enable us to find them from the nightly run.


benbp (approved):

weshaggard:
Scoping to only eng/common for now as there are a lot of broken links in the rest of the repo that we probably should clean-up but I'm not wanting to tackle that issue currently.

PR #6202, opened 2023-05-18
Adding initial commit of TypeSpecE2E common script

raych1:
This is the script spec:

```
2.2 TypeSpec-Project-Process.ps1

What does this script do?
- fetch tspconfig.yaml from remote if it doesn’t exist locally
- parse tspconfig.yaml
- create an sdk project folder if none exists
- create/update tsp-location.yaml
- call TypeSpec-Project-Sync.ps1
- call TypeSpec-Project-Generate.ps1
input:
- typespecProjectDirectory (required) either a folder of tspconfig.yaml or a remoteUrl of tspconfig.yaml
- commitSha (optional)
- repoUrl (optional)
output: n/a
```

See [here](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/typespec-end-to-end-scenarios.md#four-main-user-scenarios-to-support) for detail e2e spec.


raych1:
@weshaggard  @m-nash , this is the initial commit of new common script, please review.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+  try {
+    Invoke-WebRequest $TypeSpecProjectDirectory -OutFile $tspConfigContent -MaximumRetryCount 3
+  }
+  catch {
+    Write-Host "Failed to download '$TypeSpecProjectDirectory'"
+    Write-Error $_.Exception.Message
+    return
+  }
+  $RepoUrl = $TypeSpecProjectDirectory
+  $regex = "(?<=specification/).*?(?=/tspconfig.yaml)"
```
weshaggard, line 99: We should just put this regex into line 89 match and you can then pull out.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+  else {
+    Write-Host "Missing package-dir in $emitterName options of tspconfig.yaml."
+    exit
+  }
+  return $packageDir
+}
+
+$tspConfigContent = ""
+# example url of tspconfig.yaml: https://github.com/Azure/azure-rest-api-specs-pr/blob/724ccc4d7ef7655c0b4d5c5ac4a5513f19bbef35/specification/containerservice/Fleet.Management/tspconfig.yaml
+if ($TypeSpecProjectDirectory -match '^https://github.com/.*/tspconfig.yaml$') {
```
weshaggard, line 89: For a path like:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml

```suggestion
if ($TypeSpecProjectDirectory -match '^(?<repo>https://github.com/Azure/azure-rest-specs-apis(-pr)?)/specification/(?<path>.*)/tspconfig.yaml$') {
```
you can then get the repo from `$matches["path"]` and the relative path in `$matches["path"]`.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+  }
+  catch {
+    Write-Host "Failed to download '$TypeSpecProjectDirectory'"
+    Write-Error $_.Exception.Message
+    return
+  }
+  $RepoUrl = $TypeSpecProjectDirectory
+  $regex = "(?<=specification/).*?(?=/tspconfig.yaml)"
+  $TypeSpecProjectDirectory = [regex]::Match($TypeSpecProjectDirectory, $regex).Value
+  # TODO get the commit hash from the url
```
weshaggard, line 101: You can tweak the regex more to get the commit/branch as well.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+} else {
+  $TspConfigPath = Join-Path $TypeSpecProjectDirectory "tspconfig.yaml"
+  if (!(Test-Path $TspConfigPath)) {
+    Write-Host "Failed to find tspconfig.yaml in '$TypeSpecProjectDirectory'"
+    return
+  }
+  $tspConfigContent = Get-Content $TspConfigPath -Raw
+}
+$tspConfigYaml = ConvertFrom-Yaml $tspConfigContent
```
weshaggard, line 110: You will need to install the powershell-yaml module to call this function

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+function CreateUpdate-TspLocation() {
+  $serviceDir = ""
+  $additionalDirs = @()
+
+  # Parse tspcofig.yaml to get service-dir, additionalDirectories, and package-dir
+  if ($tspConfigYaml["parameters"] -and $tspConfigYaml["parameters"]["service-dir"]) {
+    $serviceDir = $tspConfigYaml["parameters"]["service-dir"]["default"];
+  }
+  else {
+    Write-Host "Missing service-dir in parameters section of tspconfig.yaml."
```
weshaggard, line 29: Write-Error and we should exit 1 so it is a non-zero exit code so callers will know this failed.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+  # Update tsp-location.yaml
+  $tspLocationYaml["commit"] = $CommitHash
+  $repo = ($RepoUrl -split "/")[3..4] -join "/"
+  $tspLocationYaml["repo"] = $repo
+  $tspLocationYaml["directory"] = $TypeSpecProjectDirectory
+  $tspLocationYaml["additionalDirectories"] = $additionalDirs
+  $tspLocationYaml |ConvertTo-Yaml | Out-File $tspLocationYamlPath
+}
+
+function Get-PackageDir() {
```
weshaggard, line 72: We should pass in the tspConfigYaml as a parameter instead of relying on globals.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+  $TspConfigPath = Join-Path $TypeSpecProjectDirectory "tspconfig.yaml"
+  if (!(Test-Path $TspConfigPath)) {
+    Write-Host "Failed to find tspconfig.yaml in '$TypeSpecProjectDirectory'"
+    return
+  }
+  $tspConfigContent = Get-Content $TspConfigPath -Raw
+}
+$tspConfigYaml = ConvertFrom-Yaml $tspConfigContent
+# call CreateUpdate-TspLocation function
+CreateUpdate-TspLocation
```
weshaggard, line 112: You should pass parameters to this function instead of relying on global variables.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+    New-Item -Path $packageDir -ItemType Directory
+    Write-Host "created package folder $packageDir"
+  }
+
+  # Load tsp-location.yaml if exist
+  $tspLocationYamlPath = Join-Path $packageDir "tsp-location.yaml"
+  $tspLocationYaml = @{}
+  if (Test-Path -Path $tspLocationYamlPath) {
+    $tspLocationYaml = Get-Content -Path $tspLocationYaml -Raw | ConvertFrom-Yaml
+  }
```
weshaggard, line 61: If it doesn't exist you need to initialize a new tspLocationYaml object

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+
+  # Update tsp-location.yaml
+  $tspLocationYaml["commit"] = $CommitHash
+  $repo = ($RepoUrl -split "/")[3..4] -join "/"
```
weshaggard, line 65: What is this trying to do? We should probably do this via a regex so it is clear what we are getting.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+  $tspLocationYaml["repo"] = $repo
+  $tspLocationYaml["directory"] = $TypeSpecProjectDirectory
+  $tspLocationYaml["additionalDirectories"] = $additionalDirs
```
weshaggard, line 68: What type is additionaDirs? Is it an array or a string? Will this work if there is more than one?

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -0,0 +1,115 @@
...
+  $tspLocationYaml |ConvertTo-Yaml | Out-File $tspLocationYamlPath
```
weshaggard, line 69: Probably worth a log message saying we wrote this file.

eng/common/scripts/TypeSpec-Project-Process.ps1
raych1: Yes, it's set to @{} by default and would add other config later.

eng/common/scripts/TypeSpec-Project-Process.ps1
raych1: Agree

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -57,59 +51,85 @@ function CreateUpdate-TspLocation() {
...
+  }
+  else {
+    Write-Host "Parameter 'RepoUrl' has incorrect value: $RepoUrl. It should be similar like 'https://github.com/Azure/azure-rest-api-specs'"
+    exit 1
+  }
+}
+
+$repoRootPath =  (Join-Path $PSScriptRoot .. .. ..)
+$repoRootPath = Resolve-Path $repoRootPath
+$repoRootPath = $repoRootPath -replace "\\", "/"
```
weshaggard, line 103: Is this slash replacement needed? I wouldn't expect so give Join-Path below should work.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -57,59 +51,85 @@ function CreateUpdate-TspLocation() {
...
+$tspConfigPath = Join-Path $repoRootPath 'tspconfig.yaml'
 # example url of tspconfig.yaml: https://github.com/Azure/azure-rest-api-specs-pr/blob/724ccc4d7ef7655c0b4d5c5ac4a5513f19bbef35/specification/containerservice/Fleet.Management/tspconfig.yaml
-if ($TypeSpecProjectDirectory -match '^https://github.com/.*/tspconfig.yaml$') {
+if ($TypeSpecProjectDirectory -match '^https://github.com/(?<repo>Azure/azure-rest-api-specs(-pr)?)/blob/(?<commit>[0-9a-f]{40})/(?<path>.*)/tspconfig.yaml$') {
   try {
-    Invoke-WebRequest $TypeSpecProjectDirectory -OutFile $tspConfigContent -MaximumRetryCount 3
+    $TypeSpecProjectDirectory = $TypeSpecProjectDirectory -replace "github.com", "raw.githubusercontent.com"
```
weshaggard, line 108: See https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Generate-ReleaseNotes.ps1#L24 where I do a similar replacement in one line.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -57,59 +51,85 @@ function CreateUpdate-TspLocation() {
...
+  }
+}
+
+$repoRootPath =  (Join-Path $PSScriptRoot .. .. ..)
+$repoRootPath = Resolve-Path $repoRootPath
+$repoRootPath = $repoRootPath -replace "\\", "/"
+$tspConfigPath = Join-Path $repoRootPath 'tspconfig.yaml'
 # example url of tspconfig.yaml: https://github.com/Azure/azure-rest-api-specs-pr/blob/724ccc4d7ef7655c0b4d5c5ac4a5513f19bbef35/specification/containerservice/Fleet.Management/tspconfig.yaml
-if ($TypeSpecProjectDirectory -match '^https://github.com/.*/tspconfig.yaml$') {
+if ($TypeSpecProjectDirectory -match '^https://github.com/(?<repo>Azure/azure-rest-api-specs(-pr)?)/blob/(?<commit>[0-9a-f]{40})/(?<path>.*)/tspconfig.yaml$') {
```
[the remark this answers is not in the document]
weshaggard: I see you have a TODO below to support that. If you don't add that support initially, please be sure to file an issue for it.

eng/common/scripts/TypeSpec-Project-Process.ps1
```diff
@@ -57,59 +51,85 @@ function CreateUpdate-TspLocation() {
...
+  else {
+    Write-Host "Parameter 'RepoUrl' has incorrect value: $RepoUrl. It should be similar like 'https://github.com/Azure/azure-rest-api-specs'"
+    exit 1
+  }
+}
+
+$repoRootPath =  (Join-Path $PSScriptRoot .. .. ..)
+$repoRootPath = Resolve-Path $repoRootPath
+$repoRootPath = $repoRootPath -replace "\\", "/"
+$tspConfigPath = Join-Path $repoRootPath 'tspconfig.yaml'
```
weshaggard, line 104: Is this correct? Do we ever have a case where the tspconfig.yaml is in the repo root?

weshaggard: Looks like you are only using it in the download case. I'd suggest not putting that into the repo root as that will add a file to git which we likely don't want. We should think about another location and should probably scope it in the if block below.

weshaggard (approved):
I left a few more comments but nothing that is blocking. We should also update the docs TypeSpec-Project-Scripts.md file with information about this script.

raych1:
> I left a few more comments but nothing that is blocking. We should also update the docs TypeSpec-Project-Scripts.md file with information about this script.

@weshaggard , I will fix the comments in the new PR, thanks.

weshaggard:
@raych1 it looks like you didn't follow the outlined approach for syncing this changes https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/README.md#workflow. As a result your new PR is failing. I'll try to fix the existing issue but lets try to follow the process this time.

PR #6211, opened 2023-05-19
Update stress workbook parameters to filter by time->namespace->pod

benbp:


tools/stress-cluster/cluster/azure/monitoring/stress-test-workbook.bicep
```diff
@@ -86,19 +53,82 @@ var workbookContent = {
...
+            resourceType: 'microsoft.operationalinsights/workspaces'
+            value: null
           }
         ]
         style: 'pills'
         queryType: 0
         resourceType: 'microsoft.operationalinsights/workspaces'
       }
-      name: 'parameters - 1'
+      name: 'parameters - 5'
```
ckairen, line 131: Remind me again what the parameters - 5 stands for?

benbp: I think this is just a generated number for "you didn't specify the parameter name" and it happens to be the 5th one (because I created 3 new ones and was developing with two pre-existing ones).

ckairen:
and do we not want the kube-system and stress-infra namespaces to be excluded anymore? can we get valuable info from those

benbp:
> and do we not want the kube-system and stress-infra namespaces to be excluded anymore? can we get valuable info from those

@ckairen They are effectively excluded still because you now choose a namespace and we filter against that. The new namespace dropdown parameter does exclude kube-system (query below), but I don't exclude stress-infra because it could be nice to view our service stats from this dashboard.

```
KubePodInventory\r\n| distinct Namespace\r\n| where Namespace !in ("kube-system", "kubernetes-dashboard", "gatekeeper-system")
```

ckairen (approved):

PR #6214, opened 2023-05-22
[http-fault-injector] Set UseAppHost=false

mikeharder:
- Enables HTTPS on Mac with no keychain popup
- Workaround for https://github.com/dotnet/sdk/issues/22544


mikeharder:
Closing in favor of https://github.com/Azure/azure-sdk-tools/pull/6216.  The changes in this PR work when using `dotnet run` from the source tree on MacOS, but they don't work after the tool has been packed and installed.

PR #6216, opened 2023-05-22
[http-fault-injector] Target net7.0 to enable HTTPS on Mac with no keychain popup

mikeharder:
- Fixed in https://github.com/dotnet/aspnetcore/issues/41879

### DIFF
--- a/tools/http-fault-injector/Azure.Sdk.Tools.HttpFaultInjector/Azure.Sdk.Tools.HttpFaultInjector.csproj
+++ b/tools/http-fault-injector/Azure.Sdk.Tools.HttpFaultInjector/Azure.Sdk.Tools.HttpFaultInjector.csproj
@@ -3,9 +3,7 @@
   <PropertyGroup>
     <VersionPrefix>0.2.0</VersionPrefix>
 
-    <!--
-      net7.0 enables HTTPS on Mac with no keychain popup (https://github.com/dotnet/aspnetcore/issues/41879)
-    -->
+    <!-- net7.0 enables HTTPS on Mac with no keychain popup (https://github.com/dotnet/aspnetcore/issues/41879) -->
     <TargetFramework>net7.0</TargetFramework>
 
     <IsPackable>true</IsPackable>
--- a/tools/http-fault-injector/Azure.Sdk.Tools.HttpFaultInjector/Azure.Sdk.Tools.HttpFaultInjector.csproj
+++ b/tools/http-fault-injector/Azure.Sdk.Tools.HttpFaultInjector/Azure.Sdk.Tools.HttpFaultInjector.csproj
@@ -3,7 +3,10 @@
   <PropertyGroup>
     <VersionPrefix>0.2.0</VersionPrefix>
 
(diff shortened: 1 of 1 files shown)


mikeharder:
@weshaggard: Are you OK with `http-fault-injector` targeting `net7.0`?  I know we are trying to make all our tools consistent and target `net6.0`, but the easiest way to avoid this problem is upgrading to `net7.0`.  `http-fault-injector` doesn't have many users, and I'm fine with forcing them to use .NET 7.

mikeharder:
> I don't have a concern if we have a good reason. Do we have people hitting this issue that need this fix? If so will we have this issue in other tools like Test Proxy?

Alan is trying to use http-fault-injector in a pipeline that runs on MacOS.  I will need to investigate why test-proxy isn't hitting this on Mac.  There might be a slight difference in the tool code itself, or in how the tools are being launched.

benbp:
> @weshaggard: Are you OK with `http-fault-injector` targeting `net7.0`? I know we are trying to make all our tools consistent and target `net6.0`, but the easiest way to avoid this problem is upgrading to `net7.0`. `http-fault-injector` doesn't have many users, and I'm fine with forcing them to use .NET 7.

We should aim to publish standalone executables for all our tools, then we don't have this problem where users are forced to have the right NET installed. It's a usability nightmare getting 6 and 7 installed side by side on linux, for example.

mikeharder:
> > @weshaggard: Are you OK with `http-fault-injector` targeting `net7.0`? I know we are trying to make all our tools consistent and target `net6.0`, but the easiest way to avoid this problem is upgrading to `net7.0`. `http-fault-injector` doesn't have many users, and I'm fine with forcing them to use .NET 7.
> 
> We should aim to publish standalone executables for all our tools, then we don't have this problem where users are forced to have the right NET installed. It's a usability nightmare getting 6 and 7 installed side by side on linux, for example.

@benbp: Do you know what needs to change in `HttpFaultInjector.csproj` to enable this?  Do we have existing examples of publishing standalone executables in the tools repo?

benbp:
@mikeharder here is how proxy does it. @scbedd did a lot of work to get our tools dotnet pipeline able to sign mac binaries, etc.

https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/ci.yml#L52-L68

scbedd:
> @mikeharder here is how proxy does it. @scbedd did a lot of work to get our tools dotnet pipeline able to sign mac binaries, etc.
> 
> https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/ci.yml#L52-L68
> 
> [edit] in the csproj, probably these: https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Azure.Sdk.Tools.TestProxy.csproj#L6-L8

To be clear though, when you package for standalone the executable becomes the actual namespace. so `Azure.Sdk.Tools.TestProxy` is the standalone executable regardless of `<packastool>` or not.

I agree with the suggestion to just add the standalone matrix. That _should_ "just work" ™️

PR #6217, opened 2023-05-23
Add differentiation between a null body and an empty body 

scbedd:
Before this PR, a user could start a recording with a a target file, or _no_ target file. No target file being provided means the recording should be entirely in-memory.

```jsonc
// POST /Record/Playback
// body
{
  "x-recording-file": "blah.json"
}
// returns x-recording-id for use by recording framework
```

```jsonc
// POST /Record/Playback
// body
{
}
// returns x-recording-id for use by recording framework
```

@samvaity recently was debugging a proxy start, and while the issue ended up being that the request body wasn't being formed correctly, the test-proxy didn't _inform_ Sameeksha about that problem. To the test-proxy, @samvaity was just starting an in-memory recording.

The perf framework currently sends _no_ body when attempting to start an in-memory recording. We will take advantage of that fact to ship a breaking change. If you want an in-memory recording, _send no body_.

This change allows us to easily bifurcate these behaviors, and give our users some actual detail to their errors!

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Playback.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Playback.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 using Azure.Sdk.Tools.TestProxy.Common;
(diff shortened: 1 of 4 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Playback.cs
```diff
@@ -47,7 +52,7 @@ public async Task Start()
             }
             else
             {
-                throw new HttpException(HttpStatusCode.BadRequest, "At least one of either JSON body key 'x-recording-file' or header 'x-recording-id' must be populated when starting playback.");
+                throw new HttpException(HttpStatusCode.BadRequest, "When starting playback, a body must be provided with least one JSON body key 'x-recording-file' or header 'x-recording-id' must be populated when starting playback.");
```
scbedd, line 55:
```suggestion
                throw new HttpException(HttpStatusCode.BadRequest, "At least one of either JSON body key 'x-recording-file' or header 'x-recording-id' must be populated when starting playback.");
```

benbp (approved):

PR #6244, opened 2023-05-25
Use test registry in npmrc for alpha packages

hallipr:
Currently, the Autorest Regen Preview pipeline for autorest.csharp publishes alpha versions of the c# emitter package to an alternate package feed. This is breaking the validation builds of azure-sdk-for-net because the alpha version specified doesn't exist in the default public repository.

To fix this, if we detect an alpha package version number in emitter-package.json, we create an npmrc file prior to running the code generation check.

### DIFF
--- a/eng/common/scripts/TypeSpec-Project-Generate.ps1
+++ b/eng/common/scripts/TypeSpec-Project-Generate.ps1
@@ -45,6 +45,14 @@ function NpmInstallForProject([string]$workingDirectory) {
 
         Write-Host("Copying package.json from $replacementPackageJson")
         Copy-Item -Path $replacementPackageJson -Destination "package.json" -Force
+
+        $useAlphaNpmRegistry = (Get-Content $replacementPackageJson -Raw).Contains("-alpha.")
+
+        if($useAlphaNpmRegistry) {
+            Write-Host "Package.json contains '-alpha.' in the version, Creating .npmrc using public/azure-sdk-for-js-test-autorest feed."
+            "registry=https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js-test-autorest/npm/registry/ `n`nalways-auth=true" | Out-File '.npmrc'
+        }
+        
         npm install --no-lock-file
         if ($LASTEXITCODE) { exit $LASTEXITCODE }
     }


hallipr:
To verify, I ran in azure-sdk-for-net using these pipelines:

Should pass because doesn't use alpha package:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=2799062&view=results

Should fail because uses alpha package and didn't regenerate code in branch:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=2799252&view=results

Should pass because uses alpha package and did regenerate code in branch:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=2801691&view=results

m-nash (approved):

eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -45,6 +45,14 @@ function NpmInstallForProject([string]$workingDirectory) {
 
         Write-Host("Copying package.json from $replacementPackageJson")
         Copy-Item -Path $replacementPackageJson -Destination "package.json" -Force
+
+        $useAlphaNpmRegistry = (Get-Content $replacementPackageJson -Raw).Contains("-alpha.")
```
weshaggard, line 49: I'm not sure but I suspect this could cause issues if we have a mix of alpha and non-alpha versions in our package.json file as npm doesn't allow you to restore from 2 feeds. @mikeharder can correct me if I'm wrong.

mikeharder: npm only supports one registry per "scope":

https://docs.npmjs.com/cli/v9/using-npm/scope#associating-a-scope-with-a-registry

@hallipr: What package are we publishing to the DevOps feed?  And do we consume any other packages from the same scope?

hallipr: We publish `autorest/csharp` and `azure-tools/typespec-csharp`, but only `azure-tools/typespec-csharp` is included in `emitter-package.json`

The feed uses npm.org as an upstream for all other packages.

weshaggard: @hallipr have we tested this upstream behavior? I know we have had issues with it in other contexts so I do suggest you testing that out end-to-end in a CI machine which likely has different permissions then you would doing it locally.

hallipr: It was tested in autorest.csharp -> azure-sdk-for-net validation builds before moving the change to azure-sdk-tools.
https://dev.azure.com/azure-sdk/public/_build/results?buildId=2801691&view=results

PR #6245, opened 2023-05-25
Add Sync-AzsdlLabels.ps1 script

heaths:


heaths:
I tested this with a couple spcific repos and seemed to work just fine, but have not run it for all repos yet.

benbp (approved):

tools/github/scripts/Sync-AzsdkLabels.ps1
```diff
@@ -0,0 +1,82 @@
...
+
+# Filter out the source repository.
+$Repositories = $Repositories.Where({$_ -ne $SourceRepository})
+
+foreach ($repo in $Repositories) {
+    if ($Force -or $PSCmdlet.ShouldProcess(
+        "Cloning labels from $SourceRepository to $repo",
+        "Clone labels from $SourceRepository to $repo?",
+        "Clone labels")) {
+        $result = gh -R "$repo" label clone "$SourceRepository" --force 2>&1
```
heaths, line 53: The `--force` here is to allow overwriting labels. If, for example, the description or color varies, it will be made consistent. I considered tying this to the scripts `-Force`, but that's really more for `ShouldProcess` to skip confirmation depending on the user's `$ConfirmPreference`. I could, though.

heaths: Offline, @benbp was fine with this.

heaths:
There's no other pipelines to run, but check-enforcer was still waiting for...nothing?

tools/github/scripts/Sync-AzsdkLabels.ps1
```diff
@@ -0,0 +1,82 @@
...
+        if ($LASTEXITCODE) {
+            Write-Error "Failed to clone labels from $SourceRepository to ${repo}: $result"
+        }
+    }
+}
+
+<#
```
konrad-jamrozik, line 60: @heaths shouldn't docs be first, by convention? I've never seen docs being at the bottom.

tools/github/scripts/Sync-AzsdkLabels.ps1
```diff
@@ -0,0 +1,82 @@
...
+
+if ($PSCmdlet.ParameterSetName -eq 'RepositoryFile') {
+    $Repositories = Get-Content $RepositoryFilePath
+}
+
+# Filter out the source repository.
+$Repositories = $Repositories.Where({$_ -ne $SourceRepository})
+
+foreach ($repo in $Repositories) {
+    if ($Force -or $PSCmdlet.ShouldProcess(
```
konrad-jamrozik, line 49: @heaths we do have `NeedsARMTriage` custom label in `azure-sdk-tools` repo. Please do not delete it :D

heaths: `gh label clone` is non-destructive. It does upserts only.

tools/github/scripts/Sync-AzsdkLabels.ps1
heaths: I always put it at the bottom. It's not the important stuff when you want to read/edit the code. You should be looking at it with `help <cmdlet|script>` to better render it anyway.

konrad-jamrozik: I would argue that natural language description in comments is easier to comprehend, and one should delve into the code only if they need deeper insight / more details. But to each their own, I guess.

heaths: The `help` (`get-help`) cmdlet should be their first recourse for getting help. It's meant to be universal. If they have to open the script to get help, I argue the damage was already done. And if script writers are using non-doc-style comments that don't generate MAML (explicitly or implicitly), that's on them. ;)

tools/github/scripts/Sync-AzsdkLabels.ps1
heaths: If you're interested in the code, see https://github.com/cli/cli/blob/b292dc43b25183024fb12b78459e0d940f524027/pkg/cmd/label/clone.go and https://github.com/cli/cli/blob/b292dc43b25183024fb12b78459e0d940f524027/pkg/cmd/label/create.go#L128

PR #6270, opened 2023-05-31
Fix Failing Proxy LiveTests

scbedd:
@samvaity if I don't want these to explode in runtime, so it's a good motivator to fix the race condition :)

### DIFF
--- a/tools/test-proxy/tests.yml
+++ b/tools/test-proxy/tests.yml
@@ -257,7 +257,6 @@ stages:
             git clone https://github.com/$(REPO) $(CLONE_LOCATION) --depth 1
           displayName: Clone Repo
 
-
         - pwsh: |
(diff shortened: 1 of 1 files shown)


samvaity (approved):

tools/test-proxy/tests.yml
```diff
@@ -270,6 +269,17 @@ stages:
...
 
+        # we want the proxy available so we can restore recordings and avoid the race condition error
+        # however, we also want to not run it such that the local executable is used by the java tests.
+        - template: /eng/pipelines/templates/steps/test-proxy-local-tool.yml
+          parameters:
+            runProxy: false
+
+        - pwsh: |
+            Get-ChildItem -Recurse -Path $(CLONE_LOCATION)/sdk/ -Filter assets.json | `
+              ForEach-Object { test-proxy restore -a $_.FullName }
```
benbp, line 280: @scbedd probably useful to handle LASTEXITCODE here as well

PR #6271, opened 2023-06-01
Updated doc about typespec common scripts

raych1:


raych1:
@weshaggard , can you review this docs PR?

PR #6274, opened 2023-06-01
Add openapi hub redirect doc landing page

benbp:


maririos (approved):

weshaggard (approved):
Looks good. I'd also suggest we create an aka.ms link for the usages to this page so we can more easily move that later if needed.

benbp:
> Looks good. I'd also suggest we create an aka.ms link for the usages to this page so we can more easily move that later if needed.

Are you thinking as an auxiliary link, or one that we can point to from the DNS/front door? If the latter, it's also pretty easy to just change the front door endpoint to point anywhere we want as well.

PR #6300, opened 2023-06-07
Add fallback for the test-proxy command in migration script

scbedd:
A couple languages are using the `standalone` exe primarily. That is not a tool, and as such the exe is named for the namespace of the package.

To account for that, we are updating the migration script to handle the proxy executable.

@samvaity for FYI


eng/common/testproxy/transition-scripts/generate-assets-json.ps1
```diff
@@ -348,9 +355,29 @@ $language = Get-Repo-Language
...
+    }
+
+    $standalonePresent = Test-Exe-In-Path -ExeToLookFor $StandaloneProxyExe -ExitOnError $false
+
+    if ($standalonePresent) {
+      Write-Host "Default proxy exe $TestProxyExe is not present, but standalone tool $StandaloneProxyExe is. Updating proxy exe to standalone version."
+      $TestProxyExe = $StandaloneProxyExe
+    }
+    else {
+      Write-Error "The user has selected option InitialPush to push their assets, neither $TestProxyExe nor StandaloneProxyExe are installed on this machine."
```
konrad-jamrozik, line 375:
```suggestion
      Write-Error "The user has selected option InitialPush to push their assets, neither $TestProxyExe nor $StandaloneProxyExe are installed on this machine."
```

konrad-jamrozik: When `$StandaloneProxyExe` unrolls tot the value I think it might not be immediately evident the path points to standalone proxy exe, so it might be worth prefixing it, like `neither test-proxy exe $TestProxy nor standalone test-proxy exe $StandaloneProxyExe`

Another minor nit: for symmetry I would rename $StandaloneProxyExe to $StandaloneTestProxyExe

konrad-jamrozik (approved):

eng/common/testproxy/transition-scripts/generate-assets-json.ps1
```diff
@@ -348,9 +355,29 @@ $language = Get-Repo-Language
...
+    $StandaloneProxyExe = "Azure.Sdk.Tools.TestProxy"
+
+    if ($IsWindows) {
+      $StandaloneProxyExe += ".exe"
+    }
+
+    $standalonePresent = Test-Exe-In-Path -ExeToLookFor $StandaloneProxyExe -ExitOnError $false
+
+    if ($standalonePresent) {
+      Write-Host "Default proxy exe $TestProxyExe is not present, but standalone tool $StandaloneProxyExe is. Updating proxy exe to standalone version."
```
samvaity, line 371:
```suggestion
      Write-Host "Default proxy exe $TestProxyExe is not present, but standalone tool $StandaloneProxyExe is. Updating proxy exe to use the standalone version."
```

samvaity (approved):

eng/common/testproxy/transition-scripts/generate-assets-json.ps1
```diff
@@ -348,9 +355,29 @@ $language = Get-Repo-Language
 # in the path and that we're able to map the language's recording
 # directories
 if ($InitialPush) {
-  Test-Exe-In-Path -ExeToLookFor $TestProxyExe
+  $proxyPresent = Test-Exe-In-Path -ExeToLookFor $TestProxyExe -ExitOnError $false
+
+  # try to fall back 
+  if (-not $proxyPresent) {
+    $StandaloneTestProxyExe = "Azure.Sdk.Tools.TestProxy"
```
benbp, line 362: I think I missed this. Why is this the standalone tool name and not `test-proxy`? It strikes me as pretty odd that a cli tool would be named like this. Is the dotnet tool install also using this name or still `test-proxy` (and if the latter I think we want our tool names to be consistent from the docs).

scbedd: No, when you produce a standalone executable it's named for the namespace of the project.

I can change it, but then it would make the produced dll `test-proxy.dll`, which I do not want. When I last dug into this to produce the signed standalone executables, that was all that I could find.

scbedd: @benbp I'm going to merge this to unblock the java folks, then I'm going to make the standalone just named `test-proxy`.

The latter one is going to be a fun update to `azure-sdk-build-tools` (there is a reason I didn't do it at first 😂) but you're absolutely right about your original point so...

PR #6303, opened 2023-06-07
Asset-Sync Documentation Update

scbedd:
@konrad-jamrozik  you said that we never really specified what an `asset` was in another PR. I thought you were right right up until I read the intro to the doc

```
# Asset Sync (Retrieve External Test Recordings)

The `test-proxy` optionally offers integration with other git repositories for **storing** and **retrieving** recordings. This enables the proxy to work against repositories that do not emplace their test recordings directly alongside their test implementations.
```

I feel like this is extremely clear about  `asset-sync` does, but I never actually tied it to the keyword `assets`. This PR does that for clarity 👍

### DIFF
--- a/tools/test-proxy/documentation/asset-sync/README.md
+++ b/tools/test-proxy/documentation/asset-sync/README.md
@@ -2,6 +2,8 @@
 
 The `test-proxy` optionally offers integration with other git repositories for **storing** and **retrieving** recordings. This enables the proxy to work against repositories that do not emplace their test recordings directly alongside their test implementations.
 
+Colloquially, any file that is stored externally using the `asset-sync` feature of the `test-proxy` is called an `asset`.
+
 ![asset sync block diagram](../_images/asset_sync_block_diagram.png)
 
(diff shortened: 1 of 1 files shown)


konrad-jamrozik:
I mean, yeah, I agree one could have deduced from the docs what assets are, but I had some remarks about the docs talking about assets for a bit before the concept is introduced and then still one having to go through a bit of indirection to figure it out. Basically, I was looking for a like a noun-oriented dictionary / glossary of terms. I think the sentence you added in this PR helps a lot. Now if one searches for `asset` they will land on that sentence and any guesswork will be squashed 👌🏻

konrad-jamrozik (approved):

scbedd:
> but I had some remarks about the docs talking about assets for a bit before the concept is introduced and then still one having to go through a bit of indirection to figure it out.

Absolutely, which is why I added this line item here. Short-circuiting is always better.

Adding a `glossary` of terms is a great suggestion and I'll add that as asset-sync readme section.

PR #6305, opened 2023-06-08
Added switch to skip sync and generate script call

raych1:
Per the agreement with @m-nash, adding this switch in `Process` script to be used by dotnet automation script.

With this switch, dotnet can keep most of current script code when adopting new common scripts. Dotnet would also keep using `dotnet build /t:GenerateCode` to generate sdk code.  

In CI pipeline we have consistent experience, `TypeSpec-Project-Process.ps1` would be called by each language automation script to interact with `tspconfig.yaml` and create/update `tsp-location.yaml`. 

@weshaggard , can you review?


weshaggard (approved):
I don't mind having the option that can be used but I'm not sure I exactly understand why it is needed. Aren't the existing scripts already calling sync and generate? If so, why wouldn't they just let the process script handle that.

In any case the biggest value is to have a common script to parse the tspconfig and create/update the tsp-location file. So as long as we not duplicating that and parsing the tspconfig in any other place I'm good with this.

raych1:
> I don't mind having the option that can be used but I'm not sure I exactly understand why it is needed. Aren't the existing scripts already calling sync and generate? If so, why wouldn't they just let the process script handle that.
> 
> In any case the biggest value is to have a common script to parse the tspconfig and create/update the tsp-location file. So as long as we not duplicating that and parsing the tspconfig in any other place I'm good with this.

@weshaggard , `dotnet build` command invokes sync and generate script and we want to keep this rather than switching to call process script.  Instead, `Process` script would be called in the dotnet automation script to interact with `tspconfig.yaml` and create/update `tsp-location.yaml`.

weshaggard:
>@weshaggard , dotnet build command invokes sync and generate script and we want to keep this rather than switching to call process script. Instead, Process script would be called in the dotnet automation script to interact with tspconfig.yaml and create/update tsp-location.yaml.

@raych1 how would they know which projects to invoke the dotnet build command on? Mostly I want to avoid needing to read the tspconfig.yaml file directly in any other places.

PR #6330, opened 2023-06-13
Allow `Restore` to discard pending changes

scbedd:
The original thought is that if you had pending changes, but wanted to swap to a different checked out tag for an `assets.json`, you would be **prevented** from doing so with

> you have pending changes

So that folks can avoid discarding recordings. However, because a lot of these are running behind a shim or some other surrounding script, folks don't see the proxy output clearly. Because they can't see what's going wrong with the proxy, they don't know what to try next.

This PR will eliminate this scenario.

CC @mssfang @samvaity  for FYI

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
@@ -162,7 +162,7 @@ namespace Azure.Sdk.Tools.TestProxy.Store
                 InitializeAssetsRepo(config);
             }
 
-            CheckoutRepoAtConfig(config);
+            CheckoutRepoAtConfig(config, true);
             await BreadCrumb.Update(config);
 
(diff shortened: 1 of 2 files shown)


mikeharder (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
```diff
@@ -297,12 +300,23 @@ private void HideOrigin(GitAssetsConfiguration config)
...
                 return;
             }
+            // if we are NOT on our targeted tag, before we attempt to switch we need to reset without asking for permission
+            else
+            {
+                if (cleanEnabled)
+                {
+                    Clean(config);
+                }
+            }
```
benbp, line 319:
```suggestion
            else if (cleanEnabled)
            {
                Clean(config);
            }
```

benbp (approved):

PR #6343, opened 2023-06-14
Replace secret rotation --expiring option with --force

hallipr:
The current default behavior is to rotate all secrets, including those that are not expiring, unless the --expiring option is used.

This change replaces that with a default behavior of only rotating expiring secrets unless the --force options is used, which would then also rotate the secrets that are not yet expiring.

### DIFF
--- a/tools/secret-management/Azure.Sdk.Tools.SecretManagement.Cli/Commands/RotateCommand.cs
+++ b/tools/secret-management/Azure.Sdk.Tools.SecretManagement.Cli/Commands/RotateCommand.cs
@@ -1,4 +1,4 @@
-﻿using System.CommandLine;
+using System.CommandLine;
 using System.CommandLine.Invocation;
 using Azure.Sdk.Tools.SecretRotation.Configuration;
 using Azure.Sdk.Tools.SecretRotation.Core;
@@ -7,19 +7,19 @@ namespace Azure.Sdk.Tools.SecretManagement.Cli.Commands;
 
 public class RotateCommand : RotationCommandBase
 {
-    private readonly Option<bool> expiringOption = new(new[] { "--expiring", "-e" }, "Only rotate expiring secrets");
+    private readonly Option<bool> forceOption = new(new[] { "--force", "-f" }, "Force rotation of secrets outside of their expiration window.");
     private readonly Option<bool> whatIfOption = new(new[] { "--dry-run", "-d" }, "Preview the changes that will be made without submitting them.");
 
     public RotateCommand() : base("rotate", "Rotate one, expiring or all secrets")
     {
-        AddOption(this.expiringOption);
+        AddOption(this.forceOption);
         AddOption(this.whatIfOption);
     }
 
     protected override async Task HandleCommandAsync(ILogger logger, RotationConfiguration rotationConfiguration,
         InvocationContext invocationContext)
     {
-        bool onlyRotateExpiring = invocationContext.ParseResult.GetValueForOption(this.expiringOption);
+        bool force = invocationContext.ParseResult.GetValueForOption(this.forceOption);
         bool whatIf = invocationContext.ParseResult.GetValueForOption(this.whatIfOption);
 
         var timeProvider = new TimeProvider();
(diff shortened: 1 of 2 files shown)


konrad-jamrozik:
I think `force` flag may be a little bit confusing. `force` in my mind means: "do what I want to do in spite of any concerns" but here we are using it as "do more things". I think something like `rotate --all` would be possibly more intuitive.

As a side note: given secret rotation is hard to revert, I think it would be beneficial if by default the tool doesn't rotate anything, but instead would tell me what it is going to rotate.

I am envisioning the most usable session snippet would look like this to me (psuedo-output)

``` txt
> rotate
Discovered 17 secrets, 3 of them expiring. Going to rotate the 3 expiring secrets:
- secret1: expires at dd/mm/yyyy (in X days)
- secret2: expires at dd/mm/yyyy (in X days)
- secret3: expires at dd/mm/yyyy (in X days)

To actually conduct the rotations, run 'rotate --apply'. To see all secrets, including non-expiring, run 'rotate --listAll'. To rotate all secrets, including the non-expiring, run 'rotate --all'.

> rotate --all
Discovered 17 secrets, 3 of them expiring. Going to rotate all the 17 secrets:
- secret1: expires at dd/mm/yyyy (in X days)
- secret2: expires at dd/mm/yyyy (in X days)
- secret3: expires at dd/mm/yyyy (in X days)
- secret4: expires at dd/mm/yyyy (in X days) (this secret is not expiring soon)
- secret5: expires at dd/mm/yyyy (in X days) (this secret is not expiring soon)
...
To actually conduct the rotations, run 'rotate --all --apply'.

> rotate --all --apply
Discovered 17 secrets, 3 of them expiring. Rotating all the 17 secrets. Progress:
- secret 1: expires at dd/mm/yyyy (in X days)
- secret1: expires at dd/mm/yyyy (in X days)
- secret2: expires at dd/mm/yyyy (in X days)
- secret3: expires at dd/mm/yyyy (in X days)
- secret4: expires at dd/mm/yyyy (in X days) (this secret is not expiring soon)
- secret5: expires at dd/mm/yyyy (in X days) (this secret is not expiring soon)
...

rotating secret 1... DONE. New expiration date: dd/mm/yyyy (in Y days)
rotating secret 2... DONE. New expiration date: dd/mm/yyyy (in Y days)
...
```

konrad-jamrozik (approved):

hallipr:
@konrad-jamrozik, for `--listAll`, that's already included in the `secrets list` command.

`secrets list` lists all the plans
`secrets status` checks the status of the plans and groups them into `Expired`, `Expiring` and `Up-to-date`
`secrets rotate` rotates the secrets that aren't `up-do-date`, with an optional --dry-run option to just emit log statements instead of performing any write action.  This doesn't just list the secrets that would be rotated but also includes logs for the individual actions taken during rotation.

We could either invert the --dry-run option, converting it to --apply, or we could add an additional --apply option.

I'm in favor of the additional option --apply, so there's additive behavior from`secrets rotate` to `secrets rotate --dry-run` to `secrets rotate --apply`.

There would be no difference between `--dry-run --apply` and `--dry-run`, but I don't see a need to prevent both.

benbp:
I don't think it's necessary to add a flag `--apply` when we already have `--dry-run`. It seems unconventional to me to have a base command have dry run behavior only. If we're concerned about preventing mistakes, then I think an interactive confirmation after listing secrets would be better and more consistent with general CLI behavior (e.g. azure cli) than adding another flag. Additionally, if some of our cli tools have dry run behavior as the default, people may come to expect that other tools also have dry run behavior by default, and run commands that are actually destructive.

My preference here is `--rotate-all` or `--all` instead of `--force` and keeping the remaining behavior change of this PR (default to expiring only).

konrad-jamrozik:
Yeah, I think it is best to stick with the existing conventions while still adding interactive confirmation by default, as @benbp outlined. Then I guess we would need some flag that bypasses the interactive prompt, to be used in CI scenarios. Something like `--force` or `--non-interactive`.

benbp:
Or `-y/--yes`

PR #6347, opened 2023-06-14
Add Support custom subscription config/creds for stress tests

benbp:
This PR adds support for "bring your own subscription" when deploying test resources. This will enable a new us to bill deployed resources to a separate team that owns their own test subscription (in this case expensive storage and cosmos test resources).

The process is very similar to how we do this for live testing. We set up a new secret for a partner team which they can update with service principal credentials and a subscription id to their test subscription. Then additionally they can specify which subscription config they would like to target on a per-scenario or all-scenario basis in scenario matrix config.

See the README updates in this PR for more details on the process.


tools/stress-cluster/cluster/kubernetes/stress-test-addons/templates/stress-test-static-secret-provider.tpl
```diff
@@ -0,0 +1,37 @@
...
+{{- $global := . }}
+{{- $subChart := get $global.Values "stress-test-addons" }}
+{{- $subConfigs := list (get $subChart.subscription $subChart.env) }}
+{{- range $global.Values.scenarios }}
+{{- $subConfigs = append $subConfigs (coalesce .subscriptionConfig "") }}
+{{- end }}
+{{- $subConfigs = compact $subConfigs | uniq }}
+{{- range $subConfigs }}
+---
+apiVersion: secrets-store.csi.x-k8s.io/v1
```
benbp, line 13: This file/resource is pretty similar to the yaml file deleted in this PR (I can't get git to show a diff as a moved file). The main changes is the SecretProviderClass is specified in a loop for all possible `subscriptionConfiguration` values found in the matrix, so that we download all possible secrets referenced.

Additionally, the SecretProviderClass resource now has a dynamic naming convention that includes the secret name. This way individual scenario tests just reference their respective subscription config from the new constant `{{ .Stress.SubscriptionConfig }}` via the template, so they only load the appropriate config and don't clobber with any of the others.

tools/stress-cluster/cluster/kubernetes/stress-test-addons/templates/stress-test-static-secret-provider.tpl
```diff
@@ -0,0 +1,37 @@
+{{- define "stress-test-addons.static-secrets" -}}
+{{- /* The subscriptionConfig key allows a user to use a custom set of subscription+creds uploaded manually to the static keyvault */}}
+{{- /* This template finds all unique subscription configs in scenarios and creates a single secret provider for each */}}
+{{- $global := . }}
+{{- $subChart := get $global.Values "stress-test-addons" }}
+{{- $subConfigs := list (get $subChart.subscription $subChart.env) }}
```
ckairen, line 6: is there a particular reason why we are using different subConfigs for stress-test-static-secret-provider.tpl and _util.tpl?

tools/stress-cluster/chaos/README.md
```diff
@@ -234,6 +235,57 @@ output RESOURCE_GROUP string = resourceGroup().name
...
+AZURE_SUBSCRIPTION_ID=<Subscription ID>
+```
+
+3. Update your scenarios-matrix.yaml file to set the `subscriptionConfig` field for the scenarios that should deploy to the custom subscription.
+   The value should match the secret name in keyvault. See [scenarios-matrix.yaml](examples/stress-deployment-example/scenarios-matrix.yaml) or below for examples.
+
+Override subscription for ALL scenarios
+```
+matrix:
+  subscriptionConfig: yourCustomSubscriptionConfigSecret
```
ckairen, line 265: maybe keep the name consistent? `staticTestSecretsKeyvaultName`

tools/stress-cluster/chaos/README.md
```diff
@@ -234,6 +235,57 @@ output RESOURCE_GROUP string = resourceGroup().name
...
+      foo: bar2
+```
+
+Override subscription for individual scenarios
+```
+  scenarios:
+    deploy-default:
+      foo: bar
+    deploy-custom:
+      subscriptionConfig: yourCustomSubscriptionConfigSecret
```
ckairen, line 279: same here

tools/stress-cluster/chaos/README.md
```diff
@@ -234,6 +235,57 @@ output RESOURCE_GROUP string = resourceGroup().name
...
+      foo: bar
+```
+
+As an example, for the above samples, the following command would set up the custom subscription for use in the Azure SDK Engineering System `pg` cluster:
+
+```
+az keyvault secret set --vault-name stress-secrets-pg -n yourCustomSubscriptionConfigSecret -f <path to sub config>
```
ckairen, line 286: and here

tools/stress-cluster/chaos/README.md
benbp: This points to a keyvault secret, not a keyvault resource, so I guess that would be `staticTestKeyvaultSecretName`. However that's a value that I don't think gives much context to a doc reader, so I wanted to have an example name that was a bit more clear about what the value should be replaced with.

benbp: Perhaps?

```
  subscriptionConfig: <your subscription config secret name>
```

tools/stress-cluster/cluster/kubernetes/stress-test-addons/templates/stress-test-static-secret-provider.tpl
benbp: The subConfig in the util template populates the subscriptionConfig value set for an individual scenario. The `$subConfigs` here collects all subscription configs defined across a matrix config so we can create a secret provider for each.

ckairen (approved):

Issue #4487, opened 2022-10-21
Allow GitHub Groups in CODEOWNERS


needuv:
Currently, the [CODEOWNERS in `azure-sdk-for-python`](https://github.com/Azure/azure-sdk-for-python/blob/main/.github/CODEOWNERS) does not allow GitHub groups to be listed as owners for code paths. 

The [AzureML SDK](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/ml/azure-ai-ml) was recently onboarded to `azure-sdk-for-python`, and we have a large organization with at least 6 or 7 teams (each having at least 5-6 developers) owning/contributing to distinct areas of the SDK. We need to be able to list a single owning team for a particular code path rather than having to list out individual owners, as it is hard to maintain `CODEOWNERS` currently if people leave or if there are re-orgs. It could also benefit other orgs with maintaining ownership of their SDKs, as they only have to worry about team membership rather than updating the `CODEOWNERS` file itself.

Based on a conversation with @scbedd, it should be possible with a change in this file to query for team membership: https://github.com/Azure/azure-sdk-tools/blob/5cf9572442e9a52d011e7789c3cb74ff91cded20/tools/notification-configuration/github-codeowner-subscriber/Program.cs#L132, but it would be great to get some eyes on this and make the change.

weshaggard:
@needuv can you give more details about what team you would like to use in the code owners file? Even if we were to fix this for the pipeline notifications there would be other things that would be broken. For example, to be able to use a team in a codeowners file that team must also have write permissions (https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) to the repo which is generally not the case for most teams. So we wouldn't be able to simply use any random team there would need to be other work necessary in order to be able to use that team.

konrad-jamrozik:
@weshaggard  I have some questions regarding our implementation of our relevant feature.

Q1: I understand the GitHub natively supports teams in CODEOWNERS file, but with the write-access-required limitation as mentioned by you. From the link you shared:

```
# Teams can be specified as code owners as well. Teams should
# be identified in the format @org/team-name. Teams must have
# explicit write access to the repository. In this example,
# the octocats team in the octo-org organization owns all .txt files.
*.txt @octo-org/octocats
```

[...]

PR #6355, opened 2023-06-15
Add GitHub Team User Store to tools sources

JimSuplizio:
This tool simply creates a dictionary of team/users and stores them as a json string in azure blob storage. This is first part of the fix for https://github.com/Azure/azure-sdk-tools/issues/4487. The second part of the fix will be modifying CodeOwnersParser to get the list of users for the team and adding them to the Owners instead of [throwing them out](https://github.com/Azure/azure-sdk-tools/blob/ec7fdf88739040b22ecaa92be67ae8472cb4ad48/tools/code-owners-parser/CodeOwnersParser/CodeownersEntry.cs#L150).

A large part of the why this was done this way was due to the permissions required to access the teams in GitHub. A standard GitHub token cannot have the permissions necessary to read org/team membership. It requires a fine-grained token created by someone with the organization permissions which, unfortunately, excludes bot accounts meaning it has to be from an actual user. The resulting blob string is only 44k and would only need to be pulled by the CodeOwnersParser once, if teams are encountered. 

The GitHub token and blob SAS have already been added to the secret-management tool as manual secrets in the azure-sdk-build-tools repository ([PR 747](https://dev.azure.com/azure-sdk/internal/_git/d16f44db-afeb-4fa1-9724-4389e240f15f/pullrequest/747)). I've also created the variable group, azuresdkartifacts azure-sdk-write-teams variables, which will be used in the pipeline when I create it, after this PR has been merged. Last but not least, this tool won't be published to nuget, the pipeline will build and run the tool as part of the pipeline. Also of note, because it's using a generated GitHub token, it won't affect the repository limits where it runs.


konrad-jamrozik:
> A large part of the why this was done this way was due to the permissions required to access the teams in GitHub.  A standard GitHub token cannot have the permissions necessary to read org/team membership.

How the implementation would differ if the standard GitHub token could have the permissions?

> The GitHub token and blob SAS have already been added to the secret-management tool as manual secrets in the azure-sdk-build-tools repository ([PR 747](https://dev.azure.com/azure-sdk/internal/_git/d16f44db-afeb-4fa1-9724-4389e240f15f/pullrequest/747))

So the GH token you created and put in the KV is your personal token? And the idea is that whomever rotates it next with the secrets tool, will have their own personal token put in KV instead?


> I've also created the variable group, azuresdkartifacts azure-sdk-write-teams variables

My brain's English parser throws exception on this, sorry x-D. Is `azuresdkartifacts` a variable group and `azure-sdk-write-teams` is a variable belonging to it? Could you perhaps link to the variable group from the PR description?

What is the purpose of this variable group and variable?

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Constants/ProductAndTeamConstants.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GitHubTeamUserStore.Constants
+{
```
konrad-jamrozik, line 8:
```suggestion

```

Since .net 6, you can reduce the file indentation everywhere by using [file-scoped namespaces](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces).

tools/github-team-user-store/README.md
```diff
@@ -0,0 +1,28 @@
+# GitHub Team User Store
+
+## Overview
+
+The github-team-user-store is an internal only tool. The tool will get the list of teams/users under azure-sdk-write and store the results as a json string in Azure blob storage. The purpose of this is to enable team usage in CODEOWNERS by allowing the CodeOwnersParser to pull the list of users when it encounters a non-user, or team, entry. The reason why this code isn't directly in the CodeOwnersParser is because it requires a specific, fine-grained Personal Access Token (PAT) in order to be able to get the users for teams which isn't something that's readily available or prudent to be added to all of the places where CodeOwnersParser is used.
```
konrad-jamrozik, line 5: Can we clarify what kind of entity here is the `azure-sdk-write` ? Eg. env variable.

konrad-jamrozik, line 5: So if I understand correctly running this tool puts the `teams <-> user logins` mapping into the blob store, which will be read upon runs of `CodeOwnersParser`. The mapping is read from the `azure-sdk-write` string.

My questions are: 
1. how are we supposed to populate `azure-sdk-write` string? 
2. how and how often are we supposed to execute this tool? Manually, by hand?
3. what will happens if the mapping is outdated because the tool wasn't yet run to reflect recent changes?

I think it would be cool if the README has sections answering all these questions.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+            stopWatch.Start();
+
+            GitHubEventClient gitHubEventClient = new GitHubEventClient(ProductAndTeamConstants.ProductHeaderName);
+            await gitHubEventClient.WriteRateLimits("RateLimit at start of execution:");
+            await TeamUserGenerator.GenerateTeamUserList(gitHubEventClient);
+            await gitHubEventClient.WriteRateLimits("RateLimit at end of execution:");
+            bool storedEqualsGenerated = await TeamUserGenerator.VerifyStoredTeamUsers(gitHubEventClient);
+
+            stopWatch.Stop();
+            // Get the elapsed time as a TimeSpan value.
```
konrad-jamrozik, line 20: This comment seems redundant to me.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/GitHubEventClient.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            var miscRateLimit = await GetRateLimits();
+            CoreRateLimit = miscRateLimit.Resources.Core.Limit;
+            // Get the Minutes till reset.
+            TimeSpan span = miscRateLimit.Resources.Core.Reset.UtcDateTime.Subtract(DateTime.UtcNow);
+            // In the message, cast TotalMinutes to an int to get a whole number of minutes.
+            string rateLimitMessage = $"Limit={miscRateLimit.Resources.Core.Limit}, Remaining={miscRateLimit.Resources.Core.Remaining}, Limit Reset in {(int)span.TotalMinutes} minutes.";
+            if (prependMessage != null)
+            {
+                rateLimitMessage = $"{prependMessage} {rateLimitMessage}";
```
konrad-jamrozik, line 72:
```suggestion
                rateLimitMessage = $"{prependMessage}{rateLimitMessage}";
```

I think that with that space, if there is no `prependMessage`, the message will always be indented by one space, which I guess we don't want.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/GitHubEventClient.cs
```diff
@@ -0,0 +1,199 @@
...
+                {
+                    return await _gitHubClient.Organization.Team.GetAllMembers(team.Id, apiOptions);
+                }
+                // This is what gets thrown if we try and get a userList for certain special teams on GitHub. 
+                // None of these teams are used directly in anything and neither team is a child team of 
+                // azure-sdk-write. If a ForbiddenException is encountered, then report it and return an
+                // empty list.
+                catch (Octokit.ForbiddenException forbiddenEx)
+                {
+                    Console.WriteLine($"{team.Name} cannot be retrieved using a GitHub PAT.");
```
konrad-jamrozik, line 113:
```suggestion
                    Console.WriteLine($"Team named {team.Name} cannot be retrieved using a GitHub PAT.");
```

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/GitHubEventClient.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// <summary>
+        /// Given an Octokit.Team, call to get all child teams.
+        /// </summary>
+        /// <param name="team">Octokit.Team, the team whose child teams to retrieve.</param>
+        /// <returns>IReadOnlyList of Octkit.Team</returns>
+        /// <exception cref="ApplicationException">Thrown if GetAllChildTeams fails after all retries have been exhausted</exception>
+        public async Task<IReadOnlyList<Team>> GetAllChildTeams(Team team)
+        {
+            ApiOptions apiOptions = new ApiOptions();
+            apiOptions.PageSize = 100;
```
konrad-jamrozik, line 138: Make this `100` into class member? Same for all such inline magical constants.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/TeamUserGenerator.cs
```diff
@@ -0,0 +1,133 @@
...
+            bool hasError = false;
+            string rawJson = await gitHubEventClient.GetTeamUserBlobFromStorage();
+            var list = JsonSerializer.Deserialize<List<KeyValuePair<string, List<string>>>>(rawJson);
+            var storedDictionary = list.ToDictionary((keyItem) => keyItem.Key, (valueItem) => valueItem.Value);
+
+            // Verify the dictionary from storage contains everything from the _teamUserDict
+            if (_teamUserDict.Keys.Count != storedDictionary.Keys.Count)
+            {
+                // At this point list the teams and return, don't bother looking at the users.
+                Console.WriteLine($"Error::Created dictionary has {_teamUserDict.Keys.Count} teams and stored dictionary has {storedDictionary.Keys.Count} teams.");
```
konrad-jamrozik, line 92:
```suggestion
                Console.WriteLine($"Error: Created dictionary has {_teamUserDict.Keys.Count} teams and stored dictionary has {storedDictionary.Keys.Count} teams.");
```

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/TeamUserGenerator.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// Generate the team/user lists for each and every team under azure-sdk-write. Every team and user in a CODEOWNERS
+        /// file must have azure-sdk-write permissions in order to be in there which means every team/user will be under
+        /// of azure-sdk-write. This is done to limit the number of calls made because if team/user data for every team
+        /// under Azure was pulled there would be roughly 1933 teams. Getting the team/user data for that many teams would
+        /// end up taking 2500-3000 GitHub API calls whereas getting the team/user data for azure-sdk-write and its child
+        /// teams is less than 1/10th of that. The team/user data is serialized into json and stored in azure blob storage.
+        /// </summary>
+        /// <param name="gitHubEventClient">Authenticated GitHubEventClient</param>
+        /// <returns></returns>
+        public static async Task GenerateTeamUserList(GitHubEventClient gitHubEventClient)
```
konrad-jamrozik, line 28: Based on this method name I would expect for it to return e.g. list of user names, but instead it returns void and has the side-effect of uploading to blob storage.

A name like `UploadTeamUserListToBlobStorage` would be better IMO.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/TeamUserGenerator.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        /// <summary>
+        /// Call GitHub to get users for the team and add a dictionary entry for the team/users. Note: GitHub returns 
+        /// a distinct list of all of the users, including users from any/all child teams. After that, get the list
+        /// of child teams and recursively call GetUsersForTeam for each one of those to create their team/user entries.
+        /// </summary>
+        /// <param name="gitHubEventClient">Authenticated GitHubEventClient</param>
+        /// <param name="team">Octokit.Team to get users for.</param>
+        public static async Task GetUsersForTeam(GitHubEventClient gitHubEventClient, Team team)
```
konrad-jamrozik, line 46: Given this method starts with `Get` I would expect for it to return some value, but it returns nothing.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Constants/StorageConstants.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GitHubTeamUserStore.Constants
+{
+    internal class StorageConstants
+    {
+        public const string AzureBlobAccountName = "azuresdkartifacts";
```
weshaggard, line 11: These should not be hard-coded this should be a configuration that we pass into the tool when we run it.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Program.cs
```diff
@@ -0,0 +1,40 @@
+using System.Diagnostics;
+using GitHubTeamUserStore.Constants;
+
+namespace GitHubTeamUserStore
+{
+    internal class Program
+    {
+        static async Task Main(string[] args)
```
weshaggard, line 8: Where is the pipeline that we will be running this tool in?

konrad-jamrozik: I think it is to-be-created per PR description:

> I've also created the variable group, azuresdkartifacts azure-sdk-write-teams variables, which will be used in the pipeline when I create it, after this PR has been merged.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/GitHubEventClient.cs
JimSuplizio: I'm not changing this. If prepend message is null it simply doesn't add a prependMessage to the rateLimitMessage

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/GitHubEventClient.cs
JimSuplizio: We're in a function that gets members for a team, which can throw the ForbiddenException if it can't be retrieved via a PAT. Furthermore, I'm outputting the exception whose message is "Exception trying to get members for Team: `Team Name`"

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Constants/ProductAndTeamConstants.cs
JimSuplizio: @konrad-jamrozik, this is a preference. I prefer the namespace in the file otherwise I'd select the checkbox to use file-scoped namespaces.

JimSuplizio:
> > A large part of the why this was done this way was due to the permissions required to access the teams in GitHub.  A standard GitHub token cannot have the permissions necessary to read org/team membership.
> 
> How the implementation would differ if the standard GitHub token could have the permissions?
The standard GitHub Token cannot have Organization->Membership:Read permissions which means they cannot query for team membership. Furthermore, in order to query teams and membership for an org, the account that created the token needs to be a member of Azure. This is something @weshaggard and I have discussed and it was decided to use a generated token.
> 
> > The GitHub token and blob SAS have already been added to the secret-management tool as manual secrets in the azure-sdk-build-tools repository ([PR 747](https://dev.azure.com/azure-sdk/internal/_git/d16f44db-afeb-4fa1-9724-4389e240f15f/pullrequest/747))
> 
> So the GH token you created and put in the KV is your personal token? And the idea is that whomever rotates it next with the secrets tool, will have their own personal token put in KV instead?
Yes. The secrets are in the azure-sdk-build-tools repository in the tools/secret-management/manual/AzureSDKEngKeyVault. The json files are azuresdkartifacts-azure-sdk-write-teams-github-pat.json and azuresdkartifacts-azure-sdk-write-teams-sas.json, both of which describe what needs to be done.
> 
> > I've also created the variable group, azuresdkartifacts azure-sdk-write-teams variables
> 
> My brain's English parser throws exception on this, sorry x-D. Is `azuresdkartifacts` a variable group and `azure-sdk-write-teams` is a variable belonging to it?
> 
> Could you perhaps link to the variable group from the PR description?
> 
> What is the purpose of this variable group and variable? I know they will be used in the pipeline, but how?
I don't normally link variable groups in PRs because they're public. This is the [variable group](https://dev.azure.com/azure-sdk/internal/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=133&path=azuresdkartifacts%20azure-sdk-write-teams%20variables) that holds these two variables. Like other variables in our pipelines, these are linked to the keyvault secrets and pull their values from there which is the standard. When the values change, after a rotation, the variable group won't need to change because it'll just start pulling the new values.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Program.cs
JimSuplizio: @weshaggard and @konrad-jamrozik 
I need to get this merged prior to creating the pipeline. Unlike other tools, this won't be being published to the NET dev feed because it's unusable outside of pipeline so there's really no point. The only test that makes sense here is to pull back what was just stored and ensure it matches the dictionary created from the team/user data and this done immediately after the call to store it. The pipeline just needs to build and run the tool to do both.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Program.cs
JimSuplizio: Removed with next commit.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/TeamUserGenerator.cs
JimSuplizio: Done. Will be pushed in the next commit. In the future, please make the suggestion and I'll do things locally. I don't like to commit suggestions, I'd rather make the change locally and add a commit.

tools/github-team-user-store/README.md
JimSuplizio: azure-sdk-write is a team. It's the team that everything with write permissions is supposed to inherit from. Because these teams are going to be in CODEOWNERS they're all required to have write permission. Rather than getting all 1933 teams in the Azure organization, it was decided that, because of the write permission limitation, to only get the teams under azure-sdk-write. 

I'll clarify the comment with the next commit.

JimSuplizio: To answer the questions:
1. As previously mentioned, there is no azure-sdk-write string, it's the root team for that all azure sdk teams with write permission are supposed to be children of. The dictionary of team/users recursively includes all the teams under, and including, azure-sdk-write.
2. As stated in the PR description, this tool will run in a pipeline which has not been created yet. I plan on having it run at least once daily and we'll tweak accordingly.
3. Anything running that uses this data will run with the data it has at the time.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/GitHubEventClient.cs
JimSuplizio: Made the API options a static variable on the class with the default, from a const, MaxPageSize.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/TeamUserGenerator.cs
JimSuplizio: Renamed to CreateTeamUserEntry

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Constants/StorageConstants.cs
JimSuplizio: I've added command line (via System.CommandLine) arguments which will still use these values as defaults. I've renamed the class to DefaultStorageConstants and prepended each const name with "Default". This will be updated with the next commit.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Program.cs
```diff
@@ -7,20 +8,62 @@ internal class Program
     {
         static async Task Main(string[] args)
         {
+            var azureBlobAccountNameOption = new Option<string>
```
weshaggard, line 11: Any reason these need to be split apart? I was expecting just a blob uri. As in `string blobUri = $"https://{DefaultStorageConstants.DefaultAzureBlobAccountName}.blob.core.windows.net/{DefaultStorageConstants.DefaultAzureSdkWriteTeamsContainer}/{DefaultStorageConstants.DefaultAzureSdkWriteTeamsBlobName}";`

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/GitHubEventClient.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        /// <summary>
+        /// Store the team/user blob in Azure blob storage.
+        /// </summary>
+        /// <param name="rawJson">The json string, representing the team/user information, that will be uploaded to blob storage.</param>
+        /// <returns></returns>
+        /// <exception cref="ApplicationException">If there is no AZURE_SDK_TEAM_USER_STORE_SAS in the environment</exception>
+        public async Task UploadToBlobStorage(string rawJson)
+        {
+            var azureSdkWriteTeamsSAS = Environment.GetEnvironmentVariable("AZURE_SDK_TEAM_USER_STORE_SAS");
```
weshaggard, line 170: Similar to the blob client this should probably commandline parameter, or potentially be part of the blob uri.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Program.cs
weshaggard: If you need to have the different pieces of the uri you can use https://learn.microsoft.com/en-us/dotnet/api/azure.storage.blobs.bloburibuilder.-ctor?view=azure-dotnet to parse it into the pieces.

JimSuplizio: The next commit will be taking the blob URI in from the command line and using the BlobUriBuilder

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/GitHubEventClient.cs
JimSuplizio: This will be gone from the next commit.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/Constants/StorageConstants.cs
JimSuplizio: The storage constants are now gone, the URI will be a command line option and BlobUriBuilder will parse the options

weshaggard (approved):

PR #6362, opened 2023-06-16
[Pylint] httpresponse error

l0lawrence:
https://github.com/Azure/azure-sdk-for-python/issues/30785


tools/pylint-extensions/azure-pylint-guidelines-checker/README.md
```diff
@@ -87,3 +87,4 @@ docstring-missing-type | pylint:disable=docstring-missing-type    |       Docstr
 docstring-missing-return | pylint:disable=docstring-missing-return    |       Docstring missing return.                        | [link](https://azure.github.io/azure-sdk/python_documentation.html#docstrings)           |
 docstring-missing-rtype | pylint:disable=docstring-missing-rtype    |       Docstring missing return type.                        | [link](https://azure.github.io/azure-sdk/python_documentation.html#docstrings)           |
 docstring-should-be-keyword | pylint:disable=docstring-should-be-keyword    |       Docstring should use keywords.                        | [link](https://azure.github.io/azure-sdk/python_documentation.html#docstrings)           |
+no-azure-core-http-response-import | pylint:disable=no-azure-core-http-response-import    |      Do not import HttpResponse from Azure Core.                       | [link](https://github.com/Azure/azure-sdk-for-python/issues/30785)           |
```
kristapratico, line 90: Maybe let's re-word to something like "no legacy import". Otherwise this makes me think I should never import HttpResponse from azure.core

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2178,6 +2178,38 @@ def visit_functiondef(self,node):
...
+
+    AZURE_CORE_NAME = "azure.core"
+    AZURE_CORE_TRANSPORT_NAME = "azure.core.pipeline.transport"
+    RESPONSE_CLASSES = {"HttpResponse", "AsyncHttpResponse"}
+
+    def visit_importfrom(self, node):
+        """Check that we aren't importing HttpResponse."""
+        if node.root().name.startswith(self.AZURE_CORE_NAME): 
+            return
+        if node.modname == self.AZURE_CORE_TRANSPORT_NAME:
```
kristapratico, line 2203: Does this check specifically for the `from x import y` type of import? Do we also want to look for usage of `import azure.core.transport.pipeline` / `azure.core.transport.pipeline.HttpResponse` in the code?

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2178,6 +2178,38 @@ def visit_functiondef(self,node):
...
+
+    def visit_importfrom(self, node):
+        """Check that we aren't importing HttpResponse."""
+        if node.root().name.startswith(self.AZURE_CORE_NAME): 
+            return
+        if node.modname == self.AZURE_CORE_TRANSPORT_NAME: 
+            for name, _ in node.names:
+                if name in self.RESPONSE_CLASSES:
+                    self.add_message(
+                        msgid=f"no-azure-core-http-response-import",
```
kristapratico, line 2207: nit, no need for f-string
```suggestion
                        msgid="no-azure-core-http-response-import",
```

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2205,6 +2237,7 @@ def register(linter):
     linter.register_checker(NameExceedsStandardCharacterLength(linter))
     linter.register_checker(DeleteOperationReturnStatement(linter))
     linter.register_checker(ClientMethodsHaveTracingDecorators(linter))
+    linter.register_checker(NoAzureCoreHttpResponseImport(linter))
```
kristapratico, line 2240: would also put "legacy" in the rule name

tools/pylint-extensions/azure-pylint-guidelines-checker/dev_requirements.txt
```diff
@@ -1,3 +1,3 @@
-pylint==2.14.5
+pylint==2.16.0
```
kristapratico, line 1: Does this bump match what we'll run in the azure-sdk-for-python repo once we merge pylint-next?

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2196,6 +2196,52 @@ def visit_functiondef(self,node):
...
+            for name, _ in node.names:
+                if name in self.RESPONSE_CLASSES:
+                    self.add_message(
+                        msgid=f"no-legacy-azure-core-http-response-import",
+                        node=node,
+                        confidence=None,
+                    )
+    
+    def visit_import(self, node):
+        """Check that we aren't importing import azure.core.pipeline.transport.HttpResponse."""
```
kristapratico, line 2231: I think we might have to include 2 separate checks here, but let me know if I'm wrong. In my head, we do one check in the file for a `import azure.core.pipeline.transport` and then we have to check/visit the code in the file for any usage `azure.core.pipeline.transport.HttpResponse`.

l0lawrence: I tried a global search in the python repo currently and it doesn't look like there are any instances of just import azure.core.pipeline.transport, so I don't know how often this would be used but I can add it! Maybe it will be more likely to pop up when typing gets added places?

Finding every individual usage of httpResponse (without digging into astroid yet -- there could be a faster way) would require iteration through every funcDef and then within that funcDef check every assign node when we see a file that has `import azure.core.pipeline.transport` at the top 

i.e something like : response = azure.core.pipeline.transport.HttpResponse(.....

tools/pylint-extensions/azure-pylint-guidelines-checker/dev_requirements.txt
l0lawrence: Small edit should be 2.15.8, not bumping to 2.16 in september b/c libraries I think will still be working through their docstrings  -- will update :)

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
kristapratico: FYI I'm fine with just checking the `from x import y` style import, but wanted to double-check since I think the issue you created had said something like check *all* ways of importing it. If you think this will be a big perf hit for a situation unlikely to happen then maybe we don't need it

kristapratico: So should we remove the visit_import function?

kristapratico (approved):

PR #6363, opened 2023-06-19
Improvements to APIView/Java tooling

JonathanGiles:
A number of fixes to the Java APIView tooling, discovered and developed as I went about creating APIView reviews for the Java Semantic Kernel library. You can see the generated APIView reviews for these libraries here:

https://apiview.dev/Assemblies/Review/8aa5b4eae87c47a9996bf9c80163812e
https://apiview.dev/Assemblies/Review/f52ac2e951734ee0a2c010481892fe32
https://apiview.dev/Assemblies/Review/73736094cf044f70b82a0618a97776f4

Here are the changes I have made:

 * Support for when a pom.xml does not explicitly state a groupId or version - look up in parent, if it is available.
 * Don't analyse files in META-INF, as some builds put .java files in there, and they are not always valid Java.
 * Fixing issue where not all inner types are discovered when calling getClasses. This also resulted in some additional bugs being identified in diagnostics, where the reported issue was at the parent type rather than the child type, so I fixed those.
 * Add new diagnostic to warn developers to use proper casing for enum constant values.
 * Made it so that Azure SDK specific errors are only applied to libraries with com.azure package namespaces.

### DIFF
--- a/src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
+++ b/src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
@@ -150,8 +150,10 @@ public class JavaASTAnalyser implements Analyser {
 
     private boolean filterFilePaths(Path filePath) {
         String fileName = filePath.toString();
-        // Skip paths that are directories or are in implementation.
-        if (Files.isDirectory(filePath) || fileName.contains("implementation")) {
+        // Skip paths that are directories, in implementation, or are files contained within META-INF
+        if (Files.isDirectory(filePath)
+                || fileName.contains("implementation")
+                || fileName.contains("META-INF")) {
             return false;
         } else {
             // Only include Java files.
@@ -210,9 +212,9 @@ public class JavaASTAnalyser implements Analyser {
 //            combinedTypeSolver.add(new SourceJarTypeSolver(inputFile));
 
             ParserConfiguration parserConfiguration = new ParserConfiguration()
-                .setStoreTokens(true)
-                .setSymbolResolver(new JavaSymbolSolver(combinedTypeSolver))
-                .setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_11);
+                    .setStoreTokens(true)
+                    .setSymbolResolver(new JavaSymbolSolver(combinedTypeSolver))
+                    .setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_11);
 
             // Configure JavaParser to use type resolution
             StaticJavaParser.setConfiguration(parserConfiguration);
@@ -223,10 +225,10 @@ public class JavaASTAnalyser implements Analyser {
             if (path.endsWith("package-info.java")) {
                 compilationUnit.getPackageDeclaration().ifPresent(pd -> {
                     compilationUnit.getAllComments().stream()
-                        .filter(Comment::isJavadocComment)
-                        .map(Comment::asJavadocComment)
-                        .findFirst()
-                        .ifPresent(comment -> packageNameToPackageInfoJavaDoc.put(pd.getNameAsString(), comment));
+                            .filter(Comment::isJavadocComment)
+                            .map(Comment::asJavadocComment)
+                            .findFirst()
+                            .ifPresent(comment -> packageNameToPackageInfoJavaDoc.put(pd.getNameAsString(), comment));
                 });
 
                 return Optional.empty();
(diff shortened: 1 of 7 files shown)


src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/rules/MissingAnnotationsDiagnosticRule.java
```diff
@@ -7,61 +7,60 @@
...
-                        "https://azure.github.io/azure-sdk/java_introduction.html#service-client"));
+                    if (annotatedMethodCount.get() / (double) methodCount.get() < 0.75) {
+                        // warn user to double check
+                        listing.addDiagnostic(new Diagnostic(
+                            WARNING,
+                            makeId(typeDeclaration),
+                        "There is a low number of methods annotated with @ServiceMethod. " +
+                                "Please review to ensure all appropriate methods have this annotation.",
+                            "https://azure.github.io/azure-sdk/java_introduction.html#service-client"));
+                    }
```
JonathanGiles, line 62: FYI - the bulk of this change is simply an indentation by one level. The only notable change is the line which has the `.filter(type -> getPackageName(type).equals("com.azure"))` on it.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/util/ASTUtils.java
```diff
@@ -90,7 +90,15 @@ public static Optional<String> getClassName(CompilationUnit cu) {
...
-        return cu.getTypes().stream();
+        // previously we simply return 'cu.getTypes().stream()', but this had the effect of not returning all inner
+        // types, as was expected. This is because the 'getTypes()' method only returns the top-level types, and not
+        // member types. To fix this, we now return a stream of all types, including member types.
+        return Stream.concat(
+                cu.getTypes().stream(), // top-level types
+                cu.getTypes().stream()  // member types
+                     .flatMap(type -> type.getMembers().stream()
+                          .filter(member -> member instanceof TypeDeclaration<?>)
+                          .map(member -> (TypeDeclaration<?>) member)));
```
srnagar, line 101: 👍 good catch!

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/rules/MissingAnnotationsDiagnosticRule.java
```diff
@@ -7,61 +7,60 @@
...
-                // check if the @ServiceClient annotation is present
-                if (!typeDeclaration.isAnnotationPresent("ServiceClient")) {
-                    listing.addDiagnostic(new Diagnostic(
-                        INFO,
-                        makeId(cu),
-                        "Classes named *Client are potential candidates to have the @ServiceClient annotation applied.",
-                        "https://azure.github.io/azure-sdk/java_introduction.html#service-client"));
-                }
+        getClasses(cu)
+            .filter(type -> getPackageName(type).equals("com.azure"))   // we only want to give this guidance to Azure SDK developers
```
srnagar, line 18: Should this be `.startsWith("com.azure")` instead of `equals`?

srnagar (approved):

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/rules/MissingAnnotationsDiagnosticRule.java
JonathanGiles: Good call - I'll make that change

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -615,9 +617,10 @@ private void getEnumEntries(EnumDeclaration enumDeclaration) {
                 // annotations
                 getAnnotations(enumConstantDeclaration, false, false);
 
-                // create a unique id for enum constants
+                // create a unique id for enum constants by using the fully-qualified constant name
+                // (package, enum name, and enum constant name)
                 final String name = enumConstantDeclaration.getNameAsString();
-                final String definitionId = makeId(enumDeclaration.getFullyQualifiedName().get() + "." + counter);
```
praveenkuttappan, line 620: I believe this will cause to create a diff and a new revision though nothing is different in UI because diff comparison compares both text and underlying ID. One option to avoid that issue is to increase the version of parser to rebuild all existing revisions.

praveenkuttappan (approved):
Looks good to me. I think we should increment the parser version to avoid having a new revision though nothing was modified at code level.

JonathanGiles:
Merging now - Praveen has offered to do the version increments when he does the next APIView release.

PR #6366, opened 2023-06-19
CodeOwnersParser Team/User lookup

JimSuplizio:
When CodeOwnersParser encounters a team in the CODEOWNERS file it is parsing add the team's users, instead of the team, to the Owners list for the path. When adding the users for a team to the Owners list, it's done using a union to ensure that the resulting list is distinct. The list is pulled from blob data. 

In the case where a team is encountered in the CODEOWNERS file that doesn't exist in the team/user data, the team will get added as an owner which is the behavior today. Because of this, the ExcludeNonUserAliases function still needs to be in there.

I'd also fixed an [error](https://github.com/Azure/azure-sdk-tools/pull/6366/files#diff-f071416a991afecb5cbab353457ad32f1cbb8149ef4ba45db3c97e43a8b1befaR113) I'd encountered with parsing where, if a line didn't have any owner entries, the owner was being set to the path expression.  [Here's what the CODEOWNERS line looks like when this occurs.](https://github.com/Azure/azure-sdk-for-js/blob/main/.github/CODEOWNERS#L68)

The other piece I've added to the CodeOwnersParser solution is a manual test project, CodeOwnersManualTester, which loads the selected CODEOWNERS files for testing/inspection purposes. Since this is not the default project for the solution, debugging just requires right clicking on the project and selecting Debug from there.


tools/code-owners-parser/CodeOwnersParser/CodeownersFile.cs
```diff
@@ -3,18 +3,48 @@
...
 using System.IO;
 using System.Linq;
+using System.Text.Json;
 
 namespace Azure.Sdk.Tools.CodeOwnersParser
 {
     public static class CodeownersFile
     {
+
+        private static string teamUserBlobUri = $"https://{StorageConstants.AzureBlobAccountName}.blob.core.windows.net/{StorageConstants.AzureSdkWriteTeamsContainer}/{StorageConstants.AzureSdkWriteTeamsBlobName}";
```
weshaggard, line 13: I'm OK with this be a default value but I think this needs to be an option that we can pass in at a command line parameter as well.

tools/code-owners-parser/CodeOwnersParser/CodeownersFile.cs
```diff
@@ -3,18 +3,48 @@
...
+        private static Dictionary<string, List<string>>? teamUserDict = null;
         public static List<CodeownersEntry> GetCodeownersEntriesFromFileOrUrl(
             string codeownersFilePathOrUrl)
         {
             string content = FileHelpers.GetFileOrUrlContents(codeownersFilePathOrUrl);
             return GetCodeownersEntries(content);
         }
 
+        public static Dictionary<string, List<string>>?GetTeamUserData()
```
konrad-jamrozik, line 22:
```suggestion
        public static Dictionary<string, List<string>>? GetTeamUserData()
```

weshaggard, line 22:
```suggestion
        public static Dictionary<string, List<string>>? GetTeamUserData()
```

tools/code-owners-parser/CodeOwnersParser/CodeownersFile.cs
konrad-jamrozik, line 13: Should `private static` fields start with small letter?

tools/code-owners-parser/CodeOwnersManualTester/CodeOwnerUtils.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Azure.Sdk.Tools.CodeOwnersParser;
+
+namespace CodeOwnersManualTester
+{
+    internal class CodeOwnerUtils
```
konrad-jamrozik, line 10: I see inconsistency in naming:
- `CodeOwner`
- `CodeOwners`

I would suggest to always use `Codeowners`, as this is the convention I adopted in this codebase.

tools/code-owners-parser/CodeOwnersParser/CodeownersFile.cs
weshaggard, line 13: Where is the tool/code that will create this file? Is that part of another PR?

tools/code-owners-parser/CodeOwnersManualTester/CodeOwnerUtils.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// Wrapper function to load the CODEOWNERS file from a given path or URL and return
+        /// the list of codeowners entries.
+        /// </summary>
+        /// <param name="codeOwnersFilePath"></param>
+        /// <returns>List of CodeownersEntry</returns>
+        public static List<CodeownersEntry> GetCodeOwnerEntries(string codeOwnersFilePath)
```
konrad-jamrozik, line 18: I feel this method belongs in the `CodeownersFile` class.

tools/code-owners-parser/CodeOwnersManualTester/CodeOwnersManualTester.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
konrad-jamrozik, line 7: Did you consider making them [errors](https://stackoverflow.com/a/62116924/986533) ?

tools/code-owners-parser/CodeOwnersManualTester/Program.cs
```diff
@@ -0,0 +1,38 @@
+using Azure.Sdk.Tools.CodeOwnersParser;
+
+namespace CodeOwnersManualTester
+{
+    internal class Program
```
konrad-jamrozik, line 5: Do we need this tool in addition to [CodeownersManualAnalysisTests](https://github.com/Azure/azure-sdk-tools/blob/c37701ec43d68f96f480ddf7fc502841be7fbaed/tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners.Tests/CodeownersManualAnalysisTests.cs#L40) ?

tools/code-owners-parser/CodeOwnersParser/CodeownersEntry.cs
```diff
@@ -106,11 +106,69 @@ public void ParseOwnersAndPath(string line)
...
+                            var teamUsers = GetUsersForTeam(author.Trim());
+                            // If the team is found in team user data, add the list of users to
+                            // the owners and ensure the end result is a distinct list
+                            if (teamUsers.Count > 0)
+                            {
+                                // The union of the two lists will ensure the result a distinct list
+                                Owners = Owners.Union(teamUsers).ToList();
+                            }
+                            // Else, the team user data did not contain an entry or there were no user
+                            // for the team. In that case, just add the team to the list of authors
```
konrad-jamrozik, line 131: > In that case, just add the team to the list of authors

I understand this is our "best guess" what is going on, and this may actually be incorrect interpretation? If so, what bad things will happen?

tools/code-owners-parser/CodeOwnersParser/CodeownersEntry.cs
```diff
@@ -106,11 +106,69 @@ public void ParseOwnersAndPath(string line)
...
+                            else
+                            {
+                                Owners.Add(author);
```
konrad-jamrozik, line 134: Should we also add `.Trim()` to `author` here?

tools/code-owners-parser/CodeOwnersParser/CodeownersEntry.cs
```diff
@@ -106,11 +106,69 @@ public void ParseOwnersAndPath(string line)
             line = ParsePath(line);
             line = RemoveCommentIfAny(line);
 
-            foreach (string author in SplitLine(line, OwnerSeparator).ToList())
+            // If the line doesn't contain the OwnerSeparator AKA no owners, then the foreach loop below
+            // won't work. For example, the following line would end up causing "/sdk/communication" to
+            // be added as an owner when one is not listed
+            // /sdk/communication/
+            if (line.Contains(OwnerSeparator))
```
konrad-jamrozik, line 113: I think it would be nice to have automated unit tests covering this new logic that run as part of the automated regression test suite, not manually.

tools/code-owners-parser/CodeOwnersParser/CodeownersEntry.cs
```diff
@@ -106,11 +106,69 @@ public void ParseOwnersAndPath(string line)
...
             {
-                if (!string.IsNullOrWhiteSpace(author))
-                    Owners.Add(author.Trim());
+                foreach (string author in SplitLine(line, OwnerSeparator).ToList())
+                {
+                    if (!string.IsNullOrWhiteSpace(author))
```
konrad-jamrozik, line 117: To aid readability, I would extract the body of this `foreach` to a method with intention-revealing name.

tools/code-owners-parser/CodeOwnersParser/CodeownersFile.cs
JimSuplizio: Will be updated with the next commit.

JimSuplizio: Will be updated with next commit.

tools/code-owners-parser/CodeOwnersManualTester/CodeOwnerUtils.cs
JimSuplizio: Sure. I'll update with the next commit.

tools/code-owners-parser/CodeOwnersManualTester/CodeOwnerUtils.cs
JimSuplizio: The wrapper class is gone. Will be updated with the next commit.

tools/code-owners-parser/CodeOwnersManualTester/CodeOwnersManualTester.csproj
JimSuplizio: Sure, why not. Nothing in the ad-hoc project is using them but they'll be consistent with the rest of the projects in the solution.

tools/code-owners-parser/CodeOwnersParser/CodeownersFile.cs
JimSuplizio: The tool that creates this is in the [GitHub Team User Store PR](https://github.com/Azure/azure-sdk-tools/pull/6355)

tools/code-owners-parser/CodeOwnersManualTester/CodeOwnersManualTester.csproj
JimSuplizio: I'm going to be removing this project altogether. You have something similar already in the solution.

tools/code-owners-parser/CodeOwnersManualTester/Program.cs
JimSuplizio: I'm removing this project with the next commit

tools/code-owners-parser/CodeOwnersParser/CodeownersEntry.cs
JimSuplizio: The fact that there was a trim for anything calling returned SplitLine call is a problem with that function. It's already using the `StringSplitOptions.RemoveEmptyEntries` in the call and should have been `StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries` to that line so everything returned is trimmed. Further, the documentation for [String.Split states](https://learn.microsoft.com/en-us/dotnet/api/system.stringsplitoptions?view=net-7.0) that if RemoveEmptyEntries and TrimEntries are specified together, then substrings that consist only of white-space characters are also removed from the result. Since string.split cannot return null, those IsNullOrWhiteSpace checks that are done on each entry can be removed. This will be updated with the next commit.

tools/code-owners-parser/CodeOwnersParser/CodeownersEntry.cs
JimSuplizio: This isn't a guess. The team is either in the team/user data and I'm adding the list of users for that team to the Owners instead of the team. If the team is not in the team/user data, then I'm just adding the team to the list as it previously did.

tools/code-owners-parser/Azure.Sdk.Tools.RetrieveCodeOwners/Program.cs
```diff
@@ -48,7 +50,9 @@ public static class Program
         string codeownersFilePathOrUrl,
         bool excludeNonUserAliases = false,
         string? targetDir = null,
-        string ignoredPathPrefixes = DefaultIgnoredPrefixes)
+        string ignoredPathPrefixes = DefaultIgnoredPrefixes,
+        string? teamStorageURI = null,
+        string? ownersDataOutputFile = null)
```
JimSuplizio, line 55: @weshaggard, I had to add another command line option to RetrieveCodeOwners to allow writing the json output to the file. The fact is, the [get-codeowners.lib.ps1](https://github.com/Azure/azure-sdk-tools/blob/952b6db75fd998adc6cf83af03da3a158ab1ee80/eng/common/scripts/get-codeowners.lib.ps1#L125) was incorrectly assuming any/all output from the executable was nothing more than a json codeowner entry. Using CodeOwnersParser FileHelper [GetUrlContents](https://github.com/Azure/azure-sdk-tools/blob/main/tools/code-owners-parser/CodeOwnersParser/FileHelpers.cs#L38) to fetch the blob, will cause output that'll break this assumption. The correct fix for this is the add the option to write the entry to a file. The problem here is because of locked version in the [eng/common/get-codeowners.lib.ps1](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/get-codeowners.lib.ps1#L81) script, I can't make the changes, to use the file instead of output, to the script until this is checked in and another version has been published.
@konrad-jamrozik and I already chatted about this and both agreed this was the correct way to fix this.

tools/code-owners-parser/CodeOwnersParser/CodeownersFile.cs
JimSuplizio: As per our discussion, this is now an option that can be given to the parser and has been added as a command line option to the RetrieveCodeOwners executable.

tools/code-owners-parser/CodeOwnersParser/TeamUserHolder.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            if (null == _teamUserDict)
+            {
+                string rawJson = FileHelpers.GetFileOrUrlContents(TeamUserStorageURI);
+                var list = JsonSerializer.Deserialize<List<KeyValuePair<string, List<string>>>>(rawJson);
+                if (null != list)
+                {
+                    return list.ToDictionary((keyItem) => keyItem.Key, (valueItem) => valueItem.Value);
+                }
+                Console.WriteLine($"Error! Unable to deserialize json team/user data. rawJson={rawJson}");
```
weshaggard, line 50: Probably worth including the TeamUserStorageURI as part of the error message as well.

tools/code-owners-parser/CodeOwnersParser/TeamUserHolder.cs
```diff
@@ -0,0 +1,79 @@
...
+            string teamWithoutOrg = teamName.Trim();
+            if (teamWithoutOrg.Contains('/'))
+            {
+                teamWithoutOrg = teamWithoutOrg.Split("/")[1];
+            }
+            if (TeamUserDict != null)
+            {
+                if (TeamUserDict.ContainsKey(teamWithoutOrg))
+                {
+                    Console.WriteLine($"Found team entry for {teamWithoutOrg}");
```
weshaggard, line 71: We might want to think about making this logging configurable or use a logging interface at some point as generally it isn't great for shared libraries to have a bunch of logging.

weshaggard (approved):

tools/code-owners-parser/CodeOwnersParser/TeamUserHolder.cs
JimSuplizio: I'll add this in another commit before I merge.

tools/code-owners-parser/CodeOwnersParser/TeamUserHolder.cs
JimSuplizio: That won't be part of this PR. I created a [new issue](https://github.com/Azure/azure-sdk-tools/issues/6438) to track this.

PR #6375, opened 2023-06-21
Add openapi hub redirect app

benbp:
This is a simple web app we deploy to app services that `portal.azure-devex-tools.com` lands on. This app does a client-side check to see if `openapihub.azure-devex-tools.com` is accessible and if so redirects to it, preserving any path and query parameters. If it is not accessible, the site redirects to https://aka.ms/openapihub-landing which tells users how to get access to corp net if they are not on campus.

### DIFF
--- a/tools/openapihub-redirect/README.md
+++ b/tools/openapihub-redirect/README.md
@@ -1,0 +1,20 @@
+This is a simple web app to redirect users to either [openapihub](https://openapihub.azure-devex-tools.com) if they are on corp net, or [openapi hub access docs](https://aka.ms/openapihub-landing) if they are off corp net.
+This site is configured to load from `https://portal.azure-devex-tools.com` via the [app service domain config](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/8c0e0d10-c2d6-42b8-b1d9-e7132b6f18a3/resourceGroups/openapi-platform-preview/providers/Microsoft.Web/sites/hub-redirect-app/domainsandsslv2) and the [azure-devex-tools DNS zone](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/8c0e0d10-c2d6-42b8-b1d9-e7132b6f18a3/resourceGroups/openapi-platform-preview/providers/Microsoft.Network/dnszones/azure-devex-tools.com/overview).
+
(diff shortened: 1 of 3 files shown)


tools/openapihub-redirect/README.md
```diff
@@ -0,0 +1,20 @@
...
+This site is configured to load from `https://portal.azure-devex-tools.com` via the [app service domain config](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/8c0e0d10-c2d6-42b8-b1d9-e7132b6f18a3/resourceGroups/openapi-platform-preview/providers/Microsoft.Web/sites/hub-redirect-app/domainsandsslv2) and the [azure-devex-tools DNS zone](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/8c0e0d10-c2d6-42b8-b1d9-e7132b6f18a3/resourceGroups/openapi-platform-preview/providers/Microsoft.Network/dnszones/azure-devex-tools.com/overview).
+
+To update this app:
+
+1. Update `app.py` and/or `index.html`
+1. Enable basic authentication under General Settings [here](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/8c0e0d10-c2d6-42b8-b1d9-e7132b6f18a3/resourceGroups/openapi-platform-preview/providers/Microsoft.Web/sites/hub-redirect-app/configuration)
+1. Run the following commands from this directory:
+
+```
+az login
```
weshaggard, line 11: Should we create this as a pipeline instead of just text?

tools/openapihub-redirect/requirements.txt
```diff
@@ -0,0 +1 @@
+flask
```
weshaggard, line 1: Do we need this file if we only have the one requirement?

tools/openapihub-redirect/README.md
benbp: I hope to never have to deploy this again. I would say if we deploy it a second or third time we should create a pipeline, but we would also need to figure out deployment auth. Right now I'm using user credentials which only works if basic auth is enabled. Our subscription policy disables it after about 30 seconds, so in a pipeline we would need to set all that up as well.

weshaggard (approved):

PR #6381, opened 2023-06-21
Remove openapi hub landing docs in favor of wiki

benbp:


kurtzeborn (approved):

doc/openapi_hub_redirect.md
```diff
@@ -0,0 +1,23 @@
...
+
+There are several options for users who do not have corp net access:
+
+* [AVD Workstation](https://microsoft.sharepoint.com/sites/Security_Tools_Services/SitePages/WindowsVirtualDesktop/WVD-Workstation.aspx) - This is the easiest option to set up for occasional use, but the least convenient option to use on a frequent basis. Files can be dragged or copy/pasted to the workstation, or are accessible from your OneDrive.
+* [Azure Virtual Desktop](https://microsoft.sharepoint.com/sites/security_tools_services/sitepages/windowsvirtualdesktop/windows-virtual-desktop.aspx) - Use this if you use OpenAPI hub more frequently and need a more responsive machine to interact with the site. This option is a shared VM/jumpbox with no cost to your team.
+* [Microsoft Dev Box](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/dev-box/dev-box) - Best option as a daily driver VM for performing other tasks along with OpenAPI hub. A dev box resource will be billed to your team.
+
+For a detailed comparision of all access options, see [Modern Access](https://microsoft.sharepoint.com/sites/Identity_Authentication/SitePages/ModernAccess/About-Modern-Access.aspx)
```
mikeharder, line 19:
```suggestion
Choose one of the options for [remote access](https://microsoft.sharepoint.com/sites/Identity_Authentication/SitePages/ModernAccess/About-Modern-Access.aspx).  We recommend these two:

* Azure Virtual Desktop (vDesktop) - Lightweight solution if you only need to access the OpenAPI hub
* Microsoft Dev Box - Personal VM you can use for all development work
```

mikeharder (changes requested):
Simplified recommendations

doc/openapi_hub_redirect.md
benbp: The main reason I didn't just link to the parent doc is when I was setting all this up I found all the landing docs to offer way too much information, when all I really needed to do was join a security group and navigate to a website. I think it's useful to offer this information for people who want to understand their options more, but I'd like to also direct link to the preferred options to reduce the mental overhead required for a likely already frustrated user.

benbp: I agree with removing the workstation, it seems to be pretty much identical to the jumpbox experience?

mikeharder: Feel free to edit my suggestion, but at a minimum I think we should change the following:

1. Do not recommend "AVD Workstation".  I think the other two options are strictly better.
2. To minimize confusion, use the exact same names listed at https://microsoft.sharepoint.com/sites/Identity_Authentication/SitePages/ModernAccess/About-Modern-Access.aspx
3. Do not say DevBox will be billed to your team.  It's currently not true, though it may become true at some point in the future.

mikeharder (approved):

PR #6384, opened 2023-06-22
Support ShouldPublishSymbols param in pipeline and set it to false for tool net-changelog-gen 

RodgeFu:
* Support ShouldPublishSymbols param in pipeline with default value "true"
* Set it to false for net-changelog-gen because the tool needs to be PackAsTool and has some native binaries in its dependency which would be complained by the PublishSymbol step for missing BuildId for these native binaries. It's fine to not publish symbols for the tool.


weshaggard:
I triggered a publish run at https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2866365&view=logs&j=6d7966cd-0423-5022-78e0-a1ae4dd99f62

weshaggard (approved):

PR #6391, opened 2023-06-22
Update API view parser version for Java

praveenkuttappan:
Increment Java API review parser to auto upgrade existing reviews.


praveenkuttappan:
@JonathanGiles FYI...

benbp (approved):

JonathanGiles:
Does the version in the JSON output also need incrementing? I'm not sure when that is required.

praveenkuttappan:
We should ideally set that value to parser value but not required at the moment. APIView overrides the value currently based on parser version set within the APIView.

PR #6405, opened 2023-06-26
[Pylint] allow for spacing in type

l0lawrence:


tools/pylint-extensions/azure-pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
```diff
@@ -3441,3 +3441,29 @@ def function_foo(*x, y, z):
...
         ):
             self.checker.visit_functiondef(node)
+
+    def test_docstring_type_has_space(self):
+        # Don't error if there is extra spacing in the type
+        node = astroid.extract_node(
+            """
+            def function_foo(x):
+                '''
+                :param dict[str, int] x: x
```
kristapratico, line 3451: I want to check one thing before we merge this - I don't think sphinx renders the type correctly in the docs when it's inline like this and has extra spaces. If I remembered correctly, we probably shouldn't allow this and raise a better error message.

l0lawrence: https://learn.microsoft.com/en-us/python/api/azure-core/azure.core.polling.base_polling.operationresourcepolling?view=azure-python if you look at `lro_options` here it generates with the extra space correctly. I believe this is off of sphinx but I could be wrong

kristapratico: Nice.

kristapratico (approved):

PR #6445, opened 2023-06-30
[Pylint] Infer raise

l0lawrence:
It seems that raise exceptions as inferred as inferrable when alone in a function:

def func()
  raise ValueError --- uniferable

def func()
   print("waht")
   raise ValueError ---- None

I believe this is a bug in astroid that we can work-around with this fix

### DIFF
--- a/tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
+++ b/tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
@@ -1384,7 +1384,7 @@ class CheckDocstringParameters(BaseChecker):
         function_decorators = node.decoratornames()
         try:
             returns = next(node.infer_call_result()).as_string()
-            if returns == "None":
+            if returns == "None" or returns == astroid.Uninferable:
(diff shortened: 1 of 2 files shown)


tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -1384,7 +1384,8 @@ def check_return(self, node):
         function_decorators = node.decoratornames()
         try:
             returns = next(node.infer_call_result()).as_string()
-            if returns == "None":
+            # If returns None, or raises an error ignore
+            if returns == "None" or returns == astroid.Uninferable:
```
kristapratico, line 1388: we call as_string() on L1386, is `returns` always string or does the == `astroid.Uninferable` comparison work?

l0lawrence: Calling as_string() on Uninferable returns Uninferable

kristapratico (approved):

PR #6446, opened 2023-07-01
Add generated files to gitignore so we don't accidentally commit them…

scbedd:
… or need to clean them up

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -496,6 +496,16 @@ tools/oav-traffic-converter/input-example/
 tools/oav-traffic-converter/output-example/
 tools/oav-traffic-converter/build/
 
+# assets reporting
+tools/assets-automation/assets-reporting/generated/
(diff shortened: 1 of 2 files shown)


mikeharder (approved):

benbp (approved):

benbp:
IIRC Wes has preferred to keep things in a root .gitignore as a repo convention, not sure if he would hold that same opinion here.

weshaggard:
I generally do prefer not having nested gitignore files as the don't chain together nicely. I would suggest putting it in the root one.

scbedd:
> I generally do prefer not having nested gitignore files as the don't chain together nicely. I would suggest putting it in the root one.

I've added a commit that moves the rules established in the new `.gitignore` to the base `.gitignore`. Thanks for the feedback.

PR #6458, opened 2023-07-06
Fix Playback Discarding Pending Changes

scbedd:
The proxy maintains a tiny dictionary of `<assets-json-location>: <target-tag>`

We use this little dictionary to [short circuit](https://github.com/Azure/azure-sdk-tools/blob/b14dd081686ceb639049a2e0a7762eb7d179da27/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs#L324) our checkout process. However, when I refactored to allow automatic checkout to discard pending changes, I didn't properly account for loading this dictionary into memory. This caused us to always treat even the same tag that was currently checked out as a `new` tag, which causes us to discard changes when `restoring`.

By quickly refreshing what the proxy sees at startup, we can avoid this "first restore" issue causing us to discard.


scbedd:
Closing this PR as it accidentally includes changes that are not ready for primetime.

PR #6468, opened 2023-07-07
Update typespec common scripts

raych1:
What are the changes?
- doc update
- support sync script and generate script call when it's new sdk project
- remove duplicated 'specification' in `localSpecRepoPath` (reverted this point)

@weshaggard , please review.

### DIFF
--- a/eng/common/scripts/TypeSpec-Project-Sync.ps1
+++ b/eng/common/scripts/TypeSpec-Project-Sync.ps1
@@ -103,7 +103,9 @@ $specSubDirectory = $configuration["directory"]
 
 # use local spec repo if provided
 if ($LocalSpecRepoPath) {
-  $specCloneDir = $LocalSpecRepoPath
+  $LocalSpecRepoPath = $LocalSpecRepoPath.Replace("\", "/")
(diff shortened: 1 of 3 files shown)


eng/common/scripts/TypeSpec-Project-Sync.ps1
```diff
@@ -103,7 +103,9 @@ $specSubDirectory = $configuration["directory"]
 
 # use local spec repo if provided
 if ($LocalSpecRepoPath) {
-  $specCloneDir = $LocalSpecRepoPath
+  $LocalSpecRepoPath = $LocalSpecRepoPath.Replace("\", "/")
+  # remove duplicated 'specification' in path
+  $specCloneDir = $LocalSpecRepoPath.Replace("/specification","")
```
weshaggard, line 108: What cases do we have where there is duplicate "/specification" in the path? Should we fix those instead of adding this here? If we have a good reason we cannot fix the other places then lets add a comment here about why we need to do this.

weshaggard (approved):

weshaggard (approved):

PR #6496, opened 2023-07-12
Update Azure.ClientSdk.Analyzers to fix incorrect behaviors

pshao25:
When updating Azure.ClientSdk.Analyzers in azure-sdk-for-net to uptake https://github.com/Azure/azure-sdk-tools/pull/6156, there are some errors not taken into consideration before:
- Some methods should not check AZC0004: DO provide both asynchronous and synchronous variants for all service methods. Examples: constructors, properties, GetSubClients methods.
- Missing valid return type in the previous PR. `Operation<Pageable<T>>` should also be a valid return type for protocol methods.
- The scenarios described in https://github.com/Azure/autorest.csharp/issues/3571, we will change logic to https://github.com/Azure/autorest.csharp/issues/3571#issuecomment-1635227389

After this PR, will do https://github.com/Azure/azure-sdk-for-net/pull/37037

Unblocks https://github.com/Azure/azure-sdk-for-net/pull/37447

Overall, together with https://github.com/Azure/azure-sdk-tools/pull/6156, the changes are:
1. AZC0004: From "if there is an async method then it should have corresponding sync method" to "if there is an async method then it should have corresponding sync method, and if there is a sync method it should have corresponding async mthod".
2. AZC0017: Do ensure convenience method not take RequestContent as parameter type.
3. AZC0018: Do ensure protocol method take a RequestContext parameter called context and not take models as parameter type or return type.
4. AZC0019: Check ambiguity for protocol method.


m-nash (approved):

pallavit:
@pshao25 Could you please update the PR description with what are the unexpected behaviors?

jsquire:
+1 to Pallavi's request.  Given that this change will impact everyone working in the repository, let's please leave detailed context so that we can understand and troubleshoot via search in the future if need be.

jsquire (changes requested):
Blocking merge until we have more context to understand this PR and the wider team has a chance to offer feedback.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -185,19 +187,44 @@ private static void CheckProtocolMethodParameters(ISymbolAnalysisContext context
...
             // Optional RequestContent or required RequestContent is checked in AnalyzeCodeBlock.
         }
 
-        // A protocol method should not have model as type. Accepted return type: Response, Task<Response>, Pageable<BinaryData>, AsyncPageable<BinaryData>, Operation<BinaryData>, Task<Operation<BinaryData>>, Operation, Task<Operation>
+        // A protocol method should not have model as type. Accepted return type: Response, Task<Response>, Pageable<BinaryData>, AsyncPageable<BinaryData>, Operation<BinaryData>, Task<Operation<BinaryData>>, Operation, Task<Operation>, Operation<Pageable<BinaryData>>, Task<Operation<AsyncPageable<BinaryData>>>
         private static void CheckProtocolMethodReturnType(ISymbolAnalysisContext context, IMethodSymbol method)
         {
+            bool IsValidPageable(ITypeSymbol typeSymbol)
+            {
+                if (typeSymbol.Name != "Pageable" && typeSymbol.Name != "AsyncPageable")
```
jsquire, line 208: This feels brittle.  What happens if it's a derrived type or we rename?   At the very least, let's please abstract all these magic strings into a constants file.   This set, for example, is used at least three times.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
```diff
@@ -120,8 +120,8 @@ internal class Descriptors
 
         public static DiagnosticDescriptor AZC0018 = new DiagnosticDescriptor(
             nameof(AZC0018),
-            "Do ensure protocol method take a RequestContext parameter called context and not take models as parameter type or return type.",
-            "Protocol method should have requestContext as the last parameter and don't have model as parameter type or return type.",
+            "Do ensure protocol method take a RequestContext parameter called context and not take models as parameter type or return type. Do ensure protocol method will not cause ambiguous call with convenience method.",
```
jsquire, line 123:
```suggestion
            "Do ensure protocol methods take a RequestContext parameter called `context` and do not take models as parameter or return types. Do ensure protocol methods will not cause an ambiguous overload resolution with convenience methods.",

```

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
```diff
@@ -120,8 +120,8 @@ internal class Descriptors
 
         public static DiagnosticDescriptor AZC0018 = new DiagnosticDescriptor(
             nameof(AZC0018),
-            "Do ensure protocol method take a RequestContext parameter called context and not take models as parameter type or return type.",
-            "Protocol method should have requestContext as the last parameter and don't have model as parameter type or return type.",
+            "Do ensure protocol method take a RequestContext parameter called context and not take models as parameter type or return type. Do ensure protocol method will not cause ambiguous call with convenience method.",
+            "Protocol method should have requestContext as the last parameter and don't have model as parameter type or return type. Protocol method should not have optional parameters if ambiguity exists between protocol method and convenience method.",
```
jsquire, line 124:
```suggestion
            "Protocol method should have `requestContext` as the last parameter and not use a model as parameter or return types. Protocol methods should not have optional parameters if ambiguity exists between the protocol method and convenience methods.",

```

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
jsquire: Not blocking for this PR, but let's please open an issue or follow-up to refactor this.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAnalyzerBase.cs
```diff
@@ -26,16 +26,26 @@ public override void Analyze(ISymbolAnalysisContext context)
...
             public static ParameterEquivalenceComparer Default { get; } = new ParameterEquivalenceComparer();
 
-            public bool Equals(IParameterSymbol x, IParameterSymbol y)
+            public virtual bool Equals(IParameterSymbol x, IParameterSymbol y)
             {
                 return x.Name.Equals(y.Name) && TypeSymbolEquals(x.Type, y.Type) && x.IsOptional == y.IsOptional;
             }
 
-            private bool TypeSymbolEquals(ITypeSymbol x, ITypeSymbol y)
+            protected bool TypeSymbolEquals(ITypeSymbol x, ITypeSymbol y)
```
jsquire, line 48: We should consider making this static or virtual, depending on whether we see a need to override in the future.

jsquire (approved):

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
pshao25: Why a protocol method would return a derived type or we would return? They are all both breaking change.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAnalyzerBase.cs
pshao25: I would prefer not to change it until we see a real need.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
pshao25: Already fixed in this PR.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAnalyzerBase.cs
jsquire: If not virtual, there's no benefit to this being an instance member and no risk to elevating it.    Unless we have a reason to keep it from being overridden, let's make it virtual, please.

pshao25: Makes sense. Updated.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
```diff
@@ -120,8 +120,8 @@ internal class Descriptors
 
         public static DiagnosticDescriptor AZC0018 = new DiagnosticDescriptor(
             nameof(AZC0018),
-            "Do ensure protocol method take a RequestContext parameter called context and not take models as parameter type or return type.",
-            "Protocol method should have requestContext as the last parameter and don't have model as parameter type or return type.",
+            "Do ensure protocol methods take a RequestContext parameter called `context` and do not take models as parameter or return types. Do ensure protocol methods will not cause an ambiguous overload resolution with convenience methods.",
```
annelo-msft, line 123: I believe this is the warning "title", so ideally it would be shorter.  Detail can be put in the messageFormat.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
```diff
@@ -120,8 +120,8 @@ internal class Descriptors
 
         public static DiagnosticDescriptor AZC0018 = new DiagnosticDescriptor(
             nameof(AZC0018),
-            "Do ensure protocol method take a RequestContext parameter called context and not take models as parameter type or return type.",
-            "Protocol method should have requestContext as the last parameter and don't have model as parameter type or return type.",
+            "Do ensure protocol methods take a RequestContext parameter called `context` and do not take models as parameter or return types. Do ensure protocol methods will not cause an ambiguous overload resolution with convenience methods.",
+            "Protocol method should have `requestContext` as the last parameter and not use a model as parameter or return types. Protocol methods should not have optional parameters if ambiguity exists between the protocol method and convenience methods.",
```
annelo-msft, line 124: Apologies if I'm confused, but isn't the name of the parameter `context` and not `requestContext`?

annelo-msft, line 124: > Protocol methods should not have optional parameters if ambiguity exists between the protocol method and convenience methods.

Would it make sense to say that `context` must be required if the convenience method creates an ambiguous overload?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
annelo-msft, line 123: Would it make sense to have different errors for (1) requiring `context` and prohibiting model types and (2) identifying ambiguous overloads with convenience methods?  As a developer, I might want to handle these separately.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -184,120 +193,152 @@ private static void CheckProtocolMethodParameters(ISymbolAnalysisContext context
...
 
-            var requestContent = method.Parameters.FirstOrDefault(p => p.Type.Name == "RequestContent");
-            if (requestContent == null)
+            var requestContent = method.Parameters.FirstOrDefault(IsRequestContent);
+            if (requestContent == null && method.Parameters.Last().IsOptional)
             {
-                if (method.Parameters.LastOrDefault().IsOptional)
+                INamedTypeSymbol type = (INamedTypeSymbol)context.Symbol;
+                var methodList = type.GetMembers(method.Name).OfType<IMethodSymbol>().Where(member => !SymbolEqualityComparer.Default.Equals(member, method));
+                var convenienceMethod = FindMethod(methodList, method.TypeParameters, method.Parameters.RemoveAt(method.Parameters.Length - 1), symbol => IsCancellationToken(symbol), ParameterEquivalenceComparerOptionalIgnore.Default);
```
annelo-msft, line 201: It would be helpful to break this statement into multiple lines.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -184,120 +193,152 @@ private static void CheckProtocolMethodParameters(ISymbolAnalysisContext context
...
             }
 
-            var requestContent = method.Parameters.FirstOrDefault(p => p.Type.Name == "RequestContent");
-            if (requestContent == null)
+            var requestContent = method.Parameters.FirstOrDefault(IsRequestContent);
+            if (requestContent == null && method.Parameters.Last().IsOptional)
             {
-                if (method.Parameters.LastOrDefault().IsOptional)
+                INamedTypeSymbol type = (INamedTypeSymbol)context.Symbol;
+                var methodList = type.GetMembers(method.Name).OfType<IMethodSymbol>().Where(member => !SymbolEqualityComparer.Default.Equals(member, method));
```
annelo-msft, line 200: nit: please use type name instead of `var`

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -184,120 +193,152 @@ private static void CheckProtocolMethodParameters(ISymbolAnalysisContext context
...
-
         }
 
+        private bool IsCheckExempt(IMethodSymbol method)
+        {
+            return method.MethodKind != MethodKind.Ordinary ||
+                method.DeclaredAccessibility != Accessibility.Public ||
+                method.OverriddenMethod != null ||
+                method.IsImplicitlyDeclared ||
+                (method.Name.StartsWith("Get") && method.Name.EndsWith(ClientSuffix));
```
annelo-msft, line 332: There may be some subclients that don't have the `Client` suffix: https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-subclient-factory-methods-suffix

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -184,120 +193,152 @@ private static void CheckProtocolMethodParameters(ISymbolAnalysisContext context
...
                 originalType.Name.EndsWith(ClientSuffix))
             {
                 return;
             }
 
             context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0015, method.Locations.FirstOrDefault(), originalType.ToDisplayString()), method);
-
         }
 
+        private bool IsCheckExempt(IMethodSymbol method)
```
annelo-msft, line 326: Curious: How did we decide on this logic?  It makes me a little nervous to start exempting a large group of methods from analysis checks.  How do we know we won't accidentally exempt something we previously checked?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
```diff
@@ -496,5 +496,54 @@ await Verifier.CreateAnalyzer(code)
...
+            _id = id;
+        }
+    }
+}";
+            await Verifier.CreateAnalyzer(code)
+                .RunAsync();
+        }
+
+        [Fact]
+        public async Task AZC0004NotProducedForGetSubClientMethod()
```
annelo-msft, line 528: Could we ever have a service method that ends with the `Client` suffix?  It feels more robust to me to check that the return type is a class in the library.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAnalyzerBase.cs
```diff
@@ -26,16 +26,26 @@ public override void Analyze(ISymbolAnalysisContext context)
             AnalyzeCore(context);
         }
 
+        protected class ParameterEquivalenceComparerOptionalIgnore : ParameterEquivalenceComparer
```
annelo-msft, line 29: Curious: why do we want to compare parameters but ignore whether they are optional?  This feels like something we might want to handle through a separate code path to make it extra clear what we're doing, for anyone handling maintenance of this analyzer in the future.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
pshao25: If `context` is required then every parameters should be required. I think it would be more specific to say all the parameters should be required.

pshao25: You are correct.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
pshao25: I didn't find one in azure-sdk-for-net. Do you have one?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
pshao25: We are using this rule to make sure service method is not returning arbitrary thing. Think about this method:
```C#
void ABC method(){}
```
How could you tell ABC is a sub client or an invalid return?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAnalyzerBase.cs
pshao25: Because we are looking for the convenience method of a protocol method. Therefore, it might be not consistence between their parameters that one might be optional and the other one is required.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
pshao25: I added a new rule AZC0019 for identifying ambiguity.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
annelo-msft: Nit: it might be a little easier to understand to phrase the last sentence as "If ambiguity exists between the protocol method and the convenience method, all parameters of the protocol method should be required." (Instead of "Protocol methods should not have optional parameters if ambiguity exists between the protocol method and convenience methods.")

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
annelo-msft: I didn't have a specific definition in mind, but here are some thoughts - a subclient would have service methods on it in most cases?  So if a return type has service methods, we could have high confidence it is a subclient.

Do we ever have service methods that return types that are not: Response, Response\<T\>, Pageable\<T\>, Operation, Operation\<T\>?  This might be a simple way to differentiate between service methods and subclient methods.

pshao25: We should keep this in mind: we first get a method, then we start to check it. We check whether its input is valid and its return is valid. You could see that in my code, I also make sure the return type should be Response, Response<T>, Pageable<T>...etc...

In my example, it returns ABC and you say it is not a service method. Then what? We regard it as a subclient method and let it pass check? Doesn't it mean you don't even need to check it?

> So if a return type has service methods, we could have high confidence it is a subclient.

It is an infinite recursion, I'm afraid.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
pshao25: Oh sorry I missed that guideline link above. Then there do have defect here.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
pshao25: Ordinary: means it is not a constructor, not a property, etc

OverriddenMethod: below is an example. Think if service method is an overridden method, then it means client should inherited from others. I never see such case. So I'm just exempting the overridden methods from `object`.
image

IsImplicitlyDeclared: it was added before and I don't want to touch it. From comments I believe it is used to exempt the methods automatically created by the compiler, for example, default constructor.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
pshao25: I move this sentence to AZC0019, since AZC0019 is used to detect ambiguity.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAnalyzerBase.cs
pshao25: I was wrong. When we try to look for the convenience method, the requestcontext must be optional. Therefore we don't need to introduce the OptionalIgnore comparer. I removed it.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0002Tests.cs
```diff
@@ -13,19 +13,21 @@ public class AZC0002Tests
...
             const string code = @"
+using Azure;
 using System.Threading.Tasks;
 
 namespace RandomNamespace
 {
     public class SomeClient
     {
-        public virtual Task {|AZC0002:GetAsync|}()
+        public virtual Response {|AZC0002:GetAsync|}()
```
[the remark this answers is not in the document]
annelo-msft: (Please apply to other test cases as well)

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0002Tests.cs
```diff
@@ -38,20 +40,22 @@ await Verifier.CreateAnalyzer(code)
...
+using Azure;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace RandomNamespace
 {
     public class SomeClient
     {
-        public virtual Task {|AZC0002:GetAsync|}(CancellationToken cancellation = default)
+        public virtual Response {|AZC0002:GetAsync|}(CancellationToken cancellation = default)
```
annelo-msft, line 51: Would it make sense for tests of convenience methods to use `Response<Model>` instead of just response, to better simulate a real client? Should we include tests for `Pageable<Model>` and `Operation<Model>` as well?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
```diff
@@ -229,8 +233,15 @@ public async Task AZC0004NotProducedForGenericMethodsWithSyncAlternative()
 
 namespace RandomNamespace
 {
+public class CapacityReservationCollection{
```
annelo-msft, line 236: Is this added to test subclient factory methods?  It would be good to indicate that in the test name, or a comment somewhere?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
```diff
@@ -323,32 +336,6 @@ await Verifier.CreateAnalyzer(code)
                 .RunAsync();
         }
 
-        [Fact]
-        public async Task AZC0004NProducedForMethodsWithoutArgMatchedSyncAlternative()
```
annelo-msft, line 327: Why is this test being removed?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0015Tests.cs
```diff
@@ -9,7 +9,7 @@ namespace Azure.ClientSdk.Analyzers.Tests
 {
     public class AZC0015Tests
     {
-        [Theory]
+        [Theory(Skip = "We use return type as an exit criteria to check method, so this test set is not applicable until we find a better criteria.")]
```
annelo-msft, line 12: Please don't skip tests - which of these are valid/invalid?  If a different analyzer is catching these errors now, would it make sense to update the test to detect both?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0018Tests.cs
```diff
@@ -21,8 +21,20 @@ public async Task AZC0018NotProducedForCorrectReturnType()
 
 namespace RandomNamespace
 {
+public class AssetConversion {}
```
annelo-msft, line 24: Curious: why is this added?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0018Tests.cs
```diff
@@ -193,29 +215,28 @@ await Verifier.CreateAnalyzer(code)
         }
 
         [Fact]
-        public async Task AZC0018ProducedForMethodsWithParameterModel()
+        public async Task AZC0018ProducedForMethodsWithOperationOfPageableModel()
```
annelo-msft, line 218: Would it be possible to preserve the existing test and add a new one for OperationOfPageable, instead of replacing the existing test?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0018Tests.cs
```diff
@@ -193,29 +215,28 @@ await Verifier.CreateAnalyzer(code)
...
             const string code = @"
 using Azure;
 using Azure.Core;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Collections.Generic;
 
 namespace RandomNamespace
 {
-    public struct Model
```
annelo-msft, line 207: Same comment about keeping the old test and adding a new one.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0018Tests.cs
```diff
@@ -264,33 +289,17 @@ public async Task AZC0018ProducedForMethodsWithRequiredRequestContentAndRequired
 using System.Threading.Tasks;
 using System.Collections.Generic;
 
-namespace Azure.Core
```
annelo-msft, line 267: Why is this removed?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAnalyzerBase.cs
```diff
@@ -91,7 +91,6 @@ protected static IMethodSymbol FindMethod(IEnumerable<IMethodSymbol> methodSymbo
 
             return methodSymbols.SingleOrDefault(symbol =>
             {
-
```
annelo-msft, line 94: If there are no changes to this file, it would be better to revert it so it doesn't appear in the list of changed files.  This makes it easier to understand how the repo has changed in the future.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/SymbolAnalyzerBase.cs
```diff
@@ -16,7 +16,7 @@ public abstract class SymbolAnalyzerBase : DiagnosticAnalyzer
 
         protected INamedTypeSymbol ClientOptionsType { get; private set; }
 
-        public override void Initialize(AnalysisContext context)
+        public sealed override void Initialize(AnalysisContext context)
```
annelo-msft, line 19: Curious: why make this sealed?

src/dotnet/Azure.ClientSdk.Analyzers/readme.md
```diff
@@ -0,0 +1,7 @@
+## How to test WIP Azure.ClientSdk.Analyzers against azure-sdk-for-net
```
annelo-msft, line 1: Thank you for this, by the way - I am using this guidance today, and very grateful to have it!

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0002Tests.cs
pshao25: It is just unit test and we are not testing the response. What's the value we could gain to change to `Response<Model>`? 

> Should we include tests for Pageable<Model> and Operation<Model>

We have a lot of cases in AZC0018Tests.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
pshao25: It was added accidently. Removed.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0002Tests.cs
```diff
@@ -285,5 +295,43 @@ await Verifier.CreateAnalyzer(code)
...
+        }
+
+        public virtual Response Get(string s, CancellationToken cancellationToken)
+        {
+            return null;
+        }
+    }
+}";
+            await Verifier.CreateAnalyzer(code)
+                .WithDisabledDiagnostics("AZC0015")
```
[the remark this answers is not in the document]
pshao25: Just copied from existing code. Removed.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
pshao25: Because in the previous discussion it is invalid. But now it comes to be valid. Added back.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0015Tests.cs
pshao25: All of these cases, we don't check them anymore. ANY method without valid return type, analyzer just skips check. So these tests are invalid until we find a better way to tell the methods we want to check. This is what you brought up, and @m-nash agrees with you.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0018Tests.cs
pshao25: Accident. Removed.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/SymbolAnalyzerBase.cs
pshao25: It was `sealed` before. I removed it in my previous PR to do some change. Now this change is revert so the `sealed` need to be added back.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0018Tests.cs
pshao25: Because we change the logic to detect ambiguity. AssetNotNull is not needed any more to detect ambiguity. Current logic is in https://github.com/Azure/azure-sdk-for-net/pull/37795/files

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0018Tests.cs
pshao25: AZC0018ProducedForMethodsWithParameterModel still exists below. I think it is a diff issue.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0018Tests.cs
pshao25: Same above. I think I'm not deleting.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
```diff
@@ -213,57 +207,50 @@ public virtual Response Get(CancellationToken cancellationToken = default)
...
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace RandomNamespace
 {
-public class CapacityReservationCollection{
-}
     public class SomeClient
     {
-        public virtual CapacityReservationCollection GetCapacityReservations()
```
annelo-msft: Are we still able to detect and verify that subclient methods need to be virtual?

annelo-msft: (Being virtual is [required for mocking](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-mocking-virtual-getclient-method), so we need to be able to enforce it for subclient methods as well)

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
```diff
@@ -296,50 +283,76 @@ public virtual Response Get(CancellationToken cancellationToken = default)
...
         }
     }
 }";
             await Verifier.CreateAnalyzer(code)
-                .WithDisabledDiagnostics("AZC0015")
                 .RunAsync();
         }
 
         [Fact]
+        public async Task AZC0004NProducedForMethodsWithoutArgMatchedSyncAlternative()
```
annelo-msft, line 327: Nit: is the name of this method `NotProduced` or `Produced`?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0018Tests.cs
```diff
@@ -21,8 +21,6 @@ public async Task AZC0018NotProducedForCorrectReturnType()
 
 namespace RandomNamespace
 {
-public class AssetConversion {}
```
annelo-msft: Curious: why is this being removed?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
annelo-msft: Thanks!

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0015Tests.cs
annelo-msft: It looks like the result of skipping both these tests is that we don't have any tests for AZ0015, which isn't a good place for us to be for this analyzer.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0018Tests.cs
pshao25: I'm not removing. I think you are not seeing the right iteration.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
pshao25: I think it should be `NotProduced`. I'm just adding back the deleted one which called itself `NProduced`. Updated to `NotProduced`.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0004Tests.cs
pshao25: Two things here:
1. I'm not deleting anything. I think you should see all the comments together, not just one iteration.
2. Subclient cannot be checked any more because its return not is not one of Operation, Pageable, Response.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0002Tests.cs
```diff
@@ -32,7 +32,6 @@ public class SomeClient
     }
 }";
             await Verifier.CreateAnalyzer(code)
-                .WithDisabledDiagnostics("AZC0015")
```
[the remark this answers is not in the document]
pshao25: It just means we don't need to suppress AZC0015. That's it.

AZC0015 is not effective now.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0015Tests.cs
pshao25: AZ0015 just not effective now. 
image

annelo-msft: Ah, I didn't realize that implementing it this way made AZC0015 completely non-functional.

I think AZC0015 is a useful diagnostic error to have running in azure-sdk-for-net.  For example, if I write a method that accidentally returns `Task<Model>` where I meant to return `Task<Response<Model>>`, I would like a warning to help me find this, and not rely on a human architect catching my mistake.

Is it possible to change the implementation for AZC0018 and AZC0019 so that AZC0015 is still functional?  I realize that a lot of services have [opted-out of this warning](https://grep.app/search?current=2&q=AZC0015&filter[repo][0]=Azure/azure-sdk-for-net), we should see if we can remove these where it's possible, but that doesn't have to be part of this PR.

annelo-msft (changes requested):
This PR is changing a lot of existing analyzers, where I think we should preserve their functionality.  Would it be possible to reduce its scope so that it only adds the protocol-method related analyzers without changing other ones?  If we see the need to change existing analyzers to improve what's happening in azure-sdk-for-net, that is great and we should do that, but I think it would be better to file issues for each one we want to make changes to and handle them in separate PRs.

pshao25:
Already implemented this.

PR #6498, opened 2023-07-12
Add merging to more branches in specs repo

weshaggard:
Add ARMCoreRPDev and InternalARMContracts long lived branches to the set of branches to keep updated.


weshaggard:
mirror test run https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2913275&view=logs&j=b5398758-6341-5a90-e082-88d59643ae40

PR #6517, opened 2023-07-14
Add recording file location

billwert:
There are tests in the Java repo which parse the recording files for various reasons. Add a header on the response of playback start to share that location.

I also noticed that the tests were using the wrong assertion for headers, so I changed it to check for string empty/null. (Even null is unnecessary - could be a test against `String.Empty` if you prefer.)

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy.Tests/RecordTests.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy.Tests/RecordTests.cs
@@ -35,7 +35,7 @@ namespace Azure.Sdk.Tools.TestProxy.Tests
             };
             await controller.Start();
             var recordingId = httpContext.Response.Headers["x-recording-id"].ToString();
-            Assert.NotNull(recordingId);
+            Assert.False(string.IsNullOrEmpty(recordingId));
             Assert.True(testRecordingHandler.RecordingSessions.ContainsKey(recordingId));
         }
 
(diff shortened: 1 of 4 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy.Tests/RecordTests.cs
```diff
@@ -62,9 +62,9 @@ public async Task TestStartRecordInMemory()
 
         [Theory]
         [InlineData("recordings/TestStartRecordSimple_nosave.json", "request-response")]
-        [InlineData("recordings/TestStartRecordSimpl�_nosave.json", "request-response")]
+        [InlineData("recordings/TestStartRecordSimplé_nosave.json", "request-response")]
         [InlineData("recordings/TestStartRecordSimple.json", "")]
-        [InlineData("recordings/TestStartRecordSimpl�.json", "")]
+        [InlineData("recordings/TestStartRecordSimplé.json", "")]
```
billwert, line 67: I just noticed this - VS must have "helped." Should this get reverted?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -372,7 +372,7 @@ public async Task StartPlaybackAsync(string sessionId, HttpResponse outgoingResp
             {
                 await RestoreAssetsJson(assetsPath, true);
                 var path = await GetRecordingPath(sessionId, assetsPath);
-
+                outgoingResponse.Headers.Add("x-recording-file-location", path);
```
scbedd, line 375: We should probably do the same in `StartRecordingAsync` for consistency right?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
```diff
@@ -351,6 +351,8 @@ To start an individual test recording or playback, users will POST to a route on
 
 You will receive a recordingId in the reponse under header `x-recording-id`. This value should be included under header `x-recording-id` in all further requests.
 
+For playback, you will receive the location of the recording file (if it is on disk) in the header `x-recording-file-location`.
```
scbedd, line 354: If we're going to do this we might as well do it for both `Record/Start` and `Playback/Start` I think.

scbedd (approved):
Resolve the test and I'll approve. I see where you're coming from.

scbedd (approved):
Thanks @billwert

PR #6522, opened 2023-07-14
Support .NET 7

scbedd:
By producing a tool that will run on the same runtime.


mikeharder:
I defer to @weshaggard who knows more about .NET targeting

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Azure.Sdk.Tools.TestProxy.csproj
```diff
@@ -21,6 +21,11 @@
     <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="6.0.0" />
   </ItemGroup>
   
+  <ItemGroup Condition="'$(TargetFramework)'=='net7.0'">
```
weshaggard, line 24: Doesn't look like we need these split out any longer and given we don't support .NET 5.0 any longer we can probably remove that group.

weshaggard:
@scbedd can you give more context on why .NET 7 is interesting for test-proxy?

PR #6541, opened 2023-07-19
Explore why we don't get break on non-ascii header

scbedd:
@billwert 

I have no explanation for why this doesn't repro the failure we see over in [.net here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=2929862&view=ms.vss-test-web.build-test-results-tab&runId=42030283&resultId=108599&paneView=debug).


mikeharder (approved):

scbedd:
Closing. Going to do an integration test.

PR #6543, opened 2023-07-19
Make x509 certificate script from azure-sdk-for-net common to repos

benbp:
Copying from https://github.com/Azure/azure-sdk-for-net/blob/main/eng/scripts/X509Certificate2/X509Certificate2.psm1 so that we can share cert generating functionality.

Thanks for the suggestion @conniey

### DIFF
--- a/eng/common/scripts/X509Certificate2/README.md
+++ b/eng/common/scripts/X509Certificate2/README.md
@@ -1,0 +1,24 @@
+Powershell module for generating self-signed x509 certificates
+
+Usage:
+
+```powershell
+Import-Module -Name ./eng/common/scripts/X509Certificate2 # assumes $PWD is repo root
+
+$cert1 = New-X509Certificate2 -SubjectName 'E=opensource@microsoft.com, CN=Azure SDK, OU=Azure SDK, O=Microsoft, L=Redmond, S=WA, C=US' -ValidDays 3652
+
+$CaPublicKeyBase64 = $cert1 | Format-X509Certificate2 -Type CertificateBase64
+$CaPrivateKeyPem = $cert1 | Format-X509Certificate2 -Type Pkcs1
+$CaKeyPairPkcs12Base64 = $cert1 | Format-X509Certificate2 -Type Pkcs12Base64
+```
+
+With V3 extensions
+
+```powershell
+Import-Module -Name eng/scripts/X509Certificate2.psm1 # assumes $PWD is repo root
+
+$cert2 = New-X509Certificate2 -SubjectName 'CN=Azure SDK' -SubjectAlternativeNames (New-X509Certificate2SubjectAlternativeNames -EmailAddress azuresdk@microsoft.com) -KeyUsageFlags KeyEncipherment, NonRepudiation, DigitalSignature -CA -TLS -ValidDays 3652
+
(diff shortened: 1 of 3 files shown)


conniey (approved):
🥂 lgtm

eng/common/scripts/X509Certificate2/README.md
```diff
@@ -0,0 +1,24 @@
+Powershell module for generating self-signed x509 certificates
+
+Usage:
+
+```powershell
+Import-Module -Name eng/scripts/X509Certificate2.psm1 # assumes $PWD is repo root
```
weshaggard, line 6: Does this have to be a module? In general we have moved away from modules do to loading issues and just use a script with helper functions.

weshaggard (approved):

eng/common/scripts/X509Certificate2/README.md
benbp: It was developed that way, but I'm also happy to spend a few minutes turning it into a cmdlet instead.

weshaggard: A cmdlet is a module right? At any rate I'm OK with this for now and if we hit any issues we can reevaluate.

benbp: I think of top-level scripts as cmdlets

heaths: Modules contain 1 or more cmdlets (or none, actually; could just be type or view data). Scripts can be cmdlets. Modules can be unloaded, and because these are all a bunch of related cmdlets I didn't want a bunch of one-off scripts that are harder to keep in sync. In my experience, I've never had problems with modules. They solved the fundamental problems with snapins.

christothes:
Is there any plan to delete the .net repo script that this replaces, or work items to track it?

eng/common/scripts/X509Certificate2/README.md
weshaggard: We have definitely had issues with module loading and unloading, especially during development of them as you often times need to restart your shell to unload successfully. We also have issues when loading modules in parallel when running steps which causes race conditions and other interesting things. That is why we moved to having a set of shared helper scripts/functions instead of modules. That said if I don't have concerns with keeping this as a module if it is already working well as one.

benbp:
> Is there any plan to delete the .net repo script that this replaces, or work items to track it?

@christothes yes after this is merged I was going to delete the NET repo script. Is this part of people's workflows and/or other documentation that I should update? (e.g. should I leave a redirect note in the net/eng directory, or can I just delete it entirely)

christothes:
> @christothes yes after this is merged I was going to delete the NET repo script. Is this part of people's workflows and/or other documentation that I should update? (e.g. should I leave a redirect note in the net/eng directory, or can I just delete it entirely)

Thanks! - [here](https://grep.app/search?q=New-x509certificate2&filter[repo][0]=Azure/azure-sdk-for-net&filter[lang][0]=PowerShell) are the ones I am aware of

christothes (approved):

benbp:
@christothes PR for net [here](https://github.com/Azure/azure-sdk-for-net/pull/37812). Not all the references are changed because a few services have implemented their own function with the same name.

PR #6545, opened 2023-07-19
[GitHub Scripts] Recreate label tools

jsquire:
# Summary

The focus of these changes is to rebuild the scripts for label management using the GH CLI.  These had previously been using the now retired GHCreator tool.  As part of these changes, the old scripts have been removed.


tools/github/README.md
```diff
@@ -45,6 +56,38 @@ This area is focused on the management of the Azure SDK GitHub repositories, con
...
+   ./Sync-AzsdkLabels.ps1 
+
+   # Synchronize the common labels to the Azure SDK for .NET repository.
+   ./Sync-AzsdkLabels.ps1 -LabelsFilePath "../data/common-labels.csv" -Languages 'net' 
+   
+   # View the help for the full set of parameters.
+   get-help ./Sync-AzsdkLabels.ps1 -full
+   ```
+   
+  ### `Sync-AzsdkLabels.ps1`
```
weshaggard, line 75: Snapshot? Also looks like you copied some text from above.

tools/github/scripts/Snapshot-AzsdkLabels.ps1
```diff
@@ -0,0 +1,150 @@
...
+    [Parameter()]
+    [ValidateScript({Test-Path $_ -PathType 'Leaf'})]
+    [string] $LabelsFilePath = "$PSScriptRoot/../data/common-labels.csv",
+
+    [Parameter(ParameterSetName = 'RepositoryFile')]
+    [ValidateScript({Test-Path $_ -PathType 'Leaf'})]
+    [string] $RepositoryFilePath = "$PSScriptRoot/../data/repositories.txt",
+
+    [Parameter(ParameterSetName = 'Repositories')]
+    [ValidateNotNullOrEmpty()]
```
weshaggard, line 16: Why have 3 different way to specify these? Why not just use the repositories.txt?

tools/github/scripts/Snapshot-AzsdkLabels.ps1
```diff
@@ -0,0 +1,150 @@
...
+        'Azure/azure-sdk-for-java'
+        'Azure/azure-sdk-for-js'
+        'Azure/azure-sdk-for-net'
+        'Azure/azure-sdk-for-python'
+        'Azure/azure-sdk-tools'
+    ),
+
+    [Parameter(ParameterSetName = 'Languages')]
+    [ValidateNotNullOrEmpty()]
+    [string[]] $Languages = @('cpp', 'go', 'java', 'js', 'net', 'python'),
```
weshaggard, line 29: If you keep these do you not want to add android, ios, and c? We should try to keep those 9 languages in-sync.

tools/github/README.md
```diff
@@ -10,8 +10,19 @@ This area is focused on the management of the Azure SDK GitHub repositories, con
...
+
 * **root**  
-  _The root contains the core scripts for managing data and the set of centrally managed Azure SDK repositories._
+  _The root contains the scripts used for repository management and their associated data._
+
+* **scripts**  
+  _The core scripts for repository management, such as managing milestones and labels._
+
+* **data**  
+  _The data associated with the scripts.  This includes items such as the set of centrally managed Azure SDK repositories and the labels common to all repositories. set of labels for Azure SDK repositories and the set of repositories that are centrally managed._
```
heaths, line 22:
```suggestion
  _The data associated with the scripts.  This includes items such as the set of centrally managed Azure SDK repositories and the labels common to all repositories._
```

tools/github/scripts/Sync-AzsdkLabels.ps1
```diff
@@ -24,6 +24,10 @@ param (
     [ValidateNotNullOrEmpty()]
     [string[]] $Languages = @('cpp', 'go', 'java', 'js', 'net', 'python'),
 
+    [Parameter(Mandatory=$false)]
```
heaths, line 27:
```suggestion
    [Parameter()]
```

tools/github/scripts/Sync-AzsdkLabels.ps1
```diff
@@ -42,41 +46,75 @@ if ($PSCmdlet.ParameterSetName -eq 'RepositoryFile') {
...
+        foreach ($label in $labels) {
+            $result = gh label create $label.Name -d $label.Description -c $label.Color -R "$repo" --force 2>&1
+
+            if ($LASTEXITCODE) {
+                Write-Error "Failed to create or update the common labels to ${repo}: $result"
+            }
+
+            $completed = ($labelCount / $totalLabels) * 100
+            Write-Progress -Activity $activity -Status "$($repo): $($label.Name)" -PercentComplete $completed
+            $labelCount++
```
heaths, line 74: Wrong order. You already completed the label. Start with 0 and increment, then write progress. This would be +1 off when done. Probably not noticeable, but still wrong.

heaths (approved):

tools/github/README.md
jsquire: Whoops.   I did indeed.   Nice catch.

tools/github/scripts/Snapshot-AzsdkLabels.ps1
jsquire: These are just following the pattern that was previously established.  We do occasionally run one-offs for subsets, so having an override is helpful.   I'm not sure how @heaths originally settled on using both the full repository name and the language shorthand.

tools/github/scripts/Snapshot-AzsdkLabels.ps1
jsquire: No objection, though we should do that consistently across all the scripts.   I'll work that in.

weshaggard (approved):
Looks reasonable to me.

PR #6548, opened 2023-07-19
Resolve Bug in comment autoRefresh

chidozieononiwu:
Replacement for https://github.com/Azure/azure-sdk-tools/pull/6547

PR #6560, opened 2023-07-20
Fix user icon bug

yeojunh:
fixes #6548


chidozieononiwu (approved):
Did you mean to reference the issue in the PR description. Seems to be pointing to a PR instead.

PR #6561, opened 2023-07-21
Fix the support of generating Example Model of DataFactoryKeyVaultReference type of DataFactoryElement

RodgeFu:
Fix the support of generating Example Model of DataFactoryKeyVaultReference type of DataFactoryElement according to the change in Azure.Core.Expression.DataFactory

### DIFF
--- a/tools/sdk-testgen/packages/autorest.testmodeler/src/core/model.ts
+++ b/tools/sdk-testgen/packages/autorest.testmodeler/src/core/model.ts
@@ -234,7 +234,6 @@ export class ExampleValue {
     private static createSchemaForDfeObject(session: Session<TestCodeModel>, raw: any, dfeFormat: string): ObjectSchema | undefined {
         const dfeObjectType = 'type';
         const dfeObjectValue = 'value';
-        const dfeObjectTypeValues = ['Expression', 'SecureString', 'AzureKeyVaultSecretReference'];
         const dfeObjectSchemaPrefix = 'DataFactoryElement-';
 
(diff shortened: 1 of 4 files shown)


tools/sdk-testgen/packages/autorest.testmodeler/src/core/model.ts
```diff
@@ -231,17 +231,30 @@ export class ExampleValue {
         return instance;
     }
 
-    private static createSchemaForDfeObject(raw: any, dfeFormat: string): ObjectSchema | undefined {
+    private static createSchemaForDfeObject(session: Session<TestCodeModel>, raw: any, dfeFormat: string): ObjectSchema | undefined {
```
raych1, line 234: Will this breaking change break other scenarios?

raych1:
Hi @RodgeFu , which version of this package is used in dotnet sample generation? I heard from the version after 2.5.0(included) with webpack change never works.

tools/sdk-testgen/packages/autorest.testmodeler/src/core/model.ts
RodgeFu: no, this method is added in my previous check-in and not used in other scenarios.

RodgeFu:
> Hi @RodgeFu , which version of this package is used in dotnet sample generation? I heard from the version after 2.5.0(included) with webpack change fails to compile due to dependencies resolve error.

I think I updated it to use 2.6 a few weeks ago and didn't see any error in build and publish.

ArthurMa1978 (approved):

PR #6564, opened 2023-07-21
Base64 encode path

billwert:
Paths can have characters that are illegal for header values. Base64 encoding the path will prevent issues here.


scbedd (approved):
Thanks Bill. Once this merges and the CI build finishes feel free to bump your proxy PR in place (don't need to recreate). It'll do the right thing.

PR #6636, opened 2023-08-01
Sanitize requests when logging

JoshLove-msft:


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
```
scbedd, line 3: Where did this come from? I'm a bit weirded out it wasn't already there, and I don't understand why it needs to be if it wasn't before.

We don't need this if you swapped that property to `public` right?

JoshLove-msft: I made the Logger property internal again so we do need this.

scbedd (approved):
Thanks for the PR @JoshLove-msft 

Once you get this green, I'm good with these changes. Hoping our actual signing process for the .exe should be unaffected 👍

JoshLove-msft:
> Once you get this green, I'm good with these changes. Hoping our actual signing process for the .exe should be unaffected 👍

Should be unaffected as I moved it directly into the test directory. That is what was initially caught by my changes - the icon was missing when packing due to the targets file not having the icon path.

PR #6642, opened 2023-08-02
Add an analyzer to prevent use of some internal shared source types

annelo-msft:
For the [JSON Merge Patch work](https://github.com/Azure/azure-sdk-for-net/issues/27402), we want to use MutableJsonDocument and related types in Azure SDK Patch models.

The [proposal to do this](https://github.com/Azure/azure-sdk-for-net/pull/37341) involves moving MutableJsonDocument and related types into internal shared source temporarily to accelerate the generator work on models for PATCH operations.  However, we do not want libraries we don't know about taking dependencies on MutableJsonDocument, so that we have the freedom to change its API when we later make it a public type.

This analyzer will prevent projects in the azure-sdk-for-net repo from using MutableJsonDocument and related types.  Anyone who uses the type will have to knowingly suppress this warning and therefore opt-into the risk of breaking changes we might make in the future before it becomes a public type.  The analyzer is sufficiently general to be used with any other internal shared source types we want to limit use of outside Azure.Core.


src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
```diff
@@ -122,7 +122,14 @@ internal class Descriptors
             nameof(AZC0018),
             "Do ensure protocol method take a RequestContext parameter called context and not take models as parameter type or return type.",
             "Protocol method should have requestContext as the last parameter and don't have model as parameter type or return type.",
-            "Usage", DiagnosticSeverity.Warning, isEnabledByDefault: true, description: null);
+            "Usage", DiagnosticSeverity.Warning, isEnabledByDefault: true, description: null);
+
+        public static DiagnosticDescriptor AZC0020 = new DiagnosticDescriptor(
```
annelo-msft, line 127: This analyzer number skips AZC0019 to leave space for it to be added in this PR: https://github.com/Azure/azure-sdk-tools/pull/6496, as it is closely related to AZC0018.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
```diff
@@ -0,0 +1,91 @@
...
+
+                    if (namedTypeSymbol.IsGenericType)
+                    {
+                        foreach (var typeArgument in namedTypeSymbol.TypeArguments)
+                        {
+                            CheckType(context, typeArgument, symbol);
+                        }
+                    }
+                }
+            }
```
heaths, line 52: IMO, this could be a little confusing given `Analyze` is also passed a `context`, which you now shadow. And because it's the only parameter you could possibly want to use in an enclosed static method, I don't see a point in not just declaring this method on the class and passing in the context, which you have to do anyway. Would perhaps make it a little more readable / maintainable.

heaths (approved):
A few nits/tips, but otherwise LGTM.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
annelo-msft: Cool, happy to change that.  I copied this pattern from [BannedAssembliesAnalyzer](https://github.com/Azure/azure-sdk-tools/blob/main/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs), but don't feel strongly about it.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Azure.ClientSdk.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class BannedTypesAnalyzer : SymbolAnalyzerBase
+    {
+        private static HashSet<string> BannedTypes = new HashSet<string>()
+        {
+            "MutableJsonDocument",
```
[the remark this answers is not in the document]
annelo-msft: Yeah, I think this is a good idea.  What property on INamedTypeSymbol would you use instead of Name to get the fully-qualified name?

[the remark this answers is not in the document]
heaths: Looking at https://learn.microsoft.com/dotnet/api/microsoft.codeanalysis.inamedtypesymbol, perhaps `ContainingNamespace`? Not really my forte, though. I know that .NET itself has various types - like the nullable attributes - where it only cares about the namespace-qualified name to provide some meaningful features (like nullability checks). Not sure if it's worth looking at what they do. Maybe try `ContainingNamespace` first and see if that works.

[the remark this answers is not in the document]
annelo-msft: Yeah, I can just check both.  I took a quick look at the ToString() code in dotnet/roslyn, and it is quite compex.

heaths (approved):

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+        };
+
+        public override void Analyze(ISymbolAnalysisContext context)
+        {
+            if (IsAzureCore(context.Symbol.ContainingAssembly))
+            {
+                return;
+            }
+
+            switch (context.Symbol)
```
pshao25, line 41: If I understand correctly, you are making sure every method/model don't have `MutableJsonDocument`? Then should we test nested types?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+                    break;
+                case IEventSymbol eventSymbol:
+                    CheckType(context, eventSymbol.Type, eventSymbol);
+                    break;
+                case IPropertySymbol propertySymbol:
+                    CheckType(context, propertySymbol.Type, propertySymbol);
+                    break;
+                case IFieldSymbol fieldSymbol:
+                    CheckType(context, fieldSymbol.Type, fieldSymbol);
+                    break;
```
pshao25, line 57: Do we have tests for these cases? I'd like to see what kind of code is about to be banned.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+                return;
+            }
+
+            switch (context.Symbol)
+            {
+                case IParameterSymbol parameterSymbol:
+                    CheckType(context, parameterSymbol.Type, parameterSymbol);
+                    break;
+                case IMethodSymbol methodSymbol:
+                    CheckType(context, methodSymbol.ReturnType, methodSymbol);
```
pshao25, line 47: Just curious, why only return type is checked? How about the input parameters? Will it go to the first case?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0020Tests.cs
```diff
@@ -0,0 +1,104 @@
...
+using Azure.Core.Json;
+
+namespace LibraryNamespace
+{
+    public class Model
+    {
+        MutableJsonElement {|AZC0020:_element|};
+    }
+}";
+            await Verifier.VerifyAnalyzerAsync(code, _sharedSourceFiles);
```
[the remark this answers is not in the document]
pshao25: If you are saying we cannot make `_element` public, then I think the private one should be a positive case.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
AlexanderSher: Yes

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
annelo-msft: I will add them per this comment: https://github.com/Azure/azure-sdk-tools/pull/6642#discussion_r1290303416

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
annelo-msft: I checked with @AlexanderSher, and his "Yes" is in response to the last question:
"Will it go to the first case?" -> Yes.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0020Tests.cs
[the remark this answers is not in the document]
annelo-msft: > How could we fail this?

Yes, good question!  The idea is that we will suppress AZC0020 for generated Patch models only, until we move the MutableJsonDocument types into the Azure.Core public API.  It is to prevent people from using them for purposes we don't know about, so they don't take dependencies on functionality that we plan to change.  There are more details about this in this comment: https://github.com/Azure/azure-sdk-tools/pull/6642#issue-1832206874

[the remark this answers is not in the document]
annelo-msft: > If you are saying we cannot make _element public, then I think the private one should be a positive case.

This analyzer should prevent its use for any case outside of the Azure.Core library.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        private static bool IsBannedType(INamedTypeSymbol namedTypeSymbol)
+        {
+            return BannedTypes.Contains($"{namedTypeSymbol.ContainingNamespace}.{namedTypeSymbol.Name}");
+        }
+
+        private static bool IsAzureCore(IAssemblySymbol assembly)
+        {
+            return assembly.Name.Equals("Azure.Core");
```
jsquire, line 103: Should we allow anything in the Azure.Core family - such as Azure.Core.Experimental?

jsquire (approved):

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedTypesAnalyzer.cs
annelo-msft: Good point!  Added it.

PR #6643, opened 2023-08-02
Change logging category to include TestProxy

JoshLove-msft:
This will allow users to set an env var Logging__LogLevel__Azure.Sdk.Tools.TestProxy = "Debug" instead of changing the Default or DebugLogging category. _Note changing Default will still work in the same way and I don't think we advertised "DebugLogging" to anyone so it hopefully won't break anyone._ However, changing the Default log level would break tests in .NET since it would cause tests to start emitting logs. 

By introducing a specific logging category (or more accurately renaming the specific logging category), we can let it still be configurable for users without requiring users to change the Default level.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/DebugLogger.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/DebugLogger.cs
@@ -34,7 +34,7 @@ namespace Azure.Sdk.Tools.TestProxy.Common
         {
             if (Logger == null && factory != null)
             {
-                Logger = factory.CreateLogger("TestProxy");
+                Logger = factory.CreateLogger("Azure.Sdk.Tools.TestProxy");
             }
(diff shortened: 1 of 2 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/DebugLogger.cs
```diff
@@ -40,7 +34,7 @@ public static void ConfigureLogger(ILoggerFactory factory)
         {
             if (Logger == null && factory != null)
             {
-                Logger = factory.CreateLogger("DebugLogging");
+                Logger = factory.CreateLogger("Azure.Sdk.Tools.TestProxy");
```
JoshLove-msft, line 37: The generic `CreateLogger<T>` method doesn't work with static classes.

scbedd (approved):
Makes sense. Thanks Josh!

PR #6675, opened 2023-08-03
Only have a single DefinitionId for each type

LarryOsterman:
Update ApiView tool to have a single DefinitionId for each type.


LarryOsterman:
Ping. I'd like to get the ApiView challenges sorted :).

antkmsft (approved):

PR #6678, opened 2023-08-04
Add Azure.Core.TestFramework tests to nightly integration

scbedd:
Azure.Core test framework has some excellent filename encoding tests that gotcha-ed a proxy roll-out a couple weeks ago. Adding those to our nightly integration tests so we can catch it earlier next time!

### DIFF
--- a/tools/test-proxy/tests.yml
+++ b/tools/test-proxy/tests.yml
@@ -4,186 +4,186 @@ variables:
   - template: /eng/pipelines/templates/variables/globals.yml
 
 stages:
-  # - stage: IntegrationTests
-  #   displayName: "Asset Sync Integration Tests"
-  #   jobs:
-  #   - job: Solution_Integration_Test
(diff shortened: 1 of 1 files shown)


tools/test-proxy/tests.yml
```diff
@@ -169,6 +169,50 @@ stages:
...
+
+      variables:
+        REPO: "Azure/azure-sdk-for-net"
+        CLONE_LOCATION: "$(Agent.BuildDirectory)/net_repo"
+
+      steps:
+        - template: /eng/pipelines/templates/steps/install-dotnet.yml
+
+        - pwsh: |
+            git clone https://github.com/$(REPO) $(CLONE_LOCATION) --depth 1
```
benbp, line 199: Probably easier to use sparse checkout here, but not a big deal.

benbp (approved):

PR #6695, opened 2023-08-08
[stress] test resource parallel deployment 

### LINKED ISSUE #6591: [stress] Make 'paralleism: N > 1' tests work reliably
Event Hubs testing is pushing us to want to handle larger `N` numbers for parallel instances, which makes it impractical to handle in our simplified manner (more threads, more processes in a single container). 

Our `init` containers for the pods are responsible for deploying test resources but once you increase the # of instances running in parallel you increase the possibility that the separate deployer instances will collide with each other, causing test failures.

The simplest solution is to leverage Kubernetes and do a little polling:

1. In the K8S job use the `completionMode: Indexed`, similar to what the .NET Service Bus tests do today. [(.NET's deploy-job.yaml)](https://github.com/Azure/azure-sdk-for-net/blob/d6c933b3addcc22d111f3a29d56333c05ab32bbe/sdk/servicebus/Azure.Messaging.ServiceBus/stress/templates/deploy-job.yaml#L6)
2. The 0th instance of the `init` container will do the actual deployment, just like it does today. The deployment will have a **predictable unique name**.
3. The 1st -> N `init` containers will just poll and check that **predictable unique name** deployment has completed and block until that completes.

This should make it so the individual stress apps don't need to change any code, and everything feels the same at N=1 as it does with N > 1.

ckairen:
closes #6591


tools/stress-cluster/cluster/kubernetes/stress-test-addons/images/test-resource-deployer/deploy-stress-test-resources.ps1
```diff
@@ -15,6 +15,10 @@ mkdir /azure
 Copy-Item "/scripts/stress-test/test-resources-post.ps1" -Destination "/azure/"
 Copy-Item "/mnt/testresources/*" -Destination "/azure/"
 
+while (kubectl get pods -n $env:namespace $env:BASE_NAME -o jsonpath='{.items[?(@.metadata.annotations.batch\.kubernetes\.io/job-completion-index=="0")].metadata.name}' -nq "Complete") {
```
benbp, line 18: Does this work? Usually the default pod service identity doesn't have GET access to the k8s api (unless I'm thinking of cross-namespace requests). Also how does the pod with index 0 complete if it reads its own status? Do we need an extra skip condition for index 0?

benbp: That said, I like this approach much more than polling for ARM deployment completion, pretty clever.

benbp: Also `-ne` not `-nq`

tools/stress-cluster/cluster/kubernetes/stress-test-addons/images/test-resource-deployer/deploy-stress-test-resources.ps1
```diff
@@ -15,6 +15,23 @@ mkdir /azure
 Copy-Item "/scripts/stress-test/test-resources-post.ps1" -Destination "/azure/"
 Copy-Item "/mnt/testresources/*" -Destination "/azure/"
 
+Write-Host "Job completion index $($env:JOB_COMPLETION_INDEX)"
+
+if ($env:JOB_COMPLETION_INDEX -and ($env:JOB_COMPLETION_INDEX -ne "0")) {
```
benbp, line 20: Probably good to add a comment above explaining why we need to do this.

benbp (approved):

PR #6739, opened 2023-08-09
Used ScheduledPurgeDate for both KV, MHSM

heaths:
The previous property was calculated so didn't exist from the REST-based
call. Also updated the api-version to the latest stable control plane
version.

### DIFF
--- a/eng/common/scripts/Helpers/Resource-Helpers.ps1
+++ b/eng/common/scripts/Helpers/Resource-Helpers.ps1
@@ -49,7 +49,7 @@ function Get-PurgeableResources {
   Write-Verbose "Retrieving deleted Managed HSMs from subscription $subscriptionId"
 
   # Get deleted Managed HSMs for the current subscription.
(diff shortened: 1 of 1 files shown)


mikeharder:
@heaths: Before merging this PR, I'd like to try running the cleanup pipeline using this PR.

mikeharder (changes requested):
Before merging this PR, I'd like to try running the cleanup pipeline using this PR.

mikeharder (approved):

mikeharder:
Fix prevents script from failing, but from the logs it appears we still need to do more to actually delete/purge the resources.

heaths:
> Fix prevents script from failing, but from the logs it appears we still need to do more to actually delete/purge the resources.

The script runs on a schedule as well, so it will keep trying and when the purge date expires, they will finally get cleaned up on the next run. That's how it was designed and has been working. What's concerning is how many vaults and HSMs have purge protection enabled. It should almost never be necessary. I started a thread internally to address going forward.

heaths:
I thought it was supposed to merge automatically once the final pipeline was approved.

PR #6771, opened 2023-08-15
Change AZC0002 to make method ending with required RequestContext pass

pshao25:
Currently AZC0002 fails when the last parameter is a required RequestContext, which is actually valid. This PR changes this with the below flow.

image


src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -21,111 +29,147 @@ public class ClientMethodsAnalyzer : ClientAnalyzerBase
...
+        {
+            IMethodSymbol syncMember = null;
+            var syncMemberName = method.Name.Substring(0, method.Name.Length - AsyncSuffix.Length);
+            syncMember = FindMethod(type.GetMembers(syncMemberName).OfType<IMethodSymbol>(), method.TypeParameters, method.Parameters);
+            if (syncMember == null)
+            {
+                var lastArgument = method.Parameters.LastOrDefault();
+                if (lastArgument != null && IsCancellationToken(lastArgument))
+                {
+                    syncMember = FindMethod(type.GetMembers(syncMemberName).OfType<IMethodSymbol>(), method.TypeParameters, method.Parameters.RemoveAt(method.Parameters.Length - 1), p => IsCancellationToken(p));
```
pshao25, line 150: This is for 
```C#
Download(CancellationToken cancellationToken = default);
DownloadAsync(CancellationToken cancellationToken);
DownloadAsync();
```

Why we don't have this logic before? Because when we look for sync peer before, we are not requiring the same requiredness before.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -21,104 +28,126 @@ public class ClientMethodsAnalyzer : ClientAnalyzerBase
             Descriptors.AZC0015
         });
 
-        private static void CheckClientMethod(ISymbolAnalysisContext context, IMethodSymbol member)
+        private static void CheckCancellationTokenOrRequestContext(ISymbolAnalysisContext context, IMethodSymbol member)
```
christothes, line 31:
```suggestion
        private static void CheckIsLastArgumentCancellationTokenOrRequestContext(ISymbolAnalysisContext context, IMethodSymbol member)
```

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -21,104 +28,126 @@ public class ClientMethodsAnalyzer : ClientAnalyzerBase
...
-            if (!member.IsVirtual && !member.IsOverride)
+            static bool IsCancellationOrRequestContext(IParameterSymbol parameterSymbol)
             {
-                context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0003, member.Locations.First()), member);
+                return IsCancellationToken(parameterSymbol) || IsRequestContext(parameterSymbol);
             }
 
             var lastArgument = member.Parameters.LastOrDefault();
-            var supportsCancellations = lastArgument != null && SupportsCancellationsParameter(lastArgument);
+            var isCancellationOrRequestContext = lastArgument != null && IsCancellationOrRequestContext(lastArgument);
```
christothes, line 49:
```suggestion
            var isLastArgumentCancellationOrRequestContext = lastArgument != null && IsCancellationOrRequestContext(lastArgument);
```

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -21,104 +28,126 @@ public class ClientMethodsAnalyzer : ClientAnalyzerBase
...
+            }
 
-            if (!member.IsVirtual && !member.IsOverride)
+            static bool IsCancellationOrRequestContext(IParameterSymbol parameterSymbol)
             {
-                context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0003, member.Locations.First()), member);
+                return IsCancellationToken(parameterSymbol) || IsRequestContext(parameterSymbol);
             }
 
             var lastArgument = member.Parameters.LastOrDefault();
```
christothes, line 48: Should we check early that the member has at least 1 Parameter and bail out early if so? This would avoid a few null checks below for `lastArgument` as well as any other analysis of the parameters.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -21,104 +28,126 @@ public class ClientMethodsAnalyzer : ClientAnalyzerBase
...
                 {
-                    // Skip methods that have overloads with cancellation tokens
-                    return;
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0002, member.Locations.FirstOrDefault()), member);
                 }
-
-                context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0002, member.Locations.FirstOrDefault()), member);
             }
-            else if (!lastArgument.IsOptional)
+            else if (IsCancellationToken(lastArgument))
```
christothes, line 64: Can we collapse the nested `if` statements here, since it appears there is not alternate branch for either?
```suggestion
            else if (IsCancellationToken(lastArgument) && !lastArgument.IsOptional)
```

m-nash (approved):

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
```diff
@@ -21,104 +28,126 @@ public class ClientMethodsAnalyzer : ClientAnalyzerBase
             Descriptors.AZC0015
         });
 
-        private static void CheckClientMethod(ISymbolAnalysisContext context, IMethodSymbol member)
+        private static void CheckCancellationTokenOrRequestContext(ISymbolAnalysisContext context, IMethodSymbol member)
         {
-            static bool SupportsCancellationsParameter(IParameterSymbol parameterSymbol)
+            static bool IsRequestContext(IParameterSymbol parameterSymbol)
```
annelo-msft, line 33: Do these need to be local functions?  I think it would make the code more readable (and reusable) if they were declared on the type.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0002Tests.cs
```diff
@@ -1,275 +1,303 @@
...
+        }
+
+        public virtual Response Get2(RequestContext context)
+        {
+            return null;
+        }
+    }
+}";
+            await Verifier.CreateAnalyzer(code)
+                .WithDisabledDiagnostics("AZC0018")
```
annelo-msft, line 262: Why does 0018 need to be disabled in this test?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs
pshao25: There are no checks below if `lastArgument` is null. What additional checks you are mentioning?

christothes: Doesn't the line right below check?

```c#
var isCancellationOrRequestContext = lastArgument != null && IsCancellationOrRequestContext(lastArgument);
```

If `members.Parameters` has no members, the rest of these checks are not necessary.

pshao25: I'm afraid we cannot. If `members.Parameters` has no members, we still want to find the overload below:
```C#
var overloadSupportsCancellations = FindMethod(
    member.ContainingType.GetMembers(member.Name).OfType<IMethodSymbol>(),
    member.TypeParameters,
    member.Parameters,
    p => IsCancellationToken(p));
```

christothes (approved):

PR #6772, opened 2023-08-15
flask support for python reviewer

xiangyan99:


packages/python-packages/apiview-gpt/app.py
```diff
@@ -0,0 +1,15 @@
+from flask import Flask, request, jsonify
+from src import review_python
+
+app = Flask(__name__)
+
+@app.route('/')
+def index():
+    return 'Hello, World!'
```
tjprescott, line 8: Is this really necessary?

packages/python-packages/apiview-gpt/src/_python_api.py
```diff
@@ -0,0 +1,6 @@
+from ._gpt_reviewer import GptReviewer
+
+def review_python(code):
```
tjprescott, line 3: This should really accept code and language, as apiview-gpt should be able to work with any language. APIView (web) would be responsible for passing the appropriate language.

tjprescott (approved):
Some comments, but we can always tweak this code later.

packages/python-packages/apiview-gpt/app.py
xiangyan99: I use it to detect if service is online.

I will remove it.

packages/python-packages/apiview-gpt/src/_python_api.py
xiangyan99: In the plan, this file will contain all Python related services. e.g. BreakingChangeDetector. And .net, js, java will have their own _language_api.py.

tjprescott (approved):

PR #6782, opened 2023-08-16
Create Epic work item type as Service or product

praveenkuttappan:
Epic type is not set when prepare release script creates a work item for a new service or product. This change is to set epic type as service or product so it aligns with new dashboard changes for service and product work items.

### DIFF
--- a/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
+++ b/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
@@ -135,7 +135,7 @@ function BuildHashKey()
 }
 
 $parentWorkItems = @{}
-function FindParentWorkItem($serviceName, $packageDisplayName, $outputCommand = $false)
+function FindParentWorkItem($serviceName, $packageDisplayName, $outputCommand = $false, $ignoreReleasePlannerTests = $true)
 {
   $key = BuildHashKey $serviceName $packageDisplayName
   if ($key -and $parentWorkItems.ContainsKey($key)) {
@@ -154,10 +154,12 @@ function FindParentWorkItem($serviceName, $packageDisplayName, $outputCommand = 
   else {
     $serviceCondition = "[ServiceName] <> ''"
   }
-
+  if($ignoreReleasePlannerTests){
+    $serviceCondition += " AND [Tags] NOT CONTAINS 'Release Planner App Test'"
+  }
   $query = "SELECT [ID], [ServiceName], [PackageDisplayName], [Parent] FROM WorkItems WHERE [Work Item Type] = 'Epic' AND ${serviceCondition}"
 
-  $fields = @("System.Id", "Custom.ServiceName", "Custom.PackageDisplayName", "System.Parent")
+  $fields = @("System.Id", "Custom.ServiceName", "Custom.PackageDisplayName", "System.Parent", "System.Tags")
 
   $workItems = Invoke-Query $fields $query $outputCommand
 
@@ -181,7 +183,7 @@ function FindParentWorkItem($serviceName, $packageDisplayName, $outputCommand = 
 }
 
 $releasePlanWorkItems = @{}
-function FindReleasePlanWorkItem($serviceName, $packageDisplayName, $outputCommand = $false)
+function FindReleasePlanWorkItem($serviceName, $packageDisplayName, $outputCommand = $false, $ignoreReleasePlannerTests = $true)
 {
   $key = BuildHashKey $serviceName $packageDisplayName
   if ($key -and $releasePlanWorkItems.ContainsKey($key)) {
@@ -200,10 +202,14 @@ function FindReleasePlanWorkItem($serviceName, $packageDisplayName, $outputComma
   else {
     $condition = "[ServiceName] <> ''"
   }
-  $condition += " AND [System.State] <> 'Finished'"  
+  $condition += " AND [System.State] <> 'Finished'"
+  if($ignoreReleasePlannerTests){
+    $condition += " AND [Tags] NOT CONTAINS 'Release Planner App Test'"
+  }
+
   $query = "SELECT [ID], [ServiceName], [PackageDisplayName], [Parent] FROM WorkItems WHERE [Work Item Type] = 'Release Plan' AND ${condition}"
 
-  $fields = @("System.Id", "Custom.ServiceName", "Custom.PackageDisplayName", "System.Parent")
+  $fields = @("System.Id", "Custom.ServiceName", "Custom.PackageDisplayName", "System.Parent", "System.Tags")
 
   $workItems = Invoke-Query $fields $query $outputCommand
 
@@ -252,7 +258,7 @@ function FindLatestPackageWorkItem($lang, $packageName, $outputCommand = $true)
   return $latestWI
 }
 
-function FindPackageWorkItem($lang, $packageName, $version, $outputCommand = $true, $includeClosed = $false)
+function FindPackageWorkItem($lang, $packageName, $version, $outputCommand = $true, $includeClosed = $false, $ignoreReleasePlannerTests = $true)
 {
   $key = BuildHashKeyNoNull $lang $packageName $version
   if ($key -and $packageWorkItems.ContainsKey($key)) {
@@ -264,6 +270,7 @@ function FindPackageWorkItem($lang, $packageName, $version, $outputCommand = $tr
   $fields += "System.State"
   $fields += "System.AssignedTo"
   $fields += "System.Parent"
+  $fields += "System.Tags"
   $fields += "Custom.Language"
   $fields += "Custom.Package"
   $fields += "Custom.PackageDisplayName"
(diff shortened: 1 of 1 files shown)


maririos (approved):

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -180,6 +182,56 @@ function FindParentWorkItem($serviceName, $packageDisplayName, $outputCommand =
...
+    else {
+      $condition += " AND [PackageDisplayName] = ''"
+    }
+  }
+  else {
+    $condition = "[ServiceName] <> ''"
+  }
+  $condition += " AND [System.State] <> 'Finished'"
+  if($ignoreReleasePlannerTests){
+    $condition += " AND [Tags] NOT CONTAINS 'Release Planner App Test'"
```
maririos, line 207: is there a differente between `NOT CONTAINS` and `<>` ?

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -490,21 +545,45 @@ function CreateOrUpdatePackageWorkItem($lang, $pkg, $verMajorMinor, $existingIte
...
-  $parentItem = FindOrCreatePackageGroupParent $serviceName $pkgDisplayName -outputCommand $false
+  $parentItem = FindOrCreateReleasePlanParent $serviceName $pkgDisplayName -outputCommand $false
   $workItem = CreateWorkItem $title "Package" "Release" "Release" $fields $assignedTo $parentItem.id -outputCommand $outputCommand
   Write-Host "[$($workItem.id)]$lang - $pkgName($verMajorMinor) - Created"
   return $workItem
 }
 
+function FindOrCreateReleasePlanParent($serviceName, $packageDisplayName, $outputCommand = $true)
+{
+  $existingItem = FindReleasePlanWorkItem $serviceName $packageDisplayName -outputCommand $outputCommand
```
maririos, line 561: should we pass `ignoreReleasePlannerTests ` here too?

maririos: same in the other calls in this function

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -490,21 +545,45 @@ function CreateOrUpdatePackageWorkItem($lang, $pkg, $verMajorMinor, $existingIte
...
+  $title = "Release Plan - $($packageDisplayName)"
+  $workItem = CreateWorkItem $title "Release Plan" "Release" "Release" $fields $null $productParentItem.id
+
+  $localKey = BuildHashKey $serviceName $packageDisplayName
+  Write-Host "[$($workItem.id)]$localKey - Created release plan work item"
+  $releasePlanWorkItems[$localKey] = $workItem
+  return $workItem
+}
+
 function FindOrCreatePackageGroupParent($serviceName, $packageDisplayName, $outputCommand = $true)
```
maririos, line 582: this function created the product wi right? If so we could updae line 600 so it says "Created product work item" .
Currently it says "Parent" which is confusing

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -490,21 +545,45 @@ function CreateOrUpdatePackageWorkItem($lang, $pkg, $verMajorMinor, $existingIte
...
   $workItem = CreateWorkItem $title "Package" "Release" "Release" $fields $assignedTo $parentItem.id -outputCommand $outputCommand
   Write-Host "[$($workItem.id)]$lang - $pkgName($verMajorMinor) - Created"
   return $workItem
 }
 
+function FindOrCreateReleasePlanParent($serviceName, $packageDisplayName, $outputCommand = $true)
+{
+  $existingItem = FindReleasePlanWorkItem $serviceName $packageDisplayName -outputCommand $outputCommand
+  if ($existingItem) {
+    Write-Host "Found existing release plan work item [$($existingItem.id)]"
```
JonathanCrd, line 563: Coul a test on Dev triggers this function?
Wondering what would happen if this gets trigger and then it doesn't find the existing work item because it has the test tag.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
praveenkuttappan: One is string not equal and the other one is to check if it is in the list

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
praveenkuttappan: Can you explain what do you mean by running on Dev?

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
praveenkuttappan: I will change it.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
praveenkuttappan: If an existing release plan is not found, then it will create a new one. But this gets triggered only when someone runs prepare release script and it should always exclude test work items.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -522,21 +602,23 @@ function FindOrCreatePackageGroupParent($serviceName, $packageDisplayName, $outp
...
+  $serviceParent = FindParentWorkItem $serviceName -packageDisplayName $null -outputCommand $outputCommand -ignoreReleasePlannerTests $ignoreReleasePlannerTests
   if ($serviceParent) {
+    Write-Host "Found existing service work item [$($serviceParent.id)]"
     return $serviceParent
   }
 
   $fields = @()
   $fields += "`"PackageDisplayName=`""
   $fields += "`"ServiceName=${serviceName}`""
+  $fields += "`"Custom.EpicType=Service`""
```
weshaggard, line 616: Do we need to care about Sub-Service type or Component type?

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -218,6 +270,7 @@ function FindPackageWorkItem($lang, $packageName, $version, $outputCommand = $tr
   $fields += "System.State"
   $fields += "System.AssignedTo"
   $fields += "System.Parent"
+  $fields += "System.Tags"
```
weshaggard, line 273: What are we getting from the Tags? Is that only to filter out the tests? If so we might want to consider other potential ways to filter these out, like perhaps having some way to close them all so they don't show up in the queries.

praveenkuttappan: This is to avoid showing up work items created from release planner app in dev and PPE.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
praveenkuttappan: We create another Epic Item as child of Service Epic for the product. I am assuming that's what used currently for subservice name as well.

maririos: We are managing subservice as another EPIC. Same as previously done before

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
weshaggard: Are tags the best way of doing this? These will also start to show up in other query or dashboards that folks might be using directly in DevOps. Can we instead have something that goes in and closes all the dev and PPE created items?

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
weshaggard: It is fine if it is another Epic but does this automation need to handle those as well?

praveenkuttappan: Prepare release scrip is already handling it by creating a child under Service WI and then creating individual language package items under this subservice/product WI.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
praveenkuttappan: Yes, we can mark them as closed by automation pipeline. I will discuss with @maririos and come up with a plan. In the meanwhile, I think we should include current changes so work items created by prepare release script won't attach package WI to incorrect service and product.

weshaggard (approved):

PR #6786, opened 2023-08-16
Copilot comments search

### LINKED ISSUE #6460: APIView comments database
Issue to track the work that might come to APIView as we look into adding ChatGPT to it

```[tasklist]
### Tasks
- [ ] https://github.com/Azure/azure-sdk-tools/issues/6647
- [ ] https://github.com/Azure/azure-sdk-tools/issues/6648
- [ ] https://github.com/Azure/azure-sdk-tools/issues/6649
- [ ] https://github.com/Azure/azure-sdk-tools/issues/6650
```

yeojunh:
Fixes #6460 (and #6650 #6776)

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Controllers/CopilotCommentsController.cs
+++ b/src/dotnet/APIView/APIViewWeb/Controllers/CopilotCommentsController.cs
@@ -9,7 +9,7 @@ using APIViewWeb.Filters;
 
 namespace APIViewWeb.Controllers
 {
-    //[TypeFilter(typeof(ApiKeyAuthorizeAsyncFilter))]
+    [TypeFilter(typeof(ApiKeyAuthorizeAsyncFilter))]
 
     public class CopilotCommentsController : Controller
     {
--- a/src/dotnet/APIView/APIViewWeb/Controllers/CopilotCommentsController.cs
+++ b/src/dotnet/APIView/APIViewWeb/Controllers/CopilotCommentsController.cs
@@ -1,0 +1,103 @@
+using System.Threading.Tasks;
+using APIViewWeb.Managers;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+
+namespace APIViewWeb.Controllers
+{
+    //[TypeFilter(typeof(ApiKeyAuthorizeAsyncFilter))]
+
+    public class CopilotCommentsController : Controller
+    {
+        private readonly ICopilotCommentsManager _copilotManager;
+        private readonly ILogger _logger;
+
+        public CopilotCommentsController(ICopilotCommentsManager copilotManager, ILogger<CopilotCommentsController> logger)
+        {
+            _copilotManager = copilotManager;
+            _logger = logger;
+        }
+
+        [HttpPost]
+        public async Task<ActionResult> CreateDocument(string language, string badCode, string goodCode = null, string comment = null, string[] guidelineIds = null)
+        {
+            if (badCode == null || language == null)
+            {
+                _logger.LogInformation("Request does not have the required badCode or language fields for CREATE.");
+                return StatusCode(statusCode: StatusCodes.Status400BadRequest);
+            }
+
+            var id = await _copilotManager.CreateDocumentAsync(language, badCode, goodCode, comment, guidelineIds, User.GetGitHubLogin());
+            _logger.LogInformation("Added a new document to database.");
+            return StatusCode(statusCode: StatusCodes.Status201Created, id);
+        }
+
+        [HttpPut]
+        public async Task<ActionResult> UpdateDocument(string id, string language, string badCode = null, string goodCode = null, string comment = null, string[] guidelineIds = null)
+        {
+            if (id == null || language == null)
+            {
+                _logger.LogInformation("Request does not have the required ID or language fields for UPDATE.");
+                return StatusCode(statusCode: StatusCodes.Status400BadRequest);
+            }
+
+            var result = await _copilotManager.UpdateDocumentAsync(id, language, badCode, goodCode, comment, guidelineIds, User.GetGitHubLogin());
+            if (result != null)
+            {
+                _logger.LogInformation("Found existing document with ID. Updating document.");
+                return Ok(result);
+            } else
+            {
+                _logger.LogInformation("Could not find a match for the given ID.");
+                return StatusCode(statusCode: StatusCodes.Status404NotFound);
+            }
+        }
+
+        [HttpGet]
+        public async Task<ActionResult> GetDocument(string id, string language)
+        {
+            if (id == null || language == null)
+            {
+                _logger.LogInformation("Request does not have the required ID or language fields for GET.");
+                return StatusCode(statusCode: StatusCodes.Status400BadRequest);
+            }
+
+            var document = await _copilotManager.GetDocumentAsync(id, language);
+            if (document != null)
+            {
+                return Ok(document);
+            } else
+            {
+                _logger.LogInformation("No document with this id exists in the database.");
+                return StatusCode(statusCode: StatusCodes.Status404NotFound);
+            }
+        }
+
+        [HttpDelete]
+        public async Task<ActionResult> DeleteDocument(string id, string language)
+        {
+            if (id == null || language == null)
+            {
+                _logger.LogInformation("Request does not have the required ID or language fields for DELETE.");
+                return StatusCode(statusCode: StatusCodes.Status400BadRequest);
+            }
+
+            await _copilotManager.DeleteDocumentAsync(id, language, User.GetGitHubLogin());
+            return StatusCode(statusCode: StatusCodes.Status204NoContent);
+        }
+
+        [HttpGet]
+        public async Task<ActionResult> SearchDocument(string language, string code, float threshold, int limit)
+        {
+            if (code == null || language == null)
+            {
+                _logger.LogInformation("Request does not have the required code or language fields for SEARCH.");
+                return StatusCode(statusCode: StatusCodes.Status400BadRequest);
+            }
+
+            var document = await _copilotManager.SearchDocumentsAsync(language, code, threshold, limit);
+            return Ok(document);
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
```diff
@@ -31,163 +31,136 @@ IConfiguration configuration
...
-            var embedding = await GetEmbeddingsAsync(badCode);
-
             var document = new CopilotCommentModel()
             {
+                Language = language,
                 BadCode = badCode,
                 GoodCode = goodCode,
-                Embedding = embedding,
-                Language = language,
+                Embedding = await GetEmbeddingsAsync(badCode),
```
kristapratico, line 41: I don't know what .NET returns here - is this just the embedding data or does it return an object with metadata and the embedding?

src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
```diff
@@ -31,163 +31,136 @@ IConfiguration configuration
...
 
-            var documentJson = JsonConvert.SerializeObject(document);
+            var searchResults = new List<CopilotSearchModel>();
+            foreach (var document in documents)
+            {
+                var similarity = CosineSimilarity(document.Embedding, embeddings);
+                if (similarity < threshold)
+                    continue;
+                searchResults.Add(new CopilotSearchModel(similarity, document));
+            }
```
kristapratico, line 113: Mostly just curious, does the cosmos db vector search require you to iterate over each document to check for similarity? Is there no convenience function? With other indexes, I've seen that you just give it the index name and call search with your embedding and it presumably does it for you, returning the top X results

kristapratico: edit: scrolled a little and now I see we write our own similarity check. Does cosmos not support a built-in way to do this?

src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
```diff
@@ -31,163 +31,136 @@ IConfiguration configuration
...
             {
+                Language = language,
                 BadCode = badCode,
                 GoodCode = goodCode,
-                Embedding = embedding,
-                Language = language,
+                Embedding = await GetEmbeddingsAsync(badCode),
                 Comment = comment,
-                GuidelineIds = guidelineIds,
+                GuidelineIds = SplitGuidelineIds(guidelineIds),
```
tjprescott, line 43: Does the database not do this automatically?

tjprescott (commented):
Will defer to @chidozieononiwu and @mpodwysocki for C#.

src/dotnet/APIView/APIViewWeb/Controllers/CopilotCommentsController.cs
```diff
@@ -19,73 +23,126 @@ public CopilotCommentsController(ICopilotCommentsManager copilotManager, ILogger
...
+            }
+            catch (Exception err)
+            {
+                _logger.LogInformation("Error: unsuccessful DELETE request. " + err.Message);
+                return StatusCode(statusCode: StatusCodes.Status500InternalServerError);
+            }
+        }
+
+        [HttpGet]
+        public async Task<ActionResult> SearchDocument(string language, string code, float threshold, int limit = -1)
```
tjprescott, line 128: What is the purpose of this endpoint? The GET is supposed to be by ID, not language and "badCode".

src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
tjprescott, line 113: Agree, can't we just search the database directly? It looks like we are just getting everything for a given language and then manually comparing every document, which seems like it would negate the benefit of having a DB search index.

src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
yeojunh: `GetEmbeddingsAsync` is a function I defined to extract just the vector values from an `Embeddings` object which contains metadata and the vector. We don't need the metadata since we're just computing the cosine similarity.

src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
yeojunh: Unfortunately yes, as far as I know. The reason we wanted to use Azure Cosmos DB for MongoDB as our database is because they supported a built-in vector search, but they don't support the scoring method (MongoDB Atlas does but we want to stay within Azure), which we need for searching values that are more similar than `threshold`. Which other index were you referencing? I'd love to take a look!

src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
yeojunh: When I was testing with Postman, it sent the guideline ids as a string like `"id1, id2, id3"`. `SplitGuidelineIds` just parses it into an array, but if you know how to send an array with a REST API call, I will happily change it.

src/dotnet/APIView/APIViewWeb/Controllers/CopilotCommentsController.cs
```diff
@@ -19,73 +23,126 @@ public CopilotCommentsController(ICopilotCommentsManager copilotManager, ILogger
...
+            if (language == null || code == null)
+            {
+                _logger.LogInformation("Request does not have the required code or language fields for SEARCH.");
+                return StatusCode(statusCode: StatusCodes.Status400BadRequest);
+            }
+
+            try
+            {
+                var document = await _copilotManager.SearchDocumentsAsync(language, code, threshold, limit);
+                return Ok(document);
```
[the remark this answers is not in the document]
yeojunh: Yes! Thanks for catching that :)

src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
```diff
@@ -31,163 +31,136 @@ IConfiguration configuration
...
+            }
 
-            return documentJson;
+            if (limit < 0)
+            {
+                limit = defaultLimit;
+            }
+
+            var topResults = searchResults.OrderByDescending(item => item.similarity).Take(limit);
+            return JsonSerializer.Serialize(topResults);
```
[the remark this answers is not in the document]
yeojunh: `topResults` is a list of `CopilotSearchModel`. It returns the JSON string of the list to return for the `Controller`. No worries, I'm also still learning 😃

src/dotnet/APIView/APIViewWeb/Controllers/CopilotCommentsController.cs
yeojunh: This is the API that calls the vector search with `badCode`. There's another endpoint `GetDocument(id, language)` that gets an item with the id. Do you have a suggestion for another endpoint name?

src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
kristapratico: Have we tried Azure Cognitive Search? It seems like we can connect our data from Cosmos DB to the search index and then use built-in search queries. https://learn.microsoft.com/en-us/azure/search/search-indexer-overview

yeojunh: From what I understand so far, the way to add a search index in Cosmos is using Azure Cognitive Search. I can optimize it a bit using language as a partition, which we can do natively in Cosmos DB. Though it might be that I haven't fully grasped the concept of a DB search index and all the functionalities in Cosmos. Most of this code is [referenced from a previous cosmos database search](https://github.com/Azure/azure-sdk-tools/blob/086f9e22a83d9ed1f832ea9aee18605330ea030d/src/dotnet/APIView/APIViewWeb/Repositories/CosmosCommentsRepository.cs#L25) where they search for the `reviewId` and `lineId`. I can continue to look into it.

src/dotnet/APIView/APIViewWeb/Controllers/CopilotCommentsController.cs
yeojunh: I should also ask, `GetDocument` takes 2 arguments - `id` and `language`. This is from the assumption that if they know the ID, they will also know the language. `language` is also the partition key for the DB so it queries faster. Is this okay or would you prefer using `id` only for GET?

src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
yeojunh: Thanks I'll give it a shot! I'm not sure if I fully understand search index yet, but I'll give it a try and see if I can get it working.

src/dotnet/APIView/APIViewWeb/Managers/CopilotCommentsManager.cs
```diff
@@ -31,163 +31,101 @@ IConfiguration configuration
...
         }
 
-        private FilterDefinition<CopilotCommentModel> GetIdFilter(string id)
+        public async Task<string> SearchDocumentsAsync(string language, string badCode, float threshold, int limit)
         {
-            return Builders<CopilotCommentModel>.Filter
-                .Where(f => f.Id.ToString() == id && f.IsDeleted == false);
+            var embeddings = await GetEmbeddingsAsync(badCode);
+            var searchResults = await _copilotCommentsRepository.SimilaritySearchAsync(language, embeddings, threshold, limit);
+            var topResults = searchResults.Where(x => x.similarity >= threshold).AsEnumerable();
```
tg-msft, line 105: Minor nit - Do we need the `.AsEnumerable()` at the end?  [`.Where(...)`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.where?view=net-7.0#system-linq-enumerable-where-1(system-collections-generic-ienumerable((-0))-system-func((-0-system-boolean)))) also returns `IEnumerable<T>`.

tg-msft (approved):
Everything looks great!

JonathanCrd:
Thank you @yeojunh for your contributions! 🚀

kristapratico (approved):
Awesome work! Thank you so much @yeojunh 🐱

PR #6811, opened 2023-08-18
remove loadenv

xiangyan99:
Seems like loadenv() w/o .env makes web app behave weird.


tjprescott (changes requested):
Don't agree with this change. `.env` is way more convenient that having to manually enter environment variables every time someone loads their terminal. If the Flask app doesn't behave correctly, can't we just put an `.env` file on it?

tjprescott (approved):

PR #6823, opened 2023-08-21
Use emitter-package-lock.json and @local feed view

hallipr:
- Use local feed view in the generated npmrc file
- Use eng/emitter-package-lock.json if present

### DIFF
--- a/eng/common/scripts/TypeSpec-Project-Generate.ps1
+++ b/eng/common/scripts/TypeSpec-Project-Generate.ps1
@@ -46,14 +46,36 @@ function NpmInstallForProject([string]$workingDirectory) {
         Write-Host("Copying package.json from $replacementPackageJson")
         Copy-Item -Path $replacementPackageJson -Destination "package.json" -Force
 
+        #default to root/eng/emitter-package-lock.json but you can override by writing
+        #Get-${Language}-EmitterPackageLockPath in your Language-Settings.ps1
+        $emitterPackageLock = Join-Path $PSScriptRoot "../../emitter-package-lock.json"
+        if (Test-Path "Function:$GetEmitterPackageLockPathFn") {
+            $emitterPackageLock = &$GetEmitterPackageLockPathFn
+        }
+
+        $usingLockFile = Test-Path $emitterPackageLock
+
+        if ($usingLockFile) {
+            Write-Host("Copying package-lock.json from $emitterPackageLock")
+            Copy-Item -Path $emitterPackageLock -Destination "package-lock.json" -Force
+        }
+
         $useAlphaNpmRegistry = (Get-Content $replacementPackageJson -Raw).Contains("-alpha.")
 
         if($useAlphaNpmRegistry) {
             Write-Host "Package.json contains '-alpha.' in the version, Creating .npmrc using public/azure-sdk-for-js-test-autorest feed."
-            "registry=https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js-test-autorest/npm/registry/ `n`nalways-auth=true" | Out-File '.npmrc'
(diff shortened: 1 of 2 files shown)


eng/common/scripts/TypeSpec-Project-Generate.ps1
```diff
@@ -46,14 +46,26 @@ function NpmInstallForProject([string]$workingDirectory) {
...
         $useAlphaNpmRegistry = (Get-Content $replacementPackageJson -Raw).Contains("-alpha.")
 
         if($useAlphaNpmRegistry) {
             Write-Host "Package.json contains '-alpha.' in the version, Creating .npmrc using public/azure-sdk-for-js-test-autorest feed."
-            "registry=https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js-test-autorest/npm/registry/ `n`nalways-auth=true" | Out-File '.npmrc'
+            "registry=https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js-test-autorest@local/npm/registry/ `n`nalways-auth=true" | Out-File '.npmrc'
         }
 
-        npm install --no-lock-file
+        npm install
```
mikeharder, line 68: I think we should use `npm ci` if there is a lock file, and `npm install` if not.

mikeharder (changes requested):

eng/common/scripts/TypeSpec-Project-Generate.ps1
hallipr: Updated to conditionally use `npm ci` or `npm install`

mikeharder (approved):

eng/common/scripts/common.ps1
```diff
@@ -60,3 +60,4 @@ $GetRepositoryLinkFn = "Get-${Language}-RepositoryLink"
 $GetEmitterAdditionalOptionsFn = "Get-${Language}-EmitterAdditionalOptions"
 $GetEmitterNameFn = "Get-${Language}-EmitterName"
 $GetEmitterPackageJsonPathFn = "Get-${Language}-EmitterPackageJsonPath"
+$GetEmitterPackageLockPathFn = "Get-${Language}-EmitterPackageLockPath"
```
weshaggard, line 63: Do we need another function? Any reason we cannot assume it is in the same path as the packageJson?

mikeharder: This is a good point.  I think it's fine to require lockfile in same folder.

hallipr: The path is a full path.   The override Get-${Language}-EmitterPackageJsonPath allows the author to point to any path. It would be odd to assume that there will be `emitter/emitter-package-lock.json` beside `emitter/package.json`

weshaggard: If we do stick with both of these can we have a common convention, we can implement here instead of requiring them to be implemented in all languages. They can still be overridden if we do have cases that are special.

mikeharder: @hallipr: Can you share an example of where an SDK dev would want to put the lockfile in a different folder?

hallipr: No, but in the same way that I can't share an example of where an SDK dev would want to put the emitter-package.json in a different location.  Get-${Language}-EmitterPackageJsonPath is currently unused in any repo.

Rather than let the someone explicitly override `emitter-package.json` but leave `emitter-package-lock.json` inferred from that path, I'd rather we have both or neither.

hallipr: > If we do stick with both of these can we have a common convention, we can implement here instead of requiring them to be implemented in all languages. They can still be overridden if we do have cases that are special.

There is a common convention of `eng/emitter-package(-lock).json` with the override if required.  No one currently overrides either.

weshaggard: OK that is good, I didn't realize no one was overriding this yet. If that is the case I have to wonder if we should just remove these until a point where we need someone to override.

weshaggard (approved):

PR #6843, opened 2023-08-23
Create tool tsp-client

xirzec:
Per Issue #6838, this PR scaffolds out a new tool to invoke the TypeSpec compiler using Node/npm.

The largest enhancement is avoiding the need to clone the specs repository as URLs are supported as entrypoints rather than requiring the TSP files all be locally available.

Eventually this tool will grow to support all of the scenarios covered in https://github.com/Azure/azure-sdk-tools/blob/main/doc/common/TypeSpec-Project-Scripts.md

### DIFF
--- a/tools/tsp-client/README.md
+++ b/tools/tsp-client/README.md
@@ -1,0 +1,46 @@
+# tsp-client
+
+A simple command line tool for generating TypeSpec clients.
+
+### Usage
+```
+tsp-client [options] <outputDir>
+```
+
+## Options
+
+### <outputDir> (required)
+
+The only positional parameter. Specifies the directory to pass to the language emitter.
+
+### --emitter, -e (required)
+
+Specifies which language emitter to use. Current choices are "csharp", "java", "javascript", "python", "openapi".
+
+Aliases are also available, such as cs, js, py, and ts.
+
+### --mainFile, -m
+
+Used when specifying a URL to a TSP file directly. Not required if using a `tsp-location.yaml`
+
+### --debug, -d
+
+Enables verbose debug logging to the console.
+
+### --no-cleanup
+
+Disables automatic cleanup of the temporary directory where the TSP is written and referenced npm modules are installed.
+
+## Examples
+
+Generating from a TSP file to a particular directory:
+
+```
+tsp-client -e openapi -m https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/cognitiveservices/OpenAI.Inference/main.tsp ./temp
+```
+
+Generating in a directory that contains a `tsp-location.yaml`:
+
+```
+tsp-client sdk/openai/openai
+```
--- a/tools/tsp-client/package.json
+++ b/tools/tsp-client/package.json
@@ -10,7 +10,7 @@
     "example": "npx ts-node src/index.ts -e openapi -m https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/cognitiveservices/OpenAI.Inference/main.tsp ./temp",
     "test": "mocha"
   },
-  "author": "Jeff Fisher <jeffish@microsoft.com>",
+  "author": "Microsoft Corporation",
   "license": "MIT",
   "type": "module",
   "engines": {
--- a/tools/tsp-client/package.json
+++ b/tools/tsp-client/package.json
@@ -25,6 +25,7 @@
     "@types/chai": "^4.3.5",
     "@types/mocha": "^10.0.1",
     "@types/node": "^20.4.8",
+    "@types/prompt-sync": "^4.2.1",
     "chai": "^4.3.7",
     "mocha": "^10.2.0",
     "prettier": "^3.0.1",
@@ -38,7 +39,7 @@
     "yaml": "^2.3.1"
   },
   "peerDependencies": {
-    "@typespec/compiler": "^0.47.1"
+    "@typespec/compiler": ">=0.47.1 <1.0.0"
   },
   "mocha": {
     "extension": [
--- a/tools/tsp-client/package.json
+++ b/tools/tsp-client/package.json
@@ -1,0 +1,50 @@
+{
+  "name": "@azure-tools/tsp-client",
+  "version": "1.0.0",
+  "description": "A tool to generate Azure SDKs from TypeSpec",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "npm run clean && npm run build:tsc",
+    "build:tsc": "tsc",
+    "clean": "rimraf ./dist ./types",
+    "example": "npx ts-node src/index.ts -e openapi -m https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/cognitiveservices/OpenAI.Inference/main.tsp ./temp",
+    "test": "mocha"
+  },
+  "author": "Jeff Fisher <jeffish@microsoft.com>",
+  "license": "MIT",
+  "type": "module",
+  "engines": {
+    "node": ">=18.0.0"
+  },
+  "bin": "cmd/tsp-client.js",
+  "files": [
+    "dist",
+    "cmd/tsp-client.js"
+  ],
+  "devDependencies": {
+    "@types/chai": "^4.3.5",
+    "@types/mocha": "^10.0.1",
+    "@types/node": "^20.4.8",
+    "chai": "^4.3.7",
+    "mocha": "^10.2.0",
+    "prettier": "^3.0.1",
+    "rimraf": "^5.0.1",
+    "ts-node": "^10.9.1",
+    "typescript": "^5.1.6"
+  },
+  "dependencies": {
+    "@azure/core-rest-pipeline": "^1.12.0",
+    "chalk": "^5.3.0",
+    "yaml": "^2.3.1"
+  },
+  "peerDependencies": {
+    "@typespec/compiler": "^0.47.1"
+  },
+  "mocha": {
+    "extension": [
+      "ts"
+    ],
+    "spec": "test/**/*.spec.ts",
+    "loader": "ts-node/esm"
+  }
+}
--- a/tools/tsp-client/package.json
+++ b/tools/tsp-client/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@azure-tools/tsp-client",
-  "version": "1.0.0",
+  "version": "0.0.1",
+  "private": "true",
   "description": "A tool to generate Azure SDKs from TypeSpec",
   "main": "dist/index.js",
   "scripts": {
@@ -16,7 +17,9 @@
   "engines": {
     "node": ">=18.0.0"
   },
-  "bin": "cmd/tsp-client.js",
+  "bin": {
+    "tsp-client": "cmd/tsp-client.js"
+  },
   "files": [
     "dist",
     "cmd/tsp-client.js"
--- a/tools/tsp-client/package.json
+++ b/tools/tsp-client/package.json
@@ -36,10 +36,11 @@
   "dependencies": {
     "@azure/core-rest-pipeline": "^1.12.0",
     "chalk": "^5.3.0",
+    "prompt-sync": "^4.2.0",
     "yaml": "^2.3.1"
   },
   "peerDependencies": {
-    "@typespec/compiler": ">=0.47.1 <1.0.0"
+    "@typespec/compiler": ">=0.48.1 <1.0.0"
   },
   "mocha": {
     "extension": [
--- a/tools/tsp-client/src/fs.ts
+++ b/tools/tsp-client/src/fs.ts
@@ -65,14 +65,15 @@ export async function readTspLocation(rootDir: string): Promise<[string, string,
       }
       return [ directory, commit, repo, additionalDirectories ];
     }
+    throw new Error("Could not find tsp-location.yaml");
   } catch (e) {
     Logger.error(`Error reading tsp-location.yaml: ${e}`);
+    throw e;
   }
-  throw new Error("Could not find tsp-location.yaml");
 }
 
 
-export async function getEmitterFromRepoConfig(emitterPath: string): Promise<string | undefined> {
+export async function getEmitterFromRepoConfig(emitterPath: string): Promise<string> {
   await access(emitterPath);
   const data = await readFile(emitterPath, 'utf8');
   const obj = JSON.parse(data);
--- a/tools/tsp-client/src/fs.ts
+++ b/tools/tsp-client/src/fs.ts
@@ -82,8 +82,10 @@ export async function getEmitterFromRepoConfig(emitterPath: string): Promise<str
   const languages: string[] = ["@azure-tools/typespec-", "@typespec/openapi3"];
   for (const lang of languages) {
     const emitter = Object.keys(obj.dependencies).find((dep: string) => dep.startsWith(lang));
-    Logger.info(`Found emitter package ${emitter}`);
-    return emitter;
+    if (emitter) {
+      Logger.info(`Found emitter package ${emitter}`);
+      return emitter;
+    }
   }
   throw new Error("Could not find emitter package");
 }
--- a/tools/tsp-client/src/fs.ts
+++ b/tools/tsp-client/src/fs.ts
@@ -1,0 +1,49 @@
+import { mkdir, rm, writeFile, mkdtemp, stat, readFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { FileTreeResult } from "./fileTree.js";
+import * as path from "node:path";
+import { Logger } from "./log.js";
+import { parse as parseYaml } from "yaml";
+
+export async function ensureDirectory(path: string) {
+  await mkdir(path, { recursive: true });
+}
+
+export async function removeDirectory(path: string) {
+  await rm(path, { recursive: true, force: true });
+}
+
+export async function createTempDirectory(): Promise<string> {
+  const tempRoot = await mkdtemp(path.join(tmpdir(), "tsp-client-"));
+  Logger.debug(`Created temporary working directory ${tempRoot}`);
+  return tempRoot;
+}
+
+export async function writeFileTree(rootDir: string, files: FileTreeResult["files"]) {
+  for (const [relativeFilePath, contents] of files) {
+    const filePath = path.join(rootDir, relativeFilePath);
+    await ensureDirectory(path.dirname(filePath));
+    Logger.debug(`writing ${filePath}`);
+    await writeFile(filePath, contents);
+  }
+}
+
+export async function tryReadTspLocation(rootDir: string): Promise<string | undefined> {
+  try {
+    const yamlPath = path.resolve(rootDir, "tsp-location.yaml");
+    const fileStat = await stat(yamlPath);
+    if (fileStat.isFile()) {
+      const fileContents = await readFile(yamlPath, "utf8");
+      const locationYaml = parseYaml(fileContents);
+      const { directory, commit, repo } = locationYaml;
+      if (!directory || !commit || !repo) {
+        throw new Error("Invalid tsp-location.yaml");
+      }
+      // make GitHub URL
+      return `https://raw.githubusercontent.com/${repo}/${commit}/${directory}/`;
+    }
+  } catch (e) {
+    Logger.error(`Error reading tsp-location.yaml: ${e}`);
+  }
+  return undefined;
+}
--- a/tools/tsp-client/src/fs.ts
+++ b/tools/tsp-client/src/fs.ts
@@ -1,5 +1,4 @@
-import { mkdir, rm, writeFile, mkdtemp, stat, readFile } from "node:fs/promises";
-import { tmpdir } from "node:os";
+import { mkdir, rm, writeFile, stat, readFile, access } from "node:fs/promises";
 import { FileTreeResult } from "./fileTree.js";
 import * as path from "node:path";
 import { Logger } from "./log.js";
@@ -13,8 +12,9 @@ export async function removeDirectory(path: string) {
   await rm(path, { recursive: true, force: true });
 }
 
-export async function createTempDirectory(): Promise<string> {
-  const tempRoot = await mkdtemp(path.join(tmpdir(), "tsp-client-"));
+export async function createTempDirectory(outputDir: string): Promise<string> {
+  const tempRoot = path.join(outputDir, "TempTypeSpecFiles");
+  await mkdir(tempRoot, { recursive: true });
   Logger.debug(`Created temporary working directory ${tempRoot}`);
   return tempRoot;
 }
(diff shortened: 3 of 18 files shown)


tools/tsp-client/src/npm.ts
```diff
@@ -0,0 +1,51 @@
...
+    dependencies,
+  });
+
+  const filePath = path.join(rootPath, "package.json");
+  await writeFile(filePath, packageJson);
+}
+
+export function installDependencies(workingDir: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const npm = spawn("npm", ["install", "--no-lock-file"], {
```
weshaggard, line 23: I suspect this was mimicing the existing scripts. Note there is a PR in flight for supporting of the lock file, see https://github.com/Azure/azure-sdk-tools/pull/6823.

tools/tsp-client/src/network.ts
```diff
@@ -0,0 +1,92 @@
...
+        }
+      } else {
+        Logger.debug(`discovered module import ${fileImport}`);
+        moduleImports.add(fileImport);
+      }
+    }
+  }
+
+  // look for a tspconfig.yaml next to the root
+  try {
```
weshaggard, line 79: Will the tsp and tspconfig files be enough? Do we need any other files to make this project work?

tools/tsp-client/src/network.ts
```diff
@@ -0,0 +1,92 @@
...
+  const moduleImports = new Set<string>();
+  // fetch the root file
+  const filesToProcess = [rootUrl];
+  seenFiles.add(rootUrl);
+  const fileTree = createFileTree(rootUrl);
+
+  while (filesToProcess.length > 0) {
+    const url = filesToProcess.shift()!;
+    const sourceFile = await fetch(url);
+    fileTree.addFile(url, sourceFile);
```
weshaggard, line 51: Will this handle relative paths outside of the main directory? That is an expected scenario for any shared tsp projects.

tools/tsp-client/src/network.ts
```diff
@@ -0,0 +1,92 @@
...
+  const seenFiles = new Set<string>();
+  const moduleImports = new Set<string>();
+  // fetch the root file
+  const filesToProcess = [rootUrl];
+  seenFiles.add(rootUrl);
+  const fileTree = createFileTree(rootUrl);
+
+  while (filesToProcess.length > 0) {
+    const url = filesToProcess.shift()!;
+    const sourceFile = await fetch(url);
```
weshaggard, line 50: Depending on the number of files fetching them each individually from github has proven to be error prone due to rate limiting. That is why the PS version of the script did a partial git clone instead.

tools/tsp-client/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+  emitter,
+  noCleanup,
+  emitterOutputPath,
+}: {
+  rootUrl: string;
+  emitter: string;
+  noCleanup: boolean;
+  emitterOutputPath: string;
+}) {
+  const { moduleImports, fileTree } = await downloadTsp(rootUrl);
```
weshaggard, line 39: We should probably support a mode that can work against a local spec clone as well to by-pass the need to download any files.

tools/tsp-client/src/network.ts
catalinaperalta: I was talking to Laurent yesterday and we'll start off with the sparse clone for now and mimic what the pwsh scripts are doing

tools/tsp-client/src/npm.ts
xirzec: This is mostly because I'm dynamically generating the lockfile as outside of JS the other repo lockfiles only appear to pull in the emitter package rather than all the dependencies the TSP has. 

It'd still be useful to parse the local lockfile in the repo, but I figured that could be added in.

tools/tsp-client/src/network.ts
xirzec: As far as I know, nothing else is read by `tsp` when compiling a folder (with the minor exception of package.json/node_modules where it's checking for a different compiler version)

tools/tsp-client/src/network.ts
xirzec: It will handle whatever files are relatively referenced so long as they are not on a different URL host. We could extend it to support multiple hosts, but I didn't think it would be likely to be needed for now.

tools/tsp-client/src/network.ts
xirzec: Yeah, not too difficult to support both ways of doing it, this was just the fastest way I could figured out which worked for any web-accessible URL rather than something that needs a git repo

tools/tsp-client/src/index.ts
xirzec: Yeah! We could either copy from a local location or even work out of it if we weren't worried about polluting the source with `node_modules` folders. It might be a nice way to bootstrap a "tweak the TSP and iterate" workflow.

tools/tsp-client/src/network.ts
benbp: If examples of sparse cloning would be helpful you can find some here: https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/steps/sparse-checkout.yml#L46-L75, however you likely won't need to set `--no-cone` mode.

tools/tsp-client/src/fs.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export async function tryReadTspLocation(rootDir: string): Promise<string | undefined> {
+  try {
+    const yamlPath = path.resolve(rootDir, "tsp-location.yaml");
+    const fileStat = await stat(yamlPath);
+    if (fileStat.isFile()) {
+      const fileContents = await readFile(yamlPath, "utf8");
+      const locationYaml = parseYaml(fileContents);
+      const { directory, commit, repo } = locationYaml;
```
raych1, line 38: Is `additionalDirectories` removed from the schema of `tsp-location.yaml`? How are the referenced typespec identified when generate sdk code at sdk repository?

tools/tsp-client/README.md
```diff
@@ -0,0 +1,46 @@
+# tsp-client
+
+A simple command line tool for generating TypeSpec clients.
```
raych1, line 3: Does this tool target to be consumed in CI pipeline?

tools/tsp-client/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Usage
+```
+tsp-client [options] <outputDir>
+```
+
+## Options
+
+### <outputDir> (required)
```
raych1, line 12: Can `outputDir` is made as optional and compute the value by parsing tspconfig.yaml as `service-dir` and `package-dir` are specified there?

tools/tsp-client/README.md
```diff
@@ -0,0 +1,46 @@
...
+### --no-cleanup
+
+Disables automatic cleanup of the temporary directory where the TSP is written and referenced npm modules are installed.
+
+## Examples
+
+Generating from a TSP file to a particular directory:
+
+```
+tsp-client -e openapi -m https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/cognitiveservices/OpenAI.Inference/main.tsp ./temp
```
raych1, line 39: Can we support github normal URL and transform it to raw content url in code? That would be more friendly from user perspective.

tools/tsp-client/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+  const { mainFilePath, files } = await fileTree.createTree();
+  const tempRoot = await createTempDirectory();
+  const srcDir = path.join(tempRoot, "src");
+  await writeFileTree(srcDir, files);
+
+  const emitterPackage = getEmitterPackage(emitter);
```
raych1, line 45: Can you help me understand where to load the emitter version?

tools/tsp-client/src/fs.ts
weshaggard: Given that this tool is analyzing the imports we might not need the additionalDirectories any longer but I wouldn't remove that until we have verification that this is working as expected.

tools/tsp-client/README.md
weshaggard: I would hope it does get used by any automation that needs to do these workflows.

witemple-msft:
This looks very nice.

I'm wondering a little bit if we could do anything with the TSP compiler host API to make it so that you don't even have to parse the files and traverse the import nodes, and we could just invoke `compile` with our own custom host. For example, if we could wire up a host that by default does a sparse checkout of the entrypoint's directory, then simply checks out more directories if they're imported from outside of that tree, you could just call `compile` and have it be done... maybe?

tools/tsp-client/src/fs.ts
xirzec: ah since openai wasn't using this feature I didn't write support for it and as Wes mentions, it's technically only needed for clone support, but since we want to support that we can fish it out of here as well.

tools/tsp-client/README.md
xirzec: Yeah, it should work as part of a CI pipeline, not sure if we have any particular requirements for CLI tools to make them more CI friendly, but happy to adjust anything we need.

tools/tsp-client/README.md
xirzec: ah so you'd pass the `tspconfig.yaml` in the specs repo directly and then we'd calculate the output location? I suppose this would work okay though we'd need to know the root of where you are generating (e.g. your clone of the language repo) if it wasn't the cwd

tools/tsp-client/README.md
xirzec: Yes, we do this already

tools/tsp-client/src/index.ts
xirzec: Currently I don't try to figure out the emitter version and just grab latest. I was thinking that if we notice the output path has a folder of the format `azure-sdk-for-{something}` we could look for the `package.json` and get the emitter version from there. We could also choose the appropriate language emitter this way.

xirzec:
> I'm wondering a little bit if we could do anything with the TSP compiler host API to make it so that you don't even have to parse the files and traverse the import nodes, and we could just invoke `compile` with our own custom host. For example, if we could wire up a host that by default does a sparse checkout of the entrypoint's directory, then simply checks out more directories if they're imported from outside of that tree, you could just call `compile` and have it be done... maybe?

Potentially! Though I think the default algorithm that looks for `package.json` up the tree might cause us to sparse checkout every level of the path up to the root of the repo. Unless we tried to handle that specially and ignore it.

tools/tsp-client/src/index.ts
catalinaperalta: My idea is to look for the emitter-package.json to get the emitter version. From my discussions with @lmazuel I think one of the goals here is to use this tool from the language repo clone, with something as simple as: `tsp-client generate`. If folks pass in certain flags or directories then we could customize it to the given parameters.

catalinaperalta: FYI I'm working on this already and will open a PR against this branch once it's ready.

tools/tsp-client/src/fs.ts
raych1: That works if this tool can analyze the entry tsp file then fetch the referenced typespec from github.

tools/tsp-client/src/index.ts
hallipr: @xirzec or @catalinaperalta, autorest.csharp was planning on floating the typespec dependencies and moving them to peerDependencies in the emitters, then .   This would allow typespec version flexibility for emitter consumers, but still pinning the version in azure-sdk-for-net in emitter-package.json as a devDependency:

[/eng/emitter-package.json](https://github.com/azure-sdk/azure-sdk-for-net/blob/auto-update-typespec-emitter-alpha-20230913.1/eng/emitter-package.json)
``` json
  "dependencies": {
    "@azure-tools/typespec-csharp": "0.2.0-alpha.20230913.1.x"
  },
  "devDependencies": {
    "@typespec/compiler": "0.47.1",
    "@typespec/versioning": "0.47.0",
    "@typespec/rest": "0.47.0",
    "@typespec/eslint-config-typespec": "0.47.0",
    "@azure-tools/typespec-client-generator-core": "0.33.0",
    "@typespec/http": "0.47.0",
    "@azure-tools/typespec-azure-core": "0.33.1"
  }
```

We also pin the versions in the accompanying [emitter-package-lock.json](https://github.com/azure-sdk/azure-sdk-for-net/blob/auto-update-typespec-emitter-alpha-20230913.1/eng/emitter-package-lock.json).

Will tsp-client use the lock file and devDependencies if they exists?

mikeharder: I agree, this new tool will need to include the changes @hallipr is working on.

tools/tsp-client/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "description": "A tool to generate Azure SDKs from TypeSpec",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "npm run clean && npm run build:tsc",
+    "build:tsc": "tsc",
+    "clean": "rimraf ./dist ./types",
+    "example": "npx ts-node src/index.ts -e openapi -m https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/cognitiveservices/OpenAI.Inference/main.tsp ./temp",
+    "test": "mocha"
+  },
+  "author": "Jeff Fisher <jeffish@microsoft.com>",
```
xirzec, line 13:
```suggestion
  "author": "Microsoft Corporation",
```

tools/tsp-client/src/fs.ts
```diff
@@ -0,0 +1,89 @@
...
+export async function getEmitterFromRepoConfig(emitterPath: string): Promise<string | undefined> {
+  await access(emitterPath);
+  const data = await readFile(emitterPath, 'utf8');
+  const obj = JSON.parse(data);
+  if (!obj || !obj.dependencies) {
+    throw new Error("Invalid emitter-package.json");
+  }
+  const languages: string[] = ["@azure-tools/typespec-", "@typespec/openapi3"];
+  for (const lang of languages) {
+    const emitter = Object.keys(obj.dependencies).find((dep: string) => dep.startsWith(lang));
```
xirzec, line 84: `find` will always return a matching item or `undefined`, which means the `return` two lines down will always happen and we'll never look for `@typespec/openapi3` or throw the Error at the bottom.

I think we need to test if `emitter` is truthy before returning? And perhaps return undefined at the bottom instead of throwing?

tools/tsp-client/src/fs.ts
```diff
@@ -0,0 +1,89 @@
...
+      Logger.info(`Additional directories: ${additionalDirectories}`)
+      if (!additionalDirectories) {
+        additionalDirectories = [];
+      }
+      return [ directory, commit, repo, additionalDirectories ];
+    }
+  } catch (e) {
+    Logger.error(`Error reading tsp-location.yaml: ${e}`);
+  }
+  throw new Error("Could not find tsp-location.yaml");
```
xirzec, line 71: maybe just rethrow the above exception in the catch instead?

tools/tsp-client/src/git.ts
```diff
@@ -0,0 +1,82 @@
+import { execSync, spawn } from "child_process";
+
+export function getRepoRoot(): string {
+    return execSync('git rev-parse --show-toplevel').toString().trim();
```
xirzec, line 4: use `exec` instead of `execSync` no sense in hanging the process while git does its thing

tools/tsp-client/src/git.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+  
+export async function cloneRepo(rootUrl: string, cloneDir: string, repo: string): Promise<void> {
+    return new Promise((resolve, reject) => {
+      const git = spawn("git", ["clone", "--no-checkout", "--filter=tree:0", repo, cloneDir], {
```
xirzec, line 9: seems like we could make a spawn helper since there's a lot of repeated code in this file

Also keep in mind any strings passed to spawn allow for shell injection attacks

tools/tsp-client/src/index.ts
```diff
@@ -0,0 +1,281 @@
...
+    commit: string | undefined;
+    repo: string | undefined;
+    isUrl: boolean;
+  }): Promise<string> {
+  if (isUrl) {
+    // URL scenario
+    const resolvedConfigUrl = await resolveTspConfigUrl(config);
+    Logger.debug(`Resolved config url: ${resolvedConfigUrl.resolvedUrl}`)
+    const tspConfig = await fetch(resolvedConfigUrl.resolvedUrl);
+    const configYaml = parseYaml(tspConfig);
```
xirzec, line 35: Can you declare an interface for the expected shape of tspConfig so we don't have to do the heavy `["foo"]` syntax all over the place?

tools/tsp-client/src/index.ts
```diff
@@ -0,0 +1,281 @@
...
+    if (configYaml["parameters"] && configYaml["parameters"]["service-dir"]){
```
xirzec, line 36: You can use optional chaining to make this so much cleaner looking
```suggestion
    if (configYaml.parameters?.service-dir){
```

tools/tsp-client/src/index.ts
```diff
@@ -0,0 +1,281 @@
...
+      const serviceDir = configYaml["parameters"]["service-dir"]["default"];
+      if (serviceDir === undefined) {
```
xirzec, line 38: style nit
```suggestion
      if (!serviceDir) {
```

tools/tsp-client/src/index.ts
```diff
@@ -0,0 +1,281 @@
...
+        Logger.error(`Parameter service-dir is not defined correctly in tspconfig.yaml. Please refer to https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml for the right schema.`)
+      }
+      Logger.debug(`Service directory: ${serviceDir}`)
+      let additionalDirs: string[] = [];
+      if (configYaml["parameters"]["dependencies"] && configYaml["parameters"]["dependencies"]["additionalDirectories"]) {
+        additionalDirs = configYaml["parameters"]["dependencies"]["additionalDirectories"];
+      }
```
xirzec, line 45: the nullish coalescing operator is your friend here
```suggestion
      const additionalDirs: string[] = configYaml?.parameters?.dependencies?.additionalDirectories ?? [];
```

tools/tsp-client/src/index.ts
```diff
@@ -0,0 +1,281 @@
...
+      let packageDir;
+      if (configYaml["options"][emitter] && configYaml["options"][emitter]["package-dir"]) {
+        packageDir = configYaml["options"][emitter]["package-dir"];
+      }
```
xirzec, line 49:
```suggestion
      const packageDir = configYaml?.options?[emitter]?.package-dir;
```

tools/tsp-client/src/options.ts
```diff
@@ -0,0 +1,168 @@
+import { parseArgs } from "node:util";
+import { Logger, printUsage, printVersion } from "./log.js";
+import * as path from "node:path";
+import { knownLanguages, languageAliases } from "./languageSettings.js";
+import { doesFileExist } from "./network.js";
+import PromptSync from "prompt-sync";
```
xirzec, line 6: I saw typings for this, but not an actual dependency line?

tools/tsp-client/src/typespec.ts
```diff
@@ -0,0 +1,154 @@
...
+    if (!(emitterName! in options)) {
+      options[emitterName!] = {};
+    }
+    options[emitterName!]![key!] = optionParts[1]!;
+  }
+  return options;
+}
+
+
+export async function resolveTspConfigUrl(configUrl: string): Promise<{
```
xirzec, line 73: not sure why this is async given you're doing a regex?

tools/tsp-client/src/typespec.ts
```diff
@@ -0,0 +1,154 @@
...
+  } else {
+    const configData = await readFile(path.join(tempRoot, "tspconfig.yaml"), "utf8");
+    const configYaml = parseYaml(configData);
+  
+    if (configYaml["options"] && configYaml["options"][emitter]){
+      emitterOptions[emitter] = configYaml["options"][emitter];
+      for (const key in emitterOptions[emitter]!) {
+        Object.keys(emitterOptions![emitter]!).forEach(
+          (k) => {
+            if (`{${k}}` === emitterOptions[emitter]![key]) {
```
xirzec, line 24: what is this trying to do? It almost looks like it's trying to replace the curly brace tokens in emitter options, but the logic doesn't really make sense to me

tools/tsp-client/src/typespec.ts
```diff
@@ -0,0 +1,154 @@
...
+  emitterOptions[emitter] = {};
+  if (additionalOptions) {
+    emitterOptions = resolveCliOptions(additionalOptions.split(","));
+  } else {
+    const configData = await readFile(path.join(tempRoot, "tspconfig.yaml"), "utf8");
+    const configYaml = parseYaml(configData);
+  
+    if (configYaml["options"] && configYaml["options"][emitter]){
+      emitterOptions[emitter] = configYaml["options"][emitter];
+      for (const key in emitterOptions[emitter]!) {
```
xirzec, line 21: for/of Object.entries would likely suit you much better

xirzec (commented):
Apparently since I made the PR I can't request changes, but left some comments on a few things I noticed

tools/tsp-client/src/npm.ts
catalinaperalta: Tracking issue to use lock file: https://github.com/Azure/azure-sdk-tools/issues/7057

tools/tsp-client/src/index.ts
catalinaperalta: We're currently using the `emitter-package.json` as is to install dependencies, so that should just work for you. Here's a tracking issue for supporting the lock file: https://github.com/Azure/azure-sdk-tools/issues/7057

tools/tsp-client/src/fs.ts
catalinaperalta: Done, still throwing at the bottom because we need an emitter package to be able to generate.

tools/tsp-client/src/fs.ts
catalinaperalta: I'm throwing here because the typing system didnt like that there was no return at the bottom of this function and I didnt want to add a `return undefined` here which would just make us check for undefined after the function returns then throw there. Let me know if there's something else that should be done here!

tools/tsp-client/src/git.ts
catalinaperalta: Tracking issue for future PR: https://github.com/Azure/azure-sdk-tools/issues/7045

tools/tsp-client/src/git.ts
catalinaperalta: Tracking issue for future PR: https://github.com/Azure/azure-sdk-tools/issues/7046

tools/tsp-client/src/index.ts
catalinaperalta: Tracking issue to improve in a future PR: https://github.com/Azure/azure-sdk-tools/issues/7047

tools/tsp-client/src/index.ts
catalinaperalta: Will do this refactoring in a future PR since I'm doing this in several places, tracking issue: https://github.com/Azure/azure-sdk-tools/issues/7048

tools/tsp-client/src/index.ts
catalinaperalta: Similar to the above, I'm doing this in several places, will address style improvements in a future PR: https://github.com/Azure/azure-sdk-tools/issues/7049

tools/tsp-client/src/options.ts
catalinaperalta: It's used on line 139-140 to confirm the correct output directory for the generated code.

tools/tsp-client/src/typespec.ts
catalinaperalta: There's a very common config in the tspconfig.yaml files where people are defining `package-name` by referencing their `{package-dir}` option. I updated this code and simplified it to only account for this case for now. 

Here is a tracking issue to investigate whether we need to do something beyond this simple replacement and work on a more generic solution: https://github.com/Azure/azure-sdk-tools/issues/7056

tools/tsp-client/src/typespec.ts
catalinaperalta: This code was changed.

tools/tsp-client/src/fs.ts
xirzec: I think if you change the above to 

```ts
  } catch (e) {
    Logger.error(`Error reading tsp-location.yaml: ${e}`);
    throw e;
  }
```

or if you want to wrap the original error

```ts
  } catch (e) {
    Logger.error(`Error reading tsp-location.yaml: ${e}`);
    throw new Error(`Error reading tsp-location.yaml: ${e}`);
  }
```

Then you can move the `throw new Error("Could not find tsp-location.yaml");` line after the `if (fileStat.isFile()) {` check so it looks a little tidier. Really just a style nit, it felt odd to catch an error and log then immediately throw afterwards

tools/tsp-client/src/fs.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+
+export async function getEmitterFromRepoConfig(emitterPath: string): Promise<string | undefined> {
```
xirzec, line 75: can change the return here to be `Promise<string>` if we are not allowing returning undefined

xirzec (commented):
Alas I still can't give you approval because I count as the PR originator:

image


But morally I approve ✅ (and perhaps you could approve this PR yourself?)

tools/tsp-client/src/git.ts
mikeharder, line 1: You might want to use a library like `simple-git` instead of directly calling the `git` executable.  We use it in TypeSpecValidation:

https://github.com/Azure/azure-rest-api-specs/blob/a26794fe26cdc68f4d18bc782e0b4f2f29e24b40/eng/tools/TypeSpecValidation/src/rules/git-diff.ts#L11

mikeharder (approved):
Approved with optional feedback.

allenjzhang (approved):

PR #6849, opened 2023-08-24
Update the team/user store to store additional information to be used in the upcoming CODEOWNERS linter

JimSuplizio:
The team/user store used to only store team/user data for teams with write permissions (teams under, and including, azure-sdk-write). _This data is used to expand teams in CODEOWNERS files for Azure repositories._ During the analysis for the development of a CODEOWNERS linter, it was quickly discovered that we needed more than just the team/user data. We needed the following:

1. User/Org visibility data. Our [documentation](https://eng.ms/docs/products/azure-developer-experience/onboard/access-repos) states that user's Azure org membership must be public and this information was needed to detect this specific case. The fact is, even GitHub's CODEOWNERS verification (which only verifies owners on source path/owner lines) flags owners as invalid if their membership to Azure is private, regardless of whether or not the owner has write permission for the repository. This information is also not a security risk as anyone can look at a GitHub user and see what they're public members of. [This is underlying GitHub API call Octokit.Net is using](https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28#check-public-organization-membership-for-a-user). Only the users that are part of azure-sdk-write are checked. The data is stored as a Dictionary<string, bool>
2. Repository Label data. There is metadata in our CODEOWNERS files that includes labels (PRLabel and ServiceLabel) which need to be verified. The [list of repositories](https://github.com/Azure/azure-sdk-tools/blob/main/tools/github/data/repositories.txt) that labels are being collected for is the same list of repositories that we sync a common set of labels to. The common list can't be used because it only contains the labels that are common to all of the repositories and each one has their repository specific labels. 

There was a secondary change that's not reflected in this PR. The GitHub PAT being used to collect this data had to be updated. Previously, the PAT only required Organization->Members:read-only to get the user/org data. The label data, specifically for our private repositories, required Repository->Issues:read-only and Repository->PullRequests:read-only permissions.

This does increate the number of calls made to the GitHub API, when the pipeline-owners-extraction pipeline runs, from roughly 200 to about 1500 which is still way under the 5000/hour limit. Given the pipeline only runs once daily, that's okay.

Q) Why do we need to get this data here, instead of having the linter get it?
A) The plan is to have the linter run in CI pipeline which means no secrets. While the data being stored in blob storage isn't a security risk, having a PAT in a CI pipeline most certainly would be.

### DIFF
--- a/tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/RepositoryLabelGenerator.cs
+++ b/tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/RepositoryLabelGenerator.cs
@@ -1,0 +1,119 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Azure.Storage.Blobs;
+using GitHubTeamUserStore.Constants;
+using Octokit;
+
+namespace GitHubTeamUserStore
+{
+    public class RepositoryLabelGenerator
+    {
+        // Repository name is the key, with the list of that repository's labels as the as the value
+        private static Dictionary<string, HashSet<string>> _repoLabelDict = new Dictionary<string, HashSet<string>>();
+
+        public static async Task<bool> GenerateAndStoreRepositoryLabels(GitHubEventClient gitHubEventClient, string repoLabelBlobStorageUri, string repositoryListFile)
+        {
+            Uri repoLabelBlobUri = new Uri(repoLabelBlobStorageUri);
+            BlobUriBuilder repoLabelBlobUriBuilder = new BlobUriBuilder(repoLabelBlobUri);
+
+            // Load the repostiory list file
+            string fullPath = Path.GetFullPath(repositoryListFile);
+            if (!File.Exists(fullPath))
+            {
+                throw new ArgumentException($"The path provided '{repositoryListFile}' does not exist");
+            }
+
+            var repositories = File.ReadAllLines(fullPath);
+            foreach (string repository in repositories)
+            {
+                if (string.IsNullOrWhiteSpace(repository))
+                {
+                    continue;
+                }
+                // The repositories in the file will all start with "Azure/" which is fine for storage considering
+                // that the $(Build.Repository.Name), in a pipeline, will also start with "Azure/" but "Azure/"
+                // needs to be stripped off for the GetRepositoryLabels call which requires the Org and repository
+                // be separate arguments. The dictionary key will be full repository name.
+                string repoWithoutOrg = repository;
+                if (repoWithoutOrg.StartsWith($"{ProductAndTeamConstants.Azure}/"))
+                {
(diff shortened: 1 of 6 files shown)


tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/RepositoryLabelGenerator.cs
```diff
@@ -0,0 +1,119 @@
...
+using Azure.Storage.Blobs;
+using GitHubTeamUserStore.Constants;
+using Octokit;
+
+namespace GitHubTeamUserStore
+{
+    public class RepositoryLabelGenerator
+    {
+        // Repository name is the key, with the list of that repository's labels as the as the value
+        private static Dictionary<string, HashSet<string>> _repoLabelDict = new Dictionary<string, HashSet<string>>();
```
weshaggard, line 16: Why does this need to be static? I would suggest making this a normal member and make this a class or local to a single static method. In general folks should try avoid having global static state like this.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/TeamUserGenerator.cs
```diff
@@ -15,6 +16,10 @@ public class TeamUserGenerator
         // that this is the list of Logins (Octokit.User.Login) which are what's used in CODEOWNERS, @mentions, etc.
         private static Dictionary<string, List<string>> _teamUserDict = new Dictionary<string, List<string>>();
 
+        // The user list of azure-sdk-write will contain a distinct list of all users from all child teams. For
+        // each one, check if they're a member of azure and store that in here.
+        private static Dictionary<string, bool> _userOrgDict = new Dictionary<string, bool>();
```
weshaggard, line 21: Similar feedback about having unnecessary static global values (similar to the teamUserDic which I would have similar feedback on). While what you have will work it is generally a bad pattern if this library did get used by other things.

weshaggard (approved):
Left a couple comments about avoiding global static variables but otherwise the change looks good.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/RepositoryLabelGenerator.cs
JimSuplizio: This will be updated with the next commit.

tools/github-team-user-store/GitHubTeamUserStore/GitHubTeamUserStore/TeamUserGenerator.cs
JimSuplizio: This will be updated with the next commit.

JimSuplizio:
The pipeline-owners-extraction was run against this PR and passed (No, it shouldn't automatically trigger because of the number of GitHub API calls it makes). I'm going to override check-enforcer.

PR #6855, opened 2023-08-25
Integration with Vector DB

tjprescott:
This PR provides scripts to interact with the Vector DB endpoints. 

Updates the grounding logic based on the results of the semantic similarity search


packages/python-packages/apiview-gpt/src/__init__.py
```diff
@@ -13,7 +15,7 @@ def console_entry_point():
     print("Running apiview-gpt version {}".format(__version__))
     reviewer = GptReviewer()
     # FIXME: Make this a proper CLI
-    input_filename = "test2.txt"
+    input_filename = "test.txt"
```
kristapratico, line 18: It looks like this file got deleted?

packages/python-packages/apiview-gpt/src/_gpt_reviewer.py
```diff
@@ -29,50 +31,70 @@ class GptReviewer:
...
+
         self.prompt_template = PromptTemplate(
-            input_variables=["apiview", "guidelines", "language"],
+            input_variables=["apiview", "guidelines", "language", "extra_comments", "class_list"],
             partial_variables={"format_instructions": self.output_parser.get_format_instructions()},
             template="""
-                Given the following {language} Azure SDK Guidelines:
-                  {guidelines}
-                Verify whether the following code satisfies the guidelines:
+                You are trying to analyze an API for {language} to determine whether it meets the SDK guidelines. We only provide one class at a time right now, but if you need it, here's a list of all the classes in this API:
```
kristapratico, line 40: Did you see better results by providing all the class names?

packages/python-packages/apiview-gpt/src/_gpt_reviewer.py
```diff
@@ -29,50 +31,70 @@ class GptReviewer:
...
+        all_guidelines = self.retrieve_guidelines(language)
 
         chunked_apiview = SectionedDocument(apiview.splitlines(), chunk=True)
 
         final_results = GuidelinesResult(status="Success", violations=[])
         for chunk in chunked_apiview.sections:
             if self.should_evaluate(chunk):
-                results = self.chain.run(apiview=str(chunk), guidelines=guidelines, language=language)
+                # retrieve the most similar comments to identify guidelines to check
+                semantic_matches = VectorDB().search_documents(language, chunk)
```
kristapratico, line 72: for now, are we just taking all the matches? edit: nvm I see we're using the defaults currently

kristapratico (approved):
Looks good!

mpodwysocki (approved):

packages/python-packages/apiview-gpt/src/__init__.py
tjprescott: Well, now you actually pass in the value, so I should refactor this logic away.

packages/python-packages/apiview-gpt/src/_gpt_reviewer.py
tjprescott: Sort of. It no longer sees just one client (since it runs class by class) and complains "You should provide a sync and async client!"

PR #6862, opened 2023-08-26
Disable GPG signing on commit

### LINKED ISSUE #6861: Error 5 (access denied) when commit.gpgsign is true
I have GPG signing enabled globally, which causes an error 5 (access denied) without prompting as git normally does (off hand, I'm not sure why that is). Perhaps we should pass `-c commit.gpgsign=false` to make sure it's not a problem.

@scbedd and I had previously talked about this in relation to #4720 and `--no-sign` was added to `git tag`, but commit signing is still a problem.

heaths:
Fixes #6861


scbedd (approved):
👍 Thanks Heath!

scbedd:
DevOps doesn't handle quick event flow properly.

image

Cancelled the damn build. I just kicked it again.

PR #6887, opened 2023-08-31
[tsp-client] Refactor tsp-client

catalinaperalta:
Addressing #6838 
Refactoring to require less input from the user + separate commands out into `init`, `sync`, `generate`, and `update` to correspond with the behavior of the current powershell scripts.


tools/tsp-client/src/fs.ts
```diff
@@ -13,8 +13,9 @@ export async function removeDirectory(path: string) {
   await rm(path, { recursive: true, force: true });
 }
 
-export async function createTempDirectory(): Promise<string> {
-  const tempRoot = await mkdtemp(path.join(tmpdir(), "tsp-client-"));
+export async function createTempDirectory(outputDir: string): Promise<string> {
```
catalinaperalta, line 16: TODO look into combining the helper functions for both the download and github clone scenarios

tools/tsp-client/src/fs.ts
```diff
@@ -47,3 +48,48 @@ export async function tryReadTspLocation(rootDir: string): Promise<string | unde
...
+  } catch (e) {
+    Logger.error(`Error reading tsp-location.yaml: ${e}`);
+  }
+  throw new Error("Could not find tsp-location.yaml");
+}
+
+
+export async function findEmitterPackage(emitterPath: string): Promise<string | undefined> {
+  if (existsSync(emitterPath)) {
+    const emitter = await readFile(emitterPath, 'utf8').then(data => {
```
xirzec, line 78: please don't mix `await` and promise chaining with `then`, it can create some headaches.

tools/tsp-client/src/fs.ts
```diff
@@ -47,3 +48,48 @@ export async function tryReadTspLocation(rootDir: string): Promise<string | unde
...
+      return [ directory, commit, repo, additionalDirectories ];
+    }
+  } catch (e) {
+    Logger.error(`Error reading tsp-location.yaml: ${e}`);
+  }
+  throw new Error("Could not find tsp-location.yaml");
+}
+
+
+export async function findEmitterPackage(emitterPath: string): Promise<string | undefined> {
```
xirzec, line 76: I think this method could be named a little better since what it appears to do is take the path to a repository `package.json` that *might* contain an emitter package dependency, but that's not very clear

Also it's a little strange that we resolve to a single exporter when theoretically the `emitter-package.json` could have multiple emitters. Should we support that?

tools/tsp-client/src/fs.ts
```diff
@@ -47,3 +48,48 @@ export async function tryReadTspLocation(rootDir: string): Promise<string | unde
...
+  if (existsSync(emitterPath)) {
+    const emitter = await readFile(emitterPath, 'utf8').then(data => {
+      const obj = JSON.parse(data);
+      if (!obj || !obj.dependencies) {
+        throw new Error("Invalid emitter-package.json");
+      }
+      var languages: string[] = ["@azure-tools/typespec-csharp", "@azure-tools/typespec-java", "@azure-tools/typespec-ts", "@azure-tools/typespec-python", "@typespec/openapi3"];
```
xirzec, line 83: avoid `var` at all costs, always use `let` or `const`

tools/tsp-client/src/fs.ts
```diff
@@ -47,3 +48,48 @@ export async function tryReadTspLocation(rootDir: string): Promise<string | unde
...
+      for (var lang in languages) {
```
xirzec, line 84: avoid `for/in` and always use `for/of`, it will save you indexing into things needlessly.

tools/tsp-client/src/fs.ts
```diff
@@ -47,3 +48,48 @@ export async function tryReadTspLocation(rootDir: string): Promise<string | unde
...
+        if (obj.dependencies[languages[lang]!]) {
```
xirzec, line 85: instead of checking a hardcoded list, could we reduce the dependencies array to look for something that starts with `@azure-tools/typespec-` and take the first one? I suppose we'd still have to handle openapi3 specially, but otherwise it seems more extensible

tools/tsp-client/src/gh.ts
```diff
@@ -0,0 +1,97 @@
+import { execSync, spawn } from "child_process";
+import { doesFileExist } from "./network.js";
+import { removeDirectory } from "./fs.js";
+import { mkdir } from "fs/promises";
+import { Logger } from "./log.js";
+
+export function getRepoRoot(): string {
+    return execSync('git rev-parse --show-toplevel').toString().trim();
```
xirzec, line 8: never use a `sync` method -- it blocks the whole process for no good reason now that we can await things and have readable code.

tools/tsp-client/src/gh.ts
```diff
@@ -0,0 +1,97 @@
...
+  
+export async function cloneRepo(rootUrl: string, cloneDir: string, repo: string): Promise<void> {
+    if (await doesFileExist(cloneDir)) {
+      Logger.debug(`Removing existing sparse-checkout directory ${cloneDir}`);
+      await removeDirectory(cloneDir);
+    }
+    await mkdir(cloneDir, { recursive: true });
+    Logger.debug(`Created temporary sparse-checkout directory ${cloneDir}`);
+    return new Promise((resolve, reject) => {
+      const git = spawn("git", ["clone", "--no-checkout", "--filter=tree:0", repo, cloneDir], {
```
xirzec, line 19: the dangerous thing about `spawn` and `shell: true` is this opens us up to injection attacks unless we can validate all passed parameters. Do we need it for calling git?

tools/tsp-client/src/gh.ts
```diff
@@ -0,0 +1,97 @@
...
+      });
+      git.once("error", (err) => {
+        reject(new Error(`git sparse-checkout failed with error: ${err}`));
+      });
+    });
+  }
+  
+  export async function addSpecFiles(cloneDir: string, subDir: string): Promise<void> {
+    return new Promise((resolve, reject) => {
+      const git = spawn("git", ["sparse-checkout", "add", subDir], {
```
xirzec, line 60: same risk for parameters here

tools/tsp-client/src/gh.ts
```diff
@@ -0,0 +1,97 @@
...
+      });
+      git.once("error", (err) => {
+        reject(new Error(`git sparse-checkout add failed with error: ${err}`));
+      });
+    });
+  }
+  
+  export async function checkoutCommit(cloneDir: string, commit: string): Promise<void> {
+    return new Promise((resolve, reject) => {
+      const git = spawn("git", ["checkout", commit], {
```
xirzec, line 80: and here depending on where commit is coming from

tools/tsp-client/src/gh.ts
```diff
@@ -0,0 +1,97 @@
+import { execSync, spawn } from "child_process";
```
xirzec, line 1: nit the file name should be `git.ts` I think since `gh` makes me think of github and this isn't really github specific

tools/tsp-client/src/index.ts
```diff
@@ -1,62 +1,217 @@
...
+import { cp, existsSync } from "node:fs";
+import { addSpecFiles, checkoutCommit, cloneRepo, getRepoRoot, sparseCheckout } from "./gh.js";
+import { doesFileExist, fetch } from "./network.js";
+import { parse as parseYaml } from "yaml";
 
-async function discoverMainFileUrl(outputDir: string): Promise<string> {
-  const tspUrl = await tryReadTspLocation(outputDir);
-  if (!tspUrl) {
-    throw new Error("No tsp-location.yaml found, and no main file specified");
+// async function discoverMainFileUrl(outputDir: string): Promise<string> {
```
xirzec, line 14: remove commented out code?

tools/tsp-client/src/typespec.ts
```diff
@@ -53,3 +54,34 @@ export async function compileTsp({
...
+  emitter,
+  emitterOptions,
+}: {
+  tempDir: string;
+  mainFilePath: string;
+  emitter: string;
+  emitterOptions: string;
+}): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const git = spawn("tsp", ["compile", mainFilePath, `--emit=${emitter}`, emitterOptions], {
```
xirzec, line 71: why are we shelling out to tsp instead of invoking the compiler directly?

another shell injection attack here as well

tools/tsp-client/src/options.ts
```diff
@@ -50,35 +64,53 @@ export async function getOptions(): Promise<Options> {
...
-  const outputDir = path.resolve(path.normalize(positionals[0]));
+
+  if (positionals[0] === "init") {
+    if (!values.tspConfig) {
+      Logger.error("tspConfig is required");
+      printUsage();
+      process.exit(1);
+    }
+  }
+  // By default, assume that the command is run from the output directory
```
xirzec, line 99: Assuming the cwd is the output directory seems.... not super reliable if someone is trying out the command from the repo root for example. Is there a reason we can't have this be explicit?

tools/tsp-client/src/index.ts
```diff
@@ -1,62 +1,217 @@
...
+    // repoUrl, // TODO
+  }: {
+    config: string;
+    outputDir: string;
+    // commit: string;
+    // repoUrl: string;
+  }): Promise<string> {
+  if (await doesFileExist(config)) {
+    // URL scenario
+    const tspConfigUrl = new URL(config).toString();
```
xirzec, line 83: what's the point of going through the URL constructor?

tools/tsp-client/src/index.ts
```diff
@@ -1,62 +1,217 @@
...
+    const matchRes = tspConfigUrl.match('^https://(?<urlRoot>github).com/(?<repo>[^/]*/azure-rest-api-specs(-pr)?)/blob/(?<commit>[0-9a-f]{40})/(?<path>.*)/tspconfig.yaml$')
```
xirzec, line 84: yuck, can we abstract this into a helper that parses the URL if we're going to go down this path?

tools/tsp-client/src/index.ts
```diff
@@ -1,62 +1,217 @@
...
+        if (matchRes.groups["urlRoot"]! === "github") {
+          var resolvedConfigUrl = tspConfigUrl.replace("github.com", "raw.githubusercontent.com");
+          resolvedConfigUrl = resolvedConfigUrl.replace("/blob/", "/");
+          Logger.debug(`Resolved config url: ${resolvedConfigUrl}`)
+          const tspConfig = await fetch(resolvedConfigUrl);
+          const configYaml = parseYaml(tspConfig);
+          if (configYaml["parameters"] && configYaml["parameters"]["service-dir"]){
+            const serviceDir = configYaml["parameters"]["service-dir"]["default"];
+            Logger.debug(`Service directory: ${serviceDir}`)
+            mkdir(path.join(outputDir, serviceDir), { recursive: true }).then(() => {
```
xirzec, line 96: missing await and again, don't use `then`

tools/tsp-client/src/gh.ts
xirzec: also we should cache this

tools/tsp-client/src/index.ts
```diff
@@ -1,62 +1,217 @@
...
+  await cloneRepo(tempRoot, cloneDir, `https://github.com/${repo}.git`);
+  await sparseCheckout(cloneDir);
+  await addSpecFiles(cloneDir, directory)
+  Logger.info(`Processing additional directories: ${additionalDirectories}`)
+  for (const dir of additionalDirectories) {
+    await addSpecFiles(cloneDir, dir);
+  }
+  await checkoutCommit(cloneDir, commit);
+  
+  cp(path.join(cloneDir, directory), srcDir, { recursive: true }, (err) => {
```
xirzec, line 151: can you use the promisified version of cp instead of the old callback version? This won't do what you think it will.

tools/tsp-client/src/index.ts
```diff
@@ -66,36 +221,55 @@ async function downloadAndCompile({
...
 }
 
+async function syncAndCompile({
+  outputDir,
+  noCleanup,
+}: {
+  outputDir: string;
+  noCleanup: boolean;
+}) {
+  syncTspFiles(outputDir).then(() => {
```
xirzec, line 231: again missing await and unwanted chaining

tools/tsp-client/src/index.ts
```diff
@@ -66,36 +221,55 @@ async function downloadAndCompile({
...
-    emitter: options.emitter,
-    noCleanup: options.noCleanup,
-    emitterOutputPath: options.outputDir,
-  });
+  switch (options.command) {
+      case "init":
+        if (options.tspConfig === undefined) {
+          throw new Error("tspConfig is undefined");
+        }
+        sdkInit({config: options.tspConfig, outputDir: rootUrl}).then((result) => {
```
xirzec, line 254: another `then()`

xirzec (commented):
Some highlevel feedback and style issues, I'll take another look at some of the logic though `index.ts` is getting a bit overrun, perhaps some of the logic around managing the local git clone can be pulled out into its own module?

tools/tsp-client/src/options.ts
catalinaperalta: We want to reduce the number of inputs that a service partner needs in order to use this tool. For service partners working from their project directory it makes it simpler, but that is why we still allow them to use the output directory if they're not in the service project directory

tools/tsp-client/src/fs.ts
catalinaperalta: From what I read in the instructions for the powershell scripts, `emitter-package.json` is part of the required set up in the language repos. I would expect it to always be there, but can clarify the name of this function more. Maybe `findEmitterPackageJson`? Or `findEmitterPackageConfig`? Not sure about whether a language repo would support multiple emitters, at least not initially, we could work on adding this later if that's a possible use case

catalinaperalta: Also, this was a workaround meanwhile I'm work on adding the custom language setting functionality that existed previously with the powershell scripts.

tools/tsp-client/src/gh.ts
catalinaperalta: removed

tools/tsp-client/src/options.ts
xirzec: yeah I'm more thinking, can we perhaps have a prompt before execution like `generating in <cwd here>, continue? [Y/n]` to prevent careless pollution of the tree.

Alternatively, I was thinking it would be somewhat trivial to check the path and look to see if we're in `<something>azure-sdk-for-<thing>/sdk/<someservice>/<somepackage>` and warn/error if not given an explicit path

I'll leave it up to you what seems most useful

tools/tsp-client/src/fs.ts
xirzec: I think what was confusing me is "findEmitterPackage" makes it sound like this is where the typespec emitter lives, rather than the repo config for which emitter package to use, so perhaps `getEmitterFromRepoConfig` or somesuch?

xirzec:
Also I was very flat by the time I got around to reviewing this yesterday, but thank you for contributing! I'm excited that we're getting closer to supplementing what the pwsh scripts are doing. 😄

allenjzhang:
Please also update [doc in rest-api-spec repo](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/typespec-rest-api-dev-process.md#4-generate-or-refresh-sdk-code-from-a-typespec-project) to use tsp-client command

tools/tsp-client/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "ts-node": "^10.9.1",
+    "typescript": "^5.1.6"
+  },
+  "dependencies": {
+    "@azure/core-rest-pipeline": "^1.12.0",
+    "chalk": "^5.3.0",
+    "yaml": "^2.3.1"
+  },
+  "peerDependencies": {
+    "@typespec/compiler": "^0.47.1"
```
mikeharder, line 41:
```suggestion
    "@typespec/compiler": ">=0.47.1 <1.0.0"
```

I recommend using this version range, which is equivalent to `^0.47.1` but also allows pre-release versions like `0.48.0-dev.18`.  This is useful for testing dev builds of TypeSpec packages.

tools/tsp-client/src/options.ts
```diff
@@ -0,0 +1,152 @@
...
+  command: string;
+  tspConfig?: string;
+  emitter?: string;
+  mainFile?: string;
+  outputDir: string;
+  noCleanup: boolean;
+  skipSyncAndGenerate: boolean;
+  commit?: string;
+  repo?: string;
+  isUrl: boolean;
```
raych1, line 18: What is this option `isUrl` used for?

tools/tsp-client/src/options.ts
```diff
@@ -0,0 +1,152 @@
...
+      Logger.error("tspConfig is required");
+      printUsage();
+      process.exit(1);
+    }
+    if (await doesFileExist(values.tspConfig)) {
+      isUrl = true;
+    }
+    if (!isUrl) {
+      if (values.commit === undefined || values.repo === undefined) {
+        Logger.error("The commit and repo options are required when tspConfig is a local directory");
```
raych1, line 124: I think we should support the local run scenario using local specs without commit and repo in the option. They can be optional options.

tools/tsp-client/src/index.ts
```diff
@@ -0,0 +1,266 @@
...
+  }): Promise<string> {
+  if (isUrl) {
+    // URL scenario
+    const resolvedConfigUrl = await resolveTspConfigUrl(config);
+    Logger.debug(`Resolved config url: ${resolvedConfigUrl.resolvedUrl}`)
+    const tspConfig = await fetch(resolvedConfigUrl.resolvedUrl);
+    const configYaml = parseYaml(tspConfig);
+    if (configYaml["parameters"] && configYaml["parameters"]["service-dir"]){
+      const serviceDir = configYaml["parameters"]["service-dir"]["default"];
+      Logger.debug(`Service directory: ${serviceDir}`)
```
raych1, line 38: Can you throw the exception if `serviceDir` doesn't exist?

tools/tsp-client/src/options.ts
catalinaperalta: Tracking issue for this: https://github.com/Azure/azure-sdk-tools/issues/7010. If we do this, then we cant create a valid tsp-location.yaml file so the init would be "incomplete" in that sense. I'll check on this with the issue I linked above

tools/tsp-client/src/options.ts
catalinaperalta: This is related to your comment below about requiring the commit and repo for local specs. On lines 112-129, we check if there tspconfig.yaml is in the local file system or if it is a URL. If it's a local file we need a commit sha and the repo url in order to create a valid tsp-location.yaml file. I'll use the same issue linked below to track this and change it, currently I'm just targeting matching the original Process script.

PR #6917, opened 2023-09-07
[stress] Add maintenance window and enable node security upgrades

benbp:
This enables AKS node upgrades for security patches in stress clusters, and specifies a time window (early Mondays) for them to happen.

### DIFF
--- a/tools/stress-cluster/cluster/azure/cluster/cluster.bicep
+++ b/tools/stress-cluster/cluster/azure/cluster/cluster.bicep
@@ -53,7 +53,7 @@ var agentPools = [
     defaultAgentPool
 ]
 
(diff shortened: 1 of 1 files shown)


richardpark-msft:
@benbp - is there any way to get Kubernetes to show that the node was going down for maintenance in the job/pod termination status?

richardpark-msft:
(or even something programmatic we can pick up in our dashboard)?

benbp:
@richardpark-msft I think we can correlate with VMSS events in the logs, but it's cumbersome. Regardless we should have a dashboard widget for this, but it's much harder to correlate with a specific pod. I'll dig in on some of the queries I had the last time we investigated some of these issues.

richardpark-msft:
Filed https://github.com/Azure/azure-sdk-tools/issues/6921 to track work to show status for this, if needed.

PR #6927, opened 2023-09-08
Add Kusto function definition of SpecsPipelinesBuildsJobLogs.kql

konrad-jamrozik:
See the comment in the added function for details.

I followed the guidance I wrote down earlier here:
https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/625/Kusto-Query-Catalog?anchor=adding-functions-to-%60azsdkengsys%60-cluster


weshaggard (approved):
I assume you are just backing up this here which is fine

konrad-jamrozik:
> I assume you are just backing up this here which is fine

Exactly!

PR #6931, opened 2023-09-08
Fixed several ApiView parser issues.

LarryOsterman:
- Fixed issue creating ApiView for azure core
- Added output for override and final attributes
- Added diagnostics for the following cases:
  - [destructors that aren't virtual](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c35-a-base-class-destructor-should-be-either-public-and-virtual-or-protected-and-non-virtual).
  - method overrides without the override keyword.

### DIFF
--- a/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
+++ b/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
@@ -1487,7 +1487,7 @@ public:
 
     if (templateParam->hasDefaultArgument())
     {
-      auto const&defaultArg = templateParam->getDefaultArgument().getArgument();
+      auto const& defaultArg = templateParam->getDefaultArgument().getArgument();
       switch (defaultArg.getKind())
       {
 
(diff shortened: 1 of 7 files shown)


tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -1826,6 +1830,55 @@ class AstMethod : public AstFunction {
       : AstFunction(method, database, parentNode), m_isVirtual(method->isVirtual()),
         m_isPure(method->isPure()), m_isConst(method->isConst())
   {
+    // We assume that this is an implicit override if there are overriden methods. If we later find
+    // an override attribute, we know it's not an implicit override.
+    //
+    // Note that we don't do this for destructors, because they don't have an override attribute.
```
antkmsft, line 1836: Hmm, they do...

```cpp
struct B {
    virtual ~B() {}
};

struct D : B {
    ~D() override {}
};
```

Unless I misunderstood some nuance of what you meant in that comment.

antkmsft (approved):

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
LarryOsterman: The comment is wrong. But I don't think we want to suggest that people add `override` to their dtors anyway. The point of adding the diagnostic is to ensure that if the base method is changed, all overrides are changed (if a method has the same name and signature as a method in the parent but the parent method is not virtual, that's not an override, it's just a method with the same name and signature). 

But the signature of a dtor is fixed, so there's no need to do check for that.

PR #6933, opened 2023-09-11
tox-monorepo: Refer people to --root for tox 4 instead of tox-monorepo

koffie:
Since azure-sdk-tools itself doesn't use tox-monorepo anymore and instead decided to start using the new `tox --root` command line arguement, we should also point people possibly intersted in the tox-monorepo plugin to this option.


koffie:
Could this still be merged?

Issue #5945, opened 2023-04-10
Implement new CODEOWNERS syntax for labelling workflow


konrad-jamrozik:
As in the email thread with subject `RE: (merged threads) Re: Question about the ServiceLabel tags in CODEOWNERS files + Re: Inquiry:  SDK team owners data source`

Relevant snippets:

```
# ServiceLabel: %Event Hubs 
#/<NotInRepo>/            @jsquire, @someother
 
# ServiceLabel: %Event Hubs %Service Attention
#/<NotInRepo>/            @serkant, @james
 
# ServiceLabel: %Event Hubs
/sdk/eventhub/             @jsquire, @somearchitect, @serkant, @otherservice
 
NEW SYNTAX:
# ServiceLabel: Event Hubs
# Service owners:   @serkant @james
# Azure SDK owners: @jsqure @someother
# PR reviewers:
/sdk/eventhubs/     @jsquire @somearchitect @serkant @otherservice
``` 
 
```
# ServiceLabel: %Event Hubs %Service Attention
#/<NotInRepo>/            @serkant, @james
 
# ServiceLabel: %Event Hubs
/sdk/eventhub/             @jsquire, @someperson
 
NEW SYNTAX:
# ServiceLabel: Event Hubs
# Service owners:   @serkant @james
# Azure SDK owners & PR reviewers:
/sdk/eventhubs/     @jsquire @someperson
```

```  
# ServiceLabel: %Event Hubs %Service Attention
#/<NotInRepo>/            @Azure/eh-engineers
 
# ServiceLabel: %Event Hubs %Service Attention
/sdk/eventhub/             @serkant, @james
 
NEW SYNTAX:
# ServiceLabel: Event Hubs
# Service owners:   @Azure/eh-engineers
# PR reviewers:
/sdk/eventhubs/     @serkant @james
```
  
```  
# ServiceLabel: %Event Hubs %Service Attention
/sdk/eventhub/           @serkant, @james
 
NEW SYNTAX:
# ServiceLabel: Event Hubs
# Service owners & PR reviewers:
/sdk/eventhubs/     @serkant @james
```

JimSuplizio:
I'm recommending a few changes to this.

1. There's a missing, existing, moniker in the suggested changes and that's **PRLabel**. We need to keep that. The **PRLabel** and the **ServiceLabel** don't have to be the same and they serve different purposes.
[...]

PR #6952, opened 2023-09-12
Add Codeowners linter tool

JimSuplizio:
This is the initial PR for the Azure.Sdk.Tools.CodeownersLinter tool. 
What this tool does:
This tool will lint a CODEOWNERS file including our specific brand of Metadata. Note: The linter can handle the [new metadata tags/format](https://github.com/Azure/azure-sdk-tools/issues/5945) as well as the existing metadata. Our CODEOWNERS files contain more than source path/owner information, we also have metadata tags that are used in various workflows. All of these tags require more than one line in the CODEOWNERS file and these are referred to as **blocks**. _For example, a **PRLabel** entry needs to be part of a block that ends in a source path/owner line.
Note: Changes to the existing CodeownersParser are not functional outside of FileHelpers.LoadCodeownersFileAsStringList. The other changes were to declare monikers and separators as constants which are being used by the linter.

**What changes will be required for every owner?**

- Owners, whether they're users or teams, need to have write permissions if they're going to be in a CODEOWNERS file. For teams this means that the team needs to be under the azure-sdk-write team. For owners this means they'll need to be under azure-sdk-write or in a team under azure-sdk-write. This is already true for any owner on a source path/owner line in CODEOWNERS but it also needs to be true for the metadata entries for our workflows.
- Owners, specifically users, need to be **public** members of the Azure org. This was already true and the [instructions for requesting Azure repository access](https://eng.ms/docs/products/azure-developer-experience/onboard/access) clearly state this. The reason is that some workflows, like GitHub Actions processing, check Azure org membership and are unable to see Private members. This results in workflow processing that thinks a user isn't a member of Azure and is why some people file an issue and it gets labeled as Customer Reported even though they're an Azure org member.

**What is verified during linting?**

- **Owners**
    - Users and Teams are verified to have write permissions. (see Owners above)
    - Users are also verified to be Public members of Azure.
- **Labels**
    - Whether or not the label exists in a particular repository.
- **Source paths/Globs** (not something GitHub verifies)
    - Does the path (directory or file) exist?
    - Does the glob contain [forbidden characters](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax)?
    - Does the glob match anything in the repository?
- **Block formatting** - There are several existing metadata tags and a couple more getting added or changed. Each tag is associated with an owner or label and those pieces are verified in their respective verifications. The block formatting verification ensures that things within a block are correct. Examples would be: 
    - A PRLabel needs to be part of a block that ends in a source path/owner line 
    - A ServiceLabel needs to be part of a block that ends in a source path/owner line OR needs to be immediately followed by a #/&lt;NotInRepo&gt;/ line with declared owners. 

**Where does the owner/label data come from?**
The team/user, org visibility and repository label data are pre-populated by the [github-team-user-store](https://github.com/Azure/azure-sdk-tools/tree/main/tools/github-team-user-store) which runs daily as part of the pipeline-owners-extractor and is stored in blob data that workflows can use. This had to be done this way because of the specific permissions required to get the information from GitHub. This also means that some types of fixes, such as user changing their Azure org membership from private to public, won't be immediately picked up as the pipeline-owners-extractor pipeline would need to update the data in storage first.
 
**Where will this run?**
The goal here is to run this for at least the azure-sdk-for-&lt;language&gt; public and private repositories as both a nightly pipeline as well as a CI pipeline when CODEOWNERS is changed. Where and when we want to run this is the reason the data had to be in blob storage as we can't have tokens in public pipelines.

**What about existing errors?**
Even in [repositories whose CODEOWNERS files are valid](https://github.com/Azure/azure-sdk-for-go/blob/main/.github/CODEOWNERS), according to GitHub, have errors. Because of this, repositories where this will be enabled will have a CODEOWNERS_baseline_errors.txt file which contains a deduped list of every unique error. Output from tool will filter out the baseline errors which will at least prevent new errors from being added to CODEOWNERS. As baseline issues are fixed, it's expected they'll be removed from the baseline file. The linter is capable of generating the baseline file as well as using the baseline file to filter out errors.

**Known limitations**

- As previously mentioned, the verification data for team/user, owner org visibility and repo/label data must be generated using a GitHub token with specific permissions on a private pipeline that runs daily. This means that fixing some types of issues will have a delay before their data in Azure storage is updated.
- The baseline file means that an existing error can be added to a new line and wouldn't get flagged in CI run. Yes, this is an unfortunate side effect of the baseline file however if this wasn't done, the nightly pipelines running the linter would start out failing and probably never pass. Similarly, if the pipeline runs as part of the PR that has CODEOWNERS changes and there wasn't a baseline then I'll just say hilarity would ensue.

@weshaggard - As per our conversation I'm creating the PR as a draft PR today. The work left to do, outside of any feedback, is mostly around finishing up block formatting error verification, CodeownersFormat.ProcessCodeownersFile tests, baseline file generation and filtering and the README.md for the tool. Though, with this PR description, I've got a good start on the README.


tools/codeowners-linter/ci.yml
```diff
@@ -0,0 +1,26 @@
+# NOTE: Please refer to https://aka.ms/azsdk/engsys/ci-yaml before editing this file.
```
weshaggard, line 1: Why do we need a new ci? Is there a reason we don't include this under the existing codeowners tools directory? They seem like they are connected enough to be together.

tools/codeowners-linter/Azure.Sdk.Tools.CodeownersLinter/Azure.Sdk.Tools.CodeownersLinter.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
weshaggard, line 1: As we talked at some point we should merge codeowners parser and this linter so that we don't have this logic duplicated as we shouldn't need both.

tools/codeowners-linter/ci.yml
JimSuplizio: @weshaggard because this is a separate tool that's being built and uploaded to the dev feed. The codeowners ci.yml isn't for the codeowners library but for the retrieve-codeowners tool. I'd like to keep things separate.

weshaggard: OK once we combine the libraries we should probably put all the codeowners tools under the same directory and have the one pipeline. It is OK if we release more than one library or tool from the same pipeline.

weshaggard (approved):
Left a few general comments. I didn't deep dive into the code but did scan through it and nothing stood out to me. Lets get take this to the next steps of getting a pipeline to run it and also work on merging the parser and linter.

JimSuplizio:
Closing this PR. This one was linter only and the new [PR](https://github.com/Azure/azure-sdk-tools/pull/7097) rewrites things which are now Utils that can both parse and lint which will be replacing the existing parser.

PR #6966, opened 2023-09-13
Skip cleaning up child resource groups

benbp:


eng/scripts/live-test-resource-cleanup.ps1
```diff
@@ -309,6 +309,14 @@ function HasDoNotDeleteTag([object]$ResourceGroup) {
   return $doNotDelete -ne $null
 }
 
+function IsChildResource([object]$ResourceGroup) {
+  if ($ResourceGroup.ManagedBy) {
```
weshaggard, line 313: For my context how does this work? I didn't know there was a notion of child resource groups.

benbp: It was something the script hit on the latest pipeline run. In this case it looks like the databricks service creates an inner resource group with the service as parent, and AKS creates a resource group for the VMSS resource, with the AKS resource as parent. In the past I've handled this via prefixes in the allowlist file, which is still necessary for some resources like synapse which don't have ManagedBy set.

PR #6971, opened 2023-09-14
update the document for semicolon in `TypeSpec-Project-Generate.ps1`

ArcturusZhang:
Despite the document says we could use semicolon to separate multiple options in this parameter, but it will not work if we just do this:
```
dotnet build /t:GenerateCode /p:typespecAdditionalOptions="csharpGeneratorPath=./autorest.csharp/artifacts/bin/AutoRest.CSharp/Debug/net6.0/AutoRest.CSharp.dll;debug=true"
```
which eventually calls the `TypeSpec-Project-Generate.ps1` script.

The option after the semicolon is somehow discarded. But if we escape this semicolon, everything works fine
```
dotnet build /t:GenerateCode /p:typespecAdditionalOptions="csharpGeneratorPath=./autorest.csharp/artifacts/bin/AutoRest.CSharp/Debug/net6.0/AutoRest.CSharp.dll%3Bdebug=true"
```

Fixes https://github.com/Azure/autorest.csharp/issues/3362

### DIFF
--- a/eng/common/scripts/TypeSpec-Project-Generate.ps1
+++ b/eng/common/scripts/TypeSpec-Project-Generate.ps1
@@ -5,7 +5,7 @@ param (
     [Parameter(Position=0)]
     [ValidateNotNullOrEmpty()]
     [string] $ProjectDirectory,
-    [string] $TypespecAdditionalOptions = $null, ## addtional typespec emitter options, separated by semicolon if more than one, e.g. option1=value1;option2=value2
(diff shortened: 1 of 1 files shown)


ArcturusZhang:
@m-nash @benbp @weshaggard  could you also take a look?

I wonder if we have other options or other places to fix to let this work as is instead of escaping this character.

ArcturusZhang:
Close this PR since the issue only happens when we are using `msbuild`
I will open another PR to update document somewhere else

PR #6975, opened 2023-09-14
[stress] Add parallel pod template to stress-test-addons

benbp:
This PR adds built-in support for pod parallelism within a job (https://kubernetes.io/docs/concepts/workloads/controllers/job/#completion-mode). 

See the [README update](https://github.com/Azure/azure-sdk-tools/blob/0659ed265ce4e896d04ee4408a945d76e133337c/tools/stress-cluster/chaos/README.md#run-multiple-pods-in-parallel-within-a-test-job) for details.


tools/stress-cluster/chaos/examples/parallel-pod-example/templates/parallel-pod.yaml
```diff
@@ -0,0 +1,18 @@
+{{- include "stress-test-addons.parallel-deploy-job-template.from-pod" (list . "stress.parallel-pod-example" 3) -}}
```
ckairen, line 1: a comment here saying 3 means 3 pods in parallel might be helpful for users

benbp: @ckairen fixed

ckairen: nvm, the doc mentioned it, but I think keeping the doc and example consistrnt might be good. so like 3 or 5 but @ both places

benbp: I'll change both to three.

ckairen (approved):

PR #7002, opened 2023-09-21
[stress] Add LockDeletionForDays parameter to set PodDisruptionBudget and cleanup job

benbp:
### Locking

This adds a parameter `LockDeletionForDays` to the stress deploy script, which will prevent a test and all its pods from being killed for a time period by the kubernetes api server. Possible reasons a pod may be killed:

1. Node rebalancing/scale down
2. Node security patch updates
3. Misc. node evictions and scheduler based pod deletions

It also creates a cron job resource set to delete the pod disruption budget after `LockDeletionForDays` days, so that evictions can then occur (meaning the test pods won't necessarily be killed, but they will no longer be locked).

### Parallel

This updates the parallel jobs feature to use the matrix config instead of a separate addons template. This is both more intuitive and allows for setting different parallel values per scenario (rather than per test).

### DIFF
--- a/tools/stress-cluster/chaos/README.md
+++ b/tools/stress-cluster/chaos/README.md
@@ -6,6 +6,7 @@ The chaos environment is an AKS cluster (Azure Kubernetes Service) with several 
 
   * [Installation](#installation)
   * [Deploying a Stress Test](#deploying-a-stress-test)
+     * [Locking a test to run for a minimum number of days](#locking-a-test-to-run-for-a-minimum-number-of-days)
   * [Creating a Stress Test](#creating-a-stress-test)
      * [Layout](#layout)
      * [Stress Test Metadata](#stress-test-metadata)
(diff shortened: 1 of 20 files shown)


tools/stress-cluster/cluster/kubernetes/stress-test-addons/templates/_pod_disruption_budget.tpl
```diff
@@ -0,0 +1,75 @@
...
+metadata:
+  name: disruption-budget-{{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    release: {{ .Release.Name }}
+spec:
+  # Jobs do not implement `scale` otherwise we could set `minAvailable: 100%` or `maxUnavailable: 0` instead.
+  # Work around this by setting `minAvailable` to a number that will never be reached to simulate 100%
+  # so that the disruption budget will work in parallel pod scenarios (completionMode: indexed)
+  minAvailable: 10000
```
ckairen, line 14: this also prevents watcher from deleting the pod right?

benbp: No, direct pod deletions are still allowed, it's node evictions that are prevented.

tools/stress-cluster/chaos/README.md
```diff
@@ -113,6 +113,27 @@ you can quick check the local logs:
 kubectl logs -n <stress test namespace> <stress test pod name>
 ```
 
+### Locking a test to run for a minimum number of days
```
ckairen, line 116: maybe add to table of contents?

benbp: Fixed

ckairen (approved):

PR #7009, opened 2023-09-25
Additional Documentation for `Matcher` and `Transform` concepts

scbedd:
Just a readme update. @samvaity for FYI.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
@@ -15,6 +15,8 @@ to documentation in your specific language repository in order to configure reco
 ## Table of contents
 
 - [Azure SDK Tools Test Proxy](#azure-sdk-tools-test-proxy)
+      - [Test documentation by language:](#test-documentation-by-language)
+  - [Table of contents](#table-of-contents)
   - [Installation](#installation)
     - [Via Local Compile or .NET](#via-local-compile-or-net)
     - [Via Docker Image](#via-docker-image)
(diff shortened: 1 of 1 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
```diff
@@ -582,12 +592,70 @@ In some cases, users need to register a lot (10+) of sanitizers. In this case, g
 ]
 ```
 
+### Set a Matcher
+
+Setting a matcher is just like adding a `sanitizer`. Set the `x-abstraction-identifier` value to the name of the matcher you want to instantiate, and provide the proper constructor arguments in the body! Check `Info/Available/` for available matchers.
```
samvaity, line 597: when adding a sanitizer. The latter set sanitizer overrides the one sent previously though, right?

samvaity (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
scbedd: Nope. Sanitizers can be stacked.

scbedd: A `matcher` needs to take a unit of input, then compare it to all the possible entries in a set.

It does this by walking each possible entry, and comparing URI + Body + Headers. As there are identical values found, an integer is incremented. At the end, the "most matching" entry that has not been disqualified is popped off the stack. (We can't present the same response twice this way).

Knowing that, it becomes quite difficult to figure out how one would combine a matcher.

A `sanitizer` on the other hand, merely run against two targets:

1. Every entry in the recording, prior to saving to disk.
2. Before matching a request to a recording.

We can handle multiple sanitizers by simply running them one after another.

PR #7013, opened 2023-09-26
[stress] Change prod maintenance day. Fix stress watcher deploy

benbp:


tools/stress-cluster/cluster/kubernetes/stress-infrastructure/templates/stresswatcher.yaml
```diff
@@ -34,3 +34,5 @@ spec:
         {{- include "stress-test-addons.env-volumes" $ctx | nindent 8 }}
         # Volume template for mounting azure file share for debugging
         {{- include "stress-test-addons.debug-file-volumes" $ctx | nindent 8 }}
+
+{{- include "stress-test-addons.static-secrets" . }}
```
ckairen, line 38: why are we adding this to watcher?

benbp: @ckairen It's necessary for mounting the credentials to access our subscription. It used to be included automatically when it was a yaml file, but a while back I changed it to a `tpl` file which meant it no longer got included by default in the helm release.

ckairen (approved):

PR #7033, opened 2023-09-29
[JimBot] Remove CXP Attention rule

jsquire:
# Summary

The focus of these changes is to remove the rule attached to the "CXP Attention" label that has been removed from the Azure SDK repositories.  Also riding along is a tweak to the "needs-team-triage" criteria to avoid applying it when an issue has already been assigned.  Minor errors in the rule for fully automated initial triage have also been corrected.


tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/EventProcessing/IssueProcessing.cs
```diff
@@ -188,44 +187,13 @@ public static void ServiceAttention(GitHubEventClient gitHubEventClient, IssueEv
...
         /// <summary>
         /// Manual Triage After External Assignment
         /// Trigger: issue unlabeled
         /// Conditions: Issue is open
+        ///             Issue is unassigned
         ///             Has "customer-reported" label
-        ///             Label removed is "Service Attention" OR "CXP Attention"
-        ///             Issue does not have "Service Attention" OR "CXP Attention"
+        ///             Label removed is "Service Attention"
         ///             (in other words if both labels are on the issue and one is removed, this
```
JimSuplizio, line 197: The comment "(in other words..." is no longer valid and should be removed.

JimSuplizio (approved):
Please remove the CXP Attention from the constants. I don't believe you need wait for this to go in, if a rule isn't in the config it defaults to false and prints a warning out but otherwise does nothing.

PR #7068, opened 2023-10-05
[githubio-linkcheck] Reduce timeout from infinite to 6 hours

mikeharder:
- Pipeline normally runs in 3 hours, so 6 hours is plenty of time
- We should not use infinite timeouts, to avoid blocking build agents indefinitely

### DIFF
--- a/eng/pipelines/githubio-linkcheck.yml
+++ b/eng/pipelines/githubio-linkcheck.yml
@@ -11,7 +11,7 @@ variables:
 jobs:
 - job: CheckLinks
   displayName: Check and Cache Links
-  timeoutInMinutes: 0
+  timeoutInMinutes: 360
   steps:
   - task: PowerShell@2
     displayName: 'azure-sdk link check'


benbp (approved):

weshaggard:
Did this cause some particular issue or just something you noticed? Setting this to 0 lets the host agent decide how long to run see https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts. Not sure if 1ES agents have a fallback timeout.

mikeharder:
> Did this cause some particular issue or just something you noticed? Setting this to 0 lets the host agent decide how long to run see https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts. Not sure if 1ES agents have a fallback timeout.

Since we are using self-hosted agents, I believe setting `timeoutInMinutes: 0` can cause the agents to block indefinitely.  There might be another timeout eventually, like 48 hours or something.  But I'd much rather have us use some large value like 6 hours or even 12 hours rather than "indefinite".

Here's a hung build I believe was blocking an agent for almost 48 hours until I manually cancelled it:

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3129547&view=logs&j=a460d732-23aa-5493-a411-cc406067acf8&t=71b47ba2-822c-5669-3297-76bbc5d002b7&l=35

weshaggard:
Sounds good. I was just curious of what triggered your interest in updating these.

PR #7069, opened 2023-10-05
[live-test-cleanup] Reduce timeout from infinite to default (60 minutes)

mikeharder:
- Pipeline normally runs in 15 minutes, so 60 minutes is plenty of time
- We should not use infinite timeouts, to avoid blocking build agents indefinitely

### DIFF
--- a/eng/pipelines/live-test-cleanup.yml
+++ b/eng/pipelines/live-test-cleanup.yml
@@ -62,7 +62,6 @@ stages:
 
   jobs:
   - job: Run
-    timeoutInMinutes: 0
     pool:
       name: azsdk-pool-mms-ubuntu-2204-general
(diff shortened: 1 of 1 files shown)


benbp (approved):

benbp:
I believe in the past we were timing out on large back to back purge operations. But as long as the queue is kept healthy we shouldn't need that (and can always increase timeouts temporarily).

mikeharder:
> I believe in the past we were timing out on large back to back purge operations. But as long as the queue is kept healthy we shouldn't need that (and can always increase timeouts temporarily).

I would even be fine with increasing the timeout permanently to something reasonable like 4 hours or however long you think we might need.  Just not infinite.

PR #7075, opened 2023-10-06
Set product as parent of package work item.

praveenkuttappan:
We made a change recently to create release plan work item as parent of package work item and set release plan as child of product work item. This PR is to revert some of those changes and set package work item as the children of product work item because each package work item it created for a major-minor version of a package, but release plan will be created for each individual releases including bug fixes. Devops allows setting only one work item as parent so a package work item cannot have more than one release plan as parent.


maririos:
> but release plan will be created for each individual releases including bug fixes.

Is this PR adding the related relatioship between Release Plan and Package work items?

benbp (approved):

PR #7086, opened 2023-10-09
Introduce new InternalsVisibleToAnalyzer analyzer (ACZ0112) with `FriendAttribute` concept

christothes:
# Summary
In scenarios where it is necessary to expose internal between client libraries via the `InternalVisibleToAttribute`, we want a more robust mechanism to ensure that we can be explicit about which APIs are intended to be consumed externally. 

This new analyzer validates that when a client library references an internal type, method, or property from another assembly that it is decorated with a `FriendAttribute`. 

There is a related API View PR (#7116) to treat types, methods, and properties decorated with the `FriendAttribute` as public API surface, so that changes can be raised as breaking.


KrzysztofCwalina:
@stephentoub, @terrajobst, is that something you'd like to provide in the BCL?

christothes:
> This analyzer wouldn't cause issues with the many test projects referencing internal members, would it?

No, because Test projects don't reference this analyzer by default ([link](https://github.com/Azure/azure-sdk-for-net/blob/d9948ec7486522c984de37228e1d6e33209c01b6/eng/Directory.Build.Common.props#L80))

heaths:
Maybe it's fine as-is, but would we also want to check for a namespace to avoid any other `FriendAttribute`s? What does .NET do in similar cases of look-alike attributes e.g., nullable attributes? IMO, we should follow .NET's practice.

pallavit:
Do we want to test against inheritance scenarios?

pallavit:
Also curious what are the scenarios where we see this being used? Is it a temporary feature we want to light-up or is this a more longer term dependency? We restricted the usage of InternalsVisibleToAttribute in .NET Framework so curious why do we want to make it more acceptable here?

christothes:
> Do we want to test against inheritance scenarios?

I think we have that covered for [interfaces](https://github.com/Azure/azure-sdk-tools/pull/7086/files#diff-ba50db1e90a94166cfe87d19c81888b56d68041ed5913ba287c9899c91a7889bR14) and [base classes](https://github.com/Azure/azure-sdk-tools/pull/7086/files#diff-ba50db1e90a94166cfe87d19c81888b56d68041ed5913ba287c9899c91a7889bR46).

christothes:
> Also curious what are the scenarios where we see this being used? Is it a temporary feature we want to light-up or is this a more longer term dependency? We restricted the usage of InternalsVisibleToAttribute in .NET Framework so curious why do we want to make it more acceptable here?

Azure.Identity will make use of this. I'm not sure if there are others that might in the future. In that case it was a design tradeoff we felt was the best choice given the alternatives.

pallavit:
> Azure.Identity will make use of this. I'm not sure if there are others that might in the future. In that case it was a design tradeoff we felt was the best choice given the alternatives.

Should we consider restricting it to Azure.Identity for now? Putting this in APIView ensures we scan what we are taking dependency on - which is great! However how do we ensure that the internal type will continue to be backward compatible and follow the same principles like any public type would do?

christothes:
> > Azure.Identity will make use of this. I'm not sure if there are others that might in the future. In that case it was a design tradeoff we felt was the best choice given the alternatives.
> 
> Should we consider restricting it to Azure.Identity for now? Putting this in APIView ensures we scan what we are taking dependency on - which is great! However how do we ensure that the internal type will continue to be backward compatible and follow the same principles like any public type would do?

I don't think we need to restrict it. The PR and API review processes should prevent mass adoption of this pattern. Once this is exposed in API View, any change to a decorated internal type, property, method, etc. will be flagged for review as with any other public change.

stephentoub:
> @stephentoub, @terrajobst, is that something you'd like to provide in the BCL?

This looks very familiar from a few years back ;-)

I don't fundamentally have a problem with it, though if we were to pursue it I'd want to run it through the C# / LDM, as InternalsVisibleTo is in part a compiler feature, and this is really just an extension of it.  I'd also want to name it something more aligned with InternalsVisibleTo, like VisibleTo, or InternallyVisibleTo, or something along those lines.  I'd leave it up to Immo if it's something he wants to pursue.

terrajobst:
@stephentoub @KrzysztofCwalina 

> I'd leave it up to Immo if it's something he wants to pursue.

I asked on Twitter to gauge interest. I think interest would be limited, like `private protected`, but that doesn't mean the C# compiler folks wouldn't care. They shipped a similar feature in Roslyn itself to mark their interfaces as closed, i.e. only failing people for trying to implement them.

eng/Directory.Build.targets
```diff
@@ -12,7 +12,8 @@
     <IsTestProject Condition="'$(IsTestHelperLibrary)' == 'true'">false</IsTestProject>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(SignAssembly)' != 'false'">
+<!-- Exclude TestReferenceWithInternalsVisibleTo from strong name signing, as it is required for a test scenario -->
```
heaths, line 15: Nit: 
```suggestion
  <!-- Exclude TestReferenceWithInternalsVisibleTo from strong name signing, as it is required for a test scenario -->
```

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/Descriptors.cs
```diff
@@ -172,92 +172,99 @@ internal class Descriptors
...
+        public static DiagnosticDescriptor AZC0111 = new DiagnosticDescriptor(
+            nameof(AZC0111),
+            "DO NOT use EnsureCompleted in possibly asynchronous scope.",
+            "Asynchronous method with `async` parameter can be called from both synchronous and asynchronous scopes. 'EnsureCompleted' extension method can be safely used on in guaranteed synchronous scope (i.e. `if (!async) {...}`).",
+            DiagnosticCategory.Usage,
+            DiagnosticSeverity.Warning, true);
+
+        public static DiagnosticDescriptor AZC0112 = new DiagnosticDescriptor(
+            "AZC0112",
+            "Misuse of internal type via InternalsVisibleTo.",
```
heaths, line 264: Technically,
```suggestion
            "Misuse of internal type via InternalsVisibleToAttribute.",
```

Or the attribute syntax like you did in the line below. I'd keep it consistent. But a nit, really.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/InternalsVisibleToAnalyzer.cs
```diff
@@ -0,0 +1,183 @@
...
+                AnalysisParameter(parameterSymbol, context);
+            }
+        }
+
+        private static void AnalysisParameter(IParameterSymbol symbol, SymbolAnalysisContext context)
+        {
+            var parentSymbol = context.Symbol;
+            if (symbol.Type.IsVisibleInternalWithoutFriendAttribute(context.Compilation))
+            {
+                // Check if the symbol is decorated with an attribute named FooAttribute
```
heaths, line 58: Nit: Seems, if necessary even, this should be on line 56.

src/dotnet/Azure.ClientSdk.Analyzers/TestReferenceWithInternalsVisibleTo/FriendAttribute.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+internal class FriendAttribute : Attribute
+{
+    public FriendAttribute(string friendAssembly)
```
heaths, line 8: I take it we may consider validating `friendAssembly` in the future? If so, we may want to add a property. We can analyze it either way, but less code to check it's property later and code won't have to worry about whether it exists or not since it always will have existed.

heaths (approved):

PR #7091, opened 2023-10-09
Emit the pushed commit sha as an output variable

hallipr:
Allow the devops pipeline to use the published commit id via output variable

git-branch-push.ps1 is variable on the output varable name

git-push-changes.yml is variable on the step name with a consistent output variable name 'CommitId'


weshaggard (changes requested):
As we talked I don't think we need this.

hallipr:
This ended up being not needed.  Closing PR

PR #7097, opened 2023-10-10
Add CodeownersUtils, a CodeownersParser replacement

JimSuplizio:
This is the initial PR for CodeownersUtils, the CodeownersParser replacement.  
**_Edit: I did create an [internal pipeline](https://dev.azure.com/azure-sdk/internal/_build?definitionId=6461&_a=summary) against the refs/merge of this PR. Obviously, I'll have to flip the yml location when this is checked in. Also, I canceled the release portion of the pipeline which would have run since I'd had to manually kick it._**

The parser should behave the same as the existing parser in regard to existing CODEOWNERS files. It will also be able to handle the new CODEOWNERS [syntax](https://github.com/Azure/azure-sdk-tools/issues/5945) both fully and in hybrid for (combination of both old and new). The big difference with the parser is that there is no longer a single ubiquitous list of users. We already had source owners (the users at the end of a source path/owner line) and ServiceOwners but there's now an additional piece of metadata, AzureSdkOwners. These lists are distinct and need to be separate as multiple lists can exist on a single CodeownersEntry.

The other thing this PR adds is Azure.Sdk.Tools.CodeownersLinter tool. This is the issue tracking the [linting](https://github.com/Azure/azure-sdk-tools/issues/4859)

What this tool does:
This tool will lint a CODEOWNERS file including our specific brand of Metadata. Note: The linter can handle the [new metadata tags/format](https://github.com/Azure/azure-sdk-tools/issues/5945) as well as the existing metadata. Our CODEOWNERS files contain more than source path/owner information, we also have metadata tags that are used in various workflows. All of these tags require more than one line in the CODEOWNERS file and these are referred to as **blocks**. _For example, a **PRLabel** entry needs to be part of a block that ends in a source path/owner line.

**What changes will be required for every owner?**

- Owners, whether they're users or teams, need to have write permissions if they're going to be in a CODEOWNERS file. For teams this means that the team needs to be under the azure-sdk-write team. For owners this means they'll need to be under azure-sdk-write or in a team under azure-sdk-write. This is already true for any owner on a source path/owner line in CODEOWNERS but it also needs to be true for the metadata entries for our workflows.
- Owners, specifically users, need to be **public** members of the Azure org. This was already true and the [instructions for requesting Azure repository access](https://eng.ms/docs/products/azure-developer-experience/onboard/access) clearly state this. The reason is that some workflows, like GitHub Actions processing, check Azure org membership and are unable to see Private members. This results in workflow processing that thinks a user isn't a member of Azure and is why some people file an issue and it gets labeled as Customer Reported even though they're an Azure org member.

**What is verified during linting?**

- **Owners**
    - Users and Teams are verified to have write permissions. (see Owners above)
    - Users are also verified to be Public members of Azure.
- **Labels**
    - Whether or not the label exists in a particular repository.
- **Source paths/Globs** (not something GitHub verifies)
    - Does the path (directory or file) exist?
    - Does the glob contain [forbidden characters](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax)?
    - Does the glob match anything in the repository?
- **Block formatting** - There are several existing metadata tags and a couple more getting added or changed. Each tag is associated with an owner or label and those pieces are verified in their respective verifications. The block formatting verification ensures that things within a block are correct. Examples would be: 
    - A PRLabel needs to be part of a block that ends in a source path/owner line 
    - A ServiceLabel needs to be part of a block that ends in a source path/owner line OR needs to be immediately followed by a #/&lt;NotInRepo&gt;/ or ServiceOwners line with declared owners. If part of a block that ends in a source path/owner line, the ServiceOwners are set to those owners.

**Where does the owner/label data come from?**
The team/user, org visibility and repository label data are pre-populated by the [github-team-user-store](https://github.com/Azure/azure-sdk-tools/tree/main/tools/github-team-user-store) which runs daily as part of the pipeline-owners-extractor and is stored in blob data that workflows can use. This had to be done this way because of the specific permissions required to get the information from GitHub. This also means that some types of fixes, such as user changing their Azure org membership from private to public, won't be immediately picked up as the pipeline-owners-extractor pipeline would need to update the data in storage first.
 
**Where will this run?**
The goal here is to run this for at least the azure-sdk-for-&lt;language&gt; public and private repositories as both a nightly pipeline as well as a CI pipeline when CODEOWNERS is changed. Where and when we want to run this is the reason the data had to be in blob storage as we can't have tokens in public pipelines.

**What about existing errors?**
Even in [repositories whose CODEOWNERS files are valid](https://github.com/Azure/azure-sdk-for-go/blob/main/.github/CODEOWNERS), according to GitHub, have errors. Because of this, repositories where this will be enabled will have a CODEOWNERS_baseline_errors.txt file which contains a deduped list of every unique error. Output from tool will filter out the baseline errors which will at least prevent new errors from being added to CODEOWNERS. As baseline issues are fixed, it's expected they'll be removed from the baseline file. The linter is capable of generating the baseline file as well as using the baseline file to filter out errors. It's also worth noting that the baseline file will only contain deduped single line errors and block formatting errors will not be filtered out. The reasons for this are 

1. None of the existing CODEOWNERS files have block formatting errors.
2. When parsing, any blocks with errors are thrown out. If block formatting errors were filtered, then someone adding their own entry could copy paste from a block that has errors and it wouldn't get flagged. This means that the entry that was just added would show up clean in the linting pipeline but wouldn't be parsed because of the errors.

**Known limitations**

- As previously mentioned, the verification data for team/user, owner org visibility and repo/label data must be generated using a GitHub token with specific permissions on a private pipeline that runs daily. This means that fixing some types of issues will have a delay before their data in Azure storage is updated.
- The baseline file means that an existing error can be added to a new line and wouldn't get flagged in CI run. Yes, this is an unfortunate side effect of the baseline file however if this wasn't done, the nightly pipelines running the linter would start out failing and probably never pass. Similarly, if the pipeline runs as part of the PR that has CODEOWNERS changes and there wasn't a baseline then I'll just say hilarity would ensue.

@weshaggard as per our discussion the following still needs to get done:

- The pipeline to build and publish the tool to the dev feed.
- The README.md for the tool
- The Azure-sdk [policies/opensource#codeowners](https://github.com/Azure/azure-sdk/blob/main/docs/policies/opensource.md#codeowners) needs to be updated

The following will happen after this PR has been merged.
- Update notification-configuration and pipeline-owners-extractor to use this instead of CodeownersParser. 
- Rewrite the github-event-processor InitialIssueTriage rule for the addition of AzureSdkOwners (already in that tool's README)
- Add baseline files to every repository the linter will be running in.


tools/codeowners-utils/Azure.Sdk.Tools.CodeownersUtils/Constants/OrgConstants.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Azure.Sdk.Tools.CodeownersUtils.Constants
+{
+    public class OrgConstants
+    {
+        // Azure is both the Org and the owner of the various repositories
+        public const string Azure = "Azure";
+        // Everyone with write permission in the Azure org either needs to be a direct user
+        // in the azure-sdk-write team or a user in a child team of azure-sdk-write. 
+        public const string AzureSdkWriteTeam = "azure-sdk-write";
+        public const string AzureOrgTeamConstant = $"{Azure}/";
```
weshaggard, line 16:
```suggestion
        public const string AzureOrgTeamConstant = "Azure/";
```

Any reason this isn't just a string contant like that?

tools/codeowners-utils/Azure.Sdk.Tools.CodeownersUtils/Holders/RepoLabelHolder.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    /// <summary>
+    /// Holder for the repo label dictionary. Lazy init, it doesn't pull the data from the URI until first use.
+    /// The GetRepoLabelData is interesting because the repo/label data is a dictionary of hashsets but, both
+    /// the dictionary and all of the hashsets need to be case insensitive because GitHub is case insensitive
+    /// but case preserving when it comes to labels. This means that MyLabel, mylabel and mYlAbEl are effectively
+    /// the same to GitHub and, let's face it, the CODEOWNERS files are all over the place with exact vs non-exact
+    /// casing.
+    /// </summary>
+    public class RepoLabelHolder
```
weshaggard, line 20: nit: Is Holder the right name for these classes? Would Cache be a more appropriate name?

weshaggard (approved):
I scanned over the code and things look reasonable. Let's get this initial version in and start working on your follow-up items.

tools/codeowners-utils/Azure.Sdk.Tools.CodeownersUtils/Constants/OrgConstants.cs
JimSuplizio: Other than if the org changes we'd only have to change the org constant. Six of one half a dozen of the other, I can hard code it.

weshaggard: Oh I didn't realize that was pulling from the other constant named Azure, lol. In that case I'm fine with what you have.

tools/codeowners-utils/Azure.Sdk.Tools.CodeownersUtils/Parsing/CodeownersEntry.cs
```diff
@@ -0,0 +1,171 @@
...
+                   );
+        }
+
+        /// <summary>
+        /// Remove all code owners which are not github alias.
+        /// Even with team expansion there can still be teams in lists. This
+        /// can happen if the team is not a child team under azure-sdk-write, which
+        /// are the only teams expanded.
+        /// </summary>
+        public void ExcludeNonUserAliases()
```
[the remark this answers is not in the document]
JimSuplizio: We absolutely use this but not in the parser or linter. Both [notification-configuration](https://github.com/Azure/azure-sdk-tools/blob/main/tools/notification-configuration/notification-creator/Contacts.cs#L125) and [pipeline-owners-extractor](https://github.com/Azure/azure-sdk-tools/blob/main/tools/pipeline-owners-extractor/Azure.Sdk.Tools.PipelineOwnersExtractor/Processor.cs#L124) use this. It's still necessary to have this in there. Teams are expanded into users if and only if they're child teams of azure-sdk-write, otherwise the team is returned as part of the list of owners.

tools/codeowners-utils/Azure.Sdk.Tools.CodeownersUtils/Holders/RepoLabelHolder.cs
JimSuplizio: Cache probably would be. I can rename them, and their folder, easily enough and will do so.

JimSuplizio: All of these changed from Holder to Cache as part of the 6th commit.

weshaggard (approved):

PR #7099, opened 2023-10-11
[tsp-client] Fix docs and clean up options

catalinaperalta:
- Update usage documentation
- Clean up unused options
- Use the same format for cli flags


tools/tsp-client/src/options.ts
```diff
@@ -85,19 +72,6 @@ export async function getOptions(): Promise<Options> {
     process.exit(0);
   }
 
-  if (values.emitter) {
-    let emitter = values.emitter.toLowerCase();
-    if (emitter in languageAliases) {
-      emitter = languageAliases[emitter]!;
-    }
-    if (!(knownLanguages as readonly string[]).includes(emitter)) {
```
xirzec, line 93: assuming this gets somewhere else, do we still need the `languageSettings.js` file at all or can we remove it?

tools/tsp-client/src/log.ts
```diff
@@ -55,26 +55,32 @@ export function printBanner() {
...
-  update      Sync and generate from a tsp project                      [string]
+  sync        Sync TypeSpec project specified in tsp-location.yaml      [string]
+  generate    Generate from a TypeSpec project                          [string]
+  update      Sync and generate from a TypeSpec project                 [string]
 
 Options:
-  -d, --debug      Enable debug logging                                [boolean]
-  -e, --emitter    Which language emitter to use             [required] [string]
-                  [choices: "csharp", "java", "javascript", "python", "openapi"]
-  -m, --mainFile   The url of the main tsp file to generate from        [string]
```
xirzec, line 73: a little sad that we're getting rid of generating from a remote tsp, but I guess I was the only one who thought that feature was cool?

xirzec (approved):
Minor remarks

tools/tsp-client/src/log.ts
weshaggard: Did you do this as opposed to the tspconfig? I think we still support remote tspconfig files correct?

catalinaperalta: We do support both local paths and remote urls, for the `init` command you can specify the `--tsp-config` option and we'll check if it's a remote url or local path and act accordingly. This tool currently aims to match the behavior of the powershell scripts so we should be able to do everything that the powershell scripts can do.

catalinaperalta: Later on, I want to add some sort of `main-file` or `entry-file` option to account for entry files that arent following standard names like main.tsp or client.tsp

tools/tsp-client/src/options.ts
catalinaperalta: Removed!

catalinaperalta:
/check-enforcer override

Tracking issues for adding tests (https://github.com/Azure/azure-sdk-tools/issues/7124) + setting up a pipeline (https://github.com/Azure/azure-sdk-tools/issues/7125)

PR #7130, opened 2023-10-13
Ensure `linux` instance handles `windows-style` path incoming

### LINKED ISSUE #7109: [TestProxy] Running Golang tests on Windows into TestProxy running on Docker WSL2/Linux container leads to incorrectly referenced recording files
# Scenario

- Test proxy docker version: [`azsdkengsys.azurecr.io/engsys/test-proxy:1.0.0-dev.20230818.1`](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/common/testproxy/target_version.txt)
- `azure-sdk-for-go` branch: `feature/storage/azbfs/newsubdirectoryclient`

I set up the test proxy via the eng/common script here: https://github.com/Azure/azure-sdk-for-go/blob/main/eng/common/testproxy/docker-start-proxy.ps1

Lead by instructions at https://msazure.visualstudio.com/One/_wiki/wikis/One.wiki/386238/Testing-Framework

# Expected behavior

Test recordings should be stored as a file under `sdk/storage/azdatalake/...` rather than as a file named `sdk\storage\azdatalake\...`

# Actual behavior

Test recordings get stored with their native Windows path separator, meaning that they're actually a file in the root folder, not a file under the correct folder

# Temporary mitigation

`git mv <incorrectly named test> <correct location>`

This only resolves getting tests into the pipeline, not running recordings locally.

scbedd:
Resolves #7109 

We already normalize all our strings (around assets, etc) using the `NormalizedString` type. Unfortunately, the issue here is that the  incoming file path needs to be in `json`. I'm honestly not quite sure how we've gotten this far without someone else hitting this.

Windows-style paths and `json` don't work well together, as aside from a couple specific, one cannot use escape characters in json strings. The json has gotta be quadruple escaped. If it's _not_, then `\\recordings\network` actually becomes `\r` `ecordings\network`.

When a customer discovered this issue, they were running tests against a linux docker proxy, hence the weirdness in the paths.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy.Tests/PlaybackTests.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy.Tests/PlaybackTests.cs
@@ -47,6 +47,37 @@ namespace Azure.Sdk.Tools.TestProxy.Tests
         }
 
         [Fact]
+        public async void TestStartPlaybackSimpleDeliberateCrossPlat()
+        {
+            RecordingHandler testRecordingHandler = new RecordingHandler(Directory.GetCurrentDirectory());
+            var httpContext = new DefaultHttpContext();
+            // in json cannot contain escape character aside from a couple specific ones. 
+            // `\` must become `\\`, and `\\` must become `\\\\`.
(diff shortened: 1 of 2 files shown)


scbedd:
This isn't the proper solution.

PR #7146, opened 2023-10-18
[tsp-client] Update README

catalinaperalta:
- Update README
- Add docs in preparation for release
- Update package name
- Update package version


tools/tsp-client/README.md
```diff
@@ -2,45 +2,65 @@
```
xirzec, line 2: Should we update the binary name and the header above? It feels confusing to have a mismatch between the binary command name and the package name.

Not that I think many will thank you for making them type out `typespec-client-generator-cli` in the terminal ;)

xirzec (approved):
Minor nit about the package name mismatch with the binary command, otherwise seems straightforward

tools/tsp-client/README.md
catalinaperalta: Yeah I know it's not perfect :/ but usage wise I think users will be more thankful to have a shorter alias to type out.

catalinaperalta: Plus this kind of keeps us in sync with the `tsp` cli, by making this `tsp-client` so hopefully the change isnt too jarring for folks who are moving from one to the other

tools/tsp-client/README.md
```diff
@@ -2,45 +2,65 @@
 
 A simple command line tool for generating TypeSpec clients.
 
-### Usage
+### Installation
 ```
-tsp-client [options] <outputDir>
+npm install @azure-tools/typespec-client-generator-cli
```
mikeharder, line 7: Since this doc is in `azure-sdk-tools`, it's fine to show how to install `latest`.  But if users are running this tool in a managed repo like `azure-rest-api-specs` or `azure-sdk-for-lang`:

1. The tool version should be pinned in a `package.json` and `package-lock.json`
2. Users should install the tool by running `npm ci` (assuming there is a lockfile)
3. Users should run the tool using `npx <tool-name>` (which uses the version installed from lockfile)

tools/tsp-client/README.md
```diff
@@ -2,45 +2,65 @@
...
-### <outputDir> (required)
-
-The only positional parameter. Specifies the directory to pass to the language emitter.
-
-### --emitter, -e (required)
-
-Specifies which language emitter to use. Current choices are "csharp", "java", "javascript", "python", "openapi".
+### Usage
+```
+tsp-client <command> [options]
```
mikeharder, line 12: I would go with an even shorter name.  How about `tspc`?  The TypeSpec compiler itself is `tsp`, and we currently use `tsv` for "TypeSpec Verification" (though I might be willing to change to `tspv`, if we want a suite of tools named `tsp_`.

tools/tsp-client/package.json
```diff
@@ -1,14 +1,14 @@
 {
-  "name": "@azure-tools/tsp-client",
-  "version": "0.0.1",
+  "name": "@azure-tools/typespec-client-generator-cli",
```
mikeharder, line 2:
```suggestion
  "name": "@azure-tools/typespec-client-generator",
```

Having `-cli` in the name seems redundant and inconsistent with other JS packages which provide a CLI.

tools/tsp-client/README.md
mikeharder: If the package name is `@azure-tools/typespec-client-generator`, another option for the command name could be `tspcg`.

mikeharder (changes requested):
Pending discussion of package and command names

tools/tsp-client/package.json
catalinaperalta: When we were discussing the package name offline we reached the conclusion that adding `-cli` at the end it clearer. The name `@azure-tools/typespec-client-generator` implies that this package actually does the generation, but in reality it's only a tool to help folks generate client libraries.

tools/tsp-client/README.md
catalinaperalta: `tspc` could be an option. Thoughts @lmazuel, @m-nash, @allenjzhang?

mikeharder: Or even `tcg` for `typespec-client-generator`?  I like relaly short CLI names.

catalinaperalta: So part of the original idea behind naming this `tsp-client` was to make it easier for when we ideally add it into the `tsp` cli, to make the transition easier so we go from `tsp-client` to `tsp client`. 

I think it is important to keep `tsp` as the beginning of the command. So in that sense, I'm ok with `tspc` to be the shorter option, keeping in mind that in the future we might have to replace other tools that use it with `tsp client`.

tools/tsp-client/README.md
catalinaperalta: I think this needs more thought on how we want language repositories to expose this to users. Added a tracking issue here: https://github.com/Azure/azure-sdk-tools/issues/7169

tools/tsp-client/README.md
mikeharder: If we think this is a temporary name until we use `tsp client`, then I agree `tsp-client` may be the best name.

catalinaperalta: Ok let's keep it like this since we would like to move this under `tsp` eventually.

PR #7160, opened 2023-10-18
Add max lifespan parameter to resource groups

benbp:
Changing our policies to set a max time ANY resources can live in a chosen subscription, if enabled.

### DIFF
--- a/eng/scripts/live-test-resource-cleanup.ps1
+++ b/eng/scripts/live-test-resource-cleanup.ps1
@@ -56,10 +56,10 @@ param (
   [switch] $DeleteArmDeployments,
 
   [Parameter()]
-  [int] $DeleteAfterHours = 24,
+  [Double] $DeleteAfterHours = 24,
 
   [Parameter()]
-  [int] $MaxLifespanDeleteAfterHours,
+  [Double] $MaxLifespanDeleteAfterHours,
 
   [Parameter()]
   [string] $AllowListPath = "$PSScriptRoot/cleanup-allowlist.txt",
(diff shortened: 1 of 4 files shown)


weshaggard (approved):

eng/scripts/live-test-resource-cleanup.ps1
```diff
@@ -333,8 +337,12 @@ function DeleteArmDeployments([object]$ResourceGroup) {
   if (!$DeleteArmDeployments) {
       return
   }
-  Write-Host "Deleting ARM deployments for group $($ResourceGroup.ResourceGroupName) as they may contain secrets. Deployed resources will not be affected."
-  $null = Get-AzResourceGroupDeployment -ResourceGroupName $ResourceGroup.ResourceGroupName | Remove-AzResourceGroupDeployment
+  $toDelete = Get-AzResourceGroupDeployment -ResourceGroupName $ResourceGroup.ResourceGroupName | Where-Object { $_.Outputs.Count }
```
weshaggard, line 340: Are there possibility of secrets outside of the outputs?

benbp: They could be on the inputs/parameters but we haven't seen those anywhere. I'm just adding it here because deleting deployments takes quite a while and this narrows the operations by about 90%

eng/scripts/live-test-resource-cleanup.ps1
```diff
@@ -366,22 +375,38 @@ function DeleteOrUpdateResourceGroups() {
     if ((IsChildResource $rg) -or (HasDeleteLock $rg)) {
       continue
     }
-    if ((HasDoNotDeleteTag $rg) -or (HasValidAliasInName $rg) -or (HasValidOwnerTag $rg)) {
+    if (HasDoNotDeleteTag $rg) {
```
weshaggard, line 378: Do we still want to support this tag? Is there a way to put a lock on a resource group?

benbp: Yes you can add a delete lock. Let's chat more on the Monday meeting about this. It could be nice to remove the tag, though a benefit is we know we added it manually vs. a delete lock that may have been part of a deployment or some other automatic reason.

PR #7167, opened 2023-10-19
Added doxygen comment support to C++ API Reviews.

LarryOsterman:
This PR implements 3 new features to the C++ API View tool.

1. Enable comments in API Views.
2. Add a line to each type describing where the type is located in the source tree
3. Flag typedefs in the global namespace as being challenging. This raises the visibility of using types like uint8_t in the global namespace.

### DIFF
--- a/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstDumper.hpp
+++ b/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstDumper.hpp
@@ -12,6 +12,7 @@ class AstDumper {
   std::string m_currentNamespace;
   std::vector<std::string> m_namespaceComponents;
   int m_indentationLevel{};
+  size_t m_currentCursor{};
 
   void OpenNamespace(std::string_view const& namespaceName);
   void OpenNamespaces(
@@ -22,6 +23,9 @@ class AstDumper {
       std::vector<std::string> const& namespaceComponents,
       std::vector<std::string>::iterator& current);
 
+protected:
+  void UpdateCursor(size_t cursorAdjustment) { m_currentCursor += cursorAdjustment; }
+
 public:
   // Note about the different functions.
   // Functions named "InsertXxx" and "AddXxx" are intended to insert elements into the output
@@ -30,6 +34,7 @@ public:
   void AdjustIndent(int indentDelta = 0);
   void LeftAlign();
   void Newline();
+  size_t GetCurrentCursor() { return m_currentCursor; }
   void SetNamespace(std::string_view const& currentNamespace);
 
   virtual void InsertNewline() = 0;
@@ -43,18 +48,21 @@ public:
       std::string_view const& type,
       std::string_view const& typeNavigationId)
       = 0;
-  virtual void InsertMemberName(std::string_view const& member, std::string_view const& memberFullName) = 0;
+  virtual void InsertMemberName(
+      std::string_view const& member,
+      std::string_view const& memberFullName)
+      = 0;
   virtual void InsertStringLiteral(std::string_view const& str) = 0;
   virtual void InsertLiteral(std::string_view const& str) = 0;
   virtual void InsertComment(std::string_view const& comment) = 0;
+  virtual void AddExternalLinkStart(std::string_view const& linkValue) = 0;
+  virtual void AddExternalLinkEnd() = 0;
   virtual void AddDocumentRangeStart() = 0;
   virtual void AddDocumentRangeEnd() = 0;
   virtual void AddDeprecatedRangeStart() = 0;
   virtual void AddDeprecatedRangeEnd() = 0;
   virtual void AddDiffRangeStart() = 0;
   virtual void AddDiffRangeEnd() = 0;
-  virtual void AddInheritanceInfoStart() = 0;
-  virtual void AddInheritanceInfoEnd() = 0;
 
   virtual void DumpTypeHierarchyNode(std::shared_ptr<TypeHierarchy::TypeHierarchyNode> const& node)
       = 0;
--- a/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstDumper.hpp
+++ b/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstDumper.hpp
@@ -26,7 +26,6 @@ class AstDumper {
 protected:
   void UpdateCursor(size_t cursorAdjustment) {
       m_currentCursor += cursorAdjustment; 
-      assert(m_currentCursor >= 0);
   }
 
 public:
--- a/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstDumper.hpp
+++ b/tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstDumper.hpp
@@ -24,9 +24,7 @@ class AstDumper {
       std::vector<std::string>::iterator& current);
 
 protected:
-  void UpdateCursor(size_t cursorAdjustment) {
-      m_currentCursor += cursorAdjustment; 
-  }
+  void UpdateCursor(size_t cursorAdjustment) { m_currentCursor += cursorAdjustment; }
 
 public:
   // Note about the different functions.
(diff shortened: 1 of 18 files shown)


tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -1146,6 +1188,61 @@ void AstNamedNode::DumpAttributes(AstDumper* dumper, DumpNodeOptions const& opti
...
+{
+  if (options.NeedsLeadingNewline)
+  {
+    dumper->Newline();
+  }
+  if (options.NeedsLeftAlign)
+  {
+    dumper->LeftAlign();
+  }
+  dumper->InsertComment("// ");
```
RickWinter, line 1230: InsertComment(string) w/o the "//".  Have the insertcomment method handle splitting the lines if they are too long and have the method handle inserting the "// " prefix.

If need be you could have an overload InsertComment(string stringToWrrite, bool noPrefix)

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -1703,6 +1831,13 @@ class AstFunction : public AstNamedNode {
         m_exceptionSpecification{func->getExceptionSpecType()}
   {
     m_navigationId = GetNavigationId(func);
+
+    // If the type of the return value is in the global namespace, then flag it as an error.
+    if (m_returnValue.IsTypeInGlobalNamespace())
+    {
+      database->CreateApiViewMessage(ApiViewMessages::TypedefInGlobalNamespace, m_navigationId);
+    }
```
RickWinter, line 1839: Shame this method has to be manually added to every overload.  Could this be on the base?

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -1985,6 +2130,10 @@ class AstConstructor : public AstMethod {
   }
   void DumpNode(AstDumper* dumper, DumpNodeOptions const& dumpOptions) const override
   {
+    if (dumpOptions.NeedsDocumentation)
+    {
+      DumpDocumentation(dumper, dumpOptions);
+    }
```
RickWinter, line 2136: Could simplify to avoid the if check by calling `DumpDocumentation(dumper, dumpOptions);` and letting that method do nothing if the dumpOptions.NeedsDocumentation == false.  Saves writing the if check everywhere

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/CommentExtractor.cpp
```diff
@@ -0,0 +1,875 @@
...
+  //     const clang::comments::BlockCommandComment* bc)
+  //{
+  //   auto rv{AstDocumentation::Create(bc)};
+  //   // Gather the text for the block command comment.
+  //   for (auto child = bc->child_begin(); child != bc->child_end(); child++)
+  //   {
+  //     rv->AddChild(visit(*child));
+  //   }
+  //   return rv;
+  // };
```
RickWinter, line 708: Commented code?

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/ProcessorImpl.cpp
```diff
@@ -430,7 +435,8 @@ int ApiViewProcessorImpl::ProcessApiView()
     assert(file.u8string().find(m_currentSourceRoot.u8string()) == 0);
     auto relativeFile = static_cast<std::string>(
         stringFromU8string(file.u8string().erase(0, m_currentSourceRoot.u8string().size() + 1)));
-    std::string quotedFile = replaceAll(relativeFile, "\\", "\\\\");
+ //   std::string quotedFile = replaceAll(relativeFile, "\\", "\\\\");
```
RickWinter, line 438: Removed commented out line

RickWinter (approved):
Would have been a little quicker to review if there were more comments around the new types and methods.  They make sense once you read all the code, but individually it initally took some learning to understand.  Suggestion: beef up the overview of what methods/classes do.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstDumper.hpp
```diff
@@ -22,6 +23,9 @@ class AstDumper {
       std::vector<std::string> const& namespaceComponents,
       std::vector<std::string>::iterator& current);
 
+protected:
+  void UpdateCursor(size_t cursorAdjustment) { m_currentCursor += cursorAdjustment; }
```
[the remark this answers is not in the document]
LarryOsterman: Point.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
LarryOsterman: The "Insert" (and "Add") APIs are "stupid" APIs which simply emit elements to the ApiView output. An "Insert" method inserts content which will normally be seen by the customer, a "Add" method inserts metadata.

This change includes some of the scaffolding needed to implement soft line-wrap functionality, but it's not complete (I need a way of marking a cursor and resetting to the cursor).

For simplicities sake, I chose to not build comment reflowing functionality into this code, instead I preserve the flow from clang (mostly - block comment directives are treated as being an inline type, which suppresses the leading crlf on a paragraph node).

This particular code handles inserting the source filename, which we don't ever want to flow.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
LarryOsterman: The base doesn't know what types are associated with a node - for instance, some nodes (like method nodes) have types, others (like TemplateTemplateParameter (which represents a template parameter to a template)) do not. So there's really no way of dropping it into the base class. 

Also the navigationId (which lets ApiView know where to put the error message) is top level node specific. And the database pointer is only used during the construction of the AstNode so it isn't included in the base class.

Putting that all together, if you were to put in a base class it would end up looking like:

```cpp
CheckForTypeInGlobalNamespace(database, m_returnValue, m_navigationId);
```

instead of:
```cpp
if (m_returnValue.IsTypeInGlobalNamespace())
{
  database->CreateApiViewMessage(ApiViewMessages::TypedefInGlobalNamespace, m_navigationId);
}
```

and I'm not sure it's a significant improvement.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
LarryOsterman: Great suggestion.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/CommentExtractor.cpp
LarryOsterman: Removed. This was from when I did all the work in the visitor rather than in classes derived from AstDocumentation..

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstDumper.hpp
```diff
@@ -24,7 +24,10 @@ class AstDumper {
       std::vector<std::string>::iterator& current);
 
 protected:
-  void UpdateCursor(size_t cursorAdjustment) { m_currentCursor += cursorAdjustment; }
+  void UpdateCursor(size_t cursorAdjustment) {
+      m_currentCursor += cursorAdjustment; 
+      assert(m_currentCursor >= 0);
```
RickWinter, line 29: This assert will always be true since size_t is an unsigned int.  Instead you'd need to see if cursorAdjustment is > max(size_t)-m_currentCursor.  The more I think about it, the less I believe checking matters :)

RickWinter (approved):

RickWinter (approved):

PR #7171, opened 2023-10-19
Fix update test resources tagging when tags are empty. Bump max limit

benbp:


eng/common/TestResources/Update-TestResources.ps1
```diff
@@ -136,6 +136,9 @@ try {
 
     Log "Updating DeleteAfter to '$deleteAfter'"
     Write-Warning "Any clean-up scripts running against subscription '$SubscriptionId' may delete resource group '$ResourceGroupName' after $DeleteAfterHours hours."
+    if (!$resourceGroup.Tags) {
+        $resourceGroup.Tags = @{}
```
heaths, line 140: This is actually necessary? Seems more like a bug on the Az PowerShell module. As a workaround, though, I see no harm but if it is a bug, we should make sure it's reported. PowerShell is truthy, so idiomatically so should cmdlets be.

heaths (approved):

eng/common/TestResources/Update-TestResources.ps1
benbp: It's all json at the API level so I'm not surprised if the underlying value was `null` that powershell just converted it to `$null` without initializing empty values?

heaths: That's my assumption as well...but should. If the value is falsy, it should omit the tags in the request or pass an empty object - whatever the control plane requires in that case.

benbp: The issue here is we try to index into $null a line below, not that we're submitting bad API data.

heaths: Ah, I see. Thanks for the clarification. That said, does sending whatever `@{}` serializes into cause problems?

benbp: It worked when I tested it out. The response object is also a hashtable:

```
$ $g[0].Tags.GetType()

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     Hashtable                                System.Object
```

PR #7172, opened 2023-10-19
Update tools for new Codeowners utils

JimSuplizio:
Update identity-resolution, notification-configuration and pipeline-owners-extractor to use the new CodeownersUtils instead of CodeownersParser.

identity-resolution never had a solution, Visual Studio kept creating one every time the project was loaded and wanted to save it off. I relented and I'm checking that in here.

I also fixed the ci.yml files that didn't correctly the original parser as part of their include paths. This is reason that pipeline-owners-extractor want's expanding teams when that functionality was added months ago.

### DIFF
--- a/tools/identity-resolution/Services/GitHubService.cs
+++ b/tools/identity-resolution/Services/GitHubService.cs
@@ -5,7 +5,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
-using Azure.Sdk.Tools.CodeOwnersParser;
+using Azure.Sdk.Tools.CodeownersUtils.Parsing;
+using System.IO;
 
 namespace Azure.Sdk.Tools.NotificationConfiguration
 {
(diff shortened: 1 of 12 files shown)


tools/identity-resolution/Services/GitHubService.cs
```diff
@@ -53,21 +53,31 @@ public async Task<List<CodeownersEntry>> GetCodeownersFileEntries(Uri repoUrl)
         /// </summary>
         /// <param name="repoUrl"></param>
         /// <returns></returns>
-        private async Task<List<CodeownersEntry>> GetCodeownersFileImpl(Uri repoUrl)
+        private List<CodeownersEntry> GetCodeownersFileImpl(Uri repoUrl)
```
JimSuplizio, line 56: I don't know this was originally coded as an async task that fetched the CODEOWNERS contents and passed those into the parser as a string. The original parser has its own file utilities to fetch the file and had an entry point that took a URL and did this. My parser needs the file as a list of string and only has the entry point that takes the absolute file path or URL. As such, I had to make this function and the calling function, GetCodeownersFileEntries, non-async.

weshaggard (approved):

PR #7183, opened 2023-10-24
Clean up arm deployments with secret parameters

benbp:
@weshaggard you called it.

### DIFF
--- a/eng/scripts/live-test-resource-cleanup.ps1
+++ b/eng/scripts/live-test-resource-cleanup.ps1
@@ -337,7 +337,8 @@ function DeleteArmDeployments([object]$ResourceGroup) {
   if (!$DeleteArmDeployments) {
     return
   }
-  $toDelete = @(Get-AzResourceGroupDeployment -ResourceGroupName $ResourceGroup.ResourceGroupName | Where-Object { $_ -and $_.Outputs?.Count })
+  $toDelete = @(Get-AzResourceGroupDeployment -ResourceGroupName $ResourceGroup.ResourceGroupName `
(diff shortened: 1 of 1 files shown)


eng/scripts/live-test-resource-cleanup.ps1
```diff
@@ -337,7 +337,8 @@ function DeleteArmDeployments([object]$ResourceGroup) {
   if (!$DeleteArmDeployments) {
     return
   }
-  $toDelete = @(Get-AzResourceGroupDeployment -ResourceGroupName $ResourceGroup.ResourceGroupName | Where-Object { $_ -and $_.Outputs?.Count })
+  $toDelete = @(Get-AzResourceGroupDeployment -ResourceGroupName $ResourceGroup.ResourceGroupName `
+              | Where-Object { $_ -and ($_.Outputs?.Count -or $_.Parameters?.ContainsKey('testApplicationSecret')) })
```
weshaggard, line 341: Should we just use `contains("secret")`

benbp: I think this will actually false positive a bunch of stuff, as it appears some deployments have the empty object which will match on the substring. I could check the value of all keys matching `*secrets*` but it seems excessive.

```
secrets              SecureObject               null
```

weshaggard: It seems odd to special case just this one parameter name is there something special about this specific one?

benbp: This is the one we pass in from new-testresources.ps1 so by convention a lot of live test templates use it. Otherwise I only have seen secrets in some azure function deploys from azd.

weshaggard (approved):

PR #7192, opened 2023-10-25
Add storage account immutability cleanup to live test resources script

benbp:
I think the storage account cleanup script works reliably enough now that we can integrate into the main live test resource cleanup pipeline.

### DIFF
--- a/eng/scripts/live-test-resource-cleanup.ps1
+++ b/eng/scripts/live-test-resource-cleanup.ps1
@@ -427,9 +427,17 @@ function DeleteAndPurgeGroups([array]$toDelete) {
         $purgeableResources += $purgeableResourcesFromRG
         Write-Verbose "Found $($purgeableResourcesFromRG.Count) potentially purgeable resources in resource group $($rg.ResourceGroupName)"
       }
+
       Write-Verbose "Deleting group: $($rg.ResourceGroupName)"
       Write-Verbose "  tags $($rg.Tags | ConvertTo-Json -Compress)"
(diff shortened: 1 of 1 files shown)


eng/scripts/live-test-resource-cleanup.ps1
```diff
@@ -427,9 +427,17 @@ function DeleteAndPurgeGroups([array]$toDelete) {
...
       }
+
       Write-Verbose "Deleting group: $($rg.ResourceGroupName)"
       Write-Verbose "  tags $($rg.Tags | ConvertTo-Json -Compress)"
-      Write-Host ($rg | Remove-AzResourceGroup -Force -AsJob).Name
+
+      # For storage tests specifically, if they are aborted then blobs with immutability policies
+      # can be left around which prevent deletion.
+      if ($rg.Tags?.ContainsKey('ServiceDirectory') -and $rg.Tags.ServiceDirectory -like '*storage*') {
+        $PSScriptRoot/Remove-WormStorageAccounts.ps1 -GroupPrefix $rg.ResourceGroupName
```
scbedd, line 437: We really meant `Remove-WarmStorageAccounts` right? Or is this removal script actually somewhat wormlike in that it traverses relationships?

scbedd (approved):
Wording question, but signing off otherwise.

eng/scripts/live-test-resource-cleanup.ps1
scbedd: Nope intentional: https://github.com/Azure/azure-sdk-tools/blob/main/eng/scripts/Remove-WormStorageAccounts.ps1

PR #7196, opened 2023-10-26
Adding automatic reset on failed push

### LINKED ISSUE #6664: Invalid `push` state not recoverable without additional `restore` or assets cleanup
Reported by @samvaity, thanks Sameeksha!

Let's say a user has a `test-proxy push` fail due to a permissions error. We are left in a situation where `test-proxy push` doesn't recognize that there is work to be done, so follow-up `test-proxy push` without re-recording don't actually GO anywhere. 

Test-proxy just thinks that there is no work to be pushed because there are **no pending changes in the assets repo slice**. They were all commit in the previous `push` attempt, which never went up! Need to be able to differentiate this situation.

scbedd:
Resolves #6664  

For the most part, users are entirely self-service on the test-proxy, except for one common situation. This one:

1. User is onboarding to proxy for the first time
2. User records new recordings successfully
3. User attempts to `push` the new recordings. They fail with 403. They join the appropriate group.
4. User re-attempts `push`. `Push` silently just does nothing.

The origin of this is because we only `push` when there uncommitted changes _to be pushed_. There's nothing handling when a push fails.

By resetting the changes, users are unblocked from this common scenario, and a re-entry to the `push` function will act as if the failed push never happened. 

~~TODO:~~

- [x] Add integration test, We can "fake" this scenario by restoring from public, then deliberately corrupting the github API token secret, then restoring it, and committing again.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
@@ -160,7 +160,7 @@ namespace Azure.Sdk.Tools.TestProxy.Store
                     // - ls-remote origin
                     // - push
                     // if we have a failure on either of these, we need to unstage our changes for an easy re-attempt at pushing.
-                    GitHandler.TryRun("git reset --soft HEAD^", config.AssetsRepoLocation.ToString(), out CommandResult ResetResult);
+                    GitHandler.TryRun("reset --soft HEAD^", config.AssetsRepoLocation.ToString(), out CommandResult ResetResult);
 
                     throw GenerateInvokeException(e.Result);
                 }
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
@@ -155,11 +155,21 @@ namespace Azure.Sdk.Tools.TestProxy.Store
                 catch(GitProcessException e)
                 {
                     HideOrigin(config);
+
+                    // the only executions that have a real chance of failing are
+                    // - ls-remote origin
+                    // - push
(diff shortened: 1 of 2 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
```diff
@@ -155,11 +155,21 @@ public async Task<NormalizedString> GetPath(string pathToAssetsJson)
...
+                    // if we have a failure on either of these, we need to unstage our changes for an easy re-attempt at pushing.
+                    GitHandler.TryRun("git reset --soft HEAD^", config.AssetsRepoLocation.ToString(), out CommandResult ResetResult);
+
                     throw GenerateInvokeException(e.Result);
                 }
                 await UpdateAssetsJson(generatedTagName, config);
                 await BreadCrumb.Update(config);
             }
+            // can we detect that we have unpushed commits?
+            else if
```
scbedd, line 171:
```suggestion
            else if
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
```diff
@@ -155,6 +155,13 @@ public async Task<NormalizedString> GetPath(string pathToAssetsJson)
                 catch(GitProcessException e)
                 {
                     HideOrigin(config);
+
+                    // the only executions that have a real chance of failing are
+                    // - ls-remote origin
+                    // - push
+                    // if we have a failure on either of these, we need to unstage our changes for an easy re-attempt at pushing.
+                    GitHandler.TryRun("git reset --soft HEAD^", config.AssetsRepoLocation.ToString(), out CommandResult ResetResult);
```
scbedd, line 163: We're basically ignoring the exit code here, as there is no reason to care. Wherever it failed, the changes would have already been committed.

benbp (approved):

PR #7200, opened 2023-10-27
Empty package models don't throw error

scbedd:
On `python` `pypy39` platform, this script runs into a [weird problem.](https://dev.azure.com/azure-sdk/public/_build/results?buildId=3209852&view=logs&j=8e5196d1-9c2a-56bb-41fd-e329e85c48fa&t=6cc1318c-0278-53f9-9f83-99109e387aaa)

The linked build isolates the error, along with a bunch of debug output showing where stuff is going wrong.

On this `pypy39` platform image, we are getting 2 _extra_ "blank" pkg props. That looks like this for the template project:

image

Here's the thing, these are _magically_ showing up. In the same build that produces this error for review, you can see clearly that the _actual_ `Language-Settings.ps1` function that is retrieving the package list, is **getting the correct list**. It just magically gets another two blank objects. I'm almost positive that we're probably leaking two objects from previous commands, but honestly, I've dug at this for a while and I haven't seen it yet.

The fix [proves working](https://dev.azure.com/azure-sdk/public/_build/results?buildId=3209898&view=results) in this build.

### DIFF
--- a/eng/common/scripts/Save-Package-Properties.ps1
+++ b/eng/common/scripts/Save-Package-Properties.ps1
@@ -101,34 +101,36 @@ if ($allPackageProperties)
     }
     foreach($pkg in $allPackageProperties)
     {
-        Write-Host "Package Name: $($pkg.Name)"
-        Write-Host "Package Version: $($pkg.Version)"
-        Write-Host "Package SDK Type: $($pkg.SdkType)"
-        Write-Host "Artifact Name: $($pkg.ArtifactName)"
-        Write-Host "Release date: $($pkg.ReleaseStatus)"
-        $configFilePrefix = $pkg.Name
-        if ($pkg.ArtifactName)
-        {
-          $configFilePrefix = $pkg.ArtifactName
-        }
-        $outputPath = Join-Path -Path $outDirectory "$configFilePrefix.json"
-        Write-Host "Output path of json file: $outputPath"
-        $outDir = Split-Path $outputPath -parent
-        if (-not (Test-Path -path $outDir))
-        {
-          Write-Host "Creating directory $($outDir) for json property file"
-          New-Item -ItemType Directory -Path $outDir
-        }
+        if ($pkg.Name) {
+          Write-Host "Package Name: $($pkg.Name)"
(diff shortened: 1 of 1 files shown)


mikeharder (approved):

eng/common/scripts/Save-Package-Properties.ps1
```diff
@@ -101,34 +101,36 @@ if ($allPackageProperties)
...
-        Write-Host "Package Name: $($pkg.Name)"
-        Write-Host "Package Version: $($pkg.Version)"
-        Write-Host "Package SDK Type: $($pkg.SdkType)"
-        Write-Host "Artifact Name: $($pkg.ArtifactName)"
-        Write-Host "Release date: $($pkg.ReleaseStatus)"
-        $configFilePrefix = $pkg.Name
-        if ($pkg.ArtifactName)
-        {
-          $configFilePrefix = $pkg.ArtifactName
+        if ($pkg.Name) {
```
weshaggard, line 104: I think this is better fixed directly in the python repo at https://github.com/Azure/azure-sdk-for-python/blob/main/eng/scripts/Language-Settings.ps1#L27. In general we should never be getting invalid package entries back from Get-AllPkgProperties.

scbedd: I agree, but in my log example it _is being returned properly from that function_. You can see "Length 1"? That's what is being returned from the Language-Settings function. We're _somehow_ getting two additional outputs just on this platform, but NOT from that function.

scbedd: This is what I'm talkin bout: 
image

That function _is_ returning the expected number, and then Get-AllPkgProperties is getting two additional before returning.

scbedd: "Returning PackageProps" is leaving the function that you're asking me to fix. "Length 1" is the lenght of PackageProps being returned.

weshaggard: Very strange I cannot explain this behavior either. I'm looking through your code branch to try and understand how this could happen.

Issue #7049, opened 2023-10-02
[tsp-client] Fix code style


catalinaperalta:
- Fix instances of `foo === undefined`. See comment: https://github.com/Azure/azure-sdk-tools/pull/6843#discussion_r1339104523
- Improve conditional checks and access values, like in this example: https://github.com/Azure/azure-sdk-tools/pull/6843#discussion_r1339107036

Issue #7048, opened 2023-10-02
[tsp-client] Refactor code to use optional chaining


catalinaperalta:
See comment: https://github.com/Azure/azure-sdk-tools/pull/6843#discussion_r1339104170

PR #7202, opened 2023-10-28
[tsp-client] Code fixes

catalinaperalta:
PR is mainly to improve code style, fix some logic, correct async functions
Fixes https://github.com/Azure/azure-sdk-tools/issues/7049
Fixes https://github.com/Azure/azure-sdk-tools/issues/7048


tools/tsp-client/src/git.ts
```diff
@@ -2,14 +2,17 @@ import { spawn } from "child_process";
...
-    if (err) {
-      throw err;
-    }
+  return new Promise((resolve, reject) => {
+    simpleGit(rootUrl).clone(repo, cloneDir, ["--no-checkout", "--filter=tree:0"], (err) => {
+      if (err) {
+        reject(err);
+      }
+    });
+    resolve();
```
catalinaperalta, line 15: Is this the correct way to do this?

joheredi: I would think it is inside the callback. So if you don't fall into if(err){} I guess that means clone was succesful, right? If so you can resolve there
```suggestion
  return new Promise((resolve, reject) => {
    simpleGit(rootUrl).clone(repo, cloneDir, ["--no-checkout", "--filter=tree:0"], (err) => {
      if (err) {
        reject(err);
      }
      resolve();
    });
```

tools/tsp-client/src/index.ts
```diff
@@ -234,17 +224,17 @@ async function main() {
             throw new Error("Commit SHA is required when specifying `--repo`, please specify a commit using `--commit`");
         }
         if (options.commit) {
-          let [ directory, commit, repo, additionalDirectories ] = await readTspLocation(rootUrl);
-          commit = options.commit ?? commit;
-          repo = options.repo ?? repo;
-          await writeFile(path.join(rootUrl, "tsp-location.yaml"), `directory: ${directory}\ncommit: ${commit}\nrepo: ${repo}\nadditionalDirectories: ${additionalDirectories}`);
+          let tspLocation: TspLocation = await readTspLocation(rootUrl);
```
joheredi, line 227: if you are not reassigning tspLocation it is usually recommended to make it a cosnt

joheredi: even if you modify some of the properties

tools/tsp-client/src/fs.ts
```diff
@@ -1,8 +1,8 @@
-import { mkdir, rm, writeFile, stat, readFile, access } from "node:fs/promises";
-import { FileTreeResult } from "./fileTree.js";
+import { mkdir, rm, stat, readFile, access } from "node:fs/promises";
```
joheredi, line 1: does it work if you import from "fs/promises"? without `node:` just curious, as I'm not used to this notation.

tools/tsp-client/src/git.ts
```diff
@@ -2,14 +2,17 @@ import { spawn } from "child_process";
 import { simpleGit } from "simple-git";
 
 export async function getRepoRoot(repoPath: string): Promise<string> {
-  return await simpleGit(repoPath).revparse(["--show-toplevel"]);
+  return simpleGit(repoPath).revparse(["--show-toplevel"]);
```
joheredi, line 5: There is also [Octokit](https://www.npmjs.com/package/@octokit/rest) which is github's official JS library, it is very popular. I'm not suggesting to change from simple-git, just wanted to mention it in case you need to consider alternatives in the future.

tools/tsp-client/src/index.ts
```diff
@@ -33,77 +33,67 @@ async function sdkInit(
...
+    const serviceDir = configYaml?.parameters?.["service-dir"]?.default;
+    if (!serviceDir) {
+      Logger.error(`Parameter service-dir is not defined correctly in tspconfig.yaml. Please refer to https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml for the right schema.`)
+    }
+    Logger.debug(`Service directory: ${serviceDir}`)
+    const additionalDirs: string[] = configYaml?.parameters?.dependencies?.additionalDirectories ?? [];
+    Logger.info(`Additional directories: ${additionalDirs}`)
+    const packageDir = configYaml?.options?.[emitter]?.["package-dir"];
+    if (!packageDir) {
+      throw new Error(`Missing package-dir in ${emitter} options of tspconfig.yaml. Please refer to https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml for the right schema.`);
```
joheredi, line 66: Thanks for the detailed error messages this is greatly appreciated!

joheredi (approved):
looks great! Love the improved error messages. Left a few minor comments

tools/tsp-client/src/git.ts
catalinaperalta: Thanks! I'll look into it if I need it in the future :)

tools/tsp-client/src/fs.ts
catalinaperalta: It does seem to work without it. I'm not sure why this was added for me. I could look into removing it in the future if it's better. Let me know! I'll merge this PR in without that change though

Issue #7190, opened 2023-10-25
test-proxy-tool.yml not collecting proxy output properly


scbedd:
[Context](https://dev.azure.com/azure-sdk/public/_build/results?buildId=3199449&view=results)

@gearama  reached out with some oddness that he's seeing. He can't repro these issues locally.

scbedd:
Additional context as I investigate.

It looks like the `404` is being thrown, but we aren't seeing the exception details in the proxy log. Need to understand why we're getting these 404s without a description of what is happening to cause the mismatch.

Seeing if there is a proxy issue concealing something. Published [20231025.4](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3199622) to account for possible proxy issue.

scbedd:
Ok. I understand what's going on now.

`test-proxy-tool.yml` isn't redirecting stderr to the proxy log, so we're not getting the exception logs appearing where we want. THAT is why it's appearing to `404` without actually saying "why".

This is a fairly straightforward fix in `test-proxy-tool.yml` and `test-proxy-local-tool.yml`.

PR #7204, opened 2023-10-28
Add option to allow single-log output for test-proxy

scbedd:
Addressing #7190 in a different way. Going to enable this new version of the proxy, then adjust `test-proxy-tool` to run with this new flag.

This is much much less hacky than just uploading the `stderr` output as a different file, and allows the proxy log to show up in a useful format, where errors are right next to the `info` logs that they are correlated with.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Startup.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Startup.cs
@@ -149,7 +149,7 @@ namespace Azure.Sdk.Tools.TestProxy
                         loggingBuilder.AddConfiguration(hostBuilder.Configuration.GetSection("Logging"));
                         loggingBuilder.AddConsole(options =>
                         {
-                            options.LogToStandardErrorThreshold = LogLevel.Error;
+                            options.LogToStandardErrorThreshold = startOptions.UniversalOutput ? LogLevel.None : LogLevel.Error;
                         }).AddSimpleConsole(options =>
                         {
                             options.TimestampFormat = "[HH:mm:ss] ";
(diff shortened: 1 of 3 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Startup.cs
```diff
@@ -149,7 +149,7 @@ private static void StartServer(StartOptions startOptions)
                         loggingBuilder.AddConfiguration(hostBuilder.Configuration.GetSection("Logging"));
                         loggingBuilder.AddConsole(options =>
                         {
-                            options.LogToStandardErrorThreshold = LogLevel.Error;
+                            options.LogToStandardErrorThreshold = startOptions.UniversalOutput ? LogLevel.None : LogLevel.Error;
```
scbedd, line 152: setting error threshold to non means that everything will show up on stdout.

benbp:
I'm a bit behind on the context here, why is this preferable over the conventional `2>&1` approach?

scbedd:
> I'm a bit behind on the context here, why is this preferable over the conventional `2>&1` approach?

Totally agree. If I only supported `windows` agents I think that would work. In my experience, to consistently get the proxy process running in background on windows CI agents, I needed to use `Start-Process` to reliably get it surviving after the individual agent task ended. `Start-Process` does support `RedirectStandardError`, but not to the same file that is `RedirectStandardOut`.

This seemed like the least hacky, easiest way to make that happen. The other tools pr then becomes modifying `test-proxy-tool.yml` to call with `-U`. No changes to the artifacts or log dump, which two files would necessitate. Doesn't really help the users to be separate logs in CI anyway, as it is more useful to have all errors right next to the stdout that lead up to them.

benbp (approved):

PR #7210, opened 2023-10-29
Reduce APIView search query cost when searching without filters

praveenkuttappan:
APIView search without any filters is taking around 6-7 seconds to find the records due to search against revision properties.

This change is to improve search performance to few milli seconds when searching without any filters by checking against review properties only.

Search query cost is reduced to 30 RUs from 2800 RUs and response time is few ms instead of 6 seconds with this change.

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Repositories/CosmosReviewRepository.cs
+++ b/src/dotnet/APIView/APIViewWeb/Repositories/CosmosReviewRepository.cs
@@ -207,15 +207,13 @@ namespace APIViewWeb
                 }
                 else
                 {
-                    queryStringBuilder.Append($" AND (r.Author IN {searchAsQueryStr}");
+                    queryStringBuilder.Append($" AND (");
                     foreach (var item in search) 
                     {
                         var query = '"' + $"{item}" + '"';
-                        queryStringBuilder.Append($" OR CONTAINS(ARRAY_SLICE(r.Revisions, -1)[0].Name, {query}, true)");
-                        queryStringBuilder.Append($" OR CONTAINS(r.Name, {query}, true)");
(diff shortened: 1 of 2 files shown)


benbp (approved):

src/dotnet/APIView/APIViewWeb/Repositories/CosmosReviewRepository.cs
```diff
@@ -207,15 +207,13 @@ public async Task<IEnumerable<string>> GetReviewFirstLevelPropertiesAsync(string
                 }
                 else
                 {
-                    queryStringBuilder.Append($" AND (r.Author IN {searchAsQueryStr}");
+                    queryStringBuilder.Append($" AND (");
                     foreach (var item in search) 
                     {
                         var query = '"' + $"{item}" + '"';
-                        queryStringBuilder.Append($" OR CONTAINS(ARRAY_SLICE(r.Revisions, -1)[0].Name, {query}, true)");
```
weshaggard, line 214: Does this mean that we will not find things that have the search query string in the revisions now, and is that a big deal?

weshaggard:
FYI @KrzysztofCwalina

weshaggard (approved):

src/dotnet/APIView/APIViewWeb/Repositories/CosmosReviewRepository.cs
praveenkuttappan: Name property in review level is package name. Name in the Revision has file name, and this is same in all cases. So, this won't impact search results when searching a review with a given package name. If anyone wants to find reviews submitted by a specific user, then filter `author:` can be used.

PR #7221, opened 2023-11-01
Update search query to use exists to verify revisions data

praveenkuttappan:
Search to check again revisions using exists clause instead of array_slice to improve query performance


praveenkuttappan:
Exists clause still processes records in few milli seconds. JavaScript review has different name at review and revision level due to JS namespace property in package name and this was not considered correctly in previous search query change.

weshaggard (approved):
How much increase in performance does this cause?

praveenkuttappan:
> How much increase in performance does this cause?

For few samples I tried, query response was changed from 100 ms to 500- 600ms. So it's still better. Issue is when using Array_slice within the query to find the last revision.

PR #7236, opened 2023-11-05
Updates to github-event-processor

### LINKED ISSUE #5743: Update GitHubEventProcessor's Initial Issue Triage Rule once the proposed CODEOWNERS updates have been made.
Right GitHubEventProcessor's processing for the Initial Issue Triage rule not what we'd like it to be. The rule, as written by @jsquire, requires information that we do not yet have. @konrad-jamrozik proposed updates to CODEOWNERS that would give us this information, but those changes have not yet been made and still TDB on who will make them and when. This issue to track that, when they are made, the Initial Issue Triage rule will need to be updated accordingly.

### LINKED ISSUE #6480: Issue triaging bot should tag service owners no matter which order the labels are applied
Currently as implemented (https://github.com/Azure/azure-sdk-tools/blob/main/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/EventProcessing/IssueProcessing.cs#L164) the service attention rule will only trigger when the "Service Attention" label is applied. So in a case where the "Service" label is applied after this rule will not correctly tag users in the comments.

cc @jsquire @JimSuplizio

### LINKED ISSUE #6524: [github event processor] Add retries and/or ignore failure checking rate limit api
See https://github.com/Azure/azure-sdk-for-net/actions/runs/5559064410/jobs/10154804951 for an example of a job run that failed just because of a transient github server error checking the rate limit API. It appears from the source https://github.com/Azure/azure-sdk-tools/blob/main/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/GitHubEventClient.cs#L310 that we do not do a retry on that call and do not ignore any failures (arguably we ignore a fail there).

JimSuplizio:
The following updates have been made in this PR.
1. The primary change was to swap out the old CodeownersParser for the CodeownersUtils which will parse the new moniker, AzureSdkOwners, from CODEOWNERS files. 
2. With the updated parsing the InitialIssueTriage rule has been rewritten, updated to definition provided by @jsquire. The exact rule's processing can be found in the updated Rules.md file which part of this PR. Fixes #5743 
4. The ServiceAttention rule now checks every label added. If ServiceAttention is already on the Issue and a service label is added it'll perform the rule's actions. Fixes #6480 
5. The rate limit API call now has retries. Because this is called at the start/end of processing, there is no backoff and the time in between attempts is 200ms, for a total wait time of 800ms over 5 attempts which is because we really can't afford a long delay when processing an Action. Fixes #6524 

As part of these updates, tests had to be added, or in the case of InitialIssueTriage, fully rewritten and new APIs had to be plumbed through the mock. The InitialIssueTriage can now assign someone to an issue which required new API calls. The first is the check to see if the owner has issue assignment permissions and, the second, is doing the assignment which requires the owner/repo name (part of the repository payload). This differs from other APIs that update Issues which only required the repository ID and is the reason for the GitHubIssueAssignment being added to the GitHubEventProcessor class. 

Because InitialIssueTriage will now add ServiceAttention and execute that rule, if certain conditions are met, the RuleEnabled API now has a silent mode. When checking whether not a rule is turned on, the output in the logs indicates that a rule is on and will run and anything until the next rule checkout output line is that rule running. In the case of ServiceAttention being run as part of InitialIssueTriage, it's only run if the rule is enabled for the repository. I didn't want the output going into the logs for the check done as part of the InitialIssueTriage because it's not explicitly running that rule and there's processing for the InitialIssueTriage that happens after that and anything in the output needs to be attributed to the rule that's actually running.

There were some refactors/overloads that had to be written since, ServiceAttention can be run from InitialIssueTriage which means that it's not getting the labels from the payload (a list of Octokit.Labels) but rather a list of string. Things had to handle both and the easiest way to do this was to boil everything down to the common denominator, a list of strings.

### DIFF
--- a/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/Static/IssueProcessingTests.cs
+++ b/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/Static/IssueProcessingTests.cs
@@ -15,7 +15,7 @@ namespace Azure.Sdk.Tools.GitHubEventProcessor.Tests.Static
     [Parallelizable(ParallelScope.Children)]
     public class IssueProcessingTests : ProcessingTestBase
     {
-        // JRS - With the updated Initial Triage rule I need to be able to set the following information in order to test
+        // With the updated Initial Triage rule I need to be able to set the following information in order to test
         // all of the code paths.
         // 1. mockGitHubEventClient.UserHasPermissionsReturn - whether or not the user has write or admin permissions
         // 2. mockGitHubEventClient.IsUserMemberOfOrgReturn - whether or not the user is a member of the Azure org
--- a/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/Static/IssueProcessingTests.cs
+++ b/tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/Static/IssueProcessingTests.cs
@@ -15,74 +15,297 @@ namespace Azure.Sdk.Tools.GitHubEventProcessor.Tests.Static
     [Parallelizable(ParallelScope.Children)]
     public class IssueProcessingTests : ProcessingTestBase
     {
+        // JRS - With the updated Initial Triage rule I need to be able to set the following information in order to test
+        // all of the code paths.
+        // 1. mockGitHubEventClient.UserHasPermissionsReturn - whether or not the user has write or admin permissions
+        // 2. mockGitHubEventClient.IsUserMemberOfOrgReturn - whether or not the user is a member of the Azure org
+        // 3. mockGitHubEventClient.OwnerCanBeAssignedToIssueInRepoReturn - whether or not a given owner can be assigned to an issue
+        // 4. AIReturnLabels - change to string, if null no labels returned otherwise split string into a list
+        // 5. RuleState ruleState - Whether or not the InitialIssueAssignment rule is on or off
+        // 6. RuleState ruleStateServiceAttention - Whether or not the ServiceAttention rule is on or off
+        //
+        // CODEOWNERS changes necessary to test
+        // The CODEOWNERS file needs entries that match the AIReturnLabels, in theory the AI Label Service will return one
+        // ServiceLabel and one other label. If the rule is going to process the Service Label returned needs to match one
+        // entry in the CODEOWNERS file. The AzureSdkOwners for that CODEOWNERS entry will need to have
+        // OwnerCanBeAssignedToIssueInRepoReturn set. Some entries will need AzureSdkOwners with and without assignment permission
+        // and other entries will be without AzureSdkOwners entirely.
+        //
+        // Labels being added (this entire thing only matters if there are no assinees, no labels and the AI Label Service returns
+        // labels)
+        // If there are no valid AzureSdkOwners (meaning there are none or none of the AzureSdkOwners can be assigned to an issue)
+        // AND there are ServiceOwners for the ServiceLabel
+        // AND the ServiceAttention rule is enabled
+        // add ServiceAttention label, process ServiceAttention rule and add NeedsTeamAttention label
+        // ELSE
+        // add NeedsTeamTriage label
+        //
+        // If the creator of the Issue is not a member of the Azure org and does not have write or admin permissions
+        // add CustomerReported and Question labels
+        //
+        // The ServiceAttention just adds a comment @ mentioning everyone on the ServiceOwners list from CODEOWNERS that
+        // matches the ServiceLabel entry returned from the AI Label Service.
+        // 
+
         /// <summary>
-        /// TBD: InitialIssueTriage rule is not yet complete as it requires a CODEOWNERS overhaul that is not complete.
-        /// The payload for this rule to process requires an issues opened event with no assignee and no labels.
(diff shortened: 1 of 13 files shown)


tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/Utils/CodeOwnerUtils.cs
```diff
@@ -97,66 +99,120 @@ public static List<string> GetPRAutoLabelsForFilePaths(IReadOnlyList<Label> prLa
...
         }
 
+        /// <summary>
+        /// Overloaded function to convert the list of Octokit.Labels to a list of string
+        /// </summary>
+        /// <param name="labels">The list of Octokit.Label</param>
+        /// <returns>The string which contains a unique list of individuals/teams with the names formatted to @mention in github</returns>
+        public static List<string> GetServiceOwnersForServiceLabels(IReadOnlyList<Label> labels)
+        {
+            List<string> labelOnlyList = labels.Select(l => l.Name).ToList();
```
jsquire, line 109: Why force `List<T>` for these since you're only doing a `foreach` over them?  Why not accept `IEnumerable` and save the casts/allocations?   This bubbles up to the rule implementation as well,

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/Utils/LabelUtils.cs
```diff
@@ -8,22 +9,31 @@ namespace Azure.Sdk.Tools.GitHubEventProcessor.Utils
...
+        /// </summary>
+        /// <param name="labels">List of labels</param>
+        /// <param name="labelToCheck">The label to look for in the list</param>
+        /// <returns>true if the label exists, false otherwise</returns>
+        public static bool HasLabel(List<string> labels, string labelToCheck)
         {
             if (labels != null)
             {
-                foreach (Label label in labels)
+                if (labels.Contains(labelToCheck, StringComparer.OrdinalIgnoreCase))
```
jsquire, line 34: You could simplify this to:

```csharp
return labels.Contains(labelToCheck, StringComparer.OrdinalIgnoreCase);
```

Not much of a savings, but it allows you to skip the branch condition.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor.Tests/Static/IssueProcessingTests.cs
```diff
@@ -15,74 +15,297 @@ namespace Azure.Sdk.Tools.GitHubEventProcessor.Tests.Static
     [Parallelizable(ParallelScope.Children)]
     public class IssueProcessingTests : ProcessingTestBase
     {
+        // JRS - With the updated Initial Triage rule I need to be able to set the following information in order to test
```
jsquire, line 18: Thank you for this.   Super helpful.

jsquire (approved):

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/Utils/LabelUtils.cs
JimSuplizio: Yes, but the false case is taken care of with the fall through which happens if labels is null.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/EventProcessing/IssueProcessing.cs
```diff
@@ -81,25 +89,140 @@ public static async Task InitialIssueTriage(GitHubEventClient gitHubEventClient,
...
+                                }
+                                // else there are multiple owners and a random one needs to be assigned
+                                else
+                                {
+                                    // Create a list of AzureSdkOwners that has been randomized. The reason
+                                    // the entire list is being randomed is because each person has to be
+                                    // checked to see if they can be assigned to an issue in the repository.
+                                    // and having the entire list being random simplifies processing if a given
+                                    // owner cannot be assigned.
+                                    var rnd = new Random();
```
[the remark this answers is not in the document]
JimSuplizio: @jsquire seeding this doesn't actually matter. The way it's written, right now, it'll completely randomize the list each time it runs.  There is no place to save a seed nor was there any requirement to try and round robin assignment. That can't actually happen with Actions since there's no place to save it and honestly, that would cause trouble than what it's worth. Can we end up with the same person being assigned more than once? Yes, unfortunately, that's the nature of random. 

There's another piece to this is and that's the call to ensure that the owner from the AzureSdkOwners can actually be assigned to an issue within the repository. Even if there were two items in the list, if one of them can't be assigned an issue for that repository and the other can, the other would end up getting all the assignments anyways.

tools/github-event-processor/Azure.Sdk.Tools.GitHubEventProcessor/Utils/CodeOwnerUtils.cs
JimSuplizio: I'd still need to convert the list of labels to a list of string. The AI service returns a list of string, the actions payloads return a list of Label whose label is on the Name field. It was easier to have the overload than to put the conversion inline everywhere.

PR #7251, opened 2023-11-07
Add CODEOWNERS linter yml file to eng/common/pipelines

JimSuplizio:
**codeowners-linter.yml** is going to be the pipeline yml file that'll run the linter. This is a unique pipeline in that it's the only yml file in eng/common that is a fully contained pipeline instead of a template that's included in other pipelines. This needs to be in common because, like github-actions, it needs to set the version of the CodeownersLinter to install in a common place that gets pushed out to the other repositories. 
Because the trigger is set to none, this shouldn't get picked up by pipeline generation which is we want. Once this is pushed out to the other repositories I'll need to do the following:
1. Create their baseline files and get those checked in.
2. Create the pipeline for that repository. The pipeline will run daily and the triggers are setup to run when there are changes to the CODEOWNERS or baseline files as part of PRs. Since the linter requires data that's produced as part of the pipeline-owners-extractor run, the pipelines will be scheduled to run daily _after_ that pipeline runs (which is 4am PST).

CODEOWNERS_baseline_errors.txt - This is baseline of deduped existing errors. The pipeline uses the baseline file to filter out existing errors otherwise it would start out failing. The baseline file is effectively there to prevent further bleeding but wouldn't prevent the same error from being added again (the unfortunate side effect of having to dedupe since we can't actually rely on errors per line number for obvious reasons).

### DIFF
--- a/eng/common/pipelines/codeowners-linter.yml
+++ b/eng/common/pipelines/codeowners-linter.yml
@@ -29,7 +29,7 @@ stages:
       vmImage: ubuntu-22.04
 
     variables:
-      CodeownersLinterVersion: '1.0.0-dev.20231023.2'
+      CodeownersLinterVersion: '1.0.0-dev.20231107.2'
       DotNetDevOpsFeed: "https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json"
       RepoLabelUri: "https://azuresdkartifacts.blob.core.windows.net/azure-sdk-write-teams/repository-labels-blob"
       TeamUserUri: "https://azuresdkartifacts.blob.core.windows.net/azure-sdk-write-teams/azure-sdk-write-teams-blob"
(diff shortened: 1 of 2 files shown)


eng/common/pipelines/codeowners-linter.yml
```diff
@@ -0,0 +1,47 @@
...
+  paths:
+    include:
+      - .github/CODEOWNERS
+      - .github/CODEOWNERS_baseline_errors.txt
+      - eng/common/pipelines/codeowners-linter.yml
+
+stages:
+- stage: Run
+  variables:
+    skipComponentGovernanceDetection: true
```
weshaggard, line 22: This is probably OK for now, but we might need to change this to import the /eng/pipelines/templates/variables/globals.yml in each repo as some other repos need to disable other checks in their repos.

weshaggard (approved):

eng/common/pipelines/codeowners-linter.yml
JimSuplizio: @weshaggard while I agree, not every repository has or needs a globals.yml. azure-sdk and azure-sdk-for-c are both without. It was easier just to remove the skipComponentGovernanceDetection here, for just this yml file since that's the only thing I can guarantee and control.

weshaggard: I suspect you might hit an issue in the .NET repo as we also have to turn off the nuget analysis task. However you could also set that variable in the DevOps UI if that is easier.

PR #7270, opened 2023-11-11
Fix typos in ps1 scripts

ellismg:
During Hacktober, @LukaszKokot discovered these typos is some scripts checked into `azure/azure-dev`.  He raised a PR against us to fix, them but they should actually be fixed here, since they are part of the common engineering system.

I offered to move his patch over here on his behalf (since we dropped the ball a little on getting back to him).

This change brings over what was in https://github.com/Azure/azure-dev/pull/2881, to here, which is where these file ultimately live.


weshaggard:
I'm good with these updates but currently I don't believe azure-dev is part of our eng/common sync'ing pipeline so I suspect that sync will need to be done manually.

weshaggard (approved):

weshaggard:
I'll push this through the sync process.

PR #7287, opened 2023-11-14
Add package-lock.json for tsp-client

hallipr:
tsp-client should use `npm ci` in its CI pipeline to allow for deterministic builds.  For that, we need to:
- update the pipeline yaml to use 
- stop ignoring *-lock.json files in .gitignore
- pin the floating @typespec/compiler peer dependency in devDependencies
- add tsp-client's package-lock.json to the repository

### DIFF
--- a/tools/tsp-client/ci.yml
+++ b/tools/tsp-client/ci.yml
@@ -58,18 +58,16 @@ stages:
             versionSpec: '$(NodeVersion)'
           displayName: 'Install Node.js'
 
-        - bash: |
-            npm install
-          displayName: 'npm install'
+        - script: 'npm ci'
+          displayName: 'npm ci'
           workingDirectory: $(System.DefaultWorkingDirectory)/tools/tsp-client
 
-        - bash: |
-            npm pack
+        - script: 'npm pack'
           displayName: 'npm pack'
           workingDirectory: $(System.DefaultWorkingDirectory)/tools/tsp-client
           condition: contains(variables['OSVmImage'], 'ubuntu')
 
-        - bash: 'cp azure-tools-typespec-client-generator-cli-*.tgz $(VAR_BUILD_ARTIFACT_STAGING_DIRECTORY)'
+        - script: 'cp azure-tools-typespec-client-generator-cli-*.tgz $(VAR_BUILD_ARTIFACT_STAGING_DIRECTORY)'
           displayName: 'copy to staging dir'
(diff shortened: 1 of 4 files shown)


tools/tsp-client/ci.yml
```diff
@@ -58,18 +58,16 @@ stages:
...
 
-        - bash: |
-            npm pack
+        - script: 'npm pack'
           displayName: 'npm pack'
           workingDirectory: $(System.DefaultWorkingDirectory)/tools/tsp-client
           condition: contains(variables['OSVmImage'], 'ubuntu')
 
-        - bash: 'cp azure-tools-typespec-client-generator-cli-*.tgz $(VAR_BUILD_ARTIFACT_STAGING_DIRECTORY)'
+        - script: 'cp azure-tools-typespec-client-generator-cli-*.tgz $(VAR_BUILD_ARTIFACT_STAGING_DIRECTORY)'
```
mikeharder, line 70: I think this will fail on Windows

tools/tsp-client/package.json
```diff
@@ -34,7 +34,8 @@
     "prettier": "^3.0.1",
     "rimraf": "^5.0.1",
     "ts-node": "^10.9.1",
-    "typescript": "^5.1.6"
+    "typescript": "^5.1.6",
+    "@typespec/compiler": "0.48.1"
```
mikeharder, line 38:
```suggestion
    "@typespec/compiler": "0.50.0"
```

Probably want to use latest here?

mikeharder (changes requested):

tools/tsp-client/package.json
hallipr: updated

tools/tsp-client/ci.yml
hallipr: reverted

tools/tsp-client/package.json
```diff
@@ -34,7 +34,8 @@
     "prettier": "^3.0.1",
     "rimraf": "^5.0.1",
     "ts-node": "^10.9.1",
-    "typescript": "^5.1.6"
+    "typescript": "^5.1.6",
+    "@typespec/compiler": "0.50.0"
```
mikeharder, line 38: Who is responsible for updating this pinned version, and when?  The tool uses a floating peerDep at runtime so it doesn't impact users, but it does impact our build and test.  Ideally this would be updated whenever the version is updated in azure-rest-api-specs/main.

mikeharder (approved):

PR #7305, opened 2023-11-16
Set package version in API review code File for Python

praveenkuttappan:
Update Package version for Python and upgrade existing reviews to rebuild code file.

### DIFF
--- a/packages/python-packages/apiview-stub-generator/CHANGELOG.md
+++ b/packages/python-packages/apiview-stub-generator/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Release History
 
-## Version 0.3.9 (TBD)
+## Version 0.3.10 (2023-11-17)
+Update package version in API review codefile
+
+## Version 0.3.9 (2023-11-17)
 Improve support for "CrossLanguageDefinitionId".
 
 ## Version 0.3.8 (2023-07-06)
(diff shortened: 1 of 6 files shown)


packages/python-packages/apiview-stub-generator/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Release History
 
-## Version 0.3.9 (TBD)
+## Version 0.3.10 (2023-11-17)
+Update package version in API review codefile
```
maririos, line 4: More than an update is an addition, no?

packages/python-packages/apiview-stub-generator/apiview_reqs.txt
```diff
@@ -1,2 +1,3 @@
 ./
-pylint-guidelines-checker==0.0.7
\ No newline at end of file
+pylint-guidelines-checker==0.0.7
+pylint<3.0.0
```
maririos, line 3: curious on why this addition is needed?

maririos (approved):

tjprescott (approved):

packages/python-packages/apiview-stub-generator/apiview_reqs.txt
praveenkuttappan: API review parser does not work with latest linter > 3.0 and it was released last month. Scott B has already set this package requirement is python repo when running api review parser. We didn't run into this issue so far since there was no change in parser to test this pipeline which is only triggered when modifying parser itself. We need to ensure pylint less than 3.0 is used to successfully run API parser in test pipeline.

PR #7307, opened 2023-11-16
Remove CodeOwnersParser

JimSuplizio:
CodeOwnersParser is gone, replaced by CodeownersUtils. 

github-event-processor's ci.yml file should have been changed to trigger off codeowners-utils changes earlier this week when it was updated but I'd missed that then. It's a non-issue since there were no codeowners-utils changes.


weshaggard:
We should also remove https://github.com/Azure/azure-sdk-tools/blob/main/.github/CODEOWNERS#L33. If only we had a code-owners linter that could catch this :)

JimSuplizio:
> We should also remove https://github.com/Azure/azure-sdk-tools/blob/main/.github/CODEOWNERS#L33. If only we had a code-owners linter that could catch this :)

@weshaggard, I didn't forget to update the CODEOWNERS in this PR, I was...um...testing the [linter pipeline](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3272338&view=logs&j=a460d732-23aa-5493-a411-cc406067acf8&t=259e4be2-1fd3-5c65-f373-9da11bbaf3b4)...yes, that's it! Seriously though, I'll get the CODEOWNERS updated when I create the PR for the updated codeowners-utils README.

PR #7309, opened 2023-11-16
Run upgrade of API reviews to update package version in cosmos DB

praveenkuttappan:
Run upgrade of API reviews to update package version in cosmos DB

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
+++ b/src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
@@ -352,7 +352,7 @@ namespace APIViewWeb.Managers
             {
                 if (updateDisabledLanguages.Contains(language))
                 {
-                    _telemetryClient.TrackTrace("Background task to update API review at startup is disabled for langauge " + language);
+                    _telemetryClient.TrackTrace("Background task to update API review at startup is disabled for language " + language);
                     continue;
                 }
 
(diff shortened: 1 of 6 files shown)


src/java/apiview-java-processor/pom.xml
```diff
@@ -6,7 +6,7 @@
 
   <groupId>com.azure</groupId>
   <artifactId>apiview-java-processor</artifactId>
-  <version>1.29.0</version>
+  <version>1.29.1</version>
```
maririos, line 9: If this follows our version strategy, I believe this should be `1.30.0` as `x.x.1` suggest there was a bug fix and I don't see anything in this PR that suggest that.

src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
```diff
@@ -352,23 +352,23 @@ public async Task UpdateReviewBackground(HashSet<string> updateDisabledLanguages
...
+                // If review is updated using Azure DevOps pipeline then batch process update review requests
                 if (languageService.IsReviewGenByPipeline)
                 {
                     await UpdateReviewsUsingPipeline(language, languageService, backgroundBatchProcessCount);
                 }
                 else
                 {
                     var reviews = await _reviewsRepository.GetReviewsAsync(false, language, fetchAllPages: true);
-                    foreach (var review in reviews.Where(r => IsUpdateAvailable(r)))
+                    foreach (var review in reviews.Where(r => IsUpdateAvailable(r) || r.Revisions.Any(rev => string.IsNullOrEmpty(rev.SingleFile?.PackageVersion))))
```
maririos, line 371: we want to update the review if we got the package version, right?

praveenkuttappan: This check is to find the eligible reviews to rebuild. We will rebuild review data if either parser version is different or if package version is missing in review metadata.

src/java/apiview-java-processor/pom.xml
praveenkuttappan: Done.

maririos (approved):

PR #7310, opened 2023-11-16
Force capitalize all deployment/env outputs in live test

benbp:


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -290,7 +290,14 @@ function BuildDeploymentOutputs([string]$serviceName, [object]$azContext, [objec
         }
     }
 
-    return $deploymentOutputs
+    # Force capitalize all keys to avoid azure pipelines confusion with
```
heaths, line 293: Nit:
```suggestion
    # Force capitalization of all keys to avoid Azure Pipelines confusion with
```

Or something like that.

heaths (approved):

PR #7322, opened 2023-11-19
Create new pipeline to create package work item

praveenkuttappan:
Create a new pipeline to create package work item and this can be consumed by power apps to create a package work item from release planner.


eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -348,18 +349,29 @@ function CreateWorkItem($title, $type, $iteration, $area, $fields, $assignedTo,
...
-    Invoke-AzBoardsCmd "work-item relation add" $parameters $outputCommand | Out-Null
+  if ($parentId)
+  {
+    CreateWorkItemRelation $workItem.id $parentId "parent" outputCommand
+  }
+  
+  # Add a work item as related if given.
+  if ($relatedId)
+  {
+    CreateWorkItemRelation $$workItem.id $relatedId "Related" outputCommand
```
weshaggard, line 360:
```suggestion
    CreateWorkItemRelation $workItem.id $relatedId "Related" $outputCommand
```

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -348,18 +349,29 @@ function CreateWorkItem($title, $type, $iteration, $area, $fields, $assignedTo,
...
-  if ($parentId) {
-    $parameters = $ReleaseDevOpsCommonParameters
-    $parameters += "--id", $workItem.id
-    $parameters += "--relation-type", "parent"
-    $parameters += "--target-id", $parentId
-
-    Invoke-AzBoardsCmd "work-item relation add" $parameters $outputCommand | Out-Null
+  if ($parentId)
+  {
+    CreateWorkItemRelation $workItem.id $parentId "parent" outputCommand
```
weshaggard, line 354:
```suggestion
    CreateWorkItemRelation $workItem.id $parentId "parent" $outputCommand
```

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -348,18 +349,29 @@ function CreateWorkItem($title, $type, $iteration, $area, $fields, $assignedTo,
...
   $workItem = Invoke-AzBoardsCmd "work-item create" $parameters $outputCommand
 
-  if ($parentId) {
-    $parameters = $ReleaseDevOpsCommonParameters
-    $parameters += "--id", $workItem.id
-    $parameters += "--relation-type", "parent"
-    $parameters += "--target-id", $parentId
-
-    Invoke-AzBoardsCmd "work-item relation add" $parameters $outputCommand | Out-Null
+  if ($parentId)
```
weshaggard, line 352: Should these 2 be mutually exclusive?

eng/pipelines/devops-create-package-workitem.yml
```diff
@@ -0,0 +1,68 @@
...
+steps:
+- checkout: self
+
+- task: PowerShell@2
+  displayName: Create Package Work Item
+  inputs:
+    pwsh: true
+    filePath: $(AzureSDKToolsScriptsPath)/Update-DevOps-Release-WorkItem
+    arguments: >
+      -language '$(Language)'
```
weshaggard, line 60: Don't you need to access these via `{ parameters.Language }}`

praveenkuttappan: You are right. I think I used variable way of accessing them. thanks

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
praveenkuttappan: Package work item will have product work item as `parent` and release plan work item as `related`

eng/common/scripts/Update-DevOps-Release-WorkItem.ps1
```diff
@@ -69,8 +71,12 @@ $plannedVersions = @(
     Date = $plannedDate
   }
 )
+$ignoreReleasePlannerTests = $true
+if ($tag -eq "Release Planner App Test") {
```
maririos, line 75: tag could contain multiple values. would it be better to do contains instead of eq?

praveenkuttappan: When calling from release planner app test, I believe it will only be one tag that's why I checked for eq.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -154,6 +154,9 @@ function FindParentWorkItem($serviceName, $packageDisplayName, $outputCommand =
   else {
     $serviceCondition = "[ServiceName] <> ''"
   }
+  if ($tag) {
```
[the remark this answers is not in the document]
praveenkuttappan: There are EPIC work items created from release planner test apps. Another EPIC work item exists without release planner app test tag when created from product env or by prepare release script. We don't want to link package item from Test instance to parent work item from production. So checking to ensure correct parent is selected who has `Release Planner  App Test` tag.

eng/common/scripts/Update-DevOps-Release-WorkItem.ps1
praveenkuttappan: Updated to use `Contains'

maririos: For nwo that is true. I guess if it ever changes we will just have to update it here too

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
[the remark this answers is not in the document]
maririos: I'm still confused :( isn't that the purpose of `ignoreReleasePlannerTests` ?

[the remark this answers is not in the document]
praveenkuttappan: IgnoreReleasePlannerTest is to either ignore or include work items with release planner test. But there is no way to use that flag to ensure we lookup parent with release planner test tag only. It's possible to find a parent created in production environment and attach the test package item even though it's triggered from test instance unless we explicitly search for a work item with same tag at parent level also.

maririos (approved):

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -341,25 +348,48 @@ function CreateWorkItem($title, $type, $iteration, $area, $fields, $assignedTo,
   if ($assignedTo) {
     $parameters += "--assigned-to", "`"${assignedTo}`""
   }
+  if ($tag)
+  {
+    If ($fields)
```
weshaggard, line 353:
```suggestion
    if ($fields)
```

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -185,10 +188,10 @@ function FindParentWorkItem($serviceName, $packageDisplayName, $outputCommand =
 $packageWorkItems = @{}
 $packageWorkItemWithoutKeyFields = @{}
 
-function FindLatestPackageWorkItem($lang, $packageName, $outputCommand = $true, $ignoreReleasePlannerTests = $true)
+function FindLatestPackageWorkItem($lang, $packageName, $outputCommand = $true, $ignoreReleasePlannerTests = $true, $tag = $null)
 {
   # Cache all the versions of this package and language work items
-  $null = FindPackageWorkItem $lang $packageName -includeClosed $true -outputCommand $outputCommand -ignoreReleasePlannerTests $ignoreReleasePlannerTests
+  $null = FindPackageWorkItem $lang $packageName -includeClosed $true -outputCommand $outputCommand -ignoreReleasePlannerTests $ignoreReleasePlannerTests -tag $tag
```
weshaggard, line 194: Why are you passing ignoreReleasePlannerTests twice? I'm still not a big fan of this flag.

praveenkuttappan: I didn't understand the comment about duplicate flag. ignoreReleasePlannerTests is passed to either ignore or consider the work item created by release planner test instances along with work items created by production systems (work items without release planner app test). Tag param is passed only when we need to fetch a work item with a given tag, so it won't consider a work item created by production system. Otherwise, we might end up linking a package work item from test incorrectly under a product work item created from production.

weshaggard: Nevermind, I missed read this I thought you passed `$ignoreReleasePlannerTests` twice but one is just the parameter name `-ignoreReleasePlannerTests`

eng/common/scripts/Create-Package-WorkItem.ps1
```diff
@@ -0,0 +1,47 @@
+
```
weshaggard, line 1: What is this script doing for us? It looks like it is pretty much just calling the other script

praveenkuttappan: I have removed this script now. It was added to try out az devops login, but it's not required anymore.

weshaggard (approved):

PR #7334, opened 2023-11-23
Fix typo

antkmsft:


weshaggard:
I don't mind taking the fixes but I believe these scripts will end up being removed as we move to tsp-client tool. @raych1 is going to start driving that transition.

konrad-jamrozik (approved):

raych1 (approved):

PR #7338, opened 2023-11-27
APIView - Update package name format for C

praveenkuttappan:
Update Package name to use `-` instead of `::` as scope separator.


src/dotnet/APIView/APIViewWeb/Languages/CLanguageService.cs
```diff
@@ -16,7 +16,7 @@ namespace APIViewWeb
 {
     public class CLanguageService : LanguageService
     {
-        private const string CurrentVersion = "4";
+        private const string CurrentVersion = "5";
```
maririos, line 19: is this the version of the parser? is it consumed anywhere?

maririos (approved):

PR #7346, opened 2023-11-28
Update Test-SampleMetadata.ps1 add azure-ai-contentsafety

zhaiyutong:


eng/common/scripts/Test-SampleMetadata.ps1
```diff
@@ -104,6 +104,7 @@ begin {
         "azure-active-directory-b2c",
         "azure-active-directory-domain",
         "azure-advisor",
+        "azure-ai-contentsafety",
```
weshaggard, line 107: I don't see this one in https://review.learn.microsoft.com/en-us/help/platform/metadata-taxonomies?branch=main#product so I don't expect the docs system to understand it yet.

zhaiyutong: We have name azure-ai-content-safety
But is should be updated to azure-ai-contentsafety
@catalinaperalta

weshaggard (approved):

PR #7371, opened 2023-11-30
[Pylint] Update contribution guideline

l0lawrence:


konrad-jamrozik (approved):

scbedd (changes requested):
This is your area for sure, but I don't think we should check in empty sections. Probably better to just not have the content rather than draw attention to the fact that it's missing😂

PR #7373, opened 2023-11-30
Add mirror container image pipeline

benbp:
Pipeline at https://dev.azure.com/azure-sdk/internal/_build/index?definitionId=6572

### DIFF
--- a/eng/pipelines/mirror-container-images.yml
+++ b/eng/pipelines/mirror-container-images.yml
@@ -1,0 +1,39 @@
+trigger: none
+pr: none
+
+parameters:
+  - name: Images
+    type: object
+    default:
+      - source: ghcr.io/chaos-mesh/chaos-daemon:v2.1.4
+        mirror: azsdkengsys.azurecr.io/mirror/chaos-mesh/chaos-daemon:v2.1.4
+      - source: ghcr.io/chaos-mesh/chaos-mesh:v2.1.4
+        mirror: azsdkengsys.azurecr.io/mirror/chaos-mesh/chaos-mesh:v2.1.4
+      - source: ghcr.io/chaos-mesh/chaos-dashboard:v2.1.4
+        mirror: azsdkengsys.azurecr.io/mirror/chaos-mesh/chaos-dashboard:v2.1.4
+      - source: ubuntu/squid
+        mirror: azsdkengsys.azurecr.io/mirror/ubuntu/squid
+
+jobs:
+  - job: MirrorImages
+    displayName: Mirror Container Images
+
+    pool:
+      name: azsdk-pool-mms-ubuntu-2004-general
+      vmImage: MMSUbuntu20.04
+
+    steps:
+      - ${{ each image in parameters.Images }}:
+        - task: Docker@2
+          displayName: Login to ${{ split(image.mirror, '.')[0] }}
+          inputs:
+            command: login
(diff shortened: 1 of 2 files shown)


eng/pipelines/mirror-container-images.yml
```diff
@@ -0,0 +1,39 @@
...
+  - name: Images
+    type: object
+    default:
+      - source: ghcr.io/chaos-mesh/chaos-daemon:v2.1.4
+        mirror: azsdkengsys.azurecr.io/mirror/chaos-mesh/chaos-daemon:v2.1.4
+      - source: ghcr.io/chaos-mesh/chaos-mesh:v2.1.4
+        mirror: azsdkengsys.azurecr.io/mirror/chaos-mesh/chaos-mesh:v2.1.4
+      - source: ghcr.io/chaos-mesh/chaos-dashboard:v2.1.4
+        mirror: azsdkengsys.azurecr.io/mirror/chaos-mesh/chaos-dashboard:v2.1.4
+      - source: ubuntu/squid
```
benbp, line 14: Used by c++, an alternative is we could put one of these pipelines in each language repository.

eng/scripts/mirror-container-image.ps1
```diff
@@ -0,0 +1,23 @@
...
+    Write-Host "Logging in to $mirrorRegistry"
+    az acr login -n $mirrorRegistry
+}
+
+Write-Host "docker pull $Image"
+docker pull $Image
+Write-Host "docker tag $Image $Mirror"
+docker tag $Image $Mirror
+Write-Host "docker push $Mirror"
+docker push $Mirror
```
weshaggard, line 23: How likely would these commands fail? Should we check for exit codes so that we get a failing mirroring pipeline if they do fail?

eng/pipelines/mirror-container-images.yml
```diff
@@ -0,0 +1,39 @@
...
+      name: azsdk-pool-mms-ubuntu-2004-general
+      vmImage: MMSUbuntu20.04
+
+    steps:
+      - ${{ each image in parameters.Images }}:
+        - task: Docker@2
+          displayName: Login to ${{ image.mirror }} ACR
+          inputs:
+            command: login
+            containerRegistry: ${{ split(image.mirror, '.')[0] }}
```
weshaggard, line 31: Does this work in the devops templating engine?

eng/scripts/mirror-container-image.ps1
benbp: I set the `$PSNativeCommandUseErrorActionPreference` above so it will exit out, now that the variable ships with stock powershell.

eng/pipelines/mirror-container-images.yml
benbp: Yep! Patrick tried it out on something else a few weeks ago which is how I discovered it.

benbp: Or...wait it doesn't. But I guess login still worked. Let me iterate on it.

benbp: I guess the login command seems to trim under the hood anyway, so I will change it to just pass `image.mirror`.

benbp: Oh! No it does work, but I was logging the old value. Let me fix the display name log to also trim.

weshaggard (approved):

PR #7382, opened 2023-12-01
[js-api-parser] don't add members as types

jeremymeng:
This PR uses the id of the token to determine whether it should be added as a member or a type.

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Languages/JavaScriptLanguageService.cs
+++ b/src/dotnet/APIView/APIViewWeb/Languages/JavaScriptLanguageService.cs
@@ -11,7 +11,7 @@ namespace APIViewWeb
         public override string Name { get; } = "JavaScript";
         public override string[] Extensions { get; } = { ".api.json" };
         public override string ProcessName { get; } = "node";
-        public override string VersionString { get; } = "1.0.7";
+        public override string VersionString { get; } = "1.0.8";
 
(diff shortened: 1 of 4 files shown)


jeremymeng:
before: 
image
after: 
image

jeremymeng:
Would this cause breaking changes to show up everywhere?

deyaaeldeen (approved):
LGTM, I am surprised there is no field that indicates membership

jeremymeng:
> Would this cause breaking changes to show up everywhere?

@praveenkuttappan @chidozieononiwu?

praveenkuttappan:
> > Would this cause breaking changes to show up everywhere?
> 
> @praveenkuttappan @chidozieononiwu?

Yes, unless we update the parser version and get all existing reviews updated. Can you please update JS parser version in following two locations.
https://github.com/Azure/azure-sdk-tools/blob/main/tools/apiview/parsers/js-api-parser/package.json#L3
https://github.com/Azure/azure-sdk-tools/blob/main/src/dotnet/APIView/APIViewWeb/Languages/JavaScriptLanguageService.cs#L14

This will upgrade all existing reviews using latest parser.

jeremymeng:
@praveenkuttappan updated version number.  Looking good now?

praveenkuttappan (approved):
Looks good

PR #7412, opened 2023-12-07
Fix null error

chidozieononiwu:


scbedd (approved):
This looks like a safe change to me!

PR #7417, opened 2023-12-08
Ensure `HeaderSanitizer`s apply to list-value headers

scbedd:
This was discovered by @mssfang , @billwert , and @samvaity . Example of failure:

```json
"headers": {
  "Cookie": [
    "val1",
    "val2"
  ]
}
```

We weren't sanitizing val2 in the above construct.

This PR fixes that.


billwert (approved):
Thanks for the quick turnaround!

scbedd:
> Thanks for the quick turnaround!

It's much simpler when it's something that's pretty solid to test for :D

PR #7424, opened 2023-12-09
Collect and combine inventory across agents

heaths:


heaths:
This publishes a single _inventory.csv_ artifact (well, and the ones it stitches together) that is sorted in case we wanted to check it into the repo or something.

MacOS also needs more inventory. Maybe we should run `brew` for both Mac and Ubuntu? I actually don't use Mac much so I don't know what's common, but can do some digging. Can be added later, though.

Ideally, I'd like to be able to not even show the non-executed tasks, but after a lot of trial and error, and research, it seems you can't use any sort of condition that would allow me to use template expressions or the like. I mean, there's hacks like generate a matrix and actually use a subordinate template, but that seems like overkill for something we run rarely.

weshaggard (approved):

PR #7438, opened 2023-12-12
Process swagger file cross reference when generating API review

praveenkuttappan:
Swagger file refers to a definition in another swagger file and sometimes this definition in second swagger file might have a schema reference to another definition within same file and this reference will be a local reference. API review parser incorrectly searches for this local reference in first swagger file instead of second file where it's defined.

PR has the fix to correctly refer the swagger file based on the parent reference if current reference path is local.

For e.g. data.json has a reference to 202response definition which is in common.json and 202response has a reference to longrunningoperationresult schema using local reference `#LongRunningOperationResult` and it's defined in common.json. We need to make sure parser refers common.json when processing `#LongRunningOperationResult` schema instead of data.json.

### DIFF
--- a/tools/apiview/parsers/swagger-api-parser/SwaggerApiParser/CHANGELOG
+++ b/tools/apiview/parsers/swagger-api-parser/SwaggerApiParser/CHANGELOG
@@ -1,5 +1,8 @@
 # Release History
 
+## 1.3.0 (12-12-2023)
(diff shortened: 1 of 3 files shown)


maririos (approved):

maririos:
Are there any test we can add for this?

PR #7440, opened 2023-12-12
Only trigger PR build on inventory for itself

heaths:


benbp (approved):

eng/pipelines/agent-software-inventory.yml
```diff
@@ -1,6 +1,9 @@
 # Inventory PowerShell modules installed
 
 trigger: none
+pr:
```
weshaggard, line 4: Based on my settings in the DevOps UI this will not work as it is configured to require a comment. So this shouldn't be needed.

heaths: Sure, but wouldn't it be better to:

1. Maintain more of the trigger in code, and/or
2. Not even spend the cycles determining if this pipeline is needed?

Won't it still show up in the list, just not run? Could be confusing.

weshaggard: I'm good with switching to this model if you want. Just be sure to update the DevOps settings too.

PR #7445, opened 2023-12-12
Add package filter parameter for release

benbp:
By default we recursively discover packages in our release scripts. This is causing issues in the go repo where some packages have packages nested but that are represented in separate pipelines, namely `azidentity`. This PR adds a `-NoRecurse` option to the scripts and pipeline templates which changes our search depth to the artifact location only.

```
<package>/
  go.mod
  ci.yml
  CHANGELOG.md
  <sub package>/
    go.mod
    ci.yml
    CHANGELOG.md
```

Resolves https://github.com/Azure/azure-sdk-for-go/issues/21737

### DIFF
--- a/eng/common/scripts/create-tags-and-git-release.ps1
+++ b/eng/common/scripts/create-tags-and-git-release.ps1
@@ -7,6 +7,7 @@ param (
   $artifactLocation, # the root of the artifact folder. DevOps $(System.ArtifactsDirectory)
   $workingDirectory, # directory that package artifacts will be extracted into for examination (if necessary)
   $packageRepository, # used to indicate destination against which we will check the existing version.
+  $packageFilter,
   # valid options: PyPI, Nuget, NPM, Maven, C, CPP
   # used by CreateTags
   $releaseSha, # the SHA for the artifacts. DevOps: $(Release.Artifacts.<artifactAlias>.SourceVersion) or $(Build.SourceVersion)
@@ -25,6 +26,11 @@ Write-Host "Using API URL $apiUrl"
 
 # VERIFY PACKAGES
 $pkgList = VerifyPackages -artifactLocation $artifactLocation -workingDirectory $workingDirectory -apiUrl $apiUrl -releaseSha $releaseSha -continueOnError $continueOnError
+
+if ($packageFilter) {
+    Write-Host "Filtering discovered packages to '$packageFilter'"
+    [array]$pkgList = $pkgList | Where-Object { $_.PackageId -like $packageFilter }
+}
 
 if ($pkgList) {
   Write-Host "Given the visible artifacts, github releases will be created for the following:"
--- a/eng/common/pipelines/templates/steps/create-tags-and-git-release.yml
+++ b/eng/common/pipelines/templates/steps/create-tags-and-git-release.yml
@@ -1,6 +1,7 @@
 parameters:
   ArtifactLocation: 'not-specified'
   PackageRepository: 'not-specified'
+  PackageFilter: ''
   ReleaseSha: 'not-specified'
   RepoId: $(Build.Repository.Name)
   WorkingDirectory: ''
@@ -14,6 +15,7 @@ steps:
     arguments: >
       -artifactLocation ${{ parameters.ArtifactLocation }}
       -packageRepository ${{ parameters.PackageRepository }}
+      -packageFilter ${{ parameters.PackageFilter }}
       -releaseSha ${{ parameters.ReleaseSha }}
       -repoId ${{ parameters.RepoId }}
       -workingDirectory '${{ parameters.WorkingDirectory }}'
--- a/eng/common/pipelines/templates/steps/create-tags-and-git-release.yml
+++ b/eng/common/pipelines/templates/steps/create-tags-and-git-release.yml
@@ -15,7 +15,7 @@ steps:
     arguments: >
       -artifactLocation ${{ parameters.ArtifactLocation }}
       -packageRepository ${{ parameters.PackageRepository }}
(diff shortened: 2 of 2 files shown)


eng/common/scripts/artifact-metadata-parsing.ps1
```diff
@@ -101,11 +101,16 @@ function RetrieveReleaseTag($artifactLocation, $continueOnError = $true) {
   }
 }
 
-function RetrievePackages($artifactLocation) {
-  $pkgs = Get-ChildItem -Path $artifactLocation -Include $packagePattern -Recurse -File
+function RetrievePackages($artifactLocation, [switch]$NoRecurse) {
```
weshaggard, line 104: Can you remind me which part of our release steps use this method? Vs the discovery in the language settings?

eng/common/scripts/artifact-metadata-parsing.ps1
```diff
@@ -101,11 +101,16 @@ function RetrieveReleaseTag($artifactLocation, $continueOnError = $true) {
...
+  $pkgs = @()
+  if ($NoRecurse) {
+    $pkgs = Get-ChildItem -Path $artifactLocation -Include $packagePattern -Recurse -File -Depth 0
+  } else {
+    $pkgs = Get-ChildItem -Path $artifactLocation -Include $packagePattern -Recurse -File
+  }
   if ($GetPackageInfoFromPackageFileFn -and (Test-Path "Function:$GetPackageInfoFromPackageFileFn"))
   {
-    return $pkgs, $GetPackageInfoFromPackageFileFn
+    return @($pkgs), $GetPackageInfoFromPackageFileFn
```
weshaggard, line 113: Can we not handle this inside of the language specific function when we call this?

eng/common/scripts/artifact-metadata-parsing.ps1
benbp: The origin is from https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/create-tags-and-git-release.ps1#L27 which is the core script that go uses to release.

eng/common/scripts/artifact-metadata-parsing.ps1
benbp: In the course of development I found a bug where if `$pkgs` is an empty array, powershell just ends up returning `$GetPackageInfoFromPackageFileFn` as a single value. So no matter what the caller is doing, the destructuring gets broken. This fixes that edge case issue (which was probably hidden since we have never failed to find a package before).

benbp: So it's not a casting/type issue, but a return value issue.

weshaggard: Sorry I wasn't referring to the `@(...)` part that I'm fine with I was more interested in the recurse option. We essentially call the language specific function with the set of packages can we not filter in there instead of not passing the recurse option?

benbp: I don't think this is possible without some other refactoring that will still touch eng/common (happy to brainstorm, I'm not excited about this `NoRecurse` option anyway). The code paths here are:

```
- eng/common/scripts/create-tags-and-git-release.ps1
    - eng/common/scripts/artifact-metadata-parsing.ps1
        - <go repo>/eng/scripts/Language-Settings.ps1
```

When we do call into the language specific function, it's at https://github.com/Azure/azure-sdk-for-go/blob/main/eng/scripts/Language-Settings.ps1#L69-L72 for the `Get-<lang>-PackageInfoFromPackageFile` function, from https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/artifact-metadata-parsing.ps1#L125. The only relevant context given to this function, in go's case at least, is a `$pkg` object which is just a file type that points to the `go.mod` file. So we have no way, in this context, of doing logic to compare the go.mod path against the original service directory unless we plumb it through from the top level tag/git script like I've done for `NoRecurse`. It might be a better approach, but it has the downside of introducing optional parameters to how we call the language functions, and I'm not sure introducing more variability into the signature for an untyped, dynamically loaded function is a good idea.

weshaggard: Do we ever care about this nested package? As in do we ever want to ship it? If not, could we fix this by changing the go.mod or marking it in some way as ignore?

benbp: We do and it has its own pipeline as well, it's just the parent package/pipeline will also ship it if it's unreleased.

benbp: I can see if a flattening of the directory structure would be possible, but my bet is quite a few things depend on it.

eng/common/scripts/create-tags-and-git-release.ps1
```diff
@@ -26,6 +27,11 @@ Write-Host "Using API URL $apiUrl"
 # VERIFY PACKAGES
 $pkgList = VerifyPackages -artifactLocation $artifactLocation -workingDirectory $workingDirectory -apiUrl $apiUrl -releaseSha $releaseSha -continueOnError $continueOnError
 
+if ($packageFilter) {
+    Write-Host "Filtering discovered packages to '$packageFilter'"
+    [array]$pkgList = $pkgList | Where-Object { $_.PackageId -like $packageFilter }
```
weshaggard, line 32:
```suggestion
    $pkgList = @($pkgList | Where-Object { $_.PackageId -like $packageFilter })
```

weshaggard (approved):

PR #7448, opened 2023-12-13
Automatic approval copy issue

praveenkuttappan:
Copy approval name when approval status is copied over.


maririos (approved):

chidozieononiwu (commented):
Looks fine. But is this necessary. Isn't it sufficient to just know via the note that the approval was copied over from another revisions? Since the user did not actually approve the new revisions maybe the approver should just be "azure-sdk" with the information showing the revision it was copied from.

praveenkuttappan:
> Looks fine. But is this necessary. Isn't it sufficient to just know via the note that the approval was copied over from another revisions? Since the user did not actually approve the new revisions maybe the approver should just be "azure-sdk" with the information showing the revision it was copied from.

We need to do this until we show an option in UI to show where the approval is copied from and a quick option to navigate to that review.  It's misleading to show "azure-sdk" as approver.

PR #7450, opened 2023-12-13
Fix automatic review exception

praveenkuttappan:
Handle null pointer exception to avoid invalid exceptions when processing automatic API reviews. Too many exceptions logged in app insights, makes the monitoring difficult.

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
+++ b/src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
@@ -120,7 +120,7 @@ namespace APIViewWeb.Controllers
 
             if (codeFile == null)
             {
-                return StatusCode(statusCode: StatusCodes.Status404NotFound, $"API review code file for package {packageName} is not found in DevOps pipeline artifacts.");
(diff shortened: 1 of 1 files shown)


src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
```diff
@@ -118,6 +118,10 @@ string project
                 packageName: packageName, originalFileName: originalFilePath, codeFileName: reviewFilePath, originalFileStream: memoryStream,
                 project: project);
 
+            if (codeFile == null)
+            {
+                return StatusCode(statusCode: StatusCodes.Status204NoContent, $"API review code file for package {packageName} is not found in DevOps pipeline artifacts.");
```
maririos, line 123: there is another exception on line 132, might be good to add a message to better know where it is coming from (or why)

maririos (approved):

src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
praveenkuttappan: I will check and submit another.

PR #7451, opened 2023-12-14
New validation for REST API spec location before SDK release

raych1:
We only support the SDK release for the spec checked into 'main' branch of Azure/azure-rest-api-specs repository.

- Added common script for validation on `tsp-location.yaml` (typespec case) and validation on `autorest.md` (swagger case)
- Added common pipeline

@weshaggard can you review?


eng/common/pipelines/templates/steps/verify-restapi-spec-location.yml
```diff
@@ -0,0 +1,19 @@
...
+- name: PackageName
+  type: string
+  default: ''
+
+steps:
+  - task: Powershell@2
+    inputs:
+      filePath: $(Build.SourcesDirectory)/eng/common/scripts/Verify-RestApiSpecLocation.ps1
+      arguments: >
+        -ServiceDirectory '${{ coalesce(parameters.ServiceDirectory, parameters.ServiceName) }}'
```
weshaggard, line 14:
```suggestion
        -ServiceDirectory '${{ parameters.ServiceDirectory }}'
```

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,151 @@
...
+# This function is used to verify the specific 'commit' belongs to the main branch of Azure/azure-rest-api-specs repository
+function Verify-CommitFromMainBranch([string]$commit) {
+  $mainBranch = "main"
+  $specRepoCloneDir = "./tmp_spec_repo"
+  New-Item $specRepoCloneDir -Type Directory -Force | Out-Null
+  Push-Location $specRepoCloneDir
+
+  try {
+    $repoRemoteUrl = Get-GitRemoteValue "Azure/azure-rest-api-specs"
+    git clone -b main $repoRemoteUrl .
```
weshaggard, line 56: Cloning down the specs repo in this check seems pretty heavy handed. Is there a github api we can call to check if a it is part of main? Or perhaps we could use the sparse clone from the generation command? We could even potentially add this as a check in the generate command.

weshaggard: Perhaps you could use the search api as outlined in https://stackoverflow.com/questions/30376787/github-api-check-if-a-branch-or-repository-contains-a-commit

weshaggard: Example:
`https://api.github.com/search/commits?q=repo:Azure/azure-sdk-tools+hash:dff04c39ef72f91c27cb10f164b10b6bf958edc9`

raych1: Right, I also thought that git clone is a little bit heavier. I looked at this search api previously but hesitated to use it concerning the paging behavior. I think we will not hit the paging case for current use case.

eng/common/pipelines/templates/steps/verify-restapi-spec-location.yml
```diff
@@ -0,0 +1,23 @@
...
+    inputs:
+      filePath: $(Build.SourcesDirectory)/eng/common/scripts/Verify-RestApiSpecLocation.ps1
+      arguments: >
+        -ServiceDirectory "${{ parameters.ServiceDirectory }}"
+        -PackageName "${{ parameters.PackageName }}"
+        -GitHubPat "${{  parameters.GitHubSpecRepoPat }}"
+      pwsh: true
+      workingDirectory: $(Pipeline.Workspace)
+    displayName: Verify REST API spec location for "${{ parameters.PackageName }}"
+    continueOnError: false
```
weshaggard, line 23: You can remove this given fals is the default.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  # check the commit hash belongs to main branch
+  Verify-CommitFromMainBranch $commit
+}
+
+# This function is used to verify the specific 'commit' belongs to the main branch of Azure/azure-rest-api-specs repository
+function Verify-CommitFromMainBranch([string]$commit) {
+  try {
+    $searchResult = Search-GitHubCommit -AuthToken $GitHubPat -CommitHash "0f39a2d56070d2bc4251494525cb8af88583a938" -RepoOwner "Azure" -RepoName "azure-rest-api-specs"
```
weshaggard, line 54: Is the PAT needed here? Are we worried about rate limiting?

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,121 @@
...
+  if (Test-Path -Path $tspLocationYamlPath) {
+    # typespec scenario
+    $tspLocationYaml = Get-Content -Path $tspLocationYamlPath -Raw | ConvertFrom-Yaml
+    Verify-TspLocation $tspLocationYaml
+  }
+  elseif ($Language -eq "dotnet") {
+    # only dotnet language sdk uses 'autorest.md' to configure the sdk generation, ignore this validation for the other languages
+    if (Test-Path -Path $autorestMdPath) {
+      $autorestMdContent = Get-Content -Path $autorestMdPath -Raw
+      $yamlContent = $autorestMdContent -split '``` yaml|```'
```
weshaggard, line 87: Do we need to worry about multiple yaml secions?

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,121 @@
...
+          }
+        }
+        else {
+          LogWarning "Both 'require' and 'input-file' haven't been found in $autorestMdPath."
+          exit 0
+        }
+      }
+    }
+    else {
+      LogWarning "autorest.md hasn't been found in $PackageDirectory for language: $Language"
```
weshaggard, line 110: How do we not show these warnings for any project that isn't generated? We need to be careful to not start causing a bunch of noise in the builds.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: Yes, we take the first element of it if having multiple sections. autorest.md doesn't have a title for this yaml section.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: I prefer to provide the PAT given the rate limit is three times difference. 10 calls vs 30 calls.  I would like to avoid failing the pipeline by this rate limit issue.

weshaggard, line 54: Why do we have a hard-coded commit hash here?

konrad-jamrozik: If we keep it, I would appreciate adding in a comment an URL pointing to this commit in the Web UI.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: Iterate all yaml sections now.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: Removed.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,182 @@
...
+      try {
+        $autorestMdContent = Get-Content -Path $autorestMdPath -Raw
+        Verify-Content $autorestMdContent
+      }
+      catch {
+        Write-Host "ServiceDir:$ServiceDirectory, PackageName:$PackageName. Failed to parse autorest.md file with exception:`n$_ "
+      }
+    }
+  }
+  elseif ($Language -eq "java" -or $Language -eq "js") {
```
lirenhe, line 167: Is the handling of JS the same with Java?
I found JS is using _meta.json file  [sdk/agrifood/arm-agrifood/_meta.json](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/agrifood/arm-agrifood/_meta.json)

raych1: Thanks @lirenhe for the reminder. The config format appears to be different from `readme.md` and `autorest.md`.

@qiaozha , can you confirm if `_meta.json` is the configuration file of the spec for JS management plane SDK?

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -18,19 +18,33 @@ param (
 Install-ModuleIfNotInstalled "powershell-yaml" "0.4.1" | Import-Module
 
 # This function is used to verify the 'require' and 'input-file' settings in autorest.md point to the main branch of Azure/azure-rest-api-specs repository
+# input-file may be: 
+# https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/purview/data-plane/Azure.Analytics.Purview.MetadataPolicies/preview/2021-07-01-preview/purviewMetadataPolicy.json
+# or 
+# https://github.com/Azure/azure-rest-api-specs/blob/0ebd4949e8e1cd9537ca5a07384c7661162cc7a6/specification/purview/data-plane/Azure.Analytics.Purview.Account/preview/2019-11-01-preview/account.json
 function Verify-AutorestMd([string]$fileUrl) {
-  if($fileUrl -match "^https://github.com/(?<repo>[^/]*/azure-rest-api-specs)/blob/(?<commit>[0-9a-f]{40})/(?<path>.*)") {
+  if($fileUrl -match "^https://raw.githubusercontent.com/(?<repo>[^/]*/azure-rest-api-specs)/(?<commit>[^\/]+(\/[^\/]+)*|[0-9a-f]{40})/(?<path>specification/.*)") {
```
weshaggard, line 26: You should be able to make these into one regex instead of needing to split this into multiple conditions. 
```
^https://(raw.githubusercontent.com|github.com)/(?<repo>[^/]*/azure-rest-api-specs)/(blob|tree)?/(?<commit>[^\/]+(\/[^\/]+)*|[0-9a-f]{40})/(?<path>specification/.*)
```

weshaggard (commented):
This looks reasonable to me I'll leave it to Renhe to sign-off as he will have more knowledge about the content of these files.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
qiaozha: we store the commit id for swagger in `_meta.json` , which is for regeneration for codegen change only, if it's for pipeline, I think we should always use the lastest commit id ?

raych1: @qiaozha 
Do you mean `_meta.json` is only used for re-generation when codegen changes?  
I would like to know is there any config of spec location used in JS SDK release phase? That would be the content of this validation to check.  
Or, does the JS generation script pull the spec from latest main branch of Azure/azure-rest-api-specs repo by default for release? If that's the case, then we can ignore this validation for JS mgmt. plane scenario.

raych1: @lirenhe , for Java/JS/Go mgmt. plane scenarios, they will always get the latest main branch of Azure/azure-rest-api-specs repo to release the SDK. I will ignore the validation for them.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: I thought about the combination of these two expressions but seem cannot be simplified. Anyway, I combined them with one simplified version with missing an edge case coverage. The below URL starting with `raw.githubusercontent.com` should not have `[blob|tree]` part after repository name.
```
https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/purview/data-plane/Azure.Analytics.Purview.MetadataPolicies/preview/2021-07-01-preview/purviewMetadataPolicy.json
```

raych1:
@weshaggard , I added the package version validation to stop the validation if the package is not GA version.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
weshaggard: That is why I made `((blob|tree)/)?` optional in my regex example.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,216 @@
...
+}
+
+function Verify-PackageVersion() {
+  try {
+    $resolvedPath = Resolve-Path -Path $ArtifactLocation
+    if(-not (Test-Path -Path $resolvedPath -IsValid)) {
+      LogError "ServiceDir:$ServiceDirectory, PackageName:$PackageName. Invalid artifact location: $ArtifactLocation"
+      exit(1)
+    }
+    $pkgs, $parsePkgInfoFn = RetrievePackages -artifactLocation $resolvedPath
```
weshaggard, line 138: Let's avoid using RetrievePackages here instead do something similar to what APIView does. See https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Create-APIReview.ps1#L107C63-L107C63 it reads the metadata from the packageinfo.json file.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: I know, making it as optional cannot avoid this invalid case which should not have `blob` before `main`. However, it's ok to skip this case as this script is not the place to ensure the URL is valid.
```
https://raw.githubusercontent.com/Azure/azure-rest-api-specs/blob/main/specification/purview/data-plane/Azure.Analytics.Purview.MetadataPolicies/preview/2021-07-01-preview/purviewMetadataPolicy.json
```

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: I updated it and validated locally. Can you review? @weshaggard

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,275 @@
...
+  foreach($yamlSection in $yamlContent) {
+    if ($yamlSection) {
+      try {
+        # remove the lines like: $(tag) == 'package-preview-2023-09'
+        $yamlSection = $yamlSection -replace '^\s*\$\(.+\)\s*==.+', ''
+        $yamlobj = ConvertFrom-Yaml -Yaml $yamlSection
+        if($yamlobj) {
+          $batchValue = $yamlobj["batch"]
+          $requireValue = $yamlobj["require"]
+          $inputFileValue = $yamlobj["input-file"]
```
chunyu3, line 120: Are batchValue, requireValue, inputFileValue exclusive in a readme file? If they can exist concurrently, there will be problem.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,275 @@
...
+  $tspLocationYamlPath = Join-Path $PackageDirectory "tsp-location.yaml"
+  $autorestMdPath = Join-Path $PackageDirectory "src/autorest.md"
+  $swaggerReadmePath = Join-Path $PackageDirectory "swagger/README.md"
+  $tspLocationYaml = @{}
+  if (Test-Path -Path $tspLocationYamlPath) {
+    # typespec scenario
+    $tspLocationYaml = Get-Content -Path $tspLocationYamlPath -Raw | ConvertFrom-Yaml
+    Verify-TspLocation $tspLocationYaml
+  }
+  elseif ($Language -eq "dotnet") {
```
chunyu3, line 242: Better to use existence of file `autorest.md` as if-condition, not the language name for extensibility in future.

raych1: I prefer to split the cases by languages. Go language use 'autorest.md' but we will ignore it.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: I haven't found a real case with having two of these configurations. Currently, I follow below sequence for the validation. 
Do you have any idea of how does the generation tool handle this? 
+@weidongxu-microsoft, any comment?

weidongxu-microsoft: I am OK with current order.
In my understanding, we are not achieving full support for the markdown for autorest. (one may need to use the parse code from autorest to achieve full support, e.g. https://github.com/Azure/autorest/blob/main/packages/libs/configuration/src/literate-yaml/literate-yaml.test.ts)

So, as long as it works reasonably fine, I am good with it.

raych1: Thanks for sharing the reference of autorest code logic. I learned that autorest handles the case of `json` content in the markdown file. I would like to ignore that part because I haven't found the usage of `json` in `autorest.md` and `swagger/README.md`.

lirenhe (approved):

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,281 @@
...
+  [Parameter(Position = 0)]
+  [ValidateNotNullOrEmpty()]
+  [string] $ServiceDirectory,
+  [Parameter(Position = 1)]
+  [ValidateNotNullOrEmpty()]
+  [string] $PackageName,
+  [ValidateNotNullOrEmpty()]
+  [string] $ArtifactLocation,
+  [Parameter(Position = 2)]
+  [string]$GitHubPat = $($env:GH_TOKEN),
```
raych1, line 38: Get the token from environment variable so that it won't be leaked in the pipeline log.

raych1:
@weshaggard , can you also approve this PR if the change looks good?

eng/common/pipelines/templates/steps/verify-restapi-spec-location.yml
```diff
@@ -0,0 +1,25 @@
...
+
+steps:
+  - task: Powershell@2
+    inputs:
+      filePath: $(Build.SourcesDirectory)/eng/common/scripts/Verify-RestApiSpecLocation.ps1
+      arguments: >
+        -ServiceDirectory "${{ parameters.ServiceDirectory }}"
+        -PackageName "${{ parameters.PackageName }}"
+        -ArtifactLocation: "${{ parameters.ArtifactLocation }}"
+        -GitHubPat "${{  parameters.GitHubSpecRepoPat }}"
```
weshaggard, line 20: Why do we need to set it here as well as the env? Als it doesn't look like this parameter is listed in the template.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,281 @@
...
+  [string]$PackageInfoDirectory
```
weshaggard, line 39: When do we use this parameter over the ArtifactLocation?

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        Write-Host "Version: $($version)"
+        Write-Host "SDK Type: $($pkgInfo.SdkType)"
+        Write-Host "Release Status: $($pkgInfo.ReleaseStatus)"
+
+        # Ignore the validation if the package is not GA version
+        if ($version.IsPrerelease) {
+          Write-Host "ServiceDir:$ServiceDirectory, Package $($parsedPackage.PackageId) is marked with version $($parsedPackage.PackageVersion), the version is a prerelease version and the validation of spec location is ignored."
+          exit 0
```
weshaggard, line 214: Do you want to exit the entire script here? What if there are multiple packages?

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,281 @@
...
+      }
+      catch {
+        Write-Host "ServiceDir:$ServiceDirectory, PackageName:$PackageName. Failed to parse swagger/readme.md file with exception:`n$_ "
+      }
+    }
+  }
+}
+finally {
+  Pop-Location
+}
```
weshaggard, line 281: Please add a newline at the end of the file

eng/common/scripts/Verify-RestApiSpecLocation.ps1
weshaggard: Nevermind I just looked closer at the code and there shouldn't ever be more than one package returned. I know I told you to use this same method and I do still agree with that but I may want to clean that code up at some point.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -0,0 +1,281 @@
...
+    else
+    {
+      LogError "The function for 'FindArtifactForApiReviewFn' was not found.`
+      Make sure it is present in eng/scripts/Language-Settings.ps1 and referenced in eng/common/scripts/common.ps1.`
+      See https://github.com/Azure/azure-sdk-tools/blob/main/doc/common/common_engsys.md#code-structure"
+      exit 1
+    }
+    if (-not $PackageInfoDirectory)
+    {
+      $PackageInfoDirectory = Join-Path -Path $ArtifactLocation "PackageInfo"
```
weshaggard, line 179: Is this only for local testing?

eng/common/scripts/Verify-RestApiSpecLocation.ps1
weshaggard: I believe I figured out that this is for local testing.

weshaggard (approved):
A couple comments but overall looks good to get started using.

eng/common/pipelines/templates/steps/verify-restapi-spec-location.yml
raych1: Thanks, removed. Forgot to remove this line.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: No. This is used in release pipeline where doesn't pass in the value. Because during release job stage, it doesn't have the `packageInfo.json` generated at signed artifact folder. This script needs to generate that file before getting the package version.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: Done.

PR #7456, opened 2023-12-14
Redirect and show error message when diff revision id is not active

praveenkuttappan:
Accessing a URL with diff revision ID throws error if that diff revision is no longer valid. PR has the change to redirect user to review page instead of failing with a server error message.


JonathanCrd (approved):
🚀

src/dotnet/APIView/APIViewWeb/Pages/Assemblies/Review.cshtml
```diff
@@ -482,7 +482,12 @@
                     var diffRevisionType = (Model.ReviewContent.APIRevisionsGrouped.ContainsKey("Automatic")) ? "Automatic" : Model.ReviewContent.APIRevisionsGrouped.Keys.First();
                     @if (Model.DiffRevisionId != null)
```
weshaggard, line 483: Should this DiffRevisionId condition be changed like you did below?

praveenkuttappan: Yes . I will change that too. Thanks

src/dotnet/APIView/APIViewWeb/Pages/Assemblies/Review.cshtml
```diff
@@ -480,9 +480,14 @@
                 @if (@Model.ReviewContent.APIRevisionsGrouped.SelectMany(kvp => kvp.Value).Count() > 1)
                 {
                     var diffRevisionType = (Model.ReviewContent.APIRevisionsGrouped.ContainsKey("Automatic")) ? "Automatic" : Model.ReviewContent.APIRevisionsGrouped.Keys.First();
-                    @if (Model.DiffRevisionId != null)
+                    @if (Model.ReviewContent.DiffAPIRevision != null)
                     {
-                        diffRevisionType = Model.ReviewContent.DiffAPIRevision.APIRevisionType.ToString();
+                        // Diff revision is null if someone reuses an old URL with a deleted revision as  diff revision
+                        // Users copy and shares diff URL to approvers and sometimes they delete and recreate the revision which makes the diff revision null
+                        if (Model.ReviewContent.DiffAPIRevision?.APIRevisionType != null)
```
weshaggard, line 487:
```suggestion
                        if (Model.ReviewContent.DiffAPIRevision.APIRevisionType != null)
```
Shouldn't need it given you checked for null above.

weshaggard (approved):

PR #7467, opened 2023-12-15
Modify PR api change detection to calculate diff only between base…

praveenkuttappan:
…line and modified swagger.

1. Dispose memory streams
2. Create baseline and modified swagger review versions and calculate the diff against baseline only.
3. Save baseline revision info in pull request metadata.


weshaggard (approved):

maririos (approved):
Reviewed over a call. Thank you Praveen!

Issue #7124, opened 2023-10-12
[tsp-client] Add tests


catalinaperalta:
Add tests to cover tsp-client use cases and check behavior

catalinaperalta:
Fixed in #8613

PR #7471, opened 2023-12-16
[tsp-client] Fix typespec compiler usage

catalinaperalta:
Use functions provided by `@typespec/compiler` to resolve emitter options and paths passed into the `compile()` call. 

Issues to be addressed in a follow up PR: #7124 , https://github.com/Azure/azure-sdk-tools/issues/7472 <-- will do a wider pass through all of the path operations we do in the tool and update to using the functions provided by typespec/compiler

### DIFF
--- a/tools/tsp-client/src/index.ts
+++ b/tools/tsp-client/src/index.ts
@@ -3,7 +3,7 @@ import * as path from "node:path";
 import { installDependencies } from "./npm.js";
 import { createTempDirectory, removeDirectory, readTspLocation, getEmitterFromRepoConfig } from "./fs.js";
 import { Logger, printBanner, enableDebug, printVersion } from "./log.js";
-import { TspLocation, compileTsp, discoverMainFile, getEmitterOptions, resolveTspConfigUrl } from "./typespec.js";
+import { TspLocation, compileTsp, discoverMainFile, resolveTspConfigUrl } from "./typespec.js";
 import { getOptions } from "./options.js";
 import { mkdir, writeFile, cp, readFile } from "node:fs/promises";
 import { addSpecFiles, checkoutCommit, cloneRepo, getRepoRoot, sparseCheckout } from "./git.js";
@@ -151,11 +151,9 @@ async function syncTspFiles(outputDir: string, localSpecRepo?: string) {
 async function generate({
   rootUrl,
   noCleanup,
-  additionalEmitterOptions,
 }: {
   rootUrl: string;
   noCleanup: boolean;
-  additionalEmitterOptions?: string;
 }) {
   const tempRoot = path.join(rootUrl, "TempTypeSpecFiles");
   const tspLocation = await readTspLocation(rootUrl);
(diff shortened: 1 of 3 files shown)


tools/tsp-client/src/index.ts
```diff
@@ -3,12 +3,13 @@ import * as path from "node:path";
 import { installDependencies } from "./npm.js";
 import { createTempDirectory, removeDirectory, readTspLocation, getEmitterFromRepoConfig } from "./fs.js";
 import { Logger, printBanner, enableDebug, printVersion } from "./log.js";
-import { TspLocation, compileTsp, discoverMainFile, getEmitterOptions, resolveTspConfigUrl } from "./typespec.js";
+import { TspLocation, compileTsp, discoverMainFile, resolveTspConfigUrl } from "./typespec.js";
 import { getOptions } from "./options.js";
 import { mkdir, writeFile, cp, readFile } from "node:fs/promises";
 import { addSpecFiles, checkoutCommit, cloneRepo, getRepoRoot, sparseCheckout } from "./git.js";
 import { fetch } from "./network.js";
 import { parse as parseYaml } from "yaml";
```
timotheeguerin, line 11: Are you stil needing the yaml parser?

tools/tsp-client/src/typespec.ts
```diff
@@ -114,38 +52,54 @@ export async function discoverMainFile(srcDir: string): Promise<string> {
...
-  const parsedEntrypoint = path.parse(resolvedMainFilePath);
-  const { compile, NodeHost, getSourceLocation } = await importTsp(parsedEntrypoint.dir);
-  // compile the local copy of the root file
-  const program = await compile(NodeHost, resolvedMainFilePath, {
-    outputDir: outputPath,
+  const outputDir = resolvePath(outputPath);
+  const overrideOptions: Record<string, Record<string, string>> = {};
+  overrideOptions[emitterPackage] = {"emitter-output-dir": outputDir}
+  if (saveInputs) {
+    overrideOptions[emitterPackage]!["save-inputs"] = "true";
```
timotheeguerin, line 75: I would save that object to a variable so you don't have to always `!`. Not sure you actually need to anyway

tools/tsp-client/src/typespec.ts
```diff
@@ -114,38 +52,54 @@ export async function discoverMainFile(srcDir: string): Promise<string> {
...
+  }
+  if (additionalEmitterOptions) {
```
timotheeguerin, line 77: where do those options come from?

tools/tsp-client/src/index.ts
catalinaperalta: Yes we still need the parser when we use the values from tsp-location.yaml to sync the tsp project.

tools/tsp-client/src/typespec.ts
catalinaperalta: The user provides them, we added this to be fully compatible with the original powershell scripts, see the generate step: https://github.com/Azure/azure-sdk-tools/blob/main/doc/common/TypeSpec-Project-Scripts.md#typespec-project-generateps1

tools/tsp-client/src/typespec.ts
catalinaperalta: It was strange to me as well that I got an error that overrideOptions[emitterPackage] might be undefined, because I am clearly setting an initial value for it on line 73

tools/tsp-client/src/typespec.ts
timotheeguerin: I don’t really like that there is 2 ways to provide options that are not consistent but I guess if it’s for back compat not much we can do

timotheeguerin (approved):

PR #7479, opened 2023-12-21
[tsp-client] Improve additional directory processing

catalinaperalta:
- Correct output to the `additionalDirectories` property in tsp-location.yaml
- Allow local specs with the update command
- Fix additional directory support for local specs


tools/tsp-client/src/utils.ts
```diff
@@ -0,0 +1,7 @@
+export function formatAdditionalDirectories(additionalDirectories?: string): string {
```
joheredi, line 1: should this be an string[] instead of string?

catalinaperalta: It should be string[], just pushed the updated type for this parameter. Thanks for pointing that out :)

joheredi (approved):
Looks great :shipit:

PR #7488, opened 2024-01-03
Update all projects to v5 of graph api

weshaggard:
Part of the tools used v5 and others v4 which causes runtime exceptions because of the breaking changes between those version, so updated all projects to be using v5.

Fixed bug when user doesn't have permission to update an endpoint/service connection so it correctly errors instead of throwing null ref exception.


tools/secret-management/Azure.Sdk.Tools.SecretRotation.Stores.AzureDevOps/ServiceConnectionParameterStore.cs
```diff
@@ -83,6 +83,11 @@ public class ServiceConnectionParameterStore : SecretStore
         this.logger.LogDebug("Getting service endpoint details for endpoint '{EndpointId}'", endpointId);
         ServiceEndpoint? endpointDetails = await client.GetServiceEndpointDetailsAsync(this.projectName, endpointId);
 
+        if (endpointDetails == null || endpointDetails.Authorization == null)
```
weshaggard, line 86: This is the actual bug I planned to fix but the tool wouldn't work locally while debugging because of the graph versioning issue so I fixed those as well.

konrad-jamrozik (approved):

PR #7512, opened 2024-01-11
Add script which generates a changeset for pipelinev3

### LINKED ISSUE #7504: Write `blast-radius` script which detects changed files in PR
~~This will be very similar to [Changed-Files-Functions](https://github.com/Azure/azure-rest-api-specs/blob/main/eng/scripts/ChangedFiles-Functions.ps1)~~

~~I may actually just wholesale move this file to eng/common, and add `azure-rest-api-specs` to eng/common sync.~~

This function is already present in `git-helpers.ps1`. Will re-use that. The function generating the json file at rest (with the changes) still needs to exist though. Re-purposing this issue for that.

scbedd:
Using `git-helpers` `Get-ChangedFiles`

Resolves #7504 

This will be used as the first step of a common build pipeline which we will add to all our repos.

image

### DIFF
--- a/eng/common/scripts/Generate-PR-Diff.ps1
+++ b/eng/common/scripts/Generate-PR-Diff.ps1
@@ -1,0 +1,51 @@
+<#
+.SYNOPSIS
+Script used to generate the diff.json file for a PR. Explicitly intended to work in a PR context.
+
+.DESCRIPTION
+Combines the result of git diff, some parsed details from the diff, and the PR number into a single JSON file. This JSON file is intended for use further along the pipeline.
+
+.PARAMETER ArtifactPath
+The folder in which the result will be written.
+
+.PARAMETER TargetPath
+The path under which changes will be detected.
+#>
+[CmdletBinding()]
+Param (
+  [Parameter(Mandatory=$True)]
+  [string] $ArtifactPath,
+  [Parameter(Mandatory=$True)]
+  [string] $TargetPath
+)
+
+. (Join-Path $PSScriptRoot "Helpers" git-helpers.ps1)
+
+function Get-ChangedServices {
+    Param (
+        [Parameter(Mandatory=$True)]
+        [string[]] $ChangedFiles
+    )
+    
+    $changedServices = $ChangedFiles | Foreach-Object { if ($_ -match "sdk/([a-zA-Z\-]+)") { $matches[1] } } | Sort-Object -Unique
+
+    return $changedServices
+}
+
+if (!(Test-Path $ArtifactPath)) {
+    New-Item -ItemType Directory -Path $ArtifactPath | Out-Null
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Generate-PR-Diff.ps1
```diff
@@ -0,0 +1,51 @@
...
+
+. (Join-Path $PSScriptRoot "Helpers" git-helpers.ps1)
+
+function Get-ChangedServices {
+    Param (
+        [Parameter(Mandatory=$True)]
+        [string[]] $ChangedFiles
+    )
+    
+    $changedServices = $ChangedFiles | Foreach-Object { if ($_ -match "sdk/([a-zA-Z\-]+)") { $matches[1] } } | Sort-Object -Unique
```
weshaggard, line 30:
```suggestion
    $changedServices = $ChangedFiles | Foreach-Object { if ($_ -match "sdk/([^/]+)") { $matches[1] } } | Sort-Object -Unique
```
Might be easier to match anything other than the next directory separator. I cannot remember for sure but does the Get-ChangedFiles always come back with forward slashes from git on both windows and linux?

weshaggard (approved):
Not sure what you are going to do with this info yet but looks reasonable.

eng/common/scripts/Generate-PR-Diff.ps1
```diff
@@ -0,0 +1,51 @@
...
+#>
+[CmdletBinding()]
+Param (
+  [Parameter(Mandatory=$True)]
+  [string] $ArtifactPath,
+  [Parameter(Mandatory=$True)]
+  [string] $TargetPath
+)
+
+. (Join-Path $PSScriptRoot "Helpers" git-helpers.ps1)
```
weshaggard, line 22: orthogonal to this change but at some point I wanted to update these helpers to match what we have done in the specs repo at https://github.com/Azure/azure-rest-api-specs/blob/main/eng/scripts/ChangedFiles-Functions.ps1#L1. It helps us stop depending on the devops variables.

scbedd: Ah ok. I'll submit a PR on that + check its usages in a later PR. Thanks Wes.

eng/common/scripts/Generate-PR-Diff.ps1
scbedd: I will confirm this and get back to you.

scbedd: I mean if not it just changes to `[^/\\]`

scbedd: Confirmed. This is the output for a windows agent:
image

Identical.

PR #7523, opened 2024-01-12
[Pylint] bug for paramtype

### LINKED ISSUE #7486: [pylint] false-negative for docstring-keyword-should-match-keyword-only
This pipeline shows the pylint errors for app configuration here: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3372424&view=logs&j=e5b925a9-9650-5321-24fd-5d9499fe02b8&t=e1fa7d9e-8471-5a74-cd7d-e1c9a992e07e&l=130
 
Pylint is only catching a mismatch with `etag` on the function here: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_azure_appconfiguration_client.py#L313

But there are other functions present that have the same issue: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_azure_appconfiguration_client.py#L379

The latter should also throw a linting error. It seems like there might be bug in which only one instance of violating the rule is reported on the client.

l0lawrence:
fixes #7486

### DIFF
--- a/tools/pylint-extensions/azure-pylint-guidelines-checker/CHANGELOG.md
+++ b/tools/pylint-extensions/azure-pylint-guidelines-checker/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release History
 
+## 0.3.1 (2023-1-16)
+
+- Docstring bug fix where paramtype was being consider for params
+
 ## 0.3.0 (2023-12-15)
 
 - Breaking changes involved in bump to pylint 3.0 support and bug fix to `incorrect-naming-convention` checker
--- a/tools/pylint-extensions/azure-pylint-guidelines-checker/CHANGELOG.md
+++ b/tools/pylint-extensions/azure-pylint-guidelines-checker/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.3.1 (2023-1-16)
 
-- Docstring bug fix where paramtype was being consider for params
+- Docstring bug fix where paramtype was being considered for params
 
 ## 0.3.0 (2023-12-15)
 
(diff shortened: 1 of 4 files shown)


tools/pylint-extensions/azure-pylint-guidelines-checker/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release History
 
+## 0.3.1 (2023-1-16)
+
+- Docstring bug fix where paramtype was being consider for params
```
l0lawrence, line 5: since only a bug fix and no added checks

l0lawrence, line 5:
```suggestion
- Docstring bug fix where paramtype was being considered for params
```

tools/pylint-extensions/azure-pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
```diff
@@ -4433,10 +4433,11 @@ def test_docstring_varag_no_type(self):
         # Error on documenting keyword only args as param after *args in docstring
         node = astroid.extract_node(
             """
-            def function_foo(*x, y):
```
kristapratico, line 4436: Do we have any tests with a client that has multiple methods that violate the rule?

kristapratico (approved):

Issue #7502, opened 2024-01-10
[tsp-client] Add `tsp-client convert` command


catalinaperalta:
Add a `convert` command that works similarly to what is described here: https://azure.github.io/typespec-azure/docs/next/howtos/migrate-swagger/get-started

NOTE: this command should be used instead of the convert.ps1 script in CI

PR #7538, opened 2024-01-17
[tsp-client] Support swagger to typespec conversion

catalinaperalta:
Add support for swagger->tsp conversion with `convert` command. 
Fixes https://github.com/Azure/azure-sdk-tools/issues/7502

Relies on changes in #7525 (merge after 7525)

### DIFF
--- a/tools/tsp-client/src/index.ts
+++ b/tools/tsp-client/src/index.ts
@@ -5,10 +5,12 @@ import { TspLocation, compileTsp, discoverMainFile, resolveTspConfigUrl } from "
 import { getOptions } from "./options.js";
 import { mkdir, writeFile, cp, readFile } from "node:fs/promises";
 import { addSpecFiles, checkoutCommit, cloneRepo, getRepoRoot, sparseCheckout } from "./git.js";
-import { fetch } from "./network.js";
+import { doesFileExist, fetch } from "./network.js";
 import { parse as parseYaml } from "yaml";
-import { joinPaths, resolvePath } from "@typespec/compiler";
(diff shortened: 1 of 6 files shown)


tools/tsp-client/src/index.ts
```diff
@@ -182,6 +184,28 @@ async function generate({
   }
 }
 
+
+async function convert(readme: string, outputDir: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+      const autorest = spawn("autorest", [readme, "--openapi-to-typespec", "--use=@autorest/openapi-to-typespec", `--output-folder=${outputDir}`], {
```
timotheeguerin, line 190: Should you maybe check if autorest is  installed? or maybe can use `npx autorest` to download on the fly

catalinaperalta: I was going to put autorest as a prerequisite for install on the instructions...but I think npx autorest would be nice too, I'll switch it up to npx autorest

tools/tsp-client/src/index.ts
```diff
@@ -233,6 +257,14 @@ async function main() {
         await syncTspFiles(rootUrl, options.localSpecRepo);
         await generate({ rootUrl, noCleanup: options.noCleanup, additionalEmitterOptions: options.emitterOptions});
         break;
+      case "convert":
+        Logger.info("Converting swagger to typespec...");
+        let readme = options.swaggerReadme!;
```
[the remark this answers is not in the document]
catalinaperalta: We check that the swagger readme is specified on lines 119-125 in the options.ts file so I decided not to do it again here. Let me know if you think this is still a risk!

Issue #4644, opened 2022-11-07
[stress testing] Add job/pod labels for git commit sha


benbp:
This will make it a little easier to track versions from container logs in log analytics/KubePodInventory.

benbp:
Closing in favor of https://github.com/Azure/azure-sdk-tools/issues/7521

PR #7543, opened 2024-01-17
Add git commit details to stress environment

### LINKED ISSUE #7521: Add env var with git commit hash to stress test build
I'd like to have a recorded git commit hash of the code being tested in my stress tests.
Having something like `--build-arg GIT_COMMIT=$(git rev-parse HEAD)` in the stress test deployment script would allow anyone to use this as they see please (e.g. stamp on telemetry for troubleshooting/reporting later)

EDIT (@benbp): Also add job/pod labels for git commit sha, so that queries can use the git commit to locate pods running it.

benbp:
Fixes #7521 #4644. See README.md diff in PR for details.

FYI @lmolkova @richardpark-msft

### DIFF
--- a/eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
+++ b/eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
@@ -321,7 +321,18 @@ function DeployStressPackage(
     $generatedConfigPath = Join-Path $pkg.Directory generatedValues.yaml
     $subCommand = $Template ? "template" : "upgrade"
     $subCommandFlag = $Template ? "--debug" : "--install"
-    $helmCommandArg = "helm", $subCommand, $releaseName, $pkg.Directory, "-n", $pkg.Namespace, $subCommandFlag, "--values", $generatedConfigPath, "--set", "stress-test-addons.env=$environment"
+    $helmCommandArg = @(
+        "helm", $subCommand, $releaseName, $pkg.Directory,
+        "-n", $pkg.Namespace,
+        $subCommandFlag,
+        "--values", $generatedConfigPath,
+        "--set", "stress-test-addons.env=$environment"
+    )
+
+    $gitCommit = git -C $pkg.Directory rev-parse HEAD 2>&1
+    if (!$LASTEXITCODE) {
+        $helmCommandArg += "--set", "GitCommit=$gitCommit"
+    }
 
     if ($LockDeletionForDays) {
         $date = (Get-Date).AddDays($LockDeletionForDays).ToUniversalTime()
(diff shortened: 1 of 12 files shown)


eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -321,7 +321,20 @@ function DeployStressPackage(
...
+        "helm", $subCommand, $releaseName, $pkg.Directory,
+        "-n", $pkg.Namespace,
+        $subCommandFlag,
+        "--values", $generatedConfigPath,
+        "--set", "stress-test-addons.env=$environment"
+    )
+
+    Push-Location $pkg.Directory
+    $gitCommit = git rev-parse HEAD 2>&1
+    Pop-Location
```
weshaggard, line 334:
```suggestion
    $gitCommit = git --git-dir $pkg.Directory rev-parse HEAD 2>&1
```

weshaggard (approved):
One suggestion but otherwise seems good.

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
benbp: Ooh didn't know about that!

benbp:
> One suggestion but otherwise seems good.

@weshaggard looks like `--git-dir` only works if pointing at the `.git` directory, but `-C` has the intended effect.

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -321,7 +321,18 @@ function DeployStressPackage(
...
-    $helmCommandArg = "helm", $subCommand, $releaseName, $pkg.Directory, "-n", $pkg.Namespace, $subCommandFlag, "--values", $generatedConfigPath, "--set", "stress-test-addons.env=$environment"
+    $helmCommandArg = @(
+        "helm", $subCommand, $releaseName, $pkg.Directory,
+        "-n", $pkg.Namespace,
+        $subCommandFlag,
+        "--values", $generatedConfigPath,
+        "--set", "stress-test-addons.env=$environment"
+    )
+
+    $gitCommit = git -C $pkg.Directory rev-parse HEAD 2>&1
```
weshaggard, line 332: I don't think you want `-C` here do you? When I tried that locally I don't think it worked. I thought that was referring to the configuration location.

weshaggard:
> > One suggestion but otherwise seems good.
> 
> @weshaggard looks like `--git-dir` only works if pointing at the `.git` directory, but `-C` has the intended effect.

That is interesting as that I had the opposite when I tested locally. I could point to the folder that contained the `.git` directory and I thought `-C` was for the configuration file.

PR #7554, opened 2024-01-19
[.NET] Add "Azure.Developer" to allowed namespace prefixes

joseharriaga:


jsquire (approved):

pallavit:
Looking at the 'blessed list' https://github.com/Azure/azure-sdk-tools/blob/e44b71b0dd82e138281a01d202535df37a93f395/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAssemblyNamespaceAnalyzer.cs#L15 here seems short? Is this the full list we have? If not, do we usually go the route of disabling this test since grep shows a lot of suppressions to this warning.

I guess I am curious - what is the recommended path here? Does the APIView have a check for package approval now? If so, does this warning simply becomes good to have? If not, should we also work towards cleaning up all the older supressions and make this list exhaustive?

pallavit:
/cc: @JoshLove-msft , @christothes who may know more about this.

christothes:
In my experience, these warnings are only suppressed when a new one is introduced and is still in preview. The unsuppressed warning will break the build by default, since we treat warnings as errors.

pallavit:
> In my experience, these warnings are only suppressed when a new one is introduced and is still in preview. The unsuppressed warning will break the build by default, since we treat warnings as errors.

Ahh so this list then should be exhaustive for an SDK that is going GA? In that case, I do not have any concerns with the change.

PR #7569, opened 2024-01-23
Correct the name of JS package folder

raych1:
There are two cases for the naming of JS package folder:
- RLC package case, removing the "azure-rest-" from package name then plus "-rest" to the end of the name.
- Non-RLC package case, removing the "azure-" from the package name.

The changes include:
- correct the name of package folder for JS case.
- correct the name prefix of JS ARM package.
- add exception catch handler
- style format fix

The [pipeline test run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3430267&view=logs&j=7bb69cc7-6f52-585e-8b13-5bae7c4a9fa4&t=7abbbe6b-d005-5c95-feb6-99ffe4e699e1) of JS template package release with same script change looks good.

### DIFF
--- a/eng/common/scripts/Verify-RestApiSpecLocation.ps1
+++ b/eng/common/scripts/Verify-RestApiSpecLocation.ps1
@@ -48,6 +48,8 @@ if (-not $GitHubPat) {
   exit 1
 }
 
+Write-Host "The spec used to release SDK should be from the main branch of Azure/azure-rest-api-specs repository."
+Write-Host "ServiceDir:$ServiceDirectory, PackageName:$PackageName, ArtifactLocation:$ArtifactLocation, PackageInfoDirectory:$PackageInfoDirectory."
 Install-ModuleIfNotInstalled "powershell-yaml" "0.4.1" | Import-Module
 
 # This function is used to verify the 'require' and 'input-file' settings in autorest.md point to the main branch of Azure/azure-rest-api-specs repository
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -259,8 +264,8 @@ try{
   }
   elseif ($Language -eq "java" -or $Language -eq "javascript" -or $Language -eq "python" -or $Language -eq "go") {
     # for these languages we ignore the validation because they always use the latest spec from main branch to release SDK
-    # mgmt plane packages: azure-core-management|azure-resourcemanager|azure-resourcemanager-advisor (java), azure-mgmt-devcenter (python), arm-advisor (js), armaad (go)
-    if($PackageName -match "^(arm|azure-mgmt|azure-resourcemanager|azure-core-management)[-a-z]*$") {
+    # mgmt plane packages: azure-core-management|azure-resourcemanager|azure-resourcemanager-advisor (java), azure-mgmt-devcenter (python), azure-arm-advisor (js), armaad (go)
+    if ($PackageName -match "^(arm|azure-mgmt|azure-resourcemanager|azure-core-management|azure-arm|azure-rest-arm)[-a-z]*$") {
```
raych1, line 268: @qiaozha , can you confirm the name prefix of JS arm packages?

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -227,12 +222,24 @@ function Verify-PackageVersion() {
   }
 }
 
-try{
+try {
   # Verify package version is not a prerelease version, only continue the validation if the package is GA version
   Verify-PackageVersion
 
   $ServiceDir = Join-Path $RepoRoot 'sdk' $ServiceDirectory
```
weshaggard, line 229: We don't need to build this path. You already have the information from the pkgInfo object you create in `Verify-PackageVersion`. That function is computing the properties and has all the logic. For example the package info object for `@azure-rest/core-client` has the following:
```
{
  "Name": "@azure-rest/core-client",
  "Version": "1.2.0",
  "DevVersion": "1.2.0-alpha.20240123.1",
  "DirectoryPath": "sdk/core/core-client-rest",
  "ServiceDirectory": "core",
  "ReadMePath": "sdk/core/core-client-rest/README.md",
  "ChangeLogPath": "sdk/core/core-client-rest/CHANGELOG.md",
  "Group": null,
  "SdkType": "client",
  "IsNewSdk": true,
  "ArtifactName": "azure-rest-core-client",
  "ReleaseStatus": "Unreleased"
}
```

So instead return that pkgInfo object and use those properties so we don't have to maintain this extra logic that will definitely get broken at some point. That is the reason we have these common package properties and have each language implement them directly.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
weshaggard: You should be able to use the pkgInfo.SdkType for this check.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -227,12 +222,24 @@ function Verify-PackageVersion() {
...
+    $prefixOfRlcPkg = "azure-rest-"
+    $prefixOfAzurePkg = "azure-"
+    if ($PackageName -match $prefixOfRlcPkg) {
+      $PackageDirectory = Join-Path $ServiceDir ($PackageName.Substring($prefixOfRlcPkg.Length) + "-rest")
+    }
+    else {
+      $PackageDirectory = Join-Path $ServiceDir $PackageName.Substring($prefixOfAzurePkg.Length)
+    }
+  }
   Push-Location $PackageDirectory
```
weshaggard, line 243: We should add a check to ensure that the PackageDirectory exists.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
weshaggard: To add a little more context we have a common class at https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Package-Properties.ps1#L29 that contains these properties and we have each language create this object during discover see JS's implementation for example https://github.com/Azure/azure-sdk-for-js/blob/main/eng/scripts/Language-Settings.ps1#L28 so it handles all these special edge cases.

weshaggard (changes requested):
Let's switch to using the package info properties instead of duplicating logic.

raych1:
> Let's switch to using the package info properties instead of duplicating logic.

@weshaggard , switched to rely on the package info. This is the [test run result](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3433518&view=logs&j=7bb69cc7-6f52-585e-8b13-5bae7c4a9fa4&t=7abbbe6b-d005-5c95-feb6-99ffe4e699e1).

weshaggard (approved):

PR #7585, opened 2024-01-25
Remove package retrieval when verify pkg version

raych1:
The script fails to find the package for the `Go` language because the Go language doesn't generate a package for release; instead, it simply pushes the code to the GitHub repository. 

However, since the `Verify-PackageVersion` function doesn't actually use the package found by the `APIView` related function, I removed that part of the code. Now, the script works for these five languages by retrieving package information from the `{packagename}.json` file.

In addition, I added the validation for `Go` data plane scenario.

Test run result: 
[Go](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3438606&view=logs&j=6a8552c2-2ada-5ee5-a4d2-6fcf45f87c55&t=63abb2ae-8b99-518c-37b5-ceb4c0582ab3)
[JS](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3438579&view=logs&j=fad5caa0-24eb-51d7-9ee8-d55d6e9ffe5a&t=4b903053-dc5b-56d4-4838-edf64acaab7d)

### DIFF
--- a/eng/common/scripts/Verify-RestApiSpecLocation.ps1
+++ b/eng/common/scripts/Verify-RestApiSpecLocation.ps1
@@ -9,7 +9,7 @@
   The directory path of the service.
 
 .PARAMETER PackageName
-  The name of the package.
+  The name of the package, which is the artifact name in pipeline.
 
 .PARAMETER ArtifactLocation
   The location of the generated artifact for the package.
--- a/eng/common/scripts/Verify-RestApiSpecLocation.ps1
+++ b/eng/common/scripts/Verify-RestApiSpecLocation.ps1
@@ -21,7 +21,7 @@
   The directory path where the package information is stored.
 
 .EXAMPLE
-  Verify-RestApiSpecLocation -ServiceDirectory "/home/azure-sdk-for-net/sdk/serviceab" -PackageName "MyPackage" -ArtifactLocation "/home/ab/artifacts" -GitHubPat "xxxxxxxxxxxx" -PackageInfoDirectory "/home/ab/artifacts/PackageInfo"
+  Verify-RestApiSpecLocation -ServiceDirectory "template" -PackageName "Azure.Template" -ArtifactLocation "/home/ab/artifacts" -GitHubPat "xxxxxxxxxxxx" -PackageInfoDirectory "/home/ab/artifacts/PackageInfo"
 
 #>
 [CmdletBinding()]
@@ -164,16 +164,6 @@ function Verify-YamlContent([string]$markdownContent, [string]$configFilePath) {
 
 function Verify-PackageVersion() {
   try {
-    $packages = @{}
-    if ($FindArtifactForApiReviewFn -and (Test-Path "Function:$FindArtifactForApiReviewFn")) {
-      $packages = &$FindArtifactForApiReviewFn $ArtifactLocation $PackageName
-    }
-    else {
-      LogError "The function for 'FindArtifactForApiReviewFn' was not found.`
-      Make sure it is present in eng/scripts/Language-Settings.ps1 and referenced in eng/common/scripts/common.ps1.`
-      See https://github.com/Azure/azure-sdk-tools/blob/main/doc/common/common_engsys.md#code-structure"
-      exit 1
-    }
     if (-not $PackageInfoDirectory) {
       $PackageInfoDirectory = Join-Path -Path $ArtifactLocation "PackageInfo"
       if (-not (Test-Path -Path $PackageInfoDirectory)) {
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -164,16 +164,6 @@ function Verify-YamlContent([string]$markdownContent, [string]$configFilePath) {
...
-    if ($FindArtifactForApiReviewFn -and (Test-Path "Function:$FindArtifactForApiReviewFn")) {
-      $packages = &$FindArtifactForApiReviewFn $ArtifactLocation $PackageName
-    }
-    else {
-      LogError "The function for 'FindArtifactForApiReviewFn' was not found.`
-      Make sure it is present in eng/scripts/Language-Settings.ps1 and referenced in eng/common/scripts/common.ps1.`
-      See https://github.com/Azure/azure-sdk-tools/blob/main/doc/common/common_engsys.md#code-structure"
-      exit 1
-    }
     if (-not $PackageInfoDirectory) {
```
weshaggard, line 167: Before this was only for local testing but now after your change this will happen every time and I suspect that isn't what we want. The idea was that we would get the already existing packageInfo from the artifacts but it appears that go doesn't publish those. So, the question is should we have go start publishing them? We might need to for other scenarios like APIView. However if we don't do that we shouldn't write the json files to disk we should just call the helper functions that the at save script is calling.

See https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Save-Package-Properties.ps1#L95

For example there will be cases where looking for `$PackageName.json` will fail that were already handled in that save script https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Save-Package-Properties.ps1#L115

weshaggard:
@praveenkuttappan @benbp this means that the APIView requirement check doesn't work for go. Is that a know gap? Did we never implement that block for go?

benbp:
> @praveenkuttappan Praveen Kuttappan FTE @benbp Ben Broderick Phillips FTE this means that the APIView requirement check doesn't work for go. Is that a know gap? Did we never implement that block for go?

@weshaggard to my knowledge this block was working, we implemented it last year.

weshaggard:
> > @praveenkuttappan Praveen Kuttappan FTE @benbp Ben Broderick Phillips FTE this means that the APIView requirement check doesn't work for go. Is that a know gap? Did we never implement that block for go?
> 
> @weshaggard to my knowledge this block was working, we implemented it last year.

OK looks like go doesn't use the PackageInfo files https://github.com/Azure/azure-sdk-for-go/blob/main/eng/scripts/Language-Settings.ps1#L137 it is looking for apireview files directly which is why this doesn't work in our general case. We might want to have go start producing those PackageInfo files as artifacts so things can be more consistent and we have that info for other parts of the pipeline.

raych1:
> > @praveenkuttappan Praveen Kuttappan FTE @benbp Ben Broderick Phillips FTE this means that the APIView requirement check doesn't work for go. Is that a know gap? Did we never implement that block for go?
> 
> @weshaggard to my knowledge this block was working, we implemented it last year.

@benbp @weshaggard API view works in Go. The point is Go SDK doesn't have package generated like other language which is by design. We could publish the API view artifacts which is `*.gosource` then we can use it in this script but this artifact is not real `Go SDK` package. I don't recommend this approach although it will work.
Alternatively, we could inject the script in the pipeline yml like [here](https://github.com/Azure/azure-sdk-for-go/blob/6e01b90b7fb2175ed82e40218c926a6dab12995d/eng/pipelines/templates/steps/analyze.yml#L13) or we publish the `PackageInfo` after it is created then it can be used in this script. This is only need for Go scenario as for other four languages, `PackageInfo` has been already published as artifacts.

raych1:
@weshaggard , I added the step to publish the `PackageInfo` in Go pipeline. Please see the linked PR for details.

eng/common/scripts/Verify-RestApiSpecLocation.ps1
```diff
@@ -182,39 +172,29 @@ function Verify-PackageVersion() {
         & $savePropertiesScriptPath -serviceDirectory $ServiceDirectory -outDirectory $PackageInfoDirectory
       }
     }
+    $pkgPropPath = Join-Path -Path $PackageInfoDirectory "$PackageName.json"
```
weshaggard, line 175: PackageName is actually ArtifactName https://github.com/Azure/azure-sdk-for-js/blob/main/eng/pipelines/templates/stages/archetype-js-release.yml#L49

weshaggard (approved):

eng/common/scripts/Verify-RestApiSpecLocation.ps1
raych1: I clarified the parameter description.

PR #7597, opened 2024-01-29
[.NET] Add missing prefixes to list of allowed namespace prefixes

joseharriaga:
I went through the [Azure SDK Releases page](https://azure.github.io/azure-sdk/#net) on azure.github.io and compared it to the list of approved namespace prefixes in our analyzers in order to identify and close any gaps. After talking to Ted, I'm adding: 1. Azure.Health, 2. Azure.Maps, and 3. Azure.Verticals, while I'm also removing 1. Azure.Learn (previously used for a now deprecated training project) and 2. Azure.Management (previously used for management plane but now replaced by Azure.ResourceManager). Finally, I'm removing Azure.Template from the list of approved prefixes but adding code to treat like the special case it is.


src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAssemblyNamespaceAnalyzer.cs
```diff
@@ -73,6 +73,14 @@ public override void Analyze(ISymbolAnalysisContext context)
                 {
                     return;
                 }
+
+                // "Azure.Template" is not an approved namespace prefix, but we have a project template by that name
```
jsquire, line 77: Nice.  I like the distinction here!

jsquire (approved):

PR #7615, opened 2024-02-01
Prepare-Release.ps1: Make dateTime.ToString("MM/dd/yyyy") to work on exotic setups

antkmsft:
On my machine, I experimented with the registry, and the worst part is that I don't remember/know how to reset it back.

The work items that script produces, do have datetimes for the upcoming releases in the `MM-dd-yyyy` format, and then I have to correct them by hand.

`dateTime.ToString("MM/dd/yyyy")` does produce the date in the format of `MM-dd-yyyy` on my machine. This also happens if I write a corresponding .NET app.

The fix that I am proposing makes it work on my specific setup and hopefully breaks no one else. I understand if you are hesitant to take it. Let me know, I'll see how I can restore my setting.

But on the other hand, I don't think it makes anything worse, it only makes things more robust, so maybe take it?

### DIFF
--- a/eng/common/scripts/Prepare-Release.ps1
+++ b/eng/common/scripts/Prepare-Release.ps1
@@ -112,7 +112,7 @@ else
 # On some machines, where people modified date format via Windows Registry,
 # "$ParsedReleaseDate.ToString("MM/dd/yyyy")" produces the date as "MM-dd-yyyy", which is probably a bug somewhere,
 # yet if we concatenate the parts individually, everything works as expected.
-$releaseDateString = $ParsedReleaseDate.ToString("MM") + "/" + $ParsedReleaseDate.ToString("dd") + "/" + $ParsedReleaseDate.ToString("yyyy")
+$releaseDateString = $ParsedReleaseDate.ToSTring("MM/dd/yyyy", [CultureInfo]::InvariantCulture)
 $month = $ParsedReleaseDate.ToString("MMMM")
 
 Write-Host "Assuming release is in $month with release date $releaseDateString" -ForegroundColor Green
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Prepare-Release.ps1
```diff
@@ -109,7 +109,10 @@ else
   $ParsedReleaseDate = [datetime]$ReleaseDate
 }
 
-$releaseDateString = $ParsedReleaseDate.ToString("MM/dd/yyyy")
+# On some machines, where people modified date format via Windows Registry,
+# "$ParsedReleaseDate.ToString("MM/dd/yyyy")" produces the date as "MM-dd-yyyy", which is probably a bug somewhere,
+# yet if we concatenate the parts individually, everything works as expected.
+$releaseDateString = $ParsedReleaseDate.ToString("MM") + "/" + $ParsedReleaseDate.ToString("dd") + "/" + $ParsedReleaseDate.ToString("yyyy")
```
weshaggard, line 115:
```suggestion
$releaseDateString = $ParsedReleaseDate.ToSTring("MM/dd/yyyy", [CultureInfo]::InvariantCulture)
```
Can you try that on your machine to see if it works?

antkmsft: Thank you! It works. It is a better solution.

eng/common/scripts/Prepare-Release.ps1
```diff
@@ -109,7 +109,8 @@ else
   $ParsedReleaseDate = [datetime]$ReleaseDate
 }
 
-$releaseDateString = $ParsedReleaseDate.ToString("MM/dd/yyyy")
+# Use InvariantCulture so that the date format is consistent on all machines
+$releaseDateString = $ParsedReleaseDate.ToSTring("MM/dd/yyyy", [CultureInfo]::InvariantCulture)
```
weshaggard, line 113:
```suggestion
$releaseDateString = $ParsedReleaseDate.ToString("MM/dd/yyyy", [CultureInfo]::InvariantCulture)
```

weshaggard (approved):

Issue #7621, opened 2024-02-02
Failed to execute Update-GitHubIssue


Alancere:
It was successful before, but it's a problem now.

Update-GitHubIssue: https://github.com/Azure/azure-sdk-for-go/blob/main/eng/common/scripts/Invoke-GitHubAPI.ps1#L367

pipeline failure link: https://teams.microsoft.com/l/message/19:b76b7ede-5516-43b4-96c2-f580fa4309d7_eaa95837-fda2-47de-949f-654b6771e436@unq.gbl.spaces/1706860464853?context=%7B%22contextType%22%3A%22chat%22%7D

### curl: (successful)
```
curl -L -X PATCH -H "Authorization: Bearer $env:GITHUB_TOKEN" https://api.github.com/repos/Azure/azure-sdk-for-go/issues/22333 -d '{"state":"open","labels":["Beta"],"assignees":[]}
```
#### Failure: The given key 'Content-Type' was not present in the dictionary.
```
Invoke-RestMethod -Method PATCH -Body '{"state":"open","labels":["Beta"],"assignees":[]}' -Uri https://api.github.com/repos/Azure/azure-sdk-for-go/issues/22333 -Headers @{"Authorization" = "Bearer $env:GITHUB_TOKEN"} -MaximumRetryCount 3
```
we ran this test locally using latest powershell 7.4.1.
#### Adding Content-Type in Headers was successful (Content-Type: "application/json")
```
Invoke-RestMethod -Method PATCH -Body '{"state":"open","labels":["Beta"],"assignees":[]}' -Uri https://api.github.com/repos/Azure/azure-sdk-for-go/issues/22333 -Headers @{"Authorization" = "Bearer $env:GITHUB_TOKEN"; "Content-Type" = "application/json"} -MaximumRetryCount 3
```

raych1:
@weshaggard , do you know why this issue doesn't happen in eng sync pipeline? I haven't found the reason.

weshaggard:
I cannot open the teams link for some reason. Can you directly link the failing build so I can have a look? I know that the new PS version has started rolling out to our pipelines and it has causes some encoding issues.

Alancere:
[...]

PR #7630, opened 2024-02-05
Add ContentType header to Update-GitHubIssue

Alancere:
image

This PR is to fix the issue 
https://github.com/Azure/azure-sdk-tools/issues/7621#issue-2114362657

### DIFF
--- a/eng/common/scripts/Invoke-GitHubAPI.ps1
+++ b/eng/common/scripts/Invoke-GitHubAPI.ps1
@@ -402,7 +402,7 @@ function Update-GitHubIssue {
           -Body ($parameters | ConvertTo-Json) `
           -Uri $uri `
           -Headers (Get-GitHubApiHeaders -token $AuthToken) `
-          -MaximumRetryCount 3
+          -MaximumRetryCount 3 `
           -ContentType "application/json"
 }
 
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Invoke-GitHubAPI.ps1
```diff
@@ -403,6 +403,7 @@ function Update-GitHubIssue {
           -Uri $uri `
           -Headers (Get-GitHubApiHeaders -token $AuthToken) `
           -MaximumRetryCount 3
+          -ContentType "application/x-www-form-urlencoded"
```
raych1, line 406: Can we use `application/json` as the body is converted to JSON in line#402?

weshaggard (approved):

danieljurek (changes requested):
The current syntax will not pick up the `-ContentType` parameter because the script does not have a backtick to continue parsing to the next line.

weshaggard:
@raych1 Lets please be sure to test this change.

danieljurek (approved):

raych1:
> @raych1 Lets please be sure to test this change.

@weshaggard , this is the [test pipeline run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3471789&view=logs&j=33796ffc-d0c5-548c-0c1d-9d1ddb84594a&t=49452e74-a04a-59c1-118d-9b0749e7229c)

weshaggard:
@raych1 I don't suspect that is hitting this code path on PS7.4. Just please be sure to check the same pipeline that was failing in the go repo with this fix.

raych1:
> @raych1 I don't suspect that is hitting this code path on PS7.4. Just please be sure to check the same pipeline that was failing in the go repo with this fix.

@weshaggard , the failure was caught in the Go release pipeline, and we will verify it during this month's release.

PR #7635, opened 2024-02-05
[tsp-client] Fix tspconfig.yaml processing in init command

catalinaperalta:
- Fix a user reported bug related to issues when processing tspconfig.yaml:
  - We need to read the contents of the file to parse the data. 
  - Support passing in the directory that contains the tspconfig.yaml instead of the full path to the file. 
- Updating the package-lock.json which hadnt been done for a while 🙃

### DIFF
--- a/tools/tsp-client/src/index.ts
+++ b/tools/tsp-client/src/index.ts
@@ -42,7 +42,15 @@ async function sdkInit(
     const tspConfigPath = joinPaths(resolvedConfigUrl.path, "tspconfig.yaml");
     await addSpecFiles(cloneDir, tspConfigPath)
     await checkoutCommit(cloneDir, resolvedConfigUrl.commit);
-    const data = await readFile(joinPaths(cloneDir, tspConfigPath), "utf8");
+    let data;
+    try {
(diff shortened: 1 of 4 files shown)


tools/tsp-client/src/index.ts
```diff
@@ -63,8 +64,10 @@ async function sdkInit(
     return newPackageDir;
   } else {
     // Local directory scenario
-    let configFile = joinPaths(config, "tspconfig.yaml")
-    const data = await readFile(configFile, "utf8");
+    if (!config.endsWith("tspconfig.yaml")) {
+      config = joinPaths(config, "tspconfig.yaml");
```
joheredi, line 68: Do we get a meaningful error if the file doesn't exist? If we don't get an actionable error from node we may want to catch the exception and rise a custom error to help customers

joheredi (approved):

tools/tsp-client/src/index.ts
catalinaperalta: Yeah I didnt see a good one returned from node...it was a ENOENT error will catch is on the line below and raise a better error for customers. Thanks for pointing that out!

Issue #7636, opened 2024-02-05
[tsp-client] Improve sparse-spec clean up


catalinaperalta:
When we catch any errors during program execution we need to check if the sparse-spec/ directory wasn't delete it and remove it.

PR #7637, opened 2024-02-05
[tsp-client] Delete sparse-spec dir if it exists

catalinaperalta:
Adding a temporary workaround to delete the sparse-spec/ directory to reduce issues for folks who ran into errors during program execution. Have a tracking issue to do this at the end of the program instead: https://github.com/Azure/azure-sdk-tools/issues/7636

### DIFF
--- a/tools/tsp-client/src/index.ts
+++ b/tools/tsp-client/src/index.ts
@@ -3,7 +3,7 @@ import { createTempDirectory, removeDirectory, readTspLocation, getEmitterFromRe
 import { Logger, printBanner, enableDebug, printVersion } from "./log.js";
 import { TspLocation, compileTsp, discoverMainFile, resolveTspConfigUrl } from "./typespec.js";
 import { getOptions } from "./options.js";
-import { mkdir, writeFile, cp, readFile } from "node:fs/promises";
+import { mkdir, writeFile, cp, readFile, access } from "node:fs/promises";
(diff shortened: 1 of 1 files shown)


tools/tsp-client/src/index.ts
```diff
@@ -229,9 +229,17 @@ async function main() {
     rootUrl = resolvePath(options.outputDir);
   }
 
+  // FIXME: this is a workaround meanwhile we fix the issue with failing to delete the sparse-spec directory
+  // Tracking issue: https://github.com/Azure/azure-sdk-tools/issues/7636
+  const repoRoot = await getRepoRoot(rootUrl);
+  access(joinPaths(repoRoot, "..", "sparse-spec")).then(() => {
```
joheredi, line 235: Since this is already inside an async function, I would recommend using await and try/catch to keep it consistent and easier to read. Also, do we want to silently swallow the error with an empty catch?

joheredi: consider logging information that can help tsp-config customers send actionable issue reports (maybe there is nothing to log and that's fine, just wanted to mention this)

joheredi (approved):

tools/tsp-client/src/index.ts
catalinaperalta: sounds good, will switch to a try/catch. Also, since this is only meant to check if the sparse-spec/ directory exists if there are any other errors related to this, then I didnt want to throw on that. Also, I noticed we get an error if the directory doesnt exist which in this case is what we want, so we're ignoring the error.

PR #7639, opened 2024-02-06
Update githubissut 0205 0206

Alancere:
testting....


eng/pipelines/templates/stages/archetype-sdk-tool-repo-sync.yml
```diff
@@ -76,11 +76,6 @@ stages:
             runOnce:
               deploy:
                 steps:
-                  - template: /eng/common/pipelines/templates/steps/sparse-checkout.yml
```
raych1, line 79: Can you add `- checkout: self` explicitly?

PR #7651, opened 2024-02-07
Fix letter case of appSettings

RodgeFu:
1. fix letter case of appSettings for OS whose path is case sensitive
2. Upgrade to .net7


live1206 (approved):

tools/net-changelog-gen-mgmt/Azure.SDK.Management.ChangelogGen/Azure.SDK.Management.ChangelogGen.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
```
weshaggard, line 5: Why did you convert to .NET 7? We have been trying to keep our tooling on LTS which is .NET 6.

RodgeFu: just know we are trying to keep on .net 6. will revert it back to .net6. thanks.

PR #7678, opened 2024-02-13
New script to sync specs repo structure

JonathanCrd:
This new script will replace the Azure Function we use to sync the specs repos structure cache in OpenAPI Hub.


eng/pipelines/OpenApiSpecs/GetRepoTree.py
```diff
@@ -0,0 +1,118 @@
+from git import Repo
```
weshaggard, line 1: move this under `eng/scripts`

JonathanCrd: Moved. I also rename the file to be more descriptive.

JonathanCrd:
Azure DevOps Pipeline: [automation - tools - repo-structure-cache-job](https://dev.azure.com/azure-sdk/internal/_build?definitionId=6714&_a=summary)

eng/pipelines/OpenApiSpecs/repo-structure-cache-job.yml
```diff
@@ -0,0 +1,39 @@
...
+trigger:
+- main
+
+pool:
+  vmImage: ubuntu-latest
+
+steps:
+
+- script: |
+    pip install GitPython
```
scbedd, line 28: Pin `GitPython` version. or move these requirements to a text file with it pinned, then install the `requirements.txt` file using `pip install -r <file>`.

eng/pipelines/OpenApiSpecs/repo-structure-cache-job.yml
```diff
@@ -0,0 +1,39 @@
...
+  type: string
+  default: "openapi-portal"
+
+- name: COLLECTION_NAME
+  type: string
+  default: "repoStructureCache"
+
+- name: REPOS_URL_LIST
+  type: string
+  default: "https://github.com/test-repo-billy/azure-rest-api-specs/tree/main,https://github.com/test-repo-billy/azure-rest-api-specs-pr/tree/RPSaaSDev,https://github.com/test-repo-billy/azure-rest-api-specs-pr/tree/main,https://github.com/test-repo-billy/azure-rest-api-specs/tree/RPSaaSDev"
```
scbedd, line 17: I would normally recommend making this an actual list, type `object`, and then you can have

```yml
- name: REPOS_URL_LIST
  type: string
  default: 
  - https://github.com/test-repo-billy/azure-rest-api-specs/tree/main
  - https://github.com/test-repo-billy/azure-rest-api-specs-pr/tree/RPSaaSDev
  - https://github.com/test-repo-billy/azure-rest-api-specs-pr/tree/main
  - https://github.com/test-repo-billy/azure-rest-api-specs/tree/RPSaaSDev"
```

...but you pass this directly into a script further down, which would mean passing this to the script:

`REPOS_URL_LIST: '${{ convertToJson(parameters.REPOS_URL_LIST) }}'`

and then parsing it as json on the script side. I can see why you didn't, but just FYI.

eng/pipelines/OpenApiSpecs/repo-structure-cache-job.yml
```diff
@@ -0,0 +1,39 @@
...
+pool:
+  vmImage: ubuntu-latest
+
+steps:
+
+- script: |
+    pip install GitPython
+    pip install pymongo==3.13.0
+    pip list
+    python 'eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py'
```
scbedd, line 31: nit: From a usability standpoint it would probably be better to make the arguments actually arguments passed to the script, versus env settings.

eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py
```diff
@@ -0,0 +1,118 @@
...
+    repo_url: str = "https://{}@github.com/{}/{}".format(github_token, owner, repo_name)
+    clone_path: str = os.path.join(repo_clone_path, "{}{}".format(owner, repo_name))
+    new_document["commitSha"], new_document["repoStructure"] = get_repo_tree(
+        repo_url, clone_path, branch
+    )
+
+    return new_document
+
+
+def mongo_operations(
```
scbedd, line 57: `mongo_operations` is not a very descriptive function name 👍 

maybe `update_mongo_database` or something like that?

eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py
```diff
@@ -0,0 +1,118 @@
...
+    new_document: dict, DB_NAME: str, COLLECTION_NAME: str, CONNECTION_STRING: str
+) -> None:
+    print("Creating MongoDB client...")
+    client = MongoClient(CONNECTION_STRING)
+    db = client[DB_NAME]
+    collection = db[COLLECTION_NAME]
+
+    print("Using database '{}' and collection '{}'".format(DB_NAME, COLLECTION_NAME))
```
scbedd, line 65: Just FYI if you're on python 3.6 or newer, you can use `f` strings.

`print(f"Using database '{DB_NAME}' and collection '{COLLECTION_NAME}'")`

eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py
```diff
@@ -0,0 +1,118 @@
...
+
+    if collection.count_documents({"key": new_document["key"]}) == 0:
+        print("Document does not exist, inserting '{}'...".format(new_document["key"]))
+        collection.insert_one(new_document)
```
scbedd, line 69: wrap in `try: catch` w/ error reporting. You won't have seen anything locally, but CI is def gonna run into issues running this at some point.

eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py
```diff
@@ -0,0 +1,118 @@
...
+
+    if (
+        GITHUB_TOKEN is None
+        or REPOS_URL_LIST is None
+        or DB_NAME is None
+        or COLLECTION_NAME is None
+        or CONNECTION_STRING is None
+        or REPO_CLONE_PATH is None
+    ):
+        raise Exception("Environment variables not set")
```
scbedd, line 93: nit: I would raise a `RunTimeError` here.

eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py
```diff
@@ -0,0 +1,118 @@
...
+
+    repos: List[str] = REPOS_URL_LIST.split(",")
```
scbedd, line 95:
```suggestion
    repos: List[str] = [repo.strip() for repo in REPOS_URL_LIST.split(",")]
```

to eliminate easy errors down the line if you keep the `,` methodology for the argument.

eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py
```diff
@@ -0,0 +1,118 @@
...
+    tree: Repo.tree = repo.tree()
+    print("Repo cloned successfully")
+
+    repo_structure: dict = {}
+    for blob in tree:
+        if blob.name == "specification" and blob.type == "tree":
+            repo_structure[blob.name] = get_inner_blobs_structure(blob)
+
+    if "specification" not in repo_structure:
+        raise Exception("No 'specification' folder found in the repo.")
```
scbedd, line 36: nit: Probably `RunTimeError` here too. maybe [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError)?

eng/pipelines/OpenApiSpecs/repo-structure-cache-job.yml
```diff
@@ -0,0 +1,39 @@
...
+
+- name: REPOS_URL_LIST
+  type: string
+  default: "https://github.com/test-repo-billy/azure-rest-api-specs/tree/main,https://github.com/test-repo-billy/azure-rest-api-specs-pr/tree/RPSaaSDev,https://github.com/test-repo-billy/azure-rest-api-specs-pr/tree/main,https://github.com/test-repo-billy/azure-rest-api-specs/tree/RPSaaSDev"
+
+trigger:
+- main
+
+pool:
+  vmImage: ubuntu-latest
```
weshaggard, line 23: We should use our image pools. Example https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/githubio-linkcheck.yml#L5-L6

eng/pipelines/OpenApiSpecs/repo-structure-cache-job.yml
```diff
@@ -0,0 +1,39 @@
+# Starter pipeline
```
weshaggard, line 1: Lets move this directly into the eng/pipelines directory and name change the name slightly. I'm assuming this is only about caching the spec information so perhaps `cache-specs-in-spec-repo.yml` might be slightly better name to represent what it is happening.

eng/pipelines/OpenApiSpecs/repo-structure-cache-job.yml
```diff
@@ -0,0 +1,39 @@
...
+- script: |
+    pip install GitPython
+    pip install pymongo==3.13.0
+    pip list
+    python 'eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py'
+  displayName: Run Script
+  env: 
+    MONGO_CONNECTION_STRING: $(Open-API-MongoDB-Connection-String)
+    GITHUB_TOKEN: $(azuresdk-github-pat)
+    REPO__CLONE_PATH: $(System.DefaultWorkingDirectory)
```
weshaggard, line 36: Why is there a double underscore here?

eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py
```diff
@@ -0,0 +1,118 @@
...
+    print("\033[92mDocument saved\033[0m")
+
+
+def main() -> None:
+    GITHUB_TOKEN: str = os.environ.get("GITHUB_TOKEN")
+    REPOS_URL_LIST: str = os.environ.get("REPOS_URL_LIST")
+    DB_NAME: str = os.environ.get("DB_NAME")
+    COLLECTION_NAME: str = os.environ.get("COLLECTION_NAME")
+    CONNECTION_STRING: str = os.environ.get("MONGO_CONNECTION_STRING")
+    REPO_CLONE_PATH: str = os.getenv("REPO__CLONE_PATH")
```
weshaggard, line 83: These should be parameters instead of env variables. Having these all be env variables makes running this script locally a lot more difficult.

eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py
JonathanCrd: Totally! I didn't know about this.
I got rid of all `.format` and used f string instead

eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py
JonathanCrd: Based on the description of the docs, I wouldn't go with `ValueError` since this doesn't happen in an argument. The `repo_structure` is calculated based on the cloned repo. If the cloned repo doesn't have an "specification" directory, it would throw the error. 

I'm going to go with `RunTimeError` here.

eng/scripts/OpenAPI Hub/sync-repo-structure-cache.py
JonathanCrd: Good point.
I wrapped each mongo operation inside a `try except`, so we can point which one failed.

eng/pipelines/OpenApiSpecs/repo-structure-cache-job.yml
JonathanCrd: This was a typo.
I'm removing this from the env vars since now it is an argument as suggested in your other comment.

praveenkuttappan (approved):
Looks good.

Issue #7057, opened 2023-10-02
[tsp-client] Add support for lock files


catalinaperalta:
See comment: https://github.com/Azure/azure-sdk-tools/pull/6843#discussion_r1303624591

PR #7680, opened 2024-02-14
[tsp-client] Add support for package-lock.json files

catalinaperalta:
Fixes https://github.com/Azure/azure-sdk-tools/issues/7057

### DIFF
--- a/tools/tsp-client/src/index.ts
+++ b/tools/tsp-client/src/index.ts
@@ -1,9 +1,9 @@
-import { npmCIDependencies, npmInstallDependencies } from "./npm.js";
+import { npmInstallationCommand } from "./npm.js";
 import { createTempDirectory, removeDirectory, readTspLocation, getEmitterFromRepoConfig } from "./fs.js";
 import { Logger, printBanner, enableDebug, printVersion } from "./log.js";
 import { TspLocation, compileTsp, discoverMainFile, resolveTspConfigUrl } from "./typespec.js";
 import { getOptions } from "./options.js";
-import { mkdir, writeFile, cp, readFile, access } from "node:fs/promises";
+import { mkdir, writeFile, cp, readFile, access, stat } from "node:fs/promises";
 import { addSpecFiles, checkoutCommit, cloneRepo, getRepoRoot, sparseCheckout } from "./git.js";
 import { doesFileExist } from "./network.js";
 import { parse as parseYaml } from "yaml";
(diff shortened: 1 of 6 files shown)


tools/tsp-client/src/index.ts
```diff
@@ -172,8 +172,18 @@ async function syncTspFiles(outputDir: string, localSpecRepo?: string) {
     await removeDirectory(cloneDir);
   }
 
-  const emitterPath = joinPaths(repoRoot, "eng", "emitter-package.json");
-  await cp(emitterPath, joinPaths(srcDir, "package.json"), { recursive: true });
+  try {
+    const emitterLockPath = joinPaths(repoRoot, "eng", "emitter-package-lock.json");
+    await cp(emitterLockPath, joinPaths(srcDir, "package-lock.json"), { recursive: true });
+  } catch (err) {
+    Logger.debug("emitter-package-lock.json not found, will look for emitter-package.json");
```
joheredi, line 179: There could be other reasons (maybe not too common) why copy failed, for example if the disk is full or for some reason `srcDir` doesn't exist. If users fall into those situations, this log may be missleading

tools/tsp-client/src/index.ts
```diff
@@ -201,8 +211,19 @@ async function generate({
   const mainFilePath = await discoverMainFile(srcDir);
   const resolvedMainFilePath = joinPaths(srcDir, mainFilePath);
   Logger.info("Installing dependencies from npm...");
-  await installDependencies(srcDir);
-
+  const args: string[] = [];
+  try {
+    // Check is package-lock.json exists, if it does, we'll install dependencies through `npm ci`
```
joheredi, line 216:
```suggestion
    // Check if package-lock.json exists, if it does, we'll install dependencies through `npm ci`
```

tools/tsp-client/src/npm.ts
```diff
@@ -19,9 +19,9 @@ export async function createPackageJson(rootPath: string, deps: Set<string>): Pr
   await writeFile(filePath, packageJson);
 }
 
-export function installDependencies(workingDir: string): Promise<void> {
+export async function npmInstallationCommand(workingDir: string, args: string[]): Promise<void> {
```
joheredi, line 22: Is it fair to say that this function became `npmCommand` with the change of taking args and is no longer scoped to the install command?

joheredi: If so, maybe we should rename

joheredi (approved):
Left a few minor comments

tools/tsp-client/src/index.ts
catalinaperalta: That's true, I updated the log to include the original error but still wont fail in this case if we have an emitter-package.json. If the issue happens with emitter-package.json then we will fail

tools/tsp-client/src/npm.ts
catalinaperalta: Renamed

PR #7682, opened 2024-02-14
Support resolving environment variable references in matrix config

benbp:
_sorry for all the formatting updates, vscode got more opinionated this time and I don't want to fight it_

This adds the ability to reference environment variables to be resolved at runtime for matrix values, for example:

```
"matrix": {
  "foo": "env:bar"
}
```

The primary use case is so we can reference common values like `LinuxPool` that we define as top level variables in an azure pipelines context, so we don't need to repetitively hardcode them in all the matrix configs like we do today.

### DIFF
--- a/eng/common/scripts/job-matrix/job-matrix-functions.ps1
+++ b/eng/common/scripts/job-matrix/job-matrix-functions.ps1
@@ -388,7 +388,7 @@ function ProcessEnvironmentVariableReferences([array]$matrix, $displayNamesLooku
                     if (!$value) {
                         Write-Warning "Environment variable `"$envKey`" was not found or is empty."
                     }
-                    $flattened.Set($value, $flattened.Name)
+                    $perm.Set($value, $flattened.Name)
                 }
             }
 
--- a/eng/common/scripts/job-matrix/job-matrix-functions.ps1
+++ b/eng/common/scripts/job-matrix/job-matrix-functions.ps1
@@ -76,7 +76,7 @@ class MatrixParameter {
             $displayName = $this.Value.ToString()
         }
 
-        if ($displayNamesLookup.ContainsKey($displayName)) {
+        if ($displayNamesLookup -and $displayNamesLookup.ContainsKey($displayName)) {
             $displayName = $displayNamesLookup[$displayName]
         }
 
@@ -339,6 +339,9 @@ function ProcessReplace {
 
     foreach ($element in $matrix) {
         $replacement = [MatrixParameter[]]@()
+        if (!$element -or $element.Count -eq 0) {
+            continue
+        }
 
         foreach ($perm in $element._permutation) {
             $replace = $perm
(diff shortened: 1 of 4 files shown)


eng/common/scripts/job-matrix/job-matrix-functions.ps1
```diff
@@ -365,18 +367,46 @@ function ProcessReplace
...
             return $_
-        } else {
+        }
+        else {
             $importPath = $_.Value
         }
     }
     if ((!$matrix -and !$importPath) -or !$importPath) {
-        return $matrix, @(), @{}
+        return $matrix, @(), $displayNamesLookup
```
benbp, line 409: Also fixed a separate bug here where display names lookup was getting lost if we didn't have imports

scbedd (approved):
You have a test-branch where you're actually trying this, but this code update looks fine to me. Nothing is jumping out at me 😆

benbp:
@mikeharder for example, this is what I'm thinking of switching our matrix configs to:

```
"matrix: {
  "Agent": {
    "ubuntu": {
      "Pool": "env:LinuxPool",
      "Demands": "env:LinuxImageDemands"
      "osName": "env:LinuxOS"
    }
  }
}
```

Then we can define all these values in a single place like `image.yml` referenced by `eng/pipelines/templates/variables/globals.yml`

eng/common/scripts/job-matrix/job-matrix-functions.ps1
```diff
@@ -365,18 +370,49 @@ function ProcessReplace
...
+    foreach ($element in $matrix) {
+        if (!$element || [bool]$element.PSobject.Properties.name -notmatch "_permutation") {
+            continue
+        }
+        $updated = [MatrixParameter[]]@()
+
+        foreach ($perm in $element._permutation) {
+            # Iterate nested permutations or run once for singular values (int, string, bool)
+            foreach ($flattened in $perm.Flatten()) {
+                if ($flattened.Value?.GetType() -eq "".GetType() -and $flattened.Value.StartsWith("env:")) {
```
weshaggard, line 385: What is `"".GetType()` about? If you are just checking if it is a string wny not just `$flattened.Value -is [string]`

eng/common/scripts/job-matrix/job-matrix-functions.ps1
```diff
@@ -365,18 +370,49 @@ function ProcessReplace
...
+                    $envKey = $flattened.Value.Replace("env:", "")
```
weshaggard, line 386: Couldn't you use `Get-Item $flattend.Value`" I think that would get you the name and value.

weshaggard (approved):
couple suggestions but looks good otherwise.

eng/common/scripts/job-matrix/job-matrix-functions.ps1
```diff
@@ -365,18 +370,49 @@ function ProcessReplace
...
+                    $value = [System.Environment]::GetEnvironmentVariable($envKey) ?? ""
```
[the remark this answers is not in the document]
benbp: To me the code is a bit less clear if you don't know how get-item works, and there could be issues in the future if the outer conditional gets removed, allowing file lookups, etc.

weshaggard:
Closing this for now as we don't need this feature currently.

eng/common/scripts/job-matrix/job-matrix-functions.ps1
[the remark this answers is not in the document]
weshaggard: I was just trying to avoid us needing to parse variables when PS already does that. In this case it is pretty simple so I'm not too worried about splitting and just getting the env variable directly. I'll leave it to you :)

PR #7683, opened 2024-02-14
[.NET Analyzers] Fix Template special case

jsquire:
# Summary

The focus of these changes is to move the template special case logic from an equality check to a `StartsWith`, matching the behavior of the other checks.  By design, this allows subnamespaces of `Azure.Template` which are needed due to the structure of the Template project.

### DIFF
--- a/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAssemblyNamespaceAnalyzer.cs
+++ b/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAssemblyNamespaceAnalyzer.cs
@@ -78,7 +78,7 @@ namespace Azure.ClientSdk.Analyzers
                 // "Azure.Template" is not an approved namespace prefix, but we have a project template by that name
                 // to help customers get started. We do not want our template to include a suppression for this
                 // descriptor out of the box, so we need to treat it as a special case.
-                if (displayString == "Azure.Template")
+                if (displayString.StartsWith("Azure.Template"))
                 {
                     return;
                 }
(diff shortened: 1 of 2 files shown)


src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAssemblyNamespaceAnalyzer.cs
```diff
@@ -78,7 +78,7 @@ public override void Analyze(ISymbolAnalysisContext context)
                 // "Azure.Template" is not an approved namespace prefix, but we have a project template by that name
                 // to help customers get started. We do not want our template to include a suppression for this
                 // descriptor out of the box, so we need to treat it as a special case.
-                if (displayString == "Azure.Template")
+                if (displayString.StartsWith("Azure.Template"))
```
heaths, line 81: I recommend:
```suggestion
                if (displayString.StartsWith("Azure.Template."))
```

That way it doesn't allow "Azure.TemplateHaHaGotYouJesse.MyService" namespaces. Test should be added to check.

heaths (approved):

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientAssemblyNamespaceAnalyzer.cs
heaths: You should add a test for the negative, too.

christothes: Do we still need to match the bare `Azure.Template` ? If so we'd need to or the existing check with the `.StartsWith("Azure.Template.")` (or regex it)

jsquire: Yeah, I think we'll need an equality check for the exact text and a `StartsWith` followed by the `.`

jsquire: I've got both patterns covered now and some additional negative tests around the case that Heath pointed out.

christothes (approved):

PR #7688, opened 2024-02-15
Stress: adding in a wrapper image for the otel collector so we can use it with our stress cluster

richardpark-msft:
This creates a Docker image that wraps the actual otel collector in a container which can pull the proper env variable and write out a config file.

Note, this is dependent on a fix that injects the .env file into all containers in a pod, not just one, but can be built and pushed independent of that work.

CC: @LarryOsterman

### DIFF
--- a/tools/stress-cluster/services/otelcollector/otel-collector-config.yml
+++ b/tools/stress-cluster/services/otelcollector/otel-collector-config.yml
@@ -2,7 +2,6 @@ receivers:
   otlp:
     protocols:
       http:
-      grpc:
  
 processors:
   batch:
@@ -16,18 +15,14 @@ service:
     metrics:
       receivers: [otlp]
       processors: [batch]
-      exporters: [prometheus, azuremonitor]
+      exporters: [azuremonitor]
     logs:
       receivers: [otlp]
       processors: [batch]
       exporters: [azuremonitor]
 
 exporters:
-  prometheus:
-    endpoint: "0.0.0.0:8889"
-    resource_to_telemetry_conversion: 
-      enabled: true
   logging:
-    verbosity: detailed
+    verbosity: normal
   azuremonitor:
 #    connection_string: "%APPLICATIONINSIGHTS_CONNECTION_STRING%"
--- a/tools/stress-cluster/services/otelcollector/otel-collector-config.yml
+++ b/tools/stress-cluster/services/otelcollector/otel-collector-config.yml
@@ -1,0 +1,33 @@
+receivers:
+  otlp:
+    protocols:
+      http:
+      grpc:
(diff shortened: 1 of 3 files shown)


tools/stress-cluster/services/otelcollector/otel-collector-config.yml
```diff
@@ -0,0 +1,33 @@
...
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [logging, azuremonitor]
+    metrics:
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [prometheus, azuremonitor]
```
LarryOsterman, line 19: The prometheus exporter is not configured, and should probably not be used for metrics.

tools/stress-cluster/services/otelcollector/otel-collector-config.yml
```diff
@@ -0,0 +1,33 @@
...
+    logs:
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [azuremonitor]
+
+exporters:
+  prometheus:
```
LarryOsterman, line 26: Remove references to the prometheus exporter.

tools/stress-cluster/services/otelcollector/otel-collector-config.yml
```diff
@@ -0,0 +1,33 @@
...
+    endpoint: "0.0.0.0:8889"
+    resource_to_telemetry_conversion: 
+      enabled: true
+  logging:
+    verbosity: detailed
```
LarryOsterman, line 31:
```suggestion
    verbosity: normal
```

This reduces the noise level in the collector dramatically.

tools/stress-cluster/services/otelcollector/otel-collector-config.yml
LarryOsterman: Unless we're also configuring a prometheus service client in our stress cluster, or have configured prometheus for stress metrics.

richardpark-msft: No, we're not (and we could always bring this back later).

tools/stress-cluster/services/otelcollector/otel-collector-config.yml
richardpark-msft: Done!

tools/stress-cluster/services/otelcollector/otel-collector-config.yml
richardpark-msft: Removed!

benbp (approved):

PR #7698, opened 2024-02-16
[http-fault-injector] Add support for interrupting request

### LINKED ISSUE #7545: [http-fault-injector] Add support for interrupting request with TCP RST
> There isn’t currently support for triggering an RST on receiving a request or in the middle of a request body. It’s currently only supported for triggering before a response or mid response body.

If I understand correctly, I think this should involve:

1. Add a custom middleware that can abort requests (TCP FIN or RST)
2. Interactive mode:
   1. Prompt the user at the start of receiving the request for whether to abort immediately, abort mid body, or continue.
3. Automated mode:
   1. For aborting in the middle of a request body, use a request header
   2. For aborting before even reading the request headers, it might work read something from the URL, like "?faultinjector-response-option=<option>"
 
If you need to abort even earlier (like when the TCP connection is established), it's probably going to start getting more complicated, including how to control the behavior.

CC: @ibrahimrabab

mikeharder:
- Fixes #7545

The changes under `Azure.Sdk.Tools.HttpFaultInjector` deserve more scrutiny.  The changes under `sample-[clients|servers]` need less scrutiny since they are just samples.


tools/http-fault-injector/Azure.Sdk.Tools.HttpFaultInjector/FaultInjectingMiddleware.cs
```diff
@@ -67,13 +67,10 @@ private async Task<UpstreamResponse> SendUpstreamRequest(HttpRequest request, st
 
             using (var upstreamRequest = new HttpRequestMessage(new HttpMethod(request.Method), upstreamUri))
             {
```
mikeharder, line 69: Should work even for requests with no body, or requests with chunked encoding.  Shouldn't need to gate on content-length header.

ibrahimrabab (approved):

tools/http-fault-injector/Dockerfile
```diff
@@ -13,6 +13,8 @@ RUN dotnet dev-certs https
 EXPOSE 7777
 EXPOSE 7778
 
+ENV ASPNETCORE_URLS=http://+:7777;https://+:7778
```
mikeharder, line 16: These are already set in `appsettings.json`, but when we tried to run the docker container, it ws using the default ports of 5000 and 5001 instead.  This seems like a reasonable setting if we want to ensure ASP.NET uses these ports.

mikeharder:
@lmolkova: I'd like for you to review this before I merge.

tools/http-fault-injector/sample-servers/net/http-server/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+                {
+                    options.Listen(IPAddress.Any, 5000);
+                    options.Listen(IPAddress.Any, 5001, listenOptions =>
+                    {
+                        listenOptions.UseHttps("testCert.pfx", "testPassword");
+                    });
+                })
+                .Configure(app => app.Run(async context =>
+                {
+                    Console.WriteLine($"Request: {context.Request.Path}");
```
lmolkova, line 25: nit: you can still do [`app.UseHttpLogging()`](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/http-logging/?view=aspnetcore-8.0) to have logging configurable, controllable and exportable

lmolkova (approved):
LGTM!

PR #7705, opened 2024-02-16
Update common publish task for 1es template pipeline

benbp:


konrad-jamrozik (approved):

eng/common/pipelines/templates/steps/publish-artifact.yml
```diff
@@ -10,18 +10,25 @@ parameters:
...
   ArtifactPath: ''
   CustomCondition: true
+  DisplayName: ''
 
 steps:
-  - task: PublishPipelineArtifact@1
+  - task: 1ES.PublishPipelineArtifact@1
     condition: and(succeeded(), ${{ parameters.CustomCondition }})
-    displayName: 'Publish ${{ parameters.ArtifactName }} Artifacts'
+    ${{ if parameters.DisplayName }}:
```
weshaggard, line 18: Is this necessary? Do we really need a custom display name? Or perhaps we can set the DisplayName to a default value in the parameter definition to avoid the need for an extra condition here.

benbp: I was finding several places where we were doing publishes in the pipeline that had descriptive names that I thought would be useful to preserve. The conflict is preserving the "failed" description, which I also think was useful. I think we have to pick one or the other.

benbp: @weshaggard updated a bit to remove the conditional

benbp:
closing in favor of https://github.com/Azure/azure-sdk-tools/pull/7758

PR #7733, opened 2024-02-22
Fix errors in repo structure cache script and improve pipeline

JonathanCrd:
In this PR I fixed errors found when running the script against PPE.

Also, I had a chat with @weshaggard and we decided to use the same pipeline for both PPE and PROD environments. I'm separating the pipeline into three steps:

1. Install dependencies
2. Update PROD cache
3. Update PPE cache

Also, following Wes's advice, I removed the parameters for the pipelines since it is not expected for them to be changed. Removing them will reduce room for error when running this pipeline.

### DIFF
--- a/eng/pipelines/openapitools/repo-structure-cache-job.yml
+++ b/eng/pipelines/openapitools/repo-structure-cache-job.yml
@@ -1,30 +1,26 @@
-parameters:
-- name: DB_NAME
-  type: string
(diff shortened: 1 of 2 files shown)


praveenkuttappan (approved):

JonathanCrd:
PPE cache was updated succesfully

PR #7754, opened 2024-02-26
Change the way ServiceLabel parses with AzureSdkOwners

JimSuplizio:
This is a change requested by @jsquire for the case where a ServiceLabel only has AzureSdkOwners with no ServiceOwners.

Previously, ServiceLabel required service owners in whatever shape that took (either ServiceOwners/NotInRepo or the block ended in a source path/owner line).

With this change, ServiceLabel still **_requires_** service owners unless the block only consists of ServiceLabel and AzureSdkOwners.

To be very clear what this allows, the following is now a valid block:
**AzureSdkOwners:** @owner1 @owner2
**ServiceLabel:** %SomeLabel

This does mean that the owners of CODEOWNERS files, whomever they might actually be, need to pay attention to changes to ensure that if a ServiceLabel requires both AzureSdkOwners and ServiceOwners that they're both there.


tools/codeowners-utils/METADATA.md
```diff
@@ -10,7 +10,7 @@ Azure-sdk* repository CODEOWNERS files contain data beyond the normal source pat
...
-- **ServiceLabel:** - This moniker contains the service label that's used to figure out what users need to be @ mentioned in an issue when the Service Attention label is added. This moniker must be part of a block that either ends in a source path/owner line, the ServiceOwners moniker or the /&lt;NotInRepo&gt;/ moniker. If the ServiceLabel is part of a block that ends in the source path/owner line, the service owners are inferred from that.
+- **ServiceLabel:** - This moniker contains the service label that's used to figure out what users should be assigned at triage to own issues,, whether part of the Azure SDK team or service partners.  If the label is owned by a service partner or has a set of service owners included, these users would be @ mentioned in an issue when the Service Attention label is added. This moniker must be part of a block that either ends in a source path/owner line, the AzureSdkOwners moniker, ServiceOwners moniker, or the /&lt;NotInRepo&gt;/ moniker. If the ServiceLabel is part of a block that ends in the source path/owner line, the service owners are inferred from that.
```
weshaggard, line 13:
```suggestion
- **ServiceLabel:** - This moniker contains the service label that's used to figure out what users should be assigned at triage to own issues, whether part of the Azure SDK team or service partners. If the label is owned by a service partner or has a set of service owners included, these users would be `@` mentioned in an issue when the Service Attention label is added. This moniker must be part of a block that either ends in a source path/owner line, the AzureSdkOwners moniker, ServiceOwners moniker, or the /&lt;NotInRepo&gt;/ moniker. If the ServiceLabel is part of a block that ends in the source path/owner line, the service owners are inferred from that.
```

weshaggard (approved):

PR #7758, opened 2024-02-26
Add 1es changes for job/matrix generation and publish

benbp:
Adding these files as separate copies of their originals so we can do a rollout more easily.

1. Update `archetype-sdk-tests-generate.yml` to `generate-job-matrix.yml`. Changes to this file include generating a matrix with corresponding env vars for OS pools/agents, and passing in the now required `OSName` parameter which cannot be set as a variable for the 1es templates.
2. Moving `publish-artifact.yml` to `publish-1es-artifact.yml` to use the 1es publishing wrapper task. I don't think this task is available in a backwards compatible way without the repo reference and/or onboarding we're having to do, see https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3503459&view=logs&j=96791242-dbf3-587e-3a06-ae5af5c1a705&t=8e5c4a57-b21e-59c3-59e8-95279bb1be65&l=16 for example.

### DIFF
--- a/eng/common/pipelines/templates/jobs/generate-job-matrix.yml
+++ b/eng/common/pipelines/templates/jobs/generate-job-matrix.yml
@@ -46,13 +46,14 @@ parameters:
 - name: Pools
   type: object
   default:
-    - name: LinuxPool
+    - name: Linux
+      filter: .*Linux.*Pool$
       os: linux
-    - name: LinuxNextPool
-      os: linux
-    - name: WindowsPool
+    - name: Windows
+      filter: .*Windows.*Pool$
       os: windows
-    - name: MacPool
+    - name: Mac
+      filter: .*MacPool$
       os: macOS
 
 jobs:
@@ -94,7 +95,7 @@ jobs:
                 -ConfigPath ${{ config.Path }}
                 -Selection ${{ config.Selection }}
                 -DisplayNameFilter '$(displayNameFilter)'
-                -Filters '${{ join(''',''', parameters.MatrixFilters) }}', 'container=^$', 'SupportedClouds=^$|${{ parameters.CloudConfig.Cloud }}', 'Pool=.*${{ pool.name }}$'
+                -Filters '${{ join(''',''', parameters.MatrixFilters) }}', 'container=^$', 'SupportedClouds=^$|${{ parameters.CloudConfig.Cloud }}', 'Pool=${{ pool.filter }}'
                 -Replace '${{ join(''',''', parameters.MatrixReplace) }}'
                 -NonSparseParameters '${{ join(''',''', config.NonSparseParameters) }}'
             displayName: Create ${{ pool.name }} Matrix ${{ config.Name }}
@@ -109,7 +110,7 @@ jobs:
                 -ConfigPath ${{ config.Path }}
                 -Selection ${{ config.Selection }}
                 -DisplayNameFilter '$(displayNameFilter)'
-                -Filters '${{ join(''',''', parameters.MatrixFilters) }}', 'container=^$', 'SupportedClouds=^$|${{ parameters.CloudConfig.Cloud }}', 'Pool=.*${{ pool.name }}$'
+                -Filters '${{ join(''',''', parameters.MatrixFilters) }}', 'container=^$', 'SupportedClouds=^$|${{ parameters.CloudConfig.Cloud }}', 'Pool=${{ pool.filter }}'
                 -NonSparseParameters '${{ join(''',''', config.NonSparseParameters) }}'
             displayName: Create ${{ pool.name }} Container Matrix ${{ config.Name }}
             name: container_job_matrix_${{ config.Name }}_${{ pool.name }}
--- a/eng/common/pipelines/templates/jobs/generate-job-matrix.yml
+++ b/eng/common/pipelines/templates/jobs/generate-job-matrix.yml
@@ -1,0 +1,139 @@
+parameters:
+- name: AdditionalParameters
+  type: object
+- name: DependsOn
+  type: object
+  default: null
+- name: CloudConfig
+  type: object
+  default: {}
+- name: MatrixConfigs
+  type: object
+  default: []
+- name: MatrixFilters
+  type: object
+  default: []
+- name: MatrixReplace
+  type: object
+  default: {}
+- name: JobTemplatePath
+  type: string
+# Set this to false to do a full checkout for private repositories with the azure pipelines service connection
+- name: SparseCheckout
+  type: boolean
+  default: true
+- name: SparseCheckoutPaths
+  type: object
+  default: []
+- name: Pool
+  type: string
+  default: azsdk-pool-mms-ubuntu-2204-general
+- name: OsVmImage
+  type: string
+  default: ubuntu-22.04
+- name: Os
+  type: string
+  default: linux
+# This parameter is only necessary if there are multiple invocations of this template within the SAME STAGE.
+# When that occurs, provide a name other than the default value.
+- name: GenerateJobName
+  type: string
+  default: 'generate_job_matrix'
+- name: PreGenerationSteps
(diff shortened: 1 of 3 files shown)


eng/common/pipelines/templates/jobs/generate-job-matrix.yml
```diff
@@ -0,0 +1,139 @@
...
+      os: linux
+    - name: WindowsPool
+      os: windows
+    - name: MacPool
+      os: macOS
+
+jobs:
+- job: ${{ parameters.GenerateJobName }}
+  variables:
+    - template: /eng/pipelines/templates/variables/image.yml
```
scbedd, line 61: `image.yml` should be part of this PR I think.

benbp: Originally the idea was to have image.yml be repo specific, but it is a hard requirement since this has a reference.

scbedd (approved):
Based on offline discussion about `image.yml`, signing off.

eng/common/pipelines/templates/jobs/generate-job-matrix.yml
benbp: @weshaggard maybe this is something we do want in eng/common just to avoid hidden file dependencies in our templates. However if we wanted to roll out image changes per-repo it would not work.

weshaggard: I don't mind having a default one in eng\common and perhaps we can come up with a way to override the image in the language repo if needed. We could even make this a default template parameter that could be set to something as needed.

eng/common/pipelines/templates/jobs/generate-job-matrix.yml
```diff
@@ -0,0 +1,139 @@
...
+- name: SparseCheckout
+  type: boolean
+  default: true
+- name: SparseCheckoutPaths
+  type: object
+  default: []
+- name: Pool
+  type: string
+  default: azsdk-pool-mms-ubuntu-2204-general
+- name: OsVmImage
```
weshaggard, line 31: Is this still interesting after the migration?

weshaggard: Similar question for the Pool parameter.

weshaggard (approved):
A couple questions but otherwise seems fine

eng/common/pipelines/templates/steps/publish-1es-artifact.yml
```diff
@@ -0,0 +1,34 @@
...
+parameters:
+  ArtifactName: ''
+  ArtifactPath: ''
+  CustomCondition: true
+  DisplayName: ''
+
+steps:
+  - task: 1ES.PublishPipelineArtifact@1
+    condition: and(succeeded(), ${{ parameters.CustomCondition }})
+    ${{ if parameters.DisplayName }}:
```
weshaggard, line 18: As mentioned in the other PR I'd suggest just moving to DisplayName everywhere and for the existing places just map them to call `DisplayName: Publish ArtifactName Artifacts" if we really care about the string being the same. We could also potentially use a `coalesce` but that might not work as well for the failed condition.

eng/common/pipelines/templates/jobs/generate-job-matrix.yml
benbp: Correct me if I'm wrong @scbedd, but we found demands does not work for the Azure Pipelines mac pool, and we still need to pass `vmImage` for that. Additionally for the migration we'll want to use the main pools and specify the alternate image.

scbedd: > Correct me if I'm wrong @scbedd, but we found demands does not work for the Azure Pipelines mac pool, and we still need to pass `vmImage` for that. Additionally for the migration we'll want to use the main pools and specify the alternate image.

Correct. The specific combo of

```
pool:
  name: 
  vmImage: 
  os: linux|windows|macOS
```

works for both devops + the additional agent checks. I swear I tried everything else.

eng/common/pipelines/templates/jobs/generate-job-matrix.yml
benbp: One way would be to have the underlying values get set via coalesce e.g. `$[ coalesce(variables.LinuxPoolOverride, variables.LinuxPool) ]` but I'm not too excited about the config variable bloat of doubling all the values with `*Override`.

eng/common/pipelines/templates/steps/publish-1es-artifact.yml
benbp: Updated a bit to remove the conditionals. I think keeping the naming for now is helpful but we can go in and remove custom ones after.

weshaggard: Looks like you updated the publish-artifacts.yml file but not this one. Do you plan to keep both for the transition or did you edit the wrong one.

benbp:
Thanks for catching that, fixed.

eng/common/pipelines/templates/steps/publish-1es-artifact.yml
benbp: Thanks for catching that, fixed.

eng/common/pipelines/templates/steps/publish-1es-artifact.yml
```diff
@@ -0,0 +1,30 @@
...
+    condition: and(succeeded(), ${{ parameters.CustomCondition }})
+    displayName: Set Artifact Name
+
+  - pwsh: |
+      Write-Host "##vso[task.setvariable variable=PublishArtifactName;]${{ parameters.ArtifactName }}-FailedAttempt$(System.JobAttempt)"
+    condition: and(failed(), ${{ parameters.CustomCondition }})
+    displayName: Set Failed Artifact Name
+
+  - task: 1ES.PublishPipelineArtifact@1
+    condition: and(succeeded(), ${{ parameters.CustomCondition }})
```
weshaggard, line 26: Shouldn't we remove or at least update this condition to be failedOrSucceeded()?

benbp: Ah, yes, good catch. I'll do another PR. FYI @scbedd but shouldn't be necessary for your testing.

benbp: https://github.com/Azure/azure-sdk-tools/pull/7810

eng/common/pipelines/templates/steps/publish-1es-artifact.yml
```diff
@@ -0,0 +1,30 @@
...
+# CustomCondition - Used if there is additional logic necessary to prevent attempt of publish.
+
+parameters:
+  ArtifactName: ''
+  ArtifactPath: ''
+  CustomCondition: true
+
+steps:
+  - pwsh: |
+      Write-Host "##vso[task.setvariable variable=PublishArtifactName;]${{ parameters.ArtifactName }}"
```
weshaggard, line 16: To help reduce the number of injected steps can we use `$env:AGENT_JOBSTATUS -eq "Failed"` see https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml for definition of that variable. https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#failed

eng/common/pipelines/templates/steps/publish-1es-artifact.yml
weshaggard: I would suggest succededOrFailed here as well in your pwsh step to set the artifact name. https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#succeededorfailed

PR #7778, opened 2024-02-28
[APIView] Update local development docs

maorleger:
Local APIView development requires an app config instance to connect to. For basic scenarios
it does not require any values, so the documentation was updated to add the minimal config
required


maririos (approved):

maririos:
FYI @WhitShake

PR #7793, opened 2024-02-29
remove autorest.python from label syncing

iscai-msft:
We want to remove `autorest.python` from the label syncing because we have labels that are very different from the rest of our sdk repos. We would like to keep only our small list of labels for greater labeling clarity in the `autorest.python` repo, and not have a bunch of labels synced into our repo that pollute the label namespace and are of no use to us

### DIFF
--- a/tools/github/data/repositories.txt
+++ b/tools/github/data/repositories.txt
@@ -24,7 +24,6 @@ Azure/azure-sdk-actions
 Azure/autorest
 Azure/autorest.csharp
 Azure/autorest.java
-Azure/autorest.python
 Azure/autorest.typescript
 Azure/autorest.go
 Azure/autorest.swift


jsquire:
The AutoRest repositories were added to the common set by leadership request for automation, metrics, and reporting purposes.  Removing this from common label sync is likely to break parts of automation going forward.

I have no objection, but this decision is up to @ronniegeraghty, as the owner of repository process and reporting.

ronniegeraghty:
Sorry for the delay. I must have missed the notification telling me I was mentioned here. 
The autorest repos are not currently integrated into any of our repo tracking, but that's not to say they shouldn't be or won't ever be. Let me ask around to see if tracking customer interactions in this repo is something we want to do.

ronniegeraghty (approved):

PR #7806, opened 2024-03-02
Move sparse checkout to downloadable script

benbp:
With the move to 1es pipeline templates, some of our pipelines are again exceeding the azure pipelines max template size (4mb??). I found that about 3k lines of our 9k lines in core (pre 1es pt) is due to sparse checkout inlining duplication. We still have a chicken/egg problem so the script can't be referenced via file, but we CAN download the script from github raw content to save a bit of space. Another benefit is our checkout logic is finally in a script that other people can call if they so desire.

### DIFF
--- a/eng/common/pipelines/templates/steps/sparse-checkout.yml
+++ b/eng/common/pipelines/templates/steps/sparse-checkout.yml
@@ -22,84 +22,20 @@ steps:
     ${{ else }}:
       displayName: 'Sparse checkout repositories'
     inputs:
+      # Use inline instead of file because of the chicken/egg problem with loading a script
+      # when nothing has been checked out yet. Download script to save size on the generated
+      # template to stay below the devops max.
       targetType: inline
-      # Define this inline, because of the chicken/egg problem with loading a script when nothing
-      # has been checked out yet.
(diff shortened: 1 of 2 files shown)


eng/common/pipelines/templates/steps/sparse-checkout.yml
```diff
@@ -22,84 +22,20 @@ steps:
...
-              # This will use the default branch if repo.Commitish is empty
-              git -c advice.detachedHead=false checkout $commitish --
-            } else {
-              Write-Host "Skipping checkout as repo has already been initialized"
-            }
-
-            Pop-Location
-        }
-
+        $req = Invoke-WebRequest "https://raw.githubusercontent.com/$(Build.Repository.Name)/main/eng/common/scripts/sparse-checkout.ps1"
```
mikeharder, line 30: Should we add error handling for when this call fails (which it inevitably will)?  Or just let the exception bubble up?

mikeharder (approved):

weshaggard (changes requested):
I have some concerns with this and would like to talk offline about it before we take this change.

PR #7819, opened 2024-03-05
Update package version on automatic API revision

praveenkuttappan:
Update package version on API revision even if API surface matches to reflect correct package version info. This change along with tagging a revision after package release will ensure that package version is updated correctly(A separate change from updating package version).

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
+++ b/src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
@@ -46,9 +46,9 @@ namespace APIViewWeb.Controllers
                         runAnalysis: false, memoryStream: memoryStream);
 
                     var apiRevision = await CreateAutomaticRevisionAsync(codeFile: codeFile, label: label, originalName: file.FileName, memoryStream: memoryStream, compareAllRevisions);
-
                     if (apiRevision != null)
                     {
+                        apiRevision = await _apiRevisionsManager.UpdateRevisionPackageVersionAsync(apiRevision, codeFile.PackageVersion);
                         var reviewUrl = $"{this.Request.Scheme}://{this.Request.Host}/Assemblies/Review/{apiRevision.ReviewId}?revisionId={apiRevision.Id}";
                         return apiRevision.IsApproved ? Ok(reviewUrl) : StatusCode(statusCode: StatusCodes.Status201Created, reviewUrl);
                     }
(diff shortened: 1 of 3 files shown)


chidozieononiwu (approved):

src/dotnet/APIView/APIViewWeb/Managers/APIRevisionsManager.cs
```diff
@@ -828,5 +828,15 @@ private async Task GenerateAPIRevisionInExternalResource(ReviewListItemModel rev
             }
             return result;
         }
+
+        public async Task<APIRevisionListItemModel> UpdateRevisionPackageVersionAsync(APIRevisionListItemModel revision, string packageVersion)
+        {
+            if (packageVersion != null && !packageVersion.Equals(revision.Files[0].PackageVersion))
```
maririos, line 834: just making sure, does `revision.Files[0]` always exists? or would it be better to check and avoid a null pointer

praveenkuttappan: It should always exists for a revision. Otherwise a revision is does not have anything to show in APIView.

praveenkuttappan:
@chidozieononiwu @maririos I have pushed additional changes to update label and also to add a new package version param to send actual package version instead of daily build alpha version from CI. I have fixed another issue in copying approval status to automatic review.

src/dotnet/APIView/APIViewWeb/Managers/APIRevisionsManager.cs
maririos: One eye is twitching bc this is a function that anyone can call and I don't want people to get cunexpected behavior. But I understand that it doesn't make sense and it should be good

src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
```diff
@@ -35,7 +35,7 @@ public class AutoReviewController : Controller
         }
 
         [HttpPost]
-        public async Task<ActionResult> UploadAutoReview([FromForm] IFormFile file, string label, bool compareAllRevisions = false)
+        public async Task<ActionResult> UploadAutoReview([FromForm] IFormFile file, string label, bool compareAllRevisions = false, string packageVersion = null)
```
maririos, line 38: would it be helpful to document in which cases the function expects `packageVersion` not to be `null`?

maririos (approved):

src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
praveenkuttappan: I added it as null for now until all individual caller from language repo starts to send it. Once change is merged to individual repos, I will change this as required param.

PR #7825, opened 2024-03-06
Update opentelemetry collector config to read appinsights string from…

LarryOsterman:
… environment and add more logging.


richardpark-msft (approved):
Looks good, I also removed the line that we had that injected the connection string into the file.

LarryOsterman:
> Looks good, I also removed the line that we had that injected the connection string into the file.

Thanks - I was planning on doing that as well but you got there ahead of me.

richardpark-msft:
I wanted to be a part of something. Something big.

PR #7835, opened 2024-03-07
Update Prepare-Release.ps1 to handle only one previous release

weshaggard:
In the case there is exactly one previous release PS will return the single object and thus the Count property will not exist. Instead this change ensures that we always have a list.


benbp (approved):

benbp:
I wish there was a way to turn this off :D

PR #7843, opened 2024-03-08
1es compatible publish proxy logs

scbedd:
See title.


eng/common/testproxy/publish-proxy-logs.yml
```diff
@@ -1,16 +1,24 @@
 parameters:
   rootFolder: '$(Build.SourcesDirectory)'
+  OneES: false
```
weshaggard, line 3: Is there any global parameters or variables set in the 1es case we can use? If not adding a new one should be fine.

weshaggard (approved):

PR #7845, opened 2024-03-08
Create a separate job for events requiring Az CLI

JimSuplizio:
There were recent changes to the azure/login@v1 step that inject pre/post Az CLI commands. The pre Az CLI command, more often than not, costs more in terms of time than it does to install the event processor and process the action. The injected commands do not honor the conditionals around the step, meaning that every action would pay this price regardless not just the issues opened which is the only command that needs the Az CLI to get the Label Service key.

The solution was to create two jobs with conditions around each job. The first job will only run for issues opened, the second job will run for all the other events. The only difference between the two are the Az CLI usage, everything else is the same. This does mean that there are two jobs for each Action but one job is always skipped due to the conditions and this doesn't cause any inflation of the Action run time.

Here's an [example of an Action processing and issues opened event](https://github.com/azure-sdk/github-event-processor-test/actions/runs/8206020842). Notice that it has all of the Az CLI steps.

Here's an [example of an Action processing anything other than issues opened](https://github.com/azure-sdk/github-event-processor-test/actions/runs/8206037856). Notice there are no Az CLI steps.


jsquire (approved):

.github/workflows/event-processor.yml
```diff
@@ -26,17 +30,16 @@ jobs:
       contents: read
     name: Handle ${{ github.event_name }} ${{ github.event.action }} event
```
weshaggard, line 31:
```suggestion
    name: Handle ${{ github.event_name }} ${{ github.event.action }} event with azure login
```

I know we talked about this but I do think we should at least add something about what is different about this job.

weshaggard (approved):

PR #7854, opened 2024-03-11
Skip env processing on matrix import step

benbp:
Matrix filters are intended to work against the environment variable key, not the resolved value. There is a bug where resolve environment variables on an imported matrix, before processing any matrix filters passed in from the top level. This PR changes import behavior to skip env var lookup until after the imported matrix has been combined with the parent matrix.


scbedd (approved):
Yep, def makes sense. Thanks @benbp

weshaggard (approved):

PR #7855, opened 2024-03-11
Pipeline template to validate package and update package work item

praveenkuttappan:
This PR has changes to create a new pipeline template which will run version validation, change log validation, API review status check and package name approval status and create or update this information in package work item in Azure DevOps.

### DIFF
--- a/eng/common/scripts/ChangeLog-Operations.ps1
+++ b/eng/common/scripts/ChangeLog-Operations.ps1
@@ -148,6 +148,10 @@ function Confirm-ChangeLogEntry {
       Message = ""
     }
   }
+  else {
+    # Do not stop the script on error when status object is passed as param
+    $ErrorActionPreference = 'Continue'
+  }
   $changeLogEntries = Get-ChangeLogEntries -ChangeLogLocation $ChangeLogLocation
   $changeLogEntry = $changeLogEntries[$VersionString]
 
--- a/eng/common/scripts/ChangeLog-Operations.ps1
+++ b/eng/common/scripts/ChangeLog-Operations.ps1
@@ -138,14 +138,23 @@ function Confirm-ChangeLogEntry {
     [Parameter(Mandatory = $true)]
     [String]$VersionString,
     [boolean]$ForRelease = $false,
-    [Switch]$SantizeEntry
+    [Switch]$SantizeEntry,
+    [PSCustomObject]$ChangeLogStatus = $null
   )
 
+  if (!$ChangeLogStatus) {
+    $ChangeLogStatus = [PSCustomObject]@{
+      IsValid = $false
+      Message = ""
+    }
+  }
   $changeLogEntries = Get-ChangeLogEntries -ChangeLogLocation $ChangeLogLocation
   $changeLogEntry = $changeLogEntries[$VersionString]
 
   if (!$changeLogEntry) {
-    LogError "ChangeLog[${ChangeLogLocation}] does not have an entry for version ${VersionString}."
+    $ChangeLogStatus.Message = "ChangeLog[${ChangeLogLocation}] does not have an entry for version ${VersionString}."
+    $ChangeLogStatus.IsValid = $false
+    LogError "$($ChangeLogStatus.Message)"
     return $false
   }
 
@@ -161,14 +170,16 @@ function Confirm-ChangeLogEntry {
   Write-Host "-----"
 
   if ([System.String]::IsNullOrEmpty($changeLogEntry.ReleaseStatus)) {
-    LogError "Entry does not have a correct release status. Please ensure the status is set to a date '($CHANGELOG_DATE_FORMAT)' or '$CHANGELOG_UNRELEASED_STATUS' if not yet released. See https://aka.ms/azsdk/guideline/changelogs for more info."
+    $ChangeLogStatus.Message = "Entry does not have a release status. Please ensure the status is set to a date '($CHANGELOG_DATE_FORMAT)' or '$CHANGELOG_UNRELEASED_STATUS' if not yet released. See https://aka.ms/azsdk/guideline/changelogs for more info."
+    $ChangeLogStatus.IsValid = $false
+    LogError "$($ChangeLogStatus.Message)"
     return $false
   }
 
   if ($ForRelease -eq $True)
   {
     LogDebug "Verifying as a release build because ForRelease parameter is set to true"
-    return Confirm-ChangeLogForRelease -changeLogEntry $changeLogEntry -changeLogEntries $changeLogEntries
+    return Confirm-ChangeLogForRelease -changeLogEntry $changeLogEntry -changeLogEntries $changeLogEntries -ChangeLogStatus $ChangeLogStatus
   }
 
   # If the release status is a valid date then verify like its about to be released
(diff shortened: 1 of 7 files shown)


eng/common/scripts/Validate-Package.ps1
```diff
@@ -0,0 +1,312 @@
...
+    [string]$Message
+
+    ValidationStatus([string]$name, [string]$status)
+    {
+        $this.Name = $name
+        $this.Status = $status
+    }
+}
+
+class PackageDetails
```
weshaggard, line 62: What value do you get by defining a class over just using a custom object with properties? I've always found classes in PS to be trouble, especially during development given you have to restart your shell everytime you change one.

eng/common/scripts/Validate-Package.ps1
```diff
@@ -0,0 +1,312 @@
...
+    }    
+    return $validationStatus
+}
+
+function VerifyAPIReview($packageDetails, $packageName, $packageVersion, $language)
+{
+    $packageDetails.APIReviewValidation = [ValidationStatus]::new("API Review Approval", "Pending")
+    $packageDetails.PackageNameValidation = [ValidationStatus]::new("Package Name Approval", "Pending")
+
+    try
```
weshaggard, line 122: Should we be sharing this logic with other places?

eng/common/scripts/Validate-Package.ps1
```diff
@@ -0,0 +1,312 @@
...
+    PackageDetails([string]$name)
+    {
+        $this.Name = $name
+    }
+}
+
+function ValidateVersion($packageName, $versionString)
+{
+    $validationStatus = [ValidationStatus]::new("Version Validation", "Success")
+    $version = [AzureEngSemanticVersion]::ParseVersionString($versionString)
```
weshaggard, line 80: Why are we validating the version? If we do validate it, shouldn't we do something like ensure it hasn't been released or something along those lines?

eng/common/scripts/Validate-Package.ps1
praveenkuttappan: Current Prepare release script already uses this API. I will update it to pass package version param. Create API review gets the status of matching revision, so this API is not really required in that scenario to check the status.

eng/common/scripts/Validate-Package.ps1
praveenkuttappan: Current validation is just to ensure version is in expected format. We don't create any status update in package work item if it finds that version was already shipped.

eng/common/scripts/Validate-Package.ps1
praveenkuttappan: Changed to custom object.

eng/common/scripts/Validate-Package.ps1
praveenkuttappan: Removed version validation.

eng/common/scripts/Helpers/ApiView-Helpers.ps1
```diff
@@ -29,15 +29,17 @@ function Check-ApiReviewStatus($packageName, $packageVersion, $language, $url, $
...
-  $body = @{
-    language = $lang
-    packageName = $packageName
-    packageVersion = $packageVersion
-  }
 
   try
   {
-    $response = Invoke-WebRequest $url -Method 'GET' -Headers $headers -Body $body
+    $requestUrl = $url+"?language=" + $lang +"&packageName=" +$packageName
```
weshaggard, line 35:
```suggestion
    $requestUrl = "${url}?language=${lang}&packageName=${packageName}"
```

eng/common/scripts/Validate-All-Packages.ps1
```diff
@@ -0,0 +1,62 @@
...
+    $pkgPropPath = Join-Path -Path $ConfigFileDir "$PackageName.json"
+    if (-Not (Test-Path $pkgPropPath))
+    {
+        Write-Host " Package property file path $($pkgPropPath) is invalid."
+        return $false
+    }
+
+    $pkgInfo = Get-Content $pkgPropPath | ConvertFrom-Json
+    &$EngCommonScriptsDir/Validate-Package.ps1 `
+        -PackageName $PackageName `
```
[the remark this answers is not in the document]
weshaggard: Actually do you even need it here at all?

eng/common/scripts/Validate-Package.ps1
```diff
@@ -0,0 +1,291 @@
...
+        Message = ""
+    }
+
+    try
+    {
+        $changeLogFullPath = Join-Path $RepoRoot $changeLogPath
+        Write-Host "Path to change log: [$changeLogFullPath]"        
+        if (Test-Path $changeLogFullPath)
+        {
+            $errOutput = $( $validChangeLog = & Confirm-ChangeLogEntry -ChangeLogLocation $changeLogFullPath -VersionString $versionString -ForRelease $true ) 2>&1
```
weshaggard, line 64: This line needs a little work. Perhaps you need to make changes to Confirm-ChangeLogEntry

eng/common/scripts/Validate-Package.ps1
```diff
@@ -0,0 +1,291 @@
...
+    
+    if ($LASTEXITCODE -ne 0)
+    {
+        Write-Host "Update of the Devops Release WorkItem failed."
+        return $false
+    }
+    return $true
+}
+
+function UpdateValidationStatus($pkgvalidationDetails)
```
weshaggard, line 205: I would suggest moving this to the DevOps helpers

eng/common/scripts/Validate-Package.ps1
```diff
@@ -0,0 +1,291 @@
...
+Write-Host "Output: $($output)"
+
+# Create json token file in artifact path
+$tokenFile = Join-Path $ArtifactPath "$PackageName-Validation.json"
+$output | Out-File -FilePath $tokenFile -Encoding utf8
+
+# Create DevOps work item
+$updatedWi = CreateUpdatePackageWorkItem $pkgInfo
+
+# Update validation status in package work item
```
weshaggard, line 287: Do we need to return an exit code?

praveenkuttappan: Added a check to verify status of create/update work item and removed explicit exit code from this script.

eng/common/scripts/ChangeLog-Operations.ps1
```diff
@@ -138,14 +138,23 @@ function Confirm-ChangeLogEntry {
...
+    }
+  }
   $changeLogEntries = Get-ChangeLogEntries -ChangeLogLocation $ChangeLogLocation
   $changeLogEntry = $changeLogEntries[$VersionString]
 
   if (!$changeLogEntry) {
-    LogError "ChangeLog[${ChangeLogLocation}] does not have an entry for version ${VersionString}."
+    $ChangeLogStatus.Message = "ChangeLog[${ChangeLogLocation}] does not have an entry for version ${VersionString}."
+    $ChangeLogStatus.IsValid = $false
+    LogError "$($ChangeLogStatus.Message)"
```
weshaggard, line 157: Didn't you want to prevent the errors from failing fast in this scenario? I think LogError will still fail in a pipeline.

weshaggard (approved):

eng/common/scripts/ChangeLog-Operations.ps1
praveenkuttappan: I have updated the PR to set Error Action preference to continue when status object is passed from caller.

eng/common/scripts/ChangeLog-Operations.ps1
```diff
@@ -148,6 +148,10 @@ function Confirm-ChangeLogEntry {
       Message = ""
     }
   }
+  else {
+    # Do not stop the script on error when status object is passed as param
+    $ErrorActionPreference = 'Continue'
```
weshaggard, line 153: Is this needed? In general it is not a great idea to set these preference variables conditionally like this.

praveenkuttappan: I have removed it in most recent PR and suppressed error if caller has passed an object for status

Issue #7851, opened 2024-03-11
Release planner - Enable cloud flow to create package work item


praveenkuttappan:
Cloud flow to create package work item is disabled now due to empty package root path. Enable the cloud flow to create package work item once the issue is fixed to set repo path by identifying the repo path within the pipeline,

maririos:
@praveenkuttappan could you verify that the cloud flow is on after the deployment from last week?

praveenkuttappan:
Yes. It was turned back on. Closing this issue.

PR #7859, opened 2024-03-12
Detect package root path from language repo when creating work item

praveenkuttappan:
Package root path is not available when creating a package work item from release planner. Thich change is to automatically detect the package root path from language repo by the pipeline, so it won't be empty for any package work items created from release planner.

### DIFF
--- a/eng/scripts/Create-Package-WorkItem.ps1
+++ b/eng/scripts/Create-Package-WorkItem.ps1
@@ -55,8 +55,8 @@ try
 {
     . "eng/common/scripts/common.ps1"
     # Parse package properties from cloned repo and find the package repo path
-    $packageProperties = Get-PkgProperties -PackageName $PackageName
-    if (!$packageProperties)
+    $PkgProperties = Get-PkgProperties -PackageName $PackageName
+    if (!$PkgProperties)
     {
         Write-Error "Failed to find package properties for package name [$PackageName]."
         exit 1
--- a/eng/scripts/Create-Package-WorkItem.ps1
+++ b/eng/scripts/Create-Package-WorkItem.ps1
@@ -9,8 +9,6 @@ param (
   [Parameter(Mandatory = $true)]
   [string]$PackageDisplayName,
   [Parameter(Mandatory = $true)]
-  [string]$PackageType,
-  [Parameter(Mandatory = $true)]
   [string]$PackageVersion,
   [Parameter(Mandatory = $true)]
   [string]$ReleaseDate,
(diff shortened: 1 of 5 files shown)


praveenkuttappan:
Fixes https://github.com/Azure/azure-sdk-tools/issues/7851

eng/scripts/Create-Package-WorkItem.ps1
```diff
@@ -0,0 +1,98 @@
...
+        "C" { return "azure-sdk-for-c" }
+        "C++" { return "azure-sdk-for-cpp" }
+    }
+    Write-Host "Unknown language to map it to repo name. Language: [$language]"
+    return ""
+}
+
+function Clone-Repo($repoName, $targetDir)
+{
+    git clone "https://github.com/azure/$repoName.git"  $targetDir
```
weshaggard, line 44: Might not be necessary but we should consider a shallow clone to speed things up.

eng/scripts/Create-Package-WorkItem.ps1
```diff
@@ -0,0 +1,98 @@
...
+        exit 1
+    }
+    $clonedRepoPath = Join-Path $WorkingDir $repoName
+    Clone-Repo $repoName $clonedRepoPath
+    
+    # load common script helpers from cloned language repo
+    . (Join-Path $clonedRepoPath "eng/common/scripts/common.ps1")
+    Write-Host "Language repo root path: [$RepoRoot]"
+    # Parse package properties from cloned repo and find the package repo path
+    $packageProperties = Get-PkgProperties -PackageName $PackageName -ServiceDirectory ""
```
weshaggard, line 69:
```suggestion
    $packageProperties = Get-PkgProperties -PackageName $PackageName
```

eng/scripts/Create-Package-WorkItem.ps1
```diff
@@ -0,0 +1,98 @@
...
+
+Write-Host "Package Root Path: [$PackageRootPath]"
+# Create or update package work item  
+&$EngCommonScriptsDir/Update-DevOps-Release-WorkItem.ps1 `
+    -language $Language `
+    -packageName $PackageName `
+    -version $PackageVersion `
+    -plannedDate $ReleaseDate `
+    -packageRepoPath $PackageRootPath `
+    -packageType $PackageType `
```
weshaggard, line 93: What about `-packageNewLibrary`? https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Prepare-Release.ps1#L149

Should we use more of the package metadata we get from Get-PkgProperties? I also wonder if we should potentially validate those values with what you passing to pipeline.

weshaggard:
What is your goal with these changes? Is this only for temporary purposes as we talked about deprecating this pipeline.

eng/scripts/Create-Package-WorkItem.ps1
praveenkuttappan: We don't know the service directory to sparse checkout only the package path. Are you suggesting to skip any test directories? I can update the depth to make it as shallow clone.

praveenkuttappan:
> What is your goal with these changes? Is this only for temporary purposes as we talked about deprecating this pipeline.

Correct. It's only temporary work around until we remove the need to create package work item from release planner.

eng/scripts/Create-Package-WorkItem.ps1
weshaggard: I was just meaning to we could do a sparse clone which doesn't include any history which reduces the cost of cloning. Either way it is just an optimization.

weshaggard (approved):

PR #7868, opened 2024-03-13
Use subset of Go module identity for package name

jhendrixMSFT:
This will prevent collisions as the subset is guaranteed to be unique. Updated dependencies.


jhendrixMSFT:
The build for APIViewWeb is broken, can somebody please take a look?

maririos:
@chidozieononiwu could  please take a look?

chlowell (approved):

RickWinter (approved):

src/go/cmd/module.go
```diff
@@ -39,6 +39,23 @@ type Module struct {
...
 }
 
+var majorVerSuffix = regexp.MustCompile(`/v\d+$`)
+
+// fetches the value for Module.PackageName from the full module path
+func getPackageNameFromModPath(modPath string) string {
+	// for official SDKs, use a subset of the full module path
+	// e.g. github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/appcomplianceautomation/armappcomplianceautomation
+	// becomes resourcemanager/appcomplianceautomation/armappcomplianceautomation
+	if suffix, ok := strings.CutPrefix(modPath, "github.com/Azure/azure-sdk-for-go/sdk/"); ok {
```
praveenkuttappan, line 49: I just noticed that package work items and release data tracker csv uses package name with `sdk/` as prefix. So might have to have this prefix here as well. I wasn't aware that package name is created using this format so sorry for my earlier suggestion to remove the prefix. We need to keep package name same across all the tools..
https://github.com/Azure/azure-sdk/blob/main/_data/releases/latest/go-packages.csv

for e.g. `sdk/data/azappconfig`

@benbp FYI..

praveenkuttappan (changes requested):
Package name must include `sdk/` as prefix to match the package name expected in all other tooling for Go.

src/go/cmd/module.go
jhendrixMSFT: Done.

PR #7873, opened 2024-03-13
Move git check for changes to script

benbp:
Cutting off some size from an inline script that gets duplicated dozens of times in repos like Java.


weshaggard (approved):

benbp:
Closing in favor of https://github.com/Azure/azure-sdk-tools/pull/7892

PR #7880, opened 2024-03-15
Add an additional timeout handler for mac-errors

### LINKED ISSUE #7784: Test-Proxy git clone not retrying properly
The JS Folks (@jeremymeng @HarshaNalluru ) have been tracking a weird failure to start playback on one of their unit tests.

[Link to one such failure](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3493350&view=logs&j=5ac9d23f-2893-5951-84af-95de25a7d6e2&t=3402a3b1-1688-5075-ca6d-8be6d6d8fabd&l=4339)

At first, there was nothing in the logs that highlighted the issue for us, so we activated `debug` mode on the `js - core` job to enable higher log output.

@ckairen dug into this a bit, and [located this log message](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3493350&view=logs&j=5ac9d23f-2893-5951-84af-95de25a7d6e2&t=3402a3b1-1688-5075-ca6d-8be6d6d8fabd&l=4297) in a recent failure.

This indicates to me that the [retry on transient git failure](https://github.com/Azure/azure-sdk-tools/blob/b7b097c4f5cb10c573a7301348ccd0c2fd62d288/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitProcessHandler.cs#L220) is not working properly. We need to fix this.

Fix might be a simple as adding this specific error message to retry handler:

```
      fatal: unable to access 'https://github.com/Azure/azure-sdk-assets/': Recv failure: Operation timed out
```

scbedd:
Attempts to resolve #7784

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitProcessHandler.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitProcessHandler.cs
@@ -250,6 +250,12 @@ namespace Azure.Sdk.Tools.TestProxy.Store
                 {
                     return true;
                 }
+
+                // fatal: unable to access 'https://github.com/Azure/azure-sdk-assets/': Recv failure: Operation timed out
+                if (result.StdErr.Contains("Recv failure: Operation timed out"))
+                {
+                    return true;
+                }
             }
 
             return false;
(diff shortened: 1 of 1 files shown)


mikeharder (approved):
LGTM, but I also wonder if we could just retry on any error with "timed out"?

scbedd:
> LGTM, but I also wonder if we could just retry on any error with "timed out"?

I thought about this. Why not. I'll make the change.

PR #7882, opened 2024-03-15
Initial check in for Azure SDK AI Bots

raych1:
The code includes:
- QA bot
- embedding tool
- scripts


tools/sdk-ai-bots/AzureSdkQaBot/LocalPkg/NuGet_Package_PlaceHolder
```diff
@@ -0,0 +1 @@
+Put the Microsoft.TeamsAI.<version>.nupkg here to be referenced by sample project.
```
lirenhe, line 1: Do we need to merge those nuget packages above into the repo?

raych1: I removed the unused packages. For the only one remained, we need to check in it as it might hit issues to fetch it from github feed. I will update the code to use the GA library later on when we can get rid of this package from the repo.

lirenhe (approved):

tools/sdk-ai-bots/.gitignore
```diff
@@ -0,0 +1,122 @@
+# Logs
```
weshaggard, line 1: Do we really need a specialized ignore file? Usually we just depend on the one in the root.

tools/sdk-ai-bots/AzureSdkQaBot/LocalPkg/NuGet_Package_PlaceHolder
weshaggard: Is the package not available on a feed?

weshaggard:
@konrad-jamrozik @maririos FYI Here is the start of the Chat bot.

@raych1 I assume you plan to also create a pipeline and put it into this directory as well?

maririos:
> @konrad-jamrozik @maririos FYI Here is the start of the Chat bot.

Is this the one in the Teams Channel that we tried a couple of months ago?

weshaggard:
Ye

> > @konrad-jamrozik @maririos FYI Here is the start of the Chat bot.
> 
> Is this the one in the Teams Channel that we tried a couple of months ago?

Yes that is the one. Ray is working to get it into a pipeline so we can update it regularly and others can also make updates as needed.

maririos:
FYI @MrJustinB and @mario-guerra

MrJustinB:
@mario-guerra on the surface the bot you're working on (i.e. scraping the Teams Onboarding channel chat) and this one has interesting synergies. I don't think we should work to publish two separate bots. Would it be difficult to contribute your work to this bot?

mario-guerra:
@MrJustinB I would definitely prefer to lever this work than create a whole other bot. I am guessing that we should be able to integrate the work I'm doing with this bot, I'll have to sync with @raych1 on how we could combine our efforts.

PR #7900, opened 2024-03-16
Stress: build our own custom otel collector

richardpark-msft:
Build our own custom otel distribution using their official `otb` tool.

It reduces the amount of code significantly by cutting it down to azure monitor and removing all other contrib components. As a bonus we can also control the Docker images used, which is helpful for other things.


richardpark-msft:
@lmolkova, some security related things has made me more motivated on controlling the images we use and also strip out code we don't need. We've talked about this before, hoping you can see if I did it the right way or if there are potentially even easier ways to do this.

benbp (approved):

lmolkova (approved):
LGTM! :shipit:

PR #7924, opened 2024-03-20
Migrate more tools pipelines to 1es pipeline templates

benbp:
Additional update: only auth to ACR on internal builds (not sure if this is necessary but just in case)

Pipelines:
tools - apiview-stub-generator
tools - apiview-typespec-emitter
tools - azure-pylint-guidelines-checker
tools - js - apiview-parser
tools - swaggerapiparser
tools - tsp-client


scbedd (approved):
This looks great.

PR #7951, opened 2024-03-26
Call GetDocsMsDevLanguageSpecificPackageInfoFn if it's defined

JimSuplizio:
The GetDocsMsDevLanguageSpecificPackageInfoFn pointer was only being called if the DevVersion was set and should have been being called all the time, if the function pointer was defined. The reason for this is because the only language that had the function defined was Javascript and it only needed to be called if the DevVersion was set. As such, the fix to this file is to call it all the time and, for Javascript, the Get-javascript-DocsMsDevLanguageSpecificPackageInfo, needs to be updated to only process if the DevVersion is set.


danieljurek (approved):
Before this change merges, a JS change needs to be made. The two can merge at roughly the same time as long as the PRs merge before docindex runs for the day.

The `$GetDocsMsDevLanguageSpecificPackageInfoFn` function is now more universally applicable here. This is how it should've been to begin with.

PR #7955, opened 2024-03-27
Support for client core libraries

JonathanGiles:


src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/util/MiscUtils.java
```diff
@@ -87,6 +87,19 @@ private static String wrapLine(final String line, final int lineLength) {
...
     }
 
+    public static String upperCase(String s) {
+        return upperCase(s, 0);
+    }
+
+    /**
+     * Makes all characters in the string uppercase, other than the given index.
+     */
+    public static String upperCase(String s, int index) {
```
alzimmermsft, line 97: Isn't this doing the exact opposite of what the Javadoc say?

alzimmermsft (approved):

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/util/MiscUtils.java
JonathanGiles: Ha, yes, typo, will fix. Generally I hate these two methods, it hurts. I should make it better.

alzimmermsft:
```bash
rm -rf /*
```

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/Diagnostics.java
```diff
@@ -1,90 +1,136 @@
...
+                addAzureCoreDiagnostics();
+                break;
+            }
+            case GENERIC: {
+                System.out.println("    Applying io.clientcore specific diagnostics...");
+                addClientCoreDiagnostics();
+                break;
+            }
+            case UNKNOWN:
+            default: {
```
srnagar, line 49: For default, should we fallback to using Azure diagnostics to work with older versions of Azure libraries that have already shipped?

srnagar (approved):

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/Diagnostics.java
JonathanGiles: Sometimes I'm bothered by e.g. Semantic Kernel falling into our Azure SDK diagnostics. I think I would rather we keep it as-is, and make sure all existing libraries are picked up as Azure libraries.

chidozieononiwu (approved):

PR #7975, opened 2024-03-28
Pass Group in request to get package status

praveenkuttappan:
Java packages contains `group name:` as a prefix in APIView.  API status request should pass full package name including group name when it's present for a language to get correct status.

### DIFF
--- a/eng/common/scripts/Validate-Package.ps1
+++ b/eng/common/scripts/Validate-Package.ps1
@@ -212,7 +212,11 @@ $changeLogStatus = [PSCustomObject]@{
 ValidateChangeLog $changeLogPath $versionString $changeLogStatus
 
 # API review and package name validation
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Validate-Package.ps1
```diff
@@ -212,10 +212,14 @@ $changeLogStatus = [PSCustomObject]@{
 ValidateChangeLog $changeLogPath $versionString $changeLogStatus
 
 # API review and package name validation
-$apireviewDetails = VerifyAPIReview $PackageName $pkgInfo.Version $Language
+$fulPackageName = $pkgName
+if ($pkgInfo.Group){
```
weshaggard, line 216: This will fail in strict mode. You need to check if this property exists. Something similar to https://github.com/Azure/azure-sdk/blob/main/eng/scripts/PackageList-Helpers.ps1#L259

weshaggard (approved):

PR #7986, opened 2024-04-02
A pshell update (7.4?) caused `git show` to stop working

scbedd:
- `git show head:sdk\storage\Azure.Storage.Blob\assets.json` no longer results in visible output.
  - The script grabs the tag from the **left** side of the merge, holds it constant, and merges the incoming branch tag _into_ the assets. If we aren't able to properly utilize `git show`, we can't generate the `left` or `right` side of the assets.json, and end up setting the `assets.json` to **empty**.
- `git show head:sdk/storage/Azure.Storage.Blob/assets.json` does still work, so amending the command


JonathanCrd (approved):
LGTM

PR #8019, opened 2024-04-03
Add retry on error for sparse checkout

benbp:


eng/common/pipelines/templates/steps/sparse-checkout.yml
```diff
@@ -48,10 +65,10 @@ steps:
 
               if ($repository.Commitish -match '^refs/pull/\d+/merge$') {
                 Write-Host "git clone --no-checkout --filter=tree:0 -c remote.origin.fetch='+$($repository.Commitish):refs/remotes/origin/$($repository.Commitish)' https://github.com/$($repository.Name) ."
```
weshaggard, line 67: Should we move the command logging into this retry function now as well? Just a way to save some lines of our inline script and also reduce some duplication of the command.

danieljurek (approved):

eng/common/pipelines/templates/steps/sparse-checkout.yml
benbp: Yeah I could move all logging into the function, and make it a generic runner with a retries int, that way we can pass all commands through it.

benbp: @weshaggard ok made this update here. Overdue anyway.

weshaggard (approved):

eng/common/pipelines/templates/steps/sparse-checkout.yml
```diff
@@ -47,23 +74,18 @@ steps:
...
 
-              Write-Host "git sparse-checkout init"
-              git sparse-checkout init
+              Run git sparse-checkout init
 
               # Set non-cone mode otherwise path filters will not work in git >= 2.37.0
               # See https://github.blog/2022-06-27-highlights-from-git-2-37/#tidbits
-              Write-Host "git sparse-checkout set --no-cone '/*' '!/*/' '/eng'"
               git sparse-checkout set --no-cone '/*' '!/*/' '/eng'
             }
```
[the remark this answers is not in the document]
benbp: I had a lot of issues with commands that contained any sort of glob/wildcard when sourcing those values dynamically. We may be able to improve upon this by using invoke expression everywhere, I will have to test it out.

PR #8020, opened 2024-04-04
Merge `eng/` from `main` on assets push

scbedd:
@weshaggard  I'm extremely pleased with this little diff methodology. Let me know what you think.

- [x] Handle when the patch is empty (call out to git apply --check)
- [x] PR another change that fixes the tools build. 
- [x] Resolve integration issue on pure windows `git` discovered in integration tests. Why is a soft reset failing?
  - This was failing due to a case that should only really ever occur in our test cases, where we _force_ a failed auth by setting an invalid GIT_TOKEN. This meant that we failed to retrieve remote `main` for the patch. The weird error we were seeing on CLI integration for windows, is that a soft reset with _uncommitted changes_ resulted in a weird error state that git couldn't understand how to resolve automatically. (`unable to retrieve promisor remote`) By placing the soft reset behind the changes actually being committed, we totally sidestep the entire problem.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
@@ -149,7 +149,7 @@ namespace Azure.Sdk.Tools.TestProxy.Store
                         File.Delete(engPatchLocation);
                     }
 
-                    GitHandler.Run($"diff --output=changes.patch --no-color --binary --no-prefix HEAD main -- .gitignore", config);
+                    GitHandler.Run($"diff --output=changes.patch --no-color --binary HEAD main -- .gitignore", config);
                     if (GitHandler.TryRun($"apply --check changes.patch", config.AssetsRepoLocation.ToString(), out var applyResult))
                     {
                         GitHandler.Run($"apply changes.patch", config);
@@ -582,11 +582,11 @@ namespace Azure.Sdk.Tools.TestProxy.Store
 
             if (combinedPath.ToLower() == AssetsJsonFileName)
             {
-                return "./ eng/";
+                return "./ eng/ .gitignore";
             }
             else
             {
-                return combinedPath.Substring(0, combinedPath.Length - (AssetsJsonFileName.Length + 1)) + " eng/";
+                return combinedPath.Substring(0, combinedPath.Length - (AssetsJsonFileName.Length + 1)) + " eng/ .gitignore";
             }
         }
         #endregion
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
@@ -139,15 +139,24 @@ namespace Azure.Sdk.Tools.TestProxy.Store
                      * push action will properly put their repo into the expected "ready to push" state that a failed
                      * merge would NOT.
                      */
+                    var engPatchLocation = Path.Combine(config.AssetsRepoLocation, "changes.patch");
                     GitHandler.Run($"diff --output=changes.patch --no-color --binary --no-prefix HEAD main -- eng/", config);
-                    if (GitHandler.TryRun($"apply --check --directory=eng/ changes.patch", config.AssetsRepoLocation.ToString(), out var applyResult))
+                    if (GitHandler.TryRun($"apply --check --directory=eng/ changes.patch", config.AssetsRepoLocation.ToString(), out var engPatchResult))
                     {
                         GitHandler.Run($"apply --directory=eng/ changes.patch", config);
                     }
-                    var pathLocation = Path.Combine(config.AssetsRepoLocation, "changes.patch");
+                    if (File.Exists(engPatchLocation)) {
+                        File.Delete(engPatchLocation);
+                    }
 
-                    if (File.Exists(pathLocation)) {
-                        File.Delete(pathLocation);
+                    GitHandler.Run($"diff --output=changes.patch --no-color --binary --no-prefix HEAD main -- .gitignore", config);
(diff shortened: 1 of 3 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
```diff
@@ -124,11 +124,37 @@ public async Task<NormalizedString> GetPath(string pathToAssetsJson)
...
+                     * This code works by generating a patch file for SPECIFICALLY the eng folder from main.
+                     * Given that these changes appear as "new" changes, they just look like normal file additions. 
+                     * This totally eliminates the possibility of weird historical merge if main has code that we don't expect. 
+                     * Under azure-sdk-assets, we should never see this, but we have already seen it with specific integration
+                     * test tags under azure-sdk-assets-integration. By keeping it as "patch", the soft RESET on unsuccessful 
+                     * push action will properly put their repo into the expected "ready to push" state that a failed
+                     * merge would NOT.
+                     */
+                    GitHandler.Run($"diff --output=changes.patch --no-color --binary --no-prefix HEAD main -- eng/", config);
+                    if (GitHandler.TryRun($"apply --check --directory=eng/ changes.patch", config.AssetsRepoLocation.ToString(), out var applyResult))
```
weshaggard, line 143: Should we include everything outside of the sdk folder? Or at least the files in the root like the .gitignore?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
```diff
@@ -124,11 +124,37 @@ public async Task<NormalizedString> GetPath(string pathToAssetsJson)
...
                     string gitUserName = GetGitOwnerName(config);
                     string gitUserEmail = GetGitOwnerEmail(config);
+                    string assetMessage = "Automatic asset update from test-proxy.";
+                    string configurationString = $"-c user.name=\"{gitUserName}\" -c user.email=\"{gitUserEmail}\"";
+
                     GitHandler.Run($"branch {branchGuid}", config);
                     GitHandler.Run($"checkout {branchGuid}", config);
+
+                    /*
+                     * This code works by generating a patch file for SPECIFICALLY the eng folder from main.
```
weshaggard, line 134: While I'm not against the patch approach I'm curious what you are worried about with the merge approach? If you are worried about conflicts that is still a possibility with patches as well.

scbedd: > While I'm not against the patch approach I'm curious what you are worried about with the merge approach? If you are worried about conflicts that is still a possibility with patches as well.

Merge approach immediately makes the `push failed revert to a pushable state` more complex. This is due to the fact I'll need to handle whether or not a merge actually happened from main in the `revert to ready to push` phase of exception handling. By treating this as a patch, it appears as actual changes to the files, and as such gets swept up in the `git add -A .`.

scbedd: And, in our integration tests, some of our tags have code in main under the same folder that they have recordings. So when we merge `main` we actually override the changes to a couple files in our recordings directory. While this _shouldn't_ happen in `main` from azure-sdk-assets, by keeping the patch to `eng/` we're absolutely certain.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
scbedd: I'll figure out how to make `.gitignore` addition work. See where you're coming from.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
```diff
@@ -124,11 +124,37 @@ public async Task<NormalizedString> GetPath(string pathToAssetsJson)
...
+                     * Given that these changes appear as "new" changes, they just look like normal file additions. 
+                     * This totally eliminates the possibility of weird historical merge if main has code that we don't expect. 
+                     * Under azure-sdk-assets, we should never see this, but we have already seen it with specific integration
+                     * test tags under azure-sdk-assets-integration. By keeping it as "patch", the soft RESET on unsuccessful 
+                     * push action will properly put their repo into the expected "ready to push" state that a failed
+                     * merge would NOT.
+                     */
+                    GitHandler.Run($"diff --output=changes.patch --no-color --binary --no-prefix HEAD main -- eng/", config);
```
weshaggard, line 142: Given main should never have anything under the sdk path maybe just always produce a patch for everything. I'm guessing there is also a way to exclude or revert certain paths if you want to be certain nothing comes under there.

weshaggard: Actually I guess that won't work with the sparce checkout though.

weshaggard (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
scbedd: Yeah. I'm torn. I honestly like the comfort of

> pull eng/ and gitignore and nothing else. It should just work given that main is definitely newer than any of the tags.

We already have a couple test cases that are exercising when there _are no_ changes from `main`, and those work just fine.

I feel like just diffing main will definitely _work_, but then we're back to the 

> but what if

To be clear that's just FUD on my part.

The real benefit that I discovered today is really that the patches from main come in as _part of the changeset_ so there is just a single commit with the users name, versus each tag having an "additional" tag on top of it that is merging main. Meaning that the tag won't properly display the recording changes unless the merge did nothing. Otherwise the tag commit is just a commit with the changes to `eng`, and you'll need to `browse files` to see just the recording changes.

I may swap to just merge everything? I'm trying a couple experiments locally to double check all the functionality regardless.

PR #8030, opened 2024-04-06
Add warning state to rotation

hallipr:
This change allows for a secret to be in a warning state (should have been rotated previously and is close to expiration) and replaces the state booleans `Expired`, `ThresholdExpired`, and now `WarningExpired` with a single state enum:
```
    Error,
    UpToDate,
    Rotate,
    Warning,
    Expired,
```

If any secret enters the WarningThreshold, the cli will categorize it as Expiring and the status command will return a non-zero exit code.

The WarningThreshold for plans should be configured so there is reasonable time past the RotationThreshold to allow for rotation and prevent unnecessary warnings.

### DIFF
--- a/tools/secret-management/Azure.Sdk.Tools.SecretRotation.Core/RotationPlan.cs
+++ b/tools/secret-management/Azure.Sdk.Tools.SecretRotation.Core/RotationPlan.cs
@@ -16,6 +16,7 @@ public class RotationPlan
         SecretStore primaryStore,
         IList<SecretStore> secondaryStores,
         TimeSpan rotationThreshold,
+        TimeSpan? warningThreshold,
         TimeSpan rotationPeriod,
         TimeSpan? revokeAfterPeriod)
     {
@@ -25,6 +26,7 @@ public class RotationPlan
         OriginStore = originStore;
         PrimaryStore = primaryStore;
         RotationThreshold = rotationThreshold;
+        WarningThreshold = warningThreshold;
         RotationPeriod = rotationPeriod;
         RevokeAfterPeriod = revokeAfterPeriod;
         SecondaryStores = new ReadOnlyCollection<SecretStore>(secondaryStores);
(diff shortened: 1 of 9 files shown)


tools/secret-management/Azure.Sdk.Tools.SecretRotation.Core/RotationPlan.cs
```diff
@@ -102,21 +106,31 @@ public async Task<RotationPlanStatus> GetStatusAsync()
 
             SecretState[] allStates = secondaryStoreStates.Prepend(primaryStoreState).ToArray();
 
-            DateTimeOffset thresholdDate = this.timeProvider.GetCurrentDateTimeOffset().Add(RotationThreshold);
+            DateTimeOffset rotationThresholdDate = this.timeProvider.GetCurrentDateTimeOffset().Add(RotationThreshold);
+
+            DateTimeOffset? warningThresholdDate = WarningThreshold.HasValue
+                ? this.timeProvider.GetCurrentDateTimeOffset().Add(WarningThreshold.Value)
+                : default;
```
weshaggard, line 113: should we have a smart default? Perhaps half the time of the expire?

weshaggard (approved):

tools/secret-management/Azure.Sdk.Tools.SecretRotation.Core/RotationPlan.cs
hallipr: I think that will be `rotationThreshold / 2`.   for the config:
```json
{
  "rotationPeriod": ".180.00:00:00",
  "rotationThreshold": "30.00:00:00"
}
```

we'd get the implicit:
```json
 "warning": "15.00:00:00"
```

meaning, we create a 180 day secret and rotate it 30 days before expiration.  If we haven't rotated it 15 days before expiration, we start reporting a warning

PR #8033, opened 2024-04-08
Use azure-json in ApiView Generator

alzimmermsft:
- Removes usage of Jackson, replaces it with `azure-json`.
- Replaces outdated `commons-lang` and replaces it with an HTML escaping specific library.
- Small API clean ups to improve performance.


src/java/apiview-java-processor/pom.xml
```diff
@@ -50,14 +50,14 @@
...
       <scope>test</scope>
     </dependency>
-      <dependency>
-          <groupId>commons-lang</groupId>
-          <artifactId>commons-lang</artifactId>
-          <version>2.6</version>
-      </dependency>
+    <dependency>
+      <groupId>org.unbescape</groupId>
+      <artifactId>unbescape</artifactId>
```
srnagar, line 58: Should we use [commons-lang3](https://mvnrepository.com/artifact/org.apache.commons/commons-lang3/3.14.0) which seems to be well-maintained compared to [unbescape](https://github.com/unbescape/unbescape) which hasn't been updated for many years?

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/util/ASTUtils.java
```diff
@@ -209,7 +207,34 @@ public static String makeId(EnumConstantDeclaration enumDeclaration) {
...
+        for (int i = 0; i < length; i++) {
+            char c = fullPath.charAt(i);
+            if (c == '"' || c == ' ') {
+                if (sb == null) {
+                    sb = new StringBuilder(length);
+                }
+
+                if (prevStart != i) {
+                    sb.append(fullPath, prevStart, i);
+                }
```
srnagar, line 227: Should we also append `-` after this to keep the ids backward compatible?

alzimmermsft: Good catch, missed that (more likely I had it then reverted a change and forgot it again)

src/java/apiview-java-processor/pom.xml
alzimmermsft: Went with this as it's a smaller library and escaping HTML really shouldn't need to change often, if at all.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -1515,11 +1527,10 @@ private void unindent() {
...
     private Token makeWhitespace() {
-        StringBuilder sb = new StringBuilder();
-        for (int i = 0; i < indent; i++) {
-            sb.append(" ");
-        }
-        return new Token(WHITESPACE, sb.toString());
+        byte[] bytes = new byte[indent];
+        Arrays.fill(bytes, (byte) ' ');
+
+        return new Token(WHITESPACE, new String(bytes, StandardCharsets.UTF_8));
```
JonathanGiles, line 1533: why? feels like readability goes down with this.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/util/ASTUtils.java
```diff
@@ -209,7 +214,36 @@ public static String makeId(EnumConstantDeclaration enumDeclaration) {
     }
 
     public static String makeId(String fullPath) {
-        return MAKE_ID.matcher(fullPath).replaceAll("-");
+        if (fullPath == null || fullPath.isEmpty()) {
```
JonathanGiles, line 217: what motivated this change?

alzimmermsft: Hot path with a regex that can be replaced with a simple character search

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/Main.java
```diff
@@ -205,6 +183,43 @@ private static void processJavaSourcesJar(File inputFile, APIListing apiListing)
...
+     * object.
+     *
+     * @param fs the {@link FileSystem} representing the jar file
+     * @param apiListing the {@link APIListing} object to set the {@link ApiViewProperties} on
+     * @param artifactId the artifact ID of the jar file
+     */
+    private static void tryParseApiViewProperties(FileSystem fs, APIListing apiListing, String artifactId) {
+        // the filename is [<artifactid>_]apiview_properties.json
+        String filename = (artifactId != null && !artifactId.isEmpty()
+            ? (artifactId + "_") : "") + "apiview_properties.json";
```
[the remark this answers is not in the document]
alzimmermsft: This was copied from main, just moved to a different spot 😄

JonathanGiles (approved):

PR #8037, opened 2024-04-09
Deal with friends from multiple classes

LarryOsterman:


tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -1318,6 +1318,45 @@ static std::string GetNavigationId(clang::FunctionDecl const* func)
   return navigationId;
 }
 
+static std::string GetNavigationId(
+    clang::FriendDecl const* friendDecl,
+    std::string const& parentNodeNavigationId)
+{
+  std::string navigationId = parentNodeNavigationId + "_friend_";
```
RickWinter, line 1325: Can the parentNodeNavigationId be ""? Can we assert these premises.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -1318,6 +1318,45 @@ static std::string GetNavigationId(clang::FunctionDecl const* func)
   return navigationId;
 }
 
+static std::string GetNavigationId(
```
RickWinter, line 1321: Add function comment

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
```diff
@@ -1318,6 +1318,45 @@ static std::string GetNavigationId(clang::FunctionDecl const* func)
...
+    {
+      navigationId
+          += QualType::getAsString(friendDecl->getFriendType()->getType().split(), LangOptions{});
+    }
+    else
+    {
+      friendDecl->dump(llvm::outs());
+    }
+  }
+  return navigationId;
```
RickWinter, line 1357: If you end up in the final else of the initial if/else then you could end up with navigationId being `<parentNodeId>_friend_`. Is a navigationId ending in "_" a valid format?

RickWinter (approved):

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
LarryOsterman: It cannot be empty, I'll add an error message if it is.

tools/apiview/parsers/cpp-api-parser/ApiViewProcessor/AstNode.cpp
LarryOsterman: All strings are valid, the only restriction is that the names must be unique.

PR #8038, opened 2024-04-09
[TestProxy] Common Sanitizers

scbedd:
This PR moves the test-proxy from a static append-only set of sanitizers, to an actual data structure with proper addition/removal of applied sanitizers.

- [x] Each recording or playback session now gets a modifiable list of sanitizers that are applied.
  - [x] Session sanitizers are now cut at playback or record start. Any further session sanitizer addition will not directly affect the recording that has been started.
- [x] As a consequence of all these additional sanitizers, **stored recordings** will now be sanitized. The way this will work is
  - Client starts playback session, gets a recordingId back
  - Client uses recordingId to add a couple additional sanitizers
  - Client starts firing actual requests from the test
  - The proxy will sanitize the loaded recording according to which sanitizers are active when the **first real request for the playback session occurs**.
- [x] Upon calling `/Admin/AddSanitizers` or `Admin/AddSanitizer`, an object or list of objects representing the registered sanitizers is now returned to the calling client. These objects contain the base sanitizer properties + the `id` of the registered sanitizer.
- [x] Added route `/Admin/RemoveSanitizers` which can apply to either the session-level or an individual playback/record session.

Observers please add suggestions either in discussion or directly onto the branch.


tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -985,11 +987,108 @@ public void SetDefaultExtensions(string recordingId = null)
...
+                    // "token": "sv=...ss=...srt=...se=...sp=...sig=" request body
+                    new BodyKeySanitizer("$..WEBSITE_AUTH_ENCRYPTION_KEY"),
+                    new BodyKeySanitizer("$..decryptionKey"),
+                    // "ServiceBusDlqSupplementaryAuthorization": "SharedAccessSignature sr=...\sig=...se=….skn=….",
+                    // "ServiceBusSupplementaryAuthorization": "SharedAccessSignature sr=...sig=...se=...skn=...",
+                    // "RequestBody": "client_id=...grant_type=...client_info=...client_secret=…scope=...",
                     new BodyKeySanitizer("$..access_token"),
-                    new BodyKeySanitizer("$..refresh_token")
+                    new BodyKeySanitizer("$..AccessToken"),
+                    //(client_id=)[^&]+ body regex
```
chlowell, line 1034: client IDs aren't secret

chlowell:
Does the proxy already have a way to deduplicate sanitizers so we don't run all of these twice? If not, I imagine it could compare two sanitizers with hashes of their string parameters

scbedd:
> Does the proxy already have a way to deduplicate sanitizers so we don't run all of these twice? If not, I imagine it could compare two sanitizers with hashes of their string parameters

It does not at this time.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -985,11 +998,177 @@ public void SetDefaultExtensions(string recordingId = null)
...
+                // - create an option in sanitizer to get access to 'ignore case' and other regex customizations
+                // - turn commented lines into actual sanitizers after discussion
+                // - add a query param sanitizer
                 Sanitizers = new List<RecordedTestSanitizer>
                 {
+                    // basic RecordedTestSanitizer handles Authorization header
                     new RecordedTestSanitizer(),
+
+                    // new defaults begin
+                    new GeneralRegexSanitizer(regex: "SharedAccessKey=(?<key>[^;\\\"]+)", groupForReplace: "key"),
```
scbedd, line 1015: here.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -985,11 +998,177 @@ public void SetDefaultExtensions(string recordingId = null)
...
+                    new BodyKeySanitizer("$..atlasKafkaPrimaryEndpoint"),
+                    new BodyKeySanitizer("$..atlasKafkaSecondaryEndpoint"),
+                    new BodyKeySanitizer("$..certificatePassword"),
+                    new BodyKeySanitizer("$..storageAccountPrimaryKey"),
+                    new BodyKeySanitizer("$..privateKey"),
+                    new BodyKeySanitizer("$..fencingClientPassword"),
+                    new BodyKeySanitizer("$..acrToken"),
+                    new BodyKeySanitizer("$..scriptUrlSasToken"),
+                    new BodyKeySanitizer("$..refresh_token"),
+                    new BodyRegexSanitizer(regex: "(?<=<UserDelegationKey>).*?(?:<Value>)(.*)(?:</Value>)"),
```
JoshLove-msft, line 1126: I think we need to use GroupForReplace for these - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs#L160

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -985,11 +998,177 @@ public void SetDefaultExtensions(string recordingId = null)
...
+                    // "token": "sv=...ss=...srt=...se=...sp=...sig=" request body covered by 1052
+                    new BodyKeySanitizer("$..WEBSITE_AUTH_ENCRYPTION_KEY"),
+                    new BodyKeySanitizer("$..decryptionKey"),
+                    new HeaderRegexSanitizer("ServiceBusDlqSupplementaryAuthorization", regex: "(?:(sv|sig|se|srt|ss|sp)=)(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
+                    new HeaderRegexSanitizer("ServiceBusSupplementaryAuthorization", regex: "(?:(sv|sig|se|srt|ss|sp)=)(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
+                    // "RequestBody": "client_id=...grant_type=...client_info=...client_secret=…scope=...", covered by 1036/1037
                     new BodyKeySanitizer("$..access_token"),
-                    new BodyKeySanitizer("$..refresh_token")
+                    new BodyKeySanitizer("$..AccessToken"),
+                    //(client_id=)[^&]+ body regex // disabled, not a secret?
```
JoshLove-msft, line 1045: client_id is still considered PII

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -985,11 +998,177 @@ public void SetDefaultExtensions(string recordingId = null)
...
+                // - add a query param sanitizer
                 Sanitizers = new List<RecordedTestSanitizer>
                 {
+                    // basic RecordedTestSanitizer handles Authorization header
                     new RecordedTestSanitizer(),
+
+                    // new defaults begin
+                    new GeneralRegexSanitizer(regex: "SharedAccessKey=(?<key>[^;\\\"]+)", groupForReplace: "key"),
+                    new GeneralRegexSanitizer(regex: "AccountKey=(?<key>[^;\\\"]+)", groupForReplace: "key"),
+                    // "containerUrl": "<URL>/<Container>?sp=...st=...se=...spr=...sv=...sr=...sig="
```
JoshLove-msft, line 1017: Do we need a UriRegexSanitizer here?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -985,11 +998,177 @@ public void SetDefaultExtensions(string recordingId = null)
...
+                    // new HeaderRegexSanitizer("x-ms-encryption-key"), dup
+                    new HeaderRegexSanitizer("x-ms-rename-source"),
+                    new HeaderRegexSanitizer("x-ms-file-rename-source"),
+                    new HeaderRegexSanitizer("x-ms-copy-source"),
+                    new HeaderRegexSanitizer("x-ms-copy-source-authorization"),
+                    new HeaderRegexSanitizer("x-ms-file-rename-source-authorization"),
+                    new HeaderRegexSanitizer("x-ms-encryption-key-sha256"),
+                    new BodyKeySanitizer("$..uploadUrl"),
+                    new BodyKeySanitizer("$..logLink"),
+                    new HeaderRegexSanitizer("aeg-sas-token"),
```
JoshLove-msft, line 1095: nit: would it be worth sorting these based on the sanitizer type, and potentially alphabetically within each type?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -5,23 +5,31 @@
...
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Composition;
 using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Security;
+using System.Net.Sockets;
```
JoshLove-msft, line 23: Seem to be a few unneeded usings.

scbedd: Updated in `removable-sanitizers` branch.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: I'll do it now before I assign them internal ids 👍

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: Done.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: I think this one should actually be `$..containerUrl` getting sanitized out.

Looking at the common list + this master list, I don't see the SAS URI Suffix (for want of a better term) as a global sanitizer. So I'll add that.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: Added.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -985,11 +998,177 @@ public void SetDefaultExtensions(string recordingId = null)
...
+                    new BodyRegexSanitizer(regex: "(client_id=)(?<cid>[^&]+)", groupForReplace: "cid"),
+                    new BodyRegexSanitizer(regex: "client_secret=(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
+                    new BodyRegexSanitizer(regex: "client_assertion=(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
+                    // new BodyKeySanitizer("$..targetModelLocation"), disabled, not a secret?
+                    new BodyKeySanitizer("$..targetResourceId"),
+                    new BodyKeySanitizer("$..urlSource"),
+                    new BodyKeySanitizer("$..azureBlobSource.containerUrl"),
+                    new BodyKeySanitizer("$..source"),
+                    // new BodyKeySanitizer("$..resourceLocation"), disabled, not a secret?
+                    new HeaderRegexSanitizer("Location"),
```
chlowell, line 1054: The replacement value should be a valid URL to avoid breaking polling scenarios

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,8 +44,13 @@ public class SanitizerDictionary
             ResetSessionSanitizers();
         }
 
+        // TODO
+        // - create an option in sanitizer to get access to 'ignore case' and other regex customizations
+        // - turn commented lines into actual sanitizers after discussion
+        // - add a query param sanitizer
         public List<RegisteredSanitizer> DefaultSanitizerList = new List<RegisteredSanitizer>
```
scbedd, line 51: here.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: Updated to `example.com`

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -56,6 +62,606 @@ public class SanitizerDictionary
...
+                    new BodyKeySanitizer("$..azureBlobSource.containerUrl"),
+                    "AZSDK036"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..source"),
+                    "AZSDK037"
+                ),
+                // new BodyKeySanitizer("$..resourceLocation"), disabled, not a secret?
+                new RegisteredSanitizer(
+                    new HeaderRegexSanitizer("Location", value: "example.com"),
```
chlowell, line 208: Sorry I didn't underline this earlier, this value needs to a scheme e.g. `https://example.com` so it parses as a URL

scbedd: Updated. Thanks for the callout @chlowell

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: They are now organized by sanitizer type, and the default rules list has gaps if we need to insert some additional bodyregex vs bodykey vs generalregex sanitizers.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,654 @@ public class SanitizerDictionary
...
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..appId"),
+                    "AZSDK3432"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..userId"),
+                    "AZSDK3433"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..id"),
```
pvaneck, line 421: Duplicate of "AZSDK3430"?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,654 @@ public class SanitizerDictionary
...
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..managedResourceGroupName"),
+                    "AZSDK3441"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..createdBy"),
+                    "AZSDK3442"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..tenantId"),
```
JoshLove-msft, line 457: Can this be sanitized with Guid.Empty? The C# mgmt model deserializes this into a Guid - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/src/Resources/Generated/SubscriptionData.Serialization.cs#L164

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
scbedd: Thanks!

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,652 @@ public class SanitizerDictionary
...
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "client_assertion=(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
+                    "AZSDK3002"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "(?:(sv|sig|se|srt|ss|sp)=)(?<secret>[^&\\\"\\s]*)", groupForReplace: "secret"),
+                    "AZSDK3003"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "token=(?<token>[^&]+)($|&)", groupForReplace: "token"),
```
pvaneck, line 245: Noticed that this regex is breaking some tests since double quotes aren't checked. Should we use similar regex to how `client_assertion` and `client_secret` is handled?

For context, I am encountering an issue in some of the keyvault tests where I have a response body that ends in:
```
...
          {
            "kid": "https://vaultname.vault.azure.net/keys/livekvtestkey08d2f37be",
            "attributes": {
              "enabled": true,
              "created": 1667331450,
              "updated": 1667331450,
              "recoveryLevel": "CustomizedRecoverable\u002BPurgeable"
            }
          }
        ],
        "nextLink": "https://vaultname.vault.azure.net:443/keys?api-version=2016-10-01\u0026$skiptoken=abcdefc12323"
      }
```
This gets sanitized to `-....01&$skiptoken=Sanitized`. And the `"}` at the end gets cut off, leading to malformed JSON.

scbedd: Definitely worth the fix on this side. Will work it in.

scbedd: @pvaneck I'm making this change for _all_ of these BodyRegexSanitizers 🦖

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,652 @@ public class SanitizerDictionary
...
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "client_secret=(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
+                    "AZSDK3001"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "client_assertion=(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
+                    "AZSDK3002"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "(?:(sv|sig|se|srt|ss|sp)=)(?<secret>[^&\\\"\\s]*)", groupForReplace: "secret"),
```
pvaneck, line 241: Encountered false positives in some translation tests: `<span class=notranslate` -> `<span class=Sanitized`. Since this is generally for URI-formatted strings, should we check if these patterns are preceded by a `?` or `&`? (I _think_ the following should work: `(?:[\\?&](sv|sig|se|srt|ss|sp)=)(?<secret>[^&\\\"\\s]*)`). However, do these regexes still work if ampersands are encoded (`\u0026`)?

scbedd: I'll definitely take a look and adjust the regex. thanks @pvaneck

scbedd: > However, do these regexes still work if ampersands are encoded (\u0026)?

I want to say I think so but I'll make absolutely certain with a unit test in the proxy.

scbedd: I added a [test](https://github.com/Azure/azure-sdk-tools/pull/8038/commits/4fba36e5d41644de0d034c771ebe47ff22cd1339) that ensures we sanitize that ampersand. When we read it from the json file it's definitely decoded.

pvaneck: Thanks for verifying!

pvaneck: Also, in one of the storage tests, a SAS URI snuck through unsanitized through a `x-ms-previous-snapshot-url` header.  Since we have several places where we check for SAS URI query params (`AZSDK4002`, `AZSDK3003`, `AZSDK2002`, `AZSDK2003`, can we cover _all_ SAS URIs with a single GeneralRegexSanitizer?

scbedd: Let's try it. I'll do my best to add something.

scbedd: I added `AZSDK1007` which honors your request. Should do the trick. New version of the proxy will be `20240426.1` when it completes.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,641 @@ public class SanitizerDictionary
...
+                    "AZSDK1005",
+                    "ACS Identity leverages these strings to store identity information."
+                ),
+                new RegisteredSanitizer(
+                    new GeneralRegexSanitizer(regex: "/identities/(?<realm>[^/?]+)", groupForReplace: "realm"),
+                    "AZSDK1006",
+                    "ACS Identity leverages these strings to store identity information."
+                ),
+                new RegisteredSanitizer(
+                    new GeneralRegexSanitizer(regex: "(?:[?&](sv|sig|se|srt|ss|sp)=)(?<secret>[^&\\\"\\s]*)", groupForReplace: "secret"),
```
pvaneck, line 101:
```suggestion
                    new GeneralRegexSanitizer(regex: "(?:[?&](sv|sig|se|srt|ss|sp|st)=)(?<secret>[^&\\\"\\s]*)", groupForReplace: "secret"),
```

Forgot to mention this, but I noticed that this `st` query param (Start Time) would sometimes cause record mismatches in some python playback tests. I think it should be fine to sanitize this as well.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,641 @@ public class SanitizerDictionary
...
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..createdBy"),
+                    "AZSDK3442"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..tenantId", value: EMPTYGUID),
+                    "AZSDK3443"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..principalId"),
```
JoshLove-msft, line 452:
```suggestion
                    new BodyKeySanitizer("$..principalId", value: EMPTYGUID),
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,641 @@ public class SanitizerDictionary
...
+                    "AZSDK3444"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..clientId"),
```
JoshLove-msft, line 456:
```suggestion
                    new BodyKeySanitizer("$..clientId", value: EMPTYGUID),
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,641 @@ public class SanitizerDictionary
...
+                    "AZSDK1005",
+                    "ACS Identity leverages these strings to store identity information."
+                ),
+                new RegisteredSanitizer(
+                    new GeneralRegexSanitizer(regex: "/identities/(?<realm>[^/?]+)", groupForReplace: "realm"),
+                    "AZSDK1006",
+                    "ACS Identity leverages these strings to store identity information."
+                ),
+                new RegisteredSanitizer(
+                    new GeneralRegexSanitizer(regex: "(?:[?&](sv|sig|se|srt|ss|sp|st)=)(?<secret>[^&\\\"\\s]*)", groupForReplace: "secret"),
```
JoshLove-msft, line 101: Suggest removing "se" as this is an expiration time.

JoshLove-msft: Actually.. there are a bunch of other parameters that break .NET tests - essentially everything other than "sig" and "sv". I don't think the other parameters are secrets or even PII.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,641 @@ public class SanitizerDictionary
...
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "token=(?<token>[^&\\\"]+)($|&)", groupForReplace: "token"),
+                    "AZSDK3003"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "-----BEGIN PRIVATE KEY-----\\n(?<cert>.+\\n)*-----END PRIVATE KEY-----\\n", groupForReplace: "cert"),
+                    "AZSDK3004"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "(?<=<UserDelegationKey>).*?(?:<Value>)(?<group>.*)(?:</Value>)", groupForReplace: "group"),
```
JoshLove-msft, line 246: The value here is a Base64 string, so we should replace it with one. In .NET we use "Kg==" as a default base64 string.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,641 @@ public class SanitizerDictionary
...
+                    "AZSDK3005"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "(?<=<UserDelegationKey>).*?(?:<SignedTid>)(?<group>.*)(?:</SignedTid>)", groupForReplace: "group"),
```
JoshLove-msft, line 250: This one should use the empty GUID.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,641 @@ public class SanitizerDictionary
...
+                    "AZSDK3006"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "(?<=<UserDelegationKey>).*?(?:<SignedOid>)(?<group>.*)(?:</SignedOid>)", groupForReplace: "group"),
```
JoshLove-msft, line 254: This one should use the empty GUID.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
scbedd: Agreed. @pvaneck actually suggested we add one, but going the opposite direction will also help him.

So remove everything other than `sv|sig` from this?

scbedd: Eh posted a discussion item.

scbedd: After offline discussion reducing to just `sig`.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,642 @@ public class SanitizerDictionary
...
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..certificatePassword"),
+                    "AZSDK3459"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..clientSecret"),
+                    "AZSDK3460"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..keyVaultClientSecret"),
```
JoshLove-msft, line 521: Also need one for keyVaultClientId that should be a guid

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,642 @@ public class SanitizerDictionary
...
+                    "AZSDK3496"
+                ),
+                #endregion
+                #region UriRegex
+                new RegisteredSanitizer(
+                    new UriRegexSanitizer(regex: "sig=(?<sig>[^&]+)", groupForReplace: "sig"),
+                    "AZSDK4000"
+                ),
+                new RegisteredSanitizer(
+                    new UriRegexSanitizer("REDACTED", regex: "(?<=http://|https://)(?<host>[^/?\\.]+)", groupForReplace: "host"),
```
JoshLove-msft, line 675: Why do we use REDACTED here but Sanitized everywhere else?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,646 @@ public class SanitizerDictionary
...
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "client_secret=(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
+                    "AZSDK3001"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "client_assertion=(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
+                    "AZSDK3002"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "token=(?<token>[^&\\\"]+)($|&)", groupForReplace: "token"),
```
JoshLove-msft, line 239: I don't think the ($|&) is correct. I think it can be dropped.

JoshLove-msft, line 239:
```suggestion
                    new BodyRegexSanitizer(regex: "token=(?<token>[^&\\\"]+)", groupForReplace: "token"),
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,646 @@ public class SanitizerDictionary
...
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..accountKey"),
+                    "AZSDK3477"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..accountName"),
+                    "AZSDK3478"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..applicationId"),
```
JoshLove-msft, line 593:
```suggestion
                    new BodyKeySanitizer("$..applicationId", value: EMPTYGUID),
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -43,21 +47,646 @@ public class SanitizerDictionary
...
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..inputDataUri"),
+                    "AZSDK3427"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..containerUri"),
+                    "AZSDK3428"
+                ),
+                new RegisteredSanitizer(
+                    new BodyKeySanitizer("$..sasUri"),
```
JoshLove-msft, line 397: I think we need a regex here similar to https://github.com/Azure/azure-sdk-tools/pull/8038/files#diff-39fd1d9deed4a1781b44707fba78dcefcdd783c54d9bf446a2a18ccca1280ba6R671 so that we still have a valid URI

JoshLove-msft, line 397:
```suggestion
                    new BodyKeySanitizer("$..sasUri", regex: "sig=(?<sig>[^&]+)", groupForReplace: "sig"),
```

scbedd: Ah ok. Got it.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -32,52 +36,682 @@ public static ulong GetNextId()
...
+                    "AZSDK1005",
+                    "ACS Identity leverages these strings to store identity information."
+                ),
+                new RegisteredSanitizer(
+                    new GeneralRegexSanitizer(regex: "/identities/(?<realm>[^/?]+)", groupForReplace: "realm"),
+                    "AZSDK1006",
+                    "ACS Identity leverages these strings to store identity information."
+                ),
+                new RegisteredSanitizer(
+                    new GeneralRegexSanitizer(regex: "(?:[?&](sig)=)(?<secret>[^&\\\"\\s]*)", groupForReplace: "secret"),
```
JoshLove-msft, line 104:
```suggestion
                    new GeneralRegexSanitizer(regex: "(?:[?&](sig)=)(?<secret>[^&\\\"\\s\n,\\\\]*)", groupForReplace: "secret"),
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -32,52 +36,682 @@ public static ulong GetNextId()
...
+                    "AZSDK2028"
+                ),
+                new RegisteredSanitizer(
+                    new HeaderRegexSanitizer("x-ms-ests-server"),
+                    "AZSDK2029"
+                ),
+                #endregion
+                #region BodyRegex
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "(client_id=)(?<cid>[^&\\\"]+)", groupForReplace: "cid"),
```
JoshLove-msft, line 229:
```suggestion
                    new BodyRegexSanitizer(regex: "(client_id=)(?<cid>[^&\\\"\\s\n,\\\\]+)", groupForReplace: "cid"),
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -32,52 +36,682 @@ public static ulong GetNextId()
...
+                    "AZSDK3000"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "client_secret=(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
```
JoshLove-msft, line 233:
```suggestion
                    new BodyRegexSanitizer(regex: "client_secret=(?<secret>[^&\\\"\\s\n,\\\\]+)", groupForReplace: "secret"),
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -32,52 +36,682 @@ public static ulong GetNextId()
...
+                    "AZSDK3001"
                 ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "client_assertion=(?<secret>[^&\\\"]+)", groupForReplace: "secret"),
```
JoshLove-msft, line 237:
```suggestion
                    new BodyRegexSanitizer(regex: "client_assertion=(?<secret>[^&\\\"\\s\n,\\\\]+)", groupForReplace: "secret"),
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -32,52 +36,682 @@ public static ulong GetNextId()
...
+                    "AZSDK3002"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "token=(?<token>[^&\\\"]+)", groupForReplace: "token"),
```
JoshLove-msft, line 241:
```suggestion
                    new BodyRegexSanitizer(regex: "token=(?<token>[^&\\\"\\s\n,\\\\)", groupForReplace: "token"),
```

scbedd, line 241:
```suggestion
                    new BodyRegexSanitizer(regex: "token=(?<token>[^&\\\"\\s\\n,\\\\]+)", groupForReplace: "token"),
```

I think it should be this right? To escape the newline.

scbedd: I'm taking the change, just slightly adjusting it 👍

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -32,52 +36,682 @@ public static ulong GetNextId()
...
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "client_secret=(?<secret>[^&\\\"\\s\\n,\\\\]+)", groupForReplace: "secret"),
+                    "AZSDK3001"
                 ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "client_assertion=(?<secret>[^&\\\"\\s\\n,\\\\]+)", groupForReplace: "secret"),
+                    "AZSDK3002"
+                ),
+                new RegisteredSanitizer(
+                    new BodyRegexSanitizer(regex: "token=(?<token>[^&\\\"\\s\\n,\\\\])", groupForReplace: "token"),
```
pvaneck, line 241: For Python, we added this as a general regex sanitizer, since we ran into cases where a Request URI containing `skiptoken` depended on a URI value in the previous request's body. In these cases,  `skiptoken` would be sanitized in bodies, but not Request URIs, causing recording mismatches. 

Others may encounter this, so it might be worth making this General or adding this as a URIRegexSantizer as well.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
```diff
@@ -32,52 +36,682 @@ public static ulong GetNextId()
...
+                    "AZSDK2002"
+                ),
+                new RegisteredSanitizer(
+                    new HeaderRegexSanitizer("Location", value: "https://example.com"),
+                    "AZSDK2003"
+                ),
+                new RegisteredSanitizer(
+                    new HeaderRegexSanitizer("subscription-key"),
+                    "AZSDK2004"
+                ),
```
samvaity, line 125:
```suggestion
                new RegisteredSanitizer(
                    new HeaderRegexSanitizer("Location", value: "https://example.com"),
                    "AZSDK2003"
                ),
                new RegisteredSanitizer(
                    new HeaderRegexSanitizer("subscription-key"),
                    "AZSDK2004"
                ),
                 new RegisteredSanitizer(
                    new HeaderRegexSanitizer("operation-location", value: "https://example.com"),
                    "AZSDK2004"
                ),
                new RegisteredSanitizer(
                    new HeaderRegexSanitizer("Ocp-Apim-Subscription-Key"),
                    "AZSDK2005"
                ),
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
JoshLove-msft: This is happening in .NET as well.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
scbedd: Took this suggestion, but not taking the direct commit because the ID numbers also need to increment properly.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/SanitizerDictionary.cs
scbedd: Sounds like I should make this `General`. I'll upgrade it.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitAssetsConfiguration.cs
```diff
@@ -127,7 +127,7 @@ public bool IsAssetsRepoInitialized(bool autoCreate = true)
         {
             var location = Path.Join(ResolveAssetRepoLocation().ToString(), ".git");
 
-            return Directory.Exists(location);
+            return Directory.Exists(location) | File.Exists(location);
```
benbp, line 130: I assume you mean:

```suggestion
            return Directory.Exists(location) || File.Exists(location);
```

benbp (approved):

scbedd:
Thank you to all the reviewers on this PR: @pvaneck , @JoshLove-msft , @chlowell, @samvaity . Hugely appreciate all your feedback.

`go`, `python,` `.net`, and `cpp` have shifted onto a proxy version produced by this PR. Only `java` and `js` remain to migrate, and they are _almost_ there. Given all that. I'm going to merge this PR, as I have further PRs to assist debugging that should _include_ these changes.

benbp (approved):

PR #8053, opened 2024-04-09
Pass flag to suppress errors correctly when verifying release change log

praveenkuttappan:
SuppressErrors flag was incorrectly set as true when verifying change log for releases.

### DIFF
--- a/eng/common/scripts/ChangeLog-Operations.ps1
+++ b/eng/common/scripts/ChangeLog-Operations.ps1
@@ -188,7 +188,7 @@ function Confirm-ChangeLogEntry {
   if ($ForRelease -eq $True)
   {
     LogDebug "Verifying as a release build because ForRelease parameter is set to true"
-    return Confirm-ChangeLogForRelease -changeLogEntry $changeLogEntry -changeLogEntries $changeLogEntries -ChangeLogStatus $ChangeLogStatus
+    return Confirm-ChangeLogForRelease -changeLogEntry $changeLogEntry -changeLogEntries $changeLogEntries -ChangeLogStatus $ChangeLogStatus -SuppressErrors $suppressErrors
   }
 
   # If the release status is a valid date then verify like its about to be released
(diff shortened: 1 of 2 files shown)


eng/common/scripts/ChangeLog-Operations.ps1
```diff
@@ -361,26 +361,24 @@ function Confirm-ChangeLogForRelease {
...
   )
 
-  $suppressErrors = $false
   if (!$ChangeLogStatus) {
     $ChangeLogStatus = [PSCustomObject]@{
       IsValid = $false
       Message = ""
     }
   }
-  else {
```
weshaggard, line 374: I think you still want this else. If someone calls this function directly passing in empty changelogStatus you want the suppressErrors to be true.

weshaggard (approved):
One comment otherwise looks good.

eng/common/scripts/ChangeLog-Operations.ps1
praveenkuttappan: But then passed value for SuppressErrors will be overwritten here when both suppressErrors and changelogStatus are passed. I will need a change to avoid passing changelogStatus from caller when suppression is not required. changelogstatus should only be passed when suppression is required.

praveenkuttappan: @weshaggard By default errors will be reported as errors by confirmchangelog for release function unless caller explicitly says to suppress errors. Flipping this flag based on changelogStatus here causes the same issue since changelogStatus is always set when verifying from PR.

praveenkuttappan: I have created a test PR using this change to verify the error.
https://dev.azure.com/azure-sdk/public/_build/results?buildId=3681685&view=logs&jobId=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=c9190431-7417-53a5-3fc1-7c7e026cd79d

weshaggard: Should we follow this same pattern in the Confirm-ChangeLog function then as well and force folks to pass in the suppressErrors flag? Otherwise depending on which gets called there might be different functionality.

weshaggard (approved):

PR #8061, opened 2024-04-10
[Pylint] Don't allow typing under TYPE_CHECKING

l0lawrence:
Fixes https://github.com/Azure/azure-sdk-tools/issues/3230

https://pylint.pycqa.org/projects/astroid/en/latest/api/nodes/astroid.nodes.If.html Since an `IF` statement doesn't track what the logical argument is the best we can do is check if typing is being imported under an "if". I can't think of a scenario where this would happen as we no longer have to deal with py3.6 back compat


tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2708,6 +2708,31 @@ def visit_importfrom(self, node):
...
+    msgs = {
+        "C4760": (
+            "Do not import from typing inside of TYPE_CHECKING.",
+            "no-typing-import-in-type-check",
+            "Do not import from typing inside of TYPE_CHECKING. You can import from typing outside of TYPE_CHECKING.",
+        ),
+    }
+
+    def visit_importfrom(self, node):
+        """Check that we aren't importing from azure.core.pipeline.transport import HttpResponse."""
```
kristapratico, line 2727: Comment seems like artifact of copy/paste

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2708,6 +2708,31 @@ def visit_importfrom(self, node):
...
+        if isinstance(node.parent, astroid.If) and node.modname == "typing":
```
kristapratico, line 2728: Can we also check for typing_extensions?

kristapratico (approved):
Looks good to me.

PR #8080, opened 2024-04-12
Switch away from SAS for upload link cache file

weshaggard:


eng/pipelines/githubio-linkcheck.yml
```diff
@@ -6,92 +6,14 @@ pool:
   vmImage: ubuntu-22.04
 
 variables:
-  cachefile: verify-links-cache.txt
+  cachefile: verify-links-cache-test.txt
```
weshaggard, line 9:
```suggestion
  cachefile: verify-links-cache.txt
```

weshaggard:
Looks like AzureFileCopy task only supports windows https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/azure-file-copy-v6?view=azure-pipelines#remarks so we will be forced to use the AzurePowershell approach.

PR #8086, opened 2024-04-13
Authenticate requests from APIView to DevOps using managed Identity

praveenkuttappan:
Changes to authenticate requests to DevOps using managed identity.

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Repositories/DevopsArtifactRepository.cs
+++ b/src/dotnet/APIView/APIViewWeb/Repositories/DevopsArtifactRepository.cs
@@ -13,6 +13,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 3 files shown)


src/dotnet/APIView/APIViewWeb/Repositories/DevopsArtifactRepository.cs
```diff
@@ -1,160 +1,136 @@
...
-            }
-            return 0;
+        private async Task<int> GetPipelineId(string pipelineName, BuildHttpClient client, string projectName)
+        {          
+            var pipelines = await client.GetFullDefinitionsAsync2(project: projectName);
+            if (pipelines != null)
+            {
+                return pipelines.Single(p => p.Name == pipelineName).Id;
+            }
+            return 0;
```
praveenkuttappan, line 133:
```suggestion
        private async Task<int> GetPipelineId(string pipelineName, BuildHttpClient client, string projectName)
        {          
            var pipelines = await client.GetFullDefinitionsAsync2(project: projectName);
            if (pipelines != null)
            {
                return pipelines.Single(p => p.Name == pipelineName).Id;
            }
            return 0;
```

weshaggard (approved):
Let's be sure to clean-up this PAT once this gets deployed and becomes functional.

PR #8112, opened 2024-04-17
Add polly retry to get artifact URL from DevOps

praveenkuttappan:
Request to get artifact URL is still failing with 404 error. Adding polly retry to get URL.URL is working fine when attempted again.

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Repositories/DevopsArtifactRepository.cs
+++ b/src/dotnet/APIView/APIViewWeb/Repositories/DevopsArtifactRepository.cs
@@ -56,11 +56,10 @@ namespace APIViewWeb.Repositories
 
         private async Task<string> getDownloadArtifactUrl(string buildId, string artifactName, string project)
         {
-            var maxRetryAttempts = 5;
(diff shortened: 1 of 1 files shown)


src/dotnet/APIView/APIViewWeb/Repositories/DevopsArtifactRepository.cs
```diff
@@ -56,10 +56,25 @@ public async Task<Stream> DownloadPackageArtifact(string repoName, string buildI
 
         private async Task<string> getDownloadArtifactUrl(string buildId, string artifactName, string project)
         {
+            var pauseBetweenFailures = TimeSpan.FromSeconds(2);
```
maririos, line 59: is this number arbritary? or it has history to it

maririos (approved):

src/dotnet/APIView/APIViewWeb/Repositories/DevopsArtifactRepository.cs
praveenkuttappan: 2 seconds should be ideally enough to start the delay span. So, it's just arbitrary.

PR #8113, opened 2024-04-17
Add indentation handling method in Java ApiView tool

alzimmermsft:
Adds `indentedBlock` to handle indenting and unindenting a section of tokens rather than relying on the methods indent and unindent to be called before and after. Also changes indentation handling to be based on the indentation level in case we want to offer different indentation size (right now it's always 4 whitespaces). Lastly, added a cache for common indentation levels so they can be reused.

### DIFF
--- a/src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
+++ b/src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
@@ -106,7 +106,13 @@ public class JavaASTAnalyser implements Analyser {
 
     private static final boolean SHOW_JAVADOC = true;
 
+    // Keep a list of indent levels so we don't need to recreate a string each time indenting is done.
+    // Start with an initial set of levels.
+    private static final List<String> INDENT_LEVELS;
+
     private static final Map<String, AnnotationRule> ANNOTATION_RULE_MAP;
+    private static final JavaParserAdapter JAVA_8_PARSER;
+    private static final JavaParserAdapter JAVA_11_PARSER;
     static {
         /*
          For some annotations, we want to customise how they are displayed. Sometimes, we don't show them in any
@@ -120,12 +126,7 @@ public class JavaASTAnalyser implements Analyser {
 
         // we always want @Metadata annotations to be fully expanded, but in a condensed form
         ANNOTATION_RULE_MAP.put("Metadata", new AnnotationRule().setShowProperties(true).setCondensed(true));
-    }
 
-    private static final JavaParserAdapter JAVA_8_PARSER;
-    private static final JavaParserAdapter JAVA_11_PARSER;
-
-    static {
         // Configure JavaParser to use type resolution
         JAVA_8_PARSER = JavaParserAdapter.of(new JavaParser(new ParserConfiguration()
             .setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_8)
@@ -134,6 +135,9 @@ public class JavaASTAnalyser implements Analyser {
         JAVA_11_PARSER = JavaParserAdapter.of(new JavaParser(new ParserConfiguration()
             .setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_11)
             .setDetectOriginalLineSeparator(false)));
+
+        INDENT_LEVELS = new ArrayList<>(Arrays.asList("", makeNewWhitespace(4), makeNewWhitespace(8),
+            makeNewWhitespace(12), makeNewWhitespace(16), makeNewWhitespace(20), makeNewWhitespace(24)));
     }
 
     // This is the model that we build up as the AST of all files are analysed. The APIListing is then output as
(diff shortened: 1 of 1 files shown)


src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -1518,22 +1525,32 @@ private void visitJavaDoc(JavadocComment javadoc) {
...
-        byte[] bytes = new byte[indent];
+        return new Token(WHITESPACE, makeWhitespace(indentLevel));
+    }
+
+    private static String makeWhitespace(int indentLevel) {
+        if (indentLevel < INDENT_LEVELS.size()) {
+            return INDENT_LEVELS.get(indentLevel);
+        }
+
+        return makeNewWhitespace(indentLevel * 4);
```
srnagar, line 1546: Should we add the new indent level to the cached list (or use a map instead) to avoid recreating it again?

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -302,96 +302,103 @@ private void tokeniseMavenPom(Pom mavenPom) {
...
+            //            addToken(new Token(PUNCTUATION, "{"), NEWLINE);
+            //            indent();
+            //            mavenPom.getAllowedDependencies().stream().forEach(value -> {
+            //                addToken(INDENT, new Token(TEXT, value, value), NEWLINE);
+            //            });
+            //            unindent();
+            //            addToken(INDENT, new Token(PUNCTUATION, "}"), NEWLINE);
+            //        }
+
+            // close maven
```
srnagar, line 399: Do we need to retain this? Seems like this has been commented out from a long time.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -134,6 +135,9 @@ public class JavaASTAnalyser implements Analyser {
         JAVA_11_PARSER = JavaParserAdapter.of(new JavaParser(new ParserConfiguration()
             .setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_11)
             .setDetectOriginalLineSeparator(false)));
+
+        INDENT_LEVELS = new ArrayList<>(Arrays.asList("", makeNewWhitespace(4), makeNewWhitespace(8),
```
srnagar, line 139: We may not reach 24 very often or even 20. Should we lazily grow this and add to this cache when we create it the first time? Probably 4, 8, 12, 16 are more common and we can start there?

srnagar (approved):

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
JonathanGiles: It makes me nervous to hard code this. I would rather just compute it as necessary.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
alzimmermsft: I'm not really sure, this has been commented out for a while and I'm guessing this is something we just haven't implemented but want to? @JonathanGiles

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
alzimmermsft: I can remove this if we don't think it's worthwhile

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
alzimmermsft: I'm not expecting for there to be indentation levels this deep and if there are they can be created on use

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
JonathanGiles: yeah, it migh as well go for now.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
JonathanGiles: I would probably remove it

PR #8115, opened 2024-04-17
Introducing a test framework for Java APIView processor

JonathanGiles:
Introducing a test framework for Java APIView processor to compare output for published libraries against their known-good representations, to help avoid regrettable regressions. So far there are no tests included - this is just the framework.

### DIFF
--- a/src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/Main.java
+++ b/src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/Main.java
@@ -22,6 +22,7 @@ import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.jar.JarEntry;
@@ -33,7 +34,7 @@ import static com.azure.tools.apiview.processor.model.TokenKind.LINE_ID_MARKER;
 public class Main {
     // expected argument order:
     // [inputFiles] <outputDirectory>
-    public static void main(String[] args) throws IOException {
+    public static void main(String[] args) {
         if (args.length != 2) {
             System.out.println("Expected argument order: [comma-separated sources jarFiles] <outputFile>, e.g. /path/to/jarfile.jar ./temp/");
             System.exit(-1);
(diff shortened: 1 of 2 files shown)


src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/Main.java
```diff
@@ -43,28 +44,34 @@ public static void main(String[] args) throws IOException {
...
-            final File outputFile = new File(outputDir, jsonFileName);
-            processFile(file, outputFile);
+        if (!jarFile.exists()) {
+            System.out.printf("Cannot find file '%s'%n", jarFile);
+            System.exit(-1);
         }
+
+        if (!outputDir.exists()) {
+            if (!outputDir.mkdirs()) {
+                System.out.printf("Failed to create output directory %s%n", outputDir);
```
srnagar, line 67: Should we `System.exit(-1)` from here if output directory cannot be created?

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/Main.java
```diff
@@ -126,13 +133,22 @@ private static void processFile(final File inputFile, final File outputFile) thr
...
         }
 
-        // Write out to the filesystem
-        try (JsonWriter jsonWriter = JsonProviders.createWriter(Files.newBufferedWriter(outputFile.toPath()))) {
-            apiListing.toJson(jsonWriter);
+        try {
+            // Write out to the filesystem, make the file if it doesn't exist
+            if (!outputFile.exists()) {
+                if (!outputFile.createNewFile()) {
+                    System.out.printf("Failed to create output file %s%n", outputFile);
```
srnagar, line 140: Same here - exit if file cannot be created.

srnagar (approved):

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/Main.java
```diff
@@ -126,13 +133,22 @@ private static void processFile(final File inputFile, final File outputFile) thr
...
+                }
+            }
+            try (JsonWriter jsonWriter = JsonProviders.createWriter(Files.newBufferedWriter(outputFile.toPath()))) {
+                apiListing.toJson(jsonWriter);
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
alzimmermsft, line 147: Why do we choose to print the exception stack trace instead of failing the run?

alzimmermsft (approved):

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/Main.java
JonathanGiles: in this case I think we keep going - just because we support multiple files, and the others might work.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/Main.java
JonathanGiles: Done

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/Main.java
JonathanGiles: Diagnostics - also, soon APIView will start to show the output from the parser, so we can start to diagnose issues with generating reviews from within the UI.

PR #8120, opened 2024-04-18
Removable Sanitizers

scbedd:
TODO:

- [x] Sanitizers now have an `id`
- [x] Existing behavior of the proxy WRT registering a sanitizer for a session/recording is maintained, but now sanitizers are stored in a central location
- [x] Investigate 2 failing unit tests from JS integration test run, resolve.
  - ~~Getting some assistance from @HarshaNalluru  on this failure, thank you Harsha!~~
  - `2ece462` resolved issues. Glad the tests caught it thanks again Harsha!
- [x] POST Admin/AddSanitizer now returns a body containing the newly added sanitizerId
- [x] GET Admin/Sanitizers is a new route that returns either the session sanitizers (with ids) or specific recordingId based on the presence of `x-recording-id` header
- [x] POST Admin/RemoveSanitizer
- [x] and POST Admin/RemoveSanitizers should _remove_ the given sanitizer Ids from the active session or recording
  - Arguably I should use `DELETE` here, but some clients run into problems sending along a body with a DELETE request, so I'm just going to keep this `POST` and keep it dumb simple.
- [x] Tests for RemoveSanitizer

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/launchSettings.json
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/launchSettings.json
@@ -2,7 +2,6 @@
   "profiles": {
     "Azure.Sdk.Tools.TestProxy": {
       "commandName": "Project",
-      "commandLineArgs": "--storage-location=C:/repo/sdk-for-js",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
         "Logging__LogLevel__Microsoft": "Information"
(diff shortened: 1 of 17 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/launchSettings.json
```diff
@@ -2,6 +2,7 @@
   "profiles": {
     "Azure.Sdk.Tools.TestProxy": {
       "commandName": "Project",
+      "commandLineArgs": "--storage-location=C:/repo/sdk-for-js",
```
benbp, line 5: Dev needs to be removed?

benbp (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/launchSettings.json
scbedd: Yes absolutely. Thanks for catching it.

PR #8153, opened 2024-04-23
Remove failure analysis and apply code cleanup

hallipr:


tools/pipeline-witness/infrastructure/Assign-StoragePermissions.ps1
```diff
@@ -0,0 +1,70 @@
...
+    Write-Error "Failed to grant access"
+    exit 1
+  }
+    
+  Write-Host "Granting 'Storage Queue Data Contributor' access to $appResourceGroupName/$appStorageAccountName"
+  $scope = "/subscriptions/$subscriptionId/resourceGroups/$appResourceGroupName/providers/Microsoft.Storage/storageAccounts/$appStorageAccountName"
+  $output = Invoke "az role assignment create --assignee '$azAdGroupId' --role 'Storage Queue Data Contributor' --scope '$scope' --output none"
+    
+  if ($LASTEXITCODE -ne 0) {
+    Write-Output $output
```
benbp, line 63: I'm ok with this but is there a reason you didn't put the role assignments into the bicep templates?

benbp (approved):

tools/pipeline-witness/infrastructure/Assign-StoragePermissions.ps1
hallipr: The storage assignments are only needed when a dev is running the service locally and connecting to live resources. This doesn't need to be persistent access

hallipr: Also, in the pipeline, the deployment principal doesn't have permission to query Entra.  We'd need to hard code the group's object id into the parameters and that didn't seem correct.

PR #8160, opened 2024-04-26
Use azcopy to access azure blob with FWI

raych1:
Use the federated workload identity to access azure blob.

### DIFF
--- a/tools/sdk-ai-bots/.pipelines/build-document-embeddings.yml
+++ b/tools/sdk-ai-bots/.pipelines/build-document-embeddings.yml
@@ -19,7 +19,7 @@ variables:
   asch-endpoint: $(azure-search-endpoint)
   asch-index-name: $(azure-search-index-name)
   aoai-embedding-model: $(azure-openai-embedding-model)
-  service-connection-name: $(service-connection-name)
+  azure-subscription-name: $(service-connection-name)
 
 parameters:
 - name: incrementalEmbedding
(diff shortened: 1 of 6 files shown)


tools/sdk-ai-bots/.pipelines/build-document-embeddings.yml
```diff
@@ -35,12 +35,13 @@ stages:
...
     - checkout: git://internal/_git/azure-sdk-docs-eng.ms
       displayName: 'Checkout azure-sdk-docs-eng.ms repository'
-    - task: Powershell@2
+    - task: AzurePowerShell@5
       inputs:
-        filePath: $(Build.SourcesDirectory)/azure-sdk-tools/tools/sdk-ai-bots/Scripts/Build-EngHubDocEmbeddings.ps1
-        arguments: >
-          -IncrementalEmbedding "${{ parameters.incrementalEmbedding }}"
-        pwsh: true
+        azureSubscription: 'Azure SDK Playground'
```
lirenhe, line 40: Can we make this as a parameter instead of hard coded value?

tools/sdk-ai-bots/Scripts/Common.ps1
```diff
@@ -31,6 +31,39 @@ function Clone-Repository {
...
+
+function Download-AzCopy {
+    param (
+        [Parameter(Position = 0)]
+        [ValidateNotNullOrEmpty()]
+        [string] $DestinationPath
+    )
+
+    try {
+        $AzCopyUrl = "https://azcopyvnext.azureedge.net/release20220315/azcopy_windows_amd64_10.14.1.zip"
```
lirenhe, line 55: This is also a fixed value for the location, will it change in the future and can we use a parameter with default value?

tools/sdk-ai-bots/.pipelines/build-document-embeddings.yml
raych1: Updated.

tools/sdk-ai-bots/Scripts/Common.ps1
raych1: This URL will only be used locally and won't be used in pipeline scenario.

lirenhe (approved):

raych1:
The pipeline failed due to the azure blob access policy change which doesn't allow the public internet access.

PR #8162, opened 2024-04-26
Lock session during recording

scbedd:
The instant we start handling a request get a lock on the session so it can't be touched by another thread.


scbedd:
Nope.

PR #8180, opened 2024-05-01
[ci] Add support for sdktype "utility" from the JS repo

xirzec:
In JS we use sdk-type: "utility" in our package manifests for things in the `@azure-tools` namespace, but this isn't one of the supported values in ADO work items for releases.

This can cause problems in CI like shown here:

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3751191&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=00913ed3-6224-599f-5780-70386e2dd22d

example of the build issue

This PR works around the problem by substituting the invalid string "utility" for the permitted work item value of "tool"

### DIFF
--- a/eng/common/scripts/Validate-Package.ps1
+++ b/eng/common/scripts/Validate-Package.ps1
@@ -163,6 +163,11 @@ function CreateUpdatePackageWorkItem($pkgInfo)
         $setReleaseState = $false
         $plannedDate = "unknown"
     }
+    $sdkType = $pkgInfo.SDKType;
+    # JS uses "utility" instead of "tool" in package.json
+    if ($sdkType -eq "utility") {
+        $sdkType = "tool"
+    }
         
     # Create or update package work item  
     &$EngCommonScriptsDir/Update-DevOps-Release-WorkItem.ps1 `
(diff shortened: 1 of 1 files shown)


joheredi (approved):

HarshaNalluru (approved):

mpodwysocki (approved):

weshaggard:
@xirzec we could also add utility as a valid value in the Package work item if that is a type we want to support.

praveenkuttappan:
I would prefer to add the type `utility` as an option in work item instead of overriding it in one place to keep it consistent in repo and work item. I can add the type in work item template if you are okay with that approach.  Otherwise, we should update it as tool in repo.

xirzec:
> I would prefer to add the type `utility` as an option in work item instead of overriding it in one place to keep it consistent in repo and work item. I can add the type in work item template if you are okay with that approach. Otherwise, we should update it as tool in repo.

Yeah, I'm fine with updating the template, I just didn't know how difficult it would be to do that. I'll go ahead and close this PR.

PR #8182, opened 2024-05-02
Show the latest service version

srnagar:
This PR updates the service version type to show the service version returned by the `getLatest()` method.

image


JonathanGiles:
I've added this to the new APIView api parser implementation.

PR #8246, opened 2024-05-10
SAS token removal from team/user store

JimSuplizio:
Use azure credentials instead of the SAS token for the team/user store.

In the pipeline use the AzureCLI task to authenticate to the Azure SDK Artifacts, which has already been setup. As mentioned in the comments, we can't use the DefaultAzureCredential. The reason for this is because, in CI pipelines, it picks up the ManagedIdentityCredential over the AzureCliCredential and the AzureCliCredential is what we need to be using.


weshaggard (approved):
Just remember to undo your commented out stuff.

JimSuplizio:
I need to override check-enforcer. The reason is GitHubTeamUserStore can't automatically kick off a pipeline in PRs due to the number of GitHub API calls required to get the data (2000) and the fact that there's a 5000/hour limit in the PAT.

PR #8252, opened 2024-05-13
update dep and resolve typing issue

tadelesh:


raych1:
Why do we want to make this PR? Does it come from the security alerts?

tadelesh:
> Why do we want to make this PR? Does it come from the security alerts?

nope, it is for typing fix. when i try to upgrade in go testgen, it failed for ts typing check.

raych1 (approved):

PR #8261, opened 2024-05-13
Support Scan Breakglass

scbedd:
Getting a simple feature in ahead of the need.

I can definitely see us doing something more specific, like an internal suppression mechanism or the like, but I want to be able to unblock folks on their machines if they absolutely need it.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/CommandOptions/OptionsGenerator.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/CommandOptions/OptionsGenerator.cs
@@ -56,10 +56,9 @@ namespace Azure.Sdk.Tools.TestProxy.CommandOptions
             universalOption.AddAlias("-u");
 
             var breakGlassOption = new Option<bool>(
-                name: "--breakGlass",
+                name: "--break-glass",
                 description: "Flag; Ignore secret push protection results when pushing.",
                 getDefaultValue: () => false);
-            breakGlassOption.AddAlias("-b");
 
             var collectedArgs = new Argument<string[]>("args")
             {
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/CommandOptions/OptionsGenerator.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/CommandOptions/OptionsGenerator.cs
@@ -55,6 +55,11 @@ namespace Azure.Sdk.Tools.TestProxy.CommandOptions
                 getDefaultValue: () => false);
             universalOption.AddAlias("-u");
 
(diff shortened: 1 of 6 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/CommandOptions/OptionsGenerator.cs
```diff
@@ -55,6 +55,11 @@ public static RootCommand GenerateCommandLineOptions(Func<DefaultOptions, Task>
                 getDefaultValue: () => false);
             universalOption.AddAlias("-u");
 
+            var breakGlassOption = new Option<bool>(
+                name: "--breakGlass",
```
benbp, line 59:
```suggestion
                name: "--break-glass",
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/CommandOptions/OptionsGenerator.cs
```diff
@@ -55,6 +55,11 @@ public static RootCommand GenerateCommandLineOptions(Func<DefaultOptions, Task>
...
+                description: "Flag; Ignore secret push protection results when pushing.",
+                getDefaultValue: () => false);
+            breakGlassOption.AddAlias("-b");
```
benbp, line 62: For stuff like this I would say let's not add a shorthand alias.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
```diff
@@ -145,8 +146,11 @@ public bool CheckForSecrets(GitAssetsConfiguration assetsConfiguration, string[]
             {
                 if (CheckForSecrets(config, pendingChanges))
                 {
-                    Environment.ExitCode = -1;
-                    return;
+                    if (!ignoreSecretProtection)
+                    {
+                        Environment.ExitCode = -1;
```
benbp, line 151: Is this a not implemented type thing right now?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/CommandOptions/OptionsGenerator.cs
scbedd, line 62:
```suggestion
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
scbedd: I don't get what you're implying here.

benbp (approved):

PR #8262, opened 2024-05-14
Use entra app id and certificate for authentication

raych1:
Enable certificate auth for both local development and the Azure deployed service.


tools/sdk-ai-bots/AzureSdkQaBot/Program.cs
```diff
@@ -21,9 +23,44 @@
...
+        throw new Exception($"Certificate {certificateName} not found in KeyVault {config.KeyVaultUrl}");
+    }
+}
+catch (Exception ex)
+{
+    throw new Exception($"Failed to get certificate {config.CertificateName} from KeyVault {config.KeyVaultUrl}", ex);
+}
+
+// Create the ClientCredentialsFactory to user certificate authentication
+builder.Services.AddSingleton<ServiceClientCredentialsFactory>((e) => new CertificateServiceClientCredentialsFactory(certificate, config.BOT_ID, "72f988bf-86f1-41af-91ab-2d7cd011db47"));
```
lirenhe, line 63: can we load "72f988bf-86f1-41af-91ab-2d7cd011db47" from config as well?

raych1: It won't change as we stick to use this tenant.

lirenhe (approved):

Issue #8269, opened 2024-05-15
The eng-common sync pipeline fails to merge the PR


raych1:
The merge step fails due to the following error:
```powershell
Merging https://api.github.com/repos/Azure/azure-sdk-for-android/pulls/1571
##[error]Invoke-RestMethod [https://api.github.com/repos/Azure/azure-sdk-for-android/pulls/1571/merge] failed with exception:
The given key 'Content-Type' was not present in the dictionary.
##[error]PowerShell exited with code '1'.

``` 
Here is the [pipeline run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3789534&view=logs&j=aeb03408-248b-5400-e2ce-c3865561c192&t=3b398d7a-d3bc-58c2-66b1-87b9123081b2&s=28968f5d-273a-5e19-c95c-aedd7587b6da) triggered by [this PR](https://github.com/Azure/azure-sdk-tools/pull/8268). I had to manually merge all the sync PRs. I noticed that [the previous run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3726666&view=logs&j=aeb03408-248b-5400-e2ce-c3865561c192&t=3b398d7a-d3bc-58c2-66b1-87b9123081b2) of the sync pipeline already had this issue and the PR author had to merge the PR manually.

I met this issue previously when doing the Go SDK release and made a fix to the common script by this [PR](https://github.com/Azure/azure-sdk-tools/pull/7630).  The same fix should be applied to the sync script.

@weshaggard

weshaggard:
Thanks for reporting I've seen this error once before but not sure what causes it. I think this process does still work in general though. I'll keep an eye on this the next eng/common sync I see.

PR #8275, opened 2024-05-15
Add Content-Type to the GH PR api call

weshaggard:
We were hitting an error caused by some recent GH changes. Based on https://github.com/orgs/community/discussions/110246 you now need to explicitly set Content-Type in the headers to avoid the error:

The given key 'Content-Type' was not present in the dictionary.

Fixes https://github.com/Azure/azure-sdk-tools/issues/8269

### DIFF
--- a/eng/scripts/Verify-And-Merge-PRs.ps1
+++ b/eng/scripts/Verify-And-Merge-PRs.ps1
@@ -8,7 +8,7 @@ param(
 
 $ReadyForMerge = $true
 $mergablePRs = @()
-$headers = @{ }
+$headers = @{ "Content-Type" = "text/json" }
 $RetryCount = 5
 
 if ($null -eq $ShouldMerge) {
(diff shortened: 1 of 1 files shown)


weshaggard:
tested locally to merge https://github.com/Azure/azure-sdk/pull/7564

benbp (approved):

eng/scripts/Verify-And-Merge-PRs.ps1
```diff
@@ -8,15 +8,15 @@ param(
...
 $RetryCount = 5
 
 if ($null -eq $ShouldMerge) {
   $ShouldMerge = $null -ne $AuthToken;
 }
 
 if ($AuthToken) {
-  $headers = @{
+  $headers += @{
     Authorization = "bearer $AuthToken"
```
benbp, line 20: Does this work for merging objects?

weshaggard: Yes it does from my local testing. I didn't test to see what happens if there is a conflict but the syntax here does work.

weshaggard: From a quick test if the key already exists it fails, but that shouldn't be the case here.

PR #8286, opened 2024-05-16
Remove a duplicate sanitizer from the proxy config

scbedd:
Brought to my attention by @chlowell


benbp (approved):
LGTM. Wonder if a hashset would be useful here.

Hashsets don't maintain order, which we enforce for simplicity of debugging sanitizers by the users.

scbedd:
<I somehow edited your comment instead of responding? somehow?>

Hashsets don't maintain order, which we enforce for simplicity of debugging sanitizers by the users.

PR #8288, opened 2024-05-17
Remove SAS token from GitHubIO blob storage docs publish

JimSuplizio:
GitHubIO versioned docs are published to blob storage which required the SAS key to write.

As part of this work, I did have to give Storage Blob Data Contributor access to azuresdkdocs to the service principal. This is in the same resource group as azure-sdk-artifacts. The replacement of the SAS usage in the eng/common code was pretty straightforward and the language repositories will continue to work without issue. Once this is in, I'll have to go through each language repository to remove usage of azure-sdk-docs-prod-sas and then I'll be able to remove this from the variable group and, ultimately, the keyvault.

The changes were tested via [temporary PR in Java](https://github.com/Azure/azure-sdk-for-java/pull/40214) and this is the [pipeline run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3796273&view=logs&j=19b87903-f8bc-5cbb-22d9-4f6eb3469d8e&t=83618b74-9683-5f47-3a05-e2f3e3ea330a&l=17) for template which published docs

### DIFF
--- a/eng/common/pipelines/templates/steps/publish-blobs.yml
+++ b/eng/common/pipelines/templates/steps/publish-blobs.yml
@@ -1,6 +1,5 @@
 parameters:
   FolderForUpload: ''
-  BlobSASKey: ''
   TargetLanguage: ''
   BlobName: ''
   ScriptPath: ''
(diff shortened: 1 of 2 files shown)


danieljurek (approved):

eng/common/pipelines/templates/steps/publish-blobs.yml
```diff
@@ -17,17 +16,20 @@ steps:
...
+  displayName: 'Copy Docs to Blob Storage'
+  continueOnError: false
   inputs:
-    filePath: ${{ parameters.ScriptPath }}
-    arguments: >
+    azureSubscription: 'Azure SDK Artifacts'
+    ScriptType: 'FilePath'
+    ScriptPath: ${{ parameters.ScriptPath }}
+    ScriptArguments: >
       -AzCopy $(Resolve-Path "$(Build.BinariesDirectory)/azcopy/azcopy_windows_amd64_*/azcopy.exe")[0]
```
weshaggard, line 27: Orthogonal but we should probably also remove this azcopy approach given it is already on all the agent machines.

PR #8296, opened 2024-05-17
[tsp-client] Add no-prompt flag

catalinaperalta:
Support `--no-prompt` flag to skip the output directory confirmation prompt.


iscai-msft (approved):
looks good, only nit i have is if you could add why someone would use `--no-prompt` if this is something we want users to use. if it's more internal, no need to add any docs

PR #8311, opened 2024-05-21
Remove SAS token usage from apiview-sync-staging

JimSuplizio:
Replace the SAS token usage.

Because this pipeline is a standalone pipeline once these changes are tested and merged the SAS tokens (apiview-production-storage-sas and apiview-staging-storage-sas) can be removed from the [APIView daily Prod to staging sync](https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-distributed-task.hub-library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=86&path=APIView%20sync%20Staging%20variable) and from the keyvault. 

Also, both storage accounts will need to have their access by SAS turned off.

### DIFF
--- a/src/dotnet/APIView/apiview-sync-staging.yml
+++ b/src/dotnet/APIView/apiview-sync-staging.yml
@@ -26,11 +26,11 @@ stages:
         - ${{ each c in parameters.containers }}:
           - task: AzurePowerShell@5
             displayName: Copy Blobs in ${{c}}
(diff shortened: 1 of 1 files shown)


praveenkuttappan (approved):
Thanks for doing this change @JimSuplizio .

chidozieononiwu (approved):

PR #8314, opened 2024-05-21
Remove SAS token usage from apiview-review-gen

JimSuplizio:
SAS token removal. 

The [APIView-Sandboxing](https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-distributed-task.hub-library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=124&path=APIView-Sandboxing) variable group only has two SAS tokens in it, apiview-originals-sas and apiview-staging-originals-sas. The apiview-staging-originals-sas was setup, but ultimately never used, and the only use of apiview-originals-sas is being removed in this PR. In theory, the variable group can be removed once we know which pipelines it's linked to.

The above being said, APIView-Sandboxing isn't the only VG with apiview-originals-sas in it.

### DIFF
--- a/eng/pipelines/templates/steps/apiview-review-gen.yml
+++ b/eng/pipelines/templates/steps/apiview-review-gen.yml
@@ -18,7 +18,6 @@ parameters:
 steps:
 - task: AzurePowerShell@5
   displayName: 'Generate APIView Token files'
-  continueOnError: false
   inputs:
     azureSubscription: 'Azure SDK Engineering System'
     ScriptType: 'FilePath'
--- a/eng/pipelines/templates/steps/apiview-review-gen.yml
+++ b/eng/pipelines/templates/steps/apiview-review-gen.yml
@@ -1,7 +1,7 @@
 parameters:
   - name: Reviews
     type: string
-    default: ''  
+    default: ''
   - name: APIViewURL
     type: string
     default: ''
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/steps/apiview-review-gen.yml
```diff
@@ -16,19 +16,24 @@ parameters:
     default: ''
 
 steps:
-- task: Powershell@2
+- task: AzurePowerShell@5
   displayName: 'Generate APIView Token files'
+  continueOnError: false
```
weshaggard, line 21: This shouldn't be needed. Setting this to false is the default

weshaggard (approved):

praveenkuttappan:
Verified this change using pipeline run https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3813927&view=logs&j=d4367c48-f7ef-5b11-a3cc-789c1d007bb1&t=111e49f7-34c2-56f1-14a3-f4495b4ab402

praveenkuttappan (approved):

eng/pipelines/templates/steps/apiview-review-gen.yml
JimSuplizio: Removed in second commit.

Issue #8315, opened 2024-05-21
APIView - Remove checking PR owner as internal user


praveenkuttappan:
This PR is to remove the dependency on OSS API. API review will be created for PR from all users.

PR #8316, opened 2024-05-21
Remove checking PR author using OSS API

praveenkuttappan:
This PR is to remove verification of PR author as internal user.

Fixes https://github.com/Azure/azure-sdk-tools/issues/8315


weshaggard (approved):
Looks good. Just to call out we are now relying on sandboxing to isolate this processing. 

Let's be sure to remove the related secrets from the configuration once this is deployed.

chidozieononiwu (approved):

PR #8325, opened 2024-05-22
Resolve Digest Round Trip Issues

scbedd:
PR that will eventually resolve Azure/azure-sdk-for-go#22919 

This definitely checks out locally, but also requires a re-record. I'm not actually certain how it's possible that this ever worked. I'm going to have to go back in time somehow 😂


benbp (approved):

scbedd:
I've gotten a feel for how far I will need to go to merge this. Given that I will be  out tomorrow, and this IS a breaking change, I will be merging it Monday 6/10 then working through the few tests that will be affected.

scbedd:
Merging this to begin handling the breaking change issues.

PR #8340, opened 2024-05-28
Changes to scripts to deal with PATs and AccessTokens

JimSuplizio:
Update the scripts that deal with DevOps runs to be able to handle a PAT or an access token from azure.

### DIFF
--- a/eng/common/scripts/Add-RetentionLease.ps1
+++ b/eng/common/scripts/Add-RetentionLease.ps1
@@ -29,10 +29,7 @@ Set-StrictMode -Version 3
 
 . (Join-Path $PSScriptRoot common.ps1)
 
-$Base64EncodedToken=$null
-if (![string]::IsNullOrWhiteSpace($AccessToken)) {
-  $Base64EncodedToken = Get-Base64EncodedToken $AccessToken
-}
+$Base64EncodedToken = Get-Base64EncodedToken $AccessToken
 
 LogDebug "Checking for existing leases on run: $RunId"
 $existingLeases = Get-RetentionLeases -Organization $Organization -Project $Project -DefinitionId $DefinitionId -RunId $RunId -OwnerId $OwnerId -Base64EncodedToken $Base64EncodedToken
--- a/eng/common/scripts/Add-RetentionLease.ps1
+++ b/eng/common/scripts/Add-RetentionLease.ps1
@@ -18,34 +18,34 @@ param(
   [Parameter(Mandatory = $false)]
   [string]$OwnerId = "azure-sdk-pipeline-automation",
 
-  [Parameter(Mandatory = $false)]
-  [string]$AccessToken = $env:DEVOPS_PAT,
-
-  [Parameter(Mandatory = $false)]
-  [string]$AuthToken=$null
+  # This script shouldn't need anything other than the $System.AccessToken from
+  # from the build pipeline. The retain-run.yml template doesn't run outside
+  # of the pipeline it's manipulating the retention leases for.
+  [Parameter(Mandatory = $true)]
+  [string]$AccessToken = $env:DEVOPS_PAT
 )
 
 Set-StrictMode -Version 3
 
 . (Join-Path $PSScriptRoot common.ps1)
 
+$Base64EncodedToken=$null
 if (![string]::IsNullOrWhiteSpace($AccessToken)) {
-  $encodedAuthToken = Get-Base64EncodedToken $AccessToken
+  $Base64EncodedToken = Get-Base64EncodedToken $AccessToken
 }
 
 LogDebug "Checking for existing leases on run: $RunId"
-$existingLeases = Get-RetentionLeases -Organization $Organization -Project $Project -DefinitionId $DefinitionId -RunId $RunId -OwnerId $OwnerId -Base64EncodedAuthToken $encodedAuthToken
+$existingLeases = Get-RetentionLeases -Organization $Organization -Project $Project -DefinitionId $DefinitionId -RunId $RunId -OwnerId $OwnerId -Base64EncodedToken $Base64EncodedToken
 
 if ($existingLeases.count -ne 0) {
     LogDebug "Found $($existingLeases.count) leases, will delete them first."
 
     foreach ($lease in $existingLeases.value) {
         LogDebug "Deleting lease: $($lease.leaseId)"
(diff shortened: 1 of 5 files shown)


eng/common/scripts/Invoke-DevOpsAPI.ps1
```diff
@@ -16,9 +16,28 @@ function Get-Base64EncodedToken([string]$AuthToken)
...
 }
 
-function Get-DevOpsApiHeaders ($Base64EncodedToken) {
-  $headers = @{
-    Authorization = "Basic $Base64EncodedToken"
+# The Base64EncodedToken would be from a PAT that was passed in and the header requires Basic authorization
+# The AccessToken would be the querying the Azure resource with the following command:
+#   az account get-access-token --resource "499b84ac-1321-427f-aa17-267ca6975798" --query "accessToken" --output tsv
+# The header for an AccessToken requires Bearer authorization
+function Get-DevOpsApiHeaders ($Base64EncodedToken, $AccessToken) {
```
weshaggard, line 23: I know naming is not the greatest but can we call these parameters "BasicBase64EncodedToken" and "BearerToken" so it represents the type of headers we are creating from them.

eng/common/scripts/Add-RetentionLease.ps1
```diff
@@ -19,14 +19,19 @@ param(
   [string]$OwnerId = "azure-sdk-pipeline-automation",
 
   [Parameter(Mandatory = $false)]
-  [string]$AccessToken = $env:DEVOPS_PAT
```
weshaggard, line 22: Do we even need to change this script if we don't have any plans to use the bear token for it? I think it is fine to leave this script unchanged.

eng/common/scripts/Invoke-DevOpsAPI.ps1
```diff
@@ -16,9 +16,28 @@ function Get-Base64EncodedToken([string]$AuthToken)
...
+  $headers = $null
+  if (![string]::IsNullOrWhiteSpace($Base64EncodedToken) -and
+      ![string]::IsNullOrWhiteSpace($AccessToken)) {
+        LogError "Get-DevOpsApiHeaders::Unable to set the Authentication in the header because Base64EncodedToken and AccessToken are both set and only one should be."
+        exit 1
+  }
```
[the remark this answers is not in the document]
JimSuplizio: Derp derp. This will be gone with the next commit.

eng/common/scripts/Queue-Pipeline.ps1
```diff
@@ -58,10 +58,13 @@ param(
   [string]$VsoQueuedPipelines,
 
   # Already base 64 encoded authentication token
-  [string]$Base64EncodedAuthToken,
+  [string]$Base64EncodedAuthToken=$null,
```
weshaggard, line 61: I suspect we don't have any more places we use the Base64EncodedAuthToken parameter anymore and if so let's just remove it.

eng/common/scripts/Queue-Pipeline.ps1
```diff
@@ -58,10 +58,13 @@ param(
...
 
-  # Unencoded authentication token
-  [string]$AuthToken,
+  # Unencoded authentication token from a PAT
+  [string]$AuthToken=$null,
+
+  # Temp access token from the logged in az cli user for azure devops resource
+  [string]$AccessToken=$null,
```
weshaggard, line 67: Can we similarly call this BearerToken to match what it represents in the auth header.

eng/common/scripts/Invoke-DevOpsAPI.ps1
JimSuplizio: Sure. It'll be updated with the next commit.

eng/common/scripts/Queue-Pipeline.ps1
JimSuplizio: > I suspect we don't have any more places we use the Base64EncodedAuthToken parameter anymore and if so let's just remove it.
We cannot remove this at part of this PR, the script needs to handle both for a little while longer. I'd planned on updating [sync-directory.yml](sync-directory.yml) separately, after this PR has been merged and the [docindex pipelines will still be using the PAT until it can be replaced with the service principal](https://github.com/search?q=owner%3AAzure+azuresdk-apidrop-devops-queue-build-pat&type=code). The PATs need to be fully gone before this can be removed.

eng/common/scripts/Invoke-DevOpsAPI.ps1
JimSuplizio: Done. Renames were performed all over.

eng/common/scripts/Add-RetentionLease.ps1
JimSuplizio: I removed the Base64EncodedAuthToken script parameter which was never used. I also removed the BearerToken parameter considering that this script is only ever going to get called using the System.AccessToken for the pipeline the retain-run.yml file is used in.

eng/common/scripts/Queue-Pipeline.ps1
JimSuplizio: Removed with commit 7.

eng/common/scripts/Queue-Pipeline.ps1
JimSuplizio: Renamed with commit 7

eng/common/scripts/Add-RetentionLease.ps1
```diff
@@ -18,29 +18,34 @@ param(
...
   [string]$AccessToken = $env:DEVOPS_PAT
 )
 
 Set-StrictMode -Version 3
 
 . (Join-Path $PSScriptRoot common.ps1)
 
-$encodedAuthToken = Get-Base64EncodedToken $AccessToken
+$Base64EncodedToken=$null
+if (![string]::IsNullOrWhiteSpace($AccessToken)) {
```
weshaggard, line 33: If this variable is required this check shouldn't be needed but doesn't hurt to have it still if you like.

weshaggard (approved):

eng/common/scripts/Add-RetentionLease.ps1
JimSuplizio: *sigh* Gah, I'll remove the if check.

PR #8345, opened 2024-05-28
APIView - Connect to cosmos and storage using managed identity

praveenkuttappan:
Changes to connect to Cosmos and Storage using managed identity instead of using connection strings.

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/MiddleWare/UITestsMiddleWare.cs
+++ b/src/dotnet/APIView/APIViewWeb/MiddleWare/UITestsMiddleWare.cs
@@ -5,6 +5,8 @@ using Azure.Storage.Blobs.Models;
 using Azure.Storage.Blobs;
 using Microsoft.Azure.Cosmos;
 using Microsoft.Extensions.Configuration;
+using Azure.Identity;
+using System;
 
 namespace APIViewWeb.MiddleWare
 {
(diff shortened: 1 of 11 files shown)


benbp (approved):

src/dotnet/APIView/APIViewWeb/MiddleWare/UITestsMiddleWare.cs
```diff
@@ -18,19 +20,20 @@ public UITestsMiddleWare(RequestDelegate next)
 
         public async Task InvokeAsync(HttpContext httpContext, IConfiguration config)
         {
-            var cosmosClient = new CosmosClient(config["Cosmos:ConnectionString"]);
-            var dataBaseResponse = await cosmosClient.CreateDatabaseIfNotExistsAsync("APIView");
+            var cosmosClient = new CosmosClient(config["CosmosEndpoint"], new DefaultAzureCredential());
```
weshaggard, line 23: Do you really want to use DefaultAzureCredential here? Do we expect to ever use anything other than MI? What is the auth story for local testing?

weshaggard (approved):
Question about DAC but otherwise looks reasonable.

src/dotnet/APIView/APIViewWeb/MiddleWare/UITestsMiddleWare.cs
praveenkuttappan: Let me check. I think this is not required to use dummy test instance.

chidozieononiwu (approved):

src/dotnet/APIView/APIViewWeb/MiddleWare/UITestsMiddleWare.cs
praveenkuttappan: I will submit a separate PR to remove this Will file an issue to track it.

PR #8350, opened 2024-05-29
Change scope used for OSS portal client

hallipr:
`api://66b6ea26-954d-4b68-8f48-71e3faec7ad1` isn't a supported audience
`66b6ea26-954d-4b68-8f48-71e3faec7ad1` is supported

### DIFF
--- a/tools/identity-resolution/Helpers/GitHubToAADConverter.cs
+++ b/tools/identity-resolution/Helpers/GitHubToAADConverter.cs
@@ -77,7 +77,7 @@ namespace Azure.Sdk.Tools.NotificationConfiguration.Helpers
             try
             {
                 // This is aad scope of opensource rest API.
-                string[] scopes = new [] { "api://66b6ea26-954d-4b68-8f48-71e3faec7ad1/.default" };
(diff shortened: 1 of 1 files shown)


tools/identity-resolution/Helpers/GitHubToAADConverter.cs
```diff
@@ -77,7 +77,7 @@ private async Task<UserLink[]> GetPeopleLinksAsync()
             try
             {
                 // This is aad scope of opensource rest API.
-                string[] scopes = new [] { "api://66b6ea26-954d-4b68-8f48-71e3faec7ad1/.default" };
+                string[] scopes = new [] { "66b6ea26-954d-4b68-8f48-71e3faec7ad1/.default" };
```
weshaggard, line 80: When testing I actually didn't get the token to work with using the scope I had to do a resource id of "api://66b6ea26-954d-4b68-8f48-71e3faec7ad1"

weshaggard (approved):

tools/identity-resolution/Helpers/GitHubToAADConverter.cs
hallipr: I verified with local testing using SP login to az cli

PR #8369, opened 2024-06-03
Remove PATVAR from PipelineGenerator

JimSuplizio:
This is part of the PAT removal effort. The PAT being removed is azuresdk-azure-sdk-devops-pipeline-generation-pat. This PR is the first of two, the second PR will update the yml files remove the usage of azuresdk-azure-sdk-devops-pipeline-generation-pat. The second PR requires this to be published first so the version of pipeline-generator in install-pipeline-generator.yml can be updated.

There was a hacked up version of pipeline-generation.yml that was used to test these changes, this was the [successful run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3845151&view=results) after updates. The hacked up yml file has been removed from this PR.

The second PR will use an `AzureCLI@2` task to authenticate and call pipeline-generation within the task's inlinescript.

### DIFF
--- a/tools/pipeline-generator/Azure.Sdk.Tools.PipelineGenerator/PipelineGenerationContext.cs
+++ b/tools/pipeline-generator/Azure.Sdk.Tools.PipelineGenerator/PipelineGenerationContext.cs
@@ -80,6 +80,7 @@ namespace PipelineGenerator
                 );
                 var devopsCredential = new VssAzureIdentityCredential(azureCredential);
                 cachedConnection = new VssConnection(new Uri(organization), devopsCredential);
+                await cachedConnection.ConnectAsync();
             }
 
             return cachedConnection;
--- a/tools/pipeline-generator/Azure.Sdk.Tools.PipelineGenerator/PipelineGenerationContext.cs
+++ b/tools/pipeline-generator/Azure.Sdk.Tools.PipelineGenerator/PipelineGenerationContext.cs
@@ -73,11 +73,15 @@ namespace PipelineGenerator
         {
             if (cachedConnection == null)
             {
+                // JRS-REMOVE NEXT LINE
+                Console.WriteLine("*****Start Authentication*****");
                 VssCredentials credentials;
                 var azureTokenProvider = new AzureServiceTokenProvider();
                 var authenticationResult = await azureTokenProvider.GetAuthenticationResultAsync("499b84ac-1321-427f-aa17-267ca6975798");
                 credentials = new VssAadCredential(new VssAadToken(authenticationResult.TokenType, authenticationResult.AccessToken));
                 cachedConnection = new VssConnection(new Uri(organization), credentials);
+                // JRS-REMOVE NEXT LINE
+                Console.WriteLine("*****Finish Authentication*****");
             }
 
             return cachedConnection;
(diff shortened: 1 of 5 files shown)


tools/pipeline-generator/Azure.Sdk.Tools.PipelineGenerator/PipelineGenerationContext.cs
```diff
@@ -77,18 +74,9 @@ private async Task<VssConnection> GetConnectionAsync()
...
-                    var authenticationResult = await azureTokenProvider.GetAuthenticationResultAsync("499b84ac-1321-427f-aa17-267ca6975798");
-                    credentials = new VssAadCredential(new VssAadToken(authenticationResult.TokenType, authenticationResult.AccessToken));
-                }
-                else
-                {
-                    var pat = Environment.GetEnvironmentVariable(patvar);
-                    credentials = new VssBasicCredential("nobody", pat);
-                }
-
+                var azureTokenProvider = new AzureServiceTokenProvider();
```
weshaggard, line 77: @JimSuplizio are we sure this will work in the pipeline from an AzureCLI login? I'm only curious because I don't see any mention of that here.

weshaggard: @hallipr do you know if this works and if so why?

hallipr: MSAL supports fallback to AZ CLI token if it can't get a managed identity token.
https://learn.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication?view=azure-dotnet

In most other places, we use Azure.Identity rather than MSAL to get tokens.  The code is basically:
```cs
var azureCredential = new ChainedTokenCredential(
    new AzureCliCredential(),
    new AzurePowerShellCredential()
);

var devopsCredential = new VssAzureIdentityCredential(azureCredential);
var connection = new VssConnection(uri, devopsCredential);
```

This allows us to customize the credential chain and will support new credential types like [AzurePipelinesCredential](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/Credentials/AzurePipelinesCredential.cs).

The `VssAzureIdentityCredential` class is available in the NuGet package `Microsoft.VisualStudio.Services.InteractiveClient` version `19.232.0-preview`

JimSuplizio: As per @hallipr's guidance we're using the ChainedTokenCredential as per above. Also, as per suggestion, in the GetConnectionAsync() method, I've added a call to [await cachedConnection.ConnectAsync();](https://github.com/Azure/azure-sdk-tools/pull/8369/files#diff-36943188fa46e0a82c59c2e0e6390027bca618fec331da7363578706a54bc143R83) which should cause any connection failures to happen at that point instead of randomly when some other part of the code tries to use the connection.

hallipr (approved):

JimSuplizio:
I need to override check-enforcer as there's nothing that runs when this tool is changed, only when it's checked in.

benbp (approved):

PR #8370, opened 2024-06-03
PAT Removal from pipeline generation yml

JimSuplizio:
This is the second PR, the one that removes the usage of azuresdk-azure-sdk-devops-pipeline-generation-pat. 

Note: As of right now, this PR is incomplete. This PR relies on another [PR](https://github.com/Azure/azure-sdk-tools/pull/8369) being merged, publishing a new version of the pipeline-generator and the version of [install-pipeline-generation.yml](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/steps/install-pipeline-generation.yml) being updated to the published version. That needs to be part of this PR and will be updated once the other one is removed.

### DIFF
--- a/eng/common/pipelines/templates/jobs/prepare-pipelines.yml
+++ b/eng/common/pipelines/templates/jobs/prepare-pipelines.yml
@@ -35,166 +35,181 @@ jobs:
     - template: /eng/common/pipelines/templates/steps/set-default-branch.yml
     # This covers our public repos.
     - ${{ if not(endsWith(parameters.Repository, '-pr'))}}:
-      - script: >
-          $(Pipeline.Workspace)/pipeline-generator/pipeline-generator generate
-          --organization azure-sdk
-          --project public
-          --prefix ${{parameters.Prefix}}
-          --devopspath "\${{parameters.Prefix}}"
-          --path $(System.DefaultWorkingDirectory)/sdk
-          --endpoint Azure
-          --repository ${{parameters.Repository}}
-          --convention ci
-          --agentpool Hosted
-          --branch refs/heads/$(DefaultBranch)
-          --patvar PATVAR
-          --set-managed-variables
-          --debug
-          ${{parameters.CIConventionOptions}}
+      - task: AzureCLI@2
+        inputs:
+          azureSubscription: 'opensource-api-connection'
+          scriptType: pscore
+          scriptLocation: inlineScript
+          inlineScript:
+            $(Pipeline.Workspace)/pipeline-generator/pipeline-generator generate `
+            --organization azure-sdk `
+            --project public `
+            --prefix ${{parameters.Prefix}} `
+            --devopspath "\${{parameters.Prefix}}" `
+            --path $(System.DefaultWorkingDirectory)/sdk `
+            --endpoint Azure `
+            --repository ${{parameters.Repository}} `
+            --convention ci `
+            --agentpool Hosted `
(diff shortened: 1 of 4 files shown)


eng/pipelines/pipeline-generation.yml
```diff
@@ -116,114 +116,131 @@ jobs:
...
-  #     --path $(Pipeline.Workspace)/$(RepositoryName)/sdk
-  #     --endpoint Azure
-  #     --repository Azure/$(RepositoryName)-pr
-  #     --convention ci
-  #     --agentpool Hosted
-  #     --branch refs/heads/ $(DefaultBranch)
-  #     --set-managed-variables
-  #     --patvar PATVAR
-  #     --debug
+  # - task: AzureCLI@2
```
JimSuplizio, line 227: @weshaggard is this commented out task even needed anymore, or can I push another commit that completely removes it?

weshaggard: It is probably safe to remove.

eng/common/pipelines/templates/jobs/prepare-pipelines.yml
```diff
@@ -35,166 +35,181 @@ jobs:
...
-          --convention ci
-          --agentpool Hosted
-          --branch refs/heads/$(DefaultBranch)
-          --patvar PATVAR
-          --set-managed-variables
-          --debug
-          ${{parameters.CIConventionOptions}}
+      - task: AzureCLI@2
+        inputs:
+          azureSubscription: 'opensource-api-connection'
```
weshaggard, line 40: Do we know if the identity from this connection has the right level of permissions to create new pipelines? I suspect it doesn't currently so we will need to decide if we want to give it that permission or create a new connection with enough permission for it. I think I'd lean to creating a new one for these pipeline creation pipelines so we can keep it locked down to just those.

JimSuplizio: As per this morning, we know it does.

eng/common/pipelines/templates/jobs/prepare-pipelines.yml
```diff
@@ -35,166 +35,181 @@ jobs:
     - template: /eng/common/pipelines/templates/steps/set-default-branch.yml
```
weshaggard, line 35: As part of this change syncing you will need to grant permission to the new service connection to each of these. You can do that by opening the service request and adding them one at a time there.

weshaggard:
You will need to update also update https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/steps/install-pipeline-generation.yml#L12

eng/pipelines/pipeline-generation.yml
JimSuplizio: It was removed with the second commit.

JimSuplizio:
> You will need to update also update https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/steps/install-pipeline-generation.yml#L12

Yes, I called that out in the description. The other PR, with the source changes, had to get merged and published first. The 3rd commit updates the version and that file is part of this PR.

eng/common/pipelines/templates/jobs/prepare-pipelines.yml
JimSuplizio: They're all added there.

weshaggard (approved):

PR #8371, opened 2024-06-03
Restrict live test storage account access to client IP

benbp:
Not quite sure about the ordering here, if we want to set networks rules before or after post-script/deployment removal or not.

This adds a step to our deployment script that sets any deployed storage accounts to a network deny state, but punches a hole through for the client's IP. A medium term solution until better ones come online (reach out for details).

### DIFF
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -262,7 +262,7 @@ function MergeHashes([hashtable] $source, [psvariable] $dest)
 function BuildBicepFile([System.IO.FileSystemInfo] $file)
 {
     if (!(Get-Command bicep -ErrorAction Ignore)) {
-        Write-Error "A bicep file was found at '$($file.FullName)' but the Azure Bicep CLI is not installed. See https://aka.ms/install-bicep-pwsh"
+        Write-Error "A bicep file was found at '$($file.FullName)' but the Azure Bicep CLI is not installed. See aka.ms/bicep-install"
         throw
     }
 
@@ -758,6 +758,7 @@ try {
     if ($TestApplicationSecret -and $ServicePrincipalAuth) {
         $templateParameters.Add('testApplicationSecret', $TestApplicationSecret)
     }
+    # Only add subnets when running in an azure pipeline context
     if ($CI -and $Environment -eq 'AzureCloud') {
         $templateParameters.Add('azsdkPipelineSubnetList', $azsdkPipelineSubnets)
     }
@@ -838,6 +839,28 @@ try {
                                                                 -templateFile $templateFile `
                                                                 -environmentVariables $EnvironmentVariables
 
+        $storageAccounts = Retry { Get-AzResource -ResourceGroupName $ResourceGroupName -ResourceType "Microsoft.Storage/storageAccounts" }
+        # Add client IP to storage account when running as local user. Pipeline's have their own vnet with access
+        if ($storageAccounts) {
+            foreach ($account in $storageAccounts) {
+                $rules = Get-AzStorageAccountNetworkRuleSet -ResourceGroupName $ResourceGroupName -AccountName $account.Name
+                if ($rules -and $rules.DefaultAction -eq "Allow") {
+                    Write-Host "Restricting network rules in storage account '$($account.Name)' to deny access by default"
+                    Retry { Update-AzStorageAccountNetworkRuleSet -ResourceGroupName $ResourceGroupName -Name $account.Name -DefaultAction Deny }
+                    if ($CI) {
+                        Write-Host "Enabling access to '$($account.Name)' from pipeline subnets"
+                        foreach ($subnet in $azsdkPipelineSubnets) {
+                            Retry { Add-AzStorageAccountNetworkRule -ResourceGroupName $ResourceGroupName -Name $account.Name -VirtualNetworkResourceId $subnet }
+                        }
+                    } else {
+                        Write-Host "Enabling access to '$($account.Name)' from client IP"
+                        $clientIp ??= Retry { Invoke-RestMethod -Uri 'https://icanhazip.com/' }  # cloudflare owned ip site
+                        Retry { Add-AzStorageAccountNetworkRule -ResourceGroupName $ResourceGroupName -Name $account.Name -IPAddressOrRange $clientIp | Out-Null }
+                    }
+                }
+            }
+        }
+
         $postDeploymentScript = $templateFile.originalFilePath | Split-Path | Join-Path -ChildPath "$ResourceType-resources-post.ps1"
         if (Test-Path $postDeploymentScript) {
             Log "Invoking post-deployment script '$postDeploymentScript'"
@@ -852,7 +875,6 @@ try {
         Write-Host "Deleting ARM deployment as it may contain secrets. Deployed resources will not be affected."
         $null = $deployment | Remove-AzResourceGroupDeployment
     }
-
 } finally {
     $exitActions.Invoke()
 }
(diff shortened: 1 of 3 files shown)


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -827,6 +827,19 @@ try {
                                                                 -templateFile $templateFile `
                                                                 -environmentVariables $EnvironmentVariables
 
+        $storageAccounts = Retry { Get-AzResource -ResourceGroupName $ResourceGroupName -ResourceType "Microsoft.Storage/storageAccounts" }
+        if ($storageAccounts) {
+            $clientIp = Retry { Invoke-RestMethod -Uri 'https://myip.dnsomatic.com/' }  # OpenDNS owned ip site. Some throttling will occur if called quickly.
```
mikeharder, line 832: This might work in most cases, but I don't think it's **guaranteed** to work, since your network might route your request over different public IPs depending on the destination.

benbp: In that case I'm not sure we have a great workaround, other than to enable storage access logs, wait for them to appear, download the audit history, and parse the client IP out of it :)

benbp: I'm also not happy taking a third party site dependency like this, as most sites I found seem to be hosted by unknown/random entities. I'd like to have a backup site to call if the main one doesn't work, but I couldn't find anything else I trusted as much as the opendns org. Would be nice if Microsoft had a site like this somewhere.

benbp: Another option is we could run a simple app service to do this for ourselves with opendns as a backup.

weshaggard:
@benbp is there a way to potential do this inside of a bicep template we can force folks to start using if they are deploying storage accounts?

benbp:
@weshaggard we could add some sort of dynamic parameter to pass in for the IP, but that won't help with new bicep files. Also we're going to go with a vnet solution for the agent pools, so we don't want client ip to be used in every case either.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -758,7 +758,8 @@ try {
     if ($TestApplicationSecret -and $ServicePrincipalAuth) {
         $templateParameters.Add('testApplicationSecret', $TestApplicationSecret)
     }
-    if ($CI -and $Environment -eq 'AzureCloud') {
+    # Only add subnets when running in an azure pipeline context
+    if ($env:SYSTEM_TEAMPROJECTID -and $Environment -eq 'AzureCloud') {
```
weshaggard, line 762: Isn't that what the CI parameter is about?

benbp: The challenge with using CI is some other things are coupled to it (parameter sets, other conditionals). Once in a while I need to actually test the CI parameter locally (e.g. if I'm making CI based behavior changes). So I prefer to be explicit here about the conditional, which allows me to still run CI mode in any environment, and restrict behavior that only works in the pipeline to that environment.

benbp: I can change it to CI and just modify conditionals if needed when testing. I agree there are trade-offs either way.

weshaggard (approved):

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -106,14 +119,27 @@ param (
...
-    ($azsdkPipelineVnet + "/subnets/pipeline-subnet-ubuntu-1804-general"),
-    ($azsdkPipelineVnet + "/subnets/pipeline-subnet-ubuntu-2004-general"),
-    ($azsdkPipelineVnet + "/subnets/pipeline-subnet-ubuntu-2204-general"),
-    ($azsdkPipelineVnet + "/subnets/pipeline-subnet-win-2019-general"),
-    ($azsdkPipelineVnet + "/subnets/pipeline-subnet-win-2022-general")
-)
+$azsdkPipelineVnetWestUS = '/subscriptions/a18897a6-7e44-457d-9260-f2854c0aca42/resourceGroups/azsdk-pools/providers/Microsoft.Network/virtualNetworks/azsdk-pipeline-vnet-wus'
+$azsdkPipelineVnetCanadaCentral = '/subscriptions/a18897a6-7e44-457d-9260-f2854c0aca42/resourceGroups/azsdk-pools/providers/Microsoft.Network/virtualNetworks/azsdk-pipeline-vnet-cnc'
+$azsdkPipelineSubnetMap = @{
+    'azsdk-pool-mms-ubuntu-1804-general' = ($azsdkPipelineVnetWestUS + '/subnets/pipeline-subnet-ubuntu-1804-general')
```
weshaggard, line 125: We should make sure to update the docs around setting up a machine pool to include this subnet setup as well.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -106,14 +119,27 @@ param (
...
+    'azsdk-pool-mms-ubuntu-2004-general' = ($azsdkPipelineVnetWestUS + '/subnets/pipeline-subnet-ubuntu-2004-general')
```
weshaggard, line 126: Explicitly mapping here means any new pool will also need to update here. Should we do it via string replacement instead so that new pools could work without a change if they follow the correct convention?

benbp: The challenge here is that the vnets have to be region paired with the pools. If we wanted to do a convention based naming scheme, I think we would also want to change our pool names to include the region they are in so we can know what vnet to pair them with. And we currently use the 1es change region feature (which changes the VMSS region despite the pool region) to make things more confusing.

Part of me doesn't like the config being stored in multiple places (especially here), and the other part is worried about the general practice of naming magic making things hard to diagnose. Open to ideas here.

Another option is we could make a call to Azure at the time of running this script to get the pool/subnet info. But that involves a few more pools that we actually use here and could add some time to the pipeline.

benbp: We also will need to come up with some scripts to auto-update all our resources. As it is right now, we have many places across our services (apiview, etc.) that have these same subnets added. Whether it is a manual or automated update there is still downtime potential (e.g. we found you have to restart app service after adding a vnet+subnet among other things)

weshaggard: My main concern is that I would prefer we not be forced to update this table anytime we add a new pool. Is there a property we can query on the pool itself? If so, maybe that call is worth having here.

benbp: Looks like we can do it like so, and it runs very quickly:

```
(Get-AzResource -ResourceGroupName azsdk-pools -Name azsdk-pool-mms-ubuntu-2004-general -ExpandProperties).Properties.networkProfile
```

benbp: And our provisioner app should already have read access here, as we need a certain level of permissions to able to link these subnets to other resources.

benbp: Testing out the following:

```
    $ctx = Get-AzContext
    Set-AzContext -Subscription 'Azure SDK Engineering System'
    $poolSubnet = (Get-AzResource -ResourceGroupName azsdk-pools -Name $env:Pool -ExpandProperties).Properties.networkProfile.subnetId
    $ctx | Set-AzContext
```

weshaggard: Lets do that then.

eng/common/TestResources/build-test-resource-config.yml
```diff
@@ -14,6 +14,24 @@ parameters:
     default: null
 
 steps:
+  - task: AzurePowerShell@5
```
weshaggard, line 17: Should we make this conditional if folks want to opt-out? Or is there a place where someone can override PoolSubnet?

eng/common/TestResources/build-test-resource-config.yml
```diff
@@ -14,6 +14,24 @@ parameters:
...
+    displayName: Set Pipeline Subnet Info
+    env: ${{ parameters.EnvVars }}
+    inputs:
+      azureSubscription: Azure SDK Engineering System
```
weshaggard, line 21: Weren't you planning to use the azure-sdk-tests identity for this?

benbp: Ah yes, thanks for the reminder, reverted and testing.

benbp:
Waiting on the java/net PRs from @danieljurek to update the federated auth conditional before this goes in (since it relies on two other PRs that are blocked on the aforementioned PRs)

PR #8385, opened 2024-06-05
Add `-UserAuth` to basic invocation example of `New-TestResources`

scbedd:
@benbp 

Just based off of a couple questions I've gotten recently. Should this be a more holistic update?


benbp:
@scbedd The `-UserAuth` parameter no longer exists as it is the default behavior. The legacy behavior can be invoked via `-ServicePrincipalAuth`.

scbedd:
Totally unncessary

PR #8386, opened 2024-06-06
Light up System.ClientModel interface decorations on public API surface

### LINKED ISSUE #8326: Include interfaces from System.ClientModel in APIView
Interfaces from System.ClientModel don't appear in .NET APIViews today. This can be problematic because architects reviewing APIs may not see that a type implements an interface such as `IJsonModel<T>` in an APIView, and may not be able to review the correctness of this for a library's public API.  For example, Azure.Core shipped a [RehydrationToken](https://apiview.dev/Assemblies/Review/ba87b735158144eea6cabe21a2c58dde#Azure.Core.RehydrationToken) type that implements `IJsonModel<T>`, which I am not sure whether architects were aware of.

Related to https://github.com/Azure/azure-sdk-tools/issues/6844

christothes:
fixes #8326
This change adds allowed .nupkg dependencies to the compilation as references so that the compilation has the full metadata for interfaces that live in those dependencies. Prior to this change, interfaces in referenced assemblies such as System.ClientModel would never show up as public APIs for types that implement them.

Azure.Core.RehydrationToken before:
image

after: (results from local testing):
image

### DIFF
--- a/src/dotnet/APIView/APIView/Languages/CompilationFactory.cs
+++ b/src/dotnet/APIView/APIView/Languages/CompilationFactory.cs
@@ -13,7 +13,9 @@ namespace ApiView
     {
         private static HashSet<string> AllowedAssemblies = new HashSet<string>(new[]
         {
-            "Microsoft.Bcl.AsyncInterfaces"
+            "Microsoft.Bcl.AsyncInterfaces",
+            "System.ClientModel"
+
         }, StringComparer.InvariantCultureIgnoreCase);
 
         public static IAssemblySymbol GetCompilation(string file)
@@ -24,7 +26,7 @@ namespace ApiView
             }
         }
 
-        public static IAssemblySymbol GetCompilation(Stream stream, Stream documentationStream)
+        public static IAssemblySymbol GetCompilation(Stream stream, Stream documentationStream, IEnumerable<string> dependencyPaths = null)
         {
             PortableExecutableReference reference;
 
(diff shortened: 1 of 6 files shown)


annelo-msft (approved):
This looks great!  (From the "after" screenshot)

Will it work for all .NET packages?  i.e. will `IJsonModel<T>` and `IPersistableModel<T>` start showing up for clients as well -- not just Azure.Core?

Many thanks!

christothes:
> Will it work for all .NET packages? i.e. will `IJsonModel<T>` and `IPersistableModel<T>` start showing up for clients as well -- not just Azure.Core?


Good question - I will need to test this since the current behavior change will interrogate the dependencies for the primary nupkg being reviewed. For clients, I think they would only depend on Azure.Core, which then depends on System.ClientModel. So I may need to recursively follow dependencies for it to work with other clients.

If you could point me to an existing client that implements a type that would be decorated with a System.ClientModel interface, I can validate.

annelo-msft:
> If you could point me to an existing client that implements a type that would be decorated with a System.ClientModel interface, I can validate.

Looks like [Azure.AI.OpenAI](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/openai/Azure.AI.OpenAI/api/Azure.AI.OpenAI.netstandard2.0.cs#L10) and [Azure.Security.CodeTransparency](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/confidentialledger/Azure.Security.CodeTransparency/api/Azure.Security.CodeTransparency.netstandard2.0.cs#L69) both do.

christothes:
Looks like clients take a direct dependency on `System.ClientModel` so we should be good!

m-nash (commented):
The imagine showing the apiview has an expansion of the interfaces which I am not sure we want.

For instance it shows `IJsonModel<T>` and `IPersitableModel<T>`.  The actual class only specifies `IJsonModel<T>` https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/RehydrationToken.Serialization.cs#L11.

I do see that we do this with `IEnumerable<T>` expanding it to `IEnumerable`, but just wondering if we really want this as it the very first thought I had was why are we explicitly specifying `IPersistableModel<T>` but I had to dig into the code to find out we don't.

chidozieononiwu (approved):

Issue #8407, opened 2024-06-10
Release planner - pipeline to create package work item from release planner is failing


praveenkuttappan:
Package work item is automatically created when CI pipeline runs. But this pipeline is still used to sync package display name if user updates it.

PR #8410, opened 2024-06-10
Update language param name to avoid collision with common eng variable

praveenkuttappan:
Create package work item uses param `$Language` and this collides with `$Language` in eng/common.ps1. This works in most cases but .NET repo his this value set to `dotnet` instead of `.NET` so create package work item request fails with invalid language value error.


praveenkuttappan:
Fixes #8407

maririos (approved):

PR #8423, opened 2024-06-11
Show Line Numbers/Show Left Nav Panel

WhitShake:
Added logic to show/hide line numbers and the left nav panel by toggling the corresponding switches in the side bar


jsquire (changes requested):
There's a lot in this PR that shouldn't be - likely from the big set of merges.  Can we please revert to your latest commit and rebase on `main` to ensure that only the changes that you're making are included here?

PR #8424, opened 2024-06-11
Add support for missing build processing

hallipr:
This creates the `PeriodicLockingBackgroundService` class to encapsulate the logic of periodically running a process in a single worker instance only if the worker is able to establish a lock on a central cosmos document.  It continually renews the lock while the processing is handled by a  ProcessAsync implementation. 

This change also starts paving the way for GitHub Actions logging by moving the Azure Pipelines related classes into 
 their own namespace.

### DIFF
--- a/tools/pipeline-witness/Azure.Sdk.Tools.PipelineWitness/AzurePipelines/AzurePipelinesProcessor.cs
+++ b/tools/pipeline-witness/Azure.Sdk.Tools.PipelineWitness/AzurePipelines/AzurePipelinesProcessor.cs
@@ -183,6 +183,30 @@ namespace Azure.Sdk.Tools.PipelineWitness.AzurePipelines
             await UploadBuildBlobAsync(account, build);
         }
 
+        public async Task<string[]> GetBuildBlobNamesAsync(string projectName, DateTimeOffset minTime, DateTimeOffset maxTime, CancellationToken cancellationToken)
+        {
+            DateTimeOffset minDay = minTime.ToUniversalTime().Date;
+            DateTimeOffset maxDay = maxTime.ToUniversalTime().Date;
+
+            DateTimeOffset[] days = Enumerable.Range(0, (int)(maxDay - minDay).TotalDays + 1)
+                .Select(offset => minDay.AddDays(offset))
+                .ToArray();
+
+            List<string> blobNames = [];
+
+            foreach (DateTimeOffset day in days)
+            {
+                string blobPrefix = $"{projectName}/{day:yyyy/MM/dd}/";
+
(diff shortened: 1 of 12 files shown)


tools/pipeline-witness/Azure.Sdk.Tools.PipelineWitness/AzurePipelines/BlobUploadProcessor.cs
```diff
@@ -810,6 +807,12 @@ public static int CalculateBatches(int startingNumber, int batchSize = ApiBatchS
             }
         }
 
+        public async Task<bool> BuildBlobExistsAsync(Build build)
```
weshaggard, line 810: As we talked it might be worth doing a blob list and checking locally instead of querying once per blob.

tools/pipeline-witness/Azure.Sdk.Tools.PipelineWitness/AzurePipelines/MissingBuildWorker.cs
```diff
@@ -0,0 +1,100 @@
...
+                string continuationToken = null;
+                do
+                {
+                    var completedBuilds = await this.buildClient.GetBuildsAsync2(project, minFinishTime: minFinishTime, maxFinishTime: maxFinishTime, statusFilter: BuildStatus.Completed, continuationToken: continuationToken, cancellationToken: cancellationToken);
+
+                    var skipCount = 0;
+                    var enqueueCount = 0;
+                    foreach (var build in completedBuilds)
+                    {
+                        if (await runProcessor.BuildBlobExistsAsync(build))
```
weshaggard, line 69: Lets try to switch to a list up front and then a look-up and this should be a lot more efficient.

tools/pipeline-witness/Azure.Sdk.Tools.PipelineWitness/AzurePipelines/BlobUploadProcessor.cs
hallipr: We now pull a list of known blob names that overlap the period we're processing, then check to see if each blob is in the list

tools/pipeline-witness/Azure.Sdk.Tools.PipelineWitness/AzurePipelines/MissingBuildWorker.cs
hallipr: done

weshaggard (approved):

hallipr:
/check-enforcer override

There is no PR pipeline for this project

PR #8430, opened 2024-06-12
Adjustment to breaking docker manifest change

scbedd:
Follow-up to #8325

We added a test case, but it turns out there is another value that _also_ indicates a manifest. header: `application/vnd.oci.image.index.v1` indicates a body containing multiple manifests. All of this needs to be stored _exactly_ as what comes, so pretty json printing is right out.

~~Continuing to dig via investigation using `go` `sdk/containers/azcontainerregistry/client_test.go::TestClient_UploadManifest` (which is where this new case showed up)~~

The "discovered" issue was that during **matching** of a `request`, we actually _normalize_ the JSON if it's a json content type. That obviously breaks the manifest matching, as the digest only works if the request or response body bytes are _exactly_ the same. This PR extends the "treat this as nontext" to this normalization logic as well.


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/ContentTypeUtilities.cs
```diff
@@ -5,11 +5,21 @@
 
 using System;
 using System.Text;
+using Microsoft.AspNetCore.Razor.Language.Intermediate;
```
scbedd, line 8:
```suggestion
```

benbp (approved):

scbedd:
/check-enforcer override

I don't know how I cancelled that job but I did.

PR #8438, opened 2024-06-13
Fix sort ordering and type alias resolution

jhendrixMSFT:
Sort diagnostics and navigation slices. ~This requires a custom JSON marshaler for Navigation as it contains a map and the default behavior is non-deterministic.~ Per the docs, maps are sorted by key so we're ok here.
Recursively resolve type aliases. This is to handle cases where a type alias is to another type alias.


src/go/cmd/module.go
```diff
@@ -114,94 +114,114 @@ func NewModule(dir string) (*Module, error) {
 	// the definition from azcore/internal/shared into the APIView for azcore, making the type's
 	// fields visible there.
 	externalPackages := map[string]*Pkg{}
+
+	// tracks which packages have had their type aliases resolved.
+	// this prevents resolving dependent packages multiple times.
+	processedPackages := map[string]struct{}{}
+
 	for _, p := range m.packages {
-		for alias, qn := range p.typeAliases {
```
jhendrixMSFT, line 118: This content was moved into `recursiveResolveTypeAliases`. The only changes to it are the recursive call to `recursiveResolveTypeAliases` and checking/updating `processedPackages`.

src/go/cmd/api_view.go
```diff
@@ -83,3 +100,20 @@ func createReview(pkgDir string) (PackageReview, error) {
 		PackageName: m.PackageName,
 	}, nil
 }
+
+func recursiveSortNavigation(n Navigation) {
+	for _, nn := range n.ChildItems {
+		recursiveSortNavigation(nn)
+	}
+	slices.SortFunc(n.ChildItems, func(a Navigation, b Navigation) int {
+		aa, err := json.Marshal(a)
```
chlowell, line 109: `Navigation` contains a map, should we take a similar approach as for `Diagnostic`?

jhendrixMSFT: I don't think we need to. I checked the docs for the JSON marshaler and the maps are sorted by key. So the output for each `Navigation` should be deterministic. Unless I misunderstood the question?

src/go/cmd/api_view_test.go
```diff
@@ -298,3 +299,19 @@ func Test_getPackageNameFromModPath(t *testing.T) {
 	require.EqualValues(t, "sdk/foo/bar", getPackageNameFromModPath("github.com/Azure/azure-sdk-for-go/sdk/foo/bar"))
 	require.EqualValues(t, "sdk/foo/bar", getPackageNameFromModPath("github.com/Azure/azure-sdk-for-go/sdk/foo/bar/v5"))
 }
+
+func TestDeterministicOutput(t *testing.T) {
```
jhendrixMSFT, line 303: I verified that without these changes this test will fail.

src/go/cmd/api_view.go
chlowell: No, I was worried that the marshaler would make these comparisons unstable by iterating the maps. I didn't know it sorts them. I'm surprised to learn it does; naively, that seems like wasted effort.

jhendrixMSFT: I was surprised too which is why I initially wrote the custom marshaler.

Here's the relevant docs taken from https://pkg.go.dev/encoding/json#Marshal

`
Map values encode as JSON objects. The map's key type must either be a string, an integer type, or implement [encoding.TextMarshaler](https://pkg.go.dev/encoding#TextMarshaler). The map keys are sorted and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above:
`

chlowell: Now I have to wonder whether json v2 behaves the same and what magnitude of perf improvement we could gain by not sorting maps 🤔

weshaggard:
@jhendrixMSFT I will sign-off for the codeowners requirement but I suggest adding an entry in the CODEOWNERS file for this tool for future approvals.

weshaggard (approved):

jhendrixMSFT:
@weshaggard I added appropriate code owners but the CI is flagging Charles' alias. Seems weird considering he's a listed code owner in the azure-sdk-for-go repo.

weshaggard:
@chlowell can you make your membership in Azure public?
image

@jhendrixMSFT he is baselined in the go repo currently https://github.com/Azure/azure-sdk-for-go/blob/main/.github/CODEOWNERS_baseline_errors.txt#L1. Folks should be a public Azure member with write access.

chlowell:
👍 have done

PR #8450, opened 2024-06-14
Fix missing owners for path error to include path

JimSuplizio:
The problem right now is that the missing owners error, `There are no owners defined for CODEOWNERS entry.`, is too generic. At the moment, the only missing owners in any of the repositories running the linter are for source path/owner lines. The fix here is to add the path to the error in order to make it distinct.

**There is one big issue here:** 
Monikers can also have missing owners but because we can't put line numbers into the baseline file, for obvious reasons, there's no good differentiation for monikers missing errors.
**The mitigation for the issue:**
The only repositories that have this error in their baseline files are azure-sdk-for-java and azure-sdk-for-js. For Java, this error is actually because of two source path/owner lines lacking owners. For JS, this error was apparently fixed and never removed from the baseline file. When this change goes in, and the linter version is updated, Java's baseline file will need to be updated with the new errors, which will include the path. This means that any Moniker's missing owners will fail the pipeline at the PR level since these don't currently exist in any repository running the linter.

This change only affects the linter. The parser, which is used by a number of tools, remains unchanged. Only the linter version is going to need to be updated when this is merged.

### DIFF
--- a/tools/codeowners-utils/Azure.Sdk.Tools.CodeownersUtils/Constants/ErrorMessageConstants.cs
+++ b/tools/codeowners-utils/Azure.Sdk.Tools.CodeownersUtils/Constants/ErrorMessageConstants.cs
@@ -23,6 +23,7 @@ namespace Azure.Sdk.Tools.CodeownersUtils.Constants
         public const string InvalidTeamPartial = " is an invalid team. Ensure the team exists and has write permissions.";
         public const string InvalidUserPartial = " is an invalid user. Ensure the user exists, is public member of Azure and has write permissions.";
         public const string MalformedTeamEntryPartial = " is a malformed team entry and should start with '@Azure/'.";
+        public const string PathEntryMissingOwners = "Path entry, {0}, is missing owners";
         public const string PathEntryMissingOwnersPartialStart = "Path entry, ";
         public const string PathEntryMissingOwnersPartialEnd = ", is missing owners";
         public const string NoOwnersDefined = "There are no owners defined for CODEOWNERS entry.";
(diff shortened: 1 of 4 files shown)


tools/codeowners-utils/Azure.Sdk.Tools.CodeownersUtils/Constants/ErrorMessageConstants.cs
```diff
@@ -23,6 +23,8 @@ public class ErrorMessageConstants
         public const string InvalidTeamPartial = " is an invalid team. Ensure the team exists and has write permissions.";
         public const string InvalidUserPartial = " is an invalid user. Ensure the user exists, is public member of Azure and has write permissions.";
         public const string MalformedTeamEntryPartial = " is a malformed team entry and should start with '@Azure/'.";
+        public const string PathEntryMissingOwnersPartialStart = "Path entry, ";
```
weshaggard, line 26: Any reason to not just have one string with a place holder? `Path entry, {0}, is missing owners.`

weshaggard (approved):
One suggestion but otherwise seems reasonable.

tools/codeowners-utils/Azure.Sdk.Tools.CodeownersUtils/Constants/ErrorMessageConstants.cs
JimSuplizio: > Any reason to not just have one string with a place holder? `Path entry, {0}, is missing owners.`
I wasn't aware this actually worked. I've updated the const. The only unfortunate thing here is that the TestCase declaration complains about `string.Format(ErrorMessageConstants.PathEntryMissingOwners, line.Trim())`  and requires the constant string. That's minor all things considered.

PR #8452, opened 2024-06-14
Pass pipeline vnet to live test arm templates

benbp:
Some azure resources (sql) cannot be accessed by VMs in a vnet even if they have public access enabled. In these cases we have to set up the vnet rules in the arm template at deploy time (instead of via azure policy).

See for example: https://github.com/Azure/azure-sdk-for-java/pull/40639

### DIFF
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -106,6 +106,15 @@ param (
 
 . $PSScriptRoot/SubConfig-Helpers.ps1
 
+$azsdkPipelineVnet = "/subscriptions/a18897a6-7e44-457d-9260-f2854c0aca42/resourceGroups/azsdk-pools/providers/Microsoft.Network/virtualNetworks/azsdk-pipeline-vnet-wus"
+$azsdkPipelineSubnets = @(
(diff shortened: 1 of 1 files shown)


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -743,13 +752,15 @@ try {
...
         $templateParameters["provisionerApplicationOid"] = "$ProvisionerApplicationOid"
     }
-
     if ($TenantId) {
         $templateParameters.Add('tenantId', $TenantId)
     }
     if ($TestApplicationSecret -and $ServicePrincipalAuth) {
         $templateParameters.Add('testApplicationSecret', $TestApplicationSecret)
     }
+    if ($CI -and $Environment -eq 'AzureCloud') {
```
weshaggard, line 761: Should we do this based on cloud or tenant?

benbp: Cross tenant vnet linking will work, but cross cloud will not, is my thinking (since the vnet is not multi cloud). So we would have to filter on public cloud only. Sovereign clouds aren't a problem for us since they are already in an isolated tenant. For public cloud it doesn't matter which tenant because the agent request will always originate from a vnet.

weshaggard (approved):

mikeharder (approved):

Issue #7941, opened 2024-03-22
1ES Transition Items - "Extras"


scbedd:
## Discovered long tail items

- [ ] Transition azure-sdk-for-android

## Designer Builds

The following builds are designer, and will not benefit from the yml updates above.

- [ ] mgmt-netsdk-sign
- [ ] mgmt-NetCore-SDK-Publish
- [ ] mgmt-netsdk-multiapi-publish
- [ ] mgmt-netsdk-sdkcommon-publish
- [ ] mgmt-netsdk-publish
- [ ] Specs Repo - Update readme.md all-api-versions tags.
- [ ] Autorest Npm Admin Task
- [ ] tools - Codex - vscode.dev
- [ ] We need to come up with a strategy for these.

## Release Builds

Need to come up with a strategy for all the internal release pipelines that do not have yaml attached to them. [Full List](https://dev.azure.com/azure-sdk/internal/_release)

kurtzeborn:
These are lo-pri... they existed before and they continue to exist after the transition.  We should not worry about these until we get dinged on them (and then we'll probably just delete them).

alzimmermsft:
At least in Java we're finding that in some CI runs virus scanning can take 10 to 15 minutes to run. Looking at the step running virus scanning, this could be that we're using `'$(System.DefaultWorkingDirectory)'` as the output parent directory, which may mean we're scanning all source code, build artifacts, and test recordings downloaded and generated during CI.

We should investigate if moving these build artifacts to a subfolder reduces the number of files that need to be virus scanned.

scbedd:
> These designer builds are lo-pri... they existed before and they continue to exist after the transition. We should not worry about these until we get dinged on them (and then we'll probably just delete them).

They were all deleted.

scbedd:
This is all done except for one item that I am blocked on. Publishing the ADO extension from openapi-alps-publish. Have a support ticket out with the [1es folks](https://dev.azure.com/mseng/1ES/_workitems/edit/2191132). Holding pattern until response.

PR #8460, opened 2024-06-17
Convert remaining packages which necessitate 1es-templates/esrp release

scbedd:
The following yml are associated with projects that need to publish to npmjs.org. This PR simply removes each's individual `release` stage in favor of a single added stage. After walking each package within `azure-sdk-tools`, I believe these are the only remaining builds that must become 1es compliant.

  - [x] ~~tools/sdk-generation-pipeline/ci.yml~~ Deleted in `main`
  - [x] tools/oav-traffic-converter/ci.yml
  - [x] tools/mock-service-host/ci.yml
  - [x] tools/js-sdk-release-tools/ci.yml
  - [x] ~~tools/apiview/parsers/js-api-parser/ci.yml~~ Only publishes to internal feed
  - [x] tools/apiview/emitters/typespec-apiview/ci.yml -- this only needs ESRP Release stage, already 1es compliant.
  - [x] ~~eng/pipelines/templates/stages/archetype-autorest-preview.yml~~ This yml is referenced _nowhere_ that I can tell. Not from another template and not direct from a build definition. 
  - [x] tools/tsp-client/ci.yml  -- this only needs ESRP Release stage, already 1es

**FOR REVIEWERS** I advise disabling whitespace diffs it massively simplifies looking at what this PR does. Much more re-use than code deletion.

Related to #7941


scbedd:
/check-enforcer override

The `tsp-client` failure is merely because I cancelled the release itself.

eng/scripts/determine-js-release-tag.ps1
```diff
@@ -0,0 +1,18 @@
+# Read the package.json file
+$packageJsonPath = "$PSScriptRoot/../../package.json"
+$packageJson = Get-Content $packageJsonPath -Raw | ConvertFrom-Json
+
+# Function to check if a version is non-GA
```
weshaggard, line 5: Is it worth using our SemVer.ps1 here? Do we have non-standard prerelease tags?

eng/pipelines/templates/stages/archetype-sdk-publish-js.yml
```diff
@@ -0,0 +1,74 @@
...
+                    - pwsh: |
+                        Write-Host "Will deploy with tag of $(Tag)"
+                        Get-ChildItem "$(Pipeline.Workspace)/drop" -Recurse -Force `
+                          | Where-Object { $_.Name -like "*.tgz" } `
+                          | Copy-Item -Destination "$(Build.ArtifactStagingDirectory)"
+
+                        Get-ChildItem "$(Build.ArtifactStagingDirectory)" -Recurse -Force | % { Write-Host $_.FullName }
+                      displayName: Move artifact to $(Build.ArtifactStagingDirectory)
+
+                    - task: EsrpRelease@7
```
weshaggard, line 58: Thank you for cleaning these up and converting them to ESRP. FYI @ckairen as we will eventually want common publish template here that can be used outside of the tools repo as well.

weshaggard (approved):
Looks reasonable assuming test pipelines work.

eng/scripts/determine-js-release-tag.ps1
scbedd: I'll take a look, thanks for the suggestion 👍

scbedd: Yes, I'll parse the version from package.json, then set `beta` if isPrerelease` is true`.

weshaggard:
Ray recently deleted `tools/sdk-generation-pipeline/ci.yml` so you can remove those changes.

scbedd:
Will do so. Needed to get in here to make the SemVer changes anyway.

scbedd:
My claim is that

1. The existing build failures are already existing. Going to ignore them.
2. For the failed `SDL Sources`, I believe this is only on [the merge PR](https://dev.azure.com/azure-sdk/public/_build/results?buildId=3890568&view=results) causing some weird interaction with the submodules. This doesn't occur when just run straight from the branch. And the branch is updated from `main` as of this morning.

Just going to merge this and then deal with the fallout.

eng/pipelines/templates/stages/archetype-sdk-publish-js.yml
```diff
@@ -0,0 +1,74 @@
...
+                    steps:
+                    - checkout: self
+
+                    - download: current
+                      artifact: ${{ parameters.ArtifactName }}
+                      timeoutInMinutes: 5
+
+                    - task: PowerShell@2
+                      inputs:
+                        filePath: '$(Build.SourcesDirectory)/eng/scripts/determine-release-tag.ps1'
```
catalinaperalta, line 45: This file doesnt match the new one being added in the PR. Seems like the file name should be `determine-js-release-tag.ps1`

PR #8467, opened 2024-06-18
Update publish template

weshaggard:
Add logging and fix display name given Agent.JobStatus doesn't evaluate currently in the display name property.


weshaggard:
image

Fixing that

scbedd (approved):

PR #8471, opened 2024-06-19
Fix  high sev security issue

wanlwanl:
https://github.com/Azure/azure-sdk-tools/security/dependabot/137
https://github.com/Azure/azure-sdk-tools/security/dependabot/175

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "tools/js-sdk-release-tools/deps/azure-js-dev-tools"]
-	path = tools/js-sdk-release-tools/deps/azure-js-dev-tools
+[submodule "tools/js-sdk-release-tools/src/deps/azure-js-dev-tools"]
+	path = tools/js-sdk-release-tools/src/deps/azure-js-dev-tools
 	url = https://github.com/ts-common/azure-js-dev-tools
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tools/js-sdk-release-tools/deps/azure-js-dev-tools"]
+	path = tools/js-sdk-release-tools/deps/azure-js-dev-tools
+	url = https://github.com/ts-common/azure-js-dev-tools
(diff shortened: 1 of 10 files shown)


raych1:
@wanlwanl , I don't see the `lodash` related upgrade. Was it removed?

raych1:
@wanlwanl , you need to join public Azure org to resolve the codeowner CI failures.

wanlwanl:
> @wanlwanl , you need to join public Azure org to resolve the codeowner CI failures.

I already member of public azure org, check it out:https://github.com/orgs/Azure/people?query=wanpeng

raych1 (approved):

qiaozha (approved):
LGTM

weshaggard:
Let's avoid overriding check-enforcer. The ci job should have ran but was disabled because of lack of running. I've re-enabled it so it should trigger on future changes.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tools/js-sdk-release-tools/deps/azure-js-dev-tools"]
```
weshaggard, line 1: Why are we using a submodule? Submodules cause all sorts of issues and we generally avoid them. If you need this dependency then we should publish the package instead of using a submodule.

weshaggard: One example of such a issue https://github.com/Azure/azure-sdk-tools/security/dependabot/178/update-logs/370202889

wanlwanl: Thanks @weshaggard for checking it and kindly share the guidance.

Previous tool owner used a package https://www.npmjs.com/package/@ts-common/azure-js-dev-tools
but looks like no one maintain it any more. The latest package is used, there's no way to easily upgrade the package to resolve the problem. So I add submodule to update the braces package. We will definitely avoid to use packages out of maintainence in the future.

No worries, I'll try shrinkwrap first. If it doesn't help, I'll publish a new package instead of using a submodule.

PR #8478, opened 2024-06-20
Fix AZC0006 and AZC0007 to accept endpoint as an exception

pshao25:
The correct appearance of constructor is https://github.com/Azure/autorest.csharp/issues/3706#issuecomment-1752374728. We should put all the optional parameters into ClientOptions instead of being constructor parameters. However, endpoint is an exception. Even it is an optional parameter, it should still be in the constructor parameter list, not in the ClientOptions.

Currently our analyzer will report false alert for that endpoint exception like below, which is actually correct.
```
public AClient(string required): this(new Uri("http://localhost:3000/"), required, new AClientOptions()){}
public AClient(Uri endpoint, string required, AClientOptions options){}
```

This PR is to fix that false alert.

### DIFF
--- a/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0006Tests.cs
+++ b/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0006Tests.cs
@@ -44,6 +44,41 @@ namespace RandomNamespace
         }
 
         [Fact]
+        public async Task AZC0006NotProducedForEquivalentOneWithoutEndpoint()
+        {
+            const string code = @"
+namespace RandomNamespace
+{
+    public class SomeClientOptions : Azure.Core.ClientOptions { }
+
+    public class SomeClient
+    {
+        public SomeClient() {}
+        public SomeClient(System.Uri endpoint, SomeClientOptions options) {}
+    }
+}";
+            await Verifier.VerifyAnalyzerAsync(code);
+        }
+
+        [Fact]
+        public async Task AZC0006NotProducedForEquivalentOneWithParameterAndWithoutEndpoint()
+        {
+            const string code = @"
+namespace RandomNamespace
+{
+    public class SomeClientOptions : Azure.Core.ClientOptions { }
+
+    public class SomeClient
+    {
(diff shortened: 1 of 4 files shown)


jsquire (changes requested):
Reading Ted's comment on the linked issue, I do not think the above is correct.    What I believe Ted stated is that, by default, we want to limit constructors to the pattern:

```csharp
public SomeType(Uri endpoint, AClientOptions options);
```

If we have a constructor pattern like the above, it would be a special case that the architects allowed to deviate from our standard patterns and, thus, would require suppressing the analyzer.

@tg-msft, @KrzysztofCwalina:  Can you offer thoughts, please?

KrzysztofCwalina:
I agree with that @jsquire said. Moreover, if we do allow an additional required parameter (per architect review), the parameter would have to always be in the same location in the signature, not like in the description of the PR, i.e. the signatures would have to be:

```
public AClient(string required): this(new Uri("http://localhost:3000/"), required, new AClientOptions()){}
public AClient(string required, Uri endpoint, AClientOptions options){}
```

pshao25:
@jsquire @KrzysztofCwalina So to confirm the optional endpoint cases:
1. if there is no required parameter, constructors should be
```C#
public AClient(): this(new Uri("http://localhost:3000/"), new AClientOptions()){}
public AClient(Uri endpoint, AClientOptions options){}
```
2. if there is required parameter, constructors should be
```C#
public AClient(string required): this(required, new Uri("http://localhost:3000/"),new AClientOptions()){}
public AClient(string required, Uri endpoint, AClientOptions options){}
```

And both of these "optional endpoint" cases should fail the analyzer to let customers suppress themselves.

@KrzysztofCwalina @tg-msft please confirm especially case 2 because our generator is not going this way now.

jsquire:
@pshao25:  For this statement:

> And both of these "optional endpoint" cases should fail the analyzer to let customers suppress themselves.

Who are "customers?"   The Azure SDK analyzers run only in the Azure SDK repositories.   So, the rule enforcement is intentionally scoped to Azure conventions.   If a customer is using TypeSpec and our C# generator for TypeSpec to create a non-Azure library, they are doing so in their own repository - which does not have these analyzers.

The root question that we need to answer here is "will Azure libraries in our repository every deviate from this rule without an architect exception?"

PR #8480, opened 2024-06-20
Revert "Fix  high sev security issue"

weshaggard:
Reverts Azure/azure-sdk-tools#8471

@wanlwanl @raych1 @qiaozha I had to revert this as it is breaking a number of pipelines that aren't setup to correctly sync submodules. We should not be using submodules in our repos because a number of these issues.


scbedd (approved):

scbedd:
I've spent the past 30 minutes attempting to debug my weird SDL `code 128` errors on exactly this. Definitely happy to see it go.

PR #8484, opened 2024-06-20
[stress testing]: Force amd64 as the build platform for images we push

richardpark-msft:
Without this building and pushing from ARM Mac's (or any ARM platform) produce images that just fail with mysterious errors because AKS and your local arch don't match.


benbp (approved):

richardpark-msft:
@benbp, looks like the 'sync-eng-common' step needs a manual approval to create some sync PRs?

Issue #8379, opened 2024-06-05
[tsp-client] support specific setting about `service-dir` for specific language emitter in tspconfig.yaml


msyyc:
According to https://github.com/Azure/typespec-azure/blob/161add8c6d5c3d6e74b5dcde4a3b6a4e41cc57a2/docs/howtos/Generating%20with%20tsp-client/tsp_client.md#per-repository-set-up, all language emitters shall share same value of `servide-dir` as part of output folder, which makes sense for new service.
![image](https://github.com/Azure/azure-sdk-tools/assets/70930885/09afba1d-2035-44dc-8b5a-555f41ec2be0)

However, for existing service, SDK across may have different output folder (e.g. avs in python is in [sdk/compute](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/compute) while in java is [sdk/avs](https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/avs)). So we hope tsp-client could support specific `service-dir` for different emitter. For example:
```
parameters:
  "service-dir":
    "@azure-tools/typespec-python": "sdk/contosowidgetmanager-python"
    default: "sdk/contosowidgetmanager"
...
```

raych1:
The Go management plane SDK is located under [`/sdk/resourcemanager`](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/resourcemanager) folder, with the corresponding `service-dir` being `/sdk/resourcemanager/{serviceName}`. This setting is not consistent with other languages. We need the ability to override the `service-dir` at the emitter level. 

@catalinaperalta , I discussed with @weshaggard, and we aligned on adding support for the override capability for `service-dir` in `tsp-client` at the SDK emitter level. This approach allows us to avoid changing the way we call the SDK emitter and the SDK emitters themselves. This feature request is a blocker for MPG. Could you please prioritize this feature request?

//CC: @msyyc @Alancere

raych1:
we might consider override it at SDK emitter config part like the following:
```yaml
"@azure-tools/typespec-go":
    service-dir: "sdk/resourcemanager/aad"
    package-dir: "armaad"
    flavor: azure
```

catalinaperalta:
Hello, I'm happy to support this new flag in tspconfig.yaml, but need more clarity? Will it be set as part of the global parameters or under the specific emitter configuration parameters? Please let me know

catalinaperalta:
[...]

PR #8492, opened 2024-06-21
[tsp-client] Prefer emitter specific service-dir if specified

catalinaperalta:
Fixes https://github.com/Azure/azure-sdk-tools/issues/8379

### DIFF
--- a/tools/tsp-client/src/index.ts
+++ b/tools/tsp-client/src/index.ts
@@ -51,7 +51,12 @@ async function sdkInit(
       throw new Error(`tspconfig.yaml is empty at ${tspConfigPath}`);
     }
     const configYaml = parseYaml(data);
-    const serviceDir = configYaml?.parameters?.["service-dir"]?.default;
+    // Check if service-dir is defined in the emitter specific configurations in tspconfig.yaml
+    let serviceDir = configYaml?.options?.[emitter]?.["service-dir"];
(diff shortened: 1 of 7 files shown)


tools/tsp-client/src/index.ts
```diff
@@ -84,7 +89,12 @@ async function sdkInit(
       throw new Error(`tspconfig.yaml is empty at ${config}`);
     }
     const configYaml = parseYaml(data);
-    const serviceDir = configYaml?.parameters?.["service-dir"]?.default;
+    // Check if service-dir is defined in the emitter specific configurations in tspconfig.yaml
```
joheredi, line 92: maybe you could encapsulate this into its own function that you can share in the 2 places where this is being done

joheredi: This is really personal preference but wanted to share in case you find it useful. You could do this conditional assignment in a single line
```ts
// Check if service-dir is defined in the emitter specific configurations in tspconfig.yaml
// Default to the top level service-dir parameter in tspconfig.yaml
const serviceDir = configYaml?.options?.[emitter]?.["service-dir"] ?? configYaml?.parameters?.["service-dir"]?.default;
```

joheredi (approved):

iscai-msft (approved):

PR #8496, opened 2024-06-24
Node Publishing Update

scbedd:
Looks like when I tested an actual release, I tested the only case that would actually work when merging #8460 

@catalinaperalta  hit issues this past Friday with this as a result. I unblocked her, but wanted to make an adjustment to re-use the `SemVer` package before merging.

Unfortunately, it looks like if I use `SemVer`, it marks packages < 1.0.0 (even if if GA-marked) as `Prerelease`. While I agree with this for our production packages, our internal tooling conflicts with this bar.

Specifically, the typespec-emitter `latest` version is `0.4.8`. If we used `SemVer` and `IsPrerelease` property to determine whether to set `latest` or `beta` tag, we would erroneously mark the latest version of `typespec-emitter` with `beta`. At least until 1.0.0. For now I'm leaving the simple tag check in place.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/launchSettings.json
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/launchSettings.json
@@ -2,6 +2,7 @@
   "profiles": {
     "Azure.Sdk.Tools.TestProxy": {
       "commandName": "Project",
+      "commandLineArgs": "start --storage-location=C:/repo/azure-sdk-for-net",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
         "Logging__LogLevel__Microsoft": "Information"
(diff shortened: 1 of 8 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/launchSettings.json
```diff
@@ -2,6 +2,7 @@
   "profiles": {
     "Azure.Sdk.Tools.TestProxy": {
       "commandName": "Project",
+      "commandLineArgs": "start --storage-location=C:/repo/azure-sdk-for-net",
```
weshaggard, line 5: Did you intend to have your local paths here?

weshaggard (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/launchSettings.json
scbedd: Frustration. Def shouldn't be present. Thanks for the catch Wes.

PR #8527, opened 2024-07-01
Use repositories file by default for GH scripts

heaths:
The default UX shouldn't change: to override any default values for disambiguating parameters, you had to pass values. If we change the default parameter set (or, for a couple scripts, leave it) to use the file we can remove the duplicate repo and language information from the `-Repositories` and `-Languages` parameters because to override those - even if they were the default parameter - you had to specify a different value. The end result should be no change.

Also cleans up warnings in the README and a couple other scripts using their respective linters I have enabled in VSCode.

### DIFF
--- a/tools/github/README.md
+++ b/tools/github/README.md
@@ -1,6 +1,6 @@
 # Overview
 
-This area is focused on the management of the Azure SDK GitHub repositories, containing tools and data used for that purpose.  
+This area is focused on the management of the Azure SDK GitHub repositories, containing tools and data used for that purpose.
 
 ## Prerequisites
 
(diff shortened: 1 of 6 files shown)


tools/github/README.md
```diff
@@ -10,83 +10,86 @@ This area is focused on the management of the Azure SDK GitHub repositories, con
 
 ## Structure
 
-* **root**
```
jsquire, line 13: You done messed with my formatting.   I want to keep the line break, please.

hallipr (approved):

tools/github/scripts/Validate-AzsdkCodeOwner.ps1
```diff
@@ -16,8 +16,8 @@ $hasPermissions = $false
 $orgResponse = (gh api "https://api.github.com/users/$UserName/orgs")
 $orgs = $orgResponse | ConvertFrom-Json
 
-if ($orgs -ne $null) {
-    $orgs = $orgs | select -Expand login
+if (!$org) {
+    $orgs = $orgs | Select-Object -Expand login
```
jsquire, line 20:
```suggestion
    try {
        $orgs = $orgs | Select-Object -Expand login
    } catch {
        $orgs = @()
   }
```

Since you're making a change anyway, I ran into an unexpected edge case today where the user was valid and there was a payload but it did not contain a login.

jsquire (changes requested):
Good with the script changes, but I'd like to retain the README formatting as it was.

tools/github/README.md
heaths: Commonmark elides those when rendered. markdown-lint was showing a lot of warnings, which tends to hide other - potentially valid - warnings.

What line break are you talking about anyway? All the space at the beginning would've been elided out.

tools/github/scripts/Validate-AzsdkCodeOwner.ps1
heaths: Better to use `-ErrorAction Ignore` in that case.

tools/github/README.md
hallipr: I think the double space at the end of a line forces a line break when rendered

hallipr: image

vs

image

tools/github/scripts/Validate-AzsdkCodeOwner.ps1
jsquire: wouldn't that cause `$orgs` to be null, then?

tools/github/README.md
heaths: That's the "wrong" way to do it, though. Git itself doesn't like trailing whitespace, and to get a line break in markdown the way that actually looks closer to how it renders (so it's readable in plain text form as well) is a separate line break e.g.,
```markdown
- foo

  bar
```
Whether using `-` (or `*`, but markdown-lint prefers `-` which is more generally what we use anyway, like even in this file) or `1.`, the number of whitespace characters has to match the first rendered line char e.g., 2 spaces for `-`.

heaths: @jsquire and I clarified what he meant offline, though, so this is already fixed.

tools/github/scripts/Validate-AzsdkCodeOwner.ps1
heaths: Using exceptions for conditional logic leads down the highway to the danger zone.

tools/github/scripts/Validate-AzsdkCodeOwner.ps1
```diff
@@ -16,8 +16,8 @@ $hasPermissions = $false
 $orgResponse = (gh api "https://api.github.com/users/$UserName/orgs")
 $orgs = $orgResponse | ConvertFrom-Json
 
-if ($orgs -ne $null) {
-    $orgs = $orgs | select -Expand login
+if (!$org) {
```
jsquire, line 19: Doesn't this reverse the logic?   Previously, it was a 'not null" check, now it's a (forgiving) null check?

tools/github/README.md
jsquire: I'm still seeing no line break.  It's rendering inline as Patrick's graphic shows.   Is there a pending change that needs to be pushed?

jsquire (approved):

PR #8528, opened 2024-07-01
Fix default value for env vars in build-test-resource-config template

benbp:
The null default breaks any pipelines that don't specify env vars

### DIFF
--- a/eng/common/TestResources/build-test-resource-config.yml
+++ b/eng/common/TestResources/build-test-resource-config.yml
@@ -8,7 +8,7 @@ parameters:
   # EnvVars is used to help diagnose variable conflict issues early
   - name: EnvVars
     type: object
-    default: null
+    default: {}
   - name: SubscriptionConfigurationFilePaths
     type: object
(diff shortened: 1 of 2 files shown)


weshaggard (approved):

eng/common/TestResources/build-test-resource-config.yml
```diff
@@ -8,14 +8,15 @@ parameters:
...
-    default: null
+    default: {}
   - name: SubscriptionConfigurationFilePaths
     type: object
     default: null
 
 steps:
   - task: AzurePowerShell@5
     displayName: Set Pipeline Subnet Info
+    condition: ne(variables['Pool'], '')
```
weshaggard, line 19: Should also add succeeded into the condition otherwise this might try to run when earlier steps have failed.

benbp: https://github.com/Azure/azure-sdk-tools/pull/8572

Issue #8523, opened 2024-07-01
[tsp-client] The output of the `tsp compile` command should be logged using the Logger


raych1:
**Description**
The output while calling the SDK emitter in `tsp-client` is currently only displayed in the console. For better troubleshooting, this output should also be emitted to the Logger so that it can be displayed in the pipeline scenario.

This is the code location:
https://github.com/Azure/azure-sdk-tools/blob/eaa42a758c1478fa7a3262a53254d3fe26d8ea37/tools/tsp-client/src/typespec.ts#L112

//CC: @catalinaperalta @chunyu3

chunyu3:
Hello @catalinaperalta in .NET generator we will log some information, but when `tsp-client` execute .Net emitter, the logs of .NET emitter are missing.

When I run local, it can see the logs, but when run in ci pipeline, the logs are missing

catalinaperalta:
Hi all, I opened a PR addressing this yesterday, you should be able to see it linked to this issue. Waiting on a review then will release a tsp-client update. Feel free to add your own review on the PR as well :)

catalinaperalta:
v0.9.1 of tsp-client was released and is not printing the diagnostics with the logger. Please update to that version and give it a try. Thanks

catalinaperalta:
@raych1 the output of the diagnostics causes a lot of noise when generating using tsp-client. Any issues if I move this to the debug level? I think you're already using the `--debug` flag in the pipeline, right?

PR #8529, opened 2024-07-01
[tsp-client] Fix tsp compile diagnostics logging

catalinaperalta:
Use logger for diagnostics and exit if there are diagnostics reported during the tsp `compile` call

Fixes https://github.com/Azure/azure-sdk-tools/issues/8523

### DIFF
--- a/tools/tsp-client/src/typespec.ts
+++ b/tools/tsp-client/src/typespec.ts
@@ -3,6 +3,7 @@ import { ModuleResolutionResult, resolveModule, ResolveModuleHost } from "@types
 import { Logger } from "./log.js";
 import { readFile, readdir, realpath, stat } from "fs/promises";
 import { pathToFileURL } from "url";
+import { exit } from "process";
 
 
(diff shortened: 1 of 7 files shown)


tools/tsp-client/src/typespec.ts
```diff
@@ -109,10 +109,11 @@ export async function compileTsp({
     for (const diagnostic of program.diagnostics) {
       const location = getSourceLocation(diagnostic.target);
       const source = location ? location.file.path : "unknown";
-      console.error(
+      Logger.error(
         `${diagnostic.severity}: ${diagnostic.code} - ${diagnostic.message} @ ${source}`,
```
timotheeguerin, line 113: Btw typespec export functions to format diagnostics nicely as it seems here you lose information

catalinaperalta: I see, let me take a look at that :)

timotheeguerin (approved):

PR #8546, opened 2024-07-02
Add the ability to manage Auto-Close on issues

RickWinter:
This PR is providing a way to skip the Enforce Max Life of Issues

### DIFF
--- a/tools/github/data/common-labels.csv
+++ b/tools/github/data/common-labels.csv
@@ -228,6 +228,7 @@ VideoAnalyzer,Azure Video Analyzer,e99695
 Visual Studio,,e99695
 Web Apps,,e99695
 WebPubSub,,e99695
+auto-close-exempt,Prevents the auto-close from closing based on max lifetime,bbbbbb
 blocking-release,Blocks release,d73a49
 breaking-change,,d73a49
 bug,This issue requires a change to an existing behavior in the product in order to be resolved.,eaa875
--- a/tools/github/data/common-labels.csv
+++ b/tools/github/data/common-labels.csv
@@ -228,7 +228,7 @@ VideoAnalyzer,Azure Video Analyzer,e99695
 Visual Studio,,e99695
 Web Apps,,e99695
 WebPubSub,,e99695
-auto-close-exempt,Prevents the auto-close from closing based on max lifetime,bbbbbb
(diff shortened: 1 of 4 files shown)


tools/github/data/common-labels.csv
```diff
@@ -228,6 +228,7 @@ VideoAnalyzer,Azure Video Analyzer,e99695
 Visual Studio,,e99695
 Web Apps,,e99695
 WebPubSub,,e99695
+auto-close-exempt,Prevents the auto-close from closing based on max lifetime,bbbbbb
```
JimSuplizio, line 231: @RickWinter, I'm going to defer to @jsquire and @ronniegeraghty on the color of this label. I'm sure one or both will want a different color.

JimSuplizio (approved):
As per our conversation, I'm fine with the changes. You added the label constant and then added it to the query's exclude list which is exactly what needed to be done, thank you.

@jsquire or @ronniegeraghty, one of you is probably going to have to run the sync script so the auto-close-exempt label gets added to all of the repositories after this PR gets merged.

jsquire:
@ronniegeraghty:  Once you tell me we're good, I'll do a run to push that label out to the managed repos.

tools/github/data/common-labels.csv
ronniegeraghty: I think the color is fine. It's very grey and dull, and I don't really want to bring any attention to this label. I want it to be a tool only the SDK team uses so an unassuming color works here.

ronniegeraghty (approved):

ronniegeraghty:
@jsquire, @RickWinter, @JimSuplizio 
All looks good to me.

tools/github/data/common-labels.csv
jsquire: I'm going to ask that we adjust to match the other greyed out labels, just for consistency.  That's the form that I've pushed out to the other repos.

jsquire, line 231:
```suggestion
auto-close-exempt,Prevents the auto-close from closing based on max lifetime,ededed
```

weshaggard:
I believe this is for issue https://github.com/Azure/azure-sdk-tools/issues/8559

PR #8547, opened 2024-07-02
[tsp-client] Fix compiler diagnostic reporting

catalinaperalta:
More fixes related to diagnostic reporting. Return boolean from compileTsp to exit after file clean up in case of diagnostics.

### DIFF
--- a/tools/tsp-client/src/typespec.ts
+++ b/tools/tsp-client/src/typespec.ts
@@ -64,7 +64,7 @@ export async function compileTsp({
   resolvedMainFilePath: string;
   additionalEmitterOptions?: string;
   saveInputs?: boolean;
-}) {
+}): Promise<boolean> {
   const parsedEntrypoint = getDirectoryPath(resolvedMainFilePath);
   const { compile, NodeHost, resolveCompilerOptions } = await importTsp(parsedEntrypoint);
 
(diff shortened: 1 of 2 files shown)


tools/tsp-client/src/typespec.ts
```diff
@@ -100,7 +100,8 @@ export async function compileTsp({
   Logger.debug(`Compiler options: ${JSON.stringify(options)}`);
   if (diagnostics.length > 0) {
     // This should not happen, but if it does, we should log it.
-    Logger.debug(`Compiler options diagnostic information: ${JSON.stringify(diagnostics)}`);
+    Logger.error(`Compiler options diagnostic information: ${JSON.stringify(diagnostics)}`);
```
catalinaperalta, line 103: @timotheeguerin can we also use `formatDiagnostic()` with the diagnostics returned from the `resolveCompilerOptions()`?

catalinaperalta: Also, do you have an example of a configuration that would cause an error? I'd like to test this out.

timotheeguerin: Yep, any typespec diagnostic

timotheeguerin (approved):

PR #8549, opened 2024-07-02
Add `test-proxy-standalone.yml`

scbedd:


eng/common/testproxy/test-proxy.ps1
```diff
@@ -0,0 +1,162 @@
...
+        default { throw "Unable to determine system architecture. uname -m returned $unameOutput." }
+    }
+}
+
+function Get-Proxy-Meta () {
+    $ErrorActionPreferenceDefault = $ErrorActionPreference
+    $ErrorActionPreference = "Stop"
+
+    $os = "unknown"
+    $machine = Get-SystemArchitecture
```
benbp, line 56: Won't this halt/throw on a non-ubuntu OS? Should you put this check within the `$IsLinux` conditional instead?

scbedd: Nope! Works just fine 👍 

Here's an [example](https://dev.azure.com/azure-sdk/public/_build/results?buildId=3929950&view=logs&j=2993fb17-9fe5-5b0b-e997-efa0cab8f900&t=cf38eb1d-d0e9-5773-1107-71d89ce67904&l=21).

benbp (approved):

PR #8558, opened 2024-07-03
Support storage network access and worm removal in remove test resources script

benbp:
We need to be running worm removal on storage accounts in the pipeline to avoid resource build-ups for our async cleanup script. Additionally, sometimes the network access rules we apply fail on the backend due to RP throttling. In those cases we can't disable WORM because the network policy is set to Deny, meaning blob's can't be deleted (pre-req for WORM disablement). This change includes updating the network policy within the remove script to prevent these issues as a failsafe.


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -131,272 +132,6 @@ if (!$PSBoundParameters.ContainsKey('ErrorAction')) {
     $ErrorActionPreference = 'Stop'
 }
 
-function Log($Message)
```
benbp, line 134: I moved all helper functions into `TestResources-Helpers.ps1`

eng/common/TestResources/TestResources-Helpers.ps1
```diff
@@ -0,0 +1,327 @@
...
+                    $ipRanges = $AllowIpRanges | ForEach-Object {
+                        @{ Action = 'allow'; IPAddressOrRange = $_ }
+                    }
+                    Retry { Update-AzStorageAccountNetworkRuleSet -ResourceGroupName $ResourceGroupName -Name $account.Name -IPRule $ipRanges | Out-Null }
+                }
+                elseif (!$CI) {
+                    Write-Host "Enabling access to '$($account.Name)' from client IP"
+                    $clientIp ??= Retry { Invoke-RestMethod -Uri 'https://icanhazip.com/' }  # cloudflare owned ip site
+                    $clientIp = $clientIp.Trim()
+                    $ipRanges = Get-AzStorageAccountNetworkRuleSet -ResourceGroupName $ResourceGroupName -Name $account.Name
```
benbp, line 297: Everything here and below is net new. I added a check for the client IP overlapping with the AllowIpRange, in cases where the storage account has corp net IP ranges added by policy, and the script tries to add the user's IP already within said range.

eng/common/scripts/Helpers/Resource-Helpers.ps1
```diff
@@ -204,3 +204,82 @@ function Wait-PurgeableResourceJob {
     }
   }
 }
+
+# Helper function for removing storage accounts with WORM that sometimes get leaked from live tests not set up to clean
+# up their resource policies
+function Remove-WormStorageAccounts() {
```
benbp, line 210: This is moved from an independent script. Instead of running that script to clean up storage accounts, we can just run `Remove-TestResources.ps1` instead.

eng/common/TestResources/Remove-TestResources.ps1
```diff
@@ -241,6 +257,9 @@ $verifyDeleteScript = {
 # Get any resources that can be purged after the resource group is deleted coerced into a collection even if empty.
 $purgeableResources = Get-PurgeableGroupResources $ResourceGroupName
 
+SetStorageNetworkAccessRules -ResourceGroupName $ResourceGroupName -AllowIpRanges $AllowIpRanges -Override -CI:$CI
```
weshaggard, line 260: Why do we need to set these in order to clean-up?

weshaggard (approved):

eng/common/TestResources/Remove-TestResources.ps1
benbp: A pre-requisite for disabling WORM (which prevents storage account deletion) is deleting any blobs that are WORM enabled.

We cannot delete blobs if the network policy is set to `Deny` without any vnet/ip rules added, so we just get a 403 error from storage trying to delete the blob and fail out.

PR #8278, opened 2024-05-15
Migrate stress cluster to use workload identity

benbp:
Notes TBD, a couple more edits need to be made (docs, test updates, and some more config deletion). Then I will publish out of draft.

PR #7830, opened 2024-03-07
Also write out an .env.sh file which can be safely be 'source'd into a shell.

richardpark-msft:

PR #8588, opened 2024-07-09
Add storage cluster to stress environments. Update federated auth documentation.

benbp:
I have built out a new default stress cluster in the storage partner team subscription that they can use. Adding it to the config defaults so they can use it out of the box.

Also updating the documentation to reflect the changes made in https://github.com/Azure/azure-sdk-tools/pull/8278 to support workload identity as the auth mode for the stress clusters.

### DIFF
--- a/tools/stress-cluster/chaos/README.md
+++ b/tools/stress-cluster/chaos/README.md
@@ -242,7 +242,7 @@ Additionally, several values are made available as environment variables via the
 - `POD_NAMESPACE` - The kubernetes namespace the container is running in, useful for custom telemetry.
 - `DEBUG_SHARE` - See [stress test file share](#stress-test-file-share)
 - `DEBUG_SHARE_ROOT` - See [stress test file share](#stress-test-file-share)
-- `AZURE_SUBSCRIPTION_ID` - The azure subscription id the stress test will authenticate and deploy resources to.
+- `AZURE_SUBSCRIPTION_ID` - The Azure subscription id the stress test will authenticate and deploy resources to.
 - `AZURE_TENANT_ID` - The Azure tenant id the stress test will authenticate to. Set by AKS.
 - `AZURE_CLIENT_ID` - The AAD principal the stress test will authenticate as. Set by AKS.
 - `AZURE_FEDERATED_TOKEN_FILE` - The path to the federated identity token that can be used to login with azure cli or the identity SDK. Set by AKS.
--- a/tools/stress-cluster/chaos/README.md
+++ b/tools/stress-cluster/chaos/README.md
@@ -243,7 +243,7 @@ Additionally, several values are made available as environment variables via the
 - `DEBUG_SHARE` - See [stress test file share](#stress-test-file-share)
 - `DEBUG_SHARE_ROOT` - See [stress test file share](#stress-test-file-share)
 - `AZURE_SUBSCRIPTION_ID` - The azure subscription id the stress test will authenticate and deploy resources to.
-- `AZURE_TENANT_ID` - The azure tenant id the stress test will authenticate to. Set by AKS.
+- `AZURE_TENANT_ID` - The Azure tenant id the stress test will authenticate to. Set by AKS.
 - `AZURE_CLIENT_ID` - The AAD principal the stress test will authenticate as. Set by AKS.
 - `AZURE_FEDERATED_TOKEN_FILE` - The path to the federated identity token that can be used to login with azure cli or the identity SDK. Set by AKS.
 
--- a/tools/stress-cluster/chaos/README.md
+++ b/tools/stress-cluster/chaos/README.md
@@ -186,14 +186,13 @@ Fields in `Chart.yaml`
 
 Stress tests are authenticated via [workload identity for AKS](https://learn.microsoft.com/azure/aks/workload-identity-overview).
 
-To authenticate to Azure, tests can choose to use `DefaultAzureCredential`, `WorkloadIdentityCredential`, or `AzureCliCredential`,
+To authenticate to Azure, tests can choose to use `DefaultAzureCredential` or `WorkloadIdentityCredential`,
 though it is recommended to use `DefaultAzureCredential` for ease of switching between local machine runs and container runs in the cluster.
 The container that a stress test runs in will have a federated identity token available in its environment that can be used to authenticate.
 This credential is made available automatically by the cluster.
 
-`DefaultAzureCredential` and `WorkloadIdentityCredential` should work automatically with no setup. `AzureCliCredential` can also be used
-provided a login step is added to the test invocation. This method is also useful for test jobs that may need to exercise additional azure
-cli commands. See an example [here](https://github.com/Azure/azure-sdk-tools/blob/main/tools/stress-cluster/chaos/examples/stress-deployment-example/templates/deploy-job.yaml)
+Azure CLI can also be used for scripting purposes provided a login step is added to the test invocation.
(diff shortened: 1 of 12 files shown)


tools/stress-cluster/chaos/README.md
```diff
@@ -181,6 +182,34 @@ Fields in `Chart.yaml`
...
+
+Stress tests are authenticated via [workload identity for AKS](https://learn.microsoft.com/azure/aks/workload-identity-overview).
+
+To authenticate to Azure, tests can choose to use `DefaultAzureCredential`, `WorkloadIdentityCredential`, or `AzureCliCredential`,
+though it is recommended to use `DefaultAzureCredential` for ease of switching between local machine runs and container runs in the cluster.
+The container that a stress test runs in will have a federated identity token available in its environment that can be used to authenticate.
+This credential is made available automatically by the cluster.
+
+`DefaultAzureCredential` and `WorkloadIdentityCredential` should work automatically with no setup. `AzureCliCredential` can also be used
+provided a login step is added to the test invocation. This method is also useful for test jobs that may need to exercise additional azure
```
richardpark-msft, line 195: I think we should just eliminate any mention of AzureCliCredential. A bit later, in this doc, you cover how they could use the Azure CLI - that seems like a reasonable spot to document the possibility.

We definitely don't want them to login to the CLI just to get their normal test running (ie, mentioning it here confuses things).

tools/stress-cluster/chaos/README.md
```diff
@@ -181,6 +182,34 @@ Fields in `Chart.yaml`
...
+cli commands. See an example [here](https://github.com/Azure/azure-sdk-tools/blob/main/tools/stress-cluster/chaos/examples/stress-deployment-example/templates/deploy-job.yaml)
+or in the following snippet:
+
+```
+command: ['bash', '-c']
+args:
+  - |
+      source $ENV_FILE &&
```
richardpark-msft, line 203: I have a PR to make the .env file sourceable (sometimes the unquoted values mess up bash). 

https://github.com/Azure/azure-sdk-tools/pull/7830

We can update this doc after that comes in, or perhaps you are quoting everything now?

tools/stress-cluster/chaos/README.md
```diff
@@ -181,6 +182,34 @@ Fields in `Chart.yaml`
...
+      az login --federated-token "$(cat $AZURE_FEDERATED_TOKEN_FILE)" --service-principal -u "$AZURE_CLIENT_ID" -t "$AZURE_TENANT_ID" &&
+      az account set -s $AZURE_SUBSCRIPTION_ID &&
+      <your custom test script here>
+```
+
+Each federated identity is backed by a managed identity with authorization to the cluster subscription. For all permissions
+given to this principal, see [workload app roles](https://github.com/Azure/azure-sdk-tools/blob/main/tools/stress-cluster/cluster/azure/cluster/workloadapproles.bicep).
+When a new namespace is created in the a stress cluster, a [federated identity](https://learn.microsoft.com/en-us/graph/api/resources/federatedidentitycredentials-overview?view=graph-rest-1.0) specific to that namespace is created against a pool of managed identities by the stress watcher service (there is a hard limit of 20 federated identities per managed identity). When a namespace is deleted, the corresponding federated identity is also deleted.
```
richardpark-msft, line 211:
```suggestion
When a new namespace is created in the stress cluster, a [federated identity](https://learn.microsoft.com/en-us/graph/api/resources/federatedidentitycredentials-overview?view=graph-rest-1.0) specific to that namespace is created against a pool of managed identities by the stress watcher service (there is a hard limit of 20 federated identities per managed identity). When a namespace is deleted, the corresponding federated identity is also deleted.
```

tools/stress-cluster/chaos/README.md
```diff
@@ -218,6 +242,10 @@ Additionally, several values are made available as environment variables via the
 - `POD_NAMESPACE` - The kubernetes namespace the container is running in, useful for custom telemetry.
 - `DEBUG_SHARE` - See [stress test file share](#stress-test-file-share)
 - `DEBUG_SHARE_ROOT` - See [stress test file share](#stress-test-file-share)
+- `AZURE_SUBSCRIPTION_ID` - The azure subscription id the stress test will authenticate and deploy resources to.
+- `AZURE_TENANT_ID` - The azure tenant id the stress test will authenticate to. Set by AKS.
```
richardpark-msft, line 246:
```suggestion
- `AZURE_TENANT_ID` - The Azure tenant id the stress test will authenticate to. Set by AKS.
```

tools/stress-cluster/chaos/README.md
```diff
@@ -218,6 +242,10 @@ Additionally, several values are made available as environment variables via the
 - `POD_NAMESPACE` - The kubernetes namespace the container is running in, useful for custom telemetry.
 - `DEBUG_SHARE` - See [stress test file share](#stress-test-file-share)
 - `DEBUG_SHARE_ROOT` - See [stress test file share](#stress-test-file-share)
+- `AZURE_SUBSCRIPTION_ID` - The azure subscription id the stress test will authenticate and deploy resources to.
```
richardpark-msft, line 245:
```suggestion
- `AZURE_SUBSCRIPTION_ID` - The Azure subscription id the stress test will authenticate and deploy resources to.
```

tools/stress-cluster/chaos/README.md
```diff
@@ -218,6 +242,10 @@ Additionally, several values are made available as environment variables via the
...
+- `AZURE_TENANT_ID` - The azure tenant id the stress test will authenticate to. Set by AKS.
+- `AZURE_CLIENT_ID` - The AAD principal the stress test will authenticate as. Set by AKS.
```
richardpark-msft, line 247:
```suggestion
- `AZURE_CLIENT_ID` - The Entra principal the stress test will authenticate as. Set by AKS.
```

tools/stress-cluster/chaos/README.md
```diff
@@ -218,6 +242,10 @@ Additionally, several values are made available as environment variables via the
...
+- `AZURE_FEDERATED_TOKEN_FILE` - The path to the federated identity token that can be used to login with azure cli or the identity SDK. Set by AKS.
```
richardpark-msft, line 248:
```suggestion
- `AZURE_FEDERATED_TOKEN_FILE` - The path to the federated identity token that can be used to login with Azure CLI or the Identity SDK. Set by AKS.
```

richardpark-msft (approved):

tools/stress-cluster/chaos/README.md
benbp: Fair point, and I'm considering not even mentioning `WorkloadIdentityCredential` either. We should just say in code, use DAC, and in a script, use az cli login?

tools/stress-cluster/chaos/README.md
benbp: Approved that one, yeah we can rebase after merge and get that one in.

tools/stress-cluster/chaos/README.md
richardpark-msft: I was okay with that, mostly because I it helps in identifying which credential was supposed to work. Went through this a little with @LibbaLawrence where we weren't sure which credential failure was the important one.

tools/stress-cluster/chaos/README.md
benbp: I said `a` stress cluster because we have many. But if you think `the` still makes more sense I can take this change.

PR #8602, opened 2024-07-10
Adjustments enabling change-specific PackageInfo

scbedd:
Effectively, this allows `Save-Package-Properties` to honor the `diff` file.

[In practice it looks like this](https://dev.azure.com/azure-sdk/public/_build/results?buildId=3950939&view=results)

todo:

- [x] Do we want to keep `ServiceInput` or duplicate this thing into another script entirely?
- [x] Remove debugging outputs

On my `build` phase, this is the yml gating access to the filter

```yml
- ${{ if eq(variables['Build.Reason'], 'PullRequest') }}:
    - pwsh: |
        mkdir -p $(Build.ArtifactStagingDirectory)/diff
      displayName: Create PR Diff Folder

    - task: PowerShell@2
      inputs:
        filePath: 'eng/common/scripts/Generate-PR-Diff.ps1'
        arguments: '-TargetPath "$(Build.SourcesDirectory)" -ArtifactPath $(Build.ArtifactStagingDirectory)/diff'
      displayName: 'Generate PR Diff'

    - pwsh: |
        Write-Host "We freshly generated the PR diff, and this is what it sees!"
        Get-ChildItem -R -Force $(Build.ArtifactStagingDirectory)/diff | % { $_.FullName }
        cat $(Build.ArtifactStagingDirectory)/diff/diff.json
      displayName: Dump PR Diff

    - task: Powershell@2
      inputs:
        filePath: $(Build.SourcesDirectory)/eng/common/scripts/Save-Package-Properties.ps1
        arguments: >
          -ServiceInput $(Build.ArtifactStagingDirectory)/diff/diff.json
          -OutDirectory $(Build.ArtifactStagingDirectory)/PackageInfo
        pwsh: true
        workingDirectory: $(Pipeline.Workspace)
      displayName: Save package properties filtered for PR
  - ${{ else }}:
    - task: Powershell@2
      inputs:
        filePath: $(Build.SourcesDirectory)/eng/common/scripts/Save-Package-Properties.ps1
        arguments: >
          -ServiceInput ${{parameters.ServiceDirectory}}
          -OutDirectory $(Build.ArtifactStagingDirectory)/PackageInfo
        pwsh: true
        workingDirectory: $(Pipeline.Workspace)
      displayName: Save package properties for service
```
Everywhere else that may need to do work against the changeset will get this `packages` artifact, then utilize it to determine targeted packages

```
  - ${{ if eq(variables['Build.Reason'], 'PullRequest') }}:
    # if we are in a PR, utilize the fact that the package properties folder will already be downloaded in the Packages folder
    # any json file that exists within the PackagePropertiesFolder will be added to the targeting string
    - pwsh: |
        $packageProperties = Get-ChildItem -Recurse -Force "${{ parameters.PackagePropertiesFolder }}" `
          | Where-Object { $_.Extension -eq '.json' } `
          | Foreach-Object { $_.Name } `
          | ForEach-Object { $_.Replace(".json", "") }

        $setting = $packageProperties -join ","

        Write-Host "Setting TargetingString to $setting"
        Write-Host "##vso[task.setvariable variable=TargetingString;]$setting"
      displayName: Check override of targeting string by variable for PR builds
```

What _this_ specifically willl vary per language. For Python it's just setting `TargetingString` variable to a comma separated list of package names.

EDIT: the rebase worked. got through android patch.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -114,10 +114,12 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
 
     foreach($pkg in $allPackageProperties)
     {
+        $pkgDirectory = Resolve-Path "$($pkg.DirectoryPath)"
+
         foreach($file in $targetedFiles)
         {
             $filePath = Resolve-Path (Join-Path $RepoRoot $file)
-            $shouldInclude = $filePath -like "$($pkg.DirectoryPath)*"
+            $shouldInclude = $filePath -like "$pkgDirectory*"
             if ($shouldInclude) {
                 $packagesWithChanges += $pkg
             }
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -108,7 +108,7 @@ function Get-PkgProperties
 function Get-PrPkgProperties([string]$InputDiffJson) {
     $packagesWithChanges = @()
 
-    $allPackageProperties = Get-AllPkgProperties -ServiceDirectory media
+    $allPackageProperties = Get-AllPkgProperties
     $diff = Get-Content $InputDiffJson | ConvertFrom-Json
     $targetedFiles = $diff.ChangedFiles
 
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -105,6 +105,20 @@ function Get-PkgProperties
     return $null
 }
 
+function Get-PrPkgProperties([string]$InputDiffJson) {
+    $pkgPropsResult = @()
+
+    Write-Host "In Get-PRPkgProperties function"
+
+    if ($GetPRPackageInfoFromRepoFn -and (Test-Path "Function:$GetPRPackageInfoFromRepoFn"))
+    {
+        Write-Host "Attempting to invoke the language one"
+        $pkgPropsResult = &$GetPRPackageInfoFromRepoFn -InputDiffJson $InputDiffJson
+    }
+
+    return $pkgPropsResult
+}
+
 # Takes ServiceName and Repo Root Directory
 # Returns important properties for each package in the specified service, or entire repo if the serviceName is not specified
 # Returns a Table of service key to array values of PS Object with properties @ { pkgName, pkgVersion, pkgDirectoryPath, pkgReadMePath, pkgChangeLogPath }
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -105,21 +105,10 @@ function Get-PkgProperties
     return $null
 }
 
-function Get-PackageFolderFromPath($file) {
-    $parts = $file -split "/"
-
-    if ($parts.Length -lt 3) {
-        return ""
-    }
-    return [System.IO.Path]::Combine($parts[0], $parts[1], $parts[2])
-}
-
-# Takes an input diff json file generated from eng/common/scripts/Generate-PR-Diff.ps1
-# On PR builds, this function should be utilized instead of Get-AllPkgProperties.
 function Get-PrPkgProperties([string]$InputDiffJson) {
     $packagesWithChanges = @()
 
-    $allPackageProperties = Get-AllPkgProperties
+    $allPackageProperties = Get-AllPkgProperties -ServiceDirectory media
     $diff = Get-Content $InputDiffJson | ConvertFrom-Json
     $targetedFiles = $diff.ChangedFiles
 
@@ -127,8 +116,9 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     {
         foreach($file in $targetedFiles)
         {
-            $pkgFolder = Get-PackageFolderFromPath($file)
-            if ($pkgFolder.EndsWith($pkg.ArtifactName)) {
+            $filePath = Resolve-Path (Join-Path $RepoRoot $file)
+            $shouldInclude = $filePath -like "$($pkg.DirectoryPath)*"
+            if ($shouldInclude) {
                 $packagesWithChanges += $pkg
             }
         }
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -105,18 +105,36 @@ function Get-PkgProperties
     return $null
 }
 
+function Get-PackageFolderFromPath($file) {
+    $parts = $file -split "/"
+
+    if ($parts.Length -lt 3) {
+        return ""
+    }
+    return [System.IO.Path]::Combine($parts[0], $parts[1], $parts[2])
+}
+
+# Takes an input diff json file generated from eng/common/scripts/Generate-PR-Diff.ps1
+# On PR builds, this function should be utilized instead of Get-AllPkgProperties.
 function Get-PrPkgProperties([string]$InputDiffJson) {
-    $pkgPropsResult = @()
+    $packagesWithChanges = @()
 
-    Write-Host "In Get-PRPkgProperties function"
+    $allPackageProperties = Get-AllPkgProperties
+    $diff = Get-Content $InputDiffJson | ConvertFrom-Json
+    $targetedFiles = $diff.ChangedFiles
 
-    if ($GetPRPackageInfoFromRepoFn -and (Test-Path "Function:$GetPRPackageInfoFromRepoFn"))
+    foreach($pkg in $allPackageProperties)
     {
-        Write-Host "Attempting to invoke the language one"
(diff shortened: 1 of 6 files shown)


eng/common/scripts/Package-Properties.ps1
```diff
@@ -105,6 +105,20 @@ function Get-PkgProperties
     return $null
 }
 
+function Get-PrPkgProperties([string]$InputDiffJson) {
+    $pkgPropsResult = @()
+
+    Write-Host "In Get-PRPkgProperties function"
+
+    if ($GetPRPackageInfoFromRepoFn -and (Test-Path "Function:$GetPRPackageInfoFromRepoFn"))
```
weshaggard, line 113: What do you expect one of these functions to look like?

eng/common/scripts/Save-Package-Properties.ps1
```diff
@@ -92,7 +92,18 @@ function GetRelativePath($path) {
...
 $exportedPaths = @{}
-$allPackageProperties = Get-AllPkgProperties $serviceDirectory
+
+$allPackageProperties = @()
+
+if ($serviceInput.endswith("diff.json")) {
+  $allPackageProperties = Get-PrPkgProperties $serviceInput
+}
+else {
+  $allPackageProperties = Get-AllPkgProperties $serviceInput
```
weshaggard, line 102: Would it be easier to get all the package infos (i.e. ignore the service directory) and then filter them outside of the common language settings?

eng/common/scripts/Package-Properties.ps1
weshaggard: I think I found your python version https://github.com/Azure/azure-sdk-for-python/commit/5cd960a276e2056c7215ac5a831c7177fc9a88e3

I guess I had a solution in my head where you would call `Get-AllPkgProperties` then use the file diff list to find all the packages that have files under their package directory and that would be the set of package infos we would write out and use for any other purposes.

eng/common/scripts/Save-Package-Properties.ps1
scbedd: I could! I'll note that you explicitly asked me to use the Save-Package-Properties to filter.

This is what it would look like [to just save package properties for the affected services](https://github.com/azure/azure-sdk-for-python/compare/2ca0b6faf470130a92df6b31b431c31c10d4dfd0...06df40dbd4c2269368382ed79e16a26c436725dd), without filtering in Save-Package-Properties. It's a fairly straightforward operation to simply run Save-Package-Properties for each affected service, and then filter to the affected packages. [This is what the current PR looks like in practice](https://github.com/Azure/azure-sdk-for-python/compare/main...pipelinev3)

eng/common/scripts/Package-Properties.ps1
scbedd: [Would look like this](https://github.com/Azure/azure-sdk-for-python/blob/6de9adc570ead012a3c81f67e2524e24893396b9/eng/scripts/Language-Settings.ps1#L37)

eng/common/scripts/Save-Package-Properties.ps1
weshaggard: I was thinking something like:

```
$packagesWithChanges = @()

foreach($pkg in $allPackageProperties)
{
  foreach($file in $allFilesInDiff)
  {
    # would need to normalized the paths to verify it is under
    if ($file is in exists under $pkg.DirectoryPath) {
      $packagesWithChanges += $pkg
    }
  }
}
```

Then we write out all $packagesWithChanges

scbedd: This is extremely close what I already do frankly. Get-PRPackageProperties avoids going after the _entire_ repo worth of metadata. For python that saves ~30 seconds worth of parsing of all the package info.

scbedd: If that's all you've got an issue with, then I'll just do that and move on tho.

weshaggard: My main goal is to ensure we only write out package info files for the packages we are running a testing as there will be a number of steps that depend on only having the exact set. Secondly, we do have the `Get-AllPkgProperties` implemented in all the languages already so we should be able to use that and implement this filtering in one place instead of having to implement it in each language independently.

scbedd: Alrighty. Did exactly this. [Working properly](https://dev.azure.com/azure-sdk/public/_build/results?buildId=3963068&view=logs&j=8e5196d1-9c2a-56bb-41fd-e329e85c48fa&t=6119a4c1-14d2-5fe9-3d3b-b95005807b68).

I'll deal with any fallout from `ServiceInput` vs `ServiceDirectory` and get this merged.

eng/common/scripts/Save-Package-Properties.ps1
```diff
@@ -33,7 +33,7 @@ Verison property in that file.
 [CmdletBinding()]
 Param (
   [Parameter(Mandatory=$True)]
-  [string] $serviceDirectory,
+  [string] $serviceInput,
```
weshaggard, line 36: It might be worth just keeping serviceDirectory and making it optional and adding a new parameter for the json file so we don't have this being a magic overloaded string.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -105,6 +105,38 @@ function Get-PkgProperties
...
+
+    $allPackageProperties = Get-AllPkgProperties
+    $diff = Get-Content $InputDiffJson | ConvertFrom-Json
+    $targetedFiles = $diff.ChangedFiles
+
+    foreach($pkg in $allPackageProperties)
+    {
+        foreach($file in $targetedFiles)
+        {
+            $pkgFolder = Get-PackageFolderFromPath($file)
```
weshaggard, line 130: Should we just check to see if the file is under $pkg.DirectoryPath?

eng/common/scripts/Package-Properties.ps1
```diff
@@ -105,6 +105,38 @@ function Get-PkgProperties
     return $null
 }
 
+function Get-PackageFolderFromPath($file) {
+    $parts = $file -split "/"
+
+    if ($parts.Length -lt 3) {
+        return ""
+    }
+    return [System.IO.Path]::Combine($parts[0], $parts[1], $parts[2])
```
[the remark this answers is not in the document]
scbedd: By checking it for lengths shorter than 3 and ensuring that it doesn't error.

eng/common/scripts/Package-Properties.ps1
scbedd: Ok.

eng/common/scripts/Save-Package-Properties.ps1
scbedd: Ok.

eng/common/scripts/Package-Properties.ps1
scbedd: I used `-like` with a path resolution to ensure that the path separators would be consistent.

I could use the same, get the string, and `StartsWith`, but I didn't see that it made much of a difference.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -105,6 +105,28 @@ function Get-PkgProperties
...
+    $allPackageProperties = Get-AllPkgProperties -ServiceDirectory media
+    $diff = Get-Content $InputDiffJson | ConvertFrom-Json
+    $targetedFiles = $diff.ChangedFiles
+
+    foreach($pkg in $allPackageProperties)
+    {
+        foreach($file in $targetedFiles)
+        {
+            $filePath = Resolve-Path (Join-Path $RepoRoot $file)
+            $shouldInclude = $filePath -like "$($pkg.DirectoryPath)*"
```
weshaggard, line 120: This should work as long as pkg.DirectoryPath is also a Resolved-Path, which I'm not sure but we can resolve it if needed.

eng/common/scripts/Save-Package-Properties.ps1
```diff
@@ -137,6 +149,6 @@ if ($allPackageProperties)
 }
 else
 {
-    Write-Error "Package properties are not available for service directory $($serviceDirectory)"
+    Write-Error "Package properties are not available for service directory $($serviceInput)"
```
weshaggard, line 152:
```suggestion
    Write-Error "Package properties are not available for service directory $serviceDirectory or $prdiff"
```

weshaggard (approved):

eng/common/scripts/Package-Properties.ps1
scbedd: You're right. I'll resolve it here just to be certain.

PR #8610, opened 2024-07-11
[tsp-client] Add autorest and plugins as dependencies

mikeharder:
- Guarantees versions of autorest and plugins are pinned after tsp-client is installed
- Allows consumers to use standard npm commands to manage autorest and plugin versions


tools/tsp-client/package.json
```diff
@@ -46,7 +46,10 @@
     "typescript": "^5.1.6"
   },
   "dependencies": {
+    "@autorest/openapi-to-typespec": "0.8.2",
```
pshao25, line 49: Can we float it or make it peer dependency? Otherwise we have to update this package every time we release a new version of @autorest/openapi-to-typespec.

joheredi: I'm not sure we want to float this version, especially since we don't have a stable release for the extension. If we ever get a stable release we may want to discuss, but now I think we should pin

mikeharder: I believe the relationship between `tsp-client`, and `autorest` and `@autorest/openapi-to-typespec` is a full dependency, not a peerDependency.  PeerDependency should only be used when one package is a plugin to another package.  So `@autorest/openapi-to-typespec` should probably have `autorest` as a peerDependency, but this is unrelated to `tsp-client`.

Since `tsp-client` calls `autorest` and `@autorest/openapi-to-typespec`, they are just normal runtime dependencies.

We can only start floating the dependency, once `@autorest/openapi-to-typepsec` releases a `1.0.0` and agrees to follow semver (major for breaking changes, minor for new features, patch for bug fixes).  Since semver allows breaking changes between any versions < 1.0.0, the only responsible way to use packages < 1.0.0, is to pin the version.

tools/tsp-client/src/npm.ts
```diff
@@ -59,6 +59,27 @@ export async function npxCommand(workingDir: string, args: string[]): Promise<vo
   });
 }
 
+export async function nodeCommand(workingDir: string, args: string[]): Promise<void> {
+  console.log(args);
+  return new Promise((resolve, reject) => {
+    const npm = spawn("node", args, {
```
joheredi, line 47: nit: maybe we call the const `node`?

joheredi (approved):

tools/tsp-client/package.json
joheredi: Even when we release `1.0.0` there may be changes in `1.1.0` that wouldn't be considered breaking change, like different formatting or in the order models are generated. Even though these are not breaking changes, they will introduce friction and impact repeatability. So even then I'm not sure if we'd want to float. Anyway, this is a conversation for another day.

catalinaperalta (approved):
Can we also update the changelog to mention that we are pinning these dependencies now?

tools/tsp-client/src/npm.ts
mikeharder: Thanks, was copy/paste bug

mikeharder:
> Can we also update the changelog to mention that we are pinning these dependencies now?

I will do this before merging

tools/tsp-client/package.json
pshao25: I see. It makes sense, just feel something wrong that every time we want to release a new version of our tool, we need to release twice! One for `@autorest/openapi-to-typespec` and the other for `tsp-client`.

mikeharder: @pshao25: It's a tradeoff between reliability and agility.  Pinning versions is more reliable but less agile (may need to release multiple packages).

A middle ground, once 1.0.0 is published, would be to float on patches only (`~1.0.0`).  This, way, you only need to update `tsp-client` for major or minor upgrades, not patch.

PR #8613, opened 2024-07-12
[tsp-client] Clean up and add tests

catalinaperalta:
- Clean up old tests
- Add an example sdk for testing tsp-client commands. Generating the contosowidgetmanager project with the typespec-ts emitter. The emitter language is not relevant for testing commands so testing with JS/TS since it seems to be one of the faster emitters.
- Add script to run tests
- Update package.json example script

Future PR will add more mocha tests with the commands and various config flags.


tools/tsp-client/src/network.ts
```diff
@@ -10,6 +10,8 @@ export function isValidUrl(url: string) {
 }
 
 // Checks if a file exists locally
-export function doesFileExist(path: string): Promise<boolean> {
-  return stat(path).then(() => true).catch(() => false);
+export async function doesFileExist(path: string): Promise<boolean> {
+  return stat(path)
+    .then(() => true)
+    .catch(() => false);
```
joheredi, line 16: Here you could use async/await. It's more code but more modern JS

```ts
try {
   await stat(path)
   return true;
} catch {
   return false
}
```

joheredi (approved):

tools/tsp-client/src/network.ts
catalinaperalta: Using await in the latest commit on this branch

PR #8623, opened 2024-07-13
Refactor PipelineWitness

hallipr:
- Make class names and namespace separation more consistent between AzurePipeline and GitHubActions
- Use GeneratedRegex attribute for all regex
- Prepare for addition of missing blob workers
  - Move common logic for distributed, peroidic background services to new abstract class
  - Nest config settings for background workers


hallipr:
This is to reduce the amount of refactoring in #8424

weshaggard (approved):

PR #8664, opened 2024-07-19
fix the using issue in dotnet examples

ArcturusZhang:
Because of a change in dotnet generator, now the dotnet generated code will not include using statements for those namespaces that are parent of the current namespace.
For example, if our current namespace is `Azure.ResourceManager.Compute.Samples`:
```csharp
namespace Azure.ResourceManager.Compute.Samples;
```
the `using Azure;`, `using Azure.ResourceManager;`, `using Azure.ResourceManager.Compute;` statements will not be generated into the sample file, even if members from those namespaces are used below.

This is fine when we are building the sample projects, but it is not fine if we extract the sample code into a standalone program which is under a different namespace.

This PR fixes this by adding the namespace trimming the `.Samples` suffix, and always add the namespace of `Azure` and `Azure.ResourceManager` because they are so common that every sample program should need them.


tools/azure-rest-api-specs-examples-automation/dotnet/test_main.py
```diff
@@ -1,6 +1,6 @@
 import unittest
-from main import break_down_aggregated_dotnet_example, format_dotnet
-
+import parameterized
```
ArcturusZhang, line 2: I added a `parameterized` package, and on my local, I have to `pip install parameterized`, do I need to do something here so that the pipeline could also have it?

weidongxu-microsoft: You can run the pip install in a step in CI

https://github.com/Azure/azure-sdk-tools/blob/a3d8c574bb81c6a99b6690d9562f95c47badc6d2/tools/azure-rest-api-specs-examples-automation/ci.yml#L38-L40

weidongxu-microsoft (approved):

Issue #8667, opened 2024-07-19
Show underlying type for type aliases


chlowell:
For example, [azblob/appendblob.ExpiryTypeAbsolute](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob@v1.4.0/appendblob#ExpiryTypeAbsolute) is an alias for an [internal type](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob@v1.4.0/internal/exported#ExpiryTypeAbsolute) having underlying type `time.Time`. The `appendblob` review should indicate this with an entry like `type ExpiryTypeAbsolute time.Time`

Issue #7866, opened 2024-03-12
apiviewgo resolves type aliases referring to external repos


chlowell:
Moving type definitions from an SDK module into another repo would create indirection apiviewgo currently can't follow. For example, we could end up with a chain of aliases like `azcore.AccessToken` -> `azcore/internal/exported.AccessToken` -> `github.com/foo/bar.AccessToken` -> `github.com/foo/bar/internal.AccessToken`. We would want an API review for azcore to include the definition of AccessToken because that's part of azcore's public API, however apiviewgo can't do this today because it doesn't follow references to packages outside the SDK repo. Displaying the definition of `bar.AccessToken` as `azcore.AccessToken` would require apiviewgo to do something like:

1. recognize the type alias as referring to an external module
     - probably want an allow list of external modules whose definitions we want to include
2. acquire that module or find it somewhere on disk (pointed to by a command line arg?)
    - note today our automation for generating API reviews provides only the SDK repo
    - an arg like `--include-module` could point at external modules and allow-list them for indexing
3. index the external module
1. add the type definition to the SDK module's API

Issue #8654, opened 2024-07-18
apiviewgo supports tree tokens


chlowell:
Part of https://github.com/Azure/azure-sdk-tools/issues/5897. Probably entails nontrivial refactoring due to the coupling of `Pkg`, `content`, and the flat token schema.

PR #8668, opened 2024-07-19
Refactor apiviewgo type definition hoisting

chlowell:
Part of #7866. Main goal here is to create a spot (the TODO in `Review.resolveAliases`) to handle type aliases (`type Event = log.Event`) referring to non-SDK modules.

- new `TypeAlias` type carries data about aliases, hoists the definition to the aliasing package when given a `typeDef`
- new `Review` type collects modules implicated in the review, resolves cross-module references, computes output
- `Module` now surfaces its cross-module type aliases for `Review` to resolve instead of resolving them itself
- removed assumption that all modules implicated in a review are in the same repo

Along the way I appear to have incidentally fixed a bug that omitted type aliases referring to underlying types. For example, [ExpiryTypeAbsolute](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob@v1.4.0/appendblob#ExpiryTypeAbsolute) doesn't appear in the latest [azblob review](https://apiview.dev/Assemblies/Review/e3f09c889fc54009a2ef911d083f8435). With this PR, it will appear without a hoisted definition; #8667 tracks improving that.

### DIFF
--- a/src/go/cmd/api_view_test.go
+++ b/src/go/cmd/api_view_test.go
@@ -163,10 +163,6 @@ func TestDiagnostics(t *testing.T) {
 }
 
 func TestAliasDefinitions(t *testing.T) {
-	priorValue := sdkDirName
-	sdkDirName = "testdata"
-	defer func() { sdkDirName = priorValue }()
-
 	for _, test := range []struct {
 		name, path, sourceName string
 		diagLevel              DiagnosticLevel
@@ -175,7 +171,7 @@ func TestAliasDefinitions(t *testing.T) {
 			diagLevel:  DiagnosticLevelWarning,
 			name:       "service_group",
 			path:       "testdata/test_service_group/group/test_alias_export",
-			sourceName: "github.com/Azure/azure-sdk-for-go/sdk/test_service_group/group/internal.Foo",
+			sourceName: "github.com/Azure/azure-sdk-tools/src/go/cmd/testdata/test_service_group/group/internal.Foo",
 		},
 		{
 			diagLevel:  DiagnosticLevelInfo,
@@ -187,11 +183,13 @@ func TestAliasDefinitions(t *testing.T) {
 			diagLevel:  DiagnosticLevelWarning,
 			name:       "external_package",
 			path:       "testdata/test_external_alias_exporter",
-			sourceName: "github.com/Azure/azure-sdk-for-go/sdk/test_external_alias_source.Foo",
+			sourceName: "github.com/Azure/azure-sdk-tools/src/go/cmd/testdata/test_external_alias_source.Foo",
 		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
-			review, err := createReview(filepath.Clean(test.path))
+			p, err := filepath.Abs(test.path)
+			require.NoError(t, err)
+			review, err := createReview(p)
 			require.NoError(t, err)
 			require.Equal(t, "Go", review.Language)
 			require.Equal(t, 1, len(review.Diagnostics))
@@ -210,10 +208,6 @@ func TestAliasDefinitions(t *testing.T) {
 }
 
 func TestRecursiveAliasDefinitions(t *testing.T) {
-	priorValue := sdkDirName
-	sdkDirName = "testdata"
-	defer func() { sdkDirName = priorValue }()
-
 	for _, test := range []struct {
 		name, path, sourceName string
 		diagLevel              DiagnosticLevel
--- a/src/go/cmd/api_view_test.go
+++ b/src/go/cmd/api_view_test.go
@@ -286,12 +286,12 @@ func TestVars(t *testing.T) {
 	require.True(t, hasHTTPClient)
 }
 
-func Test_getPackageNameFromModPath(t *testing.T) {
-	require.EqualValues(t, "foo", getPackageNameFromModPath("foo"))
-	require.EqualValues(t, "foo", getPackageNameFromModPath("foo/v2"))
-	require.EqualValues(t, "sdk/foo", getPackageNameFromModPath("github.com/Azure/azure-sdk-for-go/sdk/foo"))
-	require.EqualValues(t, "sdk/foo/bar", getPackageNameFromModPath("github.com/Azure/azure-sdk-for-go/sdk/foo/bar"))
-	require.EqualValues(t, "sdk/foo/bar", getPackageNameFromModPath("github.com/Azure/azure-sdk-for-go/sdk/foo/bar/v5"))
+func TestGetReviewNameFromModPath(t *testing.T) {
+	require.EqualValues(t, "foo", getReviewNameFromModPath("foo"))
+	require.EqualValues(t, "foo", getReviewNameFromModPath("foo/v2"))
+	require.EqualValues(t, "sdk/foo", getReviewNameFromModPath("github.com/Azure/azure-sdk-for-go/sdk/foo"))
+	require.EqualValues(t, "sdk/foo/bar", getReviewNameFromModPath("github.com/Azure/azure-sdk-for-go/sdk/foo/bar"))
+	require.EqualValues(t, "sdk/foo/bar", getReviewNameFromModPath("github.com/Azure/azure-sdk-for-go/sdk/foo/bar/v5"))
 }
 
 func TestDeterministicOutput(t *testing.T) {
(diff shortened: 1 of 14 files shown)


src/go/cmd/module.go
```diff
@@ -279,42 +180,33 @@ func parseModFile(dir string) (*modfile.File, error) {
...
-		for _, pkg := range packages {
-			if pkg.p.Name == pkgName {
-				source = pkg
+	// source doesn't define typeName; it must export typeName via an alias.
+	// Recurse into the package from which source imports typeName.
+	for _, a := range source.TypeAliases {
+		if a.Name == typeName {
+			// a.QualifiedName == github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container.DeleteOptions
+			dot := strings.LastIndex(a.QualifiedName, ".")
+			if dot < 0 {
```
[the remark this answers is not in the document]
RickWinter: We should also error out if `.` is the last character found in a qualifiedname.. Though maybe you have protection for that elsewhere. However if we ever did hit it, then you'd hit an index out of range on line 204

src/go/cmd/pkg.go
```diff
@@ -294,6 +298,97 @@ func (pkg Pkg) addTypeNavigator(oriVal string, imports map[string]string) string
...
+					Level:    DiagnosticLevelError,
+					TargetID: t.ID(),
+					Text:     missingAliasFor + fieldTypeName,
+				})
+			}
+		case *ast.Ident:
+			t = a.Package.c.addSimpleType(*a.Package, a.Name, a.Package.Name(), def.n.Type.(*ast.Ident).Name, nil)
+			hoistMethodsForType(def.p, a.Name, a.Package)
+		default:
+			fmt.Printf("unexpected node type %T\n", def.n.Type)
```
RickWinter, line 377: ErrorF

src/go/cmd/review.go
```diff
@@ -0,0 +1,202 @@
...
+	// modules maps module paths to Modules implicated in this API review. It
+	// contains only the reviewed module in most cases, however it contains
+	// more when the reviewed module exports types defined in another module.
+	modules map[string]*Module
+	// name of the APIView review e.g. "sdk/azcore"
+	name string
+	// path on disk to the reviewed module
+	path string
+	// reviewed is the module being reviewed
+	reviewed *Module
```
RickWinter, line 30: Does it make sense to store this pointer twice? Its also in the map.

src/go/cmd/review.go
```diff
@@ -0,0 +1,202 @@
...
+	diagnostics := []Diagnostic{}
+	packageNames := []string{}
+	for name, p := range r.reviewed.Packages {
+		// we use a prefixed path separator so that we can handle the "internal" module.
+		//  internal/dig
+		//  internal/errorinfo
+		//  etc.
+		// for other modules, we skip /internal subdirectories
+		//  azcore/internal/...
+		if strings.Contains(p.relName, "/internal") || p.c.isEmpty() {
```
RickWinter, line 75: Do we pop as error if a module is exporting a type from its internal package. I would consider it a bug if azure exported a type from azcore/internal

src/go/cmd/review.go
```diff
@@ -0,0 +1,202 @@
...
+	sort.Strings(packageNames)
+	for _, name := range packageNames {
+		p := r.reviewed.Packages[name]
+		n := p.relName
+		makeToken(nil, nil, "package", TokenTypeMemberName, tokenList)
+		makeToken(nil, nil, " ", TokenTypeWhitespace, tokenList)
+		makeToken(&n, nil, n, TokenTypeTypeName, tokenList)
+		makeToken(nil, nil, "", TokenTypeNewline, tokenList)
+		makeToken(nil, nil, "", TokenTypeNewline, tokenList)
+		// TODO: reordering these calls reorders APIView output and can omit content
```
RickWinter, line 89: :O

RickWinter (approved):

src/go/cmd/review.go
jhendrixMSFT: We make heavy use of this pattern to avoid circular references.

src/go/cmd/review.go
chlowell: 😇 it was like this when I got here. I'm moving this code but the problem has existed since the beginning. It could get fixed as part of #8654

src/go/cmd/api_view_test.go
```diff
@@ -187,11 +183,13 @@ func TestAliasDefinitions(t *testing.T) {
...
 			name:       "external_package",
 			path:       "testdata/test_external_alias_exporter",
-			sourceName: "github.com/Azure/azure-sdk-for-go/sdk/test_external_alias_source.Foo",
+			sourceName: "github.com/Azure/azure-sdk-tools/src/go/cmd/testdata/test_external_alias_source.Foo",
 		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
-			review, err := createReview(filepath.Clean(test.path))
+			p, err := filepath.Abs(test.path)
+			require.NoError(t, err)
```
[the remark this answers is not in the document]
chlowell: `Abs` calls `Clean`. We need an absolute path in this test because I removed the hook we formerly used to manipulate the code that finds local modules. That code now works in general cases--it doesn't assume it's always looking for SDK modules--so it doesn't require fakery and using an absolute path here tests that.

src/go/cmd/pkg.go
chlowell: This is another chunk I moved without changing. We have this case because the switch isn't exhaustive. The way it works now is when we resolve a type alias to a definition using syntax we don't explicitly handle, we give up and add the alias to the API review as-is. Returning an error instead would make the module unreviewable. Not necessarily a bad idea but we should add at least the case we need today before making the change. I opened #8670 to track that.

src/go/cmd/review.go
RickWinter: Ah ok, do we enforce that in the APIView tool. We should. Also we should add that to the GO design guidelines.

src/go/cmd/review.go
chlowell: We need either this pointer or a key for the map because iteration order over maps is undefined. The choice is pretty arbitrary but the pointer is more to the point and smaller.

src/go/cmd/module.go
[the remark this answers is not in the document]
chlowell: `0 < dot <= len(a.QualifiedName)-2` should always be true and is in fact true for the entire SDK today. I'm uncertain it's worth reworking this function to return an error when this condition doesn't hold because that would entail dropping the `, ok` idiom in favor of returning an error and we'd need a sentinel error for "definition not found". Seems like too much ceremony to catch an improbable bug. I suppose we could panic instead; that condition really should always be true.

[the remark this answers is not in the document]
jhendrixMSFT: Agreed that a `panic` for something that should never happen is preferred.

src/go/cmd/review.go
jhendrixMSFT: We surface it, but don't enforce it in APIView, and I don't think that it's the correct tool to do so (AFAIK there's no way to fail the APIView).

RickWinter: APIView can show errors/warnings, we do it in the C++ APIView.

jhendrixMSFT: Interesting. How does that impact the APIView, i.e. is it just some markup or does it prevent the APIView from being approved?

RickWinter: Its just markup. APIViews simplify show diffs, notes, comments, ... At the end of the day the architects can click Approve whenever they want

chlowell: We do the same for every type alias in Go, and the messages are just as illegible in our reviews 🙈

RickWinter: Perfect we can leverage the pattern and the code. Have it show a message that the `type` from internal is being exported. 
   "Type `x` is exported from internal. Exported types from internal must be in the ...../internal/exported module"

jhendrixMSFT (approved):

src/go/cmd/review.go
chlowell: I opened #8696. We don't consistently use this pattern today, so step 1 of the issue is determining whether we can.

Issue #8698, opened 2024-07-24
apiviewgo considers cross-module alias versions within SDK


chlowell:
For example, `azcore` exports types defined in `internal`. Today, when parsing `azcore`, apiviewgo parses the source code of `internal` it finds on disk next to `azcore`. This is incorrect because it ignores the `internal` version `azcore` requires--apiviewgo just parses whatever code is on the checked-out branch and can include irrelevant, pre-release `internal` API in `azcore` reviews. apiviewgo should instead parse the `internal` version specified in `azcore/go.mod`.

PR #8699, opened 2024-07-24
apiviewgo parses non-SDK modules as needed

### LINKED ISSUE #7866: apiviewgo resolves type aliases referring to external repos

chlowell:
When apiviewgo finds a type alias referencing a non-SDK module, it will parse that module's source from the local mod cache or download it from the Go module proxy, and hoist the definition to the API it's reviewing. Parsing and definition-hoisting logic is unchanged; this PR just adds the ability for apiviewgo to parse code not residing in the SDK repo.

Closes #7866


src/go/cmd/cache.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil, fmt.Errorf("failed to unzip module: %w", err)
+	}
+	if _, err := os.Stat(filepath.Join(d, "go.mod")); err != nil {
+		return nil, fmt.Errorf("go.mod not found in %s", d)
+	}
+	return NewModule(d)
+}
+
+// the apiviewgo cache looks like:
+//     /tmp/apiviewgo/cache
```
RickWinter, line 136: Is this temp folder version resilient? What happens if two reviews are being generated, will they stomp on each other to create the cache?

Should we make the cache tied to the review?

RickWinter:
The linked issue talks about an allow list. I don't see that, are we not doing that? How do we prevent exporting types from libraries we don't own?

src/go/cmd/cache.go
chlowell: Good question. I wrote this on the assumption that apiviewgo runs sequentially but not concurrently. I don't actually know the architecture of our automation though, so I'll have to ask someone about the details. Concurrent processes would race for the same module version, possibly causing one to fail or produce garbage due to reading torn files.

chlowell:
I decided against the allow list because I didn't see a benefit worth the friction. The review process from PR to apiview prevents exporting types we don't own; it's clear at each stage where a type comes from. It does rely totally on human judgment though. Do you think we need a hard barrier?

RickWinter:
> I decided against the allow list because I didn't see a benefit worth the friction. The review process from PR to apiview prevents exporting types we don't own; it's clear at each stage where a type comes from. It does rely totally on human judgment though. Do you think we need a hard barrier?

It should be fine to not have it since its clear in the Apiview where the types are coming from.

src/go/cmd/cache.go
jhendrixMSFT: https://pkg.go.dev/os#MkdirTemp will help you here.

jhendrixMSFT (approved):

src/go/cmd/cache.go
chlowell: `MkdirTemp` would make the cache ephemeral. I was reluctant to do that because I imagine this mechanism becoming the solution for #8698, however it's a small efficiency given we have 3 cross-module aliases in the SDK today. So, I'll make the change; we can revisit the synchronization question later, if durable caching becomes interesting.

PR #8709, opened 2024-07-26
[test-proxy] Only remove session once body has been written

mikeharder:
- Minimizes window of race condition, where client retries after test-proxy has already removed session
- Prevents needless `lock`s in `HandlePlaybackRequest`


tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -497,6 +488,23 @@ public async Task HandlePlaybackRequest(string recordingId, HttpRequest incoming
...
+            // Used by perf tests to play back the same request multiple times.
+            if (incomingRequest.Headers.TryGetValue("x-recording-remove", out var removeHeader))
+            {
+                remove = bool.Parse(removeHeader);
+            }
+
+            // Only remove session once body has been written, to minimize probability client retries but test-proxy has already removed the session
+            if (remove)
+            {
+                session.Session.Lookup(entry, session.CustomMatcher ?? Matcher, session.AdditionalSanitizers.Count > 0 ? Sanitizers.Concat(session.AdditionalSanitizers) : Sanitizers, remove: true);
```
mikeharder, line 506: Might be able to improve perf, by adding something like a `Remove()` method to `Session`, which safely removes a session (under locks), but doesn't need to re-run the lookup and sanitizer code.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -512,6 +503,23 @@ public async Task HandlePlaybackRequest(string recordingId, HttpRequest incoming
...
+            // If request contains "x-recording-remove: false", then request is not removed from session after playback.
+            // Used by perf tests to play back the same request multiple times.
+            if (incomingRequest.Headers.TryGetValue("x-recording-remove", out var removeHeader))
+            {
+                remove = bool.Parse(removeHeader);
+            }
+
+            if (remove)
+            {
+                session.Session.Lookup(entry, session.CustomMatcher ?? Matcher, sanitizers, remove: true);
```
mikeharder, line 521: Might be able to improve perf, by adding something like `Session.Remove()`, which safely removes a session (under locks), but doesn't need to re-run the lookup and sanitizer code.

scbedd: 👍 

I'm going to do that and publish this as a new proxy version to try reproing the storage issues on. I'm a bit concerned that the double sanitization won't be as stable as we expect (we sanitize the requestentry in `Lookup`), so I'm going to go after that Remove() separately.

mikeharder: I agree adding a separate `Remove()` is the correct design, this was more a quick hack to show the direction I was thinking.

benbp (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -103,6 +103,14 @@ public RecordEntry Lookup(RecordEntry requestEntry, RecordMatcher matcher, IEnum
             }
         }
 
+        public void Remove(RecordEntry entry)
+        {
+            lock (Entries)
```
mikeharder, line 108: If it would improve perf, you might be able to avoid locking by changing `Entries` from `List<RecordEntry>` to `ConcurrentQueue<RecordEntry>`, which would allow you to enqueue/peek/dequeue safely without locking.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -457,16 +457,19 @@ public async Task HandlePlaybackRequest(string recordingId, HttpRequest incoming
 
             var sanitizers = SanitizerRegistry.GetSanitizers(session);
 
-            // we don't need to re-sanitize with recording-applicable sanitizers every time. just the very first one
-            lock (session)
+            if (!session.IsSanitized)
```
mikeharder, line 460: This is probably safe enough for our purposes, but if we see bugs, or are concerned the memory model may not guarantee safeness, we could replace with `Lazy<T>`.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
scbedd: > If it would improve perf, you might be able to avoid locking by changing `Entries` from `List<RecordEntry>` to `ConcurrentQueue<RecordEntry>`, which would allow you to enqueue/peek/dequeue safely without locking.

I did look at doing this earlier in the lifecycle of this thing. I ran into really wonky matching errors that had to do with needing to effectively "recycle" the queue. Match operations from Task A would _fail_ because an item was popped off the concurrent queue from another lookup operation running from Task B.

The object is synced between the two threads, but one looked in the split second before an item was added back to the queue.

That's why I left it at `lock` originally. I entirely agree that perf wise it won't be as fast (not even close) as a ConcurrentQueue.

mikeharder: There's probably a way to make ConcurrentQueue work reliably, but this might not even be a perf bottleneck at all.

PR #8727, opened 2024-07-30
Update Test Proxy

scbedd:
Includes:

- #8681 
- #8709 

This won't be a magic bullet on what java is seeing, but I can't imagine it's not an improvement.


benbp (approved):

scbedd:
Closing. the ios PR merged and I need to make changes to the test proxy itself, so I need to abandon and get a new common PR going.

PR #8730, opened 2024-07-31
Adding swagger sorting command to tcp-client

allenjzhang:


timotheeguerin (approved):

timotheeguerin:
Be good to see what @catalinaperalta thinks of that first too

tools/tsp-client/README.md
```diff
@@ -44,6 +44,8 @@ Generate a client library from a TypeSpec project. The `generate` command should
 
 ### convert
 Convert an existing swagger specification to a TypeSpec project. This command should only be run once to get started working on a TypeSpec project. TypeSpec projects will need to be optimized manually and fully reviewed after conversion. When using this command a path or url to a swagger README file is required through the `--swagger-readme` flag.
+### sortSwagger
```
catalinaperalta, line 47: Nit: needs a space
```suggestion

### sortSwagger
```

tools/tsp-client/package.json
```diff
@@ -38,7 +38,8 @@
     "@types/mocha": "^10.0.1",
     "@types/node": "^20.4.8",
     "@types/prompt-sync": "^4.2.1",
-    "@typespec/compiler": "0.50.0",
+    "@typespec/compiler": "^0.58.0",
+    "@azure-tools/typespec-autorest": "^0.44.0",
```
catalinaperalta, line 42: Do folks need to use v0.44.0 or later to be able to run this command?

tools/tsp-client/src/index.ts
```diff
@@ -355,6 +362,16 @@ async function main() {
           }
         }
         break;
+    case "sortSwagger":
+        Logger.info("Sorting a swagger content...");
+        let swaggerFile =  options.swaggerFile;
+        if (swaggerFile === undefined || await !doesFileExist(swaggerFile)) {
+            throw new Error(`Swagger file not found: ${swaggerFile??"[Not Specified]"}`);
+        }
+        swaggerFile = normalizePath(resolve(swaggerFile));
```
catalinaperalta, line 371: Should we be doing this before we do the doesFileExist check on line 368?

tools/tsp-client/src/log.ts
```diff
@@ -68,6 +68,7 @@ Commands:
   generate    Generate from a TypeSpec project                          [string]
   update      Sync and generate from a TypeSpec project                 [string]
   convert     Convert a swagger specification to TypeSpec               [string]
+  sortSwagger Sort a swagger specification based on TypeSpec ordering   [string]
```
catalinaperalta, line 71: Can we rename this to `sort-swagger` ?

tools/tsp-client/src/options.ts
```diff
@@ -136,6 +138,15 @@ export async function getOptions(): Promise<Options> {
         process.exit(1);
       }
     }
+
+    if (command === "sortSwagger") {
```
catalinaperalta, line 142: Nit: can we rename to `sort-swagger`?

catalinaperalta (commented):
Can we also add a test for this scenario?

tools/tsp-client/package.json
allenjzhang: sorting API is introduced recently into autorest.

tools/tsp-client/src/index.ts
allenjzhang: Seeing the above code, it was done this way. I am not sure if this does anything as if `doesFileExist` works with the path, `readFile` should work as well. But I can adjust.

tools/tsp-client/package.json
catalinaperalta: Makes sense, then we should also have it under dependencies.

timotheeguerin: Ah yeah should be in dependencies not devDependencies

catalinaperalta: Also, I think we should float peerDependencies for typespec/compiler between 0.58.0 - 1.0.0

allenjzhang: ^0.44.0 is `0.44.0 <= ver < 1.0.0`

timotheeguerin: no for `0.` `^` doesn't allow next minors `^0.44.0` its just `>=0.44.0 && < 0.45.0`.

However in this case its not meant to be used as a peerDependency but as a regular dependency so this is correct.
This dependency can be updated whenever(probably will complain about the version of the compiler if yhou don't update) but in theory it doesn't matter its just loading the function from this lib

catalinaperalta: Also, the comment was pointing out that these deps should be under "dependency"

catalinaperalta: We float typespec/compiler in peerDependencies, which is where I was saying we need to update it to 0.58.0

catalinaperalta: I'll fix the peer deps in a separate pr @allenjzhang. Resolving now

catalinaperalta: https://github.com/Azure/azure-sdk-tools/issues/8766 tracking issue before next tsp-client release

catalinaperalta (approved):

catalinaperalta:
FYI are you aware of this issue: https://github.com/Azure/autorest/issues/4944 ? Should it be closed now or is it a separate effort?

catalinaperalta (approved):

Issue #8729, opened 2024-07-31
Investigate `java storage` failures


scbedd:
@samvaity reported this failing test [getAccountInfoBaseFail](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4009937&view=ms.vss-test-web.build-test-results-tab&runId=49124846&resultId=102203&paneView=debug)

And alan was report an intermittent channel timeout while awaiting a playback start.

Other than `getAcountInfoBaseFail`, the other commonly failing tests are part of the `NioInputStream` tests. [markAndReset](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4009937&view=ms.vss-test-web.build-test-results-tab&runId=49124856&resultId=100653&paneView=debug) is the most common failing niostream test that I can tell.

I believe this is all interrelated. If the proxy has any locking issues I'd expect it to turn up _much_ less consistently than what we're seeing however.

PR #8731, opened 2024-07-31
Updates to `HandlePlaybackRequest` to force synchronous handling

scbedd:
image

We pay next to nothing for stability. The failing test is the one that I think is actually going wrong on Java side.

Relevant to #8729

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
@@ -472,50 +472,53 @@ namespace Azure.Sdk.Tools.TestProxy
 
             var entry = (await CreateEntryAsync(incomingRequest).ConfigureAwait(false)).Item1;
 
-            // Session may be removed later, but only after response has been fully written
-            var match = session.Session.Lookup(entry, session.CustomMatcher ?? Matcher, sanitizers, remove: false, sessionId: recordingId);
-
-            foreach (ResponseTransform transform in Transforms.Concat(session.AdditionalTransforms))
+            lock (session.Session.Entries)
             {
-                transform.Transform(incomingRequest, match);
-            }
+                // Session may be removed later, but only after response has been fully written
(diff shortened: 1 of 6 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -470,50 +472,53 @@ public async Task HandlePlaybackRequest(string recordingId, HttpRequest incoming
 
             var entry = (await CreateEntryAsync(incomingRequest).ConfigureAwait(false)).Item1;
 
-            // Session may be removed later, but only after response has been fully written
-            var match = session.Session.Lookup(entry, session.CustomMatcher ?? Matcher, sanitizers, remove: false);
-
-            foreach (ResponseTransform transform in Transforms.Concat(session.AdditionalTransforms))
+            lock (session.Session.Entries)
```
mikeharder, line 475: Did you try adding this lock, but still writing the body bytes async?  I can see potential race conditions fixed by locking all entries until all bytes have been written, but I don't think sync vs async WriteBodyBytes should matter.

mikeharder (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
semick-dev: As far as I'm aware you can't write await within a `lock`?

Am I crazy? I'll try it!

scbedd: image

mikeharder: It can't be done using the `lock` statement itself, but conceptually it's possible to use locking semantics with async:

https://stackoverflow.com/a/18257065

scbedd: Understood! I've definitely used a SempahoreSlim before, I'll apply that and see if the experience is reproed. if so, I'll keep the async around.

PR #8752, opened 2024-08-01
Add exception for Utf8JsonReader/Writer to AZC0014

pshao25:
We have now enabled [JsonModelWriteCore](https://github.com/microsoft/typespec/blob/745b4f97229e128f61e35b725d4ae3adb436424a/packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/Models/ProjectedNameModelRequest.Serialization.cs#L39) and [JsonModelCreateCore](https://github.com/microsoft/typespec/blob/745b4f97229e128f61e35b725d4ae3adb436424a/packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/Models/ProjectedNameModelRequest.Serialization.cs#L69) in MGC. Now we are adding this WriteCore/CreateCore feature to autorest.csharp, because some customers are asking this. However, these two methods are violating “AZC0014: Types from System.Text.Json, Newtonsoft.Json, System.Collections.Immutable assemblies should not be exposed as part of public API surface. “
 
Confirmed with @KrzysztofCwalina that Utf8JsonReader/Writer should be an exception to AZC0014. You could refer to that decision in the email titled "Suppress AZC0014 on WriteCore and CreateCore".

### DIFF
--- a/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
+++ b/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
@@ -38,7 +38,10 @@ namespace Azure.ClientSdk.Analyzers
                 {
                     if (BannedAssemblies.Contains(type.ContainingAssembly.Name))
                     {
-                        context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0014, symbol.Locations.First(), BannedAssembliesMessageArgs), symbol);
+                        if (type.Name != "Utf8JsonReader" && type.Name != "Utf8JsonWriter")
+                        {
+                            context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0014, symbol.Locations.First(), BannedAssembliesMessageArgs), symbol);
+                        } 
                     }
 
                     if (namedTypeSymbol.IsGenericType)
--- a/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
+++ b/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
@@ -95,9 +95,9 @@ namespace Azure.ClientSdk.Analyzers
             return (type.Name == "Utf8JsonReader" || type.Name == "Utf8JsonWriter") && GetFullNamespace(type.ContainingNamespace) == "System.Text.Json";
         }
 
-        private static string GetFullNamespace(INamespaceSymbol type)
+        private static string GetFullNamespace(INamespaceSymbol namespaceSymbol)
         {
-            return type.ContainingNamespace.Name == "" ? type.Name : $"{GetFullNamespace(type.ContainingNamespace)}.{type.Name}";
+            return namespaceSymbol.ContainingNamespace.Name == "" ? namespaceSymbol.Name : $"{GetFullNamespace(namespaceSymbol.ContainingNamespace)}.{namespaceSymbol.Name}";
         }
     }
 }
--- a/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
+++ b/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
@@ -38,7 +38,7 @@ namespace Azure.ClientSdk.Analyzers
                 {
                     if (BannedAssemblies.Contains(type.ContainingAssembly.Name))
                     {
-                        if (type.Name != "Utf8JsonReader" && type.Name != "Utf8JsonWriter")
+                        if (!IsUtf8JsonReaderWriter(type))
                         {
                             context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0014, symbol.Locations.First(), BannedAssembliesMessageArgs), symbol);
                         } 
@@ -88,6 +88,16 @@ namespace Azure.ClientSdk.Analyzers
                     }
                     break;
             }
+        }
+
+        private static bool IsUtf8JsonReaderWriter(ITypeSymbol type)
+        {
+            return (type.Name == "Utf8JsonReader" || type.Name == "Utf8JsonWriter") && GetFullNamespace(type.ContainingNamespace) == "System.Text.Json";
+        }
+
+        private static string GetFullNamespace(INamespaceSymbol type)
+        {
+            return type.ContainingNamespace.Name == "" ? type.Name : $"{GetFullNamespace(type.ContainingNamespace)}.{type.Name}";
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
```diff
@@ -38,7 +38,10 @@ static void CheckType(ISymbolAnalysisContext context, ITypeSymbol type, ISymbol
                 {
                     if (BannedAssemblies.Contains(type.ContainingAssembly.Name))
                     {
-                        context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0014, symbol.Locations.First(), BannedAssembliesMessageArgs), symbol);
+                        if (type.Name != "Utf8JsonReader" && type.Name != "Utf8JsonWriter")
+                        {
+                            context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0014, symbol.Locations.First(), BannedAssembliesMessageArgs), symbol);
+                        }
```
ArcturusZhang, line 44: this change reads like we are excluding the `Utf8JsonReader` and `Utf8JsonWriter` from the list of triggering this rule.
Is this expected?
I think we still need to report something when these two classes are used in public APIs, but not for those MRW related methods, such as `JsonModelWriteCore` and `JsonModelCreateCore`, but I did not see these two strings in your implementation.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
```diff
@@ -38,7 +38,10 @@ static void CheckType(ISymbolAnalysisContext context, ITypeSymbol type, ISymbol
                 {
                     if (BannedAssemblies.Contains(type.ContainingAssembly.Name))
                     {
-                        context.ReportDiagnostic(Diagnostic.Create(Descriptors.AZC0014, symbol.Locations.First(), BannedAssembliesMessageArgs), symbol);
+                        if (type.Name != "Utf8JsonReader" && type.Name != "Utf8JsonWriter")
```
KrzysztofCwalina, line 41: is there a way to compare fully qualified names (at least with the namespace), not just type names?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
```diff
@@ -86,5 +89,15 @@ static void CheckType(ISymbolAnalysisContext context, ITypeSymbol type, ISymbol
                     break;
             }
         }
+
+        private static bool IsUtf8JsonReaderWriter(ITypeSymbol type)
+        {
+            return (type.Name == "Utf8JsonReader" || type.Name == "Utf8JsonWriter") && GetFullNamespace(type.ContainingNamespace) == "System.Text.Json";
```
KrzysztofCwalina, line 95: I am thinking there must be a way to get symbols for the reader and writer and then write this check as if (type == readerSymbol || type == writerSymbol).

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
```diff
@@ -86,5 +89,15 @@ static void CheckType(ISymbolAnalysisContext context, ITypeSymbol type, ISymbol
...
+        }
+
+        private static string GetFullNamespace(INamespaceSymbol type)
+        {
+            return type.ContainingNamespace.Name == "" ? type.Name : $"{GetFullNamespace(type.ContainingNamespace)}.{type.Name}";
```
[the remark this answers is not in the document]
pshao25: No, just a bad name. Changed to `namespaceSymbol`.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
```diff
@@ -86,5 +89,15 @@ static void CheckType(ISymbolAnalysisContext context, ITypeSymbol type, ISymbol
                     break;
             }
         }
+
+        private static bool IsUtf8JsonReaderWriter(ITypeSymbol type)
+        {
+            return (type.Name == "Utf8JsonReader" || type.Name == "Utf8JsonWriter") && GetFullNamespace(type.ContainingNamespace) == "System.Text.Json";
+        }
+
+        private static string GetFullNamespace(INamespaceSymbol type)
```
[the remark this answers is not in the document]
pshao25: Changed to `namespaceSymbol`.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
pshao25: Isn't the way you mentioned "to get symbols for the reader and writer" what I'm doing now, be checking the name and namespace name to get the symbol?

Do you have a better way?

ArcturusZhang: The biggest issue right now is that how we get the symbol of those type of `Utf8JsonReader` as `INamedTypeSymbol`.

I know a way that if we have a compilation here, we could get the symbol by its full name, this could be a static readonly field, and then we could compare them by
```
SymbolEqualityComparer.Default.Equals(type, readerSymbol)
```

@KrzysztofCwalina do you know other alternatives of how we could fetch the symbol of a known type?

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/BannedAssembliesAnalyzer.cs
```diff
@@ -86,5 +89,15 @@ static void CheckType(ISymbolAnalysisContext context, ITypeSymbol type, ISymbol
...
         }
+
+        private static bool IsUtf8JsonReaderWriter(ITypeSymbol type)
+        {
+            return (type.Name == "Utf8JsonReader" || type.Name == "Utf8JsonWriter") && GetFullNamespace(type.ContainingNamespace) == "System.Text.Json";
+        }
+
+        private static string GetFullNamespace(INamespaceSymbol namespaceSymbol)
+        {
+            return namespaceSymbol.ContainingNamespace.Name == "" ? namespaceSymbol.Name : $"{GetFullNamespace(namespaceSymbol.ContainingNamespace)}.{namespaceSymbol.Name}";
```
KrzysztofCwalina, line 100: what is this line supposed to do? especially "{GetFullNamespace(namespaceSymbol.ContainingNamespace)}.{namespaceSymbol.Name}"? it seems to concatenating two namespace names.

pshao25: Yes, `namespaceSymbol.Name` only contains current name of current namespace. For example, for `System.Text.Json.Utf8JsonReader`, its containing namespace is `Json` whose containing namespace is `Text` and so on. Therefore, I have to have a recursive way to get the full namespace.

KrzysztofCwalina (approved):

PR #8755, opened 2024-08-01
All semaphor all the time

scbedd:
This cuts out all `lock` usage for `await SemaphorSlim.WaitAsync()` to synchronize resource access.

Originally, we had 3 possible locks

- object SessionSanitizerLock
- Session.Entries
- object Session.SanitizerLock

This moves to using a semaphore for these. Unfortunately this has the impact of forcing `async` in a lot of places that _didn't_ use it before, so I've got some test pain to resolve as a result.

@mikeharder I'm very pleased with this one!

image
image


mikeharder (approved):
I agree this is better than synchronous locking.  You generally want to use async as much as possible inside asp.net apps.

However, it is still susceptible to deadlocks and/or race conditions, if we don't have the locking semantics/ordering exactly correct.  So if this doesn't fix all the issues, we might need to take a closer look at the algorithms we are using,.

PR #8760, opened 2024-08-01
Do not mark values as secret that are from git-hosted sub config files

benbp:


eng/common/TestResources/SubConfig-Helpers.ps1
```diff
@@ -82,7 +118,8 @@ function ShouldMarkValueAsSecret([string]$serviceName, [string]$key, [string]$va
         "TenantId",
```
weshaggard, line 118: Isn't the main values we want already in this list?

eng/common/TestResources/build-test-resource-config.yml
```diff
@@ -61,24 +61,11 @@ steps:
         }
       }
 
+      Write-Host "Merging sub config from files"
```
weshaggard, line 64: Should we add other helper functions to clean-up this inline code?

weshaggard (approved):

eng/common/TestResources/build-test-resource-config.yml
benbp: Yeah I was thinking about just eliminating the inline entirely. I'll go ahead and do that.

eng/common/TestResources/SubConfig-Helpers.ps1
benbp: It's possible that in the storage case, the value is getting set from somewhere else to a config key that is not in this exclusion list. Was it the sub or the rg you were seeing starred out?

weshaggard (approved):

PR #8781, opened 2024-08-04
Add proxy audit log and other assorted fixes

scbedd:
This PR

- [x] Fixes an identified issue with not properly utilizing the session locks -- causing a serious perf degradation when adding a bunch of sanitizers to a bunch of tests. (like 10+ minutes per massive core job bad)
- [x] Fixes a serious problem that was allowing a test to be `stopped` while the last entry in the recording is still writing it's body. That location in the code has _already_ obtained a lock on the session, so merely adding the same to `playback/stop` prevents the test from being stopped while the last entry is being streamed back. This was the source of the `Channel Response timeout` issues that `java- core` was seeing @alzimmermsft (at least, it has not reproed in the 3 core runs since I added this fix 🤞 )
- [x] Add a concurrent queue for log messages that can be dumped. Currently this endlessly grows, but pulling with /Audit/GetAuditLogs will clean up that memory.
- [x] Add audit entries for `Record` code paths
- [x] Add a few tests for the new functionality
- [ ] Get signoff from @mikeharder about the always-on per-session audit logging. We were kinda laughing about chewing up memory, but...it's not that bad :P No affect to `java - core` which beats the absolute hell out of the server.

@mikeharder I need to add audit items to the `record` pathway, but I was laser focused on getting `playback` concurrency issues resolved for the java folks. Please take a look, see how I'm retaining the various locks on resources (by reducing to a single access lock to each playback session). I _think_ I've nailed most of the edge cases. 

@alzimmermsft looking at the latest `java - core` failures running with this version of the test-proxy, I don't think the proxy is related to those. I'm not declaring total victory with this PR but the issues are seriously reduced.

This version is standing up to multiple storage runs: 
image

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Audit.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Audit.cs
@@ -1,0 +1,94 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Azure.Sdk.Tools.TestProxy.Common;
+using Azure.Sdk.Tools.TestProxy.Common.Exceptions;
+using Azure.Sdk.Tools.TestProxy.Store;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Azure.Sdk.Tools.TestProxy
+{
+    [ApiController]
+    [Route("[controller]/[action]")]
+    public sealed class Audit : ControllerBase
+    {
+        private readonly ILogger _logger;
+        private readonly RecordingHandler _recordingHandler;
+
+        public Audit(RecordingHandler recordingHandler, ILoggerFactory loggerFactory)
+        {
+            _recordingHandler = recordingHandler;
+            _logger = loggerFactory.CreateLogger<Record>();
+        }
+
+        [HttpGet]
+        public async Task GetAuditLogs()
+        {
+
+            var allAuditSessions = _recordingHandler.RetrieveOngoingAuditLogs();
+            allAuditSessions.AddRange(_recordingHandler.AuditSessions.Values);
+
+            StringBuilder stringBuilder = new StringBuilder();
+
+            foreach (var auditLogQueue in allAuditSessions) {
+                while (auditLogQueue.TryDequeue(out var logItem))
+                {
+                    stringBuilder.Append(logItem.ToCsvString() + Environment.NewLine);
+                }
+            }
+
+            Response.ContentType = "text/plain";
+
+            await Response.WriteAsync(stringBuilder.ToString());
+        }
+
+
+        [HttpPost]
+        public async Task Push([FromBody()] IDictionary<string, object> options = null)
+        {
+            DebugLogger.LogAdminRequestDetails(_logger, Request);
+
+            var pathToAssets = RecordingHandler.GetAssetsJsonLocation(StoreResolver.ParseAssetsJsonBody(options), _recordingHandler.ContextDirectory);
+
+            await _recordingHandler.Store.Push(pathToAssets);
+        }
+
+        [HttpPost]
+        [AllowEmptyBody]
+        public async Task Stop([FromBody()] IDictionary<string, string> variables = null)
+        {
+            string id = RecordingHandler.GetHeader(Request, "x-recording-id");
+            bool save = true;
+            EntryRecordMode mode = RecordingHandler.GetRecordMode(Request);
+
+            if (mode != EntryRecordMode.Record && mode != EntryRecordMode.DontRecord)
+            {
+                throw new HttpException(HttpStatusCode.BadRequest, "When stopping a recording and providing a \"x-recording-skip\" value, only value \"request-response\" is accepted.");
+            }
+
+            if (mode == EntryRecordMode.DontRecord)
+            {
+                save = false;
+            }
+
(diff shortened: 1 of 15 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Properties/launchSettings.json
```diff
@@ -2,6 +2,7 @@
   "profiles": {
     "Azure.Sdk.Tools.TestProxy": {
       "commandName": "Project",
+      "commandLineArgs": "--storage-location=C:/repo/azure-sdk-for-python",
```
scbedd, line 5:
```suggestion
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Audit.cs
```diff
@@ -0,0 +1,93 @@
...
+            DebugLogger.LogAdminRequestDetails(_logger, Request);
+
+            var pathToAssets = RecordingHandler.GetAssetsJsonLocation(StoreResolver.ParseAssetsJsonBody(options), _recordingHandler.ContextDirectory);
+
+            await _recordingHandler.Store.Push(pathToAssets);
+        }
+
+        [HttpPost]
+        [AllowEmptyBody]
+        public async Task Stop([FromBody()] IDictionary<string, string> variables = null)
```
scbedd, line 65: todo: cleanup random unused arguments. remove useless functions. etc.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -418,35 +462,58 @@ public async Task StartPlaybackAsync(string sessionId, HttpResponse outgoingResp
...
                 {
-                    throw new HttpException(HttpStatusCode.InternalServerError, $"Unexpectedly failed to remove in-memory session {session.SourceRecordingId}.");
+                    if (!InMemorySessions.TryGetValue(session.SourceRecordingId, out var inMemorySession))
+                    {
+                        throw new HttpException(HttpStatusCode.InternalServerError, $"Unexpectedly failed to retrieve in-memory session {session.SourceRecordingId}.");
+                    }
+
+                    Interlocked.Add(ref Startup.RequestsRecorded, -1 * inMemorySession.Session.Entries.Count);
+
+                    if (!InMemorySessions.TryRemove(session.SourceRecordingId, out _))
```
alzimmermsft, line 503: Couldn't just using `TryRemove(session.SourceRecordingId, out var inMemorySessions)` be sufficient here? Unless there are other locations outside of this EntryLock block which remove from the InMemorySessions which this catches a race condition between retrieval and removal.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -418,35 +462,58 @@ public async Task StartPlaybackAsync(string sessionId, HttpResponse outgoingResp
...
+                    {
+                        throw new HttpException(HttpStatusCode.InternalServerError, $"Unexpectedly failed to remove in-memory session {session.SourceRecordingId}.");
+                    }
+
+                    GC.Collect();
```
alzimmermsft, line 508: Is it common to force GC collection manually?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Audit.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            var allAuditSessions = _recordingHandler.RetrieveOngoingAuditLogs();
+            allAuditSessions.AddRange(_recordingHandler.AuditSessions.Values);
+
+            StringBuilder stringBuilder = new StringBuilder();
+
+            foreach (var auditLogQueue in allAuditSessions) {
+                while (auditLogQueue.TryDequeue(out var logItem))
+                {
+                    stringBuilder.Append(logItem.ToCsvString() + Environment.NewLine);
```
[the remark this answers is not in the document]
scbedd: Yeah. It [literally doesn't work. ](https://stackoverflow.com/a/11155844) this confused dev had the same experience I did. Extremely frustrating.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: Not really, but it does work more effectively than not having it (at least on performance profiler).

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
scbedd: I think you're right. It's not like we have an early exit. This will blow up if either of them fails. May as well reduce to one check.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
```diff
@@ -81,6 +82,48 @@ public class RecordingHandler
         public readonly ConcurrentDictionary<string, ModifiableRecordSession> PlaybackSessions
             = new ConcurrentDictionary<string, ModifiableRecordSession>();
 
+        // as sessions are popped off,
```
benbp, line 85: incomplete comment?

mikeharder (approved):

benbp (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs
semick-dev: **deep sigh**

scbedd, line 85:
```suggestion
```

PR #8791, opened 2024-08-06
Fix the issue in sorting revisions with more than one approver in req…

praveenkuttappan:
Revisions cannot be sorted when there is more than one approver in the requested list and this throws an exception `Failed to compare two elements in the array`.

Fix is to filter and project the sorting key based on the logged in user to correctly show the revisions in requested order for the logged in user.


praveenkuttappan:
@LarryOsterman This PR will resolve the failure in `Requested Reviews` page.

maririos (approved):

PR #8793, opened 2024-08-06
Add `tsp-client compare` command

tjprescott:
Basically is a pass-through to the packaged openapi-diff tool.


tools/tsp-client/src/commands.ts
```diff
@@ -284,6 +284,18 @@ export async function generateCommand(argv: any) {
   }
 }
 
+export async function compareCommand(argv: any, args: string[]) {
```
catalinaperalta, line 287: also, could we add a test just to have a starting point to check that this command works as expected? Feel free to check in example files as needed

tools/tsp-client/src/index.ts
```diff
@@ -218,6 +219,18 @@ const parser = yargs(hideBin(process.argv))
       await sortSwaggerCommand(argv);
     },
   )
+  .command(
+    "compare",
+    "Compare two Swaggers for functional equivalency. This is typically used to compare a source Swagger with a TypeSpec-generated Swagger to ensure that the TypeSpec project is functionally equivalent to the source Swagger.",
+    (yargs: any) => {
+      return yargs.help(false);
```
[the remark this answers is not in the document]
tjprescott: Well, it will do whatever the underlying script does, which is, correct, it doesn't give an error if they pass an unsupported flag. But that's something I can fix within that tool.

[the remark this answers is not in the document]
tjprescott: https://github.com/tjprescott/openapi-diff/issues/14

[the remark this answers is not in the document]
catalinaperalta: Also does the tool accept positional params? I suppose that will have to be considered in the logic of what it accepts or not. Also, we should add some info to the tsp-client readme about this new command and mention any flags folks should pass in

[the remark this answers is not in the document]
tjprescott: I'll update the README. It doesn't accept positional params. You have to specify --lhs and --rhs (or have them set in dotenv or environment variables).

tools/tsp-client/.gitignore
```diff
@@ -3,3 +3,5 @@ node_modules
 dist
 types
 temp
+tsp-output/
+output/
```
catalinaperalta, line 7: Just curious where does output/ come from?

tools/tsp-client/test/run_commands.ts
```diff
@@ -78,37 +78,40 @@ async function main() {
   // Run the convert command on a mgmt spec
   await runCommand(baseDir, ["convert", "-o", baseSpecDir, "--swagger-readme", mgmtSpecLink]);
 
+  // Run the compare command to ensure it is working
+  // TODO: Provide actual lhs and rhs values
```
catalinaperalta, line 82: Maybe you could just copy/paste the contoso widget manager generated tsp and add a couple of the handwritten modifications that you know are typically made in a handwritten swagger?

tools/tsp-client/src/index.ts
```diff
@@ -218,6 +219,18 @@ const parser = yargs(hideBin(process.argv))
       await sortSwaggerCommand(argv);
     },
   )
+  .command(
+    "compare",
+    "Compare two Swaggers for functional equivalency. This is typically used to compare a source Swagger with a TypeSpec-generated Swagger to ensure that the TypeSpec project is functionally equivalent to the source Swagger.",
```
catalinaperalta, line 224: Are we slightly contradicting ourselves here with what we're saying in the readme? Maybe we should reword to say: 
```suggestion
    "Compare two Swaggers for functional equivalency. This is typically used to compare a source Swagger with a TypeSpec project or TypeSpec generated Swagger to ensure that the TypeSpec project is functionally equivalent to the source Swagger.",
```

catalinaperalta (approved):
Looks good, just hoping we can get an example handwritten swagger to compare against

tools/tsp-client/.gitignore
tjprescott: The tool places the output files in `output` while `tsp-output` is where `tsp compile` will generate to.

tools/tsp-client/test/run_commands.ts
tjprescott: I copied in the files from one of my manual tests.

catalinaperalta (approved):

PR #8800, opened 2024-08-07
Fix issue preventing proper secret scan

scbedd:
And resolve failing `livetests` as well.

### DIFF
--- a/tools/test-proxy/tests.yml
+++ b/tools/test-proxy/tests.yml
@@ -162,11 +162,17 @@ stages:
         CLONE_LOCATION: "$(Agent.BuildDirectory)/net_repo"
 
       steps:
-        - task: UseDotNet@2 # About UseDotNet@2 task: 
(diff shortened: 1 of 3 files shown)


tools/test-proxy/tests.yml
```diff
@@ -162,12 +162,18 @@ stages:
...
 
       steps:
-        - task: UseDotNet@2 # About UseDotNet@2 task: 
+        - task: UseDotNet@2
           displayName: "Install .NET 8"
           retryCountOnTaskFailure: 3
           inputs:
             version: "8.x"
 
+        - task: UseDotNet@2
```
benbp, line 171: What is the need for 7 here?

benbp (approved):

tools/test-proxy/tests.yml
scbedd: To install it so that the .NET Core Playback tests pass compilation.

PR #8801, opened 2024-08-08
[Pylint] Disallow using legacy type hinting

MJoshuaB:
Addresses [azure-sdk-for-python:#31104](https://github.com/Azure/azure-sdk-for-python/issues/31104)
Checks for uses of legacy (&lt;Python 3.8) type hinting comments.
```python
def function(self, max_wait_time, **kwargs):
     # type: (int, Any) -> None
     pass
```


MJoshuaB:
@microsoft-github-policy-service agree

tools/pylint-extensions/azure-pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
```diff
@@ -4829,4 +4829,26 @@ def test_allowed_import_else(self):
...
+    def test_disallowed_typing(self):
+        """Check that illegal method typing comments raise warnings"""
+        fnt = astroid.extract_node(
+            """
+            def function(x):
+                # type: (str) -> str #@
+                pass
+            """
+        )
+        with self.assertAddsMessages(
```
annatisch, line 4848: Is it possible to add a negative test for `self.assertNoMessages` for a function that has been typed correctly?

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2750,6 +2750,28 @@ def visit_import(self, node):
...
+    priority = -1
+    msgs = {
+        "C4761": (
+            "Do not use legacy typing using comments.",
+            "do-not-use-legacy-typing",
+            "Do not use legacy typing using comments. Python 2 is no longer supported, use Python 3.9+ type hints instead.",
+        ),
+    }
+
+    def visit_functiontype(self, node):
```
l0lawrence, line 2767: I wasn't able to find functiontype as a node, could you point to where you found it?

tools/pylint-extensions/azure-pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
MJoshuaB: As it's written, no, as the ast [node](https://docs.python.org/3/library/ast.html#ast.FunctionType) used for the check (`FunctionType`) is exclusive to the legacy type hinting comments. Unit or functional tests could work, but fall outside of the pytest framework currently in use.

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
MJoshuaB: I have linked it in my original comment, it is part of the Python `ast` package and is not present in the `astroid` package. Here is the link in full: [https://docs.python.org/3/library/ast.html#ast.FunctionType](https://docs.python.org/3/library/ast.html#ast.FunctionType)

l0lawrence: Thanks! Okay, the pylint custom checkers use the AST representation provided by the astroid library, so the visit_XXX functions available would be nodes defined by astroid: https://github.com/pylint-dev/astroid/blob/0156c045346b231d4451e8aea85128dcf4e56c64/astroid/nodes/as_string.py#L97. Just as an idea, an alternative here could be calling visit_functiondef and diving into the function [properties ](https://pylint.pycqa.org/projects/astroid/en/latest/api/nodes/astroid.nodes.FunctionDef.html)there.

MJoshuaB: I see, that was a misunderstanding on my part. I've re-written the checker and tests, and re-run the checker against azure-sdk-for-python.

l0lawrence: Awesome thank you!

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2750,6 +2750,29 @@ def visit_import(self, node):
...
+        "C4761": (
+            "Do not use legacy typing using comments.",
+            "do-not-use-legacy-typing",
+            "Do not use legacy typing using comments. Python 2 is no longer supported, use Python 3.9+ type hints instead.",
+        ),
+    }
+
+    def visit_functiondef(self, node):
+        """Check that we aren't using legacy typing."""
+        if node.is_function and (node.type_comment_args or node.type_comment_returns):
```
l0lawrence, line 2769: nit: we can remove is_function - the method itself will only be called on function nodes

MJoshuaB: Done.

annatisch (approved):

l0lawrence (approved):

PR #8817, opened 2024-08-09
[tsp-client] Fix bundled dependency issue

tjprescott:
Includes openapi-diff as bundled dependency in package.json.


tools/tsp-client/package.json
```diff
@@ -33,6 +33,7 @@
     "tsp-client": "cmd/tsp-client.js"
   },
   "files": [
+    "openapi-diff-1.0.0.tgz",
```
catalinaperalta, line 36: Do we still need this?

tjprescott: Nope

tools/tsp-client/package.json
```diff
@@ -71,6 +72,9 @@
     "extension": [
       "ts"
     ],
+    "bundleDependency": [
```
catalinaperalta, line 75: Assuming this one does need to stay, right?

catalinaperalta (approved):
Can you also add a changelog entry for the new release?

PR #8819, opened 2024-08-10
Don't log errors that we end up raising #22557

JessicaBell00:
Addresses https://github.com/Azure/azure-sdk-for-python/issues/22557
Checks that errors raised in exception blocks aren't also logged at '`warning`' and '`error`' levels.


JessicaBell00:
@microsoft-github-policy-service agree

tools/pylint-extensions/azure-pylint-guidelines-checker/README.md
```diff
@@ -92,3 +92,4 @@ In the case of a false positive, use the disable command to remove the pylint er
...
 | docstring-type-do-not-use-class                    | Docstring type is formatted incorrectly. Do not use `:class` in docstring type.                                                                                      | pylint:disable=docstring-type-do-not-use-class                      | [link](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html)                            |
 | no-typing-import-in-type-check                     | Do not import typing under TYPE_CHECKING.                                                                                                                            | pylint:disable=no-typing-import-in-type-check                       | No Link.                                                                                                |
+| do-not-log-errors-that-get-raised                  | Do not log errors that get rasied in an exception block.                                                                                                             | pylint:disable=do-not-log-errors-that-get-raised                    | No Link.                                                                                                |
```
annatisch, line 95: Could maybe simplify this to `do-not-log-raised-errors`?

tools/pylint-extensions/azure-pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
```diff
@@ -4829,4 +4829,121 @@ def test_allowed_import_else(self):
...
+
+    CHECKER_CLASS = checker.DoNotLogErrorsEndUpRaising
+
+    def test_error_level_not_logged(self):
+        """Check that any exceptions raised aren't logged at error level in the exception block."""
+        exception_node = astroid.extract_node('''
+        try:
+            add = 1 + 2
+        except Exception as e:
+            logger.error(str(e))
```
annatisch, line 4847: The logic here is actually going to be trickier than this I think. Because it's quite possible to have lots of varying branches in an `except` clause. Some of our error handling cases have pretty complex code. For example:
```python
except SomeException as e:
    if e.code == "Retryable":
        logger.warning(f"Retriable failure occured: {e}, attempting to restart")  # This kind of logging is okay
        return True
    elif isinstance(SomeException(e, SomeDerivedException):
        logger.error(f"System shutting down due to error: {e}.")  # This logging is also okay I think
        return False
    logger.error(f"Unexpected error occurred: {e}") # This is the logging we'd want to flag as excessive.
    raise SystemError("Uh oh") from e
```
I'm not even sure if it's possible to make such a fine-grained distinction - as I'm not sure whether we can parse branches. If we can parse branches, then we'd probably want to flag branches that end in either:
- `raise`
- `raise <ExceptionType> from <exception_instance>`

Let's add some more complex test scenarios here to see how far we can push this. But if it's not possible with the data we have then we can get by.

tools/pylint-extensions/azure-pylint-guidelines-checker/README.md
JessicaBell00: Have renamed :)

tools/pylint-extensions/azure-pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
JessicaBell00: I have added new code to parse branches and sub-branches. Have re-run against the SDKs and found several files that don't comply with the new rules, I can attach these if you'd like. No false positives from testing against SDKs.

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2708,6 +2708,68 @@ def visit_importfrom(self, node):
...
+
+    def check_for_raise(self, node):
+        """ Helper function - checks for instance of 'Raise' node
+            Also checks 'If' and nested 'If' branches
+        """
+        for i in node:
+            if isinstance(i, astroid.Raise):
+                self.check_for_logging(node)
+            # Check for any nested 'If' branches
+            if isinstance(i, astroid.If):
```
annatisch, line 2744: I think this is probably a good start - I did a quick scan of all the SDKs just for logging.error, and all the cases I spotted that should be flagged by this rule were either in the `except` clause directly, or nested in an if-statement.
While this might miss some future cases that use for-loops or some other syntax - I think this will still capture the majority of scenarios.

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2708,6 +2708,68 @@ def visit_importfrom(self, node):
...
+                self.check_for_raise(i.body)
+                # Check any 'elif' and 'else' branches
+                elif_statements = i.orelse
+                while len(elif_statements) == 1:
```
annatisch, line 2748: What's the structure of `elif_statements`? I'm curious why the length must always equal 1?

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2708,6 +2708,68 @@ def visit_importfrom(self, node):
...
+                    if isinstance(elif_statements[0], astroid.If):
+                        for x in elif_statements:
+                            self.check_for_raise(x.body)
+                            elif_statements = x.orelse
```
annatisch, line 2752: I like the recursive approach here - though it's generally best to avoid reassigning the iterable while it's currently being iterated. Looking at the tests this seems to be functioning correctly - if you could link me to the docs for the `astroid.If` node and what the structure of the `orelse` property is I can see if I have any ideas.

tools/pylint-extensions/azure-pylint-guidelines-checker/tests/test_pylint_custom_plugins.py
```diff
@@ -4829,4 +4829,282 @@ def test_allowed_import_else(self):
             self.checker.visit_importfrom(ima)
             self.checker.visit_import(imb)
             self.checker.visit_import(imc)
-            self.checker.visit_importfrom(imd)
\ No newline at end of file
+            self.checker.visit_importfrom(imd)
+
+
+class TestDoNotLogErrorsEndUpRaising(pylint.testutils.CheckerTestCase):
```
annatisch, line 4835: Really great test coverage!

annatisch (commented):
Looking good - I think we're nearly there.
Do you have a sample output for SDKs in the repo that get flagged by this new rule?

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
JessicaBell00: Hi, the `orelse` property gives a list of `elif` or `else` nodes.
The length is 1 to account for multiple `elif` statements. What I found through trial and error was that for `elif` statements, `orelse` would always return a list of length 1 - one `If` node, whereas for `else` statements it will return a list of length 2 - one `Expr` node (the logging of the exception) and one `raise` node.

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
JessicaBell00: Hi, the `If` node docs can be found [here](https://pylint.readthedocs.io/projects/astroid/en/latest/api/nodes/astroid.nodes.If.html), `orelse` can be found on this page too. The `orelse` property is a list of nodes as described in the above comment :)

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
JessicaBell00: Have added checking for `for` loops, they use `orelse` as well, so simple to add in.

JessicaBell00:
> Looking good - I think we're nearly there. Do you have a sample output for SDKs in the repo that get flagged by this new rule?

The output of the Pylint testing against the SDKs is below. This was before the changes I have just made. Will run tests again later :)
[pylintreport.txt](https://github.com/user-attachments/files/16831376/pylintreport.txt)

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
annatisch: Thank you! I did some playing around with this and I figure out we don't actually need to loop over the contents of the `orelse` property - we can actually pass the entire thing into back into the `check_for_raise` function and it will run the iteration for us :)

So I think the whole thing can be simplified into this:
```python
    def check_for_raise(self, node):
        """ Helper function - checks for instance of 'Raise' node
            Also checks 'If' and nested 'If' branches
        """
        for i in node:
            if isinstance(i, astroid.Raise):
                self.check_for_logging(node)
                
            # Check for any nested 'If' branches
            if isinstance(i, astroid.If):
                self.check_for_raise(i.body)

                # Check any 'elif', 'else' or 'for' branches
                self.check_for_raise(i.orelse)
```
Let me know if this has any issues with the testing!

JessicaBell00: Ah yes! Works perfectly thank you! Think I overcomplicated it in my head :)

annatisch (approved):

JessicaBell00:
Output from re-testing against the SDKs after changes made yesterday. Looks good, all true positives :)
[pylintreport2.txt](https://github.com/user-attachments/files/16857720/pylintreport2.txt)

l0lawrence (approved):

Issue #8057, opened 2024-04-10
[SDK Automation - JS] Support SDK generation in CI pipeline for the services migrated from Swagger to TypeSpec


raych1:
If the migrated service has existing customized SDK code, the automation will keep them and report the error in the compiling step when the generated codes are not compatible with existing code. 

This refers to the option 1 described in the linked epic.

> option 1: keep the existing customized SDK code and generate the SDK code based on TypeSpec specs.
> If the customized SDK code is not compatible with the new specs, the generation should fail in the build/compile step providing the comprehensive error message, such as "The SDK has customized SDK code and might not be compatible with the new specs, the service team may need to update the customized code or the new specs."

@MaryGao

raych1:
Synced with @wanlwanl and he will verify for both modular codeGen and RLC codeGen.

wanlwanl:
Code gen method:
- autorest: codegen will clear code, so automation doesn't need to clear again
- typespec: automation should clear package folder before code gen

PR #8826, opened 2024-08-13
Clear Package Folder Before Generation

wanlwanl:
Fix: https://github.com/Azure/azure-sdk-tools/issues/8057
Test pipeline: https://github.com/test-repo-billy/azure-rest-api-specs/pull/3473

### DIFF
--- a/tools/js-sdk-release-tools/src/common/utils.ts
+++ b/tools/js-sdk-release-tools/src/common/utils.ts
@@ -20,7 +20,6 @@ const messageToTspConfigSample =
 async function loadTspConfig(typeSpecDirectory: string): Promise<Exclude<any, null | undefined>> {
     const configPath = join(typeSpecDirectory, 'tspconfig.yaml');
     const content = await readFile(configPath, { encoding: 'utf-8' });
-    console.log('content', content.toString());
     const config = parse(content.toString());
     if (!config) {
         throw new Error(`Failed to parse tspconfig.yaml in ${typeSpecDirectory}`);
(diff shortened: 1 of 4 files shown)


tools/js-sdk-release-tools/src/common/utils.ts
```diff
@@ -73,4 +93,20 @@ export function tryReadNpmPackageChangelog(packageFolderPath: string): string {
...
+    const tspConfig = await loadTspConfig(typeSpecDirectory);
+    const serviceDir = tspConfig.parameters?.['service-dir']?.default;
+    if (!serviceDir) {
+        throw new Error(`Misses service-dir in parameters section of tspconfig.yaml. ${messageToTspConfigSample}`);
+    }
+    const packageDir = tspConfig.options?.[emitterName]?.['package-dir'];
+    if (!packageDir) {
+        throw new Error(`Misses package-dir in ${emitterName} options of tspconfig.yaml. ${messageToTspConfigSample}`);
+    }
+    const packageDirFromRoot = posix.join(serviceDir, packageDir);
```
MaryGao, line 110: just curious why posix join? what if this command runs in windows environment?

MaryGao (approved):
Left small comments!

Regarding the customization code I remember Ray has mentioned we could improve our automation to have message like `The SDK has customized SDK code and might not be compatible with the new specs, the service team may need to update the customized code or the new specs`. Not sure this is feasible and we have plan for this.

tools/js-sdk-release-tools/src/common/utils.ts
wanlwanl: it's used in config/package that only accept posix format path.

wanlwanl:
> Left small comments!
> 
> Regarding the customization code I remember Ray has mentioned we could improve our automation to have message like `The SDK has customized SDK code and might not be compatible with the new specs, the service team may need to update the customized code or the new specs`. Not sure this is feasible and we have plan for this.

The current customized code is mixed in generated code, we don't know if customer has customized code. I think it's better to seperate generated code and customized code in the future.

PR #8827, opened 2024-08-13
upgrade openapi-to-typespec

pshao25:
1. upgrade openapi-to-typespec to 0.9.0
2. add --mgmt-debug.suppress-list-exception to .net generator dependency


catalinaperalta (approved):
I'll add an entry in the changelog for this in a separate pr to include this in our release later today

PR #8830, opened 2024-08-13
Add RestAPISpecsDocsRepos to branch cleanup and update the delete scripts to output limits and telemetry data

JimSuplizio:
1. Add MicrosoftDocs/AzureRestPreview to the cleanup scripts. There were about 130k branches that were deleted prior to this change. The number of branches in the repository is now at a manageable level.
2. We have a lot of repositories we're deleting branches from and we're also making gh API calls other than branch cleanup. Output some telemetry.
3. If one of the API commands fails, exit the script, don't continue to try and process. The reason being is that if we hit a rate limit we can't continue processing. 

The changes in here, right now, are going to be used for information gathering. There's currently no data of how many API calls are being made when the pipeline runs. Further updates will probably be made depending on the findings.

### DIFF
--- a/eng/common/scripts/Delete-RemoteBranches.ps1
+++ b/eng/common/scripts/Delete-RemoteBranches.ps1
@@ -80,12 +80,16 @@ $ghPRViewCalls = 0
 $ghBranchDeleteCalls = 0
 
 try {
+  # Output the core rate limit at the start of processing. There's no real need
+  # to output this at the end because the GH call counts are being output
+  $coreRateLimit = Get-RateLimit ([RateLimitTypes]::core)
+  Write-RateLimit $coreRateLimit
   # Output the GraphQL rate limit before and after the call
   $graphqlRateLimit = Get-RateLimit ([RateLimitTypes]::graphql)
-  Write-RateLimit $graphqlRateLimit
+  Write-RateLimit $graphqlRateLimit "Before GraphQL Call"
   $branches = Get-AllBranchesAndPullRequestInfo $owner $repo
   $graphqlRateLimit = Get-RateLimit ([RateLimitTypes]::graphql)
-  Write-RateLimit $graphqlRateLimit
+  Write-RateLimit $graphqlRateLimit "After GraphQL Call"
 
   if ($branches) {
     $totalBranchesFromQuery = $branches.Count
--- a/eng/common/scripts/Delete-RemoteBranches.ps1
+++ b/eng/common/scripts/Delete-RemoteBranches.ps1
@@ -15,7 +15,6 @@ param(
   $AuthToken
 )
 Set-StrictMode -version 3
-
 . (Join-Path $PSScriptRoot common.ps1)
 
 function Get-AllBranchesAndPullRequestInfo($owner, $repo) {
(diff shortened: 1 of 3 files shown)


eng/common/scripts/Delete-RemoteBranches.ps1
```diff
@@ -66,67 +65,124 @@ if ($AuthToken) {
...
+$skippedForPRNotInBranch = 0
+$skippedForPRNotInRepo = 0
+# gh call counters
+$ghPRViewCalls = 0
+$ghBranchDeleteCalls = 0
+
+try {
+  # Output the core rate limit at the start of processing. There's no real need
+  # to output this at the end because the GH call counts are being output
+  $coreRateLimit = Get-RateLimit ([RateLimitTypes]::core)
```
weshaggard, line 85: While this works for 1 parameter you shouldn't call PS functions with `(...)` as it turns them into an array.

eng/common/scripts/Delete-RemoteBranches.ps1
```diff
@@ -66,67 +65,124 @@ if ($AuthToken) {
...
+        exit $LASTEXITCODE
+      }
+      LogDebug "The branch '$branchName' at commit '$commitUrl' in '$RepoId' has been deleted."
     }
-    Write-Host "The branch '$branchName' at commit '$commitUrl' in '$RepoId' has been deleted."
+  }
+}
+finally {
+
+
```
weshaggard, line 172:
```suggestion
finally
{
```

eng/common/scripts/Helpers/git-helpers.ps1
```diff
@@ -110,3 +110,143 @@ class ConflictedFile {
...
+# call needs to select the particular property to return the right limit. This ensures
+# that rate limit type being passed to the function will exist.
+enum RateLimitTypes {
+  core
+  search
+  graphql
+}
+
+# Fetch the rate limit for the given RateLimitType
+# TODO: Possibly change the parameter to take in an array of types and return an
```
weshaggard, line 211: I would just remove this TODO.

eng/common/scripts/Helpers/git-helpers.ps1
```diff
@@ -110,3 +110,143 @@ class ConflictedFile {
...
+  $rate_limit = $returnValue | ConvertFrom-Json -AsHashtable | Select-Object -ExpandProperty resources | Select-Object -ExpandProperty $RateLimitType
+  # Add the limit type for convenance
+  $rate_limit["type"] = $RateLimitType
+  return $rate_limit
+}
+
+# Get the number of minutes until RateLimit reset rounded up to the nearest minute
+# for the passed in RateLimit. This is more applicable to the core and graphql rate
+# limits than search because the search rate limit resets every minute
+function Get-MinutesUntilRateLimitReset($RateLimit) {
```
weshaggard, line 230: I don't see this used anywhwere.

weshaggard (approved):

eng/common/scripts/Delete-RemoteBranches.ps1
JimSuplizio: For an enum, this isn't creating an array it's evaluating it. The syntax and processing for enums in powershell is a bit weird. That being said, after a little tinkering, I found `Get-RateLimit ([RateLimitTypes]::core)` is the same as `Get-RateLimit core` and I'll change them to that since it looks less...odd.

eng/common/scripts/Helpers/git-helpers.ps1
JimSuplizio: Will be removed in next commit.

eng/common/scripts/Delete-RemoteBranches.ps1
JimSuplizio: This is a nit but the if you look at finally statements in azure-sdk-tools *.ps1, the vast majority have their being brace immediately after the try or the finally statement.

eng/common/scripts/Helpers/git-helpers.ps1
JimSuplizio: No, because it isn't yet. Recall that these changes are initially going to be gathering data, until we know how many calls we're using or whether or not we hit rate limits, I can't really use this to wait until rate limit reset.

PR #8847, opened 2024-08-15
[tsp-client] Add gh action for tests

catalinaperalta:
Add a GH action for tsp-client tests


.github/workflows/tsp-client-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+  pull_request:
+    paths:
+      - tools/tsp-client/**
+
+jobs:
+  tsp-client:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        node-version: [18.19, 20]
```
mikeharder, line 16:
```suggestion
        node-version: [18, 20]
```

I think you should be able to just use 18, it should install 18.latest (or close to latest) by default.

.github/workflows/tsp-client-tests.yml
```diff
@@ -0,0 +1,55 @@
+name: tsp-client - Test
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths:
+      - tools/tsp-client/**
```
mikeharder, line 9:
```suggestion
      - .github/workflows/tsp-client-tests.yml
      - tools/tsp-client/**
```

You want to also trigger the action if the workflow YML itself is updated.  This change should also run your new action on this very PR.

mikeharder (approved):

mikeharder:
I believe this is blocked, until we change how workflows are synced across repos, since this workflow only applies to the tools repo:

https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/eng-workflows-sync.yml

PR #8849, opened 2024-08-16
Add FilePattern support for .github\workflow sync

weshaggard:
Given we cannot create subfolders under .github\workflow folder and we have repos that want additional workflows then we have in the tools repo we need to setup a file pattern matching for the sync processing to only sync the ones we want to be common.

For now just use the existing *event* pattern but we can move them to another pattern similar to *common* or *sync* in the future as we add more common workflows.

### DIFF
--- a/eng/common/pipelines/templates/steps/eng-common-workflow-enforcer.yml
+++ b/eng/common/pipelines/templates/steps/eng-common-workflow-enforcer.yml
@@ -21,7 +21,7 @@ steps:
         }
         if ((!"$(System.PullRequest.SourceBranch)".StartsWith("sync-.github/workflows")) -and "$(System.PullRequest.TargetBranch)" -match "^(refs/heads/)?$(DefaultBranch)$")
         {
-          $filesInCommonDir  = & "eng/common/scripts/get-changedfiles.ps1" -DiffPath '.github/workflows/*' -DiffFilterType ""
+          $filesInCommonDir  = & "eng/common/scripts/get-changedfiles.ps1" -DiffPath '.github/workflows/*event*' -DiffFilterType ""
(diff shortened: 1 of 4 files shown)


mikeharder (approved):

JimSuplizio (changes requested):
@weshaggard, there's one file that also needs to be updated. [eng-common-workflow-enforcer.yml](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/steps/eng-common-workflow-enforcer.yml#L22) would need to be modified so it would exclude *event* but allow other files. Right now, it'll exclude the entire .github/workflows directory.

eng/common/pipelines/templates/steps/eng-common-workflow-enforcer.yml
```diff
@@ -21,7 +21,7 @@ steps:
         }
         if ((!"$(System.PullRequest.SourceBranch)".StartsWith("sync-.github/workflows")) -and "$(System.PullRequest.TargetBranch)" -match "^(refs/heads/)?$(DefaultBranch)$")
         {
-          $filesInCommonDir  = & "eng/common/scripts/get-changedfiles.ps1" -DiffPath '.github/workflows/*' -DiffFilterType ""
+          $filesInCommonDir  = & "eng/common/scripts/get-changedfiles.ps1" -DiffPath '.github/workflows/*event*' -DiffFilterType ""
```
weshaggard, line 24: @JimSuplizio doesn't this take care of your concern around the enforcement?

JimSuplizio (approved):
@weshaggard, I'm an idiot and I literally stared right at the file in the PR and didn't see it. I'm approving.

Issue #8853, opened 2024-08-19
samples automation, handle x-ms-original-file from TypeSpec path


weidongxu-microsoft:
Investigation.

`x-ms-original-file: 2024-03-01-preview/MongoClusters_ListConnectionStrings.json`

automation read tsp-location.yaml to find out that `directory: specification/mongocluster/DocumentDB.MongoCluster.Management`

combine them (involve glob) to `specification/mongocluster/DocumentDB.MongoCluster.Management/examples/2024-03-01-preview/MongoClusters_ListConnectionStrings.json`

finally it would find it from example in Swagger e.g. `specification/mongocluster/resource-manager/Microsoft.DocumentDB/preview/2024-03-01-preview/MongoClusters_ListConnectionStrings.json`

PR #8854, opened 2024-08-19
examples automation examples-dir for tsp

weidongxu-microsoft:
fix https://github.com/Azure/azure-sdk-tools/issues/8853

main code here https://github.com/Azure/azure-sdk-tools/blob/examples-automation_examples-dir-for-tsp/tools/azure-rest-api-specs-examples-automation/directory/examples_dir.py
test in same folder


tools/azure-rest-api-specs-examples-automation/directory/test_examples_dir.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @unittest.skip(
+        'require call "_set_paths" with your local azure-rest-api-specs repository and azure-sdk-for-java repository'
+    )
+    def test_find_resource_manager_example_typespec(self):
+        example_dir = try_find_resource_manager_example(
+            "c:/github/azure-rest-api-specs",
+            "2024-03-01-preview",
+            "MongoClusters_ListConnectionStrings.json",
+            "c:/github/azure-sdk-for-java/sdk/mongocluster/azure-resourcemanager-mongocluster",
```
weshaggard, line 16: Will these local paths cause the tests to not be able to run in CI?

weshaggard (approved):
Looks reasonable once the pipeline is fixed.

tools/azure-rest-api-specs-examples-automation/directory/test_examples_dir.py
weidongxu-microsoft: Currently this test is disabled for this reason.

Let me see if I can prepare a very simple local file structure during test run (in a temp folder) to make the test run in CI.

PR #8860, opened 2024-08-20
Compare json object values instead of byte streams when matching bodies that are content-type `json`

### LINKED ISSUE #3171: [Test-Proxy] Custom matcher for body keys only
Key Vault tests currently use a custom [json_attribute_matcher](https://github.com/Azure/azure-sdk-for-python/blob/c8291ac6cb0dbd865da03a88dd2bcb9279e2c4a6/sdk/keyvault/azure-keyvault-certificates/tests/_shared/json_attribute_matcher.py) with `vcrpy` as a custom request matcher. This checks that the keys in request bodies match, without enforcing an ordering and without comparing body values.

It would be great if we could support this kind of matching with the test proxy, as it probably would have uses outside of KV for tests that have consistently-structured bodies with possibly (expectedly) inconsistent values.

@scbedd

scbedd:
Resolves #3171

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
@@ -103,6 +103,9 @@ namespace Azure.Sdk.Tools.TestProxy.Common
                 sanitizer.Sanitize(requestEntry);
             }
 
+            // normalize request body with STJ using relaxed escaping to match behavior when Deserializing from session files
+            RecordEntry.NormalizeJsonBody(requestEntry.Request);
+
             RecordEntry entry = matcher.FindMatch(requestEntry, Entries);
             if (remove)
             {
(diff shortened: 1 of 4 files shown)


ArcturusZhang (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -102,9 +102,6 @@ public RecordEntry Lookup(RecordEntry requestEntry, RecordMatcher matcher, IEnum
             {
                 sanitizer.Sanitize(requestEntry);
             }
-            // normalize request body with STJ using relaxed escaping to match behavior when Deserializing from session files
```
scbedd, line 105: We no longer need to normalize bytes because if it's JSON, we compare by value now.

scbedd: We now need to add this back now that we check the sequenceequal FIRST. the normalization still needs to happen to make certain request matches response.

JoshLove-msft (approved):

scbedd:
Located a hole in the logic where we don't properly handle when the body doesn't parse even when it's marked as `json` (this happens for deliberate error cases).

[This test](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4077071&view=ms.vss-test-web.build-test-results-tab&runId=49548003&resultId=105015&paneView=debug) in java turned it up. Newest commit protects against the unexpected exception + resolves the test invocation locally.

PR #8878, opened 2024-08-23
Pull Request automation feedback

scbedd:
From @hallipr dealing with rust implementation.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -113,7 +113,7 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     $diff = Get-Content $InputDiffJson | ConvertFrom-Json
     $targetedFiles = $diff.ChangedFiles
 
-    $dependentPackagesForInclusion = @()
+    $additionalValidationPackages = @()
     $lookup = @{}
 
     foreach ($pkg in $allPackageProperties)
@@ -130,13 +130,13 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
                 $packagesWithChanges += $pkg
 
                 if ($pkg.AdditionalValidationPackages) {
-                    $dependentPackagesForInclusion += $pkg.AdditionalValidationPackages
+                    $additionalValidationPackages += $pkg.AdditionalValidationPackages
                 }
             }
         }
     }
 
-    foreach ($addition in $dependentPackagesForInclusion) {
+    foreach ($addition in $additionalValidationPackages) {
         $key = $addition.Replace($RepoRoot, "").SubString(1)
 
         if ($lookup[$key]) {
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -15,7 +15,7 @@ class PackageProps
     [boolean]$IsNewSdk
     [string]$ArtifactName
     [string]$ReleaseStatus
-    [string[]]$CanaryPackages
+    [string[]]$AdditionalValidationPackages
 
     PackageProps([string]$name, [string]$version, [string]$directoryPath, [string]$serviceDirectory)
     {
@@ -113,11 +113,6 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     $diff = Get-Content $InputDiffJson | ConvertFrom-Json
     $targetedFiles = $diff.ChangedFiles
 
-    $customDependentPackagesHandler = $false
-    if ($DependentPackagesFromPackageSetFn -and (Test-Path "Function:$DependentPackagesFromPackageSetFn")) {
-        $customDependentPackagesHandler = $true
-    }
-
     $dependentPackagesForInclusion = @()
     $lookup = @{}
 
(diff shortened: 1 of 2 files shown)


scbedd:
I have identified the issue. `azure-sdk-for-android` is out of alignment with `azure-sdk-tools` as far as `eng/common/scripts/common.ps1`. Doing a manual comparison across the repos and finding the gaps.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -15,7 +15,7 @@ class PackageProps
     [boolean]$IsNewSdk
     [string]$ArtifactName
     [string]$ReleaseStatus
-    [string[]]$DependentPackages
+    [string[]]$CanaryPackages
```
weshaggard, line 18: Can we name this something like "AdditionalValidationPackages" or something a little meaningful to explain what they are intended for?

eng/common/scripts/Package-Properties.ps1
```diff
@@ -113,6 +113,11 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     $diff = Get-Content $InputDiffJson | ConvertFrom-Json
     $targetedFiles = $diff.ChangedFiles
 
+    $customDependentPackagesHandler = $false
+    if ($DependentPackagesFromPackageSetFn -and (Test-Path "Function:$DependentPackagesFromPackageSetFn")) {
```
weshaggard, line 117: Do we need another function here or can we handle this inside of function where were create the PackageProperties?

weshaggard: It looks like you already are in your PR https://github.com/Azure/azure-sdk-for-python/pull/36635/files#diff-1d28dc86baa237ca6efe6224431142a9e8bbff66fa10a2c20c69031a22f5b387R62

weshaggard:
> I have identified the issue. `azure-sdk-for-android` is out of alignment with `azure-sdk-tools` as far as `eng/common/scripts/common.ps1`. Doing a manual comparison across the repos and finding the gaps.
> 
> EDIT: one additional:
> 
> * [azure-sdk-for-python](https://github.com/Azure/azure-sdk-for-python/blob/main/eng/common/scripts/common.ps1)
> * [azure-sdk-for-android](https://github.com/Azure/azure-sdk-for-android/blob/main/eng/common/scripts/common.ps1)

I've just triggered a manual sync run https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4086664&view=logs&j=96791242-dbf3-587e-3a06-ae5af5c1a705&t=545ac0b0-0fa6-5135-76aa-19ebc03770c9 to try and make sure eng/common is not out-of-sync for anything else.

eng/common/scripts/Package-Properties.ps1
scbedd: That's what the `CanaryPackages` rename was supposed to be 👍 

If that's not clear enough then sure, I'll go with your name.

eng/common/scripts/Package-Properties.ps1
scbedd: If you don't implement the function, yes, we use the common one.

scbedd:
I just reverted this to _identical_ to the version from `main` + my updates. STILL no dice on this diff. Gonna have to manually repro the diff and discover how can I patch this.

weshaggard:
> I just reverted this to _identical_ to the version from `main` + my updates. STILL no dice on this diff. Gonna have to manually repro the diff and discover how can I patch this.

I missed merging the android update yesterday. I did this morning and retried your pipeline and the patching worked this round.

eng/common/scripts/Package-Properties.ps1
weshaggard: OK I could see this potentially being useful for mapping changes outside of the sdk folders, like `eng` to cause some set of test packages like template or some core ones. 

That actually makes me think we might also need a fallback so that if no packages are discovered that we use at least one test package.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -129,16 +134,25 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
...
 
-    foreach ($addition in $dependentPackagesForInclusion) {
-        if ($lookup[$addition]) {
-            $packagesWithChanges += $lookup[$addition]
+    # there is a custom function to get dependent packages from a changed package set and diff
+    if ($customDependentPackagesHandler)
+    {
+        $packagesWithChanges += &$DependentPackagesFromPackageSetFn $packagesWithChanges $diff
+    }
+    else {
```
weshaggard, line 149: Should this be an else? It seems like it should be a union of the two sets of packages.

eng/common/scripts/common.ps1
```diff
@@ -63,6 +63,7 @@ $GetEmitterNameFn = "Get-${Language}-EmitterName"
 $GetDirectoriesForGenerationFn = "Get-${Language}-DirectoriesForGeneration"
 $UpdateGeneratedSdksFn = "Update-${Language}-GeneratedSdks"
 $IsApiviewStatusCheckRequiredFn = "Get-${Language}-ApiviewStatusCheckRequirement" 
+$DependentPackagesFromPackageSetFn = "Get-${Language}-DependentPackagesFromPackageSet"
```
weshaggard, line 66: Whatever term we end up using for the PackageProperties (i.e. Canary or whatever) we should name this function similarly.

eng/common/scripts/Package-Properties.ps1
scbedd: It's intentional, but you could disagree with the intent 👍 

I came at this from the stance of:

> If you don't give me a custom function, I will use the default. Otherwise I will only use what you decide.

Hence the `else` instead of the `or` here. You want a union so that's what I'll go with.

eng/common/scripts/Package-Properties.ps1
scbedd: > That actually makes me think we might also need a fallback so that if no packages are discovered that we use at least one test package.

And this functionality will allow us to handle that!

eng/common/scripts/common.ps1
scbedd: Updated and aligned.

eng/common/scripts/Package-Properties.ps1
weshaggard: Let's start with a union and see how the usages end up looking like to better understand if the custom function needs to override the entire set or not.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -129,19 +129,26 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
...
-                if ($pkg.DependentPackages) {
-                    $dependentPackagesForInclusion += $pkg.DependentPackages
+                if ($pkg.AdditionalValidationPackages) {
+                    $dependentPackagesForInclusion += $pkg.AdditionalValidationPackages
                 }
             }
         }
     }
 
     foreach ($addition in $dependentPackagesForInclusion) {
```
weshaggard, line 139: minor given is just a local variable but we might want to make `dependentPackagesForInclusion` match the rename as well.

weshaggard (approved):

PR #8879, opened 2024-08-23
Allow save-package-properties to exit 0

hallipr:
Allow save-package-properties to exit 0 when PRs produce no package changes

### DIFF
--- a/eng/common/scripts/Save-Package-Properties.ps1
+++ b/eng/common/scripts/Save-Package-Properties.ps1
@@ -99,56 +99,63 @@ $exportedPaths = @{}
 $allPackageProperties = @()
 
 if ($prDiff) {
+  Write-Host "Getting package properties for PR diff file: $prDiff"
   $allPackageProperties = Get-PrPkgProperties $prDiff
+
+  if (!$allPackageProperties) {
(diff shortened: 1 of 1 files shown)


scbedd (approved):
I especially like the fact that the old style "service directory oriented" will still fail if detects no package. Which **IS** reasonable. This seems g2g to me.

hallipr:
/check-enforcer override

ADO check suite not updating.  Completed successfully

eng/common/scripts/Save-Package-Properties.ps1
```diff
@@ -99,56 +99,63 @@ $exportedPaths = @{}
...
-          Write-Host "Release date: $($pkg.ReleaseStatus)"
-          $configFilePrefix = $pkg.Name
-          if ($pkg.ArtifactName)
-          {
-            $configFilePrefix = $pkg.ArtifactName
-          }
-          $outputPath = Join-Path -Path $outDirectory "$configFilePrefix.json"
-          Write-Host "Output path of json file: $outputPath"
-          $outDir = Split-Path $outputPath -parent
-          if (-not (Test-Path -path $outDir))
```
weshaggard, line 130: Is there any case where this directory doesn't already exist?

PR #8902, opened 2024-08-28
Add descriptive error when variable groups are not added to pipeline

benbp:
These errors are annoying to debug because we just see indirect null object failures when expecting a json string and receiving the `$(<var name>)` literal from devops.

CC @deyaaeldeen

### DIFF
--- a/eng/common/TestResources/SubConfig-Helpers.ps1
+++ b/eng/common/TestResources/SubConfig-Helpers.ps1
@@ -196,7 +196,13 @@ function UpdateSubscriptionConfigurationWithFiles([object]$baseSubConfig, [strin
 # Helper function for processing stringified json sub configs from pipeline parameter data
 function BuildAndSetSubscriptionConfig([string]$baseSubConfigJson, [string]$additionalSubConfigsJson, [string]$subConfigFilesJson) {
   $finalConfig = @{}
+
   if ($baseSubConfigJson) {
+    # When variable groups are not added to the pipeline, secret references like
+    # $(<my secret>) are passed as a string literal instead of being replaced by the keyvault secret value
+    if ($baseSubConfig -like '$(*') {
+      throw "Expected a json dictionary object but found '$baseSubConfig'. This probably means a subscription config secret was not downloaded. The pipeline is likely missing a variable group."
+    }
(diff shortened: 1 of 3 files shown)


eng/common/TestResources/build-test-resource-config.yml
```diff
@@ -49,5 +49,13 @@ steps:
         ${{ convertToJson(parameters.SubscriptionConfigurationFilePaths) }}
       '@.Trim()
 
+      # When variable groups are not added to the pipeline, secret references like
```
weshaggard, line 52: Should we do this in the script and apply it to all inputs?

benbp: You mean check if it's a json object (e.g. starts with `[` or `{`)? That could work too, originally I was thinking this was just a pipeline specific problem hence why I surfaced the validation at this level, but I think both directions are reasonable.

weshaggard: Yes just validate all 3 inputs for this issue in the script before calling convertfrom-json

eng/common/TestResources/SubConfig-Helpers.ps1
```diff
@@ -207,6 +213,9 @@ function BuildAndSetSubscriptionConfig([string]$baseSubConfigJson, [string]$addi
     $subConfigs = $additionalSubConfigsJson | ConvertFrom-Json -AsHashtable
```
weshaggard, line 213: We should probably also check these strings.

benbp: I check them below for string in the loop. The outer type will always be array since we type check it in the pipeline yaml. I can add another string check here for safety but it will only guard against local input I think

weshaggard (approved):

eng/common/TestResources/deploy-test-resources.yml
```diff
@@ -43,7 +43,7 @@ steps:
 
   - ${{ if eq('true', parameters.UseFederatedAuth) }}:
     - task: AzurePowerShell@5
-      displayName: Deploy test resources
+      displayName: 🚀 Deploy test resources
```
weshaggard, line 46: you and your emoji's :)

PR #8904, opened 2024-08-29
upgrade openapi-to-typespec version to 0.9.1

pshao25:
As titled


catalinaperalta (commented):
@pshao25 please update the package patch version and add an entry in the changelog with tomorrow's date (Aug 30) for this change.

catalinaperalta (approved):

PR #8911, opened 2024-08-30
Adding Network - Cloud label

ronniegeraghty:


jsquire (approved):

jsquire:
Pro-tip:  When pushing this out to the repositories, making a temporary copy of the common labels file with just this entry speeds things up considerably.   😄

ronniegeraghty:
> Pro-tip: When pushing this out to the repositories, making a temporary copy of the common labels file with just this entry speeds things up considerably. 😄

That's such a good idea. Thank you

weshaggard (approved):

PR #8926, opened 2024-09-04
adding cpp parser to public

ckairen:
Parser is required to be in public artifact to consume it by PR pipelines. This change is required to support detecting API changes in PR.


benbp (approved):

tools/apiview/parsers/cpp-api-parser/vcpkg-triplets/x64-windows-static-release.cmake
```diff
@@ -1,4 +1,4 @@
 set(VCPKG_TARGET_ARCHITECTURE x64)
 set(VCPKG_CRT_LINKAGE static)
 set(VCPKG_LIBRARY_LINKAGE static)
-set(VCPKG_BUILD_TYPE release)
+set(VCPKG_BUILD_TYPE release)
```
LarryOsterman, line 4: Removed crlf at end of file?

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -76,14 +76,8 @@ stages:
...
             retryCountOnTaskFailure: 3
-          # - pwsh: |
-          #     ls ${env:VCPKG_ROOT}
-          #     ./vcpkg/vcpkg.exe install atlmfc:x64-windows
-          #   displayName: Install atlmfc.
-          #   workingDirectory: '$(Build.SourcesDirectory)'
-          #   condition: succeeded()
           - pwsh: |
-              cmake.exe -G "Visual Studio 17 2022" -DCMAKE_CXX_STANDARD:STRING="20" -DCMAKE_TOOLCHAIN_FILE:STRING=${env:VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake -DVCPKG_BUILD_TYPE:STRING="release" -DVCPKG_OVERLAY_TRIPLETS:STRING="../vcpkg-triplets" -DMSVC_RUNTIME_LIBRARY:STRING="MultiThreadedDebug" -DVCPKG_TARGET_TRIPLET:STRING="x64-windows-static-release" ..
+              cmake.exe -G "Visual Studio 16 2019" -DCMAKE_CXX_STANDARD:STRING="20" -DCMAKE_TOOLCHAIN_FILE:STRING=${env:VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake -DVCPKG_BUILD_TYPE:STRING="release" -DVCPKG_OVERLAY_TRIPLETS:STRING="../vcpkg-triplets" -DMSVC_RUNTIME_LIBRARY:STRING="MultiThreadedDebug" -DVCPKG_TARGET_TRIPLET:STRING="x64-windows-static-release" ..
```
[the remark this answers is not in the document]
ckairen: I think the issue is that for vscode 2022, we'll need to install ATL/MFC. It's no longer automatically included.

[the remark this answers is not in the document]
benbp: If it's not included now, it may not be included in the future. Eventually 2019 will be extinct. Seems like we should try to solve this problem now, or at least see if it's easy to script+takes a short time.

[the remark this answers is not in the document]
LarryOsterman: Actually there are several problems here:
1) The current RTM version of VS 2022 has a number of issues (working with the VS and 1ES teams to resolve).
2) We need a new APiView based off of clang 17, not clang15 because of changes to STL in VS 2022.

We still don't understand the afxres.h file being missing - the visual studio team says that it was *not* removed from the ATLMFC workload.

As a temporary workaround, this might work however.

LarryOsterman (approved):

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -31,10 +31,11 @@ stages:
...
         timeoutInMinutes: 300
         pool:
-          name: azsdk-pool-mms-win-2022-general
+          name: azsdk-pool-mms-win-2019-general
+          image: azsdk-pool-mms-win-2019-1espt
           os: windows
-          demands:
-          - ImageOverride -equals 1espt-win-2022-cpp-rollback
+          # demands:
+          # - ImageOverride -equals azsdk-pool-mms-win-2022-1espt
```
LarryOsterman, line 38: I think that after my changes, these should be unnecessary.

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -118,12 +119,32 @@ stages:
...
-                      packagePublishDescription: 'Azure C++ SDK API View Parser.'
+                      azureSubscription: 'Azure SDK Artifacts'
+                      ScriptType: 'InlineScript'
+                      azurePowerShellVersion: LatestVersion 
+                      pwsh: true
+                      Inline: |
+                        azcopy copy "$(Pipeline.Workspace)/release-artifacts/*" "https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-cpp/parseazuresdkcpp" --recursive
+                    env:
+                      AZCOPY_AUTO_LOGIN_TYPE: 'PSCRED'
+                  # - task: UniversalPackages@0
```
maririos, line 133: if this is not going to be used, could you please clean it up?

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -118,12 +119,32 @@ stages:
...
-                      vstsFeedPublish: 'internal/AzureSDKForCpp'
-                      vstsFeedPackagePublish: 'parseazuresdkcpp'
-                      versionOption: minor
-                      packagePublishDescription: 'Azure C++ SDK API View Parser.'
+                      azureSubscription: 'Azure SDK Artifacts'
+                      ScriptType: 'InlineScript'
+                      azurePowerShellVersion: LatestVersion 
+                      pwsh: true
+                      Inline: |
+                        azcopy copy "$(Pipeline.Workspace)/release-artifacts/*" "https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-cpp/parseazuresdkcpp" --recursive
```
weshaggard, line 130: Let's add the version to the path here so that we can reference a specific version when we are trying to consume it and we can have side-by-side versions as needed.

ckairen:
latest successful run as reference: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4164828&view=results

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -118,12 +119,30 @@ stages:
                   - checkout: none
                   - download: current
                     artifact: release-artifacts
+                  - pwsh: |
+                      $latestVersion = (azcopy list "https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-cpp/parseazuresdkcpp/") -replace '[\/;].*' | Get-Unique | sort {[version] $_} -Descending | Select-Object -First 1
```
weshaggard, line 123: Isn't the version number somewhere in the source code?

weshaggard, line 123: If you do need to parse and sort versions use our semver library for that https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/SemVer.ps1

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -118,12 +119,30 @@ stages:
...
+                    inputs:
+                      azureSubscription: 'Azure SDK Artifacts'
+                      ScriptType: 'InlineScript'
+                      azurePowerShellVersion: LatestVersion 
+                      pwsh: true
+                      Inline: |
+                        azcopy copy "$(Pipeline.Workspace)/release-artifacts/*" "https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-cpp/parseazuresdkcpp/$(APIVIEW_PARSER_VERSION)" --recursive
+                    env:
+                      AZCOPY_AUTO_LOGIN_TYPE: 'PSCRED'
                   - task: UniversalPackages@0
```
weshaggard, line 139: If we plan to consume using the bloc we should probably stop publishing here.

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -32,6 +32,7 @@ stages:
         timeoutInMinutes: 300
         pool:
           name: azsdk-pool-mms-win-2022-general
+          image: azsdk-pool-mms-win-2022-1espt
```
weshaggard, line 35: We should import image.yml and use the variable defined in there.

tools/apiview/parsers/cpp-api-parser/ci.yml
ckairen: no, the version number was determined by the universal package devops step

ckairen: it should be fine to just use the built-in version type cast though since we don't have any other beta, GA, alpha packages? the format will only be major.minor.build

tools/apiview/parsers/cpp-api-parser/ci.yml
ckairen: so are we only publishing to public then?

tools/apiview/parsers/cpp-api-parser/ci.yml
weshaggard: Ok if there is no version in the code then lets just use the build number as the version. We should just use OfficialBuildId https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/templates/variables/globals.yml#L2

tools/apiview/parsers/cpp-api-parser/ci.yml
weshaggard: I don't see any reason we need it on both.

ckairen: ok make sense

tools/apiview/parsers/cpp-api-parser/ci.yml
ckairen: is the officialBuildId strictly increasing? seems like it's going to be a bit less organized than numerical versioning?

weshaggard: It is based on the date and revision of the build pipeline. So yes it will be unique and ever increasing. It is what we base our nuget package alpha versions on.

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -118,12 +122,20 @@ stages:
...
                     artifact: release-artifacts
-                  - task: UniversalPackages@0
-                    displayName: Publish ApiView Parser.
+                  - pwsh: |
+                      $latestVersion = (azcopy list "https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-cpp/parseazuresdkcpp/") -replace '[\/;].*' | Get-Unique | sort {[version] $_} -Descending | Select-Object -First 1
+                      $major, $minor, $build = $latestVersion.Split('.')
+                      $nextVersion = [version]::new($major, 1 + $minor, $build).ToString()
+                      Write-Host "##vso[task.setvariable variable=APIVIEW_PARSER_VERSION]$nextVersion"
+                    displayName: Determine Next ApiView Parser Version
+                  - task: AzurePowerShell@5
```
weshaggard, line 131: You can use this template https://dev.azure.com/azure-sdk/internal/_git/azure-sdk-build-tools?version=GBmaster&path=/pipelines/steps/binary-signing.yml to do the signing now.

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -119,14 +120,16 @@ stages:
             runOnce: 
               deploy:
                 steps: 
-                  - checkout: none
+                  - checkout: azure-sdk-build-tools
```
weshaggard, line 129: You don't need to checkout. You just need to add a reference. Example https://github.com/Azure/azure-sdk-tools/blob/2e646542c4f21d64f493ef1e1b893a4489a2ecf0/eng/pipelines/templates/stages/1es-redirect.yml#L9

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -119,14 +120,16 @@ stages:
...
                     artifact: release-artifacts
+                  - template: pipelines/steps/binary-signing.yml@azure-sdk-build-tools
+                    parameters:
+                      BinaryPath: $(Pipeline.Workspace)/release-artifacts
                   - pwsh: |
-                      $latestVersion = (azcopy list "https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-cpp/parseazuresdkcpp/") -replace '[\/;].*' | Get-Unique | sort {[version] $_} -Descending | Select-Object -First 1
-                      $major, $minor, $build = $latestVersion.Split('.')
-                      $nextVersion = [version]::new($major, 1 + $minor, $build).ToString()
-                      Write-Host "##vso[task.setvariable variable=APIVIEW_PARSER_VERSION]$nextVersion"
+                      Write-Host "Set Api Parser version to $(OfficialBuildId)"
```
weshaggard, line 137: Why not just use that variable inline below instead of setting another variable in this step? I suspect we can just remove this step.

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -22,16 +22,27 @@ pr:
...
+  - name: ArtifactName
+    value: 'apiview'
+  - name: VcpkgRelease
+    value: '2024.08.23'
+
+resources:
+  repositories:
+    - repository: azure-sdk-build-tools
+      type: git
+      name: internal/azure-sdk-build-tools
```
weshaggard, line 36: Also include the ref https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/templates/stages/1es-redirect.yml#L10 so that these versions remain in sync.

tools/apiview/parsers/cpp-api-parser/ci.yml
```diff
@@ -22,16 +22,27 @@ pr:
       - tools/apiview/parsers/cpp-api-parser
 
 variables:
-  ArtifactName: 'apiview'
-  VcpkgRelease: '2024.08.23'
+  - template: /eng/pipelines/templates/variables/image.yml
+  - template: /eng/pipelines/templates/variables/globals.yml
+  - name: ArtifactName
```
weshaggard, line 27: It doesn't look like ArtifactName or VcpkgRelease are used anywhere and if they aren't please remove.

weshaggard (approved):
Added a couple more comments but otherwise looks good now.

PR #8929, opened 2024-09-04
Refactor prepare pipelines

weshaggard:


eng/common/pipelines/templates/jobs/prepare-pipelines.yml
```diff
@@ -17,14 +22,61 @@ parameters:
...
-    vmImage: ubuntu-22.04
+    name: $(LINUXPOOL)
+    vmImage: $(LINUXVMIMAGE)
+
+  variables:
+    RepositoryPath: "$(System.DefaultWorkingDirectory)/${{ parameters.RepositoryPath }}"
+    ProjectForPRValidation: 'public'
+    ${{ if startsWith(parameters.Repository, 'Azure/azure-sdk-for-java') }}:
+      Prefix: java
+      InternalVariableGroups: '1 58 93'
```
benbp, line 46: Optionally instead of the comment you could declare a `variables` block and just have these group numbers as named variable references instead.

benbp (approved):

eng/common/pipelines/templates/jobs/prepare-pipelines.yml
weshaggard: I didn't want them to be parameters and I couldn't get the variables to work in a reusable template so I just hard-coded them. At least they are all now in one place and so we will only need to update them here now.

weshaggard: I moved the VG's to variables and added an inline PS block to set them up.

benbp (approved):

PR #8940, opened 2024-09-09
Remove dependency of @autorest/csharp

pshao25:
Removed dependency of @autorest/csharp for @autorest/openapi-to-typespec

### DIFF
--- a/tools/tsp-client/CHANGELOG.md
+++ b/tools/tsp-client/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Release
 
+## 2024-09-09 - 0.13.0
+
+- Bumped `@autorest/openapi-to-typespec` version to `0.10.0`.
+- Removed the dependency `@autorest/csharp`.
+
 ## 2024-08-30 - 0.12.2
 
 - Bumped `@autorest/openapi-to-typespec` version to `0.9.1`.
--- a/tools/tsp-client/CHANGELOG.md
+++ b/tools/tsp-client/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release
 
-## 2024-09-09 - 0.13.0
+## 2024-09-10 - 0.13.0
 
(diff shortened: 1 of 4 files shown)


tools/tsp-client/src/commands.ts
```diff
@@ -360,35 +360,7 @@ export async function convertCommand(argv: any): Promise<void> {
...
-    }
-    try {
-      await rename(joinPaths(outputDir, "metadata.json"), joinPaths(outputDir, "resources.json"));
-    } catch (err) {
-      Logger.error(
-        `Error occurred while attempting to rename metadata.json to resources.json: ${err}`,
-      );
-      process.exit(1);
-    }
     args.push("--isArm");
```
catalinaperalta, line 364: So we wont need to generate the resources.json anymore? Folks only have to pass in the --isArm flag?

catalinaperalta (approved):

tools/tsp-client/package-lock.json
```diff
@@ -9,8 +9,7 @@
       "version": "0.12.2",
```
catalinaperalta, line 9: Actually this looks wrong like the package-lock.json wasnt updated correctly. If you do an npm install is should fix all dependencies based on package.json.

catalinaperalta (changes requested):
Please update package-lock.json before merging.

tools/tsp-client/src/commands.ts
pshao25: We generate resources.json in the openapi-to-typespec itself. So we don't need to call `@autorest/csharp`

tools/tsp-client/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Release
 
+## 2024-09-09 - 0.13.0
```
catalinaperalta, line 3:
```suggestion
## 2024-09-10 - 0.13.0
```

catalinaperalta (approved):

PR #8950, opened 2024-09-11
Fix read original changelog content issue

wanlwanl:
Problem: always read stable version's changelog
Fix: read beta version's changelog if the current package is beta


kazrael2119:
this change works well

qiaozha (approved):

PR #8962, opened 2024-09-13
Include empty requestMessage for private endpoint

hallipr:
If the requestMessage property isn't included on the managed private endpoint resource, the resource provider interprets that as setting the requestMessage to null.    The requestMessage property can't be changed after creation, so you must include the property and it must match the existing value if the resource already exists.


scbedd (approved):
nit: mention in comment why you did this.

PR #8975, opened 2024-09-14
Fix Autorest Hang Issue

wanlwanl:
Issue: autorest may hang when generation failed
Fix: add timeout to child process to avoid hang


tools/js-sdk-release-tools/src/common/utils.ts
```diff
@@ -14,6 +14,9 @@ import { SpawnOptions, spawn } from 'child_process';
 // so do NOT change the emitter
 const emitterName = '@azure-tools/typespec-ts';
 
+// 15 minutes in milliseconds unit
+export const defaultChildProcessTimeout = 15 * 60 * 1000;
```
qiaozha, line 18: I feel like 15 minutes is not enough

qiaozha (approved):

PR #9007, opened 2024-09-18
Move resource clean-up to use federated auth

weshaggard:


eng/pipelines/live-test-cleanup.yml
```diff
@@ -45,23 +50,25 @@ parameters:
...
       - DisplayName: AzureCloud Cosmos - Resource Cleanup
+        ServiceConnection: azure-sdk-tests-cosmos
         SubscriptionConfigurations:
           - $(sub-config-azure-cloud-test-resources)
           - $(sub-config-cosmos-azure-cloud-test-resources)
-      - DisplayName: AzureCloud Storage NET - Resource Cleanup
-        SubscriptionConfigurations:
-          - $(sub-config-azure-cloud-test-resources)
-          - $(sub-config-storage-test-resources-net)
+      # - DisplayName: AzureCloud Storage NET - Resource Cleanup
```
weshaggard, line 62: @benbp I couldn't find any usage of this so I didn't convert to a service connection. Do you know if this is being used anywhere? 

I do see https://github.com/search?q=org%3AAzure%20sub-config-storage-test-resources&type=code which is another sub but it isn't in the clean-up list.

benbp: We onboarded a connection to this bring-your-own sub for them, but the never switched over IIRC @weshaggard

benbp: So I think we can delete it

benbp (approved):

PR #9008, opened 2024-09-19
Discovered another issue with `AdditionalValidationPackagesFromPackageSetFn`

scbedd:
What happens when we don't have metadata for a package that we want to include? Welp we're screwed! I don't want to call Get-AllPkgProperties in my custom function when we CAN PASS IT IN.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -154,7 +154,7 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
 
     if ($AdditionalValidationPackagesFromPackageSetFn -and (Test-Path "Function:$AdditionalValidationPackagesFromPackageSetFn"))
     {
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Package-Properties.ps1
```diff
@@ -154,7 +154,7 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
 
     if ($AdditionalValidationPackagesFromPackageSetFn -and (Test-Path "Function:$AdditionalValidationPackagesFromPackageSetFn"))
     {
-        $packagesWithChanges += &$AdditionalValidationPackagesFromPackageSetFn $packagesWithChanges $diff
+        $packagesWithChanges += &$AdditionalValidationPackagesFromPackageSetFn $packagesWithChanges $diff $allPackageProperties
```
scbedd, line 157: Arguments:

- Existing set of packages we consider to be changed
- The diff object
- The entire list of resolved package properties

This way, when we identify packages that we want to use, we can just find the package props for it and add it to the list.

benbp (approved):

PR #9012, opened 2024-09-19
Update TestProxy from Net6 to Net8

JimSuplizio:
I was going to update TestProxy by itself in its own PR but, unfortunately, Azure.Sdk.Tools.Assets.MaintenanceTool.sln has a dependency on TestProxy which required both projects to be updated.

I removed the conditional includes from the TestProxy since net5 is deader than disco. When upgrading the framework to Net8 I also updated Azure.Core and Microsoft.VisualStudio.Web.CodeGeneration.Design to their latest RTM versions.

- `using Microsoft.Build.Tasks;` was removed because it no longer exists but was also a superfluous using.
- response.Headers.Add is now response.Headers.Append
- ['Exception.Exception(SerializationInfo, StreamingContext)' is now obsolete](https://github.com/dotnet/docs/issues/34893)

I also updated the version of Azure.Code and Azure.Storage.Blobs Azure.Sdk.Tools.TestProxy.StorageBlobSample.csproj to later versions which also did not require code changes.

@scbedd - I know once this is merged and published that I need to create another PR to update [eng/common/testproxy/target_version.txt](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/testproxy/target_version.txt) with that version.

### DIFF
--- a/tools/assets-automation/assets-maintenance-tool/Azure.Sdk.Tools.Assets.MaintenanceTool.Tests/Azure.Sdk.Tools.Assets.MaintenanceTool.Tests.csproj
+++ b/tools/assets-automation/assets-maintenance-tool/Azure.Sdk.Tools.Assets.MaintenanceTool.Tests/Azure.Sdk.Tools.Assets.MaintenanceTool.Tests.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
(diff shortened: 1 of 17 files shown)


benbp (approved):
LGTM but I'd get sign-off from @scbedd

JimSuplizio:
> LGTM but I'd get sign-off from @scbedd

@benbp, I wouldn't check this in without @scbedd's approval.

scbedd (approved):
Headers.Add used to `exception` if a duplicate was being added. Now it merely adds the value to the string array for the existing header.

This SHOULDN'T be a behavioral change that causes the target_version update to break anyone, but it's something to watch out for if you see weird test failures.

LMK when the eng/common prs are out, `sdk-for-net` has an additional change necessary to actually use the new proxy.

PR #9017, opened 2024-09-19
[tsp-client] add combine-swaggers command

chrisradek:
This adds a quick and dirty `combine-swaggers` command to 'merge' multiple swagger files into a single swagger file. This is meant to make it easier to compare existing APIs split across multiple files with a TypeSpec-generated API that exists in a single file.

Only top-level fields are merged at this time, and no deduplication of array values is done.

### DIFF
--- a/tools/tsp-client/src/commands.ts
+++ b/tools/tsp-client/src/commands.ts
@@ -3,6 +3,7 @@ import { joinPaths, normalizePath, resolvePath } from "@typespec/compiler";
 import { addSpecFiles, checkoutCommit, cloneRepo, getRepoRoot, sparseCheckout } from "./git.js";
 import {
   createTempDirectory,
+  ensureDirectory,
   getEmitterFromRepoConfig,
   readTspLocation,
   removeDirectory,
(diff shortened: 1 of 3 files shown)


allenjzhang (commented):
Can we support filename wildcard?

tools/tsp-client/src/commands.ts
```diff
@@ -412,3 +413,82 @@ export async function sortSwaggerCommand(argv: any): Promise<void> {
...
+    const swaggerFileCount = swaggerFiles?.length ?? 0;
+    if (swaggerFileCount <= 1) {
+        throw new Error(`2 or more swagger files must be specified, ${swaggerFileCount} provided.`);
+    }
+
+    await ensureDirectory(outputDir);
+    
+    const documents = [];
+    for (const file of swaggerFiles) {
+        if (!(await doesFileExist(file))) {
```
catalinaperalta, line 430: We should specify somewhere that this command only accepts the path to the swagger file and that we do not accept urls to a swagger since we do accept both in some commands.

tools/tsp-client/src/commands.ts
```diff
@@ -412,3 +413,82 @@ export async function sortSwaggerCommand(argv: any): Promise<void> {
...
+    const combined = documents.reduce((a: Record<string | number, any>, b: Record<string | number, any>) => {
+        const result = { ...a };
+        for (const [key, value] of Object.entries(b)) {
+            // if the key doesn't exist in result, just set it to b's value
+            if (!(key in result || typeof result[key] === "undefined" || result[key] === null)) {
+                result[key] = value;
+                continue;
+            }
+    
+            // when b value is null and any a value exists, ignore b value
```
catalinaperalta, line 452: Isnt this a regression if we add the value of a back into the combined swaggers?

catalinaperalta (commented):
Can we add some tests for this command to make sure nobody breaks current functionality in future updates?

Issue #8606, opened 2024-07-11
Update .NET tools to run against latest LTS .NET 8.0


weshaggard:
https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

.NET 6 is ending support in Nov 2024 and we will want to move to .NET 8 before then.

mikeharder:
@weshaggard: Is there anything simple we could do immediately, to allow our tools to roll-forward to .NET 8, so users wouldn't need to install 6 if they already have 8?

JimSuplizio:
Everything has been updated. Closing this out.

PR #9018, opened 2024-09-19
Update projects targeting net6 to net8 - part 2

JimSuplizio:
net6 hits EOL in November, 2024 and this is to update things to net8. This is the second round of projects. 

About tools/pixel-server/PixelServer/Startup.cs: These warnings should have been fixed when the project was upgraded to Net6 and weren't. I'm fixing them now.

- using Microsoft.Extensions.Hosting; was required to use IsDevelopment()
- The UseMVC now requires options => options.EnableEndpointRouting = false set in the services.AddMvcCore call

### DIFF
--- a/tools/secret-management/Azure.Sdk.Tools.SecretRotation.Configuration/RotationConfiguration.cs
+++ b/tools/secret-management/Azure.Sdk.Tools.SecretRotation.Configuration/RotationConfiguration.cs
@@ -1,7 +1,6 @@
 using System.Collections.ObjectModel;
 using Azure.Sdk.Tools.SecretRotation.Core;
 using Microsoft.Extensions.Logging;
-using TimeProvider = Azure.Sdk.Tools.SecretRotation.Core.TimeProvider;
 
 namespace Azure.Sdk.Tools.SecretRotation.Configuration;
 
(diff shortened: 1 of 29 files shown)


tools/secret-management/Azure.Sdk.Tools.SecretRotation.Configuration/RotationConfiguration.cs
```diff
@@ -1,6 +1,7 @@
 using System.Collections.ObjectModel;
 using Azure.Sdk.Tools.SecretRotation.Core;
 using Microsoft.Extensions.Logging;
+using TimeProvider = Azure.Sdk.Tools.SecretRotation.Core.TimeProvider;
```
weshaggard, line 4: Any reason not to just delete https://github.com/Azure/azure-sdk-tools/blob/main/tools/secret-management/Azure.Sdk.Tools.SecretRotation.Core/TimeProvider.cs and use https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider?view=net-8.0 now?

weshaggard (approved):

tools/secret-management/Azure.Sdk.Tools.SecretRotation.Configuration/RotationConfiguration.cs
JimSuplizio: I can but while upgrading the 40+ projects I didn't refactor anything outside of necessary changes for the net6->net8 upgrade. I have the time now so I can do this. I'll push another commit.

PR #9021, opened 2024-09-20
Update projects targeting net6 to net8 - part3

JimSuplizio:
net6 hits EOL in November, 2024 and this is to update things to net8. This is the third round of projects. 

### DIFF
--- a/tools/webhook-router/Azure.Sdk.Tools.WebhookRouter/Routing/Router.cs
+++ b/tools/webhook-router/Azure.Sdk.Tools.WebhookRouter/Routing/Router.cs
@@ -1,4 +1,4 @@
-﻿using Azure.Data.AppConfiguration;
+using Azure.Data.AppConfiguration;
 using Azure.Identity;
 using Azure.Messaging.EventHubs;
 using Azure.Messaging.EventHubs.Producer;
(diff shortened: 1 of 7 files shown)


tools/webhook-router/Azure.Sdk.Tools.WebhookRouter/Routing/Router.cs
```diff
@@ -160,7 +160,7 @@ private async Task<byte[]> ReadAndValidateContentFromGitHubAsync(GitHubRule rule
             return payloadContent;
         }
 
-        private SHA256CryptoServiceProvider sha256 = new SHA256CryptoServiceProvider();
+        private SHA256 sha256 = SHA256.Create();
```
JimSuplizio, line 163: In net8 the SHA256CryptoServiceProvider class is deprecated and they recommend using the SHA256CryptoServiceProvider's base class' Create method.

tools/webhook-router/Azure.Sdk.Tools.WebhookRouter/Routing/Router.cs
```diff
@@ -285,7 +285,7 @@ public async Task RouteAsync(Guid route, HttpRequest request)
                     payload
                     );
 
-                throw ex;
+                throw;
```
JimSuplizio, line 288: `throw ex` vs `throw` - The former loses the stack prior to the throw and, in net8, is now a warning. I'm assuming this was in error as it removes the relevant callstack

weshaggard (approved):

PR #9022, opened 2024-09-20
Fix sparse-checkout commit

weshaggard:
When using the template in prepare-pipelines we need to use `$(Build.SourceVersion)` as our checkout so that people can trigger on their pull requests. 

For the pipeline generation automation pipelines we want to default those to the main branches in the sdk repos.

### DIFF
--- a/eng/pipelines/pipeline-generation-single.yml
+++ b/eng/pipelines/pipeline-generation-single.yml
@@ -2,15 +2,15 @@ pr: none
 trigger: none
 
 parameters:
-- name: Branch
-  type: string
-  default: main
 - name: RepositoryName
   type: string
   default: azure-sdk-for-<lang>
 - name: RepositoryPath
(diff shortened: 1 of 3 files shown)


JimSuplizio (approved):
@weshaggard, as per our conversation I'm approving the PR.

eng/pipelines/pipeline-generation-single.yml
```diff
@@ -24,3 +24,4 @@ jobs:
   parameters:
     Repository: Azure/${{ parameters.RepositoryName }}
     RepositoryPath: ${{ parameters.RepositoryPath }}
+    RepositoryCommitish: ${{ paramters.RepositoryCommitish }}
```
benbp, line 27:
```suggestion
    RepositoryCommitish: ${{ parameters.RepositoryCommitish }}
```

PR #9027, opened 2024-09-23
[dependabot-updates] Overriding the default behavior with dependabot.yml

JackTn:
bugfix for dependabot updates errors for rushjs https://github.com/Azure/azure-sdk-tools/actions/workflows/dependabot/dependabot-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,58 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/src/dotnet/APIView/APIViewWeb/Client"
+    schedule:
+      interval: "daily"
+    
+  - package-ecosystem: "npm"
+    directory: "/src/dotnet/APIView/ClientSPA" 
+    schedule:
+      interval: "daily"
+      
+  - package-ecosystem: "npm"
+    directory: "/tools/apiview/emitters/typespec-apiview"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "npm"
+    directory: "/tools/apiview/parsers/apiview-treestyle-parser-schema"
+    schedule:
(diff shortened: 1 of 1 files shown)


weshaggard:
@JackTn how does this dependabot configuration file fix the rush issue?

JackTn:
@weshaggard 
We're utilizing [overwrite default behavior through dependabot.yml](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file), followed by [Configuring each npm repository in rushjs](https://github.com/dependabot/dependabot-core/issues/2270#issuecomment-688492530). This approach is necessary because, although dependabot doesn't directly support rushjs repositories, it does work with npm repositories. For more details, please refer to  [this](https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories#supported-ecosystems-and-repositories).

weshaggard:
@JackTn do we need to override it for all the directories or just the one where the error lives 2 ERROR `<job_890274753> Error during file fetching; aborting: /tools/sdk-testgen/common/config/rush/package.json not found` https://github.com/Azure/azure-sdk-tools/actions/runs/11016878832

JackTn:
@weshaggard  To implement Dependabot across all our package directories, including C# projects, Go projects, Python projects, and JavaScript projects, and to address the RushJS issue, we need to set up to override settings for each directory. Do we write all setting into dependabot.yml ?

weshaggard:
Perhaps my assumption is incorrect but I would only expect anything that isn't overridden in the config file to keep the defaults. I really do not want to have to maintain a list of all projects in this file.

If we cannot easily override only one directory then I suggest we fix that directory (ie. remove rush and only use plain npm) instead of adding this config file.

weshaggard:
Let's test only the one directory overriding.

weshaggard:
Looks like the one entry worked and unblocked the dependabot update PRs. https://github.com/Azure/azure-sdk-tools/actions/runs/11042246460/job/30674194826

weshaggard:
I guess I should say it seemed to generate updates for test-gen but it doesn't seem to update others. Also all the test-gen updates seem to be failing CI currently so configuring this as npm doesn't correctly update the rush lock files. I was talking with Ray and he is going to check to see if it is still used and if so move it away from Rush.

PR #9028, opened 2024-09-23
Update Stress.Watcher from net6 to net8

JimSuplizio:
net6 hits EOL in November, 2024 and this is to update things to net8.


weshaggard (approved):

benbp:
/check-enforcer override

Stress watcher fails on ARM template provision, as we have updated the cluster version in the config, but the target cluster is a bit too old. I still need to update the cluster to the latest kubernetes version to fix this. You can ignore for now @JimSuplizio

benbp (approved):

PR #9035, opened 2024-09-24
Don't verify changelogs for packages not present in Artifacts

scbedd:
Last blocker :tm: for pipelinev3 in the python repo. Got clean runs other than failing changelog verification against packages that are no longer published.

~~Along the lines of this PR...should we have a `ShouldInvokeCHeck` function that takes an `artifacts` property to check for? EG `skipVerifyChangelog` not true, `publishDocs` not `false. It's not necessarily common for each repo.~~

- ~~If in `CI`, `yq` will be present. `yq` a discovered `ci.yml` on the fly~~
- ~~If not in CI, use PSModules and install the powershell module for yml~~

~~If folks are good with above algo, I'll move the `ShouldInvokeCheck` to `common.ps1`. Otherwise this PR should work as-is.~~

The above may still stand, but we're just calling `yq` and erroring if not present on invoking machine in this PR now.

### DIFF
--- a/eng/common/pipelines/templates/steps/verify-changelogs.yml
+++ b/eng/common/pipelines/templates/steps/verify-changelogs.yml
@@ -6,6 +6,12 @@ parameters:
   default: succeeded()
 
 steps:
+  - pwsh: |
+      Get-ChildItem -Recurse -Path "$(Build.SourcesDirectory)/sdk" -Include ci.yml `
+        | % { Get-Content $_ | yq -o=json > "$($_.FullName).json" }
+    displayName: Generate ci.yml.json
+    condition: ${{ parameters.Condition }}
+
   - task: Powershell@2
     inputs:
       filePath: $(Build.SourcesDirectory)/eng/common/scripts/Verify-ChangeLogs.ps1
--- a/eng/common/pipelines/templates/steps/verify-changelogs.yml
+++ b/eng/common/pipelines/templates/steps/verify-changelogs.yml
@@ -6,12 +6,6 @@ parameters:
   default: succeeded()
 
 steps:
-  - pwsh: |
-      Get-ChildItem -Recurse -Path "$(Build.SourcesDirectory)/sdk" -Include ci.yml `
-        | ForEach-Object { Get-Content $_ -Raw | yq -o=json > "$($_.FullName).json" }
-    displayName: Generate ci.yml.json
-    condition: ${{ parameters.Condition }}
-
   - task: Powershell@2
     inputs:
       filePath: $(Build.SourcesDirectory)/eng/common/scripts/Verify-ChangeLogs.ps1
--- a/eng/common/pipelines/templates/steps/verify-changelogs.yml
+++ b/eng/common/pipelines/templates/steps/verify-changelogs.yml
@@ -8,7 +8,7 @@ parameters:
 steps:
   - pwsh: |
       Get-ChildItem -Recurse -Path "$(Build.SourcesDirectory)/sdk" -Include ci.yml `
-        | % { Get-Content $_ | yq -o=json > "$($_.FullName).json" }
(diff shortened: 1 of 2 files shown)


eng/common/pipelines/templates/steps/verify-changelogs.yml
```diff
@@ -6,6 +6,12 @@ parameters:
   default: succeeded()
 
 steps:
+  - pwsh: |
+      Get-ChildItem -Recurse -Path "$(Build.SourcesDirectory)/sdk" -Include ci.yml `
+        | % { Get-Content $_ | yq -o=json > "$($_.FullName).json" }
```
benbp, line 11:
```suggestion
        | ForEach-Object { Get-Content $_ | yq -o=json > "$($_.FullName).json" }
```

As a convention we try to use the full name for pipeline functions in scripts.

benbp: Also I've found `-Raw` is good when passing gci to convert from json, etc. so it doesn't pass things as newline separated arrays around.

benbp: Unless that's the intent here...

eng/common/scripts/Verify-ChangeLogs.ps1
```diff
@@ -7,6 +7,29 @@ Set-StrictMode -Version 3
...
+                | Select-Object -ExpandProperty name
+
+            if ($packagesCheckingChangeLog -contains $PackageName)
+            {
+                return $true
+            }
+            else {
+                return $false
+            }
+        }
```
benbp, line 29:
```suggestion
            } else {
                return $false
            }
        }
```

eng/common/pipelines/templates/steps/verify-changelogs.yml
scbedd: > As a convention we try to use the full name for pipeline functions in scripts.

Yeah and I usually remember to replace it from local script invocation.

eng/common/pipelines/templates/steps/verify-changelogs.yml
```diff
@@ -6,6 +6,12 @@ parameters:
   default: succeeded()
 
 steps:
+  - pwsh: |
+      Get-ChildItem -Recurse -Path "$(Build.SourcesDirectory)/sdk" -Include ci.yml `
+        | ForEach-Object { Get-Content $_ -Raw | yq -o=json > "$($_.FullName).json" }
+    displayName: Generate ci.yml.json
```
weshaggard, line 12: Why write these to disk as opposed to just doing it in the verify-changelog script directly?

eng/common/scripts/Verify-ChangeLogs.ps1
```diff
@@ -7,6 +7,28 @@ Set-StrictMode -Version 3
 
 . (Join-Path $PSScriptRoot common.ps1)
 
+function ShouldVerifyChangeLog ($ServiceDirectory, $PackageName) {
+    $jsonCiYmlPath = Join-Path $ServiceDirectory "ci.yml.json"
```
weshaggard, line 11: I would suggest just looking for ci.yml and parsing it on the fly right here.

scbedd: Yeah chatted offline.

eng/common/pipelines/templates/steps/verify-changelogs.yml
scbedd: Making the dependency a check in the package, and invoking there directly. Change coming.

scbedd: Ok this is done.

weshaggard (approved):

PR #9043, opened 2024-09-25
[tsp-client] Add @autorest/core as dependency

mikeharder:
Continuation of #8610


catalinaperalta (approved):
Can you also add a changelog entry for this? Also, if you want to release a new version of tsp-client with this pinned feel free to update the package version

catalinaperalta (approved):

PR #9063, opened 2024-09-26
[stress] Sync namespace federated credentials periodically and on startup

benbp:
The original implementation was skipping lookups to see if credentials already existed, on the assumption that deletion was always being handled. This resulted in multiple credentials being created for the same namespace.

Instead we can simply poll to make sure deleted namespaces don't continue to take up a federated identity credential slot as a failsafe. We also sync credential state on startup and keep an in-memory cache, so that namespace deletions/creations re-use existing credentials. There could be some edge cases here with the cache, but I think it's a better trade-off than always querying the managed identity pool every time a namespace is created, as the operation is slow and can increase stress deploy times (we need to wait on completion before deploying otherwise there is a race condition with the credential being created and the deploy container running).

EDIT: my original changes assumed DELETE events weren't being caught, but it's actually because I had added the below code. Now that is removed and the delete event handler method has been added back.

```
if (ns.Status?.Phase == "Terminating")
{
    return;
}
```

### DIFF
--- a/tools/stress-cluster/chaos/examples/stress-deployment-example/Chart.lock
+++ b/tools/stress-cluster/chaos/examples/stress-deployment-example/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: stress-test-addons
   repository: https://stresstestcharts.blob.core.windows.net/helm/
-  version: 0.3.2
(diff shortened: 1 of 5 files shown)


weshaggard (approved):

weshaggard:

benbp:
@weshaggard see my comment edit. I added back the delete handler method (facepalm). Otherwise everything is the same.

PR #9095, opened 2024-10-03
Upgrade revision when revision is requested and if a new parser is re…

praveenkuttappan:
Self-heal approach to upgrade a requested revision if that revision is still using old parser version. This will help to avoid mismatch issue when a revision is still using old parser version, and a new parser is released.


weshaggard (approved):

src/dotnet/APIView/APIViewWeb/Managers/APIRevisionsManager.cs
```diff
@@ -985,5 +987,18 @@ private async Task GenerateAPIRevisionInExternalResource(ReviewListItemModel rev
             }
             return result;
         }
+        private async Task<APIRevisionListItemModel> UpgradeAPIRevisionIfRequired(APIRevisionListItemModel revisionModel)
+        {
+            if (revisionModel == null)
```
benbp, line 992: This is not my codebase so feel free to ignore me, but braceless if statements can be pretty bug prone as it's easy to misread when maintaining/updating.

PR #9099, opened 2024-10-04
Persist oidc env vars in deploy template

benbp:
Example usage: https://github.com/Azure/azure-sdk-for-go/pull/23549

### DIFF
--- a/eng/common/TestResources/deploy-test-resources.yml
+++ b/eng/common/TestResources/deploy-test-resources.yml
@@ -8,6 +8,7 @@ parameters:
   ServiceConnection: not-specified
   ResourceType: test
   UseFederatedAuth: false
+  PersistOidcToken: false
 
 # SubscriptionConfiguration will be splatted into the parameters of the test
 # resources script. It should be JSON in the form:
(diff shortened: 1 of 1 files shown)


eng/common/TestResources/deploy-test-resources.yml
```diff
@@ -41,12 +42,33 @@ steps:
...
+        ARM_CLIENT_ID: $(ARM_CLIENT_ID)
+        ARM_TENANT_ID: $(ARM_TENANT_ID)
+      inputs:
+        azureSubscription: ${{ parameters.ServiceConnection }}
+        addSpnToEnvironment: true
+        scriptLocation: inlineScript
+        scriptType: pscore
+        inlineScript: |
+          Write-Host "##vso[task.setvariable variable=ARM_OIDC_TOKEN;issecret=true]$($env:idToken)"
+          Write-Host "##vso[task.setvariable variable=ARM_CLIENT_ID;issecret=true]$($env:servicePrincipalId)"
```
weshaggard, line 59: Do we need these others given we have them available already?

weshaggard (approved):
We should scope to just OIDC token if possible.

eng/common/TestResources/deploy-test-resources.yml
benbp: Oh I wasn't sure. I know that go is only using the oidc token, whereas net had all three set.

benbp: I think I already replaced the pre/post scripts to use the psboundparameters anyway. So I'll remove.

PR #9102, opened 2024-10-07
Parse CI Artifacts

scbedd:
FYI @JimSuplizio

This is what a result looks like:

```jsonc
{
  "Name": "spring-cloud-azure-starter-monitor",
  "Version": "1.0.0-beta.7",
  "DevVersion": null,
  "DirectoryPath": "sdk/spring/spring-cloud-azure-starter-monitor",
  "ServiceDirectory": "spring",
  "ReadMePath": "sdk/spring/spring-cloud-azure-starter-monitor/README.md",
  "ChangeLogPath": "sdk/spring/spring-cloud-azure-starter-monitor/CHANGELOG.md",
  "Group": "com.azure.spring",
  "SdkType": "spring",
  "IsNewSdk": true,
  "ArtifactName": "spring-cloud-azure-starter-monitor",
  "ReleaseStatus": "Unreleased",
  "IncludedForValidation": false,
  "AdditionalValidationPackages": null,
  // pulled directly from ci.*.yml
  "ArtifactDetails": {
    "name": "spring-cloud-azure-starter-monitor",
    "groupId": "com.azure.spring",
    "safeName": "springcloudazurestartermonitor",
    "skipPublishDocGithubIo": true,
    "skipPublishDocMs": true,
    "skipUpdatePackageJson": true,
    "skipVerifyChangelog": true,
    "releaseInBatch": true
  }
}

```

Which includes pulling from a [`ci.blah.yml`](https://github.com/Azure/azure-sdk-for-java/blob/d9093c6efbadad97a53b22b5eb3d2a00feb66e5b/sdk/spring/ci.monitor.yml#L33)

Confirmed unbroken package-properties on:

- [x] android (new unrelated failure in `main` though)
- [x] ios (no package-properties usage)
- [x] go (no artifacts to save to the packageinfo.json, but no crash either)
- [x] net
- [x] python
- [x] java
- [x] js
- [x] c (doesn't use Package-Properties)
- [x] cpp


eng/common/scripts/Package-Properties.ps1
```diff
@@ -79,6 +82,61 @@ class PackageProps
...
+            }
+            catch {
+              Write-Host "Exception while parsing yml file $($ymlPath): $_"
+            }
+        }
+
+        return $null
+    }
+
+    [void]InitializeCIArtifacts(){
```
scbedd, line 122: Before someone says

> you forgot to add `hidden` to this one too Scott

This is because `azure-sdk-for-js` sets `artifactName` _after_ they create the package prop class instance for the first time, so their `Get-PackagePropertiesFromRepo` needs to _call_ `InitializeCIArtifact` to populate the ArtifactDetails field.

eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -44,4 +44,51 @@ function GetDocsTocDisplayName($pkg) {
...
+#>
+function CompatibleConvertFrom-Yaml {
+  param(
+    [Parameter(Mandatory=$true, ValueFromPipeline=$true)]
+    [string]$Content
+  )
+
+  if (!($Content)) {
+    Write-Error "Content is required."
+    exit 1
```
scbedd, line 77: On second review, I think I will make this throw instead of exit.

eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -44,4 +44,51 @@ function GetDocsTocDisplayName($pkg) {
...
+    . (Join-Path $PSScriptRoot "../../../../eng/common/scripts/Helpers" PSModule-Helpers.ps1)
+    Install-ModuleIfNotInstalled -WhatIf:$false "powershell-yaml" "0.4.1" | Import-Module
+  }
+
+  # Process the content (for example, you could convert from YAML here)
+  if ($yqPresent) {
+      return ($content | yq -o=json | ConvertFrom-Json -AsHashTable)
+  }
+  else {
+      return (ConvertFrom-Yaml (Get-Content -Raw sdk/core/ci.yml))
```
scbedd, line 92: :joy:

eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -44,4 +44,50 @@ function GetDocsTocDisplayName($pkg) {
...
+  }
+
+  # Process the content (for example, you could convert from YAML here)
+  if ($yqPresent) {
+      return ($content | yq -o=json | ConvertFrom-Json -AsHashTable)
+  }
+  else {
+      return ConvertFrom-Yaml $content
+  }
 }
```
scbedd, line 93: Adding newline with a commit addressing any feedback

eng/common/scripts/Package-Properties.ps1
```diff
@@ -79,6 +82,61 @@ class PackageProps
...
+        $RepoRoot = Resolve-Path (Join-Path $PSScriptRoot ".." ".." "..")
+        # $ciFilePath = Join-Path -Path $RepoRoot -ChildPath (Join-Path "sdk" $this.ServiceDirectory "ci.yml")
```
weshaggard, line 124: I assume these commented out lines can just be removed.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -79,6 +82,61 @@ class PackageProps
...
+        if (Test-Path -Path $ymlPath) {
+            try {
+                $content = Get-Content -Raw -Path $ymlPath | CompatibleConvertFrom-Yaml
+                if ($content) {
+                    $artifacts = $this.GetValueSafely($content, @("extends", "parameters", "Artifacts"))
+
+                    $artifactForCurrentPackage = $artifacts | Where-Object { $_["name"] -eq $this.ArtifactName -or $_["name"] -eq $this.Name }
+
+                    if ($artifactForCurrentPackage) {
+                        return [HashTable]$artifactForCurrentPackage
```
danieljurek, line 110: Multiple returns from the pipeline could trip this up. But the engsys would probably also be in a pretty messed up state if someone had multiple artifacts with the same name to begin with...

weshaggard (approved):

eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -44,4 +44,50 @@ function GetDocsTocDisplayName($pkg) {
...
+  )
+
+  if (!($Content)) {
+    throw "Content to parse is a required input."
+  }
+
+  # Initialize any variables or checks that need to be done once
+  $yqPresent = Get-Command 'yq' -ErrorAction SilentlyContinue
+  if (-not $yqPresent) {
+    . (Join-Path $PSScriptRoot "../../../../eng/common/scripts/Helpers" PSModule-Helpers.ps1)
```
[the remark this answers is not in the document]
scbedd: `Install-ModuleIfNotInstalled` comes from it.

eng/common/scripts/Package-Properties.ps1
scbedd: Yeah if this happens it's going to hard error I believe. Because of the class usage the types actually matter, and if we try to convert an array to a HashTable I don't think that's going to work gracefully.

Let me try this with a deliberately broken artifact and let you know what happens.

eng/common/scripts/Package-Properties.ps1
scbedd: damnit I thought I cleaned everything. **shakes fist**

eng/common/scripts/Helpers/Package-Helpers.ps1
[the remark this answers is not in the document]
weshaggard: Oh sorry I miss read this I though you were importing Package-Helpers.ps1 not PSModule-Helpers.ps1.

We could probably shorten the relative path here though.

eng/common/scripts/Package-Properties.ps1
scbedd: My hunch of it being a hard error due to type conversion is correct!

image

```
Exception while parsing yml file C:\repo\azure-sdk-for-js\sdk\core\ci.yml: Cannot convert the "System.Object[]" value of type "System.Object[]" to type "System.Collections.Hashtable".
```

image

eng/common/scripts/Helpers/Package-Helpers.ps1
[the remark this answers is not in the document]
scbedd: Yeah definitely.

JimSuplizio (approved):
I'm approving this. The only concern I have is one that we'd talked about and that's not all languages have artifacts in their yml files. For example, neither Go nor C has them and I'm not currently sure what Rust is doing but they don't seem to have individual ci.yml files in sdk/ subdirectories yet...just something to be aware of.

scbedd:
> I'm approving this. The only concern I have is one that we'd talked about and that's not all languages have artifacts in their yml files. For example, neither Go nor C has them and I'm not currently sure what Rust is doing but they don't seem to have individual ci.yml files in sdk/ subdirectories yet...just something to be aware of.

I will ensure that it works across all the repos before merging any of them 👍

PR #9113, opened 2024-10-08
Use equality function check for persist oidc token step

benbp:
Use equality function check for persist oidc token step

This is breaking the conditional evaluation of `${{ if parameters.PersistOidcToken }}` because it is getting stringified.

### DIFF
--- a/eng/common/TestResources/deploy-test-resources.yml
+++ b/eng/common/TestResources/deploy-test-resources.yml
@@ -42,7 +42,7 @@ steps:
 
   - template: /eng/common/TestResources/setup-environments.yml
 
-  - ${{ if parameters.PersistOidcToken }}:
+  - ${{ if eq(parameters.PersistOidcToken, true) }}:
     - task: AzureCLI@2
       displayName: Set OIDC token
       env:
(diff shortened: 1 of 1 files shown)


eng/common/TestResources/deploy-test-resources.yml
```diff
@@ -44,7 +64,7 @@ steps:
 
   - ${{ if parameters.PersistOidcToken }}:
```
weshaggard, line 65: Are you sure you also don't need to do `${{ if eq(parameters.PersistOidcToken, 'true') }}`?

weshaggard (approved):

eng/common/TestResources/deploy-test-resources.yml
benbp: When it's typed as a boolean, you can do the conditional like this without the `eq` function.

benbp: See https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4210530&view=results as an example run

scbedd (approved):

PR #9117, opened 2024-10-09
Handle only "service" level changes

scbedd:
...where the package directory doesn't have a changed file, but the service does.

This is relevant for changes to `ci.yml` or the like.

I decided that because _everyone_ will hit this, I'd put it in the common Get-PrPkgProperties. I originally was going to place this specificity in `AdditionalValidationPackagesFromPackageSetFn` in each language, to allow folks to select a _subselection_ of packages in the touched service directory. (Eg only include `@azure/core-lro`, instead of all js packages under `core`)

Which would you prefer? I have a bad feeling that adding all the packages will begin overaggressive.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -130,7 +130,7 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
         {
             $filePath = Resolve-Path (Join-Path $RepoRoot $file)
             $shouldInclude = $filePath -like "$pkgDirectory*"
-            $pathComponents = $filel -split "/"
+            $pathComponents = $file -split "/"
 
             # sdk/<service>/<file.extension>
             if ($pathComponents.Length -eq 3 -and $pathComponents[0] -eq "sdk") {
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -134,7 +134,7 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
 
             # sdk/<service>/<file.extension>
             if ($pathComponents.Length -eq 3 -and $pathComponents[0] -eq "sdk") {
-                $servicePkgs = $allPackageProperties | ? { $_.ServiceDirectory -eq $pathComponents[1] } | Foreach-Object { $_.Name }
+                $servicePkgs = $allPackageProperties | Where-Object { $_.ServiceDirectory -eq $pathComponents[1] } | ForEach-Object { $_.Name }
 
                 if ($servicePkgs) {
                     $additionalValidationPackages += $servicePkgs
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -120,7 +120,6 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     $additionalValidationPackages = @()
     $lookup = @{}
 
-    # find all packages that have via comparison with changed files
     foreach ($pkg in $allPackageProperties)
     {
         $pkgDirectory = Resolve-Path "$($pkg.DirectoryPath)"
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Package-Properties.ps1
```diff
@@ -157,6 +168,8 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
         $packagesWithChanges += &$AdditionalValidationPackagesFromPackageSetFn $packagesWithChanges $diff $allPackageProperties
     }
 
+    $packagesWithChanges = $packagesWithChanges | Get-Unique
```
scbedd, line 171: We haven't been bitten by this, but that's luck.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -130,6 +130,17 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
         {
             $filePath = Resolve-Path (Join-Path $RepoRoot $file)
             $shouldInclude = $filePath -like "$pkgDirectory*"
+            $pathComponents = $file -split "/"
+
+            # sdk/<service>/<file.extension>
+            if ($pathComponents.Length -eq 3 -and $pathComponents[0] -eq "sdk") {
```
weshaggard, line 136: While this works it might end up pulling in more packages then necessary. Consider likes like https://github.com/Azure/azure-sdk-for-js/tree/main/sdk/eventhub or https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/storage where the service directory is split up into different pieces. Perhaps it isn't that big of a deal but something we should at least think through.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -130,6 +130,17 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
         {
             $filePath = Resolve-Path (Join-Path $RepoRoot $file)
             $shouldInclude = $filePath -like "$pkgDirectory*"
+            $pathComponents = $file -split "/"
```
weshaggard, line 133: Perhaps it is just me but I prefer a regex for this matching. Maybe we should share the one on line 216 https://github.com/Azure/azure-sdk-tools/pull/9117/files#diff-4958f9a6af9bad851c95447bb46695e45d0f344a6125a5fee7791dd1d36e3a8cR216

eng/common/scripts/Package-Properties.ps1
```diff
@@ -130,6 +130,17 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
...
+
+            # sdk/<service>/<file.extension>
+            if ($pathComponents.Length -eq 3 -and $pathComponents[0] -eq "sdk") {
+                $servicePkgs = $allPackageProperties | Where-Object { $_.ServiceDirectory -eq $pathComponents[1] } | ForEach-Object { $_.Name }
```
weshaggard, line 137: Do we really need to look at all packages here given we are in the context of the outer loop should we just check if that package is in the same service directory?

eng/common/scripts/Package-Properties.ps1
scbedd: Ok. I will say that I'm certain this is probably _safer_ than regex, I'll conform to a standard.

eng/common/scripts/Package-Properties.ps1
scbedd: I squinted at this for a couple minutes. I originally wrote this when it was a in a second pass over just the changed service files. You're absolutely right that we're going to add multiple duplicates by invoking this.

Honestly though @weshaggard , I'm leaning towards moving this to the `Language-Settings` specific, where specific languages will be able to customize this behavior.

eng/common/scripts/Package-Properties.ps1
scbedd: Your concerns are valid. I'm just going to do languages-specific additions to `AdditionalValidationPackagesFromPackageSetFn `

eng/common/scripts/Package-Properties.ps1
scbedd: I tried the regex and it just returns the path as is. Going to continue with what works.

scbedd: image

PR #9118, opened 2024-10-09
Do not hardcode boolean connection verify v2

16234397:
Updated pull request from closed request https://github.com/Azure/azure-sdk-tools/pull/9091


tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
```diff
@@ -2867,7 +2867,105 @@ def visit_import(self, node):
...
+                            )
+
+        except:
+            pass
+        try:
+            # Assigned as a positional
+            if len(node.args) > 0:
+                for arg in node.args:
+                    if type(arg.value) == bool:
+                        for item in node.parent.parent.body:
```
l0lawrence, line 2908: after we figure out arg.value is a bool, why do we have to dig in further to get the arg name? Does it not show you the name in node.args? I see CallNode.args returns a List[Node] -- could you use `arg.repr_name()` maybe?

l0lawrence (commented):
looking great :)

tools/pylint-extensions/azure-pylint-guidelines-checker/pylint_guidelines_checker.py
16234397: I'm not sure what your asking but here is a walkthrough of what it is doing and why it is digging further after we figure out the arg.value is a bool:

```
            if len(node.args) > 0: # here we are visiting every call that has arguments
                for arg in node.args: 
                    if type(arg.value) == bool: # every call that has a bool argument
                        for item in node.parent.parent.body: # going back up to the class because the argument name isn't specified in the call.
                            if item.name == node.func.name:  # checking the name of the call is the same function name
                                for variable in item.args.args: # checking the argument where the function was created
                                    if variable.name == "connection_verify": # where the function was created if a parameter is connection_verify
                                        if item.args.args.index(variable) == node.args.index(arg): # matching the index to see if the bool was the variable assigned to connection_verify
                                            self.add_message(
                                                msgid=f"do-not-hardcode-connection-verify",
                                                node=arg,
                                                confidence=None,
                                            )
```

This one was different, the name of the variable is located in the creation of the function, and the value of the bool assigned is located in where we call the function, two seperate items in the class body. Essentially irs check to see if a bool is allocated in a positional argument, and if it is, if it is allocated to a function creation of connection verify in that class. It is limited because it is only checking at a class level, it can be expanded but the run time of the checker would be expanded considerably, its already verify difficult to run locally against the sdks. 

If you have a different approach that you want it to check a positional argument let me know.

l0lawrence: No worries! this one answered my question: `for item in node.parent.parent.body: # going back up to the class because the argument name isn't specified in the call.` 

Re-reading the issue it looks like this would only apply to connection_verify being a keyword argument so you do not need to check positional arguments

16234397: Positional argument checks have been removed, and a tidy up.

l0lawrence (approved):
lgtm

16234397:
> lgtm

Cool, I have just removed the merge conflicts again :-)

PR #9134, opened 2024-10-10
Use $GITHUB_EVENT_PATH variable instead of inline file writing

benbp:
Reduce github event processor logging

FYI @JimSuplizio I didn't know until now that you can grab the webhook payload from disk (thanks to @hallipr). This saves us a lot of log storage space, as before I was putting the whole payload inline to the bash command (to work around env var max size issues).


hallipr (approved):

JimSuplizio (approved):
@benbp this is great, I didn't know this was a thing either. @hallipr, great find!

weshaggard (approved):

PR #9147, opened 2024-10-11
Changes required for sdk-repo based emitter pipelines

hallipr:
Changes required by pr https://github.com/Azure/azure-sdk-for-net/pull/46511

Chages tested in pipeline run [internal/4224817](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4224817&view=results)

`archetype-typespec-emitter.yml`:
- Created using logic from /eng/pipelines/templates/stages/archetype-autorest-preview.yml

`create-authenticated-npmrc.yml`:
- Moved into eng/common to support `archetype-typespec-emitter.yml`

`CommandInvocation-Helpers.ps1`:
- Converted `Invoke-LoggedCommand` to CmdLet to support -ErrorAction parameter
- Added $AllowedExitCodes parameter to support commands with expected non-zero exit codes
- Added Set-ConsoleEncoding parameter.  This is needed to force utf-8, non-ansi output of some commands.

`New-RegenerateMatrix.ps1`:
- Pass the $OnlyTypeSpec parameter to the language specific implementation of `Get-${Language}-DirectoriesForGeneration` and only filter packages with it in `New-RegenerateMatrix` if there's no language override.

`New-EmitterPackageJson.ps1`:
I wanted to simplify the way branded packages can force emitter peer pinning in emitter-package.json without having to restate all of the unbranded emitter's peer dependencies.  By adding a custom section to package.json we can list the devDependencies that should be forwarded to emitter-package.json for peer pinning.

instead of:
``` json
  "peerDependencies": {
    "@azure-tools/typespec-azure-core": ">=0.46.0 <1.0.0 || ~0.47.0-0",
    "@azure-tools/typespec-client-generator-core": ">=0.46.0 <1.0.0 || ~0.47.0-0",
    "@typespec/compiler": ">=0.60.0 <1.0.0 || ~0.61.0-0",
    "@typespec/http": ">=0.60.0 <1.0.0 || ~0.61.0-0",
    "@typespec/openapi": ">=0.60.0 <1.0.0 || ~0.61.0-0",
    "@typespec/rest": ">=0.60.0 <1.0.0 || ~0.61.0-0",
    "@typespec/versioning": ">=0.60.0 <1.0.0 || ~0.61.0-0"
   }
  "devDependencies": {
    "@azure-tools/typespec-azure-core": "0.46.0",
    "@azure-tools/typespec-client-generator-core": "0.46.2",
    "@typespec/compiler": "0.60.0",
    "@typespec/http": "0.60.0",
    "@typespec/openapi": "0.60.0",
    "@typespec/rest": "0.60.0",
    "@typespec/versioning": "0.60.0",
    
    "@eslint/js": "^9.2.0",
    "@types/node": "~18.13.0"
  }
```

we can:
``` json
  "devDependencies": {
    "@azure-tools/typespec-azure-core": "0.46.0",
    "@azure-tools/typespec-client-generator-core": "0.46.2",
    "@typespec/compiler": "0.60.0",
    "@typespec/http": "0.60.0",
    "@typespec/openapi": "0.60.0",
    "@typespec/rest": "0.60.0",
    "@typespec/versioning": "0.60.0",
    
    "@eslint/js": "^9.2.0",
    "@types/node": "~18.13.0"
  },
  "azure-sdk/emitter-package-json-pinning": [
    "@azure-tools/typespec-azure-core",
    "@azure-tools/typespec-client-generator-core",
    "@typespec/compiler",
    "@typespec/http",
    "@typespec/openapi",
    "@typespec/rest",
    "@typespec/versioning"
  ]
```


eng/pipelines/templates/stages/archetype-autorest-preview.yml
```diff
@@ -204,7 +204,7 @@ stages:
 
     # Create authenticated .npmrc file for publishing
     - ${{ if eq(parameters.PublishInternal, 'true') }}:
-      - template: ../steps/create-authenticated-npmrc.yml
+      - template: ../../../common/pipelines/templates/steps/create-authenticated-npmrc.yml
```
weshaggard, line 207:
```suggestion
      - template: /eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
```

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
```diff
@@ -0,0 +1,452 @@
...
+    - job: Publish
+      steps:
+      - template: /eng/common/pipelines/templates/steps/sparse-checkout.yml
+
+      - download: current
+        artifact: build_artifacts
+        displayName: Download build artifacts
+
+      # Create authenticated .npmrc file for publishing
+      - ${{ if eq(parameters.PublishInternal, 'true') }}:
```
weshaggard, line 167: Publish Internal is kind of misleading as it is going to a public DevOps Feed. You might want to flip this and make the called PublishPublicly or just call this PublishToDevOps or something similar.

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
```diff
@@ -0,0 +1,452 @@
...
+      - ${{ else }}:
+        - pwsh: |
+            "//registry.npmjs.org/:_authToken=$(azure-sdk-npm-token)" | Out-File '.npmrc'
+          displayName: Authenticate .npmrc for npmjs.org
+          workingDirectory: $(buildArtifactsPath)/packages
+
+      # per package, publishing using appropriate tool
+      - ${{ each package in parameters.Packages }}:
+        - ${{ if eq(package.type, 'npm') }}:
+          - pwsh: |
```
weshaggard, line 181: Any reason to not go ahead and use the ESRP Publish?

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
```diff
@@ -0,0 +1,452 @@
...
+              nuGetFeedType: 'internal'
+              publishVstsFeed: '29ec6040-b234-4e31-b139-33dc4287b756/fa8c16a3-dbe0-4de2-a297-03065ec1ba3f'
+
+        - task: PowerShell@2
+          displayName: Create emitter-package.json
+          inputs:
+            pwsh: true
+            filePath: ./eng/common/scripts/typespec/New-EmitterPackageJson.ps1
+            arguments: >
+              -PackageJsonPath '$(buildArtifactsPath)/lock-files/package.json'
```
weshaggard, line 203: This might need to be a parameter to the template if we end up needing to have multiple

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
```diff
@@ -0,0 +1,452 @@
...
+        parameters:
+          BaseRepoOwner: azure-sdk
+          TargetRepoName: $(Build.Repository.Name)
+          BaseRepoBranch: $(branchName)
+          CommitMsg: Update SDK code $(JobKey)
+          WorkingDirectory: $(Build.SourcesDirectory)
+          ScriptDirectory: $(Build.SourcesDirectory)/eng/common/scripts
+
+    - job: Create_PR
+      displayName: Create PR
```
weshaggard, line 325: I didn't think we needed to create a PR any longer given everything is in the one repo. Am I missing a scenario where we would do this?

weshaggard (approved):
A few comments but nothing blocking.

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
```diff
@@ -0,0 +1,452 @@
...
+- stage: Build
+  pool: ${{ parameters.Pool }}
+  jobs:
+  - job: Build
+    steps:
+    # Validate parameters and fail early if invalid
+    - ${{ each package in parameters.Packages }}:
+      - ${{ if notIn(package.type, 'npm', 'nuget') }}:
+        - script: |
+            echo "Package ${{ package.name }} has unsupported type: ${{ package.type }}"
```
benbp, line 91: I think 1es pt figured out how to surface an error message at the pipeline validation level, possibly with something like https://dev.azure.com/azure-sdk/_git/1ESPipelineTemplates?path=/v1/Stages/ReleaseStage.yml&version=GBmain&line=31&lineEnd=34&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

However I guess a downside of that approach is we will omit a check suite to github, so it can be a silent failure when expected as a triggered pipeline.

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
```diff
@@ -0,0 +1,452 @@
...
+
+# Build stage
+# Responsible for building the autorest generator and typespec emitter packages
+# Produces the artifact `build_artifacts` which contains the following:
+#   package-versions.json: Contains a map of package name to version for the packages that were built
+#   overrides.json: Contains npm package version overrides for the emitter and generator
+#   packages/: Contains the packages to publish
+#   lock-files/: Contains package.json and package-lock.json files for use in the test stage to ensure tests are run against the correct dependency versions
+- stage: Build
+  pool: ${{ parameters.Pool }}
```
benbp, line 83: Does this work in a 1es context? Since usually we have had to fully hardcode the pool no matter what.

eng/pipelines/templates/stages/archetype-autorest-preview.yml
hallipr: This template will be used by external repos.  We can't use an `@self` prefix and I think we should stay relative.

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
hallipr: I believe we'll be convention based if we go to multiple. e.g. `emitter-package-{packageName}.json`
\If that's true, I'd want to generate the path from the package rather than have it be configured

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
hallipr: CI checks post regeneration as well as regeneration for uptake as part of the release cycle on prod emitter builds

hallipr: In a prod release run, c# will want to build, test and publish the emitter to npmjs.org, then use the new emitter to regenerate all typespec clients in the repo and have a PR available that updates emitter-package.json and contains the client changes.

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
hallipr: I think we'll need to see.  I would expect it to be resolved before passing it to the underlying template as it's a hard parameter in

hallipr: Or...  I did work in my test runs

weshaggard (approved):

eng/pipelines/templates/stages/archetype-autorest-preview.yml
weshaggard: Won't every place that uses this have /eng/common? I feel like relative paths are much more fragile in devops template usage but both should work so not a big deal.

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
weshaggard: Yes I can see that as part of the post publish job isn't this about validation in a PR context for a emitter update?

eng/pipelines/templates/stages/archetype-autorest-preview.yml
hallipr: absolute paths worked across repos, even for repos without a local eng/common

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
hallipr: added.  thanks

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
hallipr: This is a full lifecycle template used for CI, PR, release and uptake.

eng/common/pipelines/templates/stages/archetype-typespec-emitter.yml
hallipr: I like the `:error` syntax, but ended up removing this whole check when I removed the `Packages` parameter

PR #9164, opened 2024-10-14
Test proxy logs - sanitizer info

HarshaNalluru:
## Test Proxy Logs - Impacted recordings and sanitizers

Introducing improvements to the test proxy logging in terms of which sanitizers are applied and how. 
Key changes below:

### Main changes

1. **RecordEntry.cs**
   - Added `Clone` method to create a copy of the `RecordEntry`, which is used primarily for sanitization logging.

2. **RecordSession.cs**
   - Capturing and comparing the state of entries before and after sanitization to log sanitizer modifications.
   - Added methods to copy `RecordEntry` and log modifications, enhancing traceability and debugging capabilities.

3. **LoggingTests.cs**
   - Enhanced `AssertLogs` method to include a `testType` parameter, adjusting the logic to handle different test types.
   - Updated `PlaybackLogsSanitizedRequest` and `RecordingHandlerLogsSanitizedRequests` tests to use the new `AssertLogs` method.

## Impact
These changes enhance the transparency and traceability of the sanitizers in test proxy by adding detailed logging for request and response modifications. This will aid in debugging and understanding the effects of sanitizers during test sessions.

## Example Output

| Central sanitizer from test proxy | User registered sanitizer |
|-----------------|-----------------|
| <details open>  <summary>Output 1</summary> </details>      | <details open> <summary>Output 2</summary> </details>         |


scbedd:
I like the idea, but we are adding a ton of workload to _every_ single request. Every time we sanitize we write _all the entries_ out. We have to put this behind a debug flag somehow. I'm not certain how big of an impact there will be, so my thoughts is once you get these tests resolved, I will help you get this upgrade to a branch off of `azure-sdk-tools` to publish it.

Easiest way to check performance of a proxy upgrade is to simply try updating the target_version.txt and see how all the repos handle it.

scbedd (changes requested):
Fixing tests, figuring out how to put the heavier consistent tostringing behind a flag, and checking against existing libraries.

Push this to an upstream branch at your convenience so we can easily publish the results as proxy versions.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -145,7 +149,13 @@ public async Task Sanitize(RecordedTestSanitizer sanitizer, bool shouldLock = tr
 
             try
             {
+                var entryPreSanitize = "";
+                var entryPostSanitize = "";
```
scbedd, line 153:
```suggestion
                var entryPostSanitize = string.Empty;
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -155,6 +165,14 @@ public async Task Sanitize(RecordedTestSanitizer sanitizer, bool shouldLock = tr
                 }
             }
         }
+        private void LogSanitizerModification(string sanitizerId, string entryPreSanitize, string entryPostSanitize)
+        {
+            var result = StringComparer.OrdinalIgnoreCase.Compare(entryPreSanitize, entryPostSanitize);
+            if (result != 0)
+            {
+                DebugLogger.LogInformation($"Sanitizer {sanitizerId} modified the entry\nbefore:\n {entryPreSanitize}\nafter:\n {entryPostSanitize}");
```
xirzec, line 173: for custom / user-specified ones can we change the message slightly to indicate the distinction? e.g. `Central sanitizer rule {sanitizerId}` vs `User specified sanitizer rule {sanitizerId}`

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -155,6 +165,14 @@ public async Task Sanitize(RecordedTestSanitizer sanitizer, bool shouldLock = tr
                 }
             }
         }
+        private void LogSanitizerModification(string sanitizerId, string entryPreSanitize, string entryPostSanitize)
+        {
+            var result = StringComparer.OrdinalIgnoreCase.Compare(entryPreSanitize, entryPostSanitize);
```
xirzec, line 170: perhaps we could detect that the entry changed without a string comparison? I assume .NET has some pattern for flagging when objects are modified?

HarshaNalluru: yeah, I did think that. It's just my C# inexperience showing here :P

Will update soon/

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -145,7 +149,13 @@ public async Task Sanitize(RecordedTestSanitizer sanitizer, bool shouldLock = tr
 
             try
             {
+                var entryPreSanitize = "";
+                var entryPostSanitize = "";
+
+                entryPreSanitize += string.Join("", this.Entries.Select(e => e.ToString()));
```
scbedd, line 155: These behind the same check I mentioned above.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -100,7 +101,10 @@ public RecordEntry Lookup(RecordEntry requestEntry, RecordMatcher matcher, IEnum
         {
             foreach (RecordedTestSanitizer sanitizer in sanitizers)
             {
+                var reqEntryPreSanitize = requestEntry.ToString();
```
scbedd, line 104: Place these calls behind `if(DebugLogger.CheckLogLevel(LogLevel.Debug))`.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
HarshaNalluru: Updated to track the property changes in the setters.
Let me know if you like that better.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordEntry.cs
```diff
@@ -26,16 +26,51 @@ public class RecordEntry
...
         public RequestOrResponse Response { get; set; } = new RequestOrResponse();
 
-        public string RequestUri { get; set; }
+        private string _requestUri;
+        public string RequestUri
+        {
+            get { return this._requestUri; }
+            set {
+                // If the requestUri is being modified, set the flag to true
+                if (DebugLogger.CheckLogLevel(LogLevel.Debug) && this._requestUri != value) requestUriIsModified = true;
```
scbedd, line 35: We don't use bare `if` checks anywhere else in the codebase. I'm good with it but it should be at least indented down.

```C#
if (DebugLogger.CheckLogLevel(LogLevel.Debug) && this._requestUri != value)
    requestUriIsModified = true;
```

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -178,5 +223,63 @@ public async Task Sanitize(IEnumerable<RecordedTestSanitizer> sanitizers, bool s
...
             }
         }
+
+        /// <summary>
+        /// Creates a copy of the provided record entry.
+        /// Used primarily for sanitization logging.
+        /// </summary>
+        /// <param name="requestEntry">The record entry to copy.</param>
+        /// <returns>The copied record entry.</returns>
+        private RecordEntry CopyRecordEntry(RecordEntry requestEntry)
```
scbedd, line 233: This should exist in the `RecordEntry` class. Call it `Clone()` IMO.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -167,7 +198,21 @@ public async Task Sanitize(IEnumerable<RecordedTestSanitizer> sanitizers, bool s
...
+                    if (DebugLogger.CheckLogLevel(LogLevel.Debug))
+                    {
+                        entriesPreSanitize = this.Entries.Select(requestEntry => { requestEntry.initializeIsModifiedFlag(); return CopyRecordEntry(requestEntry); }).ToArray();
+                    }
+                    
                     sanitizer.Sanitize(this);
+
+                    if (DebugLogger.CheckLogLevel(LogLevel.Debug))
+                    {
+                        for (int i = 0; i < entriesPreSanitize.Length; i++)
```
scbedd, line 211: Check for entriesPostSanitize[i] actually existing. Remember that sanitizers can clean out an entire entry.

Just looking to avoid out of index exception.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordEntry.cs
```diff
@@ -26,16 +26,54 @@ public class RecordEntry
...
+        }
 
-        public static RecordEntry Deserialize(JsonElement element)
+        /// <summary>
+        /// Initializes the IsModified flags for the RecordEntry instance.
+        /// Primarily used to determine if the RecordEntry has been sanitized.
+        /// </summary>
+        public void initializeIsModifiedFlag()
         {
+            if (DebugLogger.CheckLogLevel(LogLevel.Debug))
```
scbedd, line 67: I appreciate you figuring out how to make this work in `Debug` mode.

Anywhere that you have checks for the `isModified` or not though, I think I'm ok with you pulling back from the "only in debugging" stance just for code clarity.

The reason I say that is because tracking a bool that says whether or not an object is modified won't add more than a few bits per entry. I think it's better to have clean constructor and setter code.

So for `InitializeIsModifiedFlag()`. I don't think this should exist. We should _set it_ during sanitization. 

```
                this.requestUriIsModified = false;
                this.Request.IsModified = new RequestOrResponse.RequestOrResponseIsModified();
                this.Response.IsModified = new RequestOrResponse.RequestOrResponseIsModified();
```

Make `requestUrisModified` _default_ to false. Anything that touches the entry in the `setter` should set it to `true`.

Over in `RequestOrResponse` make `IsModified = new ModificationStatus()` a default value. Can avoid this function entirely.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordEntry.cs
```diff
@@ -26,16 +26,54 @@ public class RecordEntry
...
+            {
+                this.requestUriIsModified = false;
+                this.Request.IsModified = new RequestOrResponse.RequestOrResponseIsModified();
+                this.Response.IsModified = new RequestOrResponse.RequestOrResponseIsModified();
+            }
+        }
+
+        public static RecordEntry Deserialize(JsonElement element)
+            {
             var record = new RecordEntry();
```
scbedd, line 77: We should update `Deserialize` to set `Request.IsModified` and `requestUriIsModified` if putting the defaults on the property directly won't work for you.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -100,7 +102,19 @@ public RecordEntry Lookup(RecordEntry requestEntry, RecordMatcher matcher, IEnum
         {
             foreach (RecordedTestSanitizer sanitizer in sanitizers)
             {
+                RecordEntry reqEntryPreSanitize = null;
+                if (DebugLogger.CheckLogLevel(LogLevel.Debug))
+                {
+                    requestEntry.initializeIsModifiedFlag();
```
scbedd, line 108: If we do these in the deserialization property, we shouldn't need to do it here.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -167,7 +198,21 @@ public async Task Sanitize(IEnumerable<RecordedTestSanitizer> sanitizers, bool s
             {
                 foreach (var sanitizer in sanitizers)
                 {
+                    var entriesPreSanitize = Array.Empty<RecordEntry>();
+                    if (DebugLogger.CheckLogLevel(LogLevel.Debug))
+                    {
+                        entriesPreSanitize = this.Entries.Select(requestEntry => { requestEntry.initializeIsModifiedFlag(); return CopyRecordEntry(requestEntry); }).ToArray();
```
scbedd, line 204: Same cleanup if `InitializeIsModifiedFlag`. Should be possible in defaults. Then we should just `return requestEntry.Clone()`

scbedd (changes requested):
Overall this is pretty damn solid. I requested a couple refactors for clarity, but nothing that is crazy.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordEntry.cs
```diff
@@ -26,16 +26,54 @@ public class RecordEntry
 
         public RequestOrResponse Response { get; set; } = new RequestOrResponse();
 
-        public string RequestUri { get; set; }
+        private string _requestUri;
+        public string RequestUri
+        {
+            get { return this._requestUri; }
+            set {
```
scbedd, line 33: I think you went with the `InitializeIsModifiedFlag` because this `set` is causing the `isModified` to always be true. I'll help come up with a workaround.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordEntry.cs
```diff
@@ -26,16 +26,54 @@ public class RecordEntry
...
+            }
+        }
 
         public bool IsTrack1Recording { get; set; }
 
         public RequestMethod RequestMethod { get; set; }
 
         public int StatusCode { get; set; }
+        // Flag to indicate if the requestUri has been modified
+        public bool requestUriIsModified { get; set; } = false;
```
scbedd, line 49: public properties should be `PascalCase` instead of `camelCase`

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
HarshaNalluru: I didn't do that since I am not actually cloning the whole object, just the pieces I need for the comparisons.
Now that you asked, I'll do it anyway.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
HarshaNalluru: Will do.
That's interesting, I didn't realize `this.Entries[i]` can be cleaned out entirely, can you give me an example/workflow where/how that happens.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
HarshaNalluru: calling it `Reset..` now, but still keeping it as discussed offline.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordEntry.cs
HarshaNalluru: As discussed offline, we want this resetting flags to be as close as possible to Sanitization.
Before sanitization - we take it to be a clean slate
After sanitization - we get the changes from what the flags say

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
HarshaNalluru: Leaving it as is, as discussed offline.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordEntry.cs
HarshaNalluru: Leaving it as is, as discussed offline.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
```diff
@@ -224,62 +227,42 @@ public async Task Sanitize(IEnumerable<RecordedTestSanitizer> sanitizers, bool s
...
             if (entryPostSanitize.Request.IsModified.Body)
             {
-                DebugLogger.LogDebug($"Request Body is modified{Environment.NewLine}before:{Environment.NewLine} {entryPreSanitize.Request.Body}{Environment.NewLine}after:{Environment.NewLine} {entryPostSanitize.Request.Body}{Environment.NewLine}");
+                logMessage += $"Request Body is modified{before}{entryPreSanitize.Request.Body.ToString()}{after}{entryPostSanitize.Request.Body.ToString()}{Environment.NewLine}";
             }
 
             if (entryPostSanitize.Response.IsModified.Body)
             {
-                DebugLogger.LogDebug($"Response Body is modified{Environment.NewLine}before:{Environment.NewLine} {entryPreSanitize.Response.Body}{Environment.NewLine}after:{Environment.NewLine} {entryPostSanitize.Response.Body}{Environment.NewLine}");
+                logMessage += $"Response Body is modified{before}{entryPreSanitize.Response.Body.ToString()}{after}{entryPostSanitize.Response.Body.ToString()}{Environment.NewLine}";
```
HarshaNalluru, line 263: TODO: use something other than `.ToString()` for detailed logging of before and after.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
scbedd: There are couple ways you could do this.

```
for (int i = 0; i < entriesPreSanitize.Length; i++)
```

could be become

```
var max = Math.Max(entriesPreSanitize.Length, entries.Length)
for (int i = 0; i <max ; i++)
```

and then check that the two arrays have equal length. If they aren't that is also a modification 👍

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
scbedd: It's dangerous to go alone, [take this](https://github.com/Azure/azure-sdk-tools/blob/3e14dcaf28df0f1c81948114c2016b28a9991c83/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RequestOrResponse.cs#L39)
```
if (entryPostSanitize.Response.TryGetBodyAsText(out string bodyText)) {
   <use the string>
}.
```

scbedd (changes requested):
Please address [this comment](https://github.com/Azure/azure-sdk-tools/pull/9164/files#r1815543507) and then we're g2g aside from your `ToString()` todo. 

Feel free to use TryGetContentAsText if you'd prefer that to relying on `ToString()`.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
HarshaNalluru: Is it possible that both the arrays go out of sequence too?
Because the logging diff is for each entry of the array and not just the whole array at once.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
HarshaNalluru: updated with this

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
HarshaNalluru: handled all the <, >, = cases individually.
- throws if the entries increase
- does not log entries/modifications if entries decrease
- if entry count remains the same, logs the diff

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordSession.cs
HarshaNalluru: it's hard to track the object changes through setters in cases where the value for a certain key is modified in the dictionary.
Sample snippet..
https://www.programiz.com/online-compiler/6lO13l7zzXYZU
 
Essentially, 
        `obj2.Headers["key2"] = obj2.Headers["key2"].Select(x => x.Replace(x, x+"_modified")).ToArray();`
 
In this case, Headers 's setter is not invoked, which means my modify-flag will not change.
 
The `RequestUri` and the `Body` are not affected the same way.
Still, for consistency, removed the modify-flag tracking in the setters in favor of doing the comparisons in Debug mode.

scbedd (approved):
LGTM @HarshaNalluru !

PR #9181, opened 2024-10-15
Add ability to not trigger `PR` diff generation if the incoming service isn't `auto`

scbedd:
It has been really nice in the `python` repo to maintain the _same_ package properties code, but only activate the new diff logic if the service directory is one we expect.

EG: sdk/pullrequest.yml -> passes servicedirectory `auto`, we call `Save-Package-Properties.yml` with the ServiceDirectoryFilter of `auto`. This ensures the new logic will only trigger if running from a specific servicedirectory.

In the case where we are running _both_ alongside each other for integration checks, this will allow us to use the save-package-properties template while still merging to `main`. Any of the `<language> - <service> - ci` jobs will resolve the same targeted set that they _currently_ do if we set this to `auto` or some other value that only the `pullrequest` build will provide.

### DIFF
--- a/eng/common/pipelines/templates/steps/save-package-properties.yml
+++ b/eng/common/pipelines/templates/steps/save-package-properties.yml
@@ -16,6 +16,12 @@ parameters:
     default: eng/common/scripts
 
 steps:
+  # There will be transitory period for every language repo where the <language> - pullrequest build definition will run
+  # alongside the <language> - <service> - ci definitions. These pullrequest build definitions will have the ServiceDirectory parameter
+  # set to 'auto', which will allow the expanding and contracting based on PR Diff.
+
+  # The other public CI builds will pass a real service directory, which will not activate the PR diff logic and as such will operate
+  # as before this change.
   - ${{ if and(eq(variables['Build.Reason'], 'PullRequest'), eq(parameters.ServiceDirectory, 'auto')) }}:
       - task: Powershell@2
         displayName: Generate PR Diff
--- a/eng/common/pipelines/templates/steps/save-package-properties.yml
+++ b/eng/common/pipelines/templates/steps/save-package-properties.yml
@@ -14,9 +14,12 @@ parameters:
   - name: ScriptDirectory
     type: string
     default: eng/common/scripts
+  - name: ServiceDirectoryFilter
+    type: string
+    default: ""
 
 steps:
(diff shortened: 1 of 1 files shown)


scbedd:
This logic [isn't working as I expect](https://github.com/Azure/azure-sdk-for-js/pull/31420/checks?check_run_id=31588551907)

eng/common/pipelines/templates/steps/save-package-properties.yml
```diff
@@ -14,9 +14,12 @@ parameters:
   - name: ScriptDirectory
     type: string
     default: eng/common/scripts
+  - name: ServiceDirectoryFilter
+    type: string
+    default: ""
 
 steps:
-  - ${{ if eq(variables['Build.Reason'], 'PullRequest') }}:
+  - ${{ if and(eq(variables['Build.Reason'], 'PullRequest'), or(eq(parameters.ServiceDirectoryFilter, ''), eq(parameters.ServiceDirectory, parameters.ServiceDirectoryFilter))) }}:
```
weshaggard, line 22: Perhaps I'm not understanding something but I was expecting to see some sort of check for ServiceDirectory == auto in this condition.

scbedd: My is to add the parameter so we can _pass in_ `auto` as the word that we want to filter on. 

image

I just implemented the condition wrong 👎

scbedd: Oh. I was misreading the build failure. This condition is working the way I expect it. Given above config, this is what a trigger of `js - core - ci` looks like alongside `js - pullrequest`

image

weshaggard: Would the ServiceDirectoryFilter ever be anything aside from empty or auto?

scbedd: > Would the ServiceDirectoryFilter ever be anything aside from empty or auto? If not perhaps just make it a boolean.

Realistically no. I'm just avoiding needing to change this code again when someone has an issue for `auto`.

scbedd: Hardcoded to `auto` and removed param.

weshaggard (approved):

hallipr (approved):

PR #9216, opened 2024-10-17
Move perf pipelines to TME subscription

weshaggard:
- Move perf to TME
- Flip default for resource deployment and removal to true
- Remove unnecessary sub-config files


benbp (approved):

weshaggard:
Tested one perf pipeline https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4252347&view=results

PR #9229, opened 2024-10-22
Remove default sub config sub-config-azure-cloud-test-resources

weshaggard:
We no longer want to default the configuration and instead default from the service connection which had the info we need.

### DIFF
--- a/eng/common/TestResources/remove-test-resources.yml
+++ b/eng/common/TestResources/remove-test-resources.yml
@@ -3,7 +3,7 @@
 
 parameters:
   ServiceDirectory: ''
-  SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources)
+  SubscriptionConfiguration: $(SubscriptionConfiguration)
   ServiceConnection: not-specified
   ResourceType: test
   EnvVars: {}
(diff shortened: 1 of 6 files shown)


eng/common/TestResources/remove-test-resources.yml
```diff
@@ -3,7 +3,7 @@
 
 parameters:
   ServiceDirectory: ''
-  SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources)
+  SubscriptionConfiguration: $(SubscriptionConfiguration)
```
weshaggard, line 6: I've defaulted this because this is what deploy-test-resources outputs. It will fix the usages like we have in perf to ensure we have consistent configuration shared between deploy and remove. Should fix issue like https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4252994&view=logs&j=42ded549-05ee-5cdb-7ba7-7a948a0cc056&t=b938c548-640d-5101-7a83-c666c2d131df

weshaggard:
@benbp the last reference to the default sub config in tools is https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/templates/jobs/stress-cluster-provision.yml#L22. Can you take the item to update it to federated auth and TME?

weshaggard:
perf test pipeline https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4253904&view=logs&j=42ded549-05ee-5cdb-7ba7-7a948a0cc056

template tests pipeline https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4253910&view=results

PR #9230, opened 2024-10-22
Update apiview token prefix

praveenkuttappan:
Update APIvIew token schema and APIView server to support hasPrefixSpace property at token level.


src/dotnet/APIView/APIView/Model/V2/ReviewLine.cs
```diff
@@ -106,6 +106,7 @@ private string ToString(bool includeAllTokens)
             StringBuilder sb = new();
             foreach (var token in filterdTokens)
             {
+                sb.Append(token.HasPrefixSpace == true ? " " : string.Empty);
                 sb.Append(token.Value);
                 sb.Append(token.HasSuffixSpace == true ? " " : string.Empty);
```
tjprescott, line 111: This logic needs to be smarter. If the previous token has "HasSuffixSpace" set to true, then this needs to be ignored so you don't end up with a double space.

src/dotnet/APIView/APIViewWeb/Helpers/CodeFileHelpers.cs
```diff
@@ -253,6 +253,12 @@ private static CodePanelRowData GetCodePanelRowData(CodePanelData codePanelData,
             // Convert ReviewToken to UI required StructuredToken
             foreach (var token in reviewLine.Tokens)
             {
+                if (token.HasPrefixSpace == true)
+                {
+                    var spaceToken = StructuredToken.CreateSpaceToken();
+                    spaceToken.Value = " ";
+                    tokensInRow.Add(spaceToken);
+                }
```
tjprescott, line 261: Same comment here about avoiding a double-space.

tjprescott (changes requested):

src/dotnet/APIView/APIView/Model/V2/ReviewLine.cs
praveenkuttappan: Done.

tjprescott (approved):
Thank you so much @praveenkuttappan!

jeremymeng:
This would help simplify current JS api parser code!

PR #9235, opened 2024-10-22
Support branch name with special char in create review step

praveenkuttappan:
Fix the issue in apiview pipeline step when branch name has &.


maririos (approved):

benbp (approved):

weshaggard:
Change looks fine but what cases allow a space in the branch name?

praveenkuttappan:
> Change looks fine but what cases allow a space in the branch name?

It's not space. It's `&` in the branch name.

PR #9236, opened 2024-10-22
Refactor PackageProps to make the Yaml loading and processing methods common

JimSuplizio:
All I'm doing here is refactoring some of the code that was in PackageProps, specifically around Yaml loading and processing, hoisting it out to be common and usable outside of that class. Specifically, I'm going to be using the yaml loading and processing for the Get-AllPackageInfoFromRepo I'm implementing for Java.

The changes are as follows:

1.  LoadAndCompatibleConvertFrom-Yaml (this was renamed to LoadFrom-Yaml as per feedback) - this was originally part of the PackageProps::ParseYmlForArtifact function. I moved the Test-Path verification of the yml file and try/catch surrounding the call to CompatibleConvertFrom-Yaml into a common function. That way anyone using the CompatibleConvertFrom-Yaml doesn't have to dupe the checks or the try catch.
2. GetValueSafelyFrom-Yaml - this was originally PackageProps::GetValueSafely and hidden. I could have changed the class to unhide and make the function static but it seemed better to be hoisted out and sit alongside the rest of the functions in Package-Helpers. The one thing I changed in this function was that if the key isn't found or the value is null, it writes the key and the full path into the output. In theory this could have thrown but there are scenarios where we can look for something that might not exist in a yml file. An example of this would Java's AdditionalModules which may or may not exist.

For testing, to ensure I was running the code path our pipelines run, I just executed /eng/common/scripts/Save-Package-Properties.ps1 locally against my Java repo to create the PackageInfo json files for various service directories. I pulled the PackageInfo directories directly from their respective pipelines and used Beyond Compare to verify the output. The differences were that the json files from the pipelines had Namespaces and this was expected. In a Java pipeline, the PackageInfo files are created before the build and Namespaces are added to them after the build.

### DIFF
--- a/eng/common/scripts/Helpers/Package-Helpers.ps1
+++ b/eng/common/scripts/Helpers/Package-Helpers.ps1
@@ -91,3 +91,89 @@ function CompatibleConvertFrom-Yaml {
       return ConvertFrom-Yaml $content
   }
 }
+
+<#
+.SYNOPSIS
+Common function that will verify that the YmlFile being loaded exists, load the raw file and
+return the results of CompatibleConvertFrom-Yaml or report an exception and return null if
+there's a problem loading the yml file. The return is the PowerShell HashTable object.
+
+.DESCRIPTION
+Common function that will verify that the YmlFile being loaded exists, load the raw file and
+return the results of CompatibleConvertFrom-Yaml or report an exception and return null if
+there's a problem loading the yml file. This is just to save anyone needing to load yml from
+having to deal with checking the file's existence and ensure that the CompatibleConvertFrom-Yaml
+is made within a try/catch. The return is the PowerShell HashTable object from the
+CompatibleConvertFrom-Yaml call or $null if there was an issue with the convert.
+
+.PARAMETER YmlFile
+The full path of the yml file to load.
+
+.EXAMPLE
+CompatibleLoadAndConvertFrom-Yaml -YmlFile path/to/file.yml
+#>
+function LoadAndCompatibleConvertFrom-Yaml {
+  param(
+    [Parameter(Mandatory=$true)]
+    [string]$YmlFile
+  )
+  if (Test-Path -Path $YmlFile) {
+    try {
+      return Get-Content -Raw -Path $YmlFile | CompatibleConvertFrom-Yaml
+    }
+    catch {
+      Write-Host "CompatibleLoadAndConvertFrom-Yaml::Exception while parsing yml file $($YmlFile): $_"
+    }
+  }
+  else {
+    Write-Host "CompatibleLoadAndConvertFrom-Yaml::YmlFile '$YmlFile' does not exist."
+  }
+  return $null
+}
+
+<#
+.SYNOPSIS
+Given the Hashtable contents of a Yml file and an array of strings representing the keys
+return the value if it exist or null if it doesn't.
+
+.DESCRIPTION
+The Yaml file needs to be loaded via CompatibleConvertFrom-Yaml which returns the file as
+as hashtable. The Keys are basically the path in the yaml file whose value to return, or
+null if it doesn't exist. This function does is safely traverses the path, outputting an
+error if there's an issue or returning the object representing the result if successful.
+This function loops through the Keys safely trying to get values, checking each piece of
+the path to ensure it exists. Normally one would just do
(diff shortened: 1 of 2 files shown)


eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -90,4 +90,90 @@ function CompatibleConvertFrom-Yaml {
...
+The hashtable representing the yaml file contents loaded through LoadAndCompatibleConvertFrom-Yaml
+or CompatibleConvertFrom-Yaml, which is what LoadAndCompatibleConvertFrom-Yaml calls.
+
+.PARAMETER Keys
+String array representation of the path in the yaml file whose value we're trying to retrieve.
+
+.EXAMPLE
+GetValueSafelyFrom-Yaml -YamlContentAsHashtable $YmlFileContent -Keys @("extends", "parameters", "Artifacts")
+#>
+function GetValueSafelyFrom-Yaml {
```
benbp, line 160: Nit: we don't really need to follow powershell cmdlet/func naming conventions for modules. If we were, it would be `Get-ValueSafelyFromYaml` (`Get-`, `Set-`, verbs etc.) but `GetValueSafelyFromYaml` is fine here.

eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -90,4 +90,90 @@ function CompatibleConvertFrom-Yaml {
...
+is made within a try/catch. The return is the PowerShell HashTable object from the
+CompatibleConvertFrom-Yaml call or $null if there was an issue with the convert.
+
+.PARAMETER YmlFile
+The full path of the yml file to load.
+
+.EXAMPLE
+CompatibleLoadAndConvertFrom-Yaml -YmlFile path/to/file.yml
+#>
+function LoadAndCompatibleConvertFrom-Yaml {
```
scbedd, line 115: The naming here is a bit weird. The name in the `.Example` is right, but the name of the actual function is wrong. `LoadAndCompatibleConvertFrom-Yaml`. I think the name in the example makes more sense.

OR, you totally rename it to `LoadFrom-Yaml` or something like that.

eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -90,4 +90,90 @@ function CompatibleConvertFrom-Yaml {
...
+
+<#
+.SYNOPSIS
+Given the Hashtable contents of a Yml file and an array of strings representing the keys
+return the value if it exist or null if it doesn't.
+
+.DESCRIPTION
+The Yaml file needs to be loaded via CompatibleConvertFrom-Yaml which returns the file as
+as hashtable. The Keys are basically the path in the yaml file whose value to return, or
+null if it doesn't exist. This function does is safely traverses the path, outputting an
```
scbedd, line 142:
```suggestion
null if it doesn't exist. This function safely traverses the path, outputting an
```

scbedd (approved):
Gave some feedback, but I don't think anything that would block.

eng/common/scripts/Helpers/Package-Helpers.ps1
JimSuplizio: I'm just going to rename it LoadFrom-Yaml as you suggest :) It's simpler.

eng/common/scripts/Helpers/Package-Helpers.ps1
JimSuplizio: I'm making this change and will push it with the next commit. It's easier since I'm updating the file anyways.

eng/common/scripts/Helpers/Package-Helpers.ps1
JimSuplizio: Renamed with 3rd commit.

eng/common/scripts/Helpers/Package-Helpers.ps1
JimSuplizio: Description updated as per this suggestion with the 3rd commit.

weshaggard (approved):

JimSuplizio:
> Gave some feedback, but I don't think anything that would block.

Yup. I made changes for all your feedback. Everything made sense and the changes were trivial.

PR #9250, opened 2024-10-25
examples-automation, handle corner cases and invalid cases from upstream

weidongxu-microsoft:
1. "directory" in tsp-location could be windows format (with `\`)
2. golang has invalid path like `D:/w/Azure/azure-sdk-for-go/sdk/resourcemanager/computefleet/armcomputefleet/TempTypeSpecFiles/AzureFleet.Management/examples/2024-11-01/Fleets_Update.json` in file, due to bug in codegen -- this result in pipeline failure on doc site
    They had fixed, but we'd better to have more protection.

ref https://github.com/Azure/azure-rest-api-specs-examples/pull/5066

### DIFF
--- a/tools/azure-rest-api-specs-examples-automation/directory/examples_dir.py
+++ b/tools/azure-rest-api-specs-examples-automation/directory/examples_dir.py
@@ -20,8 +20,9 @@ def try_find_resource_manager_example(
             if tsp_dir:
                 # find example under directory
                 # e.g. example_path = "specification/mongocluster/DocumentDB.MongoCluster.Management/examples/2024-03-01-preview/MongoClusters_ListConnectionStrings.json"
+                example_search_glob = f"{path.join(specs_path, tsp_dir)}/**/examples/{example_dir}/{example_filename}"
                 example_paths = glob.glob(
-                    f"{path.join(specs_path, tsp_dir)}/**/examples/{example_dir}/{example_filename}", recursive=True
+                    example_search_glob, recursive=True
                 )
 
                 if len(example_paths) > 0:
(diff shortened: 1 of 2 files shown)


tools/azure-rest-api-specs-examples-automation/directory/examples_dir.py
```diff
@@ -48,7 +49,15 @@ def try_find_resource_manager_example(
...
                             example_dir = path.relpath(example_path, specs_path).replace("\\", "/")
+                else:
+                    raise RuntimeError(f"Example file not found at path {example_search_glob}")
+            else:
+                raise RuntimeError("'directory' property not found in tsp-location.yaml")
         else:
             raise RuntimeError(f"tsp-location.yaml not found in SDK package folder {sdk_package_path}")
 
+    if re.search(r"[:\"*?<>|]+", example_dir):
+        # invalid character in windows path < > : " | ? *
```
weidongxu-microsoft, line 60: `\`, `/` not excluded

`:` excluded because this should be a relative path

tools/azure-rest-api-specs-examples-automation/directory/test_examples_dir.py
```diff
@@ -103,18 +103,45 @@ def test_find_resource_manager_example_typespec(self):
...
             example_dir = try_find_resource_manager_example(
                 path.join(mock_path, "azure-rest-api-specs"),
                 path.join(mock_path, "azure-sdk-for-java/sdk/mongocluster/azure-resourcemanager-mongocluster"),
                 "2024-07-01",
                 "MongoClusters_ListConnectionStrings.json",
             )
 
+    def test_find_resource_manager_example_typespec_windows_directory_separator(self):
+        with create_mock_test_folder() as tmp_dir_name:
+            # use windows directory separator
```
XiaofeiCao, line 116: where do we use that?

XiaofeiCao (approved):

tools/azure-rest-api-specs-examples-automation/directory/test_examples_dir.py
weidongxu-microsoft: saw one in dotnet
https://github.com/Azure/azure-sdk-for-net/blob/Azure.ResourceManager.ContainerOrchestratorRuntime_1.0.0-beta.1/sdk/containerorchestratorruntime/Azure.ResourceManager.ContainerOrchestratorRuntime/tsp-location.yaml

PR #9258, opened 2024-10-26
Add PostGeneration script

scbedd:
This PR will work as-is, with a lot of reduction of complexity over in azure-sdk-for-js.

But frankly, users will have to ensure that whatever the update script is doing _does not need inputs_. This isn't a perfect solution, as users will have to ensure all the input to their script is available on disk or in environment variables.

Ben proposed offline the concept of a new extension on top of the existing matrix IL. Something along the lines of 

```
{
  "matrix": {
    "$SPARSEIMPORT": "<matrix file>",
    // this would just offer full matrix multiplication by the below package list
    "DynamicSetting": ["<pkg list 1>", "<pkg list 2>"]
  }
}
```

If we honored the above, then what we'd do is update the `pullrequest` build matrix pull from a definition like above. We will still need to do a pregeneration matrix update to _populate_ that dynamic setting with the package lists from the current run.

If we went this route, then there wouldn't be custom code to assign to the output matrix in each repo, _however_ we would lose access to any custom logic (like we'd use for direct/indirect packages). My [js - pullrequest](https://github.com/Azure/azure-sdk-for-js/pull/31420) PR has a working implementation with a heavily stripped down `distribute-packages-to-matrix`.

### DIFF
--- a/doc/common/matrix_generator.md
+++ b/doc/common/matrix_generator.md
@@ -111,7 +111,7 @@ To address this limitation, we introduce the [stepList](https://learn.microsoft.
 
 `PreGenerationSteps` allows users to update matrix config json however they like prior to actually invoking the matrix generation. Injected steps are run after the repository checkout, but before any matrix generation is invoked.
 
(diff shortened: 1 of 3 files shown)


scbedd:
This PR has been supplanted by a better proposal in #9281

PR #9281, opened 2024-10-30
`Create-PrJobMatrix`

scbedd:
This PR addresses a lot of the boilerplate code that would be necessary to distribute packages to one or multiple matrix files.

We allow our users to customize their platform using `ci.yml -> MatrixConfigs` params, but we _don't_ honor those in the PR build (which is a single unified matrix).

This PR allows the PR build to automatically use the correct matrix. Here it is [in use.](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4283836&view=results)

Pardon the long runtime, I have further fixes from my other sdk-for-js branch that aren't there, which is why the builds are timing out. Nothing to do with this matrix logic 👍

### DIFF
--- a/eng/common/pipelines/templates/jobs/generate-job-matrix.yml
+++ b/eng/common/pipelines/templates/jobs/generate-job-matrix.yml
@@ -48,6 +48,7 @@ parameters:
 - name: PRMatrixSetting
   type: string
   default: 'ArtifactPackageNames'
+# Mappings to OS name required at template compile time by 1es pipeline templates
 - name: Pools
   type: object
   default:
--- a/eng/common/pipelines/templates/jobs/generate-job-matrix.yml
+++ b/eng/common/pipelines/templates/jobs/generate-job-matrix.yml
@@ -45,9 +45,9 @@ parameters:
 - name: EnablePRGeneration
   type: boolean
   default: false
-- name: PRMatrix
-  type: object
-# Mappings to OS name required at template compile time by 1es pipeline templates
+- name: PRMatrixSetting
+  type: string
+  default: 'ArtifactPackageNames'
 - name: Pools
   type: object
   default:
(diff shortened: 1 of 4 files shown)


eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -176,4 +176,81 @@ function GetValueSafelyFrom-Yaml {
...
+    }
+
+    # Add the current item to the array for this key
+    $groupedDictionary[$key] += $item
+  }
+
+  return $groupedDictionary
+}
+
+function Split-ArrayIntoBatches {
```
scbedd, line 238: @weshaggard  following up on your comment on the js version of this PR. Do you want me to put this in a new file `Array-Helpers.ps1`?

eng/common/pipelines/templates/jobs/generate-job-matrix.yml
```diff
@@ -42,6 +42,11 @@ parameters:
 - name: PreGenerationSteps
   type: stepList
   default: []
+- name: EnablePRGeneration
+  type: boolean
+  default: false
+- name: PRMatrix
```
scbedd, line 48: I will swap PRMatrix to inheriting MatrixConfigs and pass that on to the Create-PrJobMatrix in a further commit.

eng/common/scripts/Helpers/Package-Helpers.ps1
weshaggard: I'm ok with it staying here until we decide if and how we want to reconcile.

eng/common/pipelines/templates/jobs/generate-job-matrix.yml
scbedd: Ok this is done now.

eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -170,10 +170,86 @@ function GetValueSafelyFrom-Yaml {
         $current = $current[$key]
       }
       else {
-        Write-Host "The '$key' part of the path $($Keys -join "/") doesn't exist or is null."
```
scbedd, line 173: @JimSuplizio FYI

This is very spammy when invoking for `MatrixConfigs`, as it is mostly null. We're returning `$null`, so the caller can output if they want to.

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
```diff
@@ -0,0 +1,129 @@
...
+}
+
+if (!(Test-Path $PRMatrixFile)) {
+  Write-Error "PR Matrix file doesn't exist"
+  exit 1
+}
+
+Write-Host "Generating PR job matrix for $PackagePropertiesFolder"
+
+$configs = Get-Content -Raw $PRMatrixFile | ConvertFrom-Json | ForEach-Object { [PSCustomObject]$_ }
```
benbp, line 59: Doesn't ConvertFrom-Json already convert to the PSCustomObject type when you don't pass `-AsHashtable`?

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
```diff
@@ -0,0 +1,129 @@
...
+# calculate general targeting information and create our batches prior to generating any matrix
+$packageProperties = Get-ChildItem -Recurse "$PackagePropertiesFolder" *.json `
+| ForEach-Object { Get-Content -Path $_.FullName | ConvertFrom-Json } `
+| ForEach-Object { [PSCustomObject]$_ }
+
+# set default matrix config for each package if there isn't an override
+$packageProperties | ForEach-Object {
+  if (-not $_.CIMatrixConfigs) {
+    $_.CIMatrixConfigs = $configs
+  }
```
benbp, line 70: If this is only null and not empty string you can simplify to:

```suggestion
  $_.CIMatrixConfigs = $_.CIMatrixConfigs ?? $configs
```

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
scbedd: That's a good point. I'll confirm.

scbedd: image

yep.

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
scbedd: Will wait for the rest of your comments and pull it in.

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
benbp: Yeah I wasn't sure if maybe there were inner nested type conversions you were trying to do or something, but IIRC it's PSCustomObject all the way down.

benbp (approved):

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
```diff
@@ -64,9 +64,7 @@ $packageProperties = Get-ChildItem -Recurse "$PackagePropertiesFolder" *.json `
 
 # set default matrix config for each package if there isn't an override
 $packageProperties | ForEach-Object {
-  if (-not $_.CIMatrixConfigs) {
```
scbedd: I blindly took this and it _definitely_ doesn't work. It's because `CIMatrixConfigs` isn't null, it's Empty Array.

PR #9283, opened 2024-10-30
Support Type Alias Scope Expand Detection

wanlwanl:
Fix https://github.com/Azure/azure-sdk-tools/issues/9269


tools/js-sdk-release-tools/packages/typescript-codegen-breaking-change-detector/src/azure/patch/patch-detection.ts
```diff
@@ -74,7 +74,7 @@ export function patchRoutes(astContext: AstContext): DiffPair[] {
   return [...breakingChangePairs, ...newFeaturePairs];
 }
 
-export function patchUnionType(name: string, astContext: AstContext, assignDirection: AssignDirection): DiffPair[] {
+export function patchTypeAlias(name: string, astContext: AstContext, assignDirection: AssignDirection): DiffPair[] {
```
MaryGao, line 77: Just curious if we consider other type alias cases. Below two are used in our cases. I think other cases are low priority for now.

- export type foo = A & B & C
- export type foo = () => void

Another consideration is if union is not in type alias, can we detect the breaking also?

export interface A {a: string | number | null;}

wanlwanl: > export type foo = A & B & C
export type foo = () => void

Yes, it's convered by `ts.isTypeAssignableTo`

> export interface A {a: string | number | null;}
Yes for new detection tool; No for current tool, but the RLC patch PR already cover this case, so it can be detected.

MaryGao (approved):

PR #9296, opened 2024-10-31
Add AdditionalTags processing

hallipr:
Because public builds don't have permission to edit their own tags but they can add build attachments, we use an attachment to hold "additional tags" and process them in Pipeline Witness


scbedd (approved):
❤️ thank you this Patrick!

PR #9308, opened 2024-11-04
Remove a resolve-path from `Get-PrPkgProperties`

scbedd:
This PR will enable PR invocations of `generate-job-matrix` won't crash when the files don't actually exist on disk.

My only concern is the following scenario:

I have some FUD around a scenario where a `PkgProps` file is generated on `windows` and then we check `-like PkgProps.DirectoryPath` against a path checked out on a different OS.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -168,7 +168,7 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
         $lookup[$lookupKey] = $pkg
 
         foreach ($file in $targetedFiles) {
-            $filePath = Resolve-Path (Join-Path $RepoRoot $file)
+            $filePath = (Join-Path $RepoRoot $file)
             $shouldInclude = $filePath -like "$pkgDirectory*"
             if ($shouldInclude) {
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Package-Properties.ps1
```diff
@@ -168,7 +168,7 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
         $lookup[$lookupKey] = $pkg
 
         foreach ($file in $targetedFiles) {
-            $filePath = Resolve-Path (Join-Path $RepoRoot $file)
+            $filePath = (Join-Path $RepoRoot $file)
```
scbedd, line 171: Because `AllPkgProperties` is populated in this run, we do not need to worry about path separators being cross-platform different.

benbp (approved):

weshaggard:
What cases do we have where these files aren't on disk? Even without the resolve wouldn't that cause trouble?

scbedd:
> What cases do we have where these files aren't on disk? Even without the resolve wouldn't that cause trouble?

During `generate-job-matrix` is where this popped for us.

PR #9326, opened 2024-11-05
Bump powershell yaml versions to 0.4.7

benbp:
Found some pipeline failures installing 0.4.1, not sure if it was transient but figured we might as well update anyways.

I might be breaking myself into a world of hurt here 😅


weshaggard (approved):

weshaggard:
@benbp should we try to also move this package to our DevOps feed as part of this update? 

cc @scbedd

benbp:
@weshaggard good point, I'll change this PR to incorporate that given the recent discussion.

PR #9329, opened 2024-11-05
Do not try to purge keyvaults with purge protection

benbp:
There's no point in trying to purge these keyvaults with delete protection, it will always fail. The error message that comes after  (`InvalidOperation: Operation 'DeletedVaultPurge' is not allowed.`) marks our cleanup builds with a skipped fail which makes it more annoying to see if there are new issues at a glance:
image


benbp:
Closing in favor of https://github.com/Azure/azure-sdk-tools/pull/9333

PR #9330, opened 2024-11-05
Disable `InitializeCIArtifacts` on Non-CI Builds

scbedd:
Until we fix up the loop for usage of `yq` vs `ConvertFrom-Yaml` within `Package-Properties.ps1`.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -106,7 +106,7 @@ class PackageProps {
     }
 
     [void]InitializeCIArtifacts() {
-        if ($env:SYSTEM_TEAMPROJECTID  -or $env:GITHUB_ACTIONS) {
+        if (-not $env:SYSTEM_TEAMPROJECTID  -and -not $env:GITHUB_ACTIONS) {
             return
         }
 
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -106,22 +106,24 @@ class PackageProps {
     }
 
     [void]InitializeCIArtifacts() {
-        if ($env:TF_BUILD -or $env:GITHUB_ACTIONS) {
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Package-Properties.ps1
```diff
@@ -106,20 +106,22 @@ class PackageProps {
...
-        if (-not $this.ArtifactDetails) {
-            foreach ($ciFile in $ciFiles) {
-                $ciArtifactResult = $this.ParseYmlForArtifact($ciFile.FullName)
-                if ($ciArtifactResult) {
-                    $this.ArtifactDetails = [Hashtable]$ciArtifactResult.ArtifactConfig
-                    $this.CIMatrixConfigs = $ciArtifactResult.MatrixConfigs
-                    # if this package appeared in this ci file, then we should
-                    # treat this CI file as the source of the Matrix for this package
-                    break
+        if ($env:TF_BUILD -or $env:GITHUB_ACTIONS) {
```
benbp, line 109:
```suggestion
        if ($env:SYSTEM_TEAMPROJECTID -or $env:GITHUB_ACTIONS) {
```

Either works, but this is more consistent with how we detect CI everywhere else

benbp: Also this is a bit cleaner:

```
if ($env:TF_BUILD -or $env:GITHUB_ACTIONS) {
    return
}
```

scbedd: Did both.

benbp (approved):

scbedd:
When I squash merged a couple of these PRS, the `cleanup branch` was checked. 

[The sync pr creation is failing because we are targeting a branch that doesn't exist](https://github.com/Azure/azure-sdk/tree/sync-eng/common-save-package-properties-9330).

I'm going to check enforcer override and just merge this. Won't mess up the manual merge in the future.

PR #9331, opened 2024-11-05
Finish docs cleanup

JimSuplizio:
The docker-pull-image.yml template was only used by update-docsms-metadata.yml and even then, only Java, JS and Python. All of the aforementioned languages now use the "2docfx" tools do docs generation verification and no longer require the docker images.

The scripts that took the docker image id as a parameter have had that parameter removed and no longer pass that option to any functions/function pointers.

@danieljurek removed the docker images, themselves.


weshaggard (approved):

danieljurek (approved):
JS template release succeeded https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4302407&view=results

PR #9333, opened 2024-11-05
Live test cleanup - sub deployments, immutable containers, purge skip

benbp:
Live test cleanup - sub deployments, immutable containers, purge skip

### DIFF
--- a/eng/common/scripts/Helpers/Resource-Helpers.ps1
+++ b/eng/common/scripts/Helpers/Resource-Helpers.ps1
@@ -342,6 +342,9 @@ function RemoveStorageAccount($Account) {
   }
 
   foreach ($container in $containers) {
+    if (!($container | Get-Member 'BlobContainerProperties')) {
+      continue
+    }
     if ($container.BlobContainerProperties.HasImmutableStorageWithVersioning) {
       try {
         # Use AzRm cmdlet as deletion will only work through ARM with the immutability policies defined on the blobs
@@ -395,7 +398,7 @@ function EnableBlobDeletion($Blob, $Container, $StorageAccountName, $ResourceGro
     $Blob.ICloudBlob.BreakLease()
   }
 
-  if ($container.BlobContainerProperties.HasImmutableStorageWithVersioning) {
+  if (($container | Get-Member 'BlobContainerProperties') -and $container.BlobContainerProperties.HasImmutableStorageWithVersioning) {
     $forceBlobDeletion = $true
   }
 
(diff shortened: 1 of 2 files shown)


eng/common/scripts/Helpers/Resource-Helpers.ps1
```diff
@@ -345,8 +345,9 @@ function RemoveStorageAccount($Account) {
     if ($container.BlobContainerProperties.HasImmutableStorageWithVersioning) {
       try {
         # Use AzRm cmdlet as deletion will only work through ARM with the immutability policies defined on the blobs
-        Remove-AzRmStorageContainer -Name $container.Name -StorageAccountName $Account.StorageAccountName -ResourceGroupName $Account.ResourceGroupName -Force
-        #$container | Remove-AzStorageContainer
+        # Add a retry in case blob deletion has not finished in time for container deletion, but not too many that we end up
+        # getting throttled by ARM/SRP if things are actually in a stuck state
+        Retry -Attempts 1 -Action { Remove-AzRmStorageContainer -Name $container.Name -StorageAccountName $Account.StorageAccountName -ResourceGroupName $Account.ResourceGroupName -Force }
```
weshaggard, line 350: Why only retry 1 time?

eng/scripts/live-test-resource-cleanup.ps1
```diff
@@ -507,6 +518,7 @@ if ($SubscriptionId -and ($originalSubscription -ne $SubscriptionId)) {
 
 try {
   DeleteOrUpdateResourceGroups
+  #DeleteSubscriptionDeployments
```
weshaggard, line 521: I assume you wish to re-enable this before merging.

weshaggard (approved):

eng/common/scripts/Helpers/Resource-Helpers.ps1
benbp: I don't feel 100% on this, but I think when I was trying to delete the container too quickly it would fail, as some blobs with immutability may have been cleaning up in the background. Adding the one retry gives us a 5 second sleep. However I don't want to retry multiple times, because if it is failing for valid reasons it just adds to our throttle count and we'll have that backoff/throttle problem all over again (especially if the failure itself is from throttling, we don't want to keep retrying too much).

eng/scripts/live-test-resource-cleanup.ps1
benbp: Yep disabled for testing. We have hundreds of this right now so I'm manually reconciling since it takes a very long time with that quantity.

benbp:

PR #9339, opened 2024-11-06
Disable adaptive sampling to AppInsights

hallipr:
Adaptive Sampling was causing PipelineWitness to drop a large portion of its happy-path telemetry before sending it to AppInsights.  This made it difficult to search for trace messages related to specific build processing as there was a good chance that the trace logs from processing the build were sampled away.

By disabling adaptive sampling, we're telling the app to send all emitted telemetry to the AppInsights collector for processing.


weshaggard (approved):
Not exactly sure what this means but changes seems reasonable.

PR #9354, opened 2024-11-07
Add step for emitting rate limits for a token

hallipr:


benbp (approved):

eng/common/pipelines/templates/steps/emit-rate-limit-metrics.yml
```diff
@@ -0,0 +1,26 @@
+parameters:
+  - name: GitHubUser
+    type: string
+  - name: GitHubToken
+    type: string
```
weshaggard, line 5: Should we also add some context of where we are logging from? So we can start adding this to various places to see what the limits look like in different contexts?

eng/pipelines/templates/steps/sync-directory.yml
```diff
@@ -170,3 +170,8 @@ steps:
         PushArgs: -f
         WorkingDirectory: $(System.DefaultWorkingDirectory)/${{ repo }}
         ScriptDirectory: $(System.DefaultWorkingDirectory)/$(Build.Repository.Name)/${{ parameters.ScriptDirectory }}
+
+- template: /eng/common/pipelines/templates/steps/emit-rate-limit-metrics.yml
+  parameters:
+    GitHubUser: azure-sdk
+    GitHubToken: $(azuresdk-github-pat)
```
weshaggard, line 177:
```suggestion
    GitHubToken: $(azuresdk-github-pat)

```

weshaggard (approved):
One suggestion but otherwise looks good.

eng/common/pipelines/templates/steps/emit-rate-limit-metrics.yml
benbp: That may already be included in the metric metadata as we process the build info at the same time

JimSuplizio (changes requested):
@hallipr we have a [git-helpers.ps1](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Helpers/git-helpers.ps1) PowerShell script where we've been adding common git functions and I would rather see this added in there, as common code than inline. You might even be able to reuse or slightly modify some of the existing code to add this particular scenario.

hallipr:
> @hallipr we have a [git-helpers.ps1](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Helpers/git-helpers.ps1) PowerShell script where we've been adding common git functions and I would rather see this added in there, as common code than inline. You might even be able to reuse or slightly modify some of the existing code to add this particular scenario.

I'll open another PR to move the code to the helpers script.

eng/pipelines/templates/steps/sync-directory.yml
hallipr: I'll add the line break in another PR when I move the inline script to the github helpers script

eng/common/pipelines/templates/steps/emit-rate-limit-metrics.yml
hallipr: We'd be applying false granularity to the metric.   It'd be context around the job that's checking the metric while the metric being checked is more broad than that context.

PR #9359, opened 2024-11-11
Updates to dotnet example automation for recent changes on dotnet side

ArcturusZhang:
In azure-sdk-for-net, we are changing how the namespace works.
Previously, to avoid `NUnit` related namespaces to be added into the using statements, we write code in this way:
```csharp
[NUnit.Framework.Test]
public async Task Sample_DoSomething() {}
```
Because of a refactor of code in dotnet generator, the above way is no longer possible to write - therefore the new code looks like this:
```csharp
using NUnit.Framework;

[Test]
public async Task Sample_DoSomething() {}
```

Therefore this PR is changing how we collect those namespaces in usings to exclude those `NUnit` namespaces.

Unit test cases are also updated.

Related .Net generator changes here:
- https://github.com/Azure/autorest.csharp/pull/5150
- https://github.com/Azure/autorest.csharp/pull/5107


tools/azure-rest-api-specs-examples-automation/dotnet/test_main.py
```diff
@@ -20,14 +24,15 @@
...
 
 namespace Azure.ResourceManager.ContainerInstance.Samples
 {
     public partial class Sample_ContainerGroupCollection
     {
         // ContainerGroupsListByResourceGroup
-        [NUnit.Framework.Test]
-        [NUnit.Framework.Ignore("Only verifying that the sample builds")]
+        [Test]
+        [Ignore("Only verifying that the sample builds")]
```
ArcturusZhang, line 35: I only changes the first method (this case has multiple methods) to show that we currently work with both old style and new style.

tools/azure-rest-api-specs-examples-automation/dotnet/main.py
```diff
@@ -98,7 +98,10 @@ def get_dotnet_using_statements(lines: List[str]) -> List[str]:
         "using Azure.ResourceManager;\n",
     ]
     for line in lines:
-        if line.startswith("using "):
+        if line.startsWith("using NUnit."):
```
weidongxu-microsoft, line 101: startswith

weidongxu-microsoft (approved):

PR #9379, opened 2024-11-13
Deprecated code and update dependencies

raych1:
- Removed dependencies of swagger-validation-common and azure-js-dev-tools
- Deprecated code for supporting track1 SDK generation
- Deleted invalid tests
- Updated dependencies
- Added pipelines.


weshaggard (approved):

tools/spec-gen-sdk/src/types/sdks.ts
```diff
@@ -0,0 +1,66 @@
...
+ * - https://gist.github.com/raych1/353949d19371b69fb82a10dd70032a51
+ * - https://github.com/Azure/azure-sdk-tools/issues/6327
+ * - https://microsoftapc-my.sharepoint.com/:w:/g/personal/raychen_microsoft_com/EbOAA9SkhQhGlgxtf7mc0kUB-25bFue0EFbXKXS3TFLTQA
+ */
+export type SdkName =
+  | "azure-sdk-for-go"
+  | "azure-sdk-for-java"
+  | "azure-sdk-for-js"
+  | "azure-sdk-for-net"
+  | "azure-sdk-for-net-track2"
```
maririos, line 15: could you remind me the difference between `azure-sdk-for-net` and `azure-sdk-for-net-track2` ?

tools/spec-gen-sdk/src/utils/readme.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+  return result;
+}
+
+/**
+ * Parse the contents of an AutoRest readme.md configuration file and return the parsed swagger to
+ * sdk configuration section.
+ * @param readmeMdFileContents The contents of an AutoRest readme.md configuration file.
+ */
+export function findSwaggerToSDKConfiguration(readmeMdFileContents: string | undefined): ReadmeMdSwaggerToSDKConfiguration | undefined {
```
maririos, line 57: is this something that could be used by other jobs in the specs repo that would make more sense to have in there than in here?

maririos (approved):

tools/spec-gen-sdk/src/types/sdks.ts
weshaggard: I would love to remove this "-track2" as part of this work if possible. IIRC the older was for the old code generator.

maririos: same. This causes so much confusion that if we can use this effort to clean it up, I am all up for it!

mikeharder (changes requested):
Please wait for my review

weshaggard:
@raych1 are we at a point where we can have a buildable pipeline? If so lets try to get it added.

tools/spec-gen-sdk/src/utils/readme.ts
raych1: It might be used in other place where requires to access the SDK configuration in `readme.md`. But we don't have an individual tool staying in spec repo and is published as a library in order to be consumed.

tools/spec-gen-sdk/src/types/sdks.ts
raych1: Thank you for the suggestion. I like it. I will confirm with Arthur if the name can be updated. Otherwise, I prefer to keep the name consistent with what they use.

raych1:
> @raych1 are we at a point where we can have a buildable pipeline? If so lets try to get it added.

That's the next PR I am planning to add.

tools/spec-gen-sdk/src/utils/readme.ts
maririos: Does the flow before this involves cloning the specs repo? What if there is a common script on the spec repo that different tools can call? no need to publish anything

tools/spec-gen-sdk/package.json
```diff
@@ -0,0 +1,129 @@
+{
+  "name": "spec-gen-sdk",
```
mikeharder, line 2: Will this package be published to npmjs?  If so, I believe our convention is to use prefix `@azure-tools`.  Example:

https://github.com/Azure/azure-sdk-tools/blob/main/tools/js-sdk-release-tools/package.json

tools/spec-gen-sdk/package.json
```diff
@@ -0,0 +1,129 @@
...
+    "lib/**/*.d.ts.map",
+    "lib/**/*.handlebars",
+    "images/**/*",
+    "src/**/*.ts",
+    "schemas/**/*.json",
+    "LICENSE",
+    "README.md"
+  ],
+  "license": "MIT",
+  "dependencies": {
```
mikeharder, line 32: This is still a large number of dependencies.  Do you believe they are all needed?

mikeharder (changes requested):

tools/spec-gen-sdk/package.json
raych1: Yes, I walked through the list, and they are being used.

tools/spec-gen-sdk/package.json
raych1: Yes, I updated it along with the pipeline definitions.

tools/spec-gen-sdk/src/utils/readme.ts
raych1: The new design won't have dependency on the spec repo. It makes sense to keep this function in the same code base or reference it from another npm library.

raych1:
@weshaggard 
could you review the pipeline changes?

tools/spec-gen-sdk/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+    ArtifactName: drop
+    PackageJsonPath: $(Build.SourcesDirectory)/tools/spec-gen-sdk
+    BuildStages:
+      - stage: InstallAndBuild
+        variables:
+          - template: /eng/pipelines/templates/variables/globals.yml
+          - template: /eng/pipelines/templates/variables/image.yml
+          - name: NodeVersion
+            value: '20.x'
+          - name: VAR_ARTIFACT_NAME
```
weshaggard, line 35: These variables don't seem to be used so I'd suggest removing them.

tools/spec-gen-sdk/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+pr:
+  branches:
+    include:
+      - main
+  paths:
+    include:
+      - tools/spec-gen-sdk
+
+parameters:
+  - name: Publish
```
weshaggard, line 18: Is there a reason you want to have this publish parameter? It will not actually control the real publishing at https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/templates/stages/archetype-sdk-publish-js.yml#L24. 

I would suggest removing it.

tools/spec-gen-sdk/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 2024-11-14 - 0.1.0
```
weshaggard, line 1: Please also add an entry in the CODEOWNERS file

tools/spec-gen-sdk/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+          - name: VAR_BUILD_ARTIFACT_STAGING_DIRECTORY
+            value: $(Build.ArtifactStagingDirectory)
+        jobs:
+          - job: Build_Linux
+            pool:
+              name: $(LINUXPOOL)
+              image: $(LINUXVMIMAGE)
+              os: linux
+            steps:
+              - template: /tools/spec-gen-sdk/build-spec-gen-sdk.yml
```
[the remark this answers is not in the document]
raych1: No plan at least currently. Let me add them inline.

weshaggard (approved):

raych1:
@mikeharder do you have any further feedback on this PR?

tools/spec-gen-sdk/package.json
mikeharder: It can happen in a later PR, but I think we should try to trim this down.  It seems unnecessarily large for what the tool needs to do.

mikeharder (approved):

tools/spec-gen-sdk/src/types/sdks.ts
raych1: I have initiated the discussion with Shanghai team on this name repurpose and will update this code if decided.

maririos: Thank you!

PR #6521, opened 2023-07-14
Make helper function "Install-ModuleIfNotInstalled" threadsafe

mikeharder:
- Fixes #6514

PR #9392, opened 2024-11-15
Support override of powershell module feeds

benbp:
Adding an override list of ps modules we want to redirect to our devops mirror feed.

Also simplified the module existence check (still thread safe), but I'm thinking that maybe the double existence check was there so we wouldn't have a giant lock request backlog if a ton of threads all do a module check at the same time (given it takes 1 sec to do the check if it's behind the lock). @mikeharder any recollection on this? From https://github.com/Azure/azure-sdk-tools/pull/6521

### DIFF
--- a/eng/common/scripts/Helpers/PSModule-Helpers.ps1
+++ b/eng/common/scripts/Helpers/PSModule-Helpers.ps1
@@ -1,4 +1,3 @@
-$DefaultPSRepositoryUrl = "https://www.powershellgallery.com/api/v2"
 $global:CurrentUserModulePath = ""
 
 function Update-PSModulePathForCI()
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Helpers/PSModule-Helpers.ps1
```diff
@@ -55,68 +55,74 @@ function Install-ModuleIfNotInstalled()
...
+    [string]$repositoryUrl
   )
 
-  # Check installed modules
-  $modules = (Get-Module -ListAvailable $moduleName)
-  if ($version -as [Version]) {
-    $modules = $modules.Where({ [Version]$_.Version -ge [Version]$version })
+  # List of modules+versions we want to replace with internal feed sources for reliability, security, etc.
+  $mirrorFeedOverrides = @{
+    'powershell-yaml;0.4.7' = 'https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-tools/nuget/v2'
```
weshaggard, line 63: Do we care about the version?

Also is there any drawback to query this feed for everything first and then fallback? Do they allow multiple feeds?

benbp: We only care because due to security constraints we can't set the feed as upstream in a way where we can make all the packages available. So unless I mirror every version, this simply reflects which package versions I have manually uploaded to our feed.

I like the idea of querying first and then adding fallbacks though, i.e. if one fails we go to the other. I can go and implement that. And then we wouldn't need to specify versions either.

scbedd (approved):
This seems like a reasonable middle ground until we need to change. I know we could pull the versions available from the feed, but there's something to be said for the determinism of having the `<package>;<version>` key always resolve to a feed we expect.

eng/common/scripts/Helpers/PSModule-Helpers.ps1
```diff
@@ -47,6 +46,65 @@ function Update-PSModulePathForCI()
...
+  $repoUrls = if ($packageFeedOverrides.Contains("${moduleName}")) {
+    @($packageFeedOverrides["${moduleName}"], $DefaultPSRepositoryUrl)
+  } else {
+    @($DefaultPSRepositoryUrl)
+  }
+
+  return $repoUrls
+}
+
+function moduleIsInstalled([string]$moduleName, [string]$version) {
```
weshaggard, line 65: nit:
```suggestion
function ModuleIsInstalled([string]$moduleName, [string]$version) {
```

eng/common/scripts/Helpers/PSModule-Helpers.ps1
```diff
@@ -47,6 +46,65 @@ function Update-PSModulePathForCI()
...
+    if ($modules.Count -gt 0)
+    {
+      Write-Host "Using module $($modules[0].Name) with version $($modules[0].Version)."
+      return $modules[0]
+    }
+  }
+  return $null
+}
+
+function installModule([string]$moduleName, [string]$version, $repoUrl) {
```
weshaggard, line 78: nit:
```suggestion
function InstallModule([string]$moduleName, [string]$version, $repoUrl) {
```

eng/common/scripts/Helpers/PSModule-Helpers.ps1
benbp: I marked it lowercase as a sort of "private" function, given this is a sourced file only.

eng/common/scripts/Helpers/PSModule-Helpers.ps1
benbp: Same as above comment

weshaggard (approved):

benbp:
@scbedd @weshaggard added another commit to unregister the repository after install, it's causing some weird overlap issues with `dotnet tool install` later in the pipelines, possibly b/c the url format has to be slightly different. Still trying to figure out what's actually going on here.


> Installing test-proxy version 1.0.0-dev.20241101.1
Unhandled exception: NuGet.Protocol.Core.Types.FatalProtocolException: Failed to fetch results from V2 feed at '[https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-tools/nuget/v2/FindPackagesById()?id='azure.sdk.tools.testproxy'&semVerLevel=2.0.0'](https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-tools/nuget/v2/FindPackagesById()?id=%27azure.sdk.tools.testproxy%27&semVerLevel=2.0.0%27) with following message : Response status code does not indicate success: 401 (Unauthorized - No local versions of package 'azure.sdk.tools.testproxy'; please provide authentication to access versions from upstream that have not yet been saved to your feed. (DevOps Activity ID: D6280451-EFE3-48AA-BF7C-7A31FBB5EF7E)).
 ---> System.Net.Http.HttpRequestException: Response status code does not indicate success: 401 (Unauthorized - No local versions of package 'azure.sdk.tools.testproxy'; please provide authentication to access versions from upstream that have not yet been saved to your feed. (DevOps Activity ID: D6280451-EFE3-48AA-BF7C-7A31FBB5EF7E)).
   at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()

PR #9393, opened 2024-11-16
Batching adjustments for Create-PRJobMatrix

scbedd:
I noticed something _odd_ about how my PR to use the common create-prmatrix instead of custom `distribute-packages-to-matrix` over in Azure/azure-sdk-for-python#38324 

It looked like only the second batch ran?

This PR resolves the oddness.

Details as to what this fixes:
- batches are stomping on each other.

### DIFF
--- a/eng/common/scripts/job-matrix/job-matrix-functions.ps1
+++ b/eng/common/scripts/job-matrix/job-matrix-functions.ps1
@@ -490,7 +490,7 @@ function CloneOrderedDictionary([System.Collections.Specialized.OrderedDictionar
 function SerializePipelineMatrix([Array]$matrix) {
     $pipelineMatrix = [Ordered]@{}
     foreach ($entry in $matrix) {
-        if ($pipelineMatrix.Contains($entry.Name)) {
+        if ($pipelineMatrix.Contains($entry.name)) {
             Write-Warning "Found duplicate configurations for job `"$($entry.name)`". Multiple values may have been replaced with the same value."
             continue
         }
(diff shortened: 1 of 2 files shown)


eng/common/scripts/job-matrix/job-matrix-functions.ps1
```diff
@@ -490,7 +490,7 @@ function CloneOrderedDictionary([System.Collections.Specialized.OrderedDictionar
 function SerializePipelineMatrix([Array]$matrix) {
     $pipelineMatrix = [Ordered]@{}
     foreach ($entry in $matrix) {
-        if ($pipelineMatrix.Contains($entry.Name)) {
+        if ($pipelineMatrix.Contains($entry.name)) {
```
scbedd, line 493: Bug! I managed to stumble upon this by utilizing an OrderedHashTable on the input, which caused `name` vs `Name` to blow up. I figured out

> oh Ben _never_ intended this to be capital `Name`, it was just working because the type may have been forgiving on capitalization before.

I know this because _in the error output_ you call it lower-case `name`.

scbedd: Yes. Confirmed. HashTables are case insensitive. OrderedHashTables are not. I am passing in a MatrixResult cloned, which turned it into an OrderedHashTable from a HashTable.

I tend to think this should be accessing the right property name anyway, so I should keep it, but I think I can account for this by converting the cloned result to a plain HashTable.

PR #9396, opened 2024-11-18
Fix issues in diff when end of context line has diff

praveenkuttappan:
Diff is showing added line as removed when diff is in end of context line. Empty lines from untouched classes are not removed from node and tree view if parser doesn't set empty lines as related to the node line.


maririos (approved):
Can we add a test for this?

praveenkuttappan:
> Can we add a test for this?

Sure. I will submit a separate PR to add it.

PR #9409, opened 2024-11-20
apiviewgo supports tree tokens

### LINKED ISSUE #3142: Add apiviewgo tests
First thoughts on a couple viable approaches:
1. the autorest.go approach: a test case is sample code that exercises some aspect of `apiviewgo` and the expected JSON output given that code. Running the tests means running `apiviewgo` on each case and inspecting the output. CI fails when there's a diff. Downside of this approach is that it's noisy, doesn't work for PR validation, and introduces some human error risk. Many changes in the output JSON, for example removing redundant whitespace tokens, don't impact reviews. Maintainers must reason about the diffs created by new features and "approve" them by committing updated JSON.
2. unit tests that parse snippets of sample code and assert on the resulting code model. May require refactoring to expose all the code we want to test.

SWAG 3-4 days for option 1, double that for option 2.

### LINKED ISSUE #8654: apiviewgo supports tree tokens

chlowell:
Closes #8654. Examples of the new format: [azblob](https://spa.apiviewstagingtest.com/review/e3f09c889fc54009a2ef911d083f8435?activeApiRevisionId=42515aa76fb34198811ed68428f9df00), [azcore](https://spa.apiviewstagingtest.com/review/6ad8bb9839e546ddb1f3e0fd916df3fc?activeApiRevisionId=6dcee0adab7247719d910391a67e1612)

I fixed a bug in today's parser that creates duplicate types for aliases, however this PR doesn't address other bugs or feature gaps. I haven't found anything newly broken by this PR.

This PR also closes #3142 by adding a new test that generates JSON output and fails when that changes.


src/go/cmd/testdata/test_const/go.mod
jhendrixMSFT, line 1: Are the deleted things tested in some other way now

src/go/cmd/content.go
```diff
@@ -123,7 +123,8 @@ func includesType(s []string, t string) bool {
 	return false
 }
 
-func (c *content) parseSimpleType(tokenList *[]Token) {
+func (c *content) parseSimpleType() []ReviewLine {
```
jhendrixMSFT, line 126: At some point we should add doc comments to the various funcs/methods. Might help future maintenance.

src/go/cmd/models.go
```diff
@@ -3,62 +3,126 @@
...
-type PackageReview struct {
-	Diagnostics []Diagnostic `json:"Diagnostics,omitempty"`
-	Language    string       `json:"Language,omitempty"`
-	Name        string       `json:"Name,omitempty"`
-	Tokens      []Token      `json:"Tokens,omitempty"`
-	Navigation  []Navigation `json:"Navigation,omitempty"`
-	PackageName string       `json:"PackageName,omitempty"`
+// ReviewToken corresponds to an individual token within a ReviewLine
+type ReviewToken struct {
+	HasPrefixSpace  bool `json:"HasPrefixSpace,omitempty"`
```
jhendrixMSFT, line 80: From some comments, I'm assuming this field is optional. What does `omitempty` do for the zero-value (presumably nothing)? Would it help if the optional fields are pointer-to-type?

jhendrixMSFT (approved):

src/go/cmd/models.go
chlowell: `omitempty` prevents marshalling the zero value, so we never output `"HasPrefixSpace": false`. I don't see a case in which a pointer value would help

chlowell:
> Is there any coverage data or apiview scenario coverage data?

We can generate coverage for unit tests but that's statement coverage (currently 89.3%), not useful here because there are so many conditions. We can validate our output against the JSON schema (have done) but there's no scenario coverage a la codegen, I guess because there's no consistency in input (source code) across parsers.

I suppose someone could write language-agnostic tests validating JSON details like "all LineId values are unique" and "all NavigateToId values are also LineId values". @praveenkuttappan, what do you think?

src/go/cmd/testdata/test_output/output.json
```diff
@@ -3052,6 +3029,10 @@
           "HasSuffixSpace": false
         }
       ]
+    },
+    {
+      "IsContextEndLine": true,
+      "Tokens": []
     }
   ]
 }
```
RickWinter, line 3038:
```suggestion
}

```

chidozieononiwu (approved):

PR #9412, opened 2024-11-21
`Create-ApiReview` honors PackageProperty files

scbedd:
See title.

@praveenkuttappan Detect-Api-Changes [explicitly walks the prop files](https://github.com/Azure/azure-sdk-tools/blob/715cecc17bd9175dd5c6e90d4cb9e3cc324587ec/eng/common/scripts/Detect-Api-Changes.ps1#L109) to begin with. So we're already covered there.

Pardon the other spacing updates. My `powershell` vscode extension is aggressive 👍

### DIFF
--- a/eng/common/scripts/Create-APIReview.ps1
+++ b/eng/common/scripts/Create-APIReview.ps1
@@ -3,9 +3,9 @@ Param (
   [Parameter(Mandatory=$True)]
   [array] $ArtifactList,
   [Parameter(Mandatory=$True)]
-  [string] $ArtifactPath,  
+  [string] $ArtifactPath,
   [Parameter(Mandatory=$True)]
-  [string] $APIKey,  
+  [string] $APIKey,
   [string] $SourceBranch,
   [string] $DefaultBranch,
   [string] $RepoName,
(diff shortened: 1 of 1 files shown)


benbp (approved):

eng/common/scripts/Create-APIReview.ps1
```diff
@@ -312,22 +312,27 @@ if ($PackageName)
...
 else
 {
-    # process all packages in the artifact
-    foreach ($artifact in $ArtifactList)
-    {
-        Write-Host "Processing $($artifact.name)"
-        $result = ProcessPackage -packageName $artifact.name
-        $responses[$artifact.name] = $result
+    $propFiles = Get-ChildItem -Path $ConfigFileDir -Recurse -Filter "*.json"
+    $artifactPackages = $ArtifactList | ForEach-Object { $_["name"] }
```
weshaggard, line 320: Shouldn't this be `$_.name` or is this a hashtable object?

eng/common/scripts/Create-APIReview.ps1
```diff
@@ -312,22 +312,27 @@ if ($PackageName)
...
+
+    foreach($propFile in $propFiles) {
+        $artifact = Get-Content $propFile.FullName -Raw | ConvertFrom-Json -AsHashtable
+        if ($artifactPackages -contains $propFile.Name.Replace(".json", ""))
+        {
+            Write-Host "Processing $($artifact.ArtifactName)"
```
weshaggard, line 326: Isn't the artifact name the same as the file or are their cases where they are different? If they are the same, we probably can avoid reading the json file. If they aren't the same we probably should use the artifactname in your condition instead of the file name.

weshaggard (approved):
couple questions but otherwise looks good

eng/common/scripts/Create-APIReview.ps1
scbedd: `$pkgInfo.ArtifactName` == `<package-json-file-name>`, if that follows. I'll add comments for clarity and double check, but I'm fairly certain.

praveenkuttappan (approved):
Looks good

scbedd:
Re-opening this in 2025 with a rebase.

Issue #9208, opened 2024-10-17
Address infinite loop in `gitprocessexception` while invoking `push`


scbedd:
@skadhiresan reported a very interesting issue where the `test-proxy` entered an infinite loop of retrying the `git push` command before erroring. Looking something like **[1]**.

On running the failing `push` command with `debug` mode enabled, the issue became apparent:

```powershell
PS C:\Users\skadhiresan\source\repos\azure-sdk-for-net\sdk\resources\Azure.ResourceManager.Resources> test-proxy push -a ./assets.json
Running proxy version is Azure.Sdk.Tools.TestProxy 20240919.1
git --version
StdOut: git version 2.47.0.windows.1
 
StdErr:
ExitCode: 0
git remote -v
StdOut: origin  https://github.com/skadhiresan/azure-sdk-for-net.git (fetch)
origin  https://github.com/skadhiresan/azure-sdk-for-net.git (push)
 
StdErr:
ExitCode: 0
git remote set-url origin https://github.com/Azure/azure-sdk-assets
StdOut:
StdErr: fatal: detected dubious ownership in repository at 'C:/Users/skadhiresan/source/repos/azure-sdk-for-net/.assets/sztCmI9VWt'
'C:/Users/skadhiresan/source/repos/azure-sdk-for-net/.assets/sztCmI9VWt' is owned by:
        BUILTIN/Administrators (S-1-5-32-544)
but the current user is:
        REDMOND/skadhiresan (S-1-12-1-2625456001-1339269914-2951898773-3550298200)
To add an exception for this directory, call:
 
        git config --global --add safe.directory C:/Users/skadhiresan/source/repos/azure-sdk-for-net/.assets/sztCmI9VWt
 
ExitCode: 128
Exception of type 'Azure.Sdk.Tools.TestProxy.Common.Exceptions.GitProcessException' was thrown.
Unhandled exception:
PS C:\Users\skadhiresan\source\repos\azure-sdk-for-net\sdk\resources\Azure.ResourceManager.Resources>
```

We checked that this issue disappeared if we ran the `test-proxy push` from the same elevated command prompt. It worked flawlesslyl. This proves to me that it's a permissions issue. The **real** problem is that the proxy is hiding these file permissions error behind a wall of  exception output.

This issue covers addressing the infinite loop so that the exception can dump out immediately.

**[1]**
![Image](https://github.com/user-attachments/assets/1cac9d82-ecb3-4c07-bf7e-8af0fd929307)
[...]

PR #9413, opened 2024-11-21
Avoid destructive language-repo origin updates

scbedd:
I'm chasing a hidden `GitProcessException` in #9208 . I have a hunch that that combined with a failure to properly get the target repo is causing `git` to get into an irreconcilable state. When that happens, and `git remote set` runs, it escapes the `<languagerepo>/.assets/sliceId/` folder because there _isn't_  a real git repo initialized there.

when that occurs, and we still invoke `git remote set-url`, that'll override the language repo origin. @alzimmermsft this should at least reduce the impact when something goes wrong. I'm still chasing the real origin of it all though.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
@@ -392,13 +392,30 @@ namespace Azure.Sdk.Tools.TestProxy.Store
         private void SetOrigin(GitAssetsConfiguration config)
         {
             var cloneUrl = GetCloneUrl(config.AssetsRepo, config.RepoRoot);
-            GitHandler.Run($"remote set-url origin {cloneUrl}", config);
+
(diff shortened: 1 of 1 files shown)


benbp (approved):

alzimmermsft:
Just a random thought here given the popularity of using the remote name `origin`, have we thought about using something unique for the azure-sdk-assets remote? Something like `assetsremoteorigin`, or really anything else unique, so it doesn't accidentally clobber existing `origin` remotes?

benbp:
Thinking about it, adding handling to migrate existing asset clones to a new remote within the tool is arguably not worth it (though it would be a pretty small task).

scbedd:
> Thinking about it, adding handling to migrate existing asset clones to a new remote within the tool is arguably not worth it (though it would be a pretty small task).

I think if we begin explicitly specifying an origin in all of the commands, we should be able to transition to using something not `origin`.

That being said once I understand #9208 (task for later today) I think this issue should either disappear or be seriously ameliorated.

PR #9423, opened 2024-11-22
Properly handle when `git` encounters an unrecoverable error

### LINKED ISSUE #9208: Address infinite loop in `gitprocessexception` while invoking `push`

scbedd:
Don't infinitely loop. Resolves #9208

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy.Tests/IntegrationTests/GitStoreIntegrationRestoreTests.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy.Tests/IntegrationTests/GitStoreIntegrationRestoreTests.cs
@@ -304,6 +304,44 @@ namespace Azure.Sdk.Tools.TestProxy.Tests.IntegrationTests
               ""AssetsRepoPrefixPath"": ""pull/scenarios"",
               ""AssetsRepoId"": """",
               ""TagPrefix"": ""main"",
+              ""Tag"": ""INVALID_TAG""
+        }", "Invocation of \"git fetch origin refs/tags/INVALID_TAG:refs/tags/INVALID_TAG\" had a non-zero exit code -1")]
+        [Trait("Category", "Integration")]
+        public async Task InvalidTagThrows(string inputJson, string httpException)
+        {
+            var folderStructure = new string[]
+            {
+                GitStoretests.AssetsJson
+            };
+
+            Assets assets = JsonSerializer.Deserialize<Assets>(inputJson);
+            var testFolder = TestHelpers.DescribeTestFolder(assets, folderStructure);
+
+            try
+            {
+                var jsonFileLocation = Path.Join(testFolder, GitStoretests.AssetsJson);
+
+                var parsedConfiguration = await _defaultStore.ParseConfigurationFile(jsonFileLocation);
+
+                var assertion = await Assert.ThrowsAsync<HttpException>(async () =>
+                {
+                    await _defaultStore.Restore(jsonFileLocation);
+                });
+                Assert.Contains(httpException, assertion.Message);
+            }
+            finally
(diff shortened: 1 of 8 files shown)


benbp (approved):

scbedd:
As much as I hate to admit it, I don't think this is the right solution. The `Environment.Exit` is wrong. The source of the unhandled exception is because we're running in two different "modes" in the proxy.

1. On CLI invocation of `restore` or `push`, we want to report a `git` error to CLI, then early exit from whatever command we're invoking.
2. When the proxy is running in a server, we want to throw a `GitProcessException` (which is an `HttpException`) because our custom ASP.NET middleware will turn that unhandled exception into a clean HTTP response for the user.

Effectively, my solution will cause the server to _straight crash_ if anything goes wrong with any "unhandled" `git` exception.

scbedd:
> As much as I hate to admit it, I don't think this is the right solution. The `Environment.Exit` is wrong. The source of the unhandled exception is because we're running in two different "modes" in the proxy.
> 
> 1. On CLI invocation of `restore` or `push`, we want to report a `git` error to CLI, then early exit from whatever command we're invoking.
> 2. When the proxy is running in a server, we want to throw a `GitProcessException` (which is an `HttpException`) because our custom ASP.NET middleware will turn that unhandled exception into a clean HTTP response for the user.
> 
> Effectively, my solution will cause the server to _straight crash_ if anything goes wrong with any "unhandled" `git` exception. I'm going to refactor so we can _know_ what mode we're in via function signature update.

Added a bit for this exact binary setting to `GitProcessHandler`, and honor it. That addresses the above without entirely rewriting every usage of `GitHandler.Run`.

PR #9440, opened 2024-11-27
Hide output from register/unregister ps module repository

benbp:
Hide output from register/unregister ps module repository

### DIFF
--- a/eng/common/scripts/Helpers/PSModule-Helpers.ps1
+++ b/eng/common/scripts/Helpers/PSModule-Helpers.ps1
@@ -79,7 +79,7 @@ function installModule([string]$moduleName, [string]$version, $repoUrl) {
   $repo = (Get-PSRepository).Where({ $_.SourceLocation -eq $repoUrl })
   if ($repo.Count -eq 0)
   {
-    Register-PSRepository -Name $repoUrl -SourceLocation $repoUrl -InstallationPolicy Trusted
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Helpers/PSModule-Helpers.ps1
```diff
@@ -79,7 +79,7 @@ function installModule([string]$moduleName, [string]$version, $repoUrl) {
   $repo = (Get-PSRepository).Where({ $_.SourceLocation -eq $repoUrl })
   if ($repo.Count -eq 0)
   {
-    Register-PSRepository -Name $repoUrl -SourceLocation $repoUrl -InstallationPolicy Trusted
+    Register-PSRepository -Name $repoUrl -SourceLocation $repoUrl -InstallationPolicy Trusted | Out-Null
```
weshaggard, line 82: Are these the only places? What about Set-PSRepository or Install-Module below?

weshaggard (approved):

eng/common/scripts/Helpers/PSModule-Helpers.ps1
benbp: I wasn't finding issues from those, but I'll add this to those lines to be safe regardless.

benbp: Keeping the output from install-module since it doesn't seem to mess with returns and could be useful error output when it fails.

PR #9460, opened 2024-12-04
Support Test Proxy logging error output into a different log file

alzimmermsft:
Updates the Test Proxy scripts to support outputting error logs into `test-proxy-error.log`. This should also standardize behaviors between Windows and Linux/macOs where `Start-Process` doesn't appear to include error output when `-RedirectStandardError` isn't supplied whereas `&>` for `nohup` will redirect all output to the given location.

With this there will be two files that need to be managed for Test Proxy, so `publish-proxy-logs.yml` is updated appropriately to grab `test-proxy-error.log` as well.


scbedd (approved):
👍 Before we merge this across the repos I/we gotta double check to ensure that we don't have conditional yaml in each pipeline that expects the original output of a single log file.

I don't protest the intent at all.

benbp (approved):

PR #9462, opened 2024-12-04
Add service connection permissions for generated pipelines

weshaggard:


tools/pipeline-generator/Azure.Sdk.Tools.PipelineGenerator/PipelineGenerationContext.cs
```diff
@@ -161,6 +169,83 @@ public async Task<ServiceEndpointHttpClient> GetServiceEndpointClientAsync(Cance
...
+        }
+
+        private HttpClient cachedRawHttpClient = null;
+
+        private async Task<HttpClient> GetRawHttpClient(CancellationToken cancellationToken)
+        {
+            if (this.cachedRawHttpClient == null)
+            {
+                var credential = GetAzureCredentials();
+                var tokenRequestContext = new TokenRequestContext(new[] { "499b84ac-1321-427f-aa17-267ca6975798/.default" });
```
benbp, line 199: Can we make this a named const so it's clear what it is?

weshaggard: I'll add a comment.

tools/pipeline-generator/Azure.Sdk.Tools.PipelineGenerator/Program.cs
```diff
@@ -196,6 +201,45 @@ public async Task<ExitCondition> RunAsync(
...
+                foreach (var serviceConnection in serviceConnectionObjects)
+                {
+                    // Get set of permissions for the service connection
+                    JsonNode pipelinePermissions = await context.GetPipelinePermissionsAsync(serviceConnection.Id, cancellationToken);
+
+                    var pipelines = pipelinePermissions["pipelines"].AsArray();
+                    var pipelineIdsWithPermissions = new HashSet<int>(pipelines.Select(p => p["id"].GetValue<int>()));
+
+                    bool needsUpdate = false;
+                    foreach (var definition in updatedDefinitions)
```
benbp, line 220: Does this mean if we accidentally run the generator against public with the service connection flag, it will add those pipelines to the permissions list? Or am I missing the logic somewhere else that excludes that. If not, I think it would be better to have the convention forced under the hood to prevent client level mistakes.

weshaggard: The service connection lookup is also attached to the team so it would fail for public to find those service connections. However, even if we did find them it should fail to try to add a pipeline from another team project.

benbp (approved):

PR #9466, opened 2024-12-04
Use new generation tool that supports service connections

weshaggard:
Update to use new features in https://github.com/Azure/azure-sdk-tools/pull/9462

Also opt JS and Python out of making public pipelines given we have moved to the new pullrequest model for those.

### DIFF
--- a/eng/common/pipelines/templates/jobs/prepare-pipelines.yml
+++ b/eng/common/pipelines/templates/jobs/prepare-pipelines.yml
@@ -22,8 +22,11 @@ jobs:
     Prefix: 'na'
     DevOpsPath: 'na'
     InternalVariableGroups: ''
+    InternalServiceConnections: ''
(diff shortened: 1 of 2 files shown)


eng/common/pipelines/templates/jobs/prepare-pipelines.yml
```diff
@@ -81,16 +90,19 @@ jobs:
             $internalVariableGroups = '$(AzureSDK_Maven_Release_Pipeline_Secrets) $(Release_Secrets_for_GitHub) $(APIReview_AutoCreate_Configurations)'
           }
           "js" {
+            $generatePublicCIPipeline = 'false'
```
weshaggard, line 93: @scbedd @benbp I've added a mode to disable the public jobs from being created after switching to the new pull request model.

benbp (approved):

PR #9474, opened 2024-12-05
Support new interfaces

raych1:
- input contract
- output contract
- new file logger


raych1:
@weshaggard could you review this PR? Please note that the change to support spec PR scenario including the change of how to calculate the change files in a spec PR will be in a separated pull request.

tools/spec-gen-sdk/src/automation/entrypoint.ts
```diff
@@ -1,120 +1,68 @@
...
 }
 
+type SpecPrInfo = {
+  head: {owner: string; repo: string};
+  base: {owner: string; repo: string};
+}
+
 export type SdkAutoContext = {
   config: SdkAutoOptions;
   octokit: Octokit;
```
weshaggard, line 53: Do we need OctoKit any longer what are we trying to access from github?

weshaggard (approved):
A couple questions but seems reasonable. 

I know this is still a work in progress but it will be much easier to understand once we see the reporting in action from the various scenarios.

tools/spec-gen-sdk/src/automation/entrypoint.ts
```diff
@@ -1,120 +1,68 @@
...
   getGithubAccessToken: (owner: string) => Promise<string>;
```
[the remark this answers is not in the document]
raych1: I would expect this to be deleted when refactoring the SDK breaking change related code.

tools/spec-gen-sdk/src/automation/entrypoint.ts
raych1: Same as the above reply.

PR #9493, opened 2024-12-06
Optimize Yml Extraction

scbedd:
Attempt dos. Creating this PR to avoid the merge conflicts on the original PR: #9332

### DIFF
--- a/eng/common/scripts/Helpers/PSModule-Helpers.ps1
+++ b/eng/common/scripts/Helpers/PSModule-Helpers.ps1
@@ -64,7 +64,7 @@ function Get-ModuleRepositories([string]$moduleName) {
 }
 
 function moduleIsInstalled([string]$moduleName, [string]$version) {
-  if ($null -eq $global:InstalledModules) {
+  if ($null -eq $script:InstalledModules) {
     $script:InstalledModules = @{}
   }
 
--- a/eng/common/scripts/Helpers/PSModule-Helpers.ps1
+++ b/eng/common/scripts/Helpers/PSModule-Helpers.ps1
@@ -64,7 +64,7 @@ function Get-ModuleRepositories([string]$moduleName) {
 }
 
 function moduleIsInstalled([string]$moduleName, [string]$version) {
-  if ($null -eq $script:InstalledModules) {
+  if (-not (Test-Path variable:script:InstalledModules)) {
     $script:InstalledModules = @{}
   }
 
--- a/eng/common/scripts/Helpers/PSModule-Helpers.ps1
+++ b/eng/common/scripts/Helpers/PSModule-Helpers.ps1
@@ -121,6 +121,14 @@ function installModule([string]$moduleName, [string]$version, $repoUrl) {
   return $modules[0]
 }
 
+function InstallAndImport-ModuleIfNotInstalled([string]$module, [string]$version) {
+  if ($null -eq (moduleIsInstalled $module $version)) {
+    Install-ModuleIfNotInstalled -WhatIf:$false $module $version | Import-Module
+  } elseif (!(Get-Module -Name $module)) {
+    Import-Module $module
+  }
+}
+
 # Manual test at eng/common-tests/psmodule-helpers/Install-Module-Parallel.ps1
 # If we want to use another default repository other then PSGallery we can update the default parameters
 function Install-ModuleIfNotInstalled() {
--- a/eng/common/scripts/Helpers/PSModule-Helpers.ps1
+++ b/eng/common/scripts/Helpers/PSModule-Helpers.ps1
@@ -62,8 +62,18 @@ function Get-ModuleRepositories([string]$moduleName) {
   return $repoUrls
 }
 
+if ($null -eq $script:InstalledModules) {
+  $script:InstalledModules = @{}
+}
+
 function moduleIsInstalled([string]$moduleName, [string]$version) {
-  $modules = (Get-Module -ListAvailable $moduleName)
+  if ($script:InstalledModules.ContainsKey("${moduleName}")) {
+    $modules = @($script:InstalledModules["${moduleName}"])
+  }
+  else {
+    $modules = (Get-Module -ListAvailable $moduleName)
+  }
+
   if ($version -as [Version]) {
     $modules = $modules.Where({ [Version]$_.Version -ge [Version]$version })
     if ($modules.Count -gt 0)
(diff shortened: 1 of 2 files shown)


eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -68,28 +68,21 @@ Get-Content -Raw path/to/file.yml | CompatibleConvertFrom-Yaml
...
   if (!($Content)) {
     throw "Content to parse is a required input."
   }
 
-  # Initialize any variables or checks that need to be done once
-  $yqPresent = Get-Command 'yq' -ErrorAction SilentlyContinue
-  if (-not $yqPresent) {
+  if (-not $script:IsYamlInstalled) {
     . (Join-Path $PSScriptRoot PSModule-Helpers.ps1)
     Install-ModuleIfNotInstalled -WhatIf:$false "powershell-yaml" "0.4.7" | Import-Module
```
weshaggard, line 81: @benbp should we add another helper called Import-ModuleIfNotAlready (or something like that) to make the helper handle the noop aspect of importing.

benbp: Yeah I think that's a better approach. We could do some sort of convention under the hood for the script scoped variable. But I wouldn't hold up this change on it either.

scbedd: Ya'll mind if I take this `script` level variable over to `PSModule-Helpers.ps1` then? It's not like you're asking for a big change, just moving this change over there.

benbp: Yeah that sounds good to me

eng/common/scripts/Helpers/PSModule-Helpers.ps1
```diff
@@ -11,7 +10,8 @@ function Update-PSModulePathForCI()
   if ($IsWindows) {
     $hostedAgentModulePath = $env:SystemDrive + "\Modules"
     $moduleSeperator = ";"
-  } else {
+  }
+  else {
```
benbp, line 14: Just a nit, what do you like about this half indent format?

eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -68,28 +67,18 @@ Get-Content -Raw path/to/file.yml | CompatibleConvertFrom-Yaml
...
   }
 
-  # Initialize any variables or checks that need to be done once
-  $yqPresent = Get-Command 'yq' -ErrorAction SilentlyContinue
-  if (-not $yqPresent) {
-    . (Join-Path $PSScriptRoot PSModule-Helpers.ps1)
-    Install-ModuleIfNotInstalled -WhatIf:$false "powershell-yaml" "0.4.7" | Import-Module
-  }
+  . (Join-Path $PSScriptRoot PSModule-Helpers.ps1)
+  Install-ModuleIfNotInstalled -WhatIf:$false "powershell-yaml" "0.4.7" | Import-Module
```
benbp, line 79: I think the other idea here was to replace this with below, so in the case where it's a no-op we don't waste time on `import-module`. I can take that on though if you have other stuff to do, wasn't sure which piece you were volunteering to do.

```
  InstallAndImport-ModuleIfNotInstalled -WhatIf:$false "powershell-yaml" "0.4.7"
```

eng/common/scripts/Helpers/PSModule-Helpers.ps1
scbedd: This is my `Powershell` vscode extension being opinionated. I didn't even notice the code change.

eng/common/scripts/Helpers/Package-Helpers.ps1
scbedd: How can we call a function without importing it?

scbedd: Oh I see what you're saying. You're saying I should also be able to avoid `| Import-Module`.

benbp: I think it would just be something like this, but in `PSModule-Helpers.ps1`. As implemented in this PR, the no-op check will still take a couple seconds each time because we are pipelining the already installed module to `Import-Module`.

```
function InstallAndImport-ModuleIfNotInstalled([string]$module, [string]$version) {
  if ($null -eq (moduleIsInstalled $module $version)) {
    Install-ModuleIfNotInstalled -WhatIf:$false $module $version | Import-Module
  } elseif (!(Get-Module -Name $module)) {
    Import-Module $module
  }
}   
```

eng/common/scripts/Helpers/PSModule-Helpers.ps1
benbp: Huh, that's bizarre, as I think the actual powershell convention is no newline :D

benbp (approved):

eng/common/scripts/Helpers/PSModule-Helpers.ps1
```diff
@@ -102,17 +112,26 @@ function installModule([string]$moduleName, [string]$version, $repoUrl) {
...
   # commands using the same devops feed
   Unregister-PSRepository -Name $repoUrl | Out-Null
 
   return $modules[0]
 }
 
+function InstallAndImport-ModuleIfNotInstalled([string]$module, [string]$version) {
+  if ($null -eq (moduleIsInstalled $module $version)) {
+    Install-ModuleIfNotInstalled -WhatIf:$false $module $version | Import-Module
+  } elseif (!(Get-Module -Name $module)) {
```
weshaggard, line 127: What cases do you expect to hit this elseif block? I would expect if it were not installed it would be installed by the first block.

weshaggard (approved):
One question but looks good otherwise.

eng/common/scripts/Helpers/PSModule-Helpers.ps1
scbedd: `moduleIsInstalled` function checks whether it's installed or not. If not, we install it and import it.

the `elseif` is the case where the module is _installed_, but not _imported_ yet. This short circuits the import-module call in most cases.

scbedd:
[Discovered an issue with the null check](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4404336&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=1bb4b58e-f076-55c3-0dfc-a70a7c67f4bb).

On stricter modes we fail because the path isn't defined. Latest commit should be a safe fix.

PR #9512, opened 2024-12-11
Fix Windows Path Issue in js-sdk-release-tool

wanlwanl:
Bug: Falied to resolve correct path in windows
Fix: Use posix path as it is supported in all platforms


wanlwanl:
Test in https://github.com/test-repo-billy/azure-rest-api-specs/pull/3571 to ensure it works in pipeline environment

qiaozha (approved):
LGTM

MaryGao (approved):

PR #9540, opened 2024-12-14
[TestProxy] Add auto-shutdown to the proxy

### LINKED ISSUE #9467: Proxy can auto-shutdown after X minutes
@heaths is working away at integrating the test-proxy with the azure-sdk-for-rust repo. As part of this work, we're discovering new limitations for how the proxy is used. Specifically, the proxy has no way to automatically shut itself down. This is especially meaningful in `rust`, because there _is no_ attachment point for 

> after all tests are finished do X

So there is no way to shut off the auto-started test-proxy process.

This issue covers adding a new parameter to the `start` verb:

- `--auto-shutdown X`

Where X is the number of minutes since last handled request, after which the proxy will auto-shutoff.

scbedd:
Resolves #9467 

If a user provides `--auto-shutdown-in-seconds` to the proxy, the proxy will automatically shut down after the specified number of seconds if any route isn't contacted. We reset this on _any_ handled request via middleware, so we should have 100% coverage.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/CommandOptions/OptionsGenerator.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/CommandOptions/OptionsGenerator.cs
@@ -56,7 +56,7 @@ namespace Azure.Sdk.Tools.TestProxy.CommandOptions
             universalOption.AddAlias("-u");
 
             var autoShutdownOption = new Option<int>(
-                name: "--shutdownInSeconds",
+                name: "--auto-shutdown-in-seconds",
                 description: "Integer argument; When provided, the proxy will auto-shutdown after not being contacted over any HTTP route for this number of seconds.",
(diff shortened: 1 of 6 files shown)


benbp (approved):

hallipr (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/CommandOptions/OptionsGenerator.cs
```diff
@@ -55,6 +55,11 @@ public static RootCommand GenerateCommandLineOptions(Func<DefaultOptions, Task>
                 getDefaultValue: () => false);
             universalOption.AddAlias("-u");
 
+            var autoShutdownOption = new Option<int>(
+                name: "--shutdownInSeconds",
```
scbedd, line 59:
```suggestion
                name: "--auto-shutdown-in-seconds",
```

PR #9559, opened 2024-12-18
Don't invoke `prepare-pipelines` for `CI` convention for `js`, `python`

scbedd:
An alternative to implementing it this way is to make the `disallowed` repositories arguments to the tool. EG:

`--disallowed "azure/azure-sdk-for-js,azure/azure-sdk-for-python"`

And then we split this input, and use it the same way that we're using the hardcoded values. Figured I'd do the extremely easy first, and prompt opinions from @weshaggard  and @benbp based on this simple version.

### DIFF
--- a/tools/pipeline-generator/Azure.Sdk.Tools.PipelineGenerator/Program.cs
+++ b/tools/pipeline-generator/Azure.Sdk.Tools.PipelineGenerator/Program.cs
@@ -17,6 +17,7 @@ namespace PipelineGenerator
 {
     public class Program
     {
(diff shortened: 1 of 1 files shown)


weshaggard:
With my changes to the common prepare-pipelines.yml at https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/jobs/prepare-pipelines.yml#L98 I don't think we need to hard-code this in the tool itself.

scbedd:
Closing as unnecessary.

Issue #9110, opened 2024-10-08
[tsp-client] Add command to sync emitter-package.json


catalinaperalta:
Sync emitter-package.json with the package.json of an emitter. similar to this script: https://github.com/Azure/azure-sdk-for-python/blob/main/eng/common/scripts/typespec/New-EmitterPackageJson.ps1

PR #9566, opened 2024-12-21
[tsp-client] Add command to generate config files

catalinaperalta:
Add `generate-config-files` command to generate the `emitter-package.json` and `emitter-package-lock.json` files under the `eng/` directory of the repo. 

Fixes https://github.com/Azure/azure-sdk-tools/issues/9110


tools/tsp-client/src/commands.ts
```diff
@@ -405,6 +405,61 @@ export async function convertCommand(argv: any): Promise<void> {
...
+
+  // Add emitter as dependency
+  emitterPackageJson["dependencies"][packageJson["name"]] =
+    overrideJson[packageJson["name"]] ?? packageJson["version"];
+
+  delete overrideJson[packageJson["name"]];
+  const devDependencies: Record<string, any> = {};
+
+  const possiblyPinnedPackages =
+    packageJson["azure-sdk/emitter-package-json-pinning"] ?? packageJson["peerDependencies"];
```
hallipr, line 438: You'll be processing `possiblyPinnedPackages` as a string array, so you'll want to pull the keys from `peerDependencies`.

`azure-sdk/emitter-package-json-pinning` will already be a string array.

hallipr: In powershell, we use the Keys property:
`$packageJson.peerDependencies.Keys`

Does javascript enumerate the keys of an object with a `for` loop?

hallipr, line 438: `packageJson["azure-sdk/emitter-package-json-pinning"]` will be a string array here, while `packageJson["peerDependencies"]` will be a map.

Int the "json-pinning" case, the `for..in` below will enumerate the indices of the array rather than its members.

hallipr: maybe:
```js
const peerDependencies = packageJson["peerDependencies"] ?? {};

const possiblyPinnedPackages =
    packageJson["azure-sdk/emitter-package-json-pinning"] ?? Object.keys(peerDependencies);
    
for (const pinnedPackage of possiblyPinnedPackages)
```

hallipr: A test for this case would be helpful

tools/tsp-client/test/commands.spec.ts
```diff
@@ -238,4 +240,44 @@ describe.sequential("Verify commands", () => {
...
+      const args = {
+        "package-json": joinPaths(cwd(), "./test/package.json"),
+      };
+      repoRoot = await getRepoRoot(cwd());
+      await generateConfigFilesCommand(args);
+      assert.isTrue(await doesFileExist(joinPaths(repoRoot, "eng", "emitter-package.json")));
+      const emitterJson = JSON.parse(
+        await readFile(joinPaths(repoRoot, "eng", "emitter-package.json"), "utf8"),
+      );
+      assert.equal(emitterJson["dependencies"]["@azure-tools/typespec-python"], "0.37.3");
```
hallipr, line 255: I would add a devDependency assertion:

```suggestion
      assert.equal(emitterJson["dependencies"]["@azure-tools/typespec-python"], "0.37.3");
      assert.equal(emitterJson["devDependencies"]["@azure-tools/typespec-azure-core"], "~0.49.0");
```

tools/tsp-client/src/commands.ts
catalinaperalta: Done, thanks for pointing that out :)

tools/tsp-client/test/commands.spec.ts
catalinaperalta: added

tools/tsp-client/test/examples/package-sdk-pinning.json
```diff
@@ -0,0 +1,83 @@
...
+    "rimraf": "~6.0.1",
+    "typescript": "~5.6.3",
+    "typescript-eslint": "^8.8.1",
+    "yargs": "~17.7.2",
+    "chalk": "5.3.0",
+    "@types/fs-extra": "11.0.4"
+  },
+  "azure-sdk/emitter-package-json-pinning": {
+    "@typespec/compiler": ">=0.63.0 <1.0.0"
+  }
```
hallipr, line 82: pinning here would use an array of package names instead of version specs:

```suggestion
  "azure-sdk/emitter-package-json-pinning": [
    "@typespec/compiler"
  ]
```

Also, package name pinning means we don't need the peerDependency above to allow devDependency to be forwarded:
```json
{
  "name": "@azure-tools/typespec-python",
  "version": "0.37.3",
  "peerDependencies": {
    "@typespec/http": ">=0.63.0 <1.0.0",
    "@typespec/rest": ">=0.63.0 <1.0.0"
  },
  "devDependencies": {
    "@typespec/compiler": "~0.64.0",
    "c8": "^10.1.2",
    "vitest": "^2.1.2",
    "rimraf": "~6.0.1"
  },
  "azure-sdk/emitter-package-json-pinning": [
    "@typespec/compiler"
  ]
}

hallipr: This lets us pin transitive peerDependencies without having to restate them

catalinaperalta: Just pushed the update, I noticed this yesterday but decided to wrap it up today. The fixes are in the branch now.

catalinaperalta: Also, this file is a copy/paste of an actual emitter file with some modifications so I think it's fine to leave the peerDeps as a way of ensuring that we are actually taking the values of the azure-sdk/emitter-json-pinning field

hallipr (approved):

PR #9591, opened 2025-01-07
Cleaned up dependencies and set pipeline variables

raych1:
- Cleaned up dependencies
- Set variables for pipeline scenarios
- Added alias for CLI parameters
- Fixed minor log issue


chidozieononiwu (approved):
Looks fine

PR #9592, opened 2025-01-07
APIView for Rust

HarshaNalluru:
## Issue https://github.com/Azure/azure-sdk-for-rust/issues/1621

# Rust API Parser
This tool processes Rustdoc JSON files from the azure-sdk-for-rust repository and generates API views that capture the exported API surface. It organizes API elements into a hierarchical representation of modules, functions, traits, structs, and other Rust constructs while maintaining the parent-child relationships.

## Links
- **Branch(Readme)**: [azure-sdk-tools/tools/apiview/parsers/rust-api-parser at harshan/rust-api-parser · HarshaNalluru/azure-sdk-tools](https://github.com/HarshaNalluru/azure-sdk-tools/tree/harshan/rust-api-parser/tools/apiview/parsers/rust-api-parser)
- **Output**: 
  [API View for azure_core](https://spa.apiviewstagingtest.com/review/e18a70acf99a458ab00655a202ce0c25?activeApiRevisionId=66b07e2742ff47ffaf46075da3214ae4)
  

## Related work
Output from `generate_api_report` at https://github.com/Azure/azure-sdk-for-rust/pull/2019 is fed into `rust-api-parser` here as input.

## rust-api-parser
- [x] Parses Rustdoc JSON files (e.g., `azure_core.rust.json`)
- [x] Extracts and organizes the public API surface.
- [x] Preserves the hierarchical structure of the API
- [x] Generates output compatible with the APIView JSON schema
- [x] Supports both development mode (ts-node) and production mode (compiled JavaScript)

### Usage
1. Clone the repository:
    ```sh
    git clone https://github.com/Azure/azure-sdk-tools.git
    cd azure-sdk-tools/tools/apiview/parsers/rust-api-parser
    ```

2. Install Node.js dependencies:
    ```sh
    npm install
    ```

3. Build the project:
    ```sh
    npm run build
    ```

4. Run the project:
    ```sh
    node ./dist/src/main.js {input_file_path} {output_file_path}
    ```
    - Example: 
      ```sh
      ts-node src/main.ts /workspaces/azure-sdk-tools/tools/apiview/parsers/rust-api-parser/inputs/azure_core.rust.json /workspaces/azure-sdk-tools/tools/apiview/parsers/rust-api-parser/outputs/azure_core.json

      ts-node src/main.ts /workspaces/azure-sdk-tools/tools/apiview/parsers/rust-api-parser/inputs/azure_template.rust.json /workspaces/azure-sdk-tools/tools/apiview/parsers/rust-api-parser/outputs/azure_template.json
      
      node ./dist/src/main.js /workspaces/azure-sdk-tools/tools/apiview/parsers/rust-api-parser/inputs/azure_core.rust.json /workspaces/azure-sdk-tools/tools/apiview/parsers/rust-api-parser/outputs/azure_core.json

      node ./dist/src/main.js /workspaces/azure-sdk-tools/tools/apiview/parsers/rust-api-parser/inputs/azure_template.rust.json /workspaces/azure-sdk-tools/tools/apiview/parsers/rust-api-parser/outputs/azure_template.json
      ```
    - Or if you have the package installed, you can run:
      ```sh
      rust-genapi <input_file_path> <output_file_path>
      ```

## To Dos
- [x] Transform the [current output](https://github.com/HarshaNalluru/azure-sdk-tools/blob/harshan/rust-api-parser/tools/apiview/parsers/rust-api-parser/get-api-view/outputs/docs.rs) to the JSON schema expected by the API view tool
- [x] Expand the coverage to span all kinds of elements from the exported API surface
- [x] Finalize the workflow
- [ ] Hook up with the APIView Web workflow


heaths:
> It removes private items

Interesting. Last I looked only the public API surface i.e., anything reachable via public API, was included. How did you ascertain elements were truly private?

> `cargo build`

You don't need to do this separately. Just have them run `cargo run`. Like `dotnet run` et. al., it'll restore, build, and run.

tools/apiview/parsers/rust-api-parser/clean-rust-doc-output/inputs/docs/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+
+[lints]
+workspace = true
```
heaths, line 13: Workspaces don't work if you don't actually have a root `Cargo.toml` that declares a workspace; and, in this case, I don't recommend you do. It's overhead that often isn't worth it unless you're truly taking advantage of a monorepo like we are in the Azure/azure-sdk-for-rust repo.

tools/apiview/parsers/rust-api-parser/clean-rust-doc-output/src/main.rs
```diff
@@ -0,0 +1,37 @@
+use serde_json::Value;
+use std::fs::File;
+use std::io::{self, Read, Write};
+
+fn main() -> io::Result<()> {
```
heaths, line 5: This means any error must be under `std::io`. It's better to declare:
```suggestion
fn main() -> Result<(), Box<dyn std::error::Error>> {
```
This means that any type that implements `std::error::Error` can be returned, which lets you use the `?` sugar operator on any `Result<T, E>` error type. Constraining the error type this way is really only typical/useful for specific functions. `main` should generally be generic.

tools/apiview/parsers/rust-api-parser/clean-rust-doc-output/src/main.rs
```diff
@@ -0,0 +1,37 @@
...
+    // Read the JSON file generated by rustdoc
+    let mut file = File::open("inputs/docs.json")?;
+    let mut data = String::new();
+    file.read_to_string(&mut data)?;
+
+    // Parse the JSON data
+    let v: Value = serde_json::from_str(&data).expect("Unable to parse JSON");
```
heaths, line 12: Case in point: it's weird using `?` and `expect()`, which panics. If you use the recommendation above, use `?` here and everywhere else as well.

tools/apiview/parsers/rust-api-parser/clean-rust-doc-output/src/main.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    // Extract the exported API surface
+    let mut exported_api = serde_json::Map::new();
+    exported_api.insert("crate_version".to_string(), v["crate_version"].clone());
+    exported_api.insert("includes_private".to_string(), v["includes_private"].clone());
```
heaths, line 17: All this works, but since you're importing the `derive` feature, I expected to see models. In the end, it'll probably make things easier - not just to read, but maintain. And your models only need to declare what you need; everything else will just be dropped. ultimately, it's going to be a lot less code. For all this parsing you're doing, for example, would just be a couple models and you only deserialize it all in one line.

tools/apiview/parsers/rust-api-parser/get-api-view/expected-outputs/docs.rs
```diff
@@ -0,0 +1,42 @@
+
+pub mod docs{
```
heaths, line 2: At some point you'll also want to add some attributes e.g.,

```rust
#[derive(Debug)]
pub struct foo;
```

We should see most attributes (there's relatively few we shouldn't, and I don't even know if they'd show in in JSON so let's deal with them as we find them); though, remember that for `derive` macros you'll have to piece them back together. I'm not aware of any others currently.

tools/apiview/parsers/rust-api-parser/clean-rust-doc-output/inputs/docs/Cargo.toml
HarshaNalluru: I meant to move this clean-rust-doc-output project to the rust repository.
For now, while we are developing, I thought to have them side-by-side here for easier workflow.

Will move it there and figure out how to hook up the commands in the rust repo later, after we finalize the workflow and the parser/apiview tool.

heaths: How are you testing it when this won't even compile? You can use workspace references when you're not in a workspace. `cargo` will fail fast.

heaths: Andis there a reason to move it to the rust repo when other APIView support is here as well? If we do move it, we need to talk about where. I don't want to make it part of the main workspace. There's no need to compile this each and every time we build a workspace.

tools/apiview/parsers/rust-api-parser/clean-rust-doc-output/inputs/docs.json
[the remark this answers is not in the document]
HarshaNalluru: removing it

tools/apiview/parsers/rust-api-parser/clean-rust-doc-output/README.md
```diff
@@ -0,0 +1,33 @@
...
+## Prerequisites
+
+-   Rust programming language installed. You can install Rust from rust-lang.org.
+
+## Usage
+
+1. **Get the rustdoc output for the temp project**:
+
+    ```sh
+        cargo +nightly rustdoc -Z unstable-options --output-format json --package docs --no-default-features
```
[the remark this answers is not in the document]
HarshaNalluru: this is me playing around, I'll rely on `--all-features`

tools/apiview/parsers/rust-api-parser/clean-rust-doc-output/inputs/docs/Cargo.toml
HarshaNalluru: > How are you testing it when this won't even compile? You can use workspace references when you're not in a workspace. cargo will fail fast.

Well. I have a functional version of this in my local Rust repository. This copy is intended for easier reviews and to facilitate back-and-forth discussions and for my workflow.

> And is there a reason to move it to the rust repo when other APIView support is here as well? 

My goal is to refine the rustdoc output before integrating it with the API view tool. I need to further develop this aspect.

> If we do move it, we need to talk about where. I don't want to make it part of the main workspace. There's no need to compile this each and every time we build a workspace.

I completely agree. In that case, perhaps we could utilize a PowerShell script in that location?

tools/apiview/parsers/rust-api-parser/clean-rust-doc-output/src/main.rs
HarshaNalluru: thanks @heaths :)

tools/apiview/parsers/rust-api-parser/clean-rust-doc-output/src/main.rs
HarshaNalluru: this is a great suggestion, I'll follow this to get a concise api-json.

HarshaNalluru:
> Last I looked only the public API surface i.e., anything reachable via public API, was included. How did you ascertain elements were truly private?

It's a miss on my part. By filtering with the visibility, I got a much smaller copy of the API that resembled my expected output. I shouldn't have called the others' private.

tools/apiview/parsers/rust-api-parser/README.md
```diff
@@ -0,0 +1,43 @@
+# Rust API Parser
+
+This repository contains tools for cleaning and parsing  rustdoc output into API view. 
+It includes two main folders: `clean-rust-doc-output` and `get-api-view`.
+
+## clean-rust-doc-output
+
+This tool reads the JSON output generated by `rustdoc` and extracts the public API surface. It removes private items and generates a cleaned JSON file with only the public API.
```
HarshaNalluru, line 8:
```suggestion
This tool reads the JSON output generated by `rustdoc` and strips off unneeded items/metadata from the rustdoc generated JSON output w.r.t the APIView tool. Generates a scoped down version of the rustdoc output.
```

tools/apiview/parsers/rust-api-parser/README.md
```diff
@@ -0,0 +1,43 @@
...
+1. Ensure you have Rust installed. If not, you can install it from [rust-lang.org](https://www.rust-lang.org/).
+2. Navigate to the `clean-rust-doc-output` folder.
+3. Run the tool using:
+   ```sh
+   cargo run
+   ```
+
+## get-api-view
+
+This TypeScript tool processes the cleaned JSON output from `clean-rust-doc-output` and generates a structured document representing the API view. It organizes the API elements into modules, functions, and other Rust constructs.
```
HarshaNalluru, line 21:
```suggestion
This TypeScript tool processes the cleaned JSON output from `clean-rust-doc-output` and generates a JSON document representing the API view. It organizes the API elements into modules, functions, and other Rust constructs.
```

LarryOsterman:
> > It removes private items
> 
> Interesting. Last I looked only the public API surface i.e., anything reachable via public API, was included. How did you ascertain elements were truly private?

Does this imply that Rust allows you to access the transitive closure of all types reachable via public API? That is a surprising behavior - most languages hide non-public elements of public APIs (and types referenced by public APIs). Some return errors if a non-public type is returned by a public API (I thought Rust did this), which means that private types reachable via public API were not a thing.

And wouldn't it be the job of rustdoc to figure out which types are public and which types are not public (in other words, the ApiView tool wouldn't necessarily have to handle this).

tools/apiview/parsers/rust-api-parser/ci.yml
```diff
@@ -0,0 +1,114 @@
...
+
+            steps:
+              - task: NodeTool@0
+                inputs:
+                  versionSpec: '$(NodeVersion)'
+                displayName: 'Use NodeJS $(NodeVersion)'
+
+              - script: |
+                  npm install -g npm@8.16.0
+                displayName: "Install npm 8.16.0"
```
heaths, line 53: Why does the version matter? This creates a maintenance issue if we upgrade node. Already, we've deprecated v18 for Azure SDKs so you're already behind.

tools/apiview/parsers/rust-api-parser/ci.yml
```diff
@@ -0,0 +1,114 @@
...
+
+extends:
+  template: /eng/pipelines/templates/stages/1es-redirect.yml
+  parameters:
+    stages:
+      - stage: 'Build'
+        variables:
+          - template: /eng/pipelines/templates/variables/image.yml
+          - name: NodeVersion
+            value: '18.x'
```
heaths, line 32: You should be on 22.

tools/apiview/parsers/rust-api-parser/design.md
```diff
@@ -0,0 +1,193 @@
...
+   ```
+## Additional Design Considerations
+
+- **Type Bridging**: Cross-language type safety via [typeshare](https://github.com/1password/typeshare) between the Rustdoc types and the TS types.
+- **Semantic Tokens**: Context-aware syntax highlighting and navigation.
+- **Trait Implementation Analysis**: Comprehensive handling of trait implementations, specialized handling for derived trait, logical grouping of related implementations.
+- **Re-export Resolution**: Deterministic resolution of module/item re-exports to prevent duplication in rendering and maintain correct hierarchy.
+- **Sorting**: Sorts items within modules by type and name for consistent output. Sorts re-exports after the lines are generated to avoid redundancy.
+
+## Technical Architecture
```
heaths, line 72: Is this much technical detail necessary? As we update the code, it will be out of date. To keep it up to date adds a maintenance burden. I think a high-level abstract is great, but well-commented code is going to stay up to date better. This is just technical debt.

tools/apiview/parsers/rust-api-parser/rustdoc-types/LICENSE-APACHE.txt
```diff
@@ -0,0 +1,73 @@
+                                 Apache License
```
heaths, line 1: Why Apache? We use MIT. If this is just a copy of rustc's, that needs to go into a ThirdPartyNotices.txt file.

heaths (changes requested):
Not a comprehensive review as this is an internal tool and thus much code probably needs more of a functional review, but as we've been discussing this for a while and having checkpoint meetings, that may suffice.

But, the license needs to be fixed. AFAIK, MIT is the recommended license for any OSS we release by default.

NOTICE.txt
heaths, line 1: What about our license file, though? Do we need one for each project in this repo? /cc @weshaggard

tools/apiview/parsers/rust-api-parser/ci.yml
HarshaNalluru: copied this from the JS parser, I agree, we should change this.

tools/apiview/parsers/rust-api-parser/ci.yml
HarshaNalluru: again.. copied this from the JS parser, I agree, we should change this.

tools/apiview/parsers/rust-api-parser/design.md
HarshaNalluru: This is a great suggestion; my thought process was to provide a simpler guide to future devs.
But I agree with what you said, I'll make it more concise and remove this part.

NOTICE.txt
HarshaNalluru: I don't see License files under most of the projects in this repo.
Example: js-api-parser, cpp-api-parser, csharp-api-parser do not have either

tools/apiview/parsers/rust-api-parser/design.md
HarshaNalluru: removed the mermaid diagram

tools/apiview/parsers/rust-api-parser/ci.yml
```diff
@@ -0,0 +1,114 @@
...
+
+            - pwsh: |
+                $detectedPackageName=Get-ChildItem $(Pipeline.Workspace)/$(ArtifactName)/*.tgz
+                Write-Host "Detected package name: $detectedPackageName"
+                Write-Host "##vso[task.setvariable variable=PackageArtifactFullPath]$detectedPackageName"
+              displayName: "Find package artifact"
+
+            - pwsh: |
+                @"
+                registry=https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-rust/npm/registry/
```
chidozieononiwu, line 103: Should we publish to the `azure-sdk-for-js` registry since its a JS package

chidozieononiwu:
```suggestion
                registry=https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js/npm/registry/
```

NOTICE.txt
weshaggard: In general, everything in here should be using the shared LICENSE file in the root of the repo. If we are publishing to a public package manager and the package file needs to have the file locally (like our JS SDKs) then yes go ahead and copy it into the directory as well.

heaths (approved):

tools/apiview/parsers/rust-api-parser/ci.yml
xirzec: So the question is which ADO feed to publish the packages to?

I was looking at how the tools repo handles this (https://github.com/search?q=repo%3AAzure%2Fazure-sdk-tools%20https%3A%2F%2Fpkgs.dev.azure.com%2Fazure-sdk%2Fpublic%2F_packaging&type=code ) and it seems that today we either use the `azure-sdk-for-js` one or one called `azure-sdk-for-js-test-autorest` which feels honestly worse.

I don't have super strong opinions about it, and perhaps for now it will be easier to re-use the JS SDK one, but it would be nice to make something like "azure-tools-npm" in the future so we don't have the confusion about where the projects being published here actually live.

@weshaggard do you have any strong feelings about where we should be publishing this to?

HarshaNalluru: Updated to `azure-sdk-for-js` like Dozie suggested for now to not cause any disruptions.

If there are more thoughts, I'll update accordingly.

weshaggard: In the general case we have been grouping packages by language so given this is a npm package I would suggest azure-sdk-for-js dev feed as well.

tools/apiview/parsers/rust-api-parser/ci.yml
```diff
@@ -0,0 +1,114 @@
...
+    include:
+      - main
+      - feature/*
+      - release/*
+      - hotfix/*
+  paths:
+    include:
+      - tools/apiview/parsers/rust-api-parser
+
+extends:
```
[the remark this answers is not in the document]
weshaggard: See https://github.com/Azure/azure-sdk-tools/blob/3207d5379da8ee94e60b1a254d0452389b60ddc9/tools/apiview/emitters/typespec-apiview/ci.yml#L25 as an example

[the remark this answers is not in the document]
chidozieononiwu: Is that pipeline publishing to `azure-sdk-for-js` dev feed? I cannot firn `@azure-tools/typespec-apiview` in the feed.

chidozieononiwu (approved):

tools/apiview/parsers/rust-api-parser/ci.yml
HarshaNalluru: thanks Wes

PR #9598, opened 2025-01-07
Update notes on usage of logging variables

scbedd:
CC @billwert updates for clarity

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
@@ -264,8 +264,8 @@ The test-proxy is integrated with the following environment variables.
 | Variable                       | Usage                                                                                                                                                                              |
 |--------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `TEST_PROXY_FOLDER`            | if command-line argument `storage-location` is not provided when invoking the proxy, this environment variable is also checked for a valid directory to use as test-proxy context. |
-| `Logging__LogLevel__Default` | Defaults to `Information`. Possible valid values are <br/><br/>`Debug`, `Information`, `Warning`, `Error`, `Critical`.<br><br>Do not set for .NET test runs as it would cause the tests *themselves* to start emitting logs.|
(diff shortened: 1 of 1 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
```diff
@@ -264,8 +264,8 @@ The test-proxy is integrated with the following environment variables.
...
 | `TEST_PROXY_FOLDER`            | if command-line argument `storage-location` is not provided when invoking the proxy, this environment variable is also checked for a valid directory to use as test-proxy context. |
-| `Logging__LogLevel__Default` | Defaults to `Information`. Possible valid values are <br/><br/>`Debug`, `Information`, `Warning`, `Error`, `Critical`.<br><br>Do not set for .NET test runs as it would cause the tests *themselves* to start emitting logs.|
-| `Logging__LogLevel__Azure.Sdk.Tools.TestProxy`| Set to `Debug` to see request level logs emitted by the Test Proxy.|
+| `Logging__LogLevel__Default` | Sets the global log level for .NET tools including the test-proxy. Defaults to `Information`. Possible valid values are <br/><br/>`Debug`, `Information`, `Warning`, `Error`, `Critical`. <br><br> Users are recommended to use this variable, unless globally setting .NET log level negatively affects their test experience. |
+| `Logging__LogLevel__Azure.Sdk.Tools.TestProxy`| Same possible values as `Default`, but sets the log level for **only** the test-proxy. Other .NET tool log levels will remain unaffected.|
```
billwert, line 268: > Users are recommended to use this variable 

For historical sake, why? Even if I'm running Java tests there might be other tooling in the pipeline in .NET that I don't want to slow down with logging either, right?

billwert (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
scbedd: >  Even if I'm running Java tests there might be other tooling in the pipeline in .NET that I don't want to slow down with logging either, right?

Nothing that would impact the runtime of the pipeline. _Maybe_ a couple more seconds during 1ES artifact publishing or something like that. Test-Proxy can output a ton of data during Debug, including additional processing of format strings to dump sanitizer updates etc, which is why I'm much more concerned about the impacts there.

PR #9600, opened 2025-01-07
Update Test-Proxy install instructions

heaths:
Update to modern .NET recommendation, and simplify instructions with the `--prerelease` switch I asked for and was implemented a couple years ago.


scbedd (approved):
Cheers @heaths

PR #9602, opened 2025-01-08
[tsp-client] Float @azure-tools/openapi-to-typespec dep

catalinaperalta:
tsp-client doesn't need to rely on a specific version of this dependency. Will float the dep up to GA for now, so that users can specify the version of the package they wish to use. 

@pshao25 instead of pinning the openapi-to-typespec dependency in tsp-client you will pin it in the azure-rest-api-specs package.json to update versions.


tools/tsp-client/package.json
catalinaperalta, line 1: Should we still be pinning "autorest" to `^3.7.1`?

catalinaperalta:
@mikeharder I noticed that checkenforcer doesnt complete on tsp-client PRs, I believe it's because it doesnt know that the GH action has finished the checks, do you know how we can fix that? I've had to evaluate or override to merge PRs

tools/tsp-client/package.json
```diff
@@ -51,7 +51,7 @@
   },
   "dependencies": {
     "@autorest/core": "^3.10.2",
-    "@autorest/openapi-to-typespec": "0.10.5",
+    "@autorest/openapi-to-typespec": ">=0.10.5 <1.0.0",
```
pshao25, line 54: I feel like it should be peerDependency if you expect we set it in specs repo directly. And in the tsp-client, it could be devDependency to pass the test.

But I still want to understand the reason we want to float. That changes our user experience. They have to install two dependencies, which I don't think it's good.

mikeharder: We should not get in the business of having our tools install NPM packages dynamically.  Instead, I recommend creating `package.json` and `package-lock.json` files in each repo where a user might need the tool, plus a simple command to install all the necessary tools (typically `npm ci` from the repo root).  We already do this in the `azure-rest-api-specs` repo.

I agree that tsp-client should float it's dependency on openapi-to-typespec.  I think a direct floated dependency makes more sense than a peerDependency.  PeerDependencies are more intended for cases like a React plugin, where you want to require the user (instead of your package) to choose the version.

With the current change, when package `@azure-tools/typespec-client-generator-cli` is installed, by default the latest matching version of `@autorest/openapi-to-typespec` will also be installed.  Users can override this behavior, by selecting a different version of `openapi-to-typespec` in their own package.json file.

mikeharder: > If you want to float, I think we might need some command like "tsp-client update" or something to update the dependency to latest.

The dependency should be floated, so the two components can release independently.  We shouldn't need to publish a new version of `tsp-client` every time there's a new version of `openapi-to-typespec`.

mikeharder (approved):

tools/tsp-client/package.json
catalinaperalta: I agree with Mike, we dont need to add functionality in tsp-client to update the package install. Likewise, the current experience requires the user to update their tsp-client package version, with this update they can simply update the `@azure-tools/openapi-to-typespec` package directly without having to update the whole tool and without us having to release a new version of tsp-client for this dependency. So instead of users running `npm update @azure-tools/typespec-client-generator-cli@latest` they would run `npm update @azure-tools/openapi-to-typespec@latest`.

pshao25: > We shouldn't need to publish a new version of tsp-client every time there's a new version of openapi-to-typespec. Specifically, the specs repo should be able to upgrade tsp-client without upgrading openapi-to-typespec.

> So instead of users running npm update @azure-tools/typespec-client-generator-cli they would run npm update @azure-tools/openapi-to-typespec in a repo that doesnt control dependencies through a package.json

Both sounds perfect to me! Thank you!

pshao25 (approved):

Issue #9520, opened 2024-12-12
[MONITOR-SPEC-PR] User error on tspconfig - failed to set groups of options for a specific client type, client type inference should be improved 


jianingwang123:
**Description:**
```
10:02:54.858 cmdout 	[automation_generate.sh] [INFO]: Start to generate changelog and bump version in sdk/eventgrid/eventgrid
10:02:54.859 cmdout 	[automation_generate.sh] [INFO]: SDK type 'ModularClient' is detected in '/mnt/vss/_work/1/s/azure-sdk-for-js/sdk/eventgrid/eventgrid'.
10:02:54.878 cmdout 	[automation_generate.sh] [WARN]: Failed to find extactly one REST client in pattern '/mnt/vss/_work/1/s/azure-sdk-for-js/sdk/eventgrid/eventgrid/src/api/*Context.ts', got ''.
10:02:54.879 cmdout 	[automation_generate.sh] [WARN]: Failed to find extactly one REST client in pattern '/mnt/vss/_work/1/s/azure-sdk-for-js/sdk/eventgrid/eventgrid/src/rest/*Client.ts', got ''.
10:02:54.879 cmdout 	[automation_generate.sh] [ERROR]: Failed to build typespec project: "specification/eventgrid/Azure.Messaging.EventGrid":
10:02:54.879 cmdout 	[automation_generate.sh] Err: Error: Expected 1 'parameters.ts' file, but found '' in '/mnt/vss/_work/1/s/azure-sdk-for-js/sdk/eventgrid/eventgrid/src/rest'.
10:02:54.879 cmdout 	[automation_generate.sh] Stderr: "undefined"
10:02:54.879 cmdout 	[automation_generate.sh] Stdout: "undefined"
10:02:54.879 cmdout 	[automation_generate.sh] ErrorStack: "Error: Expected 1 'parameters.ts' file, but found '' in '/mnt/vss/_work/1/s/azure-sdk-for-js/sdk/eventgrid/eventgrid/src/rest'.
10:02:54.879 cmdout 	[automation_generate.sh]     at findParametersPath (/opt/hostedtoolcache/node/18.20.5/x64/lib/node_modules/@azure-tools/js-sdk-release-tools/dist/xlc/apiVersion/utils.js:117:15)
[...]

PR #9603, opened 2025-01-08
Improve SDK Type Inference  in SDK Automation

wanlwanl:
Fix: https://github.com/Azure/azure-sdk-tools/issues/9520
Problem: Service team doesn't set `isModularClientLibrary` explicitly, we can improve user experience by inferencing sdk type automatically
Solution: add the Inference flow below
```ts
if (generateByTypeSpec) {
    if (isModularLibrary !== undefined) return options.isModularLibrary ? isModular : isRLC;
    if (flavor === "azure") {
        if (namespace.includes("Management")) { // MPG
            return isModular; // by default
        } else return isRLC; // DPG
    } else { // unbranded
        return isModular; // by default
    }
} else { // generate by autorest
    // original behavior
    if (hasIsRestLevelClientOption) return isRLC;
    return isHLC;
}
```


tools/js-sdk-release-tools/src/utils/generateInputUtils.ts
```diff
@@ -4,38 +4,49 @@ import { loadTspConfig } from "../common/utils";
...
+    if (isModularLibrary === true) return SDKType.ModularClient;
+    if (isModularLibrary === false) return SDKType.RestLevelClient;
+
+    const isAzureSDK =
+        tspConfig?.options?.["@azure-tools/typespec-ts"]?.["flavor"] ===
+        "azure";
+    // NOTE: unbranded sdk will generate modular client bt default
+    if (!isAzureSDK) return SDKType.ModularClient;
+
+    const isManagementPlane = new RegExp(/\.Management[\/\\]?$/).test(
```
MaryGao, line 45: I was wondering if we still need this, we already adopted default behavior in codegen side: https://github.com/Azure/autorest.typescript/pull/2973.

wanlwanl: If so, we don't need to check it.
However, we have an independent changelog generation tool. `GetSDKType()` will be used to bump version - need to find api-version type (stable or preview) based on different client type. I'm wondering if codegen side can add 'api-version' into package.json or somewhere else, then I don't need the check of 'management' any more.

MaryGao: Could we directly refer the package name: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.Management/tspconfig.yaml#L39?

Starting with `@azure` would be modular and starting with `@azure-rest` would be RLC.

> I'm wondering if codegen side can add 'api-version' into package.json or somewhere else

I may not get your point, the `api-version` is the package version or spec version?

PR #9608, opened 2025-01-08
Add `--update` option to `tsp-client compare`

### LINKED ISSUE #9498: Make it easier to upgrade the version of `tsp-client compare` in-place
This command wraps the rest-api-diff tool [here](https://github.com/tjprescott/rest-api-diff). Because of this, the internal tool version doesn't update when you run `npm update tsp-client`. It would be good to have a way to update the internal command version.

tjprescott:
Fixes #9498. 

This PR adds a hidden `--update` flag to `tsp-client compare` to allow for easily updating the wrapped version of `@azure-tools/rest-api-diff` in-place without having to uninstall and reinstall `tsp-client`. Because the option is hidden, the command will always end with a message describing the ability to update the command. 

This tool is under active development. If you experience issues or have questions, please contact Travis Prescott direct
ly (trpresco@microsoft.com). [Tool version: 0.1.8]
**Use `tsp-client compare --update` to update @azure-tools/rest-api-diff to the latest version**

**TODO**
- [ ] bump version
- [ ] verify this works when run in a non-dev setting

### DIFF
--- a/tools/tsp-client/package.json
+++ b/tools/tsp-client/package.json
@@ -52,7 +52,7 @@
   "dependencies": {
     "@autorest/core": "^3.10.2",
     "@autorest/openapi-to-typespec": "0.10.5",
-    "@azure-tools/rest-api-diff": ">=0.1.0 <1.0.0",
+    "@azure-tools/rest-api-diff": "^0.1.8",
     "@azure-tools/typespec-autorest": ">=0.44.0 <1.0.0",
     "@azure/core-rest-pipeline": "^1.12.0",
     "@types/yargs": "^17.0.32",
--- a/tools/tsp-client/src/commands.ts
+++ b/tools/tsp-client/src/commands.ts
@@ -27,6 +27,7 @@ import { config as dotenvConfig } from "dotenv";
 import { resolve } from "node:path";
 import { doesFileExist } from "./network.js";
 import { sortOpenAPIDocument } from "@azure-tools/typespec-autorest";
+import { execSync } from "child_process";
 
 export async function initCommand(argv: any) {
   let outputDir = argv["output-dir"];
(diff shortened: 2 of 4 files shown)


tools/tsp-client/package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "@autorest/core": "^3.10.2",
     "@autorest/openapi-to-typespec": "0.10.5",
-    "@azure-tools/rest-api-diff": ">=0.1.0 <1.0.0",
+    "@azure-tools/rest-api-diff": "^0.1.8",
```
catalinaperalta, line 55: IIRC we had to explicitly specify the range so that minor version updates are allowed, since this package doesnt have a major version yet. If you want to specify a min version you could do:
```suggestion
    "@azure-tools/rest-api-diff": ">=0.1.8 <1.0.0",
```

tools/tsp-client/src/commands.ts
```diff
@@ -311,13 +312,19 @@ export async function generateCommand(argv: any) {
...
   let outputDir = argv["output-dir"];
-  const openApiDiffPath = await getPathToDependency("@azure-tools/rest-api-diff");
-  const command = [openApiDiffPath, ...args];
-  try {
-    await nodeCommand(outputDir, command);
-  } catch (err) {
-    Logger.error(`Error occurred while attempting to compare: ${err}`);
-    process.exit(1);
+  if (argv["update"]) {
+    execSync(`npm install @azure-tools/rest-api-diff@latest`, { stdio: "inherit" });
```
catalinaperalta, line 316: We were just discussing this with @mikeharder in this other PR: #9602. I dont think we need to have this functionality in tsp-client. It seems if a user is at the point of knowing that they need to update the package version then they should simply run the npm update command directly.

catalinaperalta: Can this be as simple as updating your output message to be the following:

_This tool is under active development. If you experience issues or have questions, please contact Travis Prescott direct
ly ([trpresco@microsoft.com](mailto:trpresco@microsoft.com)). [Tool version: 0.1.8]
**Run `npm update @azure-tools/rest-api-diff` to update @azure-tools/rest-api-diff to the latest version**_

catalinaperalta: I believe `npm update @azure-tools/rest-api-diff@latest` should work as well to be very explicit but if it doesnt then npm install should be fine.

tools/tsp-client/package.json
catalinaperalta: Once the tool has a major version we should be able to simplify this

tjprescott: Yeah, I didn't actually manually change this file... Hmm.... I wonder if this command did.... Need to check that.

tools/tsp-client/src/commands.ts
tjprescott: Well, the problem is from what folks have told me, running npm update doesn't actually work. It still uses the original internal version that tsp-client originally pulled down.

catalinaperalta: I have had issues with npm update myself...maybe just recommending npm install, the same as the command you're wrapping under this command would be better?

PR #9615, opened 2025-01-10
[Python APIView] install optional dependencies on ImportError

swathipil:
addresses: #9612

Fixes the issue, but may be a temporary solution depending on further discussion.

Local testing passed. Need to test against existing APIViews in the CI and azure-quantum.

### DIFF
--- a/packages/python-packages/apistubgentest/setup.py
+++ b/packages/python-packages/apistubgentest/setup.py
@@ -63,5 +63,12 @@ setup(
     ],
     zip_safe=False,
     packages=find_packages(),
+    extras_require={
+        ":python_version<'3.7'": ["qsharp>=1.0.33,<2.0"],
+        "cirq": [
+            "cirq-core>=1.3.0,<=1.4.1",
+            "cirq-ionq>=1.3.0,<=1.4.1",
+        ]
+    },
     python_requires=">=3.8"
 )
--- a/packages/python-packages/apistubgentest/setup.py
+++ b/packages/python-packages/apistubgentest/setup.py
@@ -64,10 +64,10 @@ setup(
     zip_safe=False,
     packages=find_packages(),
     extras_require={
-        ":python_version<'3.7'": ["qsharp>=1.0.33,<2.0"],
-        "cirq": [
-            "cirq-core>=1.3.0,<=1.4.1",
-            "cirq-ionq>=1.3.0,<=1.4.1",
+        ":python_version<'3.7'": ["qsharp<0.16.0"],
+        "testing": [
+            "httpx>=0.15.4,<1.0",
+            "pandas>1.0.0,<3.0",
         ]
     },
     python_requires=">=3.8"
--- a/packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
+++ b/packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
@@ -283,17 +283,21 @@ class StubGenerator:
         zip_file.extractall(temp_pkg_dir)
         logging.debug("Extracted package files into temp path")
 
-        # Locate the .dist-info directory
-        temp_dir = Path(temp_pkg_dir)
-        dist_info_dirs = list(temp_dir.glob("*.dist-info"))
-        if not dist_info_dirs:
-            raise ValueError("No .dist-info directory found in the wheel.")
+        try:
+            # Locate the .dist-info directory
+            temp_dir = Path(temp_pkg_dir)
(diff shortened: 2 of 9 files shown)


packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
```diff
@@ -121,16 +122,52 @@ def _parse_arg(self, name):
...
+        extras_require = dist.metadata.get_all("Provides-Extra")
+
+        # Clean up the temporary directory
+        shutil.rmtree(temp_dir)
+
+        return extras_require or []
+
+    def install_extra_dependencies(self, extras_require):
+        for extra in extras_require:
+            if ':' in extra:
```
scbedd, line 145: So this will only cause an issue if the extra that is `platform` and `pyversion` specific will show up on the platform and pyversion that we utilize for generating api stubs right?

To me it seems like a great start point, especially if we don't have any situations where the conditional extras are required. I think you should talk to @tjprescott to see if its worth the extra effort.

You can use `packaging` if you need to check compatibility.

```
from packaging.requirements import Requirement
from packaging.markers import default_environment

# Example extras definition
extras_requirements = {
    "extra1": [
        "package1; python_version<'3.10'",
        "package2; sys_platform=='win32'",
    ],
    "extra2": [
        "package3; python_version>='3.7'",
    ],
}

# Function to check if an extra is compatible with the running environment
def is_extra_compatible(extra_name, extras_requirements):
    if extra_name not in extras_requirements:
        return False, "Extra not found"
    
    for req_str in extras_requirements[extra_name]:
        req = Requirement(req_str)
        if req.marker:  # Check if there's a marker
            if not req.marker.evaluate(default_environment()):
                return False, f"Incompatible: {req_str}"
    
    return True, "Compatible"

# Check compatibility of extras
extra_name = "extra1"
compatible, message = is_extra_compatible(extra_name, extras_requirements)
print(f"Extra '{extra_name}' compatibility: {message}")

extra_name = "extra2"
compatible, message = is_extra_compatible(extra_name, extras_requirements)
print(f"Extra '{extra_name}' compatibility: {message}")
```

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
```diff
@@ -121,16 +122,52 @@ def _parse_arg(self, name):
...
+                logging.info(f"Skipping conditional extra dependency: {extra}")
+                continue
+            logging.info(f"Installing extra dependency: {extra}")
+            try:
+                check_call([sys.executable, "-m", "pip", "install", f"{self.pkg_path}[{extra}]", "-q"])
+            except:
```
scbedd, line 151: Thoughts on this @tjprescott ? 

I thought this would be a great way to bypass the "DO we need extras though?", as it will be a fairly uncommon occurrence.

scbedd (approved):
I think it definitely seems reasonable based on our convo. Any thoughts from @tjprescott

packages/python-packages/apistubgentest/setup.py
```diff
@@ -63,5 +63,12 @@
     ],
     zip_safe=False,
     packages=find_packages(),
+    extras_require={
+        ":python_version<'3.7'": ["qsharp>=1.0.33,<2.0"],
+        "cirq": [
+            "cirq-core>=1.3.0,<=1.4.1",
+            "cirq-ionq>=1.3.0,<=1.4.1",
+        ]
+    },
```
tjprescott, line 72: Does this mean every Azure package that has an optional dependency will need it added here?

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
tjprescott: After chatting with @swathipil this makes sense to me.

packages/python-packages/apistubgentest/setup.py
tjprescott: This is for testing.

tjprescott (approved):
I'll defer to @praveenkuttapan and @scbedd on the best way to test this on CI to make sure it behaves, but the logic makes sense to me.

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
swathipil: As it is currently, azure-quantum is the only one to run into this issue and they don't have any conditional extras. I think this seems fine to be for now. Thoughts @tjprescott?

praveenkuttappan (approved):

PR #9658, opened 2025-01-17
handle only deleted files in a <language> - pullrequest build

scbedd:
[In action](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4488546&view=results)

I think the only real discussion item here is whether to fold the deleted files directly under `ChangedFiles`. I prefer it to be separate, as we can rely on `ChangedFiles` actually resolving to a reference on disk. However for this usage it doesn't actually matter, as we don't rely on them existing on disk to check which package they belong to in `Get-PrPackageProperties`.

Pending feedback from @weshaggard or @benbp , I have an opinion but I know ya'll will have one as well 👍 

@kristapratico when I merge this (regardless of implementation detail) this PR will resolve the issue you saw over [here.](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4484095&view=logs&jobId=22aa3ba4-644c-509d-acec-db6f6514dcba)

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -162,6 +162,14 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     $allPackageProperties = Get-AllPkgProperties
     $diff = Get-Content $InputDiffJson | ConvertFrom-Json
     $targetedFiles = $diff.ChangedFiles
+
+    if ($diff.DeletedFiles) {
+        if (-not $targetedFiles) {
+            $targetedFiles = @()
+        }
+        $targetedFiles += $diff.DeletedFiles
+    }
+
     # The exclude paths and the targeted files paths aren't full OS paths, they're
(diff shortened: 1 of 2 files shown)


benbp (approved):

eng/common/scripts/Package-Properties.ps1
```diff
@@ -162,6 +162,14 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     $allPackageProperties = Get-AllPkgProperties
     $diff = Get-Content $InputDiffJson | ConvertFrom-Json
     $targetedFiles = $diff.ChangedFiles
+
+    if ($diff.DeletedFiles) {
```
weshaggard, line 166: Will this work? If we have deleted files in the list of changed files won't we try to build projects that don't exist?

scbedd: I am relying on `save-package-properties` NOT generating a project for that deleted folder. If the package itself is deleted, then none of those files will map back to a package.

image

weshaggard (approved):

PR #9668, opened 2025-01-21
Support incrementing semver prereleases with 'zero' versions

hallipr:
- Remove the artificial `PrereleaseLabel` and `PrereleaseNumber` values from non-prerelease versions
- Add support for bumping Major, Minor or Patch versions using `IncrementAndSetToPrerelease($segment)`
  - The default segment is still `minor`
- Fix support for "zero" versions in `HasValidPrereleaseLabel`
  - previously, `[AzureEngSemanticVersion]::new("0.1.2").HasValidPrereleaseLabel()` would return
     ```
     Unexpected pre-release identifier 'zzz',  should be 'beta' or 'alpha'
     False
     ```

Parse results before:
| Property         | `0.1.2`  | `1.2.3`  | `1.2.3-alpha.1` | `1.2.3-alpha.20250121.10` |
| ---------------- | -------- | -------- | --------------- | ------------------------- |
| Major            | 0        | 1        | 1               | 1                         |
| Minor            | 1        | 2        | 2               | 2                         |
| Patch            | 2        | 3        | 3               | 3                         |
| PrereleaseLabel  | zzz      | zzz      | alpha           | alpha                     |
| BuildNumber      |          |          |                 | 10                        |
| PrereleaseNumber | 99999999 | 99999999 | 1               | 20250121                  |
| IsPrerelease     | True     | False    | True            | True                      |
| VersionType      | Beta     | Patch    | Beta            | Beta                      |

Parse results after change:
| Property         | `0.1.2` | `1.2.3` | `1.2.3-alpha.1` | `1.2.3-alpha.20250121.10` |
| ---------------- | ------- | ------- | --------------- | ------------------------- |
| Major            | 0       | 1       | 1               | 1                         |
| Minor            | 1       | 2       | 2               | 2                         |
| Patch            | 2       | 3       | 3               | 3                         |
| PrereleaseLabel  |         |         | alpha           | alpha                     |
| BuildNumber      |         |         |                 | 10                        |
| PrereleaseNumber | 0       | 0       | 1               | 20250121                  |
| IsPrerelease     | True    | False   | True            | True                      |
| VersionType      | Beta    | Patch   | Beta            | Beta                      |

### DIFF
--- a/eng/common/scripts/SemVer.ps1
+++ b/eng/common/scripts/SemVer.ps1
@@ -78,13 +78,10 @@ class AzureEngSemanticVersion : IComparable {
 
       if ($null -eq $matches['prelabel'])
       {
-        # artifically provide these values for non-prereleases to enable easy sorting of them later than prereleases.
-        $this.PrereleaseLabel = "zzz"
-        $this.PrereleaseNumber = 99999999
         $this.IsPrerelease = $false
         $this.VersionType = "GA"
         if ($this.Major -eq 0) {
-           # Treat initial 0 versions as a prerelease beta's
+          # Treat initial 0 versions as a prerelease beta's
           $this.VersionType = "Beta"
           $this.IsPrerelease = $true
         }
@@ -116,7 +113,7 @@ class AzureEngSemanticVersion : IComparable {
   # See https://azure.github.io/azure-sdk/policies_releases.html#package-versioning
   [bool] HasValidPrereleaseLabel()
   {
-    if ($this.IsPrerelease -eq $true) {
+    if ($this.PrereleaseLabel) {
       if ($this.PrereleaseLabel -ne $this.DefaultPrereleaseLabel -and $this.PrereleaseLabel -ne $this.DefaultAlphaReleaseLabel) {
         Write-Host "Unexpected pre-release identifier '$($this.PrereleaseLabel)', "`
                    "should be '$($this.DefaultPrereleaseLabel)' or '$($this.DefaultAlphaReleaseLabel)'"
@@ -136,7 +133,7 @@ class AzureEngSemanticVersion : IComparable {
   {
     $versionString = "{0}.{1}.{2}" -F $this.Major, $this.Minor, $this.Patch
 
-    if ($this.IsPrerelease -and $this.PrereleaseLabel -ne "zzz")
+    if ($this.PrereleaseLabel)
     {
       $versionString += $this.PrereleaseLabelSeparator + $this.PrereleaseLabel + `
                         $this.PrereleaseNumberSeparator + $this.PrereleaseNumber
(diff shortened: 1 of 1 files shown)


eng/common/scripts/SemVer.ps1
```diff
@@ -147,24 +144,42 @@ class AzureEngSemanticVersion : IComparable {
...
       }
-      $this.PrereleaseNumber++
+      if($Segment -eq "Minor") {
+        $this.Patch = 0
+      }
+
+      # If the major version is 0, we don't need a prerelease label
+      if ($this.Major -ne 0)
+      {
+        $this.PrereleaseLabel = $this.DefaultPrereleaseLabel
```
weshaggard, line 171: I wonder if this should be also an option as folks like JS that increment the patch don't want to turn it into a prerelease as part of the increment.

eng/common/scripts/SemVer.ps1
```diff
@@ -360,6 +394,38 @@ class AzureEngSemanticVersion : IComparable {
...
+    }
+
+    $version.IncrementAndSetToPrerelease("minor")
+    $expected = "0.3.0"
+    if ($expected -ne $version.ToString()) {
+      Write-Host "Error: version string did not correctly increment. Expected: $expected, Actual: $version"
+    }
+
+    $version.IncrementAndSetToPrerelease("major")
+    $expected = "1.0.0-beta.1"
```
weshaggard, line 418: Why would this be the expected? You might want to consider resetting or at least asserting what $version should be here if you plan to reuse the same object that is being mutated.

eng/common/scripts/SemVer.ps1
hallipr: Maybe a separate function then:

```powershell
[void] IncrementAndSetToPrerelease($Segment) {}
[void] Increment($Segment) {}
```

eng/common/scripts/SemVer.ps1
hallipr: I've updated the tests to be more explicit on the version before the increment call.

For that major bump, we do expect a "0.1.2" to become "1.0.0", but then append a prerelease tag.
So `"0.1.2" IncrementAndSetToPrerelease("major")` -> `1.0.0-beta.1`
But `"1.0.0-beta.1" IncrementAndSetToPrerelease("major")` -> `1.0.0-beta.2`

eng/common/scripts/SemVer.ps1
```diff
@@ -78,13 +78,10 @@ class AzureEngSemanticVersion : IComparable {
 
       if ($null -eq $matches['prelabel'])
       {
-        # artifically provide these values for non-prereleases to enable easy sorting of them later than prereleases.
-        $this.PrereleaseLabel = "zzz"
```
weshaggard, line 82: This will break https://github.com/Azure/azure-sdk-for-go/blob/main/eng/scripts/Update-ModuleVersion.ps1#L43 so we will need to submit a fix.

hallipr: agreed.  

There are several copycat implementations in both typescript and powershell that are only used for version comparison or metadata and don't do increment.   It would be nice to have them no longer return artificial prerelease info as part of their parse.

hallipr:
azd is manually incrementing versions and could move to the new overload:
https://github.com/Azure/azure-dev/blob/88ad9bc2cf8e63ef6e0e6b4577770127a162fc3d/eng/scripts/Update-CliVersion.ps1#L24
```powershell
    if ($version.IsPrerelease) {
        if ($version.HasValidPrereleaseLabel()) {
            # 0.1.0-beta.1 -> 0.1.0-beta.2
            # 1.0.0-beta.1 -> 1.0.0-beta.2
            $version.PrereleaseNumber++
        } else {
            # 0.1.0 -> 0.2.0-beta.1
            $version.Minor++
            $version.PrereleaseLabel = 'beta'
            $version.PrereleaseNumber = 1
        }
    } else {
        # 1.0.0 -> 1.1.0-beta.1
        $version.IncrementAndSetToPrerelease()
    }
```

eng/common/scripts/SemVer.ps1
hallipr: The method name `IncrementPrerelease` would be more appropriate.   
  Prerelease tagged input only increments the prerelease number
  Release input increments the segment, then appends a prerelease tag if  `major != 0`
  
The resulting version is always prerelease.

weshaggard (approved):
Let's be sure to look at the existing usages and fix any that are broken by this change or that can now use your changes.

eng/common/scripts/SemVer.ps1
hallipr: Fixed in -go repo

PR #9670, opened 2025-01-21
Support spec PR scenario in sdk gen pipeline template

raych1:
- use 'azure-sdk' forked repo for spec PR scenario
- use spec PR info for checkout and spec-gen-sdk

### DIFF
--- a/eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
+++ b/eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
@@ -5,7 +5,7 @@ parameters:
     type: string
   - name: SpecRepoCommit
     type: string
-    default: 'HEAD'
+    default: '$(Build.SourceBranch)'
   - name: SdkRepoCommit
     type: string
     default: 'HEAD'
@@ -47,6 +47,15 @@ jobs:
     - checkout: none
 
     - pwsh: |
+        $tspConfigPathPattern = '^specification\/([^\/]+\/)+tspconfig\.yaml$'
+        $readmePathPattern = '^specification\/([^\/]+\/){2,}readme\.md$'
+        if (('${{ parameters.ConfigType }}' -eq 'TypeSpec') -and ('${{ parameters.ConfigPath }}' -notmatch $tspConfigPathPattern)) {
+          Write-Host "##vso[task.logissue type=error]'ConfigPath' must be a valid 'tspconfig.yaml' file path when 'ConfigType' is set to 'TypeSpec'. For example, 'specification/contosowidgetmanager/Contoso.Management/tspconfig.yaml'"
+          Exit 1
+        } elseif (('${{ parameters.ConfigType }}' -eq 'OpenAPI') -and ('${{ parameters.ConfigPath }}' -notmatch $readmePathPattern)) {
+          Write-Host "##vso[task.logissue type=error]'ConfigPath' must be a valid 'readme.md' file path when 'ConfigType' is set to 'OpenAPI'. For example, 'specification/appplatform/resource-manager/readme.md'"
+          Exit 1
+        }
         $urlPattern = '^https://github\.com/(?<organization>[^/]+)/(?<repository>[^/]+)'
         if ('${{ parameters.SpecRepoUrl }}' -match $urlPattern) {
           $specRepoOwner = $Matches['organization']
@@ -64,11 +73,11 @@ jobs:
 
         if ('${{ parameters.SdkRepoUrl }}' -match $urlPattern) {
           if ('${{ parameters.SpecRepoUrl }}'.EndsWith('-pr') -and (-not '${{ parameters.SdkRepoUrl }}'.EndsWith('-pr'))) {
-            Write-Host "SdkRepoUrl must be a private repository if SpecRepoUrl is a private repository."
+            Write-Host "##vso[task.logissue type=error]SdkRepoUrl must be a private repository if SpecRepoUrl is a private repository."
             Exit 1
           }
 
-          $sdkRepoOwner = if ('$(SdkRepoOwner)' -eq '') { $Matches['organization'] } else { '$(SdkRepoOwner)' }
+          $sdkRepoOwner = $Matches['organization']
           Write-Host "##vso[task.setvariable variable=SdkRepoOwner]$sdkRepoOwner"
           Write-Host "SdkRepoOwner variable set to: $sdkRepoOwner"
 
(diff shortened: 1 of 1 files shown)


eng/common/pipelines/templates/steps/generate-sdk.yml
```diff
@@ -0,0 +1,172 @@
+parameters:
+  - name: SpecRepoOwner
+    type: string
+  - name: SpecRepoName
+    type: string
+  - name: SdkRepoOwner
+    type: string
+  - name: SdkRepoName
+    type: string
+  - name: IntegrationSdkRepoOwner
```
maririos, line 10: what is this for?

eng/common/pipelines/templates/steps/generate-sdk.yml
```diff
@@ -0,0 +1,172 @@
...
+    type: string
+    default: ''
+  - name: IntegrationSdkRepoName
+    type: string
+    default: ''
+  - name: SpecRepoCommit
```
maririos, line 16: nit: could we sort these so all spec properties are together and then the SDKs ones?

maririos (approved):

eng/common/pipelines/templates/steps/generate-sdk.yml
```diff
@@ -0,0 +1,172 @@
...
+- template: /eng/common/pipelines/templates/steps/sparse-checkout.yml
+  parameters:
+    Paths:
+    - '/*'
+    - '!sdk/**/test-recordings/*'
+    - '!sdk/**/recordings/*'
+    - '!sdk/**/SessionRecords/*'
+    - '!sdk/**/session-records/*'
+    Repositories:
+    - Name: ${{ parameters.SdkRepoOwner }}/${{ parameters.SdkRepoName }}
```
weshaggard, line 77: What is the reason to move the spec checkout to the containing yml?

eng/common/pipelines/templates/steps/generate-sdk.yml
weshaggard: I'm curious about the same. If this is about enabling forks, why can't we use the existing paramters for that?

eng/common/pipelines/templates/steps/generate-sdk.yml
raych1: Updated code per the discussion.

eng/common/pipelines/templates/steps/generate-sdk.yml
raych1: Created a new version of the pipeline based on the discussion with Wes.

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
```diff
@@ -86,12 +86,14 @@ jobs:
           Exit 1
         }
 
-        $sdkRepoNameForSpecGen = $sdkRepoName
-        if ($sdkRepoName.EndsWith('-for-net')) {
-          $sdkRepoNameForSpecGen = $sdkRepoName + "-track2"
+        if ($env:BUILD_REASON -eq 'PullRequest') {
+           $specRepoCommitVar = $env:BUILD_SOURCEBRANCH
```
weshaggard, line 90: What does this variable end in "Var"?

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
```diff
@@ -86,12 +86,14 @@ jobs:
           Exit 1
         }
 
-        $sdkRepoNameForSpecGen = $sdkRepoName
-        if ($sdkRepoName.EndsWith('-for-net')) {
-          $sdkRepoNameForSpecGen = $sdkRepoName + "-track2"
+        if ($env:BUILD_REASON -eq 'PullRequest') {
```
weshaggard, line 89: Can we default the SpecRepoCommit parameter to this variable, so we don't have to override here in code?

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
raych1: I didn't want to have same name as `parameter.SpecRepoCommit` though it's allowed.

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
raych1: Made this change and changed the value in entry yml too.
This is a [testing commit](https://github.com/azure-sdk/azure-sdk-for-java/commit/4a2db7c74eaa8b67f7fed79a39a9726ec3ea7261#diff-e0126afe55b455125a857b72ca29be278c8628ae725833622cbc5b8cc1f6e708) generated.

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
```diff
@@ -5,7 +5,7 @@ parameters:
     type: string
   - name: SpecRepoCommit
     type: string
-    default: 'HEAD'
+    default: '$(Build.SourceBranch)'
```
weshaggard, line 8: While I did suggest this, we should make sure we test this outside of the context of a PR to ensure it doesn't cause issues.

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
```diff
@@ -148,6 +151,14 @@ jobs:
           sdk_gen_info="$sdk_gen_info '${{ parameters.ConfigPath }}',"
         fi
 
+        if [ "$(Build.Reason)" = "PullRequest" ]; then
```
weshaggard, line 154: Any reason not to do this above? BTW if you set the variable in the same step you cannot use it in that same script block it can only be used in a future step.

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
```diff
@@ -148,6 +151,14 @@ jobs:
...
+          optional_params="$optional_params --n=$(System.PullRequest.PullRequestNumber)"
+          specPrUrl="${{ parameters.SpecRepoUrl }}/pull/$(System.PullRequest.PullRequestNumber)"
+          sdk_gen_info="$sdk_gen_info spec PR: $specPrUrl"
+          echo "Setting SDK repository to the forked repository under azure-sdk organization"
+          echo "##vso[task.setvariable variable=SdkRepoOwner]azure-sdk"
```
weshaggard, line 159: Why are we only doing this for pull requests? Shouldn't we always push to azure-sdk fork?

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
raych1: This change has been verified. This is a [test run](https://dev.azure.com/azure-sdk/playground/_build/results?buildId=4503033&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e)

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
raych1: It was already implemented by `git-push-changes.yml`, see this [SDK PR](https://github.com/Azure/azure-sdk-for-java/pull/43910) being created.

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
raych1: It needs to be after the sparse checkout step. So either the step of `Set SpecRepoCommit variable` or this step. I chose this step since it will only be used in the `code push` step.

weshaggard: I don't think you set and consume the variable in the same step so you will at least need to split.

eng/common/pipelines/templates/jobs/archetype-spec-gen-sdk.yml
weshaggard: That is very confusing as it seems like you are intending to use it in this step as well. However you already disabled the PR if the build reason is PullRequest so how would this ever be used?

weshaggard (approved):

PR #9680, opened 2025-01-22
Update dotnet parser to skip RequiresUnreferencedCode attribute

praveenkuttappan:
Update the list of attributes that are skipped from APIView for .NET


jsquire (approved):
Thanks, Praveen!

maririos (approved):

PR #9681, opened 2025-01-23
Sparse checkout 1espt auto-baselining file by default

benbp:
1es pt scripts normally try to download this file remotely using the pipeline gh token, but they will also honor the file on the machine. In some cases the remote fetch will fail or be unauthorized, so we should always check out 1espt config files for consistency/reliability.

### DIFF
--- a/eng/common/pipelines/templates/steps/sparse-checkout.yml
+++ b/eng/common/pipelines/templates/steps/sparse-checkout.yml
@@ -86,7 +86,10 @@ steps:
 
               # Set non-cone mode otherwise path filters will not work in git >= 2.37.0
               # See https://github.blog/2022-06-27-highlights-from-git-2-37/#tidbits
-              git sparse-checkout set --no-cone '/*' '!/*/' '/eng'
+              # '/*' '!/*/' -> only checkout files in top level directory
+              # '/eng' -> checkout required eng/ scripts/configs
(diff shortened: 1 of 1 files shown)


eng/common/pipelines/templates/steps/sparse-checkout.yml
```diff
@@ -86,7 +86,11 @@ steps:
 
               # Set non-cone mode otherwise path filters will not work in git >= 2.37.0
               # See https://github.blog/2022-06-27-highlights-from-git-2-37/#tidbits
-              git sparse-checkout set --no-cone '/*' '!/*/' '/eng'
+              # '/*' '!/*/' -> only checkout files in top level directory
+              # '/eng' -> checkout required eng/ scripts/configs
+              # '.config/1espt' -> required for files like .config/1espt/PipelineAutobaseliningConfig.yml used by 1es PT scripts
+              # '.config/guardian' -> required for files like .config/guardian/.gdnbaselines used by 1es PT scripts
+              git sparse-checkout set --no-cone '/*' '!/*/' '/eng' '/.config/1espt' '/.config/guardian'
```
weshaggard, line 93: Should we just always include /.config directory?

weshaggard (approved):

eng/common/pipelines/templates/steps/sparse-checkout.yml
benbp: Yeah, probably. I'll switch it

PR #9692, opened 2025-01-24
[Python APIView] add support for pyproject.toml only packages

swathipil:
fixes: #9617

Adding support for packages with only pyproject.toml.

TODO:
- [x] Optional dependencies for pyproject.toml only whl/sdist
- [ ] Add tests


packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
```diff
@@ -136,22 +136,34 @@ def install_extra_dependencies(self):
...
         else:
-            # package root is passed as arg to parse
-            pkg_root_path = self.pkg_path
-            pkg_name, version, namespace, self.extras_require = parse_setup_py(pkg_root_path)
+            namespace = self._get_package_root_namespace(pkg_root_path)
+        return pkg_root_path, pkg_name, version, namespace
 
-        logging.debug("package name: {0}, version:{1}, namespace:{2}".format(pkg_name, version, namespace))
+    def _get_package_root_namespace(self, pkg_root_path):
+        for root, dirs, files in os.walk(pkg_root_path):
```
swathipil, line 153: [Replacing - with .](https://github.com/Azure/azure-sdk-for-python/blob/eec0113e48c26a1f416203ba6f82c233895ec8ef/tools/azure-sdk-tools/ci_tools/parsing/parse_functions.py#L393) doesn't work for packages like `azure-eventhub-checkpointstoreblob` where the namespace includes an extra path : `azure.eventhub.extensions.checkpointstoreblob.

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
```diff
@@ -136,22 +136,34 @@ def install_extra_dependencies(self):
...
-        if self.pkg_path.endswith(".whl") or self.pkg_path.endswith(".zip"):
-            pkg_root_path = self.wheel_path
-            pkg_name, version = self._parse_pkg_name()
-            namespace = self.get_module_root_name(pkg_root_path)
+    def _get_pkg_metadata(self):
+        metadata = get_metadata(self.pkg_path)
+        pkg_name = metadata.name
+        version = metadata.version
+        pkg_root_path = self.wheel_path or self.pkg_path
+        # TODO: pyproject.toml + whl/sdist: metadata.extras_require doesn't pick up optional dependencies
```
swathipil, line 144: Specifying optional dependencies in the pyproject.toml as per [[this](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements)] doesn't work with whl/sdist. The extras are not included in the Provides-Extras section of the PKG-INFO/dist-info.

It _does_ work when the pkg path is the package itself.

packages/python-packages/apiview-stub-generator/setup.py
```diff
@@ -34,7 +34,8 @@
     install_requires=[
         "charset-normalizer",
         "pylint",
-        "azure-pylint-guidelines-checker"
+        "azure-pylint-guidelines-checker",
+        "pkginfo"
```
swathipil, line 38: pkginfo is currently used by ci_tools: https://github.com/Azure/azure-sdk-for-python/blob/21ff3945ce37cf36758563637c6e8a15101b1989/eng/ci_tools.txt#L10

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
swathipil: Currently, the extras require won't be an issue for anyone, since azure-quantum is the only package that needs extras and they're using setup.py.

Need to figure out how to get the optional deps picked up from the pyproject.toml.

packages/python-packages/apiview-stub-generator/setup.py
swathipil: @scbedd - Do you see any issues with this?

Alternatives I've tried:
- importlib.metadata --> Requires that the package is installed first to grab metadata. In the case of something like azure-quantum that fails to install without the optional deps, this won't work.
- build --> Tried a few things but didn't seem as clean as pkginfo. I can go back and test this out more if pkginfo doesn't work.

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
swathipil: I believe this is causing the following failure, so we can't do this: `Skipping module build.lib.apistubgentest. Module should start with mnt.vss._work.1.s.packages.python-packages.apiview-stub-generator....apistubgentest`

I don't see an alternative to parsing pyproject.toml/setup.py. Thoughts?

swathipil: discussion w Laurent: Can we avoid passing in namespace here? Can't we just pick up namespace as we go during parsing?

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
scbedd: > It does work when the pkg path is the package itself.

Bit more clarity here? Are you saying that if you point this metadata parse at the distribution file specifically, it parses it properly?

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
scbedd: > [Replacing - with .](https://github.com/Azure/azure-sdk-for-python/blob/eec0113e48c26a1f416203ba6f82c233895ec8ef/tools/azure-sdk-tools/ci_tools/parsing/parse_functions.py#L393) doesn't work for packages like azure-eventhub-checkpointstoreblob where the namespace doesn't follow the package name

Yeah. This is about how I would do it @swathipil . Unless the pkginfo metadata contains the top-level namespace?

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
scbedd: > Need to figure out how to get the optional deps picked up from the pyproject.toml.

I spent some time this morning digging a bit. I think this may be a part of how this gets encoded that we're not aware of. At least as far as `pkginfo` goes.

I say this because I did some local experimentation.

```python
>>> from pkginfo import Wheel
>>> mypackage = Wheel('C:/repo/azure-sdk-for-python/.artifacts/azure-core/azure_core-1.32.1-py3-none-any.whl')
>>> mypackage.requires_dist
['requests >=2.21.0', 'six >=1.11.0', 'typing-extensions >=4.6.0', "aiohttp >=3.0 ; extra == 'aio'"]
>>> mypackage.provides_extras
['aio']
```

You _can_ get the extras, you just need to do some filtering on the `requires_dist` from `pkginfo`.

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
scbedd: I do think this might traverse a bit too deep? If you absolutely need the namespace I think the first folder deeper than the LAST [pure extension dunder init](https://github.com/Azure/azure-sdk-for-python/blob/770bb3c003c348898e45869298d3e3bd48e21ff6/sdk/core/azure-core/azure/__init__.py#L1) is the one you want.

swathipil: V helpful, thanks @scbedd !

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
swathipil: Ahh, I see! I tested again this morning and it worked, so this must've been the issue. Thank you @scbedd !

packages/python-packages/apiview-stub-generator/apistub/_stub_generator.py
scbedd: I am going to add this exact concept to `parse_pyproject.toml` for discovering the root namespace of a package in azure-sdk-tools. Soon ™️

PR #9698, opened 2025-01-24
Fix pipeline generation for private repos

weshaggard:
Pipeline generation for the private repos no longer works after my recent refactoring. The reason is because sparse-checkout doesn't support cloning them without a key. I've hacked together a proof of concept to unblock folks but I'm not sure how I like this implementation. 

I'm curious of your thoughts @benbp

### DIFF
--- a/eng/common/pipelines/templates/steps/sparse-checkout.yml
+++ b/eng/common/pipelines/templates/steps/sparse-checkout.yml
@@ -11,10 +11,19 @@ parameters:
   - name: SkipCheckoutNone
     type: boolean
     default: false
+  - name: TokenToUseForAuth
+    type: string
+    default: ''
 
 steps:
   - ${{ if not(parameters.SkipCheckoutNone) }}:
       - checkout: none
+
+  - ${{ if ne(parameters.TokenToUseForAuth, '') }}:
+    - pwsh: |
+        $base64Token = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes("nobody:${{ parameters.TokenToUseForAuth }}"))
+        git config set --global "http.extraheader" "AUTHORIZATION: basic $base64Token"
+      displayName: Setup git config auth header
 
   - task: PowerShell@2
     ${{ if eq(length(parameters.Repositories), 1) }}:
@@ -126,3 +135,9 @@ steps:
         }
       pwsh: true
       workingDirectory: $(System.DefaultWorkingDirectory)
+
+  - ${{ if ne(parameters.TokenToUseForAuth, '') }}:
+    - pwsh: |
+        git config unset --global "http.extraheader"
+      displayName: Removing git config auth header
+      condition: always()
--- a/eng/common/pipelines/templates/steps/sparse-checkout.yml
+++ b/eng/common/pipelines/templates/steps/sparse-checkout.yml
@@ -22,6 +22,7 @@ steps:
   - ${{ if ne(parameters.TokenToUseForAuth, '') }}:
     - pwsh: |
         $base64Token = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes("nobody:${{ parameters.TokenToUseForAuth }}"))
+        Write-Host "##vso[task.setvariable variable=_base64AuthToken;issecret=true;]$base64Token"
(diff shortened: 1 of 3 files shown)


eng/common/pipelines/templates/steps/validate-all-packages.yml
```diff
@@ -4,7 +4,7 @@ parameters:
   ConfigFileDir: $(Build.ArtifactStagingDirectory)/PackageInfo
 
 steps:
- - ${{ if and(ne(variables['Skip.PackageValidation'], 'true'), eq(variables['System.TeamProject'], 'internal')) }}:
+ - ${{ if and(ne(variables['Skip.PackageValidation'], 'true'), and(ne(variables['Build.Reason'], 'PullRequest'), eq(variables['System.TeamProject'], 'internal'))) }}:
```
weshaggard, line 7: @praveenkuttappan do you recall why you added the `not(endsWith(variables['Build.Repository.Name'], '-pr'))` condition below? Does this not work at all in the private repos?

eng/common/pipelines/templates/steps/sparse-checkout.yml
```diff
@@ -67,16 +67,21 @@ steps:
             New-Item $dir -ItemType Directory -Force | Out-Null
             Push-Location $dir
 
+            $ghToken = ""
+            if ($repository.Name.EndsWith("-pr")) {
+              $ghToken = "$(azuresdk-github-pat)@"
```
weshaggard, line 72: I don't like this approach so I'm investigating if I can do this with a git config header.

weshaggard:
test run:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4520016&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e&t=5431112d-2b61-5a5f-7042-ef698f761043

eng/common/pipelines/templates/steps/validate-all-packages.yml
weshaggard: @benbp is conditions like these that make me worry about the test pipeline being in internal.

eng/common/pipelines/templates/steps/sparse-checkout.yml
```diff
@@ -11,11 +11,20 @@ parameters:
...
+    default: ''
 
 steps:
   - ${{ if not(parameters.SkipCheckoutNone) }}:
       - checkout: none
 
+  - ${{ if ne(parameters.TokenToUseForAuth, '') }}:
+    - pwsh: |
+        $base64Token = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes("nobody:${{ parameters.TokenToUseForAuth }}"))
+        git config set --global "http.extraheader" "AUTHORIZATION: basic $base64Token"
```
benbp, line 25: Should we mark this value as secret via devops logging just in case?

weshaggard: Yes, we should mark the base64Token as a secret as well.

weshaggard, line 25:
```suggestion
        "##vso[task.setvariable variable=_base64AuthToken;issecret=true;]$base64Token"
```

eng/common/pipelines/templates/steps/sparse-checkout.yml
```diff
@@ -11,11 +11,20 @@ parameters:
...
+    type: string
+    default: ''
 
 steps:
   - ${{ if not(parameters.SkipCheckoutNone) }}:
       - checkout: none
 
+  - ${{ if ne(parameters.TokenToUseForAuth, '') }}:
+    - pwsh: |
+        $base64Token = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes("nobody:${{ parameters.TokenToUseForAuth }}"))
```
weshaggard, line 24:
```suggestion
        $base64Token = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes("nobody:${{ parameters.TokenToUseForAuth }}"))
        Write-Host "##vso[task.setvariable variable=_base64AuthToken;issecret=true;]$base64Token"
```

PR #9700, opened 2025-01-24
Add fallback user id lookup for test resource deployment

benbp:
Add fallback user id lookup for test resource deployment


weshaggard (approved):
Looks fine as long as a user object with ID is always returned.

PR #9715, opened 2025-01-28
Introduce a root folder to save the logs and artifacts

raych1:
* Introduced a root folder to save all logs and artifacts
* Updated file paths to use the new `out/logs` and `out/generatedSdkArtifacts` directories


raych1:
@chidozieononiwu can you review the change? This is to facilitate the artifact publishing step in pipeline.

maririos (approved):

PR #9731, opened 2025-01-29
Connect to APIView app config using managed identity

praveenkuttappan:
Connect to Azure app config using managed identity so we can disable key based access to app config resource.


benbp (approved):

maririos (approved):

maririos (approved):
Is this something that needs to be added to https://github.com/Azure/azure-sdk-tools/blob/main/src/dotnet/APIView/APIViewWeb/CONTRIBUTING.md ?

weshaggard (approved):

PR #9745, opened 2025-01-30
Pass arm template parameter for TME context

benbp:
Addressing some initial needs from https://github.com/Azure/azure-sdk-for-js/pull/32769

### DIFF
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -528,7 +528,7 @@ try {
         $templateParameters.Add('azsdkPipelineSubnetList', @($env:PoolSubnet))
     }
     # Some arm/bicep templates may want to change deployment settings (e.g. local auth) in sandboxed TME tenants
-    $templateParameters.Add('tenantIsTME', ($context.Tenant.Name -like '*TME*'))
+    $templateParameters.Add('supportsSafeSecretStandard', ($context.Tenant.Name -like '*TME*'))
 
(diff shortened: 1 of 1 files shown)


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -527,6 +527,8 @@ try {
     if ($CI -and $Environment -eq 'AzureCloud' -and $env:PoolSubnet) {
         $templateParameters.Add('azsdkPipelineSubnetList', @($env:PoolSubnet))
     }
+    # Some arm/bicep templates may want to change deployment settings (e.g. local auth) in sandboxed TME tenants
+    $templateParameters.Add('tenantIsTME', ($context.Tenant.Name -like '*TME*'))
```
weshaggard, line 531: Instead of being sub based should this be feature based? I.e. supportsLocalAuth or something like that?

benbp: If possible I would prefer downstream arm templates to be not care about the tenant, so yes.

I wonder if we want to do feature based, or some sort of security level/category? I worry enumerating all the features as more and more parameters could be tedious, vs. something like `disableBroadAccess`.

weshaggard: WE could be broad with our features like SupportsSafeSecretStandard or something along those lines. Either way, I like it more than sub or tenant checking. The main question is how we would make a smart default so that both local and pipeline testing works.

PR #9746, opened 2025-01-31
Disallow absolute recording file paths when an assets.json is provided

scbedd:
During restore, make this an explicit error.

@heaths was getting badrequests due to improperly joining paths. Let's make it explicit so he isn't surprised.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
@@ -325,9 +325,11 @@ final path = C:/repo/sdk-for-net/sdk/anomalydetector/Azure.AI.AnomalyDetector/te
 
 When the user POSTS to `/Record/Stop` the recording will be written to the file as described directly above.
 
+> ⚠️ **Note!** If passing the body key `x-recording-assets-file`, DO NOT pass a fully qualified absolute path. The test-proxy must be able to combine your assets repo location with the file path.
+
 During a `playback` start, the value for `x-recording-file` is used to _load an existing recording into memory_ and serve requests from it!
 
-Please note that if a **absolute** path is presented in header `x-recording-file`. The test-proxy will write directly to that file, wherever it is. If the parent folders do not exist, they will be created at run-time during the write operation.
+Please note that if a **absolute** path is presented in the `x-recording-file` key. The test-proxy will write directly to that file, wherever it is. If the parent folders do not exist, they will be created at run-time during the write operation.
 
 ### Start the test run
 
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
@@ -329,7 +329,7 @@ When the user POSTS to `/Record/Stop` the recording will be written to the file 
 
 During a `playback` start, the value for `x-recording-file` is used to _load an existing recording into memory_ and serve requests from it!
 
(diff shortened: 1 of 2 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
```diff
@@ -325,9 +325,11 @@ final path = C:/repo/sdk-for-net/sdk/anomalydetector/Azure.AI.AnomalyDetector/te
 
 When the user POSTS to `/Record/Stop` the recording will be written to the file as described directly above.
 
+> ⚠️ **Note!** If passing the body key `x-recording-assets-file`, DO NOT pass a fully qualified absolute path. The test-proxy must be able to combine your assets repo location with the file path.
+
 During a `playback` start, the value for `x-recording-file` is used to _load an existing recording into memory_ and serve requests from it!
 
-Please note that if a **absolute** path is presented in header `x-recording-file`. The test-proxy will write directly to that file, wherever it is. If the parent folders do not exist, they will be created at run-time during the write operation.
+Please note that if a **absolute** path is presented in the `x-recording-file` key. The test-proxy will write directly to that file, wherever it is. If the parent folders do not exist, they will be created at run-time during the write operation.
```
benbp, line 332:
```suggestion
Please note that if an **absolute** path is presented in the `x-recording-file` key. The test-proxy will write directly to that file, wherever it is. If the parent folders do not exist, they will be created at run-time during the write operation.
```

This might be my greatest (smallest) nit of all time.

benbp (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
```diff
@@ -325,9 +325,11 @@ final path = C:/repo/sdk-for-net/sdk/anomalydetector/Azure.AI.AnomalyDetector/te
 
 When the user POSTS to `/Record/Stop` the recording will be written to the file as described directly above.
 
+> ⚠️ **Note!** If passing the body key `x-recording-assets-file`, DO NOT pass a fully qualified absolute path. The test-proxy must be able to combine your assets repo location with the file path.
```
benbp, line 328: We do technically have knowledge of the repo root path, so I imagine it would be possible to calculate this. Not sure if you want to go in that direction though.

heaths: FWIW, if there's good reason to allow for absolute paths, I see no reason not to support it. The way the docs read before, it sounded like you didn't have to and I incorrectly assumes the files get created there and shuffled in/out of the `.assets` file. Part of that is just my long history with the old test proxy stuff that originated in the Azure SDK for .NET team where I spent 5 years and helped with some improvements on that. That's on me for assuming.

heaths (approved):

tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md
scbedd: > if there's good reason to allow for absolute paths, I see no reason not to support it. 

We support absolute local paths. Reasoning was that Java and .NET needed it at one point. That being said, providing an absolute path while also providing an assets.json just makes no sense and the proxy should tell you that.

scbedd: > Not sure if you want to go in that direction though.

I definitely do not. I strongly believe that if users provide an assets.json, they are implicitly telling the test-proxy

> I don't know where you live, store my data for me.

PR #9747, opened 2025-01-31
Pass non sparse parameters through to PR job matrix generator

benbp:
@JimSuplizio and I were trying to figure out why the new `java - pullrequest` matrix generation was short a couple jobs from the `java - core - ci` matrix. I found that we aren't plumbing through `NonSparseParameters` from the matrix config.

@scbedd I wasn't sure if this was intended to be skipped or not. If it should be skipped I can change this PR to just add a comment explaining why.

### DIFF
--- a/eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
+++ b/eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
@@ -100,6 +100,7 @@ function GeneratePRMatrixForBatch {
     $matrixResults = @()
     foreach ($matrixConfig in $matrixConfigs) {
       Write-Host "Generating config for $($matrixConfig.Path)"
+      $nonSparse = $matrixConfig.PSObject.Properties['NonSparseParameters'] ? $matrixConfig.NonSparseParameters : @()
 
       $matrixResults = @()
       if ($directBatch) {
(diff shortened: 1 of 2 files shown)


JimSuplizio (approved):
Thank you, @benbp. I'm going to make these changes locally in my java test branch, for testing purposes, until this gets merged in.

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
```diff
@@ -100,12 +100,14 @@ function GeneratePRMatrixForBatch {
...
     foreach ($matrixConfig in $matrixConfigs) {
       Write-Host "Generating config for $($matrixConfig.Path)"
+      $nonSparse = $matrixConfig.PSObject.Properties['NonSparseParameters'] ? $matrixConfig.NonSparseParameters : @()
 
       $matrixResults = @()
       if ($directBatch) {
         $matrixResults = GenerateMatrixForConfig `
           -ConfigPath $matrixConfig.Path `
           -Selection $matrixConfig.Selection `
+          -NonSparseParameters $nonSparse `
```
scbedd, line 110: If we change it here, do you think we should also pass the NonSparseParameters through to the `indirect` matrix generation down below?

benbp: Yep I just forgot to update that :D

PR #9752, opened 2025-01-31
Update query count to fix automation

weshaggard:
Fix issue where we are over 10000 work items so we aren't finding the newer ones. This updates to 100000, hopefully we don't hit that anytime soon.

Also made some of the common logging verbose to remove noise in steps like prepare-release where we don't need to see that once per yml file we parse.

### DIFF
--- a/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
+++ b/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
@@ -55,7 +55,7 @@ function Invoke-Query($fields, $wiql, $output = $true)
   }
 
   $response = Invoke-RestMethod -Method POST `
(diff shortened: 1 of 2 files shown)


eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -55,7 +55,7 @@ function Invoke-Query($fields, $wiql, $output = $true)
   }
 
   $response = Invoke-RestMethod -Method POST `
-    -Uri "https://dev.azure.com/azure-sdk/Release/_apis/wit/wiql/?`$top=10000&api-version=6.0" `
+    -Uri "https://dev.azure.com/azure-sdk/Release/_apis/wit/wiql/?`$top=100000&api-version=6.0" `
```
weshaggard, line 58: @praveenkuttappan this should fix the work item updating issue.

maririos (approved):

PR #9760, opened 2025-02-01
Ensure that invalid variable argument doesn't silently crash

### LINKED ISSUE #9757: Test-Proxy silently crashes on invalid `variables` input
Somehow providing this invalid body:

```json
{"Variables": {"Hello": "World" }}
```

is causing a totally silent crash. No message whatsoever. Should not be possible.

We need to update the `test-proxy` typespec as well.

scbedd:
Remove `FromBody[]` everywhere in the app. It silently crashes when an invalid request is sent.

Resolves #9757 

todo:

- [x] Resolve failing .NET integration test runs


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Admin.cs
```diff
@@ -64,14 +64,21 @@ public async Task AddTransform()
...
+            {
+                PropertyNameCaseInsensitive = true
+            });
+
             var recordingId = RecordingHandler.GetHeader(Request, "x-recording-id", allowNulls: true);
 
             var removedSanitizers = new List<string>();
 
-            if (sanitizerList.Sanitizers.Count == 0)
+            if (sanitizerList.Sanitizers != null && sanitizerList.Sanitizers.Count == 0)
```
scbedd, line 81:
```suggestion
            if (sanitizerList.Sanitizers == null || sanitizerList.Sanitizers.Count == 0)
```

`default(T)` of `<RemoveSanitizerList>` is class with a single property that may be null. Need to handle empty body here.

benbp (approved):

PR #9769, opened 2025-02-04
[js-parser] Support re-exported enums

### LINKED ISSUE #9768: [JS Parser] Add support for renamed enums
Today our (JS) APIView parser uses the `text` property to generate the APIView line for a given item.

This works most of the time, but when you rename an item (either on import or export) the `text` property will contain the original name instead of the publicly visible name.

Example:

```ts
// src/generated/index.ts
export enum KnownJsonWebKeyEncryptionAlgorithm { ... }

// src/index.ts
import { KnownJsonWebKeyEncryptionAlgorithm as KnownEncryptionAlgorithm }
export { KnownEncryptionAlgorithm }
```

output doc-model from APIExtractor:

```json
{
     "canonicalReference": "@azure/keyvault-keys!KnownEncryptionAlgorithms:enum",
     "excerptTokens": [
         {
            "kind": "Content",
            "text": "export declare enum KnownJsonWebKeyEncryptionAlgorithm "
        }
     ],
     "members": [ ]
     "fileUrlPath": "src/generated/models/models.ts",
     "kind": "Enum",
     "name": "KnownEncryptionAlgorithms",
     "preserveMemberOrder": false,
     "releaseTag": "Public"
}
```

In this case, the APIView will show KnownJsonWebKeyEncryptionAlgorithm but we have all the information needed to display the correct name. This issue tracks making changes to the parser to support this scenario.

maorleger:
Adds support for re-exported enums, ensuring APIView shows the correct (i.e. exported and customer-visible) name for an enum.

I scoped this just to enums for now, but we can expand if we find other usages.

I also went ahead and added a basic test with a basic test fixture. Future PR will work with @praveenkuttappan to ideally add this to CI builds for the parser.

Resolves #9768

### DIFF
--- a/tools/apiview/parsers/js-api-parser/vitest.config.ts
+++ b/tools/apiview/parsers/js-api-parser/vitest.config.ts
@@ -1,7 +1,7 @@
 import { defineConfig } from "vitest/config";
 export default defineConfig({
   test: {
-    include: ["**/*.spec.ts"],
+    include: ["test/**/*.spec.ts"],
     watch: false,
   },
 });
--- a/tools/apiview/parsers/js-api-parser/vitest.config.ts
+++ b/tools/apiview/parsers/js-api-parser/vitest.config.ts
@@ -1,0 +1,7 @@
+import { defineConfig } from "vitest/config";
+export default defineConfig({
+  test: {
+    include: ["**/*.spec.ts"],
+    watch: false,
+  },
+});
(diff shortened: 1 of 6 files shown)


jeremymeng (approved):

tools/apiview/parsers/js-api-parser/vitest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from "vitest/config";
+export default defineConfig({
+  test: {
+    include: ["**/*.spec.ts"],
```
jeremymeng, line 4: nit: 
```suggestion
    include: ["test/**/*.spec.ts"],
```

PR #9783, opened 2025-02-05
Print clickable link to terminal

heaths:


heaths:
I ran into this the other day when I had to manually install bicep. Most modern terminals auto-detect URLs but only if there's a supported scheme. This will make the URL clickable in those terminals e.g., Windows Terminal or iTerm2 on macOS.

scbedd (approved):

PR #9786, opened 2025-02-06
fix: replace `import.meta.resolve` with `require.resolve` for Node 18+ compatibility

irmowan:
Under Node 18+, the `tsp-client convert` tool met this issue:

```
TypeError: (intermediate value).resolve is not a function
    at getPathToDependency (file:///C:/nvm/v18.15.0/node_modules/@azure-tools/typespec-client-generator-cli/dist/utils.js:55:50)
    at convertCommand (file:///C:/nvm/v18.15.0/node_modules/@azure-tools/typespec-client-generator-cli/dist/commands.js:321:32)
    at async Object.handler (file:///C:/nvm/v18.15.0/node_modules/@azure-tools/typespec-client-generator-cli/dist/index.js:170:5)
```

In Node.js, there is no built-in `import.meta.resolve()` API. That function is a browser feature and isn’t implemented in Node 18 (nor in later versions at time of writing). If you need to resolve a module path in an ES module, the simplest solution is to use createRequire from the module package. 

After I patched this commit in my local module, the issue is fixed.

### DIFF
--- a/tools/tsp-client/src/utils.ts
+++ b/tools/tsp-client/src/utils.ts
@@ -2,7 +2,6 @@ import { joinPaths, normalizeSlashes } from "@typespec/compiler";
 import { randomUUID } from "node:crypto";
 import { access, constants, mkdir, writeFile } from "node:fs/promises";
 import { dirname, join } from "node:path";
-import { fileURLToPath } from "node:url";
 import { Logger } from "./log.js";
 import { TspLocation } from "./typespec.js";
 import { normalizeDirectory } from "./fs.js";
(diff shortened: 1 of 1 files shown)


catalinaperalta (changes requested):
Thanks for reaching out about this. Unfortunately, we [require node 18.19+ to use tsp-client ](https://github.com/Azure/azure-sdk-tools/blob/dbc0bd485c6ee3d1a16e6fe8e0b980e0da8ddc7a/tools/tsp-client/package.json#L29C3-L31C5) since we use several components that are not compatible with lower node versions. I wouldnt want to add this simply to try to support a lower node version for some commands in the tool.

PR #9802, opened 2025-02-10
Fix Migration SDK Type Detection

wanlwanl:


tools/js-sdk-release-tools/src/common/utils.ts
```diff
@@ -66,20 +66,30 @@ function printErrorDetails(
...
-    const paraPath = getClassicClientParametersPath(packageRoot);
     const packageName = getNpmPackageName(packageRoot);
     if (packageName.startsWith('@azure-rest/')) {
         return SDKType.RestLevelClient;
     }
-    const exist = shell.test('-e', paraPath);
-    const type = exist ? SDKType.HighLevelClient : SDKType.ModularClient;
+    
+    const paraPath = getClassicClientParametersPath(packageRoot);
+    const npmParaPath = getNpmClassicClientParametersPath(packageRoot);
```
qiaozha, line 87: would it be problematic if the latest npm version is HLC after the esm migration and we are moving to modular with TypeSpec in the current PR?

wanlwanl: Oh, forget to rename the function, the current behavior is to check both `src/models/parameters.ts` and `dist/esm/models/parameters,js`, if one of them exists, the sdk is HLC.

qiaozha (approved):

PR #9803, opened 2025-02-10
Get-ObjectKey converts boolean values

scbedd:
This PR enhances the `Group-ByObjectKey` function to coalesce boolean-like values to their true boolean representation before getting a hashcode for them.

We use this grouping function to group a list of packages by which matrix config they belong to. The issue is that due to devops shenanigans, these matrix configs are getting their booleans a bit changed. The "default" matrix that is fed into `Create-PrJobMatrix` is a result of a devops json conversion of the static yaml parameter. Due to this, the booleans are getting messed up.

```yml
MatrixConfigs:
      - Name: NET_ci_test_base
        Path: eng/pipelines/templates/stages/platform-matrix.json
        Selection: sparse
        GenerateVMJobs: true
```

But devops _converts_ this to:

```json
{
  "Name": "NET_ci_test_base",
  "Path": "eng/pipelines/templates/stages/platform-matrix.json",
  "Selection": "sparse",
  "GenerateVMJobs": "True"
}
```

Notice that `GenerateVMJobs` value is different and not really a boolean anymore? This was causing matrix coalesce to treat these as _different matrices_ when they were NOT intended to be. This PR addresses this possible issue.

### DIFF
--- a/eng/common-tests/helpers/Package-Helpers.tests.ps1
+++ b/eng/common-tests/helpers/Package-Helpers.tests.ps1
@@ -21,7 +21,7 @@ Describe "Matrix-Collation" {
         }
     }
 
-    It "Should properly group identical matrix inputs" {
+    It -Skip "Should properly group identical matrix inputs" {
         $Pkgs = @(
             @{
                 Name = "package1"
@@ -53,7 +53,7 @@ Describe "Matrix-Collation" {
         $groupingResults[$group2][1].Name | Should -Be "package3"
     }
 
-    It "Should properly group items with no setting" {
+    It -Skip "Should properly group items with no setting" {
         $Pkgs = @(
             @{
                 Name = "package1"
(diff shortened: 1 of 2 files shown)


eng/common-tests/helpers/Package-Helpers.tests.ps1
```diff
@@ -53,7 +113,7 @@ Describe "Matrix-Collation" {
         $groupingResults[$group2][1].Name | Should -Be "package3"
     }
 
-    It "Should properly group items with no setting" {
+    It -Skip "Should properly group items with no setting" {
```
scbedd, line 116: Will revert. These are for my convenience while exploring the issue.

JimSuplizio (approved):
I'm approving this. For everyone else's benefit, these changes were tested in an existing test [PR](https://github.com/Azure/azure-sdk-for-java/pull/44121). What was going on is that the test runs weren't getting collated correctly, for example azure-core and azure-identity have the same test matrix but because this was converting values they were being computed as separate matrices which, because they're effectively the same run, were causing the duplicate runs to get dropped.

PR #9807, opened 2025-02-11
Replace `CIMatrixConfigs` with generalized `CIParameters` within `Package-Properties`

scbedd:
We need a way to make a few `ci.yml` parameters available to the `package-properties` class. We need to do this for a couple of the .NET parameterized checks. Specifically:

- `AOTCompat: boolean`
- `CheckSnippets: boolean`

These are opt-in, so anyone forwarding `CheckSnippets: false` _aren't_ expecting check snippets to run. We can't currently honor this scenario in `net - pullrequest`.

This PR does two things:

- [x] Adds `CIParameters` hashtable to `PackageProps` class.
- [x] Removes `CIMatrixConfigs` as a direct prop of `PkgProp` and instead adds it as key `CIMatrixConfigs` _within_ `CIParameters`
- [x] Externalizes the "give me the ci.yml parameters for this package" in `Package-Properties.ps1`

My intent is to utilize `GetCIYmlForArtifact` within the `Get-AllPackageInformationFromRepo` function in `.NET`. From there, I'll add key-value pairs representing these values to the `CIParameters` hashtable.

The reason I'm offering it this way, is that `Save-Package-Properties.ps1` doesn't currently offer customizations as an argument, we rely on each individual repo to implement the "give me all the metadata" functions. We will get each generated PkgProps, then before returning it we'll parse the CIYml and add these two parameters if necessary.

The `analyze` steps will then be able to check these parameter values to see which packages need these optional checks run.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -21,7 +21,6 @@ class PackageProps {
     [string[]]$AdditionalValidationPackages
     [HashTable]$ArtifactDetails
     [HashTable]$CIParameters
-    [HashTable[]]$CIMatrixConfigs
 
     PackageProps([string]$name, [string]$version, [string]$directoryPath, [string]$serviceDirectory) {
         $this.Initialize($name, $version, $directoryPath, $serviceDirectory)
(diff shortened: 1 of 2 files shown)


eng/common/scripts/Package-Properties.ps1
```diff
@@ -112,28 +114,38 @@ class PackageProps {
...
-
+    [PSCustomObject]GetCIYmlForArtifact() {
         $RepoRoot = Resolve-Path (Join-Path $PSScriptRoot ".." ".." "..")
 
         $ciFolderPath = Join-Path -Path $RepoRoot -ChildPath (Join-Path "sdk" $this.ServiceDirectory)
         $ciFiles = Get-ChildItem -Path $ciFolderPath -Filter "ci*.yml" -File
+        $ciArtifactResult = $null
+
+        foreach ($ciFile in $ciFiles) {
+            $ciArtifactResult = $this.ParseYmlForArtifact($ciFile.FullName)
```
benbp, line 125: Just reading through the current and old code, are we getting lucky that we resolve ci.yml before ci.<variant>.yml and/or how do we resolve those differences? I think it's fair for us to require anything go in ci.yml (or a variant if no ci.yml exists) if you want it to be processed by the auto-pr pipeline.

scbedd: > are we getting lucky that we resolve ci.yml before ci.variant.yml and/or how do we resolve those difference

No, because we're looking _for the artifact in the artifactlist for the ci.yml_. That is our "yes this is the right ci.yml." This logic falls through if the artifact belongs to both, but I believe that's in error if the team has split their `ci.yml` into `ci.data.yml` and `ci.mgmt.yml`.

scbedd: I see where you're coming from though. This needs further refactoring to make it work the way we expect.

scbedd: Ok I'm going to resolve this comment chain, the format of this is going to be much more obvious now, even though I need to integration test a bit.

benbp (approved):

PR #9815, opened 2025-02-11
Encode `RollForward` behavior for test-proxy

scbedd:
So that users don't need to set `DOTNET_ROLL_FORWARD=Major` if they have a NEWER version of dotnet other than what the project was compiled upon. 

CC @alzimmermsft


weshaggard (approved):
I'm good with this. I would keep .NET as it is the LTS version .NET 9 is not LTS.

PR #9821, opened 2025-02-12
Fix the expression used for AddDevVersion param in save package properties

praveenkuttappan:
Save package properties add DEV version only if `AddDevVersion` switch is passed and currently this is incorrectly set in yaml. Script also needs a change since this is wrapped in a yaml to pass `AddDevVersion` a bool to avoid duplicate step in the pipeline yaml.

### DIFF
--- a/eng/common/pipelines/templates/steps/save-package-properties.yml
+++ b/eng/common/pipelines/templates/steps/save-package-properties.yml
@@ -61,10 +61,11 @@ steps:
   - ${{ else }}:
       - task: Powershell@2
         displayName: Save package properties
+        condition: succeeded()
         inputs:
           filePath: ${{ parameters.ScriptDirectory }}/Save-Package-Properties.ps1
           arguments: >
             -ServiceDirectory '${{parameters.ServiceDirectory}}'
             -OutDirectory '${{ parameters.PackageInfoDirectory }}'
-            -AddDevVersion:$${{ eq(variables['SetDevVersion'],'true') }}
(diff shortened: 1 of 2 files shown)


eng/common/pipelines/templates/steps/save-package-properties.yml
```diff
@@ -61,10 +61,11 @@ steps:
   - ${{ else }}:
       - task: Powershell@2
         displayName: Save package properties
+        condition: succeeded()
         inputs:
           filePath: ${{ parameters.ScriptDirectory }}/Save-Package-Properties.ps1
           arguments: >
             -ServiceDirectory '${{parameters.ServiceDirectory}}'
             -OutDirectory '${{ parameters.PackageInfoDirectory }}'
-            -AddDevVersion:$${{ eq(variables['SetDevVersion'],'true') }}
```
praveenkuttappan, line 69: AddDevVersion is a switch, and we can't pass a boolean to override the value of switch. DevOps also does not process expression evaluation at this level, so I have changed it to use variable value to pass the bool flag to script.

scbedd (approved):
This should work consistently. My only concern is that on multi-platform where capitalization of key matters, should we use converted `SETDEVVERSION`? Does it matter?

weshaggard (approved):

praveenkuttappan:
> This should work consistently. My only concern is that on multi-platform where capitalization of key matters, should we use converted `SETDEVVERSION`? Does it matter?

Ideally yes. But I think it shouldn't matter in this case.

PR #9823, opened 2025-02-13
Handle defaults in `changedFiles` and `deletedFiles`

scbedd:
@hallipr this repairs the issue found in Azure/azure-sdk-for-net#48269

### DIFF
--- a/eng/common/scripts/Generate-PR-Diff.ps1
+++ b/eng/common/scripts/Generate-PR-Diff.ps1
@@ -53,12 +53,19 @@ $deletedFiles = Get-ChangedFiles -DiffPath $TargetPath -DiffFilterType "D"
 if ($changedFiles) {
   $changedServices = Get-ChangedServices -ChangedFiles $changedFiles
 }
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Generate-PR-Diff.ps1
```diff
@@ -53,13 +53,20 @@ $deletedFiles = Get-ChangedFiles -DiffPath $TargetPath -DiffFilterType "D"
...
+  $changedFiles = @()
+}
 
 # ExcludePaths is an object array with the default of [] which evaluates to null.
 # If the value is null, set it to empty list to ensure that the empty list is
 # stored in the json
 if (-not $ExcludePaths) {
   $ExcludePaths = @()
 }
+if (-not $deletedFiles) {
```
scbedd, line 67: We are really checking for `$null` but frankly I prefer `-not $obj` to `$null -ne $deletedFiles`.

hallipr (approved):

PR #9834, opened 2025-02-14
chmod +x TestResources scripts

heaths:


heaths:
For those of us working primarily in linux or mac, this really helps so I don't have to start pwsh just to run the scripts. If I forget I'm running in another shell, I may set env vars that won't float back up to the parent bash process or whatever. Gets frustrating.

weshaggard:
@heaths I'll approve the sync PRs once they get created.

Issue #9829, opened 2025-02-13
JS Automation Tool Fails to Process Multiple Packages


raych1:
This is a [batch run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4562486&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=c10d5f44-55ce-55d7-7975-407ed75d9a96&l=6889) for all the typespecs. Part of the errors are as follows:

```
Err: Error: Find unexpected changed package directory. Length: 4. Value: sdk/documentintelligence/ai-document-intelligence-rest, sdk/face/ai-vision-face-rest, sdk/vision/ai-vision-image-analysis-rest, sdk/healthinsights/azurehealthinsightsradiologyinsights-rest. Please only change files in one directory
2025-02-13T19:19:27.8957298Z 07:19:27.895 [4m[32mcmdout[39m[24m 	[automation_generate.sh] Stderr: "undefined"
2025-02-13T19:19:27.8958200Z 07:19:27.895 [4m[32mcmdout[39m[24m 	[automation_generate.sh] Stdout: "undefined"
2025-02-13T19:19:27.8960076Z 07:19:27.895 [4m[32mcmdout[39m[24m 	[automation_generate.sh] ErrorStack: "Error: Find unexpected changed package directory. Length: 4. Value: sdk/documentintelligence/ai-document-intelligence-rest, sdk/face/ai-vision-face-rest, sdk/vision/ai-vision-image-analysis-rest, sdk/healthinsights/azurehealthinsightsradiologyinsights-rest. Please only change files in one directory
2025-02-13T19:19:27.8961798Z 07:19:27.895 [4m[32mcmdout[39m[24m 	[automation_generate.sh]     at generateRLCInPipeline (/opt/hostedtoolcache/node/22.13.0/x64/lib/node_modules/@azure-tools/js-sdk-release-tools/dist/llc/generateRLCInPipeline/generateRLCInPipeline.js:184:23)
2025-02-13T19:19:27.8963516Z 07:19:27.895 [4m[32mcmdout[39m[24m 	[automation_generate.sh]     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
[...]

PR #9851, opened 2025-02-17
Remove requirement of only modifying 1 folder

wanlwanl:
Fix: https://github.com/Azure/azure-sdk-tools/issues/9829

Solution: remove the check to limit the number of changed folders more than one (while the the old tool pass only one `readmeMd` already, no need to check the number)


tools/js-sdk-release-tools/src/llc/generateRLCInPipeline/generateRLCInPipeline.ts
```diff
@@ -194,19 +194,6 @@ export async function generateRLCInPipeline(options: {
     const outputPackageInfo = getOutputPackageInfo(options.runningEnvironment, options.readmeMd, options.typespecProject);
 
     try {
-        // TODO: need to refactor
```
MaryGao, line 197: Have we verified this?

MaryGao (approved):
Feel free to merge if you verified this works in batch mode.

tools/js-sdk-release-tools/src/llc/generateRLCInPipeline/generateRLCInPipeline.ts
wanlwanl: No, since the prod pipeline has no batch mode. Only ray can help run in piepline v2 that supports batch mode.
It's a save change for prod.

PR #9862, opened 2025-02-19
Relying on `CIParameters` for storage of trigger information

scbedd:
FYI @benbp @JimSuplizio 

This PR:

- [x] Adds support for saving trigger paths from the `ci.yml` for each pkg into the `pkgprops` under `CITriggerPaths`.
- [x] ~~Honor the `CITriggerPaths` with fallback to artifact `triggerPaths` to implement non-sdk path triggers in `Get-PrPkgProperties`~~

This covers the "if eng/ changes add this package purely for validation".

The issue I discovered is that if we honor the default `trigger` from each `ci.yml`, we would end up causing `pullrequest` builds to just target the entire `servicedirectory` again. We need the inclusion to be at the _artifact_ level.

So for instances like `eventhub` for .NET. Where a given service directory contains multiple `ci.yml` that all target different artifact lists. This enables specific coverage of packages that depend on a specific directory path that might not even be a valid package. So for eventhub, that looks like this:

```
     - name: Azure.Messaging.EventHubs.Processor
       safeName: AzureMessagingEventHubsProcessor
       triggeringPaths:
         - sdk/eventhub/Azure.Messaging.EventHubs.Shared
```

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -126,14 +126,7 @@ class PackageProps {
 
             if ($ciArtifactResult) {
                 $this.ArtifactDetails = [Hashtable]$ciArtifactResult.ArtifactConfig
-                $this.CIParameters["CITriggerPaths"] = @()
                 $this.CIParameters["CIMatrixConfigs"] = @()
-
-                $triggers = GetValueSafelyFrom-Yaml $ciArtifactResult.ParsedYml @("trigger", "paths", "include")
-
-                if ($triggers) {
-                    $this.CIParameters["CITriggerPaths"] += $triggers
-                }
 
                 # if we know this is the matrix for our file, we should now see if there is a custom matrix config for the package
                 $matrixConfigList = GetValueSafelyFrom-Yaml $ciArtifactResult.ParsedYml @("extends", "parameters", "MatrixConfigs")
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -230,6 +230,11 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
             }
             $filePath = (Join-Path $RepoRoot $file)
 
+            # but I believe that we have to include the exclude paths as well
+            # otherwise we will not handle when the trigger path has exclude. like for java/core/ci.yml
+            # in that instance, the trigger path is sdk/core but we want to exclude sdk/core/azure-core-v2
+            # so we need to ensure that we are including the exclude paths when we evaluate the additional trigger paths
+            # this is still todo though, a follow-up commit will add this
             foreach($triggerPath in $triggerPaths) {
                 # utilize the various trigger paths against the targeted file here
             }
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -244,6 +244,9 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
 
                 if ($resolvedRelativePath) {
                     $shouldInclude = $shouldInclude -or ($filePath -like (Join-Path "$resolvedRelativePath" "*"))
+                    # if we reached here, this package wasn't directly changed, but instead
+                    # should be validated because it was indirectly changed
+                    $pkg.IncludedForValidation = $true
                 }
             }
 
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -210,6 +210,13 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
         $lookupKey = ($pkg.DirectoryPath).Replace($RepoRoot, "").TrimStart('\/')
         $lookup[$lookupKey] = $pkg
 
+        # resolve the trigger paths for the package, either from individual artifact config OR by falling back to associated ci.yml file trigger paths
+        $triggerPaths = $pkg.CIParameters["CITriggerPaths"]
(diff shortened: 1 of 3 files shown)


eng/common/scripts/Package-Properties.ps1
```diff
@@ -196,11 +204,21 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
...
+    # this is the primary loop that identifies the packages that have changes
     foreach ($pkg in $allPackageProperties) {
         $pkgDirectory = Resolve-Path "$($pkg.DirectoryPath)"
         $lookupKey = ($pkg.DirectoryPath).Replace($RepoRoot, "").TrimStart('\/')
         $lookup[$lookupKey] = $pkg
 
+        # we only honor the individual artifact triggers
+        # if we were to honor the ci-level triggers, we would simply
+        # end up in a situation where any change to a service would
+        # still trigger every package in that service. individual package triggers
```
weshaggard, line 216: Wouldn't these be scoped to the individual artifacts listed in a given ci.yml file?

eng/common/scripts/Package-Properties.ps1
```diff
@@ -241,10 +280,18 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
         }
     }
 
+    # now pass along the set of packages we've identified, the diff itself, and the full set of package properties
+    # to locate any additional packages that should be included for validation
     if ($AdditionalValidationPackagesFromPackageSetFn -and (Test-Path "Function:$AdditionalValidationPackagesFromPackageSetFn")) {
```
weshaggard, line 285: I assume this would be removed once we convert folks over.

scbedd: I _think_ this will have to stick around to cover a few of the more esoteric functions in `java`.

I will prove that out by actually updating the `language-settings` directly in the sync/eng branches created for this PR.

eng/common/scripts/Package-Properties.ps1
scbedd: When we discussed this offline, we started with

> can we just inherit the trigger by default and only override if necessary

Here is a package.json

```json
{
  "Name": "Azure.Messaging.EventHubs",
  "ArtifactDetails": {
    "name": "Azure.Messaging.EventHubs",
    "triggeringPaths": [
      "Azure.Messaging.EventHubs.Shared"
    ],
    "safeName": "AzureMessagingEventHubs"
  },
  "CIParameters": {
    "CITriggerPaths": [
      "sdk/eventhub/ci.yml",
      "sdk/eventhub/service.projects",
      "sdk/eventhub/Azure.Messaging.EventHubs",
      "sdk/eventhub/Azure.Messaging.EventHubs.Processor",
      "sdk/eventhub/Azure.Messaging.EventHubs.Shared"
    ],
    "CIMatrixConfigs": []
  }
}
```

See how the ci.yml triggers include every package associated with the build def? If we included every package for validation that fell under any of those paths, any change to `sdk/eventhub/azure.messaging.eventhubs` will also trigger indirect tests of `sdk/eventhub/Azure.Messaging.EventHubs.Processor.` However, the behavior we _actually_ want is for changes to `Azure.Messaging.EventHubs.Shared` to trigger indirect testing of `Azure.Messaging.EventHubs` and `Azure.Messaging.EventHubs.Processor`.

weshaggard: Ok that seems like a substring match issue which would even occur with similar trigger paths on the artifact. 

Perhaps we fix this by updating the paths to include a "/" at the end so we avoid this substring matching.

weshaggard: I think I miss understood your comment to where it actually included the other libraries, but you are suggesting here that it will include all the other track 2 libraries when we don't really need to do that. I do agree it will do that but that is also what it does today and if folks want to optimize it they can always move triggers to the artifacts to override.

benbp:
I'm fine with this approach. I think overloading this config file could get messier down the road if we keep finding new use case requirements for more config fields. I don't have a better idea right now about how to do this, so if it ends up being a problem we can always refactor.

eng/common/scripts/Package-Properties.ps1
weshaggard: Ok let's leave out the ci triggers for now, we don't even need to gather them in CIParameters if we aren't going to use them.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -126,6 +126,14 @@ class PackageProps {
 
             if ($ciArtifactResult) {
                 $this.ArtifactDetails = [Hashtable]$ciArtifactResult.ArtifactConfig
+                $this.CIParameters["CITriggerPaths"] = @()
+                $this.CIParameters["CIMatrixConfigs"] = @()
+
+                $triggers = GetValueSafelyFrom-Yaml $ciArtifactResult.ParsedYml @("trigger", "paths", "include")
```
weshaggard, line 132: If we don't plan on using these lets not even pull them in.

weshaggard (approved):
I left one comment about removing the citriggers if we aren't going to use them but otherwise I'm fine with this for now.

benbp (approved):

eng/common/scripts/Package-Properties.ps1
scbedd: Yep I just wanted to have it for discussion. I'll clean it up now.

eng/common/scripts/Package-Properties.ps1
weshaggard: I would like to understand what scenarios still need this after we do our conversations. If we don't have any or we can generalize them into common I'd like to remove this.

PR #9878, opened 2025-02-20
[SDK Gen] Add condition before transport end

JackTn:


tools/spec-gen-sdk/src/automation/logging.ts
```diff
@@ -139,8 +139,8 @@ export const loggerWaitToFinish = async (logger: winston.Logger) => {
   for (const transport of logger.transports) {
     if (transport instanceof winston.transports.File) {
       if (transport.end) {
+          await new Promise<void>((resolve) => logger.on('finish', resolve));
```
raych1, line 142: Can you try with this code:
```typescript
/**
 * Ensures all logs are written to file transports
 */
export const loggerWaitToFinish = async (logger: winston.Logger) => {
  logger.info('Wait for logger transports to complete');
  
  const fileTransports = logger.transports.filter((t) => t instanceof winston.transports.File);

  if (fileTransports.length === 0) {
    return;
  }

  await Promise.all(
    fileTransports.map(transport => 
      new Promise<void>(resolve => {
        transport.once('finish', resolve);
        transport.end();
      })
    )
  );
};
```

JackTn: ```NodeError: write after end``` happened and find another issues. Will fix!

tools/spec-gen-sdk/src/automation/logging.ts
```diff
@@ -139,8 +139,8 @@ export const loggerWaitToFinish = async (logger: winston.Logger) => {
   for (const transport of logger.transports) {
     if (transport instanceof winston.transports.File) {
       if (transport.end) {
-          transport.end();
           await setTimeout(2000);
+          transport.end();
```
raych1, line 143: Will 2000 seconds be not enough time if there are many logs?

JackTn: Based on the observation that the `NodeError: write after end` does not occur with small data sets, I analyzed the write times for each log. Regardless of the data size, the write time for each log is consistently between 1ms and 10ms. The log length itself does not cause this issue. The error arises because there are unwritten logs in the stream used by Winston. To address this, ensure a delay of a few milliseconds to write the queued logs. I believe the missing logs do not exceed 10, so a delay of **1000ms** or **2000ms** should suffice.

| Message Length | Cost Time |
| --- | --- |
| 30 | 5.408 ms |
| 196 | 4.408 ms |
| 52 | 0.590 ms |
| 149 | 3.554 ms |
| 100 | 3.554 ms |
| 28039 | 3.554 ms |

tools/spec-gen-sdk/src/automation/logging.ts
raych1: try this:
```typescript
export const loggerWaitToFinish = async (logger: winston.Logger) => {
  logger.info('Wait for logger transports to complete');
  await Promise.all(
    logger.transports.map(
      (transport) =>
        new Promise<void>((resolve) => {
          if (transport instanceof winston.transports.File && transport.end) {
            transport.on('finish', () => resolve());
            transport.end();
          } else {
            resolve();
          }
        })
    )
  );
};
```

JackTn: not works ~ `NodeError: write after end` happened

raych1 (approved):

PR #9886, opened 2025-02-20
Handle `push` update of `assets.json` when member `Tag` is not present

### LINKED ISSUE #9759: Test-Proxy push when `Tag` is not present results in successful push with no tag update
Documentation states that Tag should be added on the fly.

scbedd:
Resolves #9759

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
@@ -836,13 +836,22 @@ namespace Azure.Sdk.Tools.TestProxy.Store
                 // we deliberately do an extremely stripped down version parse and update here. We do this primarily to maintain
                 // any comments left in the assets.json though maintaining attribute ordering is also nice. To do this, we read all the file content, then
                 // simply replace the existing Tag value with the new one, then write the content back to the json file.
-
                 var currentSHA = (await ParseConfigurationFile(config.AssetsJsonLocation.ToString())).Tag;
                 var content = await File.ReadAllTextAsync(config.AssetsJsonLocation.ToString());
                 if (String.IsNullOrWhiteSpace(currentSHA))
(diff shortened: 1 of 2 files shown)


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Store/GitStore.cs
```diff
@@ -836,13 +836,22 @@ public async Task UpdateAssetsJson(string newSha, GitAssetsConfiguration config)
...
+                    {
+                        string pattern = @"""Tag"":\s*""\s*""";
+                        content = Regex.Replace(content, pattern, $"\"Tag\": \"{newSha}\"", RegexOptions.IgnoreCase);
+                    }
+                    // if not we just have to reserialize the entire thing. This edge case is not worth the amount of extra code
+                    // necessary to maintain the cohesion of the tag file.
+                    else
+                    {
+                        content = JsonSerializer.Serialize(config, new JsonSerializerOptions { WriteIndented = true });
+                    }
```
benbp, line 854: Is this whole block redundant with the existing value of content?

scbedd: I don't think so? Look at it this way:

- if currentSha is null OR whitespace
  - If the content contains `Tag`, then it's just a case of `"Tag": ""`, we can update in place and maintain existing ordering/comments in the assets.json
  - if the content doesn't contain `Tag` member, we will just reserialize, destroying any comments or ordering in the assets.json.
- else
  - It's not null or whitespace, so there is an existing valid value for `Tag` that we can just replace.

benbp: I see, so the re-serialization does have some formatting enforcement. That's what I was wondering or if we just wanted to keep the file on disk the same.

benbp (approved):

PR #9902, opened 2025-02-22
Moving more discovery into `Get-PrPkgProperties`

scbedd:
This PR handles service-level changes without language-specific settings.

- [x] Add owning `ci.yml` to `triggeringPaths` by default. 
- [x] Handle service-level changes where there is a single `ci.yml` present in the service directory for the package.
  - Just add the artifacts within the `ci.yml` to the set of artifacts for testing
- [x] When there are multiple ci.yml files in service directory, and a file that is NOT a ci.yml is changed, emit a warning

This approach is working the way I expect in: Azure/azure-sdk-for-net#48379

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -256,29 +256,27 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
             # for this task. This is because we can resolve a service directory from the ci.yml, and if
             # there is a single ci.yml in that directory, we can assume that any file change in that directory
             # will apply to all packages that exist in that directory.
-            if ($pathComponents.Length -eq 3 -and $pathComponents[0] -eq "sdk") {
-                $triggeringCIYmls = $triggeringPaths | Where-Object { $_ -like "*ci*.yml" }
+            $triggeringCIYmls = $triggeringPaths | Where-Object { $_ -like "*ci*.yml" }
 
-                foreach($yml in $triggeringCIYmls) {
-                    # given that this path is coming from the populated triggering paths in the artifact,
-                    # we can assume that the path to the ci.yml will successfully resolve.
-                    $ciYml = Join-Path $RepoRoot $yml
-                    # ensure we terminate the service directory with a /
-                    $directory = [System.IO.Path]::GetDirectoryName($ciYml).Replace("`\", "/") + "/"
-                    $soleCIYml = (Get-ChildItem -Path $directory -Filter "ci*.yml" -File).Count -eq 1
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Package-Properties.ps1
```diff
@@ -219,12 +228,13 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
...
             $shouldInclude = $filePath -like (Join-Path "$pkgDirectory" "*")
 
-            # this implementation guesses the working directory of the ci.yml
+            # handle changes to files that are RELATED to each package
             foreach($triggerPath in $triggeringPaths) {
                 $resolvedRelativePath = (Join-Path $RepoRoot $triggerPath)
-                # utilize the various trigger paths against the targeted file here
+
                 if (!$triggerPath.StartsWith("/")){
                     $resolvedRelativePath = (Join-Path $RepoRoot "sdk" "$($pkg.ServiceDirectory)" $triggerPath)
```
scbedd, line 239: We can utilize the `triggeringPaths` details here to grab the actual containing directory of the `ci.yml`....that may be FUD though and $pkg.ServiceDirectory is entirely reliable even for deeper ci.yml files.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -237,6 +247,38 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
...
+                    break
+                }
+            }
+
+            # handle service-level changes to the ci.yml files
+            # we are using the ci.yml file being added automatically to each artifactdetails as the input
+            # for this task. This is because we can resolve a service directory from the ci.yml, and if
+            # there is a single ci.yml in that directory, we can assume that any file change in that directory
+            # will apply to all packages that exist in that directory.
+            if ($pathComponents.Length -eq 3 -and $pathComponents[0] -eq "sdk") {
```
benbp, line 259: This assumption breaks in a few edge cases, namely go (e.g. https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/aztables/ci.yml)

scbedd: Yeah. This `if` check will mean we never evaluate this piece of code. I'm going to move this logic within the $triggeringPaths iteration. We're kinda uselessly iterating over the triggering paths again here.

scbedd: Ok. This version is checking out.

```
  "ChangedFiles": [
    "sdk/core/tests.yml"
  ],
```

is successfully triggering the `core` packages that belong to the single `core/ci.yml`.

weshaggard (approved):

PR #9915, opened 2025-02-25
[apiview-gpt] update to use openai sdk / o3-mini

kristapratico:


packages/python-packages/apiview-gpt/dev_requirements.txt
```diff
@@ -1,6 +1,7 @@
 flask
 knack
-langchain
-openai
-pydantic
+langchain==0.0.287
```
tjprescott, line 3: Don't we have to dump langchain?

tjprescott (approved):
LGTM, though maybe we should add the outputs to a gitignore so they aren't commited to the repo.

kristapratico:
> LGTM, though maybe we should add the outputs to a gitignore so they aren't commited to the repo.

I was thinking we might want to keep them for testing purposes / comparing outputs from different models

packages/python-packages/apiview-gpt/dev_requirements.txt
kristapratico: Removed

Issue #9587, opened 2025-01-06
[Python APIView] overload docstring not parsed properly


swathipil:
The keywords in the azure-appconfiguration-provider `load` function docstring aren't being parsed properly.

https://github.com/mrm9084/azure-sdk-for-python/blob/cf55cc0110b74b5ccf949e588f2157432f1deb66/sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py#L126

The `load` overloads are not showing in the APIView: https://apiview.dev/Assemblies/Review/109d5798466d4d33ad472e991ad9d478?revisionId=ea29d1a73cf843cc85d574c04fbef745&diffOnly=False&doc=False&diffRevisionId=20068860641846ebbb201ea005d8c9d5

swathipil:
Parser isn't checking top-level functions for overloads. Need to fix.

PR #9919, opened 2025-02-26
[APIView] parse module level function overloads

swathipil:
fixes: #9587

* Also updated timeout value during package installation step, for Windows.
* Remove unused get_navigation method


tjprescott (approved):
Looks fine to me. If feasible, it might be nice to combine the logic for parsing module-level functions and class-level methods. Right now, it looks like the logic is copy-paste-edit, which could introduce maintenance burden.

PR #9921, opened 2025-02-26
Code needs to handle more than one template library

JimSuplizio:
If there are no packages with changes detected we add the template libraries for validation. The problem here is that the code incorrectly assumed that there's only one template library but, because of batch release in Java, we have more than one. If IncludedForValidation isn't set to true for all of them then it'll end up with multiple test matrices, one for the indirect packages and one for the direct packages.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -320,7 +320,9 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     # packages. We should never return NO validation.
     if ($packagesWithChanges.Count -eq 0) {
         $packagesWithChanges += ($allPackageProperties | Where-Object { $_.ServiceDirectory -eq "template" })
-        $packagesWithChanges[0].IncludedForValidation = $true
+        foreach ($package in $packagesWithChanges) {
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Package-Properties.ps1
```diff
@@ -320,7 +320,9 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     # packages. We should never return NO validation.
     if ($packagesWithChanges.Count -eq 0) {
         $packagesWithChanges += ($allPackageProperties | Where-Object { $_.ServiceDirectory -eq "template" })
-        $packagesWithChanges[0].IncludedForValidation = $true
+        foreach ($package in $packagesWithChanges) {
```
weshaggard, line 323: While I'm not opposed to this, I'm curious if you think there is enough value to do all over just one?

benbp (approved):

eng/common/scripts/Package-Properties.ps1
JimSuplizio: Are per our conversation this adds all template libraries and, for Java, there's 3 and they all need to be set.

srnagar (approved):

PR #9922, opened 2025-02-26
[Python APIView] fix empty line rendering in diff

swathipil:
Set RelatedToLine on empty lines so that they're rendered in the APIView diff.

Tests have been updated to count RelatedToLine on empty lines.

TODO:
- [x] add test


packages/python-packages/apiview-stub-generator/tests/_test_util.py
```diff
@@ -47,16 +47,10 @@ def _check(actual, expected, client):
 MockApiView = ApiView(pkg_name="test", namespace="test")
 
 def _count_review_line_metadata(tokens, metadata):
-    lastRelatedTo = None
     for token in tokens:
-        # check that the relatedToLine is the same as the lineID of the next token
-        if lastRelatedTo:
```
swathipil, line 53: This logic no longer applies since RelatedToLine won't always be the next token, so dropping this check.

praveenkuttappan (approved):

PR #9934, opened 2025-02-27
fix: use full type name when checking for ClientOptions type

jorgerangel-msft:
This PR introduces a fix for AZC0007, where this analyzer was incorrectly firing for clients even when the expected constructors were present. Rather than trying to load the `Azure.Core.ClientOption` type during compilation start, and using this loaded type to validate if a constructor's last parameter matches it, we now simply just check the parameter's base type's name to see if it matches `Azure.Core.ClientOption`.

image

sdk-for-net pr:  https://github.com/Azure/azure-sdk-for-net/pull/48474

### DIFF
--- a/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0007Tests.cs
+++ b/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0007Tests.cs
@@ -73,5 +73,35 @@ namespace RandomNamespace
 }";
             await Verifier.VerifyAnalyzerAsync(code);
         }
+
+        // This test validates that the analyzer does not produce a diagnostic for clients with multiple constructors
+        // which end with a ClientOptions parameter.
+        [Fact]
+        public async Task AZC0007NotProducedForClientWithMultipleCtors()
+        {
+            const string code = @"
+using System;
+using Azure;
+using Azure.Core;
+
+namespace RandomNamespace.Foo
+{
+
+    public partial class RoomsClientOptions : ClientOptions {} 
+
+    public partial class RoomsClient
+    {
+        protected RoomsClient() {}
+        public RoomsClient(string connectionString) {}
+        public RoomsClient(string connectionString, RoomsClientOptions options) {}
+        public RoomsClient(Uri endpoint, AzureKeyCredential credential, RoomsClientOptions options = null) {}
+        public RoomsClient(Uri endpoint, TokenCredential credential, RoomsClientOptions options = null) {}
+    }
+
+
+}";
+            await Verifier.VerifyAnalyzerAsync(code);
+        }
     }
 }
+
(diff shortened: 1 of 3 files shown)


src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/SymbolAnalyzerBase.cs
```diff
@@ -20,17 +21,9 @@ public sealed override void Initialize(AnalysisContext context)
...
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
             context.EnableConcurrentExecution();
-            context.RegisterCompilationStartAction(CompilationStart);
             context.RegisterSymbolAction(c => Analyze(new RoslynSymbolAnalysisContext(c)), SymbolKinds);
         }
 
-#pragma warning disable RS1012 // Start action has no registered actions.
-        protected virtual void CompilationStart(CompilationStartAnalysisContext context)
-        {
-            ClientOptionsType = context.Compilation.GetTypeByMetadataName("Azure.Core.ClientOptions");
```
jorgerangel-msft, line 30: My current theory is that due to the asynchronous nature of the Roslyn analyzer system this might've been problematic.  We can achieve a similar result without having to load this type.

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers.Tests/AZC0007Tests.cs
```diff
@@ -70,8 +70,38 @@ public class SomeClient
...
+    public partial class RoomsClient
+    {
+        protected RoomsClient() {}
+        public RoomsClient(string connectionString) {}
+        public RoomsClient(string connectionString, RoomsClientOptions options) {}
+        public RoomsClient(Uri endpoint, AzureKeyCredential credential, RoomsClientOptions options = null) {}
+        public RoomsClient(Uri endpoint, TokenCredential credential, RoomsClientOptions options = null) {}
+    }
+
+
```
jsquire, line 101:
```suggestion
```

nittiest of nits.

jsquire (approved):

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/SymbolAnalyzerBase.cs
```diff
@@ -53,7 +46,12 @@ protected bool IsClientOptionsType(ITypeSymbol typeSymbol)
             ITypeSymbol baseType = typeSymbol.BaseType;
             while (baseType != null) 
             {
-                if (SymbolEqualityComparer.Default.Equals(baseType, ClientOptionsType))
+                // validate if the base type is Azure.Core.ClientOptions
+                var fullName = baseType.ContainingNamespace.GetFullNamespaceName()
+                    .Append(".")
+                    .Append(baseType.Name);
```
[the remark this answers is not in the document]
jorgerangel-msft: sure, I will open a follow up PR with this since this pr is merged.

PR #9943, opened 2025-02-28
Use subscription name to assign supportsSafeSecretStandard

benbp:
The `.Tenant.Name` property is null in the pipeline's login context.

I'm not a fan of this approach, but it's either subscription name checking or hardcoded tenant id checking.

### DIFF
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -121,6 +121,8 @@ param (
 . $PSScriptRoot/TestResources-Helpers.ps1
 . $PSScriptRoot/SubConfig-Helpers.ps1
 
+$wellKnownTMETenants = @('70a036f6-8e4d-4615-bad6-149c02e7720d')
+
 if (!$ServicePrincipalAuth) {
     # Clear secrets if not using Service Principal auth. This prevents secrets
     # from being passed to pre- and post-scripts.
@@ -527,8 +529,11 @@ try {
     if ($CI -and $Environment -eq 'AzureCloud' -and $env:PoolSubnet) {
         $templateParameters.Add('azsdkPipelineSubnetList', @($env:PoolSubnet))
     }
-    # Some arm/bicep templates may want to change deployment settings (e.g. local auth) in sandboxed TME tenants
-    $templateParameters.Add('supportsSafeSecretStandard', ($context.Tenant.Name -notlike '*TME*'))
+    # The TME tenants are our place for local auth testing so we do not support safe secret standard there.
+    # Some arm/bicep templates may want to change deployment settings like local auth in sandboxed TME tenants.
+    # The pipeline account context does not have the .Tenant.Name property, so check against subscription via
+    # naming convention instead.
+    $templateParameters.Add('supportsSafeSecretStandard', ($wellKnownTMETenants.Contains($TenantId)))
 
     $defaultCloudParameters = LoadCloudConfig $Environment
     MergeHashes $defaultCloudParameters $(Get-Variable templateParameters)


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -528,7 +528,9 @@ try {
         $templateParameters.Add('azsdkPipelineSubnetList', @($env:PoolSubnet))
     }
     # Some arm/bicep templates may want to change deployment settings (e.g. local auth) in sandboxed TME tenants
-    $templateParameters.Add('supportsSafeSecretStandard', ($context.Tenant.Name -notlike '*TME*'))
+    # The pipeline account context does not have the .Tenant.Name property, so check against subscription via
+    # naming convention instead.
+    $templateParameters.Add('supportsSafeSecretStandard', ($context.Subscription.Name -notlike '*TME*'))
```
weshaggard, line 533: This seems like it might have false positive matches if tme shows up anywhere in the sub name. While it isn't ideal perhaps just match on the TME tenant id.

benbp: @weshaggard I pushed a new approach.

weshaggard (approved):

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -270,6 +270,11 @@ try {
             '4d042dc6-fe17-4698-a23f-ec6a8d1e98f4' = 'Azure SDK Test Resources - TME '
         }
 
+        $wellKnownTenantTMEStatus = @{
+            '70a036f6-8e4d-4615-bad6-149c02e7720d' = $true  # tme testing
```
weshaggard, line 274: I would invert this bool to make it clearer and match the usages of supportsSafeSecretStandard

benbp: I'm not too pleased with the inverted conditional I added here. The tradeoff I considered when setting it this way is that we may have additional template conditionals based on TME status, so I wanted a simple hash that marked whether a tenant was TME or not. That way we could base other `supportsXYZStandard` off this. Given that, naming it the opposite `wellKnownTenantNonTMEStatus` seemed a bit odd.

I'm open to either approach here.

weshaggard: If you just want this to be well-known TME subs then I would just make it a list of only the TME tenants and contain nothing else. So, it only answers the question of is this TME or not.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -528,7 +533,13 @@ try {
         $templateParameters.Add('azsdkPipelineSubnetList', @($env:PoolSubnet))
     }
     # Some arm/bicep templates may want to change deployment settings (e.g. local auth) in sandboxed TME tenants
-    $templateParameters.Add('supportsSafeSecretStandard', ($context.Tenant.Name -notlike '*TME*'))
+    # The pipeline account context does not have the .Tenant.Name property, so check against subscription via
+    # naming convention instead.
+    if ($wellKnownTenantTMEStatus.Contains($TenantId)) {
```
weshaggard, line 538: Perhaps a note that says TME tenants allows testing local auth and thus does not support safe secret standards.

weshaggard (approved):

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -270,6 +270,10 @@ try {
             '4d042dc6-fe17-4698-a23f-ec6a8d1e98f4' = 'Azure SDK Test Resources - TME '
         }
 
+        $wellKnownTMETenants = @{
```
weshaggard, line 273: minor but this could just be an array.

benbp: :D yepppp

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -527,8 +529,11 @@ try {
     if ($CI -and $Environment -eq 'AzureCloud' -and $env:PoolSubnet) {
         $templateParameters.Add('azsdkPipelineSubnetList', @($env:PoolSubnet))
     }
-    # Some arm/bicep templates may want to change deployment settings (e.g. local auth) in sandboxed TME tenants
-    $templateParameters.Add('supportsSafeSecretStandard', ($context.Tenant.Name -notlike '*TME*'))
+    # The TME tenants are our place for local auth testing so we do not support safe secret standard there.
+    # Some arm/bicep templates may want to change deployment settings like local auth in sandboxed TME tenants.
+    # The pipeline account context does not have the .Tenant.Name property, so check against subscription via
+    # naming convention instead.
+    $templateParameters.Add('supportsSafeSecretStandard', ($wellKnownTMETenants.Contains($TenantId)))
```
deyaaeldeen, line 536: I think TME doesn't support safe secret standard, so this check should be inverted, see the earlier check above.

deyaaeldeen: I opened https://github.com/Azure/azure-sdk-tools/pull/10013

PR #9945, opened 2025-02-28
Allow Targeted Disable of PRBatching

scbedd:
See title!

I'm thinking I see some interesting test cases around if we want to force the addition of a suffix. I don't think so.

### DIFF
--- a/eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
+++ b/eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
@@ -144,6 +144,16 @@ function GeneratePRMatrixForBatch {
       }
     }
 
+    if ($matrixConfig.PSObject.Properties['PRBatch']) {
+      # if we are doing a PR Batch, we need to just add the matrix items directly to the OverallResult
+      # as the users have explicitly disabled PR batching for this matrix.
+      if (!$matrixConfig.PRBatch) {
+        Write-Host "The matrix config $($matrixConfig.Path) has PRBatch set to false, the matrix members will be directly added without batching by $PRMatrixSetting."
+        $OverallResult += $matrixResults
+        continue
+      }
+    }
+
     $packageBatches = Split-ArrayIntoBatches -InputArray $matrixBatch -BatchSize $BATCHSIZE
 
     # we only need to modify the generated job name if there is more than one matrix config + batch
--- a/eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
+++ b/eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
@@ -144,10 +144,10 @@ function GeneratePRMatrixForBatch {
       }
     }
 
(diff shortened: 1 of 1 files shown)


eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
```diff
@@ -144,6 +144,16 @@ function GeneratePRMatrixForBatch {
       }
     }
 
+    if ($matrixConfig.PSObject.Properties['PRBatch']) {
```
benbp, line 147: We should update the matrix and/or unified pr FAQ as well

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
```diff
@@ -144,6 +144,16 @@ function GeneratePRMatrixForBatch {
...
+      # if we are doing a PR Batch, we need to just add the matrix items directly to the OverallResult
+      # as the users have explicitly disabled PR batching for this matrix.
+      if (!$matrixConfig.PRBatch) {
+        Write-Host "The matrix config $($matrixConfig.Path) has PRBatch set to false, the matrix members will be directly added without batching by $PRMatrixSetting."
```
benbp, line 151: Does `.Path` point to the json file? Could be misleading since it's in the yaml.

benbp (approved):

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
JimSuplizio: @benbp the Path setting is part of the matrix configuration that identifies the where the json file is. @scbedd, maybe the name + path in the message?
```
AdditionalMatrixConfigs:
  - Name: version_overrides_tests
    Path: sdk/core/version-overrides-matrix.json
    Selection: all
    GenerateVMJobs: true
```

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
JimSuplizio: @scbedd is this supposed to be `PRBatch `or `PRBatching`? In our chat it was `PRBatching: false`

JimSuplizio (approved):
@scbedd, I'm going to pre-approve but there are a couple of questions that might require tiny changes.

eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
scbedd: I started with `PRBatching`, but ben had `PRBatch` and I just accepted the correction without it being said. That being said I prefer `PRBatching` so I'm just going to go with that.

scbedd:
> @scbedd, I'm going to pre-approve but there are a couple of questions that might require tiny changes.

Took all those adjustments. Thanks @JimSuplizio

PR #9954, opened 2025-03-03
Python API review is not generated using correct package name

praveenkuttappan:
API parser has a new change to get package name and version using pkginfo and this only works when gnerating review file from wheel or sdist. But python CI generates token from package source hence pkginfo is not available and as a result review token file is generated with wrong version and wrong token file name.

Parser also iterates all directories including tests and build and include those paths in package modules.


swathipil (approved):
Lgtm, thanks for the fix @praveenkuttappan  !!

Issue #9469, opened 2024-12-05
Refine AZC0030 Error Message Regarding `Options` Suffix


archerzz:
## Description

Right now, in `AZC0030` we prompt to replace `Options` suffix with `Config` for model names. That's too rigid. We should:
- point out that `Options` is not a good name
- recommend renaming the model according to Azure .NET SDK guidelines

Here is what to change:
- update the error message for `Options` like below:
> Model `XXXOptions` can be returned from service calls so should not use the `Options` suffix. Please rename it. For details, see: [url]

- compose a MD file give details about why and how to rename, and attach the link in the error message above

Issue #9335, opened 2024-11-06
[C# analyzers] Update AZC0030 so that it doesn't prevent clients from having serializable models with the Options suffix


annelo-msft:
It is a common pattern in data plane clients to have model types with the `Options` suffix.

This issue tracks the work to understand and document the intended scope of analyzer warning AZC0030, so it can be scoped down to address only intended cases.

This warning was introduced in https://github.com/Azure/azure-sdk-tools/pull/6916 and modified in https://github.com/Azure/azure-sdk-tools/pull/7250.  @jsquire has suggested that it was intended to prevent subtypes of `ClientOptions` from being serializable.  We need to confirm this is the intended scope so we don't change an analyzer that is needed for other cases.

If this is the correct intended scope, the analyzer should be updated to validate specifically that subtypes of `ClientOptions` do not implement the `IPersistableModel<T>` interface, and not prevent any type with the `Options` suffix from being serializable.

@archerzz, @m-nash, @christothes, would it be possible to help us understand the intended scope of this warning?

archerzz:
@annelo-msft  This rule was from .NET Mgmt SDK Review Guide: 
![Image](https://github.com/user-attachments/assets/a9c87137-626b-4610-b35d-965e7f20a77b)

The rule regarding `Options` seems have been changed since I implemented the analyzer. The latest version is opposite to what Jesse suggested.

Anyway, the key point is that when I wrote those rules, **they were for mgmt SDK only**. See the original repo: https://github.com/archerzz/Azure.MgmtSdk.Analyzers

Then when we added them into current analyzer package, we thought those are generic and we should them enable for all SDKs.

To resolve this issue, I think we could:
- remove `Options` suffix check from `AZC0030`
- create a `Azure.MgmtSdk.Analzyers` package, put that rule and any further mgmt specific rules into that package
- in our build script, enable `Azure.MgmtSdk.Analzyers` only for mgmt SDKs

cc @ArthurMa1978

ArthurMa1978:
This convention rule has been confirmed by @KrzysztofCwalina. If we need alternative suggestions, we'll require input from KC.

annelo-msft:
Super helpful context, many thanks for sharing it, @archerzz and @ArthurMa1978!

I worked with @m-nash and @KrzysztofCwalina to clarify the guidance around the options parameter pattern in our Azure SDK C# guidelines.  Existing guidance is here:
[...]

PR #9961, opened 2025-03-03
[Azure.ClientSdk.Analyzers] Refine AZC0030

jorgerangel-msft:
This PR introduces fixes for these items:
- [Refine AZC0030 Error Message Regarding Options Suffix](https://github.com/Azure/azure-sdk-tools/issues/9469)
   - The analyzer that was validating models with the Options suffix was updated to include a general diagnostic message which links to our guidelines regarding naming models. It will no longer suggest a name to use as a replacement.
- [Update AZC0030 so that it doesn't prevent clients from having serializable models with the Options suffix](https://github.com/Azure/azure-sdk-tools/issues/9335).
  - The analyzer that was validating models with the Options suffix was scoped down to management plane models.

sdk-for-net test pr: https://github.com/Azure/azure-sdk-for-net/pull/48597

### DIFF
--- a/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ModelName/OptionsSuffixAnalyzer.cs
+++ b/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ModelName/OptionsSuffixAnalyzer.cs
@@ -1,0 +1,76 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Azure.ClientSdk.Analyzers.ModelName
+{
+    /// <summary>
+    /// Analyzer that checks model names ending with "Options". This analyzer shares the diagnostic ID with <see cref="GeneralSuffixAnalyzer"/>.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class OptionsSuffixAnalyzer : SuffixAnalyzerBase
+    {
+        private const string AzureResourceManagerNamespaceName = "Azure.ResourceManager";
+        private const string OptionsSuffix = "Options";
+        private const string IUtf8JsonSerializable = "IUtf8JsonSerializable";
+        private const string JsonElement = "JsonElement";
+        private static readonly string[] suffixes = new string[] { OptionsSuffix };
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Descriptors.AZC0030); } }
+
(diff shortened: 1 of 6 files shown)


src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ModelName/OptionsSuffixAnalyzer.cs
```diff
@@ -0,0 +1,76 @@
...
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class OptionsSuffixAnalyzer : SuffixAnalyzerBase
+    {
+        private const string AzureResourceManagerNamespaceName = "Azure.ResourceManager";
+        private const string OptionsSuffix = "Options";
+        private const string IUtf8JsonSerializable = "IUtf8JsonSerializable";
+        private const string JsonElement = "JsonElement";
+        private static readonly string[] suffixes = new string[] { OptionsSuffix };
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Descriptors.AZC0030); } }
```
jorgerangel-msft, line 23: Since we want a different behavior for when a model is suffixed with "Options", I've introduced this new class which still shares the original diagnostic  ID with `GeneralSuffixAnalyzer`. My main motivation was to reduce the blast radius of this change, in the case that users have suppressed the original diagnostic ID. If we feel like these changes warrant a new ID, it should be setup to easily do so.

jsquire: Good thinking.

jsquire, line 23: This pattern will allocate each time the property is called.   Since the underlying array is immutable, we should be able to share the same instance.
```suggestion
        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } => 
            ImmutableArray.Create(Descriptors.AZC0030);
```

src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ModelName/OptionsSuffixAnalyzer.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            var fullNamespace = symbol.ContainingNamespace.GetFullNamespaceName();
+
+            if ($"{fullNamespace}".Equals(AzureResourceManagerNamespaceName)
+                || $"{fullNamespace}".StartsWith($"{AzureResourceManagerNamespaceName}."))
+            {
+                return true;
+            }
+
+            return false;
```
jsquire, line 64:
```suggestion
            return ($"{fullNamespace}".Equals(AzureResourceManagerNamespaceName)
                || $"{fullNamespace}".StartsWith($"{AzureResourceManagerNamespaceName}."));
```

jsquire (approved):

PR #9963, opened 2025-03-04
Optimize Service-Level changes, update ordering of targeted files

scbedd:
We want to catch `direct` changes _first_.

Simplified algo for matching a package to a changed file:

```
foreach changedFile:
  if file like $pkgDirectory -> direct
    break
  if file like triggeringPath -> indirect
    break
  if file changed in service directory -> indirect
    break
```

Now, take this example set of `targetedFiles` that we're iterating through:

```jsonc
{
  // ...
  "ChangedFiles": [
    "sdk/cloudmachine/Azure.Projects.sln",
    "sdk/cloudmachine/Azure.Projects/CHANGELOG.md",
  ]
  // ...
}
```

In the above situation, we are going to process `Azure.Projects.sln` _first_, evaluate

> oh that's an indirect change to this package, don't need to continue looking at the rest of the `targetedFiles`!

Which is a problem! By sorting targetedFiles by depth, we can ensure that we always catch direct changes first.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -199,6 +199,14 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     $additionalValidationPackages = @()
     $lookup = @{}
 
+    # Sort so that we very quickly find any directly changed packages before hitting service level changes.
+    # This is important because due to the way we traverse the changed files, the instant we evaluate a pkg
+    # as directly or indirectly changed, we exit the file loop and move on to the next pkg.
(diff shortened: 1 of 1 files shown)


benbp (approved):

scbedd:
> LGTM. FYI this is another assumption that will break with go, as some packages have all or almost all files at the top level (e.g. https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/data/aztables). I could see sorting by suffix or something to address this when we hit it.

Well I'm starting on that rn! You're absolutely right. Now that being said it does look like that is a single ci.yml. I'll test!

PR #9967, opened 2025-03-04
Tools typespec bump deps

weidongxu-microsoft:


tools/typespec-bump-deps/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "main": "dist/index.js",
+  "type": "module",
+  "bin": {
+    "typespec-bump-deps": "./cmd/cli.mjs"
+  },
+  "scripts": {
+    "watch": "tsc -p ./tsconfig.build.json --watch",
+    "build": "tsc -p ./tsconfig.build.json",
+    "clean": "rimraf dist/ temp/",
+    "test": "vitest run"
```
MaryGao, line 14: do we run test in ci? i have no special requirement for config and package files since they are just template codes.

weidongxu-microsoft: I haven't set up the ci yet.

It should be something similar to 
https://github.com/Azure/azure-sdk-tools/blob/main/tools/tsp-client/ci.yml

Let me add the ci.yml first.

weshaggard (approved):
Please also add a CODEOWNER entry for this tool.

PR #9993, opened 2025-03-06
Early exit non-service changes, cache getchilditem results

scbedd:
We got a diff that had ~24k files. We have a lot to check, but still need to optimize best we can.

- [x] After further discussion with @benbp and @hallipr , going to update this logic to execute a pre-pass across the changed files. We will keep around any files that correlate to a `triggeringPath`, but otherwise we are going to simplify the changed files to changed directories. This reduces a prospective change of 24k files to ~1k.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -198,6 +198,7 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
 
     $additionalValidationPackages = @()
     $lookup = @{}
+    $directoryIndex = @{}
 
     # Sort so that we very quickly find any directly changed packages before hitting service level changes.
     # This is important because due to the way we traverse the changed files, the instant we evaluate a pkg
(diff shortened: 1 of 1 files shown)


weshaggard (approved):

eng/common/scripts/Package-Properties.ps1
```diff
@@ -206,9 +262,11 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
     # To avoid this without wonky changes to the detection algorithm, we simply sort our files by their depth, so we will always
     # detect direct package changes first!
     $targetedFiles = $targetedFiles | Sort-Object { ($_.Split("/").Count) } -Descending
+    $pkgCounter = 1
```
benbp, line 265: This seems unused, aside from incrementing?

eng/common/scripts/Package-Properties.ps1
```diff
@@ -243,20 +336,12 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
...
-                # if we are including this package due to one of its additional trigger paths, we need
-                # to ensure we're counting it as included for validation, not as an actual package change
-                if ($resolvedRelativePath) {
-                    $includedForValidation = $filePath -like (Join-Path "$resolvedRelativePath" "*")
-                    $shouldInclude = $shouldInclude -or $includedForValidation
-                    if ($includedForValidation) {
-                        $pkg.IncludedForValidation = $true
-                    }
-                    break
+                $includedForValidation = $filePath -like (Join-Path "$resolvedRelativePath" "*")
```
scbedd, line 339: This is now completed when resolving the artifact to begin with. Users can place their relative paths in their `ci.yml`, but we will only ever see the full format EG:

within `sdk/eventhub/ci.yml`
```yml
- artifactname
  triggeringPaths:
    - Azure.Messaging.EventHubs.Shared
```

becomes 

```
sdk/eventhub/Azure.Messaging.EventHubs.Shared`
```

benbp (approved):

PR #10012, opened 2025-03-07
Questions

hallipr:


eng/common/scripts/Package-Properties.ps1
```diff
@@ -240,6 +240,7 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
             # handle direct changes to packages
             $shouldInclude = $filePath -like (Join-Path "$pkgDirectory" "*")
 
+            #should we only do this if $shouldInclude is false?  i.e. no reason to include as indirect if already included
```
scbedd, line 243: 👍

eng/common/scripts/Package-Properties.ps1
```diff
@@ -266,6 +270,8 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
             # will apply to all packages that exist in that directory.
             $triggeringCIYmls = $triggeringPaths | Where-Object { $_ -like "*ci*.yml" }
 
+            # what would happen if someone included a ci.yml in the triggering paths?
```
scbedd, line 273: It's already always there implicitly due to `triggeringPaths` artifact update in `InitializeCIArtifactDetails`.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -281,6 +287,11 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
                 }
 
                 if ($soleCIYml -and $filePath.Replace("`\", "/").StartsWith($directory)) {
+                    #  `sdk/computefleet/Azure.ResourceManager.ComputeFleet/CHANGELOG.md` exists
+                    # and `sdk\compute\ci.mgmt.yml` exists
+                    # does this mean that when `sdk/computefleet/Azure.ResourceManager.ComputeFleet/CHANGELOG.md` changes, the projects under `sdk/compute/ci.mgmt.yml` will be triggered?
+                    # because `sdk/computefleet/Azure.ResourceManager.ComputeFleet/CHANGELOG.md` starts with `sdk/compute`
```
scbedd, line 293: No, because we are only entering this code path if it is a changed file at the service directory level. Where service-directory-level is defined by "in the same directory as my owning ci.yml"

scbedd: Scroll up to line 278 just out of view.

eng/common/scripts/Package-Properties.ps1
hallipr: I mean... What happens if, for some reason, eventhubs/ci.yml includes:
```
      triggeringPaths:
      - /sdk/storage/ci.yml
```

Do we make assumptions about ci*.yml triggeringPaths that would be wrong if someone made an explicit, cross service reference to a ci*.yml file?

eng/common/scripts/Package-Properties.ps1
hallipr: aah... 
```
 $serviceDirectoryChange = (Split-Path $filePath -Parent).Replace("`\", "/") -eq $directory
  if (!$serviceDirectoryChange) {
      break
  }
  ```
  
This foreach loop only cares about files that are direct siblings of the `*ci*.yml` files from triggeringPaths when there was only 1 `*ci*.yml` in the folder.

PR #10019, opened 2025-03-10
typespec-bump-deps, fix packaging

weidongxu-microsoft:


tools/typespec-bump-deps/package.json
```diff
@@ -1,16 +1,21 @@
...
+  "files": [
+    "dist",
+    "cmd/cli.mjs"
+  ],
   "scripts": {
     "watch": "tsc -p ./tsconfig.build.json --watch",
     "build": "tsc -p ./tsconfig.build.json",
-    "clean": "rimraf dist/ temp/",
+    "clean": "rimraf dist/",
+    "prepack": "npm run clean && npm run build",
```
weidongxu-microsoft, line 18: The cause is this.

I use the same build ci from tsp-client, but I didn't aware that they've included the `build` step in `pack` but here I didn't.

PR #10039, opened 2025-03-13
Ensure that direct batches have an underscore in front of them

scbedd:
See title.

image

Would prefer to avoid that.


eng/common/scripts/job-matrix/Create-PrJobMatrix.ps1
```diff
@@ -205,7 +205,7 @@ function GeneratePRMatrixForBatch {
         }
 
         if ($batchSuffixNecessary) {
-          $outputItem["name"] = $outputItem["name"]  + "_$batchNamePrefix$batchCounter"
```
scbedd, line 208: I saw this when it snuck in and I didn't want to pay another eng/common round trip when I did. Finally addressing it 😆

weshaggard (approved):

PR #10047, opened 2025-03-14
[API View] Extending rust-api-parser support 

HarshaNalluru:
Follow up of https://github.com/Azure/azure-sdk-tools/pull/9592
## Improves/adds support for the following features
1. assoc_const aesthetics
2. func spaces
3. trait - remove extra punctuation
4. null check in processGenerics
5. "unsafe" traits
6. Negative implementations 
7. impl generics and where predicates
8. generics for assoc_type
9. qualified_path#generics_args
10. processGenericArgs for type.qualified_path.trait.args
11. Update RenderClasses with existing classes for now

And refactoring

### DIFF
--- a/tools/apiview/parsers/rust-api-parser/src/main.ts
+++ b/tools/apiview/parsers/rust-api-parser/src/main.ts
@@ -46,6 +46,7 @@ function addSectionHeader(codeFile: CodeFile, headerText: string): void {
         Value: `/* ${headerText} */`,
         NavigateToId: `header-${headerText}`,
         NavigationDisplayName: `/* ${headerText} */`,
+        RenderClasses: ["namespace"],
       },
     ],
   });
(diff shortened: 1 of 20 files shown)


heaths (approved):

tools/apiview/parsers/rust-api-parser/src/main.ts
```diff
@@ -46,6 +46,7 @@ function addSectionHeader(codeFile: CodeFile, headerText: string): void {
         Value: `/* ${headerText} */`,
         NavigateToId: `header-${headerText}`,
         NavigationDisplayName: `/* ${headerText} */`,
+        RenderClasses: ["namespace"],
```
chidozieononiwu, line 49: Why is punctuation set to render as a namespace?

HarshaNalluru: It was intentional.
image

Since the rendered output is too long. I wanted to show quick links to the re-exported items in particular.
It is merely a delimiter of the whole output.
I'll keep that for now, I'll change if we re-consider for the next iteration.

PR #10062, opened 2025-03-17
eng, codeowner, weidxu to azure-rest-api-specs-examples-automation

weidongxu-microsoft:
Mostly because we are having stricter merge requirement soon.

Is it preferrable we try to have 2 (or more) persons in CODEOWNERS for the tool (so if owner raise a PR, the other owner can review and approve)?


weshaggard:
You can add another owner but it isn't required.

weshaggard (approved):

PR #10064, opened 2025-03-17
Simplify find para path

wanlwanl:


tools/js-sdk-release-tools/src/mlc/apiVersion/apiVersionTypeExtractor.ts
```diff
@@ -13,8 +15,11 @@ export const getApiVersionType: IApiVersionTypeExtractor = async (
     typeFromClient = await getApiVersionTypeFromRestClient(packageRoot, clientPattern, tryFindRestClientPath);
     if (typeFromClient !== ApiVersionType.None) return typeFromClient;
 
-    const parametersFolder = "src/rest";
-    const typeFromOperations = getApiVersionTypeFromOperations(packageRoot, parametersFolder, findParametersPath);
+    const parametersPath = join(packageRoot, "src/rest/parameters.ts");
```
MaryGao, line 18: does this work for modular? i think there is no rest folder in modular now.

wanlwanl: usually client has api version, so the case for this is rare.
this used to work for modular client.
this is why I want to sync bewteen js emitter and sdk.

wanlwanl: I feel like we can keep it as before for now since this pr focusing on refactor.
We can schedule a meeting to clean up all the out-of-date code in the future.

MaryGao: sure, i believe this path `src/rest/parameters.ts` is no longer valid, feel free to check other places also.

MaryGao: then my next question is since this code doesn't work, does it produce any errors in ci?

MaryGao (approved):
Approved since just code refactoring.

tools/js-sdk-release-tools/src/mlc/apiVersion/apiVersionTypeExtractor.ts
wanlwanl: Yes, track here: https://github.com/Azure/autorest.typescript/issues/3118

PR #10097, opened 2025-03-20
[TypeSpec APIView] Update for TypeSpec 0.67

tjprescott:
Removes `--version` option that isn't compatible with TSP 0.67 and wasn't used anyways. Updates for breaking changes in TSP 0.67.

### DIFF
--- a/tools/apiview/emitters/typespec-apiview/src/apiview.ts
+++ b/tools/apiview/emitters/typespec-apiview/src/apiview.ts
@@ -1,4 +1,12 @@
 import {
+  Expression,
+  getNamespaceFullName,
+  getSourceLocation,
+  Namespace,
+  navigateProgram,
+  Program
+} from "@typespec/compiler";
+import { 
   AliasStatementNode,
   ArrayExpressionNode,
   ArrayLiteralNode,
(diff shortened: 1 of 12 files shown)


tools/apiview/emitters/typespec-apiview/src/apiview.ts
```diff
@@ -73,10 +75,10 @@ export class ApiView {
   typeDeclarations = new Set<string>();
   includeGlobalNamespace: boolean;
 
-  constructor(name: string, packageName: string, packageVersion?: string, includeGlobalNamespace?: boolean) {
+  constructor(name: string, packageName: string, includeGlobalNamespace?: boolean) {
     this.name = name;
     this.packageName = packageName;
-    this.packageVersion = packageVersion ?? "";
+    this.packageVersion = "ALL";
```
praveenkuttappan, line 81: Does this mean TypeSpec API review won't have any version info?

praveenkuttappan (approved):
Looks good to me except a question on package version

mikeharder (approved):

tools/apiview/emitters/typespec-apiview/src/apiview.ts
tjprescott: It only ever contained the tool version. The package version isn't written anywhere and that has been true the whole time since the Powershell script that invoked it never supplied a version.

PR #10098, opened 2025-03-20
[Python APIView] update namespace discovery for package

swathipil:
Azure-core tests will be included in #10032.

Whl/sdist namespace discovery logic has been updated to check extracted folder structure. Currently, whl/sdist looks at the top_level.txt to determine namespace, ("azure") rather than finding the innermost namespace ("azure.comos"). This causes inherited methods to be included in the APIView, since the package namespace is the same as the top level namespace of azure.core. (i.e. Compares azure=azure, rather than azure.core =/= azure.cosmos.)


praveenkuttappan (approved):
Looks fine to me. One question on requirements file. Is this change done to support removing setup.py? If that's the case, including all required dependencies in txt file the new approach or including a pyproject.toml

swathipil:
@praveenkuttappan The update to the apiview_reqs.txt was done so that updates to any of those dependencies won't change the parsing/token file unexpectedly.

tjprescott (approved):

PR #10099, opened 2025-03-20
Add option to create self contained test resources post script

benbp:
Add option to create self contained test resources post script

### DIFF
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -93,6 +93,11 @@ param (
     })]
     [array] $AllowIpRanges = @(),
 
+    # Instead of running the post script, create a wrapped file to run it with parameters
+    # so that CI can run it in a subsequent step with a refreshed azure login
+    [Parameter()]
+    [string] $SelfContainedPostScript,
+
     [Parameter()]
     [switch] $CI = ($null -ne $env:SYSTEM_TEAMPROJECTID),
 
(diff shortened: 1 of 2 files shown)


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -626,8 +631,37 @@ try {
...
+                    }
+                }
+                $deserialized['ResourceGroupName'] = $ResourceGroupName
+                $deserialized['DeploymentOutputs'] = $deploymentOutputs
+                $serialized = $deserialized | ConvertTo-Json
+
+                $outScript = @"
+`$parameters = `@'
+$serialized
+'`@ | ConvertFrom-Json -AsHashtable
```
weshaggard, line 653: Why serialize in the script as opposed to serializing the entire command line into a string to write to the script?

benbp: The only way I can think of to convert instances arrays/objects and switches into powershell literals is via json serialization.

benbp: Or rather, the json conversion avoids having to de-compile runtime powershell into a valid script.

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -93,6 +93,11 @@ param (
     })]
     [array] $AllowIpRanges = @(),
 
+    # Instead of running the post script, create a wrapped file to run it with parameters
+    # so that CI can run it in a subsequent step with a refreshed azure login
+    [Parameter()]
+    [switch] $SelfContainedPostScript,
```
weshaggard, line 99: I suspect we need a better name for this as it is I'm not sure I could figure it out by name. Perhaps something that indicates you are writing the post script to a wrapper and let this be the name of that wrappers script so we also don't have to have the magic variable writing here either as it will be an input and the caller can control it.

benbp: Ah ok, that's a simpler approach. So it's only hardcoded in the deploy yaml basically, and not passed around as a var.

benbp: Actually this won't work, because we still need to call the script by it's path, and the path is not just `service directory + name` but rather `service directory + template file directory + name` which is calculated in the test resources script (https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/TestResources/New-TestResources.ps1#L627).

I'm not sure if there are any places that this would cause trouble with at the moment, but I figured that the implementation should honor template file path.

weshaggard: Why wouldn't we hard-code the full path in the generated script?

eng/common/TestResources/New-TestResources.ps1
```diff
@@ -626,8 +631,36 @@ try {
...
+`$parameters = `@'
+$serialized
+'`@ | ConvertFrom-Json -AsHashtable
+# Set global variables that aren't always passed as parameters
+`$ResourceGroupName = `$parameters.ResourceGroupName
+`$DeploymentOutputs = `$parameters.DeploymentOutputs
+$postDeploymentScript `@parameters
+"@
+                $outScript | Out-File $SelfContainedPostScript
+                LogVsoCommand "##vso[task.setvariable variable=SELF_CONTAINED_TEST_RESOURCES_POST_SCRIPT;]$selfContainedScript"
```
weshaggard, line 659: I assume we don't need this variable any longer.

weshaggard (approved):

benbp:
Finally have this working, took a minute to debug with the slow dev/test in pipelines.

Example: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4689808&view=results

PR #10102, opened 2025-03-21
Update the sdk preview template to include a step that sets node version

ArcturusZhang:


eng/pipelines/templates/stages/archetype-autorest-preview.yml
```diff
@@ -1,4 +1,8 @@
 parameters:
+# Version of Node.js to use.
+- name: NodeVersion
+  type: string
+  default: '20.x'
```
hallipr, line 5:
```suggestion
  default: '20.x'

```

eng/pipelines/templates/stages/archetype-autorest-preview.yml
```diff
@@ -95,6 +99,11 @@ stages:
 
     - checkout: azure-sdk-tools
 
+    - task: NodeTool@0
```
hallipr, line 102:
```suggestion
    - task: NodeTool@1
```

hallipr (approved):

PR #10134, opened 2025-03-26
Use graphql for status lookups in eng/common sync script. Resolve AI review threads.

benbp:
The main point to this change is to support auto-resolving review conversation threads from copilot and/or other AI/bot reviewers when merging eng/common sync PRs. Many of our repos block on review conversations being resolved now, and these AI reviews happen automatically and will block merge.

A side effect of this is review threads can only be resolved via the graphql api. You can look up PR comments via the rest api (/pulls/comments), but the ID given for those reviews is different than the graphql node id for the same ones, so it doesn't work for resolving.

Additionally, the `BLOCKED` status is helpful for determining if we should process PR reviews, and this is also only available via graphql (rest only has `mergeable` and `mergeable_state`). I found that since graphql has more detailed information, I might as well migrate most of the script to graphql. It can also serve as an example if we want to make these pieces more common.

I kept the actual merge call as rest since it's a bit simpler than with graphql.

Other upsides:
- My understanding is graphql has more flexibility with batched calls vs. the rest api, and higher secondary rate limits. It may be helpful for us to use it more to avoid throttling?
- Graphql queries used to be a pain to construct, but chatgpt, etc. are extremely good at building these now.

Downsides:
- graphql calls require auth
- graphql queries are more verbose than their respective rest endpoints (though more fields are available)


scbedd (approved):

scbedd:
I don't see anything jumping out. If this works then it's really a policy discussion on increased usage of Graph.

eng/scripts/Verify-And-Merge-PRs.ps1
```diff
@@ -1,111 +1,250 @@
...
+      throw
+    }
+  }
+}
+
+function RequestGithubGraphQL([string]$query, [object]$variables = @{}) {
+  $body = @{ query = $query; variables = $variables } | ConvertTo-Json -Depth 100
+
+  try {
+    $response = Invoke-RestMethod `
```
weshaggard, line 147: I would suggest switching to using the gh cli if we are touching this. You can use it to call graphql.

weshaggard: Example of doing it https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Delete-RemoteBranches.ps1#L20

eng/scripts/Verify-And-Merge-PRs.ps1
```diff
@@ -1,111 +1,250 @@
...
+
+  # The API will return 200 with a list of error messages for bad queries
+  if ($response.errors) {
+    throw $response.errors.message -join '`n'
+  }
+
+  return $response
+}
+
+function TryResolveAIReviewThreads([string]$repoOwner, [string]$repoName, [string]$prNumber) {
```
weshaggard, line 166: Should we do this here or in our approve helper script?

eng/scripts/Verify-And-Merge-PRs.ps1
benbp: Totally in favor, just wasn't sure if we wanted to stick to the default invoke-restmethod here or not. I'll make the change.

eng/scripts/Verify-And-Merge-PRs.ps1
benbp: Good question. Seems like it does belong in the user approval stage instead. On the other hand, not everyone uses that script, so it could add more friction. If we want to make everyone use it, then I can also update the sync workflow doc to say you have to run the script.

eng/scripts/Verify-And-Merge-PRs.ps1
weshaggard: I've been moving to the gh cli over calling rest directly in PS.

eng/scripts/Verify-And-Merge-PRs.ps1
weshaggard: Yeah I think we probably should start having folks use it and I kind of feel like this belongs there, so things are resolved as the person not the bot.

benbp: Ok I'll move it over.

benbp:
@weshaggard updated - approval script does the conversation resolving, `gh cli` everywhere, and moved some functionality into `git-helpers.ps1`.

Some testing still needed, but local runs of the main flow work.

eng/scripts/Approve-Eng-Common-Sync-PRs.ps1
```diff
@@ -50,14 +55,22 @@ foreach ($repo in $repos)
   }
   
   if ($prstate.reviews.author.login -notcontains $ghloggedInUser) {
```
weshaggard, line 57: I suspect we might need to also add some logic to see if we need to re-approve based on when a new commit is pushed.

eng/scripts/Verify-And-Merge-PRs.ps1
```diff
@@ -1,111 +1,132 @@
...
 )
 
-. "${PSScriptRoot}\..\common\scripts\logging.ps1"
+. "${PSScriptRoot}/../common/scripts/common.ps1"
 
-$ReadyForMerge = $true
-$mergablePRs = @()
-$headers = @{ "Content-Type" = "text/json" }
-$RetryCount = 5
+$ResolveReviewAuthors = @('copilot-pull-request-reviewer')
```
weshaggard, line 11: I don't see this used anywhere.

eng/scripts/Verify-And-Merge-PRs.ps1
```diff
@@ -1,111 +1,132 @@
 param(
-  $PRDataArtifactPath,
-  $AuthToken,
-  $ShouldMerge
+  [string]$PRDataArtifactPath,
+  # @('Azure;azure-sdk-for-net;1', ...)
+  [array]$PRDataInline,
```
weshaggard, line 4: Did you just this just for local testing?

eng/common/scripts/Helpers/git-helpers.ps1
```diff
@@ -247,4 +274,93 @@ function Write-RateLimit {
...
+}
+'@
+
+  $threadIds = GetAIReviewThreads -repoOwner $repoOwner -repoName $repoName -prNumber $prNumber -reviewers $reviewers
+
+  if (!$threadIds) {
+    return $false
+  }
+
+  if ($SkipMerge) {
```
weshaggard, line 354: Did you intend to look at this as a global variable?

eng/scripts/Verify-And-Merge-PRs.ps1
```diff
@@ -1,111 +1,132 @@
...
-    $prUrl = "https://github.com/${repoOwner}/${repoName}/pull/${prNumber}"
-    
-    try
+  MergePRs $mergeable
+}
+
+function ProcessPRMergeStatuses([array]$prData) {
+  for ($retry = 1; $retry -le 5; $retry++)
+  {
+    $curr, $prData = $prData, @()
```
weshaggard, line 44: Is this a fancy way to pop the front of the list? I assume you are only doing it this way because you are modifying the list in the loop.

weshaggard: With that said I don' understand the `,@()` part of this expression.

eng/scripts/Approve-Eng-Common-Sync-PRs.ps1
benbp: Ah, good point on the new requirement. I'll add that in.

eng/scripts/Verify-And-Merge-PRs.ps1
benbp: Yep it's just for local. I can scrap it, but I imagine there will be some hardening so I think it's useful to keep local hooks in for now.

eng/common/scripts/Helpers/git-helpers.ps1
benbp: Another variable lost in the refactor :)

eng/scripts/Verify-And-Merge-PRs.ps1
```diff
@@ -1,111 +1,132 @@
...
-        $ReadyForMerge = $false
-        $unMergablePRs += $prDataLine
-      }
+      $prData += GetOrSetMergeablePR -repoOwner $pr.RepoOwner -repoName $pr.RepoName -prNumber $pr.Number
     }
-    catch {
-      LogError "Invoke-RestMethod ${prApiUrl} failed with exception:`n$_"
-      exit 1
+
+    if ($SkipMerge -or $prData.Retry -notcontains $true) {
```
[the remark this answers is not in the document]
benbp: Yep, missed in refactoring.

eng/scripts/Verify-And-Merge-PRs.ps1
benbp: It's destructuring the list items into variable assignments, so equivalent to:

```
$curr = $prData
$prData = @()
```

weshaggard: I see you are setting prdata to be empty I though you were trying to pull the first item from the prdata array into curr which the name kind of implied.

benbp: I was trying to think of a better var name, any ideas?

weshaggard: Perhaps something like `$currPrData` or `$currPRSet`. I would probably also just split the assignment up and reset $prData on its own line.

weshaggard (approved):

PR #10142, opened 2025-03-27
AI Issue Triage Bot

jeo02:
The underlying infrastructure for the new issue triage.

PR #10169, opened 2025-03-31
Fixes to build files

jeo02:
- Centralized the package versions. 
- Removed duplicate prop file


jeo02:
In response to the following: https://github.com/Azure/azure-sdk-tools/pull/10142#discussion_r2017474933
> Do we need this in the tools repo? I don't recall there being any redirection globally.

I do believe that when you build in the tools repo it creates a artifacts folder at the top level of the tools directory. When I was using the azure functions extension on VS code it did not like this at all, so I had made that little change so that I could deploy/test.

I removed it for now but do you think it should be something we keep in or just put in whenever we need? @jsquire

jsquire:
> In response to the following: [#10142 (comment)](https://github.com/Azure/azure-sdk-tools/pull/10142#discussion_r2017474933)
> 
> > Do we need this in the tools repo? I don't recall there being any redirection globally.
> 
> I do believe that when you build in the tools repo it creates a artifacts folder at the top level of the tools directory. When I was using the azure functions extension on VS code it did not like this at all, so I had made that little change so that I could deploy/test.
> 
> I removed it for now but do you think it should be something we keep in or just put in whenever we need? @jsquire

Better that those who need it use a local override rather than something here.   I'd avoid the extension, personally.     You can add a local `DirectoryBuild.targets` file and then hide with the [Git local exclude](https://hashnode.codeops365.com/a-guide-to-gitinfoexclude-for-local-git-ignore-settings) so that it gets applied to just your environment.

jsquire (approved):

PR #10181, opened 2025-04-01
Update logging functions for github actions

weshaggard:
We need to use Write-Host for the custom logging commands to get them correctly out into stdout which is what they need to be interpreted correctly.


weshaggard:
example usage before change https://github.com/Azure/azure-rest-api-specs/actions/runs/14201937279/job/39790725135

example usages after changes https://github.com/Azure/azure-rest-api-specs/actions/runs/14202683791/job/39793088565

raych1 (approved):

PR #10188, opened 2025-04-02
[Python APIView] renaming default mapping file

swathipil:
- Updating the default mapping path from "apiview_mapping_python.json" to "apiview-properties.json" to align with TCGC mapping file.
- Checking the old default file name if the new default is not found.


packages/python-packages/apiview-stub-generator/tests/apiview_test.py
```diff
@@ -61,6 +61,14 @@ def _add_pyproject_package_to_temp(src_dir):
     assert os.path.exists(pyproject_path)
     os.remove(setup_py_path)
     assert not os.path.exists(setup_py_path)
+
+    # Move the new mapping file to the old mapping file name
+    mapping_file_path = os.path.join(dest_dir, "apiview-properties.json")
```
iscai-msft, line 66: i think it would be better to move the old mapping to the new mapping

iscai-msft: i guess this is a test so maybe it doesn't matter that much

iscai-msft (approved):

packages/python-packages/apiview-stub-generator/tests/apiview_test.py
swathipil: Yeah, either works. My thought was that if, eventually, we can get rid of the old mapping name, it'll be easy to remove this "old mapping file" bit from here. I've updated the default testing package "apistubgentest" so that it has apiview-properties.json, since I'm assuming that'll be the most widely used.

tjprescott (approved):

PR #10190, opened 2025-04-02
Add additional logging to `test-proxy` startup in `CI` actions

### LINKED ISSUE #10166: Log some info in test proxy Azure Pipelines tasks
The download/start tasks for test proxy don't seem to log any details about what it's doing. I recently ran into an issue where I suspect a warning I saw - the proxy couldn't be started and the task will retry - might be the cause. It wasn't (it was a subtle code change with a cascading effect), but I couldn't tell from the logs.

What might be nice is just a few details, like the URL being downloaded for the download task and the command line being invoked for each try in the start task, so that if a try fails we see when it's trying again. As well as the exit code. Maybe use the `Invoke-LogCommand` helper or something like that.

scbedd:
FYI @heaths 

Resolves #10166

### DIFF
--- a/eng/common/testproxy/test-proxy-standalone-tool.yml
+++ b/eng/common/testproxy/test-proxy-standalone-tool.yml
@@ -32,23 +32,33 @@ steps:
       }
 
       Write-Host "Installing test-proxy version $version"
+      Write-Host "${{ parameters.templateRoot }}/eng/common/testproxy/install-test-proxy.ps1 -Version $version -InstallDirectory $(Build.BinariesDirectory)/test-proxy"
       ${{ parameters.templateRoot }}/eng/common/testproxy/install-test-proxy.ps1 -Version $version -InstallDirectory $(Build.BinariesDirectory)/test-proxy
     displayName: "Install test-proxy"
     condition: and(succeeded(), ${{ parameters.condition }})
 
   - pwsh: |
+      Write-Host "Prepending path with the test proxy tool install location: '$(Build.BinariesDirectory)/test-proxy'"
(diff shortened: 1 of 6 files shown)


heaths (approved):
I wasn't expecting that much. ;) @hallipr is the `Invoke-ToolCommand` or whatever you use not common? Could've saved @scbedd a bunch of typing and reduce duplication errors.

eng/common/testproxy/test-proxy-standalone-tool.yml
```diff
@@ -71,7 +88,9 @@ steps:
     - pwsh: |
         for ($i = 0; $i -lt 10; $i++) {
             try {
+                Write-Host "Invoke-WebRequest -Uri "http://localhost:5000/Admin/IsAlive" | Out-Null"
```
hallipr, line 91:
```suggestion
                Write-Host "Invoke-WebRequest -Uri `"http://localhost:5000/Admin/IsAlive`" | Out-Null"
```

hallipr (approved):

scbedd:
> I wasn't expecting that much. ;) @hallipr is the `Invoke-ToolCommand` or whatever you use not common? Could've saved @scbedd a bunch of typing and reduce duplication errors.

It's just because they are separate steps so yes I could use it but would be a hilarious:

```
. common.ps1

Invoke-LoggedCommand blah
```

And for one of them I actually _do_ need the returned Process which `Invoke-LoggedCommand` doesn't do. So I just kept it simple.

hallipr (approved):

hallipr (approved):

PR #10193, opened 2025-04-02
[tsp-client] Unblock compiler 1.0 support

catalinaperalta:
Unblocking support for `@typespec/compiler` v1.0. Still supporting 0.66.0 and later for a bit longer to give time for folks to move forward to the stable compiler version.

### DIFF
--- a/tools/tsp-client/package.json
+++ b/tools/tsp-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@azure-tools/typespec-client-generator-cli",
-  "version": "0.16.0",
+  "version": "0.17.0",
   "description": "A tool to generate Azure SDKs from TypeSpec",
   "main": "dist/index.js",
   "homepage": "https://github.com/Azure/azure-sdk-tools/tree/main/tools/tsp-client#readme",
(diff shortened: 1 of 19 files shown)


tools/tsp-client/package.json
```diff
@@ -40,7 +40,7 @@
     "@types/chai": "^4.3.5",
     "@types/node": "^20.4.8",
     "@types/prompt-sync": "^4.2.1",
-    "@typespec/compiler": "^0.58.0",
+    "@typespec/compiler": "^0.67.0",
```
allenjzhang, line 43: This is incorrect. Should use `^1.0.0-0`

tools/tsp-client/package.json
```diff
@@ -65,6 +65,6 @@
     "yargs": "^17.2.1"
   },
   "peerDependencies": {
-    "@typespec/compiler": ">=0.58.0 <1.0.0"
+    "@typespec/compiler": ">=0.66.0"
```
allenjzhang, line 68: Same as above

allenjzhang (changes requested):

allenjzhang (approved):

PR #10224, opened 2025-04-05
initial evals for apiview-copilot

kristapratico:
TODOs

- ~clean-up~
- ~retrieve guidelines from db for adding the context to test cases (will be done here if the RAG PR goes in first)~ just using static files under guidelines/ for now
- ~add docs~


packages/python-packages/apiview-copilot/guidelines/python/guidelines.json
```diff
@@ -225,154 +225,93 @@
...
+      "category": "Types (or not)",
+      "text": "DO derive from the abstract collections base classes collections.abc (or collections for Python 2.7) to provide custom mapping types."
     },
     {
-        "id": "python_implementation.html#python-codestyle-pep484",
-        "category": "Types (or not)",
-        "text": "DO provide type hints PEP484 (https://www.python.org/dev/peps/pep-0484/) for publicly documented classes and functions"
-    },
-    {
-        "id": "python_implementation.html#python-codestyle-vars-naming",
```
tjprescott, line 318: Did you change the guidelines? If not, what's on main should be more up-to-date.

packages/python-packages/apiview-copilot/evals/deconstruct_testcase.py
```diff
@@ -0,0 +1,53 @@
+import argparse
+import json
+import pathlib
+
+
+if __name__ == "__main__":
```
tjprescott, line 6: Seems like these could easily be plugged into the `cli.py` script at the root of the workspace folder. under an `eval` group.

tjprescott (commented):
Overall LGTM. A couple comments.

packages/python-packages/apiview-copilot/guidelines/python/guidelines.json
kristapratico: Discussed offline, will leave this here as representation of what was used to create baseline and the RAG PR will overwrite it

packages/python-packages/apiview-copilot/evals/deconstruct_testcase.py
kristapratico: I was able to plug in the create / deconstruct test case into the CLI, but the evals script will still need to be run by `python evals/run.py` for now (there is some issue with the evals library needing to be called from __main__)

tjprescott (approved):

PR #10230, opened 2025-04-07
[tsp-client] Fix additional directory implementation

catalinaperalta:
Given the recent changes in typespec validation for the tspconfig.yaml file, move configuration options specific to tsp-client under an entry in the `options` property of tspconfig.yaml. This change affects the way additionalDirectories will be specified.

### DIFF
--- a/tools/tsp-client/CHANGELOG.md
+++ b/tools/tsp-client/CHANGELOG.md
@@ -8,8 +8,7 @@
 options:
   "@azure-tools/typespec-client-generator-cli":
     "additionalDirectories":
-      - "specification/ai/HealthInsights/HealthInsights.Common/"
-      - "specification/ai/HealthInsights/HealthInsights.OpenAPI/"
+      - "specification/contosowidgetmanager/Contoso.WidgetManager.Shared/"
 ```
 
 ## 2025-04-03 0.17.0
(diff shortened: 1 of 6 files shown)


tools/tsp-client/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Release
 
+## 2025-04-07 - 0.18.0
+
+- Specify tsp-client specific configurations under a `@azure-tools/typespec-client-generator-cli` entry in tspconfig.yaml options. This affects the `additionalDirectories` configuration. Example entry in tspconfig.yaml:
+
+```yaml
+options:
+  "@azure-tools/typespec-client-generator-cli":
```
catalinaperalta, line 9: For ease of use I would prefer to have this be:
```suggestion
  "tsp-client":
```

timotheeguerin: it would be cleaner but also inconsitent with the other options, at least here we know who is using that info

timotheeguerin (approved):

tools/tsp-client/CHANGELOG.md
catalinaperalta: Ok, I'll leave as is for now and see if there's any push back

PR #10236, opened 2025-04-08
EmitterPackageJsonPath fixes

JoshLove-msft:
- Use repo root for emitter-package path and match package-lock naming with specified emitter path
- include azure-typespec scope in emitter list
- Update generate-config/generate-lockfile commands

Contributes to https://github.com/Azure/azure-sdk-for-net/issues/49240

### DIFF
--- a/tools/tsp-client/src/commands.ts
+++ b/tools/tsp-client/src/commands.ts
@@ -41,7 +41,7 @@ export async function initCommand(argv: any) {
 
   const emitterPackageOverride = resolveEmitterPathFromArgs(argv);
 
-  const emitter = await getEmitterFromRepoConfig(emitterPackageOverride ?? defaultRelativeEmitterPackageJsonPath);
+  const emitter = await getEmitterFromRepoConfig(emitterPackageOverride ?? joinPaths(repoRoot, defaultRelativeEmitterPackageJsonPath));
   if (!emitter) {
     throw new Error("Couldn't find emitter-package.json in the repo");
   }
@@ -495,7 +495,7 @@ export async function generateConfigFilesCommand(argv: any) {
 export async function generateLockFileCommand(argv: any) {
   await generateLockFileCommandCore(
     argv["output-dir"],
-    resolveEmitterPathFromArgs(argv) ?? joinPaths(await getRepoRoot(argv["output-dir"]), "eng", "emitter-package.json"));
+    resolveEmitterPathFromArgs(argv) ?? joinPaths(await getRepoRoot(argv["output-dir"]), defaultRelativeEmitterPackageJsonPath));
 }
 
 export async function generateLockFileCommandCore(outputDir: string, emitterPackageJsonPath: string) {
--- a/tools/tsp-client/src/commands.ts
+++ b/tools/tsp-client/src/commands.ts
@@ -481,7 +481,7 @@ export async function generateConfigFilesCommand(argv: any) {
     emitterPackageJson["overrides"] = overrideJson;
   }
   
-  const emitterPath = resolveEmitterPathFromArgs(argv) ?? defaultRelativeEmitterPackageJsonPath;
+  const emitterPath = resolveEmitterPathFromArgs(argv) ?? joinPaths(repoRoot, defaultRelativeEmitterPackageJsonPath);
   
   await writeFile(
     emitterPath,
--- a/tools/tsp-client/src/commands.ts
+++ b/tools/tsp-client/src/commands.ts
@@ -24,7 +24,7 @@ import {
 } from "./utils.js";
 import { parse as parseYaml } from "yaml";
 import { config as dotenvConfig } from "dotenv";
-import { dirname, resolve } from "node:path";
+import path, { dirname, resolve } from "node:path";
 import { doesFileExist } from "./network.js";
 import { sortOpenAPIDocument } from "@azure-tools/typespec-autorest";
 
@@ -37,8 +37,9 @@ export async function initCommand(argv: any) {
 
   const repoRoot = await getRepoRoot(outputDir);
 
-  const emitterPackageJsonPath =
-    argv["emitter-package-json-path"] ?? joinPaths(repoRoot, "eng", "emitter-package.json");
+  const emitterPackageJsonPath = getEmitterPackageJsonPath(
+    repoRoot,
+    argv["emitter-package-json-path"]);
 
   const emitter = await getEmitterFromRepoConfig(emitterPackageJsonPath);
   if (!emitter) {
@@ -169,8 +170,9 @@ export async function syncCommand(argv: any) {
     throw new Error("Could not find repo root");
   }
   const tspLocation: TspLocation = await readTspLocation(outputDir);
-  const emitterPackageJsonPath =
-    tspLocation.emitterPackageJsonPath ?? joinPaths(repoRoot, "eng", "emitter-package.json");
+  const emitterPackageJsonPath = getEmitterPackageJsonPath(
+    repoRoot,
+    tspLocation.emitterPackageJsonPath);
   const dirSplit = tspLocation.directory.split("/");
   let projectName = dirSplit[dirSplit.length - 1];
   Logger.debug(`Using project name: ${projectName}`);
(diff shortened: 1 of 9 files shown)


tools/tsp-client/src/commands.ts
```diff
@@ -478,36 +477,43 @@ export async function generateConfigFilesCommand(argv: any) {
   if (Object.keys(overrideJson).length > 0) {
     emitterPackageJson["overrides"] = overrideJson;
   }
+  const relativeEmitterPath = argv["emitter-package-json-path"] ?? joinPaths("eng", "emitter-package.json");
+  const absoluteEmitterPath = joinPaths(repoRoot, relativeEmitterPath);
```
hallipr, line 487: I wouldn't assume that `--emitter-package-json-path` is relative to the repo root.

```suggestion
  const relativeEmitterPath = argv["emitter-package-json-path"];
  const defaultEmitterPath =  joinPaths(repoRoot, "eng", "emitter-package.json");
  const absoluteEmitterPath = relativeEmitterPath ? resolve(relativeEmitterPath) : defaultEmitterPath;
```

tools/tsp-client/src/commands.ts
```diff
@@ -242,8 +242,7 @@ export async function syncCommand(argv: any) {
     if (tspLocation.emitterPackageJsonPath) {
 
       // If the emitter package json path is provided, we need to check for the lock file in the same directory with the same name prefix.
-      const emitterPackageJsonFileName = path.basename(tspLocation.emitterPackageJsonPath, path.extname(tspLocation.emitterPackageJsonPath));
-      emitterLockPath = joinPaths(dirname(emitterPackageJsonPath), `${emitterPackageJsonFileName}-lock.json`);
+      emitterLockPath = getEmitterLockPath(emitterPackageJsonPath);
```
hallipr, line 251: Why not just always do this?   won't we always use a lock file name derived from the emitterPackageJson path, even in the default case?

hallipr: The special casing for `if (tspLocation.emitterPackageJsonPath) {` doesn't seem worth the cost of always computing the name.

hallipr: seems like I worded that oddly:)   Not special casing seems worth the cost of a string compute, even in the default case.

hallipr (approved):

tools/tsp-client/src/commands.ts
```diff
@@ -24,10 +24,12 @@ import {
 } from "./utils.js";
 import { parse as parseYaml } from "yaml";
 import { config as dotenvConfig } from "dotenv";
-import { dirname, resolve } from "node:path";
+import path, { basename, dirname, resolve } from "node:path";
```
catalinaperalta, line 27: Nit: looks like we just needed to import relative here?

tools/tsp-client/src/commands.ts
```diff
@@ -567,3 +570,23 @@ export async function sortSwaggerCommand(argv: any): Promise<void> {
...
 }
+
+
+function getEmitterPackageJsonPath(repoRoot: string, tspLocation: TspLocation): string {
+  const relativePath = tspLocation.emitterPackageJsonPath ?? defaultRelativeEmitterPackageJsonPath;
+  return joinPaths(repoRoot, relativePath);
+}
+
+function getEmitterLockPath(emitterPackageJsonPath: string): string {
+  const emitterPackageJsonFileName = path.basename(emitterPackageJsonPath, path.extname(emitterPackageJsonPath));
```
catalinaperalta, line 581: Why are we including the extension here?

tools/tsp-client/src/commands.ts
```diff
@@ -485,36 +480,44 @@ export async function generateConfigFilesCommand(argv: any) {
   if (Object.keys(overrideJson).length > 0) {
     emitterPackageJson["overrides"] = overrideJson;
   }
+  
+  const emitterPath = resolveEmitterPathFromArgs(argv) ?? joinPaths(repoRoot, defaultRelativeEmitterPackageJsonPath);
```
catalinaperalta, line 484: This command doesnt support the `emitter-package-json-path` arg it supports a `package-json` path which is meant to be used with the package.json file of the target emitter to pull deps from there. It also is meant to create the default config files (emitter-package.json and emitter-package-lock.json) that tsp-client needs to be able to run commands generally. We dont need to change this. At most we could enable an option for you to specify an alternate name to save these files to.

tools/tsp-client/src/commands.ts
```diff
@@ -485,36 +480,44 @@ export async function generateConfigFilesCommand(argv: any) {
...
+  Logger.info(`${basename(emitterPath)} file generated in '${dirname(emitterPath)}' directory`);
 
-  await generateLockFileCommand(argv);
+  await generateLockFileCommandCore(outputDir, emitterPath);
 }
 
 export async function generateLockFileCommand(argv: any) {
-  const outputDir = argv["output-dir"];
-  const repoRoot = await getRepoRoot(outputDir);
+  await generateLockFileCommandCore(
```
catalinaperalta, line 496: Same comment as above this command doesnt support the flag you're trying to use...we could enable it if you need this facility.

tools/tsp-client/src/commands.ts
JoshLove-msft: Yeah I think we need the same support for specifying the emitter package path in the generate config commands. This is so that we are able to use tsp-client to regen these files after we ship the new emitter.

tools/tsp-client/src/commands.ts
JoshLove-msft: Yes, we will need it. I can update the docs if that works for you?

catalinaperalta: You need to add this flag under the command in the index.ts file and yes please update the docs as well

tools/tsp-client/src/commands.ts
catalinaperalta: So then you should revert these changes and add a flag to allow you to specify an alternate name for the files

tools/tsp-client/src/commands.ts
JoshLove-msft: This computes the file name from the file path + ext.

tools/tsp-client/src/commands.ts
catalinaperalta: How long are you planning on hosting these 2 sets of emitter-package files together? (the default file and the alternate emitter config file)

catalinaperalta: > So then you should revert these changes and add a flag to allow you to specify an alternate name for the files

In this command you'd only need to update the file name when it's being written. And yes please update the docs on this one as well

tools/tsp-client/src/commands.ts
catalinaperalta: I thought basename with the emitterPackageJsonPath would work without you needing to specify the extension? Is that not so?

tools/tsp-client/src/commands.ts
JoshLove-msft: I also updated the name of the file that is logged.

tools/tsp-client/src/commands.ts
JoshLove-msft: I believe it includes the extension.

tools/tsp-client/src/commands.ts
```diff
@@ -567,3 +570,23 @@ export async function sortSwaggerCommand(argv: any): Promise<void> {
...
+}
+
+function resolveEmitterPathFromArgs(argv: any): string | undefined {
+  const emitterPath = argv["emitter-package-json-path"];
+  if (emitterPath) {
+    return resolve(emitterPath);
+  }
+
+  return undefined;
+}
```
catalinaperalta, line 592: Nit: newline

tools/tsp-client/test/commands.spec.ts
```diff
@@ -48,6 +53,16 @@ describe.sequential("Verify commands", () => {
     }
   });
 
+  it("Generate lock file with altername package path", async () => {
+    try {
+      await generateLockFileCommand({"emitter-package-json-path": joinPaths(repoRoot, "tools/tsp-client/test/utils/alternate-emitter-package.json")});
+
+      assert.isTrue((await stat(joinPaths(repoRoot, "tools/tsp-client/test/utils/alternate-emitter-package.json"))).isFile());
```
catalinaperalta, line 60: Should we delete the lock file you have checked in under utils at the beginning of this test, then in this assert we should check for its existence?

tools/tsp-client/test/commands.spec.ts
```diff
@@ -286,6 +324,27 @@ describe.sequential("Verify commands", () => {
     }
   }, 360000);
 
+  it("Generate config files with alternate json path", async () => {
```
catalinaperalta, line 327: Same for this test, should we try deleting those files and then asserting that they exist?

catalinaperalta (commented):
looking good, just a couple of comments about the tests

catalinaperalta (approved):

PR #10238, opened 2025-04-08
Stop-gap for `go` `eng/` packages

scbedd:
This PR addresses the issue that we have over in `go` with the fact that we aren't resolving the appropriate `ci.yml` for the eng/tools directory packages.

It is NOT the correct solution even in medium term. We need to adjust this to do some smarter algo that climbs up from the package directory, rather than assuming `sdk` or `eng` (which this PR does).

I want to unblock `go`, but I due to IRL commitments I can't get to that correct solution today.

I ran a full Save-Package-Properties on every repo. Clean runs. Can't see any regressions, and we are passing our integration tests.

### DIFF
--- a/eng/common/scripts/Package-Properties.ps1
+++ b/eng/common/scripts/Package-Properties.ps1
@@ -115,11 +115,28 @@ class PackageProps {
         return $null
     }
 
-    [PSCustomObject]GetCIYmlForArtifact() {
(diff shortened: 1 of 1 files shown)


hallipr (approved):

scbedd:
I accidentally pushed merge main to this branch after all the common prs went in. Just going to bypass and merge.

PR #10241, opened 2025-04-08
Use tsp-client in emitter CI

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/49240

### DIFF
--- a/eng/common/pipelines/templates/archetype-typespec-emitter.yml
+++ b/eng/common/pipelines/templates/archetype-typespec-emitter.yml
@@ -67,6 +67,10 @@ parameters:
   type: object
   default: []
 
+- name: EmitterPackageJsonPath
+  type: string
+  default: ""
+
 extends:
   template: /eng/pipelines/templates/stages/1es-redirect.yml
   parameters:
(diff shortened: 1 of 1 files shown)


eng/common/pipelines/templates/archetype-typespec-emitter.yml
```diff
@@ -189,26 +193,19 @@ extends:
...
-              workingDirectory: $(Build.SourcesDirectory)
+          - pwsh: |
+              tsp-client generate-config-files '$(buildArtifactsPath)/lock-files/package.json' --output-dir '$(Build.ArtifactStagingDirectory)' 
+              --emitter-package-json-path  ${{ parameters.EmitterPackageJsonPath }}
+              --overrides '$(buildArtifactsPath)/overrides.json'
+            displayName: Generate emitter-package.json file
+
+          - pwsh: |
+              tsp-client generate-lock-file --output-dir '$(Build.ArtifactStagingDirectory)' --emitter-package-json-path  ${{ parameters.EmitterPackageJsonPath }}
+            displayName: Generate emitter-package-lock.json file
```
hallipr, line 208: no need for this call now.   `tsp-client generate-config-files` makes an implicit, chained call to `generate-lock-file`

hallipr (approved):

eng/common/pipelines/templates/archetype-typespec-emitter.yml
```diff
@@ -189,26 +193,25 @@ extends:
...
-            displayName: Create emitter-package-lock.json
-            inputs:
-              pwsh: true
-              filePath: ./eng/common/scripts/typespec/New-EmitterPackageLock.ps1
-              arguments: >
-                -EmitterPackageJsonPath '$(Build.ArtifactStagingDirectory)/emitter-package.json'
-                -OutputDirectory '$(Build.ArtifactStagingDirectory)'
-              workingDirectory: $(Build.SourcesDirectory)
+          - pwsh: |
+              if (Test-Path -Path '$(buildArtifactsPath)/overrides.json') {
```
weshaggard, line 201: What is the scenario where we need to override the emitter-package.json? 

cc @m-nash

weshaggard: I also chatted with @m-nash about this and I can understand using this for cases where we want to transition or use a completely different emitter but I want to make sure it is only temporary so that we don't get into a place where service directories are all using different versions.

JoshLove-msft: These aren't going to be defined at the service level. There will be 1 emitter-package.json for the old emitter and another one for the new emitter. We do eventually want to support service level customization of dependencies, but this would likely use tsp-location.yaml.

eng/common/pipelines/templates/archetype-typespec-emitter.yml
```diff
@@ -189,26 +193,25 @@ extends:
...
+                --output-dir '$(Build.ArtifactStagingDirectory)' 
+                --emitter-package-json-path  '${{ parameters.EmitterPackageJsonPath }}'
+                --overrides '$(buildArtifactsPath)/overrides.json'
+              } else {
+                tsp-client generate-config-files 
+                --package-json '$(buildArtifactsPath)/lock-files/package.json' 
+                --output-dir '$(Build.ArtifactStagingDirectory)' 
+                --emitter-package-json-path  '${{ parameters.EmitterPackageJsonPath }}'
+              }
+            displayName: Generate emitter-package.json and emitter-package-lock files
```
catalinaperalta, line 214: I'm not seeing the line where the workingDirectory is set to the repo clone..the tsp-client command will fail if we're not doing that @JoshLove-msft

PR #10252, opened 2025-04-09
Add env vars, vcpkg packages to inventory pipeline

heaths:


eng/pipelines/agent-software-inventory.yml
```diff
@@ -127,15 +127,34 @@ jobs:
...
+    displayName: (Windows) Program Files (x86)
+    workingDirectory: $(Build.ArtifactStagingDirectory)
+    condition: and(succeeded(), eq(variables.Host, 'Windows'))
+
+  - pwsh: |
+      Get-ChildItem env: | ForEach-Object {
+        "$(Host), $(OSVmImage), Environment, $($_.Name), $($_.Value), all"
+      } >> env-inventory.csv
+    displayName: Environment Variables
+    workingDirectory: $(Build.ArtifactStagingDirectory)
```
benbp, line 147: Isn't there a possibility of secrets leaking into the pipeline artifacts here?

heaths: I didn't think that was possible if they weren't added to the `env:` block explicitly.

benbp: I just always assume the shell env is likely to have some extra stuff placed in there. But if devops is keeping it sanitary, then good.

heaths: I commented it out just to be safe. Don't need to run this very often anyway.

benbp (approved):

PR #10256, opened 2025-04-09
[tsp-client] Enable tracing during compile

catalinaperalta:


tools/tsp-client/src/typespec.ts
```diff
@@ -112,6 +114,9 @@ export async function compileTsp({
     emit: [emitterPackage],
     options: overrideOptions,
   };
+  if (trace) {
```
timotheeguerin, line 117: Should we not just match and do a pas through

catalinaperalta: You mean have the option be `--trace [<emitter name]` ??

catalinaperalta: We only allow folks to emit with one emitter at a time through this tool so I thought `--trace` is cleaner for folks to pass in, thoughts?

timotheeguerin (approved):

tools/tsp-client/src/typespec.ts
catalinaperalta: If we wanted to enable tracing the compiler what does that config look like? Something like this:
`["@typespec/compiler", "<emitter name>"]`

JoshLove-msft (approved):

tools/tsp-client/src/typespec.ts
timotheeguerin: Compiler is not scoped so you’d have to include everything

timotheeguerin: https://typespec.io/docs/handbook/configuration/tracing/

catalinaperalta: Thanks for the info! Updated the code to match what the tsp cli does :)

JoshLove-msft (approved):

PR #10257, opened 2025-04-09
Fixes to generate-config step

JoshLove-msft:
Testing in https://github.com/Azure/azure-sdk-for-net/pull/49321

### DIFF
--- a/eng/common/pipelines/templates/archetype-typespec-emitter.yml
+++ b/eng/common/pipelines/templates/archetype-typespec-emitter.yml
@@ -204,13 +204,13 @@ extends:
                 --package-json '$(buildArtifactsPath)/lock-files/package.json' `
                 --output-dir '$(Build.SourcesDirectory)' `
                 --emitter-package-json-path "$(Build.SourcesDirectory)/${{ parameters.EmitterPackageJsonPath }}" `
-                --overrides '$(buildArtifactsPath)/overrides.json'
+                --overrides '$(buildArtifactsPath)/overrides.json' `
                 --version '$(emitterVersion)'
               } else {
                 tsp-client generate-config-files `
                 --package-json '$(buildArtifactsPath)/lock-files/package.json' `
                 --output-dir '$(Build.SourcesDirectory)' `
-                --emitter-package-json-path "$(Build.SourcesDirectory)/${{ parameters.EmitterPackageJsonPath }}"
+                --emitter-package-json-path "$(Build.SourcesDirectory)/${{ parameters.EmitterPackageJsonPath }}" `
                 --version '$(emitterVersion)'
               }
             displayName: Generate emitter-package.json and emitter-package-lock files
--- a/eng/common/pipelines/templates/archetype-typespec-emitter.yml
+++ b/eng/common/pipelines/templates/archetype-typespec-emitter.yml
@@ -203,18 +203,18 @@ extends:
                 tsp-client generate-config-files `
                 --package-json '$(buildArtifactsPath)/lock-files/package.json' `
                 --output-dir '$(Build.SourcesDirectory)' `
-                --emitter-package-json-path "$(Build.SourcesDirectory)/${{ parameters.EmitterPackageJsonPath }}" `
+                --emitter-package-json-path '${{ parameters.EmitterPackageJsonPath }}' `
                 --overrides '$(buildArtifactsPath)/overrides.json' `
                 --version '$(emitterVersion)'
               } else {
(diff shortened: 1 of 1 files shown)


eng/common/pipelines/templates/archetype-typespec-emitter.yml
```diff
@@ -202,16 +202,19 @@ extends:
                 Write-Host "Using overrides.json to generate emitter-package.json"
                 tsp-client generate-config-files `
                 --package-json '$(buildArtifactsPath)/lock-files/package.json' `
-                --output-dir '$(Build.ArtifactStagingDirectory)' `
-                --emitter-package-json-path  '${{ parameters.EmitterPackageJsonPath }}' `
-                --overrides '$(buildArtifactsPath)/overrides.json'
+                --output-dir '$(Build.SourcesDirectory)' `
+                --emitter-package-json-path "$(Build.SourcesDirectory)/${{ parameters.EmitterPackageJsonPath }}" `
+                --overrides '$(buildArtifactsPath)/overrides.json' `
+                --version '$(emitterVersion)'
```
hallipr, line 208: `--version` isn't a parameter to `generate-config-files`. It should be extracted from `--package-json`

hallipr, line 208:
```suggestion
                --overrides '$(buildArtifactsPath)/overrides.json' 
```

`--version` isn't a parameter to `generate-config-files`. It should be extracted from `--package-json`

eng/common/pipelines/templates/archetype-typespec-emitter.yml
```diff
@@ -202,16 +202,19 @@ extends:
...
               } else {
                 tsp-client generate-config-files `
                 --package-json '$(buildArtifactsPath)/lock-files/package.json' `
-                --output-dir '$(Build.ArtifactStagingDirectory)' `
-                --emitter-package-json-path  '${{ parameters.EmitterPackageJsonPath }}'
+                --output-dir '$(Build.SourcesDirectory)' `
+                --emitter-package-json-path "$(Build.SourcesDirectory)/${{ parameters.EmitterPackageJsonPath }}" `
+                --version '$(emitterVersion)'
```
hallipr, line 214:
```suggestion
                --emitter-package-json-path "$(Build.SourcesDirectory)/${{ parameters.EmitterPackageJsonPath }}"
```

eng/common/pipelines/templates/archetype-typespec-emitter.yml
```diff
@@ -202,16 +202,19 @@ extends:
                 Write-Host "Using overrides.json to generate emitter-package.json"
                 tsp-client generate-config-files `
                 --package-json '$(buildArtifactsPath)/lock-files/package.json' `
-                --output-dir '$(Build.ArtifactStagingDirectory)' `
-                --emitter-package-json-path  '${{ parameters.EmitterPackageJsonPath }}' `
-                --overrides '$(buildArtifactsPath)/overrides.json'
+                --output-dir '$(Build.SourcesDirectory)' `
+                --emitter-package-json-path "$(Build.SourcesDirectory)/${{ parameters.EmitterPackageJsonPath }}" `
```
hallipr, line 206: I would require parameters.EmitterPackageJsonPath be a rooted path, or have `tsp-client` resolve it using working directory.   joining the string here will always pass  `--emitter-package-json-path "$(Build.SourcesDirectory)"` when the parameter is empty.   Best to default it at the top to `$(Build.SourcesDirectory)/eng/emitter-package.json` and let users override it as needed.

eng/common/pipelines/templates/archetype-typespec-emitter.yml
JoshLove-msft: This is a parameter in generate-config-files. The issue I was hitting is that the version extracted from package.json was the base version and didn't include the suffix.

JoshLove-msft: It is set for this job and here is the successful pipeline run - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4745993&view=results

JoshLove-msft: I guess $(emitterVersion) is essentially a global available to all steps?

weshaggard: I would double check this as I don't see where this value is set in this context.

PR #10264, opened 2025-04-10
[gh bot] Restructure labeling

jeo02:
- Separated the Configuration to avoid circular dependency
- Separated out the labeler, so far only the legacy labeler but can easily add in the rag labeler and other labelers.
- Added common classes to existing shared folder.
- Importing directory.Build.local.props after the standard props did not work even though logically that makes sense?? I changed it back to doing local before the standard props.


tools/issue-labeler/src/LabelerFactory/LegacyLabeler.cs
```diff
@@ -0,0 +1,111 @@
...
+            var predictionRepositoryName = TranslateRepoName(issue.RepositoryName);
+
+            // If the model needed for this request hasn't been initialized, do so now.
+            if (!InitializedRepositories.ContainsKey(predictionRepositoryName))
+            {
+                _logger.LogInformation($"Models for {predictionRepositoryName} have not yet been initialized; loading prediction models.");
+
+                try
+                {
+                    var allBlobConfigNames = _config.GetItem($"IssueModel.{predictionRepositoryName.Replace("-", "_")}.BlobConfigNames").Split(';', StringSplitOptions.RemoveEmptyEntries);
```
jeo02, line 45: Forgot to change this will do!

tools/issue-labeler/Directory.Build.props
```diff
@@ -26,11 +26,11 @@
...
   </PropertyGroup>
-  
-     <!-- Apply the standard props first, so that any local props can override -->
-    <Import Project="../../Directory.Build.props" />
-  
+
     <!-- Apply the local props after, which allows them to override.-->
     <Import Project="Directory.Build.local.props" Condition="Exists('Directory.Build.local.props')" />
 
+     <!-- Apply the standard props first, so that any local props can override -->
```
jsquire, line 33: I'd flip the order here.

tools/issue-labeler/src/IssueLabeler.Shared/IssueOutput.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace IssueLabeler.Shared
+{
+    public class IssueOutput
```
jsquire, line 9:
```suggestion
    public class TriageOutput
```

tools/issue-labeler/src/IssueLabelerService/AzureSdkIssueLabelerService.cs
```diff
@@ -22,91 +24,118 @@ public class AzureSdkIssueLabelerService
...
                 return new JsonResult(result);
             }
             catch (Exception ex)
             {
-                _logger.LogError($"Unable to deserialize output: {ex.Message}\n\t{ex}");
+                _logger.LogError($"Error processing issue #{issue.IssueNumber} in repository {issue.RepositoryName}: {ex.Message}{Environment.NewLine}\t{ex}{Environment.NewLine}");
                 return EmptyResult;
             }
         }
+        //{
```
jsquire, line 81: Is there a plan for this or should we trim out the comments?

jeo02: I'll remove it

tools/issue-labeler/Directory.Build.props
jeo02: I don't know why but when I flip the order it won't build.

tools/issue-labeler/src/IssueLabelerService/Configuration.cs
```diff
@@ -1,6 +1,6 @@
 using Microsoft.Extensions.Configuration;
 
-namespace ConfigurationService
+namespace IssueLabelerService
```
jsquire, line 3: Can we add folders for organization.   It would be nice to have:
   - Configuration
   - Labels
   - Answers

Feel free to tune the text, but I'm thinking the discrete areas that we'd want to modularize should each have a folder.

jsquire (approved):

PR #10292, opened 2025-04-11
[gh bot] Restructuring Ai Answering

jeo02:
- Separated the answering from the function similarly to the labeling.
- Put some additional hard coded variables into the app configuration
- Dynamic prompting. We can add variables into our prompts with the function `FormatTemplate`


tools/issue-labeler/src/IssueLabeler.Shared/QnaOutput.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace IssueLabeler.Shared
+{
+    public class QnaOutput
```
jsquire, line 9: nitpicky, but I'm not a fan of "Qna".    Maybe  "ResolutionOutput" or just "AnswerOutput" to align with what we're calling the fields?

tools/issue-labeler/src/IssueLabelerService/Answers/RagQna.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Threading.Tasks;
+using IssueLabeler.Shared;
+using Newtonsoft.Json;
+using Microsoft.Extensions.Logging;
+using System.Linq;
+using System.Collections.Generic;
+
+namespace IssueLabelerService
+{
+    public class RagQna : IQnaService
```
jsquire, line 11: Should we start to call this `OpenAiAnswerService` or similar?   That way, if we add some kind of "Azure AI Foundry" AI answer service in the future, we have better differentiation?    We'd have to reflect labeling too

tools/issue-labeler/src/IssueLabelerService/Answers/RagQna.cs
```diff
@@ -0,0 +1,146 @@
...
+        public RagQna(ILogger<QnaFactory> logger, RepositoryConfiguration config, TriageRag ragService)
+        {
+            _config = config;
+            _ragService = ragService;
+            _logger = logger;
+        }
+        public async Task<QnaOutput> AnswerQuery(IssuePayload issue)
+        {
+            // Configuration for Azure services
+            var modelName = _config.OpenAIModelName;
```
jsquire, line 25: Config integration looks really nice.   Clean.  Readable.

tools/issue-labeler/src/IssueLabelerService/Answers/RagQna.cs
```diff
@@ -0,0 +1,146 @@
...
+                    rd.Item1.chunk,
+                    rd.Item1.Service,
+                    rd.Item1.Category,
+                    rd.Item1.Url,
+                    Score = rd.Item2
+                })
+                .ToList();
+
+            // Filtered out all sources for either one then not enough information to answer the issue. 
+            if (docs.Count() == 0 || issues.Count() == 0)
```
jsquire, line 71:
```suggestion
            if (docs.Count == 0 || issues.Count == 0)
```

For a list, `Count` is a property that just reads a backing field.   `Count()` is a LINQ extension method that you should always expect to walk the entire enumerable.    If you don't have a known enumerable type, `Any()` would be better than `Count()` for asking "is there anything in there?"    In this case, we know it's a list, so we can go direct.

tools/issue-labeler/src/IssueLabelerService/Answers/RagQna.cs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                message = $"Sources:\nDocumentation:\n{string.Join("\n", printableDocs)}\nGitHub Issues:\n{string.Join("\n", printableIssues)}\nThe user needs a solution to their GitHub Issue:\n{query}";
+            }
+            else
+            {
+                message = $"Sources:\nDocumentation:\n{string.Join("\n", printableDocs)}\nGitHub Issues:\n{string.Join("\n", printableIssues)}\nThe user needs suggestions for their GitHub Issue:\n{query}";
+            }
+
+            // Structured output for the model
+            var structure = BinaryData.FromBytes("""
```
jsquire, line 97: Why not just "FromString" ?    Seems like we're doing a lot of work and multiple translations/allocations.

tools/issue-labeler/src/IssueLabelerService/Answers/RagQna.cs
```diff
@@ -0,0 +1,146 @@
...
+              "required": [ "Category", "Service", "Response" ],
+              "additionalProperties": false
+            }
+            """u8.ToArray());
+
+            var response = await _ragService.SendMessageQnaAsync(instructions, message, structure);
+
+            var resultObj = JsonConvert.DeserializeObject<AIOutput>(response);
+            string intro, outro;
+            Dictionary<string, string> replacements = new Dictionary<string, string>
```
jsquire, line 114:
```suggestion
            var replacements = new Dictionary<string, string>
```

Everything else around here is `var`.   Weird that this one isn't.

tools/issue-labeler/src/IssueLabelerService/Answers/RagQna.cs
jeo02: Yea I agree

tools/issue-labeler/src/IssueLabelerService/Answers/RagQna.cs
jeo02: Wait actually we don't even need structure since we are only asking for the response now

tools/issue-labeler/src/IssueLabelerService/Answers/AnswerFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public AnswerFactory(ILogger<AnswerFactory> logger, TriageRag ragService)
+        {
+            _logger = logger;
+            _ragService = ragService;
+        }
+
+        public IAnswerService GetQna(RepositoryConfiguration config) =>
+            _qnaServices.GetOrAdd(
+                config.QnaService,
```
jsquire, line 20:
```suggestion
                config.AnswerService,
```

jsquire:
One last nit, otherwise looks great.   Let me know when you're ready for me to approve.

jeo02:
Should be good

tools/issue-labeler/src/IssueLabeler.Shared/QnaOutput.cs
jsquire: Resolved, but not, uh, resolved?    😄

jeo02: Okay resolved for real, changed so many qnas to answer I somehow missed this

jsquire (approved):

PR #10301, opened 2025-04-14
[tsp-client] Support untracked deps

catalinaperalta:
Keep any manually added dependencies specified in emitter-package.json when using the `generate-config-files` command. If the target emitter has a package-lock.json, attempt to update the manually added dependency with the pinned version, if not found, then keep the dependency version the same.

NOTE: there were some formatting updates that also made it into this PR due to some other contributions that didnt format files properly.


catalinaperalta:
@allenjzhang this will help folks keep their liftr dependencies in emitter-package.json

tools/tsp-client/src/commands.ts
```diff
@@ -480,13 +481,55 @@ export async function generateConfigFilesCommand(argv: any) {
...
+      if (untrackedDependencies) {
+        // Add untracked dependencies to the updated emitter-package.json devDependencies
+        // with the version from the existing emitter-package.json
+        for (const key of untrackedDependencies) {
+          emitterPackageJson["devDependencies"][key] =
+            existingEmitterPackageJson["devDependencies"][key];
+        }
+        // Check if the target emitter has a lock file and attempt to pull untracked
+        // dependency versions from there
+        const packageLockPath = joinPaths(dirname(packageJsonPath), "package-lock.json");
```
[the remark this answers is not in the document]
catalinaperalta: This is defined at the beginning of the function.

tools/tsp-client/package.json
```diff
@@ -38,10 +38,12 @@
   ],
   "devDependencies": {
     "@types/chai": "^4.3.5",
+    "@types/lodash": "^4.17.16",
     "@types/node": "^20.4.8",
     "@types/prompt-sync": "^4.2.1",
     "@typespec/compiler": "^1.0.0-0",
     "chai": "^4.3.7",
+    "lodash": "^4.17.21",
```
timotheeguerin, line 46: i would try not to use lodash, its now an old package that hasn't been updated in 4 years

tools/tsp-client/src/commands.ts
```diff
@@ -480,13 +480,50 @@ export async function generateConfigFilesCommand(argv: any) {
...
+    const existingEmitterPackageJson = JSON.parse(await readFile(emitterPath, "utf8"));
+    // If there's an existing emitter-package.json, we need to merge any untracked devDependencies
+    if (existingEmitterPackageJson) {
+      const currentDevDependencies = existingEmitterPackageJson["devDependencies"] ?? {};
+      const packageLockPath = joinPaths(dirname(packageJsonPath), "package-lock.json");
+      let existingPackageLockJson;
+      try {
+        // Check if package-lock.json exists
+        await stat(packageLockPath);
+        existingPackageLockJson = JSON.parse(await readFile(packageLockPath, "utf8"));
```
timotheeguerin, line 498: is there a chance that the lock file is invalid, this could log a confusing error then.

Also I think as you doing a whole catch you can just catch the read(remove the stat)

catalinaperalta: Could be, I updated the message we log to be more generic. Also removed the stat call

tools/tsp-client/src/commands.ts
```diff
@@ -480,13 +480,49 @@ export async function generateConfigFilesCommand(argv: any) {
...
+      } catch (err) {
+        Logger.debug(`Unable to read package-lock.json: ${packageLockPath}`);
+      }
+      // If there are untracked dependencies, we will check the package-lock.json of the
+      // target emitter for a compatible version, otherwise we will leave the version from
+      // the emitter-package.json unchanged
+      for (const key of Object.keys(currentDevDependencies)) {
+        if (
+          emitterPackageJson["devDependencies"] &&
```
JoshLove-msft, line 506: What if there is no "devDependencies" in the "current" not yet written version? Don't we still need to write any dev dependencies from the existing version?

tools/tsp-client/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release
 
+## 2025-04-16 - 0.21.0
+
+- Forward untracked dependencies in emitter-package.json when using the `generate-config-files` command.
```
JoshLove-msft, line 5: What do we mean by "untracked" dependencies? Are these dependencies that have been manually added to the emitter-package.json file? I thought we will always be generating the emitter-package.json file going forward with tsp-client.

catalinaperalta: We've had several discussions around this and folks require this flexibility in certain cases. Allen has been working on a specific scenario where we want to enable this flexibility so manually pinned dependencies arent erased. Likewise the rust repo also wants to allow this for their repo. Do you have any specific concerns? Repo owners should still be reviewing any PRs updating their emitter-package.json file.

catalinaperalta: I will update the wording here to say manually pinned so this change is clearer, thanks for pointing that out!

catalinaperalta: I dont think we're there yet, but if in the future someone wants to reset their emitter-package.json to be purely based on the deps in the package.json of the target emitter, we can add a flag to force that

tools/tsp-client/src/commands.ts
JoshLove-msft: But I thought this loop is trying to retain the manually added dependencies from the emitter-package-json. I'm wondering why we wouldn't have the same behavior if the generated emitter package json happens to not have a devDependencies section. i.e. why is this special?

tools/tsp-client/CHANGELOG.md
JoshLove-msft: I think the wording was my main feedback. I can think of scenarios where this would be useful for C#, e.g. for defining a service specific plugin to the emitter. That said, I think in that case we would end up duplicating the majority of the emitter-package.json between service and repo root, so I was planning on going another route to support the plugins.

JoshLove-msft: One other question - what if the actual package.json removes a devDependency? Won't the changes in this PR prevent that from ever getting removed in the emitter-package-json?

tools/tsp-client/src/commands.ts
catalinaperalta: Ah I think I misunderstood your first question, you're talking about the case where we dont have any pinned dev deps. I didnt take it into account since I dont think we have any emitter that doesnt, but that doesnt mean someone wont run into this. I'll support that corner case, thanks for the feedback!

tools/tsp-client/CHANGELOG.md
catalinaperalta: Ah yeah I think in that case, we'd probably use the alternate path for emitter-package json, right?

catalinaperalta: Tracking issue to add a flag to do a hard reset on emitter-package.json: https://github.com/Azure/azure-sdk-tools/issues/10333

Will add that in a separate PR, I dont think we need to block here on that.

catalinaperalta:
Spoke with @JoshLove-msft offline last week and this change is good to move forward so that we keep untracked deps in the emitter-package.json files when folks use this command. Tracking issue to add a flag for a hard reset when needed: https://github.com/Azure/azure-sdk-tools/issues/10333. 

@JoshLove-msft @timotheeguerin please review the pr.

tools/tsp-client/test/examples/generate-config-files/package-lock.json
timotheeguerin, line 1: do you need the lock files in the tests?

tools/tsp-client/test/commands.spec.ts
```diff
@@ -331,27 +340,102 @@ describe.sequential("Verify commands", () => {
...
+        "0.53.0",
+      );
+      assert.isUndefined(emitterJson["overrides"]);
+      assert.isTrue(
+        await doesFileExist(
+          joinPaths(repoRoot, "tools/tsp-client/test/utils/emitter-package-extra-dep-b-lock.json"),
+        ),
+      );
+      // Clean up the generated files
+      await rm(joinPaths(dirname(emitterPackageJsonPath), "emitter-package-extra-dep-b-lock.json"));
```
timotheeguerin: nit: but for consitency if you using the compiler path utils there is a getDirectoryName or something like that from there

tools/tsp-client/src/commands.ts
```diff
@@ -247,7 +247,7 @@ export async function syncCommand(argv: any) {
 
   try {
     let emitterLockPath = getEmitterLockPath(getEmitterPackageJsonPath(repoRoot, tspLocation));
-    
+
```
timotheeguerin, line 250: is there nothing that enforce formatting?

tools/tsp-client/src/commands.ts
```diff
@@ -474,13 +474,65 @@ export async function generateConfigFilesCommand(argv: any) {
...
+    const manualDevDependencies = {};
+    for (const [key, value] of Object.entries(
+      existingEmitterPackageJson["devDependencies"] ?? {},
+    )) {
+      if (!Object.keys(emitterPackageJson["devDependencies"] ?? {}).includes(key)) {
+        Object.assign(manualDevDependencies, { [key]: value });
+      }
+    }
+
+    // Attempt to read package-lock.json and find the version of the manually added dependencies
```
JoshLove-msft: Why do we need the version from the package-lock? Shouldn't we use the version from emitter package JSON file?

tools/tsp-client/test/examples/generate-config-files/package-lock.json
catalinaperalta: Yes, since we try to look for untracked deps in the lock file.

tools/tsp-client/src/commands.ts
catalinaperalta: Nope, I have to add it soon. Tracking issue for that: https://github.com/Azure/azure-sdk-tools/issues/11073

tools/tsp-client/src/commands.ts
catalinaperalta: I'd discussed this a while back with Patrick, we thought that checking the lock file would at least help update to newer versions, kind of like a best effort to update to a version pinned in the lock file that's compatible with the other deps, but I dont feel strongly about this. We can just leave the manually pinned dep version the same

catalinaperalta: Let me know and I'll update accordingly

JoshLove-msft: I would just leave it the same for simplicity's sake. I believe the behavior would be the same regardless - if a newer version is available and we aren't pinning to specific version it should still get used.

tools/tsp-client/test/examples/generate-config-files/package-lock.json
catalinaperalta: This is now removed since we're just going to keep the manual dep at the same version specified in the emitter-package.json

tools/tsp-client/test/commands.spec.ts
catalinaperalta: I'm not finding this in the available compiler methods.

tools/tsp-client/src/commands.ts
catalinaperalta: Done

tools/tsp-client/test/commands.spec.ts
catalinaperalta: going to merge so we can release this today, let me know if you have the name of the method and I'll do a follow up PR to make this change :)

PR #10320, opened 2025-04-15
Fix line id handling

### LINKED ISSUE #10316: [APIView Copilot] Improve line number detection
This has always been an issue, that the LLM isn't very good as figuring out line numbers.

We tell it to provide a single line number, but sometimes it just doesn't comply. It will output ranges (x-y) or lists (x, y). Rather than try to constraint its behavior it may be better to "roll with it" and interpret what it provides as a line number.

Separately, we also need to ensure the line number makes sense.

tjprescott:
Fixes #10316.

This handles the following possible scenarios that could be returned by the LLM for `line_no`:
- if a comma-separated list is provided, split and create a `Violation` each line in the list
- if a range is provided (i.e. 10-20) just take the first value since APIView doesn't support line ranges

Regardless, the `line_no`, while received by the LLM as a string, is cast to an int as inteded.

### DIFF
--- a/packages/python-packages/apiview-copilot/prompts/review_apiview_gpt_4o_mini.prompty
+++ b/packages/python-packages/apiview-copilot/prompts/review_apiview_gpt_4o_mini.prompty
@@ -86,8 +86,6 @@ followed.
 
 - APIView does not contain runnable code or implementations. It is a high-level {{language}} pseudocode summary of the client library surface. 
 
-- If related violations occur on multiple lines, report them as distinct violations and reference the other lines in the comment.
-
 # CONTEXT
 These are deemed the most relevant guidelines for this review. Ground your responses solely within this context.
 {{context}}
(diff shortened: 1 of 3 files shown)


packages/python-packages/apiview-copilot/cli.py
```diff
@@ -158,7 +158,12 @@ def deconstruct_test_case(language: str, test_case: str, test_file: str):
         f.write(apiview)
 
     with open(deconstructed_expected, "w") as f:
-        f.write(expected)
+        # sort violations by line number
+        expected = json.loads(expected)
+        expected["violations"] = sorted(
+            expected["violations"], key=lambda x: x["line_no"]
+        )
+        f.write(json.dumps(expected, indent=4))
```
tjprescott, line 166: @kristapratico this just makes sure it outputs violations sorted by line_id and indented

kristapratico (approved):

PR #10334, opened 2025-04-16
Update tspconfig options with kebab-case style in JS emitter

v-jiaodi:
related issue: https://github.com/Azure/autorest.typescript/issues/3067


tools/tsp-client/test/examples/specification/diagnostics/tspconfig.yaml
```diff
@@ -25,7 +25,7 @@ options:
     flavor: azure
   "@azure-tools/typespec-ts":
     package-dir: "contosowidgetmanager-rest"
-    packageDetails:
+    package-details:
```
MaryGao, line 28: @catalinaperalta JS emitter would switch to kebab-case style options so camel-case ones would be deprecated soon. So we updated this directly.

MaryGao (approved):

catalinaperalta (approved):

PR #10346, opened 2025-04-16
[js-api-parser] support `@deprecated` tag

jeremymeng:


jeremymeng:
Currently children's tokens are not set based on their parent's deprecation status, but we can change to do that if it's more desired.

jeremymeng:
Here's the example, as can be seen, only the first line of a deprecated type is marked but not its children

image

maorleger (approved):
🚀

PR #10347, opened 2025-04-16
[AI Triage] Utilizing Labels for Answer Service

jeo02:
- Filter through issues based on labels.
- Switched around the `OpenAiLabeler` output order since it was wrong.
- ~~Not filtering labels for Documents until dotnet documents have Service and Category labels.~~
- Added Filtering to Service and Category labels

### DIFF
--- a/tools/issue-labeler/src/IssueLabelerService/Answers/IAnswerService.cs
+++ b/tools/issue-labeler/src/IssueLabelerService/Answers/IAnswerService.cs
@@ -9,6 +9,6 @@ namespace IssueLabelerService
 {
     public interface IAnswerService
     {
-        public Task<AnswerOutput> AnswerQuery(IssuePayload issue);
+        public Task<AnswerOutput> AnswerQuery(IssuePayload issue, string[] labels);
     }
 }
--- a/tools/issue-labeler/src/IssueLabelerService/Answers/IAnswerService.cs
+++ b/tools/issue-labeler/src/IssueLabelerService/Answers/IAnswerService.cs
@@ -9,6 +9,6 @@ namespace IssueLabelerService
 {
     public interface IAnswerService
     {
-        public Task<AnswerOutput> AnswerQuery(IssuePayload issue, string[] labels);
+        public Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<string, string> labels);
     }
 }
--- a/tools/issue-labeler/src/IssueLabelerService/Labelers/LegacyLabeler.cs
+++ b/tools/issue-labeler/src/IssueLabelerService/Labelers/LegacyLabeler.cs
@@ -4,6 +4,7 @@ using Microsoft.Extensions.Logging;
 using IssueLabeler.Shared;
 using System.Threading.Tasks;
 using System;
+using System.Collections.Generic;
 
 
 namespace IssueLabelerService
@@ -32,7 +33,7 @@ namespace IssueLabelerService
             _config = config;
         }
 
-        public async Task<string[]> PredictLabels(IssuePayload issue)
+        public async Task<Dictionary<string, string>> PredictLabels(IssuePayload issue)
         {
             var predictionRepositoryName = TranslateRepoName(issue.RepositoryName);
 
@@ -84,7 +85,12 @@ namespace IssueLabelerService
                 }
 
                 _logger.LogInformation($"Labels were predicted for {issue.RepositoryName} using the `{predictionRepositoryName}` model for issue #{issue.IssueNumber}.  Using: [{predictions[0]}, {predictions[1]}].");
-                return [predictions[0], predictions[1]];
+                
+                return new Dictionary<string, string>
+                {
+                    { "Category", predictions[0] }, 
+                    { "Service", predictions[1] }
+                };
             }
             catch (Exception ex)
             {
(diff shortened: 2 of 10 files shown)


tools/issue-labeler/src/IssueLabelerService/Answers/IAnswerService.cs
```diff
@@ -9,6 +9,6 @@ namespace IssueLabelerService
 {
     public interface IAnswerService
     {
-        public Task<AnswerOutput> AnswerQuery(IssuePayload issue);
+        public Task<AnswerOutput> AnswerQuery(IssuePayload issue, string[] labels);
```
jsquire, line 12: Is it reasonable to make this less specific?  Like `IEnumerable<string> labels` or do we have a hard requirement on an array, specifically?

tools/issue-labeler/src/IssueLabelerService/Labelers/OpenAiLabeler.cs
```diff
@@ -90,7 +90,7 @@ public async Task<string[]> PredictLabels(IssuePayload issue)
 
             _logger.LogInformation($"Open AI Response for {issue.RepositoryName} using the Open AI Labeler for issue #{issue.IssueNumber} Service: {output.Service} Category: {output.Category}");
 
-            return [output.Service, output.Category];
+            return [output.Category, output.Service];
```
jsquire, line 93: We have to be careful not to depend on ordering here.

tools/issue-labeler/src/IssueLabelerService/TraigeRag.cs
```diff
@@ -121,8 +124,10 @@ public async Task<List<Document>> SearchDocumentsAsync(
             string field,
             string query,
             int count,
-            double scoreThreshold)
+            double scoreThreshold,
+            string[] labels = null)
```
jsquire, line 128:
```suggestion
            IEnumerable<string> labels = null)
```

tools/issue-labeler/src/IssueLabelerService/TraigeRag.cs
```diff
@@ -205,6 +213,19 @@ public double GetHighestScore(IEnumerable<Issue> issues, IEnumerable<Document> d
 
             return highestScore;
         }
+
+        public string LabelsFilter(string[] labels)
```
jsquire, line 218:
```suggestion
        public string LabelsFilter(IEnumerable<string> labels)
```

tools/issue-labeler/src/IssueLabelerService/TraigeRag.cs
```diff
@@ -205,6 +213,19 @@ public double GetHighestScore(IEnumerable<Issue> issues, IEnumerable<Document> d
...
+        {
+            if (labels != null && labels.Length >= 2)
```
jsquire, line 220:
```suggestion
            if (labels?.Take(2).Count() == 2)
```

 We will need to be able to handle other schemes from non-Azure SDK repositories.  We should move `2` into some kind of `minimum label predictions for answer` config eventually.

tools/issue-labeler/src/SearchIndexCreator/IssueIndex.cs
```diff
@@ -263,8 +263,14 @@ private SearchIndex GetSampleIndex()
                         IsSearchable = false
                     },
                     new SearchableField("Title"),
-                    new SearchableField("Service"),
-                    new SearchableField("Category"),
+                    new SearchableField("Service")
```
jsquire, line 266: At some point, we'll need to find a more generic way to do this.   Or, we'll need to deal with it being super-specific to us and allow for some dynamic plugin system for loading custom labelers/answer thingers on demand.

tools/issue-labeler/src/IssueLabelerService/Answers/IAnswerService.cs
jeo02: No hard requirement, can definitely be changed

jeo02:
Instead of replacing with `IEnumerable<string>` replaced with `Dictionary<string, string>` in all recommended spots above. This allowed me to get rid of the hard coded Category and Service parts and rely on the Dictionary key and values. 

Also added in a new config for `LabelNames`

tools/issue-labeler/src/IssueLabelerService/Labelers/LegacyLabeler.cs
```diff
@@ -84,7 +85,12 @@ public async Task<string[]> PredictLabels(IssuePayload issue)
                 }
 
                 _logger.LogInformation($"Labels were predicted for {issue.RepositoryName} using the `{predictionRepositoryName}` model for issue #{issue.IssueNumber}.  Using: [{predictions[0]}, {predictions[1]}].");
-                return [predictions[0], predictions[1]];
+                
+                return new Dictionary<string, string>
+                {
+                    { "Category", predictions[0] },
```
jsquire, line 91: Since we key on these in a couple of places, it would be a good idea to move them to a constants class.

tools/issue-labeler/src/IssueLabelerService/Answers/IAnswerService.cs
```diff
@@ -9,6 +9,6 @@ namespace IssueLabelerService
 {
     public interface IAnswerService
     {
-        public Task<AnswerOutput> AnswerQuery(IssuePayload issue);
+        public Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<string, string> labels);
```
jsquire, line 12: Your call on the dictionary vs. class.

jsquire (approved):

jsquire (approved):

PR #10369, opened 2025-04-18
[APIView Copilot] Support multiple new models

tjprescott:
Adds support for gpt-4o (non-mini), gpt-4.1 (non-mini) and o4-mini. Refactors prompty files so we don't always need a new prompt, since a lot of models have the same parameter sets.


packages/python-packages/apiview-copilot/app.py
```diff
@@ -23,5 +23,7 @@ def api_reviewer(language: str):
         return jsonify({"error": "Unsupported language"}), 400
     data = request.get_json()
     content = data["content"]
-    result = ApiViewReview(language=language, model="o3-mini").get_response(content)
+    result = ApiViewReview(language=language, model="gpt-4.1-nano").get_response(
```
kristapratico, line 26: is this the new default we want to set here?

kristapratico (approved):

packages/python-packages/apiview-copilot/app.py
tjprescott: Ah, no it isn't.

kristapratico (approved):

Issue #10310, opened 2025-04-14
[AI Triage] Prototype for AI-based label prediction


jsquire:
Based on Xiang's learnings, we know that we need an independent workflow with its own dedicated prompt to improve label prediction.   In order to test feasibility in the real world, we'll want to prototype one and enable in the .NET and Python repositories.   This prediction should include at least a rudimentary system for confidence scoring and be set at a 96% threshold for parity with the legacy ML-based label prediction.

Because of the desire to enable this more quickly, we should avoid looking into an external service for confidence scoring now and cover those efforts with a dedicated spike in the near future.

xiangyan99:
I will work with Juan on this.

PR #10379, opened 2025-04-18
[Ai Triage] Rag Labeler

jeo02:
- Store the labels in storage and get them when an issue comes back.
- Used Xiangs promp and added the set of labels from storage + evaluations.
- Fixed the deleted check for EnableAnswers

### DIFF
--- a/tools/issue-labeler/src/IssueLabelerService/AzureSdkIssueLabelerService.cs
+++ b/tools/issue-labeler/src/IssueLabelerService/AzureSdkIssueLabelerService.cs
@@ -69,32 +69,48 @@ namespace IssueLabelerService
                 return EmptyResult;
             }
 
-            try{
-
-                // Get the Qna model based on configuration
-                var qnaService = _answerServices.GetAnswerService(config);
-
-                var answer = await qnaService.AnswerQuery(issue, labels);
-
-                TriageOutput result = new TriageOutput
-                {
-                    Labels = labels.Values,
-                    Answer = answer.Answer,
-                    AnswerType = answer.AnswerType
-                };
-
-                return new JsonResult(result);
-            }
-            catch (Exception ex)
+            if(bool.Parse(config.EnableAnswers))
             {
-                _logger.LogError($"Error commenting on issue #{issue.IssueNumber} in repository {issue.RepositoryName}: {ex.Message}{Environment.NewLine}\t{ex}{Environment.NewLine}");
+                try{
 
+                    // Get the Qna model based on configuration
+                    var qnaService = _answerServices.GetAnswerService(config);
+
+                    var answer = await qnaService.AnswerQuery(issue, labels);
+
+                    TriageOutput result = new TriageOutput
+                    {
+                        Labels = labels.Values,
+                        Answer = answer.Answer,
+                        AnswerType = answer.AnswerType
+                    };
+
+                    return new JsonResult(result);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError($"Error commenting on issue #{issue.IssueNumber} in repository {issue.RepositoryName}: {ex.Message}{Environment.NewLine}\t{ex}{Environment.NewLine}");
+
+                    TriageOutput result = new TriageOutput
+                    {
+                        Labels = labels.Values,
+                        Answer = null,
+                        AnswerType = null
+                    };
+
+                    return new JsonResult(result);
+                }
+            }
+            else
+            {
                 TriageOutput result = new TriageOutput
                 {
                     Labels = labels.Values,
                     Answer = null,
                     AnswerType = null
                 };
+
+                return new JsonResult(result);
             }
         }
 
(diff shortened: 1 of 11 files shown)


tools/issue-labeler/src/IssueLabelerService/Labelers/OpenAiLabeler.cs
```diff
@@ -73,31 +86,116 @@ public async Task<Dictionary<string, string>> PredictLabels(IssuePayload issue)
...
+            }
+
+            foreach (var label in filteredOutput)
+            {
+                try
+                {
+                    var confidence = double.Parse(output[$"{label.Key}ConfidenceScore"]);
+                    if(label.Value == "UNKOWN")
+                    {
+                        throw new InvalidDataException($"Open AI Response for {issue.RepositoryName} using the Open AI Labeler for issue #{issue.IssueNumber} provided an UNKOWN label.");
```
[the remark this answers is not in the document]
jeo02: Good catch copilot

tools/issue-labeler/src/IssueLabelerService/AzureSdkIssueLabelerService.cs
```diff
@@ -69,32 +69,48 @@ public async Task<ActionResult> Run([HttpTrigger(AuthorizationLevel.Function, "P
...
 
-            try{
+            if(bool.Parse(config.EnableAnswers))
+            {
+                try{
 
-                // Get the Qna model based on configuration
-                var qnaService = _answerServices.GetAnswerService(config);
+                    // Get the Qna model based on configuration
+                    var qnaService = _answerServices.GetAnswerService(config);
```
jsquire, line 77:
```suggestion
                    var answerService = _answerServices.GetAnswerService(config);
```

tools/issue-labeler/src/IssueLabelerService/AzureSdkIssueLabelerService.cs
```diff
@@ -69,32 +69,48 @@ public async Task<ActionResult> Run([HttpTrigger(AuthorizationLevel.Function, "P
...
             }
 
-            try{
+            if(bool.Parse(config.EnableAnswers))
+            {
+                try{
 
-                // Get the Qna model based on configuration
-                var qnaService = _answerServices.GetAnswerService(config);
+                    // Get the Qna model based on configuration
```
jsquire, line 76:
```suggestion
                    // Get the answer service based on configuration
```

tools/issue-labeler/src/IssueLabelerService/AzureSdkIssueLabelerService.cs
```diff
@@ -69,32 +69,48 @@ public async Task<ActionResult> Run([HttpTrigger(AuthorizationLevel.Function, "P
...
+                catch (Exception ex)
                 {
-                    Labels = labels.Values,
-                    Answer = answer.Answer,
-                    AnswerType = answer.AnswerType
-                };
+                    _logger.LogError($"Error commenting on issue #{issue.IssueNumber} in repository {issue.RepositoryName}: {ex.Message}{Environment.NewLine}\t{ex}{Environment.NewLine}");
 
-                return new JsonResult(result);
+                    TriageOutput result = new TriageOutput
```
jsquire, line 94: We can avoid the duplication here by moving things around a bit.  Something like:

```csharp

var result = new TriageOutput { Labels = labels.Values };

if (iWantAnswersNao)
{
    try
    {
        var answers = GetAnswers(...);
        result.Answer = answers.Answer;
        result.AnswerType = answers.AnswerType;
    }
    catch (Exception ex)
    {
        // Log - ERROR, OMG OMG!

        // Very unlikely this is needed, but covers any weird corner case.
        result.Answer = null;
        result.AnswerType = null;
    }

    return new JsonResult(result);
}
```

tools/issue-labeler/src/IssueLabelerService/Labelers/LegacyLabeler.cs
```diff
@@ -88,8 +88,8 @@ public async Task<Dictionary<string, string>> PredictLabels(IssuePayload issue)
                 
                 return new Dictionary<string, string>
                 {
-                    { "Category", predictions[0] }, 
-                    { "Service", predictions[1] }
+                    { "Service", predictions[0] },
```
jsquire, line 91: We should still move the magic strings into a consts class somewhere since we set them in one place and use them from some place else.   That way, we don't ever risk a mismatch on the values/casing/whatever.

Something like:

```csharp
public static class LabelType
{
    public const string Category = "Category";
    public const string Service = "Service";
}
```

This gives us the benefit of strong typing and compile time safety for typos.

tools/issue-labeler/src/IssueLabelerService/AzureSdkIssueLabelerService.cs
```diff
@@ -69,49 +69,35 @@ public async Task<ActionResult> Run([HttpTrigger(AuthorizationLevel.Function, "P
...
-                        Answer = null,
-                        AnswerType = null
-                    };
-
-                    return new JsonResult(result);
+                    result.Answer = null;
+                    result.AnswerType = null;
                 }
             }
             else
```
jsquire, line 94: You can remove the `else`.    The `null` values are the default.   We included this explicitly in the `catch` block above just in case there was an error that occurred setting them, to ensure there's no inconsistent state (like answer being set but answer type being null).

jsquire (approved):

jsquire (approved):

jeo02:
Resolves #10310

PR #10399, opened 2025-04-21
[APIView Copilot] Merge general and guideline-oriented reviews with judge prompts

tjprescott:
- On each chunk, runs the general (currently, only Python) prompt along with the guideline-oriented prompt. Merges the result, favoring comments that cite guidelines when there are duplicates.
- Runs the combined results through a judge prompt to improve the final results.


packages/python-packages/apiview-copilot/prompts/final_comments_filter.prompty
```diff
@@ -0,0 +1,109 @@
...
+  parameters:
+    stop: []
+    frequency_penalty: 0
+    presence_penalty: 0
+    max_completion_tokens: 20000
+    reasoning_effort: "high"
+    response_format: ${file:filtered_result_schema.json}
+sample:
+  language: python
+  exceptions: |
```
kristapratico, line 24: Can this just reference the yaml file you created ?

https://github.com/Azure/azure-sdk-tools/blob/e6cf8482986e201f7f966e9b731a76deae37ef76/packages/python-packages/apiview-copilot/metadata/python/filter.yaml

packages/python-packages/apiview-copilot/prompts/final_comments_filter.prompty
```diff
@@ -0,0 +1,109 @@
...
+    6. Suggest adding docstrings
+    7. Suggest using pydantic or dataclasses for models
+    8. Comment on async list method naming
+    9. Comment on indentation or namespace declaration
+    10. Suggest consolidating multiple overloads
+    11. Suggest providing convenience methods directly on the client
+    12. Comment on non-standard use of TypedDict syntax
+    13. Comment about ivar being non-standard use
+    14. Comment about use of distributed_trace/async decorators
+    15. Comment about methods ending with :
```
kristapratico, line 39: It looks like https://github.com/Azure/azure-sdk-tools/blob/e6cf8482986e201f7f966e9b731a76deae37ef76/packages/python-packages/apiview-copilot/metadata/python/filter.yaml is missing 15

packages/python-packages/apiview-copilot/prompts/generic_review.prompty
```diff
@@ -12,14 +13,21 @@ model:
...
-    stop: []
     frequency_penalty: 0
     presence_penalty: 0
     max_completion_tokens: 20000
     reasoning_effort: "high"
-    response_format: ${file:general_review_result_schema.json}
+    response_format: ${file:generic_result_schema.json}
 sample:
   language: python
+  custom_rules: |
```
kristapratico, line 23: same comment, should this reference the yaml? https://github.com/Azure/azure-sdk-tools/blob/e6cf8482986e201f7f966e9b731a76deae37ef76/packages/python-packages/apiview-copilot/metadata/python/guidance.yaml

packages/python-packages/apiview-copilot/prompts/generic_review_judge.prompty
```diff
@@ -0,0 +1,109 @@
...
+  4. Return a filtered set of improvements in the same schema
+
+  For each improvement I will:
+  - Verify it doesn't comment on excluded aspects
+  - Ensure it focuses on API design quality
+  - Validate it improves developer experience
+  - Confirm it aligns with {{language}} best practices
+
+  The response will maintain the schema structure with:
+  - A status field ("Success" or "Error")
```
kristapratico, line 68: It doesn't look like the generic_result_schema.json has a status field anymore. I think that affects the samples below as well

packages/python-packages/apiview-copilot/prompts/guidelines_review_judge.prompty
```diff
@@ -12,18 +13,17 @@ model:
...
-    response_format: ${file:general_review_result_schema.json}
+    response_format: ${file:review_result_schema.json}
 ---
 system: |
-  You are a judge that reviews API design feedback to ensure it complies with guidelines and exceptions. Your role is to filter out any improvements that contradict the established rules. You will receive:
+  You are a judge that reviews API design feedback to ensure it complies with guidelines. Your role is to filter out any violations that contradict the established rules. You will receive:
   1. The original APIView content
   2. The initial review results
-  3. The guidelines and exceptions that must be followed
+  3. The guidelines that must be followed
```
kristapratico, line 26: Do we need this judge prompt? Or is this the one you said you were going to play more with tomorrow?

packages/python-packages/apiview-copilot/prompts/final_comments_filter.prompty
kristapratico: Also maybe I missed it and code does reference the yaml, but then should we remove the duplication here?

tjprescott: I'm not entirely sure if I can just reference the yaml file, but even if I can, you want this sample to be different. This is the sample that is used when you "Run" the prompty file. If you provide a sample as part of the context, like this does, and then give it that same input, naturally you would always expect the same exact output.

packages/python-packages/apiview-copilot/prompts/generic_review_judge.prompty
tjprescott: Ah, yes, that's true. I'll cut that part. The reason I cut it was because we weren't trusting the LLM output anyways and were setting it programmatically based on whether there were comments or not, so it was just one more thing to potentially have to "fix" (that, AFAIK, we weren't using).

packages/python-packages/apiview-copilot/prompts/guidelines_review_judge.prompty
tjprescott: Yeah, the two judge prompts aren't wired up. This PR is pretty big, so I might defer it to a separate PR since most of the value is added by the generic and filter prompts. I'll probably leave the judge prompts in though so it's easy to resume.

packages/python-packages/apiview-copilot/prompts/final_comments_filter.prompty
tjprescott: The reason I can pass the yaml metadata in below the `---` is because that is all Jinja templates, so I can load from a file and pass into the template. However everything above the `---` is regular yaml, not a jinja template, so I can't pass params in that way. I could do `{$env:SOME_ENV_VAR}` to pass things the way I do the model for some prompts. However, when you run prompty, this would require you to put that ENV_VAR in your `.env` file since that's all prompty loads before expanding the template, so even then, it wouldn't actually be loading it from the yaml.

I think in terms of inner-dev loop, the cycle would be you change the sample guidelines, gauge the results, and if they are better, save them into the metadata yaml file (exaclty like I did NOT)

tjprescott: I am curious though if you have a sense for how much passing a sample in the prompt itself improved results?

packages/python-packages/apiview-copilot/prompts/generic_review_judge.prompty
tjprescott: For this PR I removed the judge prompts

packages/python-packages/apiview-copilot/prompts/guidelines_review_judge.prompty
tjprescott: Actually, to keep things clean I removed the judge prompts. I can always get them back from git history for the next PR

packages/python-packages/apiview-copilot/prompts/final_comments_filter.prompty
tjprescott: Hm... it might be possible wiht the `${file:blah}` syntax, though I suspect I'll need to split the yaml into separate files. Let me try it.

tjprescott: Well, the problem with file syntax is that the file must exist and can't be programmatically determined, so that won't really work either. 😢  We use the language to construct the path, and apply defaults if the path doesn't exist.

tjprescott: Well, I tried another way, and the problem is that the `sample` key in the config won't read a file at all. I wonder if it's only supported for the response_format property.

tjprescott: Resolved per our call

packages/python-packages/apiview-copilot/prompts/generic_review.prompty
tjprescott: Resolved per our call

kristapratico (approved):

kristapratico (approved):

PR #10411, opened 2025-04-22
Treat triggerpath added packages as `direct`-ly changed packages

scbedd:
Instead of `indirect`. The key here is that I want to include `Azure.Core` in `build/analyze` for `net - pullrequest`, but I _don't_ want to include all the additional detected `Azure.Core` dependent packages. This change ensures that packages that are included by their `triggerPath` will be treated as `direct`ly changed. Dependency added packages will still be treated as indirect.

@JimSuplizio for FYI.

This should get the experience in `build/analyze` for `NET` that we want, without having any special case logic.

- [x] todo: update the test cases to reflect this new reality.


weshaggard:
Perhaps I'm missing something but doesn't this defeat the purpose of why we introduced the indirect references? When will indirect ever be set?

scbedd:
> Perhaps I'm missing something but doesn't this defeat the purpose of why we introduced the indirect references? When will indirect ever be set?

When the additional package comes through as a dependency from `AdditionalValidationPackages`. In that case, the package is still included as indirect.

So for `.NET`, this would apply to the long list of packages that depend on `Azure.Core` when we change `Azure.Core` directly.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -395,7 +395,7 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
                     $includedForValidation = ($filePath -like (Join-Path "$resolvedRelativePath" "*") -or $filePath -eq $resolvedRelativePath)
                     $shouldInclude = $shouldInclude -or $includedForValidation
                     if ($includedForValidation) {
-                        $pkg.IncludedForValidation = $true
+                        $pkg.IncludedForValidation = $false
```
weshaggard, line 398: Do we need to explicitly set this to false or is that the default and we can just skip this if block completely?

eng/common/scripts/Package-Properties.ps1
```diff
@@ -415,10 +415,12 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
                     $directory = [System.IO.Path]::GetDirectoryName($ciYml).Replace("`\", "/")
```
weshaggard, line 415: orthogonal just noticed while reading. Given the first think you do is resolve the full path to the triggering files each time you use them it might be worth doing that once when you capture them.

eng/common/scripts/Package-Properties.ps1
scbedd: We want to keep the `break` in place as there is no reason to continue iterating over files for this package, but I do agree about removing the `$pkg.IncludedForValidation` setting.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -415,10 +415,12 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
...
-                    # files that are directly included in triggerPaths will kept in full form, but otherwise we pre-process the targetedFiles to the
-                    # directory containing the change. Given that pre-process, we should check both direct equality (when not triggeringPath) and parent directory
-                    # for the case where the full form of the file has been left behind (because it was a triggeringPath)
-                    $serviceDirectoryChange = (Split-Path $filePath -Parent).Replace("`\", "/") -eq $directory -or $filePath.Replace("`\", "/") -eq $directory
+                    # files that are directly included in triggerPaths will kept in full form, so owning ci.yml files will be present in their full form
+                    # and not as a directory. We need to check the parent directory if the file is a file and not a directory, but otherwise
+                    # we just need to check if the file is in the service directory
+                    $normalizedFilePath = $filePath.Replace("`\", "/")
+                    $parent = if ($normalizedFilePath -match '\.[^\\/]+$') { Split-Path $normalizedFilePath -Parent } else { $normalizedFilePath }
```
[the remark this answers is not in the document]
scbedd: I agree we should normalize stuff earliest as possible so we don't need to deal with this. Your suggestion will work. Simplifications coming.

eng/common/scripts/Package-Properties.ps1
scbedd: Ok. `changedFiles` will always be in `unix` style. `TriggeringPaths` will _also_ always be normalized when we populate it for `artifactdetails`. We don't need these `replace` at all aside from when we create the initial `filePath` from `Join-Path` at the top of this function.

weshaggard (approved):

weshaggard (approved):

PR #10462, opened 2025-04-25
[APIview copilot] add some additional exceptions for python

kristapratico:


packages/python-packages/apiview-copilot/prompts/final_comments_filter.prompty
```diff
@@ -37,6 +37,7 @@ sample:
     13. Comment about ivar being non-standard use
     14. Comment about use of distributed_trace/async decorators
     15. Comment about methods ending with :
+    16. Comment about the use of **kwargs
```
tjprescott, line 40: We may want to be more specific here so it doesn't throw out legitimate comments like it was doing for "...". For example, we don't want it to filter out these:

```
YOU SHOULD NOT have methods that require more than five positional parameters. Optional/flag parameters can be accepted using keyword-only arguments, or **kwargs.
```

kristapratico: I think you're right. I'm fixing up the evals right now and I think I'll need to make a few tweaks for exceptions so they are a bit more specific

kristapratico:
Closing in favor of https://github.com/Azure/azure-sdk-tools/pull/10493

PR #10471, opened 2025-04-25
[APIView Copilot] Implement better chunking algorithm

### LINKED ISSUE #10446: [APIView Copilot] Revisit chunking
Currently, we simply pass the whole APIView and do a single pass. However, for large APIs with a lot of comments, this often results in large sections not being reviewed.

We have a rudimentary chunking mechanism, but it works by splitting the APIView by blank lines. This can result in very small chunks (as small as one line!) and very large ones (entire classes) and not all language APIView use visual whitespace to break things up.

It might make more sense to have a line length (for example 500 lines) that determines the chunk size. This would result in more consistent chunk sizes and fewer overall requests than our current chunking implementation.

tjprescott:
Fixes #10446.

This makes chunking automatic (no longer optional) based on the chunk size.

It aims to send consistently sized chunks to the LLM (configured for approximately 500 line) to ensure that the entirety of the API surface area gets adequately reviewed.

For chunks that exceed the size (for example, model namespaces), it will break that into sub-sections and append the "header" to each to retain the context.  For example:

```
# chunk 1
10: namespace models
11:   class Foo
12:   class Bar
   ...

# chunk 2
10: namespace models
511:   class Widget
512:  class Zoodle
```

Note this is likely to result in some potential comment duplication or some false positives (for example, the "you need sync and async clients" will likely crop up if they are contained in separate batches). However, we should be able to engineer away those discrepancies. 

The purpose of this change is to ensure that the LLM doesn't just "give up" making comments because the APIView is large... after all, that's what we do and why we want the AI to review it for us 😅


packages/python-packages/apiview-copilot/src/_retry.py
```diff
@@ -0,0 +1,46 @@
...
+        description: Description of the operation for logging
+
+    Returns:
+        The result of the function call, or the result of on_failure if all retries fail
+    """
+    for attempt in range(max_retries):
+        try:
+            result = func()
+            return result
+        except retry_exceptions as e:
```
kristapratico, line 29: Should we narrow this to specific retryable exceptions?

kristapratico (approved):

packages/python-packages/apiview-copilot/src/_retry.py
tjprescott: Yeah, good point. Kinda pointless to retry 5 times if you send a bad request or something like that.

kristapratico (approved):

PR #10484, opened 2025-04-28
Add azure-mcp-pr to mirror-repos pipeline

hallipr:


benbp (approved):

eng/pipelines/mirror-repos.yml
```diff
@@ -135,6 +135,10 @@ jobs:
             TargetRepos:
               azure-sdk/azure-dev-docs-pr:
 
+          Azure/azure-mcp:
+            TargetRepos:
+              Azure/azure-mcp-pr:
```
weshaggard, line 140: This is going to break right now as the main branches are not in sync. We probably should revert this change until we have agreement to overwrite the private main branch.

Issue #10445, opened 2025-04-24
[APIView Copilot] Update evals to respect new schema + others


kristapratico:
TODOs

- ~respect new schema -- "expected" no longer matches~
- ~do not call false positive on "general" comments~
- ~do we still need fuzzy match with current line_no logic?~
- ~penalty for allowing comments to escape filter?~
- speed up evals / run in parallel
- account for variance -- when should we throw out a run that has too much variance and retry?

tjprescott:
Done in #10493.

PR #10493, opened 2025-04-29
[apiview copilot] update evals to new schema + handle generic comments

kristapratico:
Part of https://github.com/Azure/azure-sdk-tools/issues/10445

- respects new review result schema
- handles "generic" comments. If deemed invalid by model judge, count as false positives.
- resets baseline since I updated all the test files
- some tweaks to exceptions


packages/python-packages/apiview-copilot/evals/run.py
```diff
@@ -5,79 +5,145 @@
...
+        """Evaluate generic comments. If they are invalid, they count as false positives and receive penalty."""
+
+        filter_path = pathlib.Path(__file__).parent.parent / "metadata" / language / "filter.yaml"
+        with open(filter_path, "r") as f:
+            filter_data = yaml.safe_load(f)
+            exceptions = filter_data["exceptions"].strip().split("\n")
+            exceptions = [e.split(". ", 1)[1] for e in exceptions]
+
+        client = openai.AzureOpenAI()
+        for comment in generic_comments:
```
tjprescott, line 76: Since this is a query per generic comment, I'd recommend running these in parallel. Assuming o4-mini takes as long as o3-mini, running these in series would really drag out the run times.

kristapratico: gpt-4o is used as the model judge for all evals (including generic comments), and it's much faster than the reasoning models. I didn't sense a difference in run times by adding this. Overall, I do need to investigate how we can speed up evals, but I'm going to work on that in another PR (tracked in https://github.com/Azure/azure-sdk-tools/issues/10445)

kristapratico:
> One comment on performance (judging the generic comments) but overall LGTM. Do you have the overall baseline results handy in the terminal format? It would be nice to see them.
> 
> Might be good to save that table as markdown or something for quick reference.

Good idea, let me add the baseline table results in a txt

tjprescott (approved):

PR #10509, opened 2025-04-30
`azsdk-cli` MCP Server

scbedd:
FYI @benbp @weshaggard 

This PR:

- [x] Creates a project `Azure.SDK.Tools.Cli` that will run in either `CLI` or `mcp` mode
- [x] Supports `--tool X,Y,Z` filters


tools/mcp/dotnet/AzureSDKMCPHub/Azure.SDK.Tools.MCP.Hub/Tools/README.md
```diff
@@ -0,0 +1,21 @@
+# Writing tools that integrate with the MCP server
+
+## How to create your tool code
+
+- Inherit from `Azure.SDK.Tools.MCP.Contract::MCPHubTool`
+- Ensure your class has the `[McpServerToolType]` attribute added to it
+  - Eventual goal is to automatically recognize any classes that have this type as `[McpServerToolType]`.
+  - For now you need to add the attribute yourself.
```
benbp, line 8: azure-mcp does this (only attributing methods, no McpServerToolType) but I haven't dug in to see how. Probably a mix of DI/reflection (https://github.com/Azure/azure-mcp/blob/main/src/Commands/Server/ServiceStartCommand.cs#L127-L181).

tools/mcp/dotnet/AzureSDKMCPHub/Azure.SDK.Tools.MCP.Hub/Tools/README.md
```diff
@@ -0,0 +1,21 @@
...
+- Mark each function with `[McpServerTool]`.
+- Leverage `HelloWorld.cs` for a functioning basic example (more advanced samples to follow)
+- If you require any helper functions or other shared code. Place your code under a _folder_ named the same as your Tool class name.
+  - See `ReleasePlanTool` for an example of this folder structure.  
+
+## To quickly dev loop on just your class.
+
+- Ensure that your class in `Tools` is inheriting from the abstract class `MCPHubTool.`
+- Update the `debug` properties for the `hub` project to target a specific tool `--tool <nameofyourclass>`
```
benbp, line 17: I think long term we'll need some official way of building or running the server with only a set of tools and/or functions available (e.g. exposing the rest-api-spec MCP to partners without also having things like `eng/common PR checker` exposed.

benbp:
My bad for some outdated docs and directories (tools/mcp/dotnet), but as I mentioned I think this should be a CLI first app (with `server start` as a sub-command), and as such should would be better placed in a location like https://github.com/Azure/azure-sdk-tools/tree/main/tools/azsdk-cli. Thoughts?

tools/mcp/dotnet/AzureSDKMCPHub/Azure.SDK.Tools.MCP.Hub/Tools/README.md
scbedd: Understood. With external packages that's simpler. With the internal classes -- we'll have to think a bit.

scbedd:
> My bad for some outdated docs and directories (tools/mcp/dotnet), but as I mentioned I think this should be a CLI first app (with `server start` as a sub-command), and as such should would be better placed in a location like https://github.com/Azure/azure-sdk-tools/tree/main/tools/azsdk-cli. Thoughts?

I'll make this change now.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/CommandFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    public class CommandFactory
+    {
+        private readonly IServiceProvider _serviceProvider;
+        private readonly ILogger<CommandFactory> _logger;
+        private readonly ILoggerFactory _loggerFactory;
+
+        private readonly RootCommand _rootCommand;
+
+        public CommandFactory(IServiceProvider serviceProvider, ILoggerFactory loggerFactory)
```
hallipr, line 21: instead of taking IServiceProvider so you can use it in create instances of all of the MCPTool implementations, leave that in your IServiceCollection registrations and take `IEnumerable<MCPTool> tools` as a parameter

```suggestion
        public CommandFactory(IEnumerable<MCPTool> tools, ILoggerFactory loggerFactory)
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            builder.AddConsole();
+            builder.SetMinimumLevel(LogLevel.Debug);
+        });
+        services.AddSingleton<CommandFactory>();
+
+        // any registrations below here are only necessary for CLI mode
+        // the hosttool mode re-inits its own service provider
+        services.AddSingleton<IAzureService, AzureService>();
+
```
hallipr, line 27: Discover and register all of your tools here:
```suggestion

        RegisterTools(services);
```

where RegisterTools can allow each tool to register additional services, or just add all tools by base class from assembly lookup.

```
   HellowWorldTool.Register(services);
   
   vs
   
   AppDomain.CurrentDomain
                         .GetAssemblies()....
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/CommandFactory.cs
scbedd: I get where you're coming from. We may still end up doing that, but I believe I have this thing working the way we want, only registering the tools that are actually registered. Just gotta work through the reflection in `CreateRootCommand` and `HostServerTool.CreateAppBuilder`

benbp (approved):

PR #10521, opened 2025-05-02
Improve logging for package detection for pullrequests

weshaggard:
This eliminates the noisy changelog lines and add output helping to understand why a particular package was included in a PR validation build.

### DIFF
--- a/eng/common/scripts/ChangeLog-Operations.ps1
+++ b/eng/common/scripts/ChangeLog-Operations.ps1
@@ -19,7 +19,7 @@ function Get-ChangeLogEntries {
     LogError "ChangeLog[${ChangeLogLocation}] does not exist"
     return $null
   }
-  LogDebug "Extracting entries from [${ChangeLogLocation}]."
+  Write-Verbose "Extracting entries from [${ChangeLogLocation}]."
   return Get-ChangeLogEntriesFromContent (Get-Content -Path $ChangeLogLocation)
 }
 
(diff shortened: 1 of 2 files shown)


eng/common/scripts/ChangeLog-Operations.ps1
```diff
@@ -19,7 +19,7 @@ function Get-ChangeLogEntries {
     LogError "ChangeLog[${ChangeLogLocation}] does not exist"
     return $null
   }
-  LogDebug "Extracting entries from [${ChangeLogLocation}]."
+  Write-Verbose "Extracting entries from [${ChangeLogLocation}]."
```
benbp, line 22: Should we just add a LogVerbose?

eng/common/scripts/Package-Properties.ps1
```diff
@@ -448,6 +451,8 @@ function Get-PrPkgProperties([string]$InputDiffJson) {
             }
 
             if ($shouldInclude) {
+
+                Write-Host $includeMsg
```
benbp, line 455:
```suggestion
                LogInfo $includeMsg
```

etc. etc.

scbedd (approved):
@benbp has some suggestions. I'm approving the logging changes though.

eng/common/scripts/ChangeLog-Operations.ps1
weshaggard: This is a PS specific form of logging. The logging functions are mostly to abstract between local, devops, and github actions. So, I don't think that is needed. Perhaps we need another level eventually.

PR #10523, opened 2025-05-02
fix: bump node version in emitter pipeline

jorgerangel-msft:
This PR fixes a dependency issue with installing & using the `tsp-client` with node v18.x.
Tested in : https://github.com/Azure/azure-sdk-for-net/pull/49789

### DIFF
--- a/eng/common/pipelines/templates/archetype-typespec-emitter.yml
+++ b/eng/common/pipelines/templates/archetype-typespec-emitter.yml
@@ -241,6 +241,11 @@ extends:
               Paths:
               - "/*"
               - "!SessionRecords"
+          
+          - task: UseNode@1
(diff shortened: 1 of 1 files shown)


JoshLove-msft (approved):

eng/common/pipelines/templates/archetype-typespec-emitter.yml
```diff
@@ -307,6 +312,10 @@ extends:
           variables:
             matrixArtifactsPath: $(Pipeline.Workspace)/matrix_artifacts
           steps:
+          - task: UseNode@1
```
JoshLove-msft, line 315: nit: should this go after sparse checkout for consistency?

PR #10536, opened 2025-05-05
can we remove node 18?

l0lawrence:
need to update packagelock

### DIFF
--- a/tools/tsp-client/package.json
+++ b/tools/tsp-client/package.json
@@ -27,7 +27,7 @@
   "license": "MIT",
   "type": "module",
   "engines": {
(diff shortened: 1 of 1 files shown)


tools/tsp-client/package.json
```diff
@@ -27,7 +27,7 @@
   "license": "MIT",
   "type": "module",
   "engines": {
-    "node": "^18.19.0 || >=20.6.0"
+    "node": ">=20.6.0"
```
catalinaperalta, line 30: Yeah this is fine, since node18 was deprecated. Just run an npm install to update the package-lock.json file

PR #10542, opened 2025-05-05
[Tools] Update Most Current Taxonom

minhanh-phan:
- The current taxonomy is not up-to-date and still includes AAD. Updates to the latest one
- The current link points to a 404 page. Updates to the current taxonomy page

### DIFF
--- a/eng/common/scripts/Test-SampleMetadata.ps1
+++ b/eng/common/scripts/Test-SampleMetadata.ps1
@@ -73,7 +73,7 @@ process {
 
                 Write-Error "File '$($file.FullName)' contains invalid product slug: $product" -TargetObject $file `
                     -Category InvalidData -CategoryTargetName $product -CategoryTargetType string `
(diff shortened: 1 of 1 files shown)


ckairen (approved):

eng/common/scripts/Test-SampleMetadata.ps1
```diff
@@ -95,14 +95,11 @@ end {
 }
 
 begin {
-    # https://review.learn.microsoft.com/help/platform/metadata-taxonomies?branch=main#product
+    # https://taxonomy.learn.microsoft.com/TaxonomyServiceAdminPage/#/taxonomy/
```
weshaggard, line 98: I don't see the list at this url. Is there some other way to see the new list?

minhanh-phan: The only way to see the new list is to link to a specific run, but I don't think we should do that since there is a new run everyday, and I don't know how far back they keep the history

weshaggard (approved):

PR #10550, opened 2025-05-06
Support `az bicep build` in New-TestResources

hallipr:
Some users already have `az bicep` installed. We should be able to use that if `bicep` cli isn't installed

### DIFF
--- a/eng/common/TestResources/TestResources-Helpers.ps1
+++ b/eng/common/TestResources/TestResources-Helpers.ps1
@@ -129,8 +129,30 @@ function MergeHashes([hashtable] $source, [psvariable] $dest) {
     }
 }
 
+function IsBicepInstalled() {
+    try {
+        bicep --version | Out-Null
+        return $LASTEXITCODE -eq 0
+    }
+    catch {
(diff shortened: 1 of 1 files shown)


eng/common/TestResources/TestResources-Helpers.ps1
```diff
@@ -129,8 +129,30 @@ function MergeHashes([hashtable] $source, [psvariable] $dest) {
...
+        return $LASTEXITCODE -eq 0
+    }
+    catch {
+        return $false
+    }
+}
+
+function IsAzCliBicepInstalled() {
+    try {
+        az bicep version | Out-Null
```
weshaggard, line 144: Is this extension installed by default with az?

weshaggard (approved):

eng/common/TestResources/TestResources-Helpers.ps1
hallipr: no. The version check is required to see if `az bicep install` has been called

weshaggard (approved):

PR #10586, opened 2025-05-09
Add dynamic output formats. Use factory for dynamic agents based on model opts

benbp:
Refactor agent service classes. Use factory for dynamic agents based on model opts.

For the response service/formatting piece, the idea is that instead of building a string for output, one can define a response class implementing BaseCommandResponse. Then one can inject the response service into the constructor and wrap their response with it. This allows us to format outputs based on CLI mode (supporting -o json or plain text) and MCP mode (where we should always return structured data).

Example:

```
var response = new MyResponseClass
{
  Message = "my output"
  FooBar = <any other properties you care about>
};
return responseService.Respond(response);  // outputs string
```

For plain text output/formatting, one can implement a ToPlainText() method on their response class, which could be as simple as return this.Message for example. See [azure-sdk-tools/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/DefaultCommandResponse.cs at 7ef2e88da9435a51d564ad9d3fef5690ff983476 · Azure/azure-sdk-tools](https://github.com/Azure/azure-sdk-tools/blob/7ef2e88da9435a51d564ad9d3fef5690ff983476/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/DefaultCommandResponse.cs) as a basic one to use.

The response format is determined at startup based on CLI options: [azure-sdk-tools/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs at 7ef2e88da9435a51d564ad9d3fef5690ff983476 · Azure/azure-sdk-tools](https://github.com/Azure/azure-sdk-tools/blob/7ef2e88da9435a51d564ad9d3fef5690ff983476/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs#L53-L75)


benbp:
Re-working this a bit more, closing so no one reviews. Will re-open.

PR #10591, opened 2025-05-09
[APIView Copilot] Reliability improvements

tjprescott:
This PR:
- Adds a timeout so that if a request just hangs, after 3 minutes it will timeout and retry
- On retry, will honor a "Retry-After" header, or apply exponential backoff if not provided
- Adds an `info.log` file alongside `error.log` so you can review all logs for diagnostic purposes


packages/python-packages/apiview-copilot/src/_diff.py
```diff
@@ -1,13 +1,12 @@
 import os
 import re
-import subprocess
-import tempfile
+import difflib
```
kristapratico, line 3: Oh cool didn't know this was part of the stdlib

praveenkuttappan (approved):

kristapratico (approved):

PR #10605, opened 2025-05-12
Update approve script to work with both sync PR types

weshaggard:
This updates the approval script to pull the sync PR data from the pipeline and then use that. This will enable us to use it for both eng/common and workflow syncs and even PRs that have both.

### DIFF
--- a/eng/scripts/Approve-Eng-Common-Sync-PRs.ps1
+++ b/eng/scripts/Approve-Eng-Common-Sync-PRs.ps1
@@ -3,7 +3,7 @@
 [CmdletBinding(SupportsShouldProcess = $true)]
 param(
   [Parameter(Mandatory = $true)]
-  [string] $engCommonSyncPRNumber
+  [string] $ToolsPRNumber
 )
 
 . "${PSScriptRoot}/../common/scripts/logging.ps1"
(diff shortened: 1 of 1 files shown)


eng/scripts/Approve-Eng-Common-Sync-PRs.ps1
```diff
@@ -16,66 +16,69 @@ if ($LASTEXITCODE -ne 0) {
...
 }
 
-$ghloggedInUser = (gh api user -q .login)
-$engCommonToolsBranch = gh pr view $engCommonSyncPRNumber -R Azure/azure-sdk-tools --json "headRefName" --jq ".headRefName"
+$ToolsRepo = "Azure/azure-sdk-tools"
 
-if (!$engCommonToolsBranch) {
-  Write-Error "Didn't find branch for PR $engCommonSyncPRNumber in Azure/azure-sdk-tools"
-  exit 1
+$ghloggedInUser = (gh api user -q .login)
```
benbp, line 21: Maybe we should set `$PSNativeCommandUseErrorActionPreference = $true` along with error action preference stop at the top level?

benbp: If we can't get logged in user do we not want to fail?

benbp (approved):

eng/scripts/Approve-Eng-Common-Sync-PRs.ps1
weshaggard: Maybe :)

I can try setting that with my next common change to ensure that the error codes work as expected.

PR #10657, opened 2025-05-15
Ensure that we log the actual `serviceDirectories` when tagging a PR build

scbedd:
for `go` especially the `ServiceDirectory` of each package doesn't necessarily align with the `sdk/<X>` pattern that we currently fill for `ChangedServices`.

Just updating this to use the generated package props Service Directories, which WILL reflect this reality.

FYI @benbp doing the needful from our convo.

### DIFF
--- a/eng/common/pipelines/templates/steps/save-package-properties.yml
+++ b/eng/common/pipelines/templates/steps/save-package-properties.yml
@@ -41,20 +41,6 @@ steps:
           pwsh: true
           workingDirectory: '${{ parameters.WorkingDirectory }}'
 
-      # When running in PR mode, we want the detected changed services to be attached to the build as tags.
-      # However, the public identity does not have the permissions to attach tags to the build.
-      # Instead, we will save the changed services to a file, attach it as an attachment for PiplineWitness to pick up and utilize.
-      - pwsh: |
-          $changedServices = (Get-Content -Path '${{ parameters.DiffDirectory }}/diff.json' -Raw | ConvertFrom-Json).ChangedServices
-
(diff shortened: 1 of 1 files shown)


weshaggard (approved):

benbp (approved):

benbp (approved):

eng/common/pipelines/templates/steps/save-package-properties.yml
```diff
@@ -41,11 +41,26 @@ steps:
...
+            -OutDirectory '${{ parameters.PackageInfoDirectory }}'
+          pwsh: true
+          workingDirectory: '${{ parameters.WorkingDirectory }}'
+
       # When running in PR mode, we want the detected changed services to be attached to the build as tags.
       # However, the public identity does not have the permissions to attach tags to the build.
       # Instead, we will save the changed services to a file, attach it as an attachment for PiplineWitness to pick up and utilize.
+      #
       - pwsh: |
-          $changedServices = (Get-Content -Path '${{ parameters.DiffDirectory }}/diff.json' -Raw | ConvertFrom-Json).ChangedServices
```
weshaggard, line 48: Do you know if the ChangedServices property is still needed in the diff.json file for any other scenarios now?

scbedd: Yeah, `js` does use `ChangedServices`. A couple `Analyze` pieces in other repos use `ChangedServices` as well. I'm torn on just updating the `diff.json` if the two arrays don't align.

PR #10667, opened 2025-05-15
Agentic Retreival for sub-queries generation

radhgupta:
The service now generates multiple subqueries based on the original issue’s title and body, retrieves relevant issues and documents for each subquery, and aggregates them with deduplication. This approach increases the diversity and relevance of retrieved context, improving the quality of answers and suggestions.
1. `GenerateSubqueriesAsync` : Generates Subqueries using RAG agent. The number of subqueries depends on the length of the original query.
2. For each subqueries and original query we call `SearchIssuesAsync` and `SearchDocumentsAsync` to retrieve relevant issues and documents.
3. We Deduplicate the resultant issue and documents as they are added.
4. We call `SendMessageQnaAsync` to generate the Open AI response.

### DIFF
--- a/tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
+++ b/tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
@@ -40,25 +40,63 @@ namespace IssueLabelerService
             double scoreThreshold = double.Parse(_config.ScoreThreshold);
             double solutionThreshold = double.Parse(_config.SolutionThreshold);
 
-            var issues = await _ragService.SearchIssuesAsync(issueIndexName, issueSemanticName, issueFieldName, query, top, scoreThreshold, labels);
-            var docs = await _ragService.SearchDocumentsAsync(documentIndexName, documentSemanticName, documentFieldName, query, top, scoreThreshold, labels);
- 
-            if (docs.Count == 0 && issues.Count == 0)
+            //Subquery generation configurations
+            int subqueryCount = query.Length < 50 ? 3 : query.Length < 150 ? 5 : 7;
+            string countString = subqueryCount.ToString();
+            var replacementSubqueriesPrompt = new Dictionary<string, string>
             {
-                throw new Exception($"Not enough relevant sources found for {issue.RepositoryName} using the Complete Triage model for issue #{issue.IssueNumber}. Documents: {docs.Count}, Issues: {issues.Count}.");
+                { "subqueryCount", countString },
+                { "query", query }
+            };
+
+            string subqueriesPrompt = AzureSdkIssueLabelerService.FormatTemplate(_config.SubqueriesGenerationPrompt, replacementSubqueriesPrompt, _logger);
+
+            //Step 1: Generate sub-queries
+            var subqueries = await _ragService.GenerateSubqueriesAsync(subqueriesPrompt, query, modelName);
+
+            //Step 2: Retrieve results for sub-queries
+            var aggregatedIssues = new List<Issue>();
+            var aggregatedDocs = new List<Document>();
+            var uniqueIssues = new HashSet<string>();
+            var uniqueDocs = new HashSet<string>();
+
+            var subqueryTasks = subqueries.Select(async subquery =>
+            {
+                var subqueryIssues = await _ragService.SearchIssuesAsync(issueIndexName, issueSemanticName, issueFieldName, subquery, top, scoreThreshold, labels);
+                var subqueryDocs = await _ragService.SearchDocumentsAsync(documentIndexName, documentSemanticName, documentFieldName, subquery, top, scoreThreshold, labels);
+                AddUniqueItems(subqueryIssues, uniqueIssues, aggregatedIssues, issue => issue.Id);
+                AddUniqueItems(subqueryDocs, uniqueDocs, aggregatedDocs, doc => doc.Url);
+                return (Issues: subqueryIssues, Docs: subqueryDocs);
+            });
+
+            var subqueryResults = await Task.WhenAll(subqueryTasks);
+
+            // Step 3: Retrieve results for the original query
+            var originalIssues = await _ragService.SearchIssuesAsync(issueIndexName, issueSemanticName, issueFieldName, query, top, scoreThreshold, labels);
+            var originalDocs = await _ragService.SearchDocumentsAsync(documentIndexName, documentSemanticName, documentFieldName, query, top, scoreThreshold, labels);
+
+
+            // Step 4: Deduplicate original query results as they are added
+            AddUniqueItems(originalIssues,uniqueIssues,aggregatedIssues, issue => issue.Id);
+            AddUniqueItems(originalDocs, uniqueDocs, aggregatedDocs, doc => doc.Url);
+            var allIssues = aggregatedIssues.ToList();
+            var allDocs = aggregatedDocs.ToList();
+
+            // Step 5: Check if there are any results
+            if (allDocs.Count == 0 && allIssues.Count == 0)
+            {
+                throw new Exception($"Not enough relevant sources found for {issue.RepositoryName} using the Complete Triage model for issue #{issue.IssueNumber}. Documents: {allDocs.Count}, Issues: {allIssues.Count}.");
             }
 
-            double highestScore = _ragService.GetHighestScore(issues, docs, issue.RepositoryName, issue.IssueNumber);
+            double highestScore = _ragService.GetHighestScore(allIssues, allDocs, issue.RepositoryName, issue.IssueNumber);
             bool solution = highestScore >= solutionThreshold;
 
             _logger.LogInformation($"Highest relevance score among the sources: {highestScore}");
 
-            // Format issues 
-            var printableIssues = string.Join("\n\n", issues.Select(issue =>
+            var printableIssues = string.Join("\n\n", allIssues.Select(issue =>
                 $"Title: {issue.Title}\nDescription: {issue.chunk}\nURL: {issue.Url}\nScore: {issue.Score}"));
 
-            // Format documents 
-            var printableDocs = string.Join("\n\n", docs.Select(doc =>
+            var printableDocs = string.Join("\n\n", allDocs.Select(doc =>
                 $"Content: {doc.chunk}\nURL: {doc.Url}\nScore: {doc.Score}"));
 
             var replacementsUserPrompt = new Dictionary<string, string>
@@ -110,10 +148,26 @@ namespace IssueLabelerService
 
             _logger.LogInformation($"Open AI Response for {issue.RepositoryName} using the Complete Triage model for issue #{issue.IssueNumber}.: \n{formatted_response}");
 
-            return new AnswerOutput {
-                Answer =  formatted_response, 
-                AnswerType = solution ? "solution" : "suggestion" 
+            return new AnswerOutput
+            {
+                Answer = formatted_response,
+                AnswerType = solution ? "solution" : "suggestion"
             };
+        }
+
+        private void AddUniqueItems<T, TKey>(
+        IEnumerable<T> items,
+        HashSet<TKey> uniqueSet,
+        List<T> aggregatedList,
+        Func<T, TKey> keySelector)
+        {
+            foreach (var item in items)
+            {
+                if (uniqueSet.Add(keySelector(item)))
+                {
+                    aggregatedList.Add(item);
+                }
+            }
         }
     }
 }
--- a/tools/issue-labeler/src/IssueLabelerService/TraigeRag.cs
+++ b/tools/issue-labeler/src/IssueLabelerService/TraigeRag.cs
@@ -225,6 +225,28 @@ namespace IssueLabelerService
 
             return null;
         }
+
+        public async Task<List<string>> GenerateSubqueriesAsync(string subqueriesPrompt, string query, string modelName)
+        {
+            _logger.LogInformation($"Starting subquery generation for query: {query}");
+            if (string.IsNullOrWhiteSpace(query))
+            {
+                _logger.LogWarning("Query is empty or null. Returning an empty list of subqueries.");
+                return new List<string>();
+            }
+            
+            _logger.LogInformation($"Instructions for subquery generation: {subqueriesPrompt}");
+
+            string response = await SendMessageQnaAsync(subqueriesPrompt, query, modelName);
+
+            var subqueries = response.Split('\n')
+                                    .Where(subquery => !string.IsNullOrWhiteSpace(subquery) && char.IsDigit(subquery.TrimStart()[0]))
+                                    .Select(subquery => subquery.Trim().Substring(2).Trim())
+                                    .ToList();
+
+            _logger.LogInformation($"Generated {subqueries.Count} subqueries: {string.Join(", ", subqueries)}");
+            return subqueries;
+        }
     }
 
     public class Issue
--- a/tools/issue-labeler/src/IssueLabelerService/Configuration/RepositoryConfiguration.cs
+++ b/tools/issue-labeler/src/IssueLabelerService/Configuration/RepositoryConfiguration.cs
@@ -58,6 +58,7 @@ namespace IssueLabelerService
         public string LabelUserPrompt => GetItem("LabelUserPrompt");
         public string LabelInstructions => GetItem("LabelInstructions");
         public string LabelNames => GetItem("LabelNames");
+        public string SubqueriesGenerationPrompt => GetItem("SubqueriesGenerationPrompt");
 
         public string GetItem(string name)
         {


tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -40,25 +40,63 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
+            var subqueryTasks = subqueries.Select(async subquery =>
+            {
+                var subqueryIssues = await _ragService.SearchIssuesAsync(issueIndexName, issueSemanticName, issueFieldName, subquery, top, scoreThreshold, labels);
+                var subqueryDocs = await _ragService.SearchDocumentsAsync(documentIndexName, documentSemanticName, documentFieldName, subquery, top, scoreThreshold, labels);
+                AddUniqueItems(subqueryIssues, uniqueIssues, aggregatedIssues, issue => issue.Id);
+                AddUniqueItems(subqueryDocs, uniqueDocs, aggregatedDocs, doc => doc.Url);
+                return (Issues: subqueryIssues, Docs: subqueryDocs);
+            });
+
+            var subqueryResults = await Task.WhenAll(subqueryTasks);
```
jeo02, line 72: This can be its own method. I'll leave it up to you whether you place it in `TriageRag` or in here

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -40,25 +40,63 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
             double scoreThreshold = double.Parse(_config.ScoreThreshold);
             double solutionThreshold = double.Parse(_config.SolutionThreshold);
 
-            var issues = await _ragService.SearchIssuesAsync(issueIndexName, issueSemanticName, issueFieldName, query, top, scoreThreshold, labels);
-            var docs = await _ragService.SearchDocumentsAsync(documentIndexName, documentSemanticName, documentFieldName, query, top, scoreThreshold, labels);
- 
-            if (docs.Count == 0 && issues.Count == 0)
+            //Subquery generation configurations
+            int subqueryCount = query.Length < 50 ? 3 : query.Length < 150 ? 5 : 7;
```
jeo02, line 44: For readability rather just do if else when its this nested.
You can also do something like divide by 15 with a max of 7 so we don't have these values hard coded but that's just me nitpicking. 
```suggestion
            const int charsPerSubquery = 15;
            const int maxSubqueries = 7;
            int subqueryCount = query.Length / querySubDivisions;
            subqueryCount = subqueryCount < maxSubqueries  ? subqueryCount : maxSubqueries;
            
```

jeo02 (commented):
GitHub won't ignore approval if i approve now and you make changes so LMK

tools/issue-labeler/src/IssueLabelerService/TraigeRag.cs
```diff
@@ -225,6 +225,28 @@ public string LabelsFilter(Dictionary<string, string> labels)
...
+                _logger.LogWarning("Query is empty or null. Returning an empty list of subqueries.");
+                return new List<string>();
+            }
+            
+            _logger.LogInformation($"Instructions for subquery generation: {subqueriesPrompt}");
+
+            string response = await SendMessageQnaAsync(subqueriesPrompt, query, modelName);
+
+            var subqueries = response.Split('\n')
+                                    .Where(subquery => !string.IsNullOrWhiteSpace(subquery) && char.IsDigit(subquery.TrimStart()[0]))
```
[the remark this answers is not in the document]
jsquire: You don't want to do it in this form, as you'll pay the `Trim` cost several times, each of which allocates a new string.   You'd want to do something like:

```csharp

var subqueries = response
    .Split('\n')
    .Select(line => line?.Trim())
    .Where((line => (!string.IsNullOrWhiteSpace(line)) && (line.Length >= 2)))
    .Select(line => line.Substring(2))
    .ToList();
```

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -208,8 +209,20 @@ private async Task RetrieveAndAggregateSubqueryResults(
                 { "PrintableDocs", printableDocs },
                 { "PrintableIssues", printableIssues }
             };
-            
-            return (printableIssues, printableDocs, replacementsUserPrompt);
+
+            return replacementsUserPrompt ;
+        }
+
+        class IssueIdComparer : IEqualityComparer<Issue>
```
jsquire, line 216:
```suggestion
        private class IssueIdComparer : IEqualityComparer<Issue>
```

We'll always want to use explicit visibility scoping, even though `private` is the default in this case.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -208,8 +209,20 @@ private async Task RetrieveAndAggregateSubqueryResults(
...
+        {
+            public bool Equals(Issue x, Issue y) => x.Id == y.Id;
+            public int GetHashCode(Issue obj) => obj.Id.GetHashCode();
+        }
+
+        class DocumentUrlComparer : IEqualityComparer<Document>
```
jsquire, line 222:
```suggestion
        private class DocumentUrlComparer : IEqualityComparer<Document>
```

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -175,22 +175,23 @@ private async Task RetrieveAndAggregateSubqueryResults(
...
             var subqueryTasks = subqueries.Select(async subquery =>
             {
                 var subqueryIssues = await _ragService.SearchIssuesAsync(issueIndexName, issueSemanticName, issueFieldName, subquery, top, scoreThreshold, labels);
                 var subqueryDocs = await _ragService.SearchDocumentsAsync(documentIndexName, documentSemanticName, documentFieldName, subquery, top, scoreThreshold, labels);
-                AddUniqueItems(subqueryIssues, uniqueIssues, aggregatedIssues, issue => issue.Id);
-                AddUniqueItems(subqueryDocs, uniqueDocs, aggregatedDocs, doc => doc.Url);
+                foreach (var subqueryIssue in subqueryIssues)
+                    uniqueIssues.Add(subqueryIssue);
+                foreach (var subqueryDoc in subqueryDocs)
+                    uniqueDocs.Add(subqueryDoc);
```
jsquire, line 189:
```suggestion
                foreach (var subqueryIssue in subqueryIssues)
                {
                    uniqueIssues.Add(subqueryIssue);
                }
                
                foreach (var subqueryDoc in subqueryDocs)
                {
                    uniqueDocs.Add(subqueryDoc);
                }
```

nit: Always use braces, even for single lines.  It helps add safety and eliminate a class of errors.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -37,37 +37,69 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
             // Query + Filtering configurations
             string query = $"{issue.Title} {issue.Body}";
             int top = int.Parse(_config.SourceCount);
```
jsquire, line 39:
```suggestion
            var query = $"{issue.Title} {issue.Body}";
            var top = int.Parse(_config.SourceCount);
```

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -37,37 +37,69 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
-            double scoreThreshold = double.Parse(_config.ScoreThreshold);
-            double solutionThreshold = double.Parse(_config.SolutionThreshold);
+            var scoreThreshold = double.Parse(_config.ScoreThreshold);
+            var solutionThreshold = double.Parse(_config.SolutionThreshold);
+
+            //Subquery generation configurations
+            int minSubqueries = 3;
+            int maxSubqueries = 7;
+            int charsPerSubquery = 30;
```
jsquire, line 46:
```suggestion
            const int MinSubqueries = 3;
            const int MaxSubqueries = 7;
            const int CharsPerSubquery = 30;
```

Let's also hoist these up to the class level, please.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -37,37 +37,69 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
+
+            // Step 4: Deduplicate original query results as they are added
+            foreach (var originalIssue in originalIssues)
+                uniqueIssues.Add(originalIssue);
+            foreach (var originalDoc in originalDocs)
+                uniqueDocs.Add(originalDoc);
+
+            // Step 5: Check if there are any results
+            var aggregatedIssues = uniqueIssues.ToList();
+            var aggregatedDocs = uniqueDocs.ToList();
```
jsquire, line 91: There's really no need to `ToList` these.   `HashSet<T>` implements `IEnumerable<T>` and has a `Count` property.   By doing a `ToList`, we're allocating a new set and copying data over.  We can avoid both costs.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -110,10 +142,87 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
-                Answer =  formatted_response, 
-                AnswerType = solution ? "solution" : "suggestion" 
+            return new AnswerOutput
+            {
+                Answer = formatted_response,
+                AnswerType = solution ? "solution" : "suggestion"
+            };
+        }
+
+        private void AddUniqueItems<T, TKey>(
```
jsquire, line 152: I think this is unused now.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -37,37 +37,69 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
+            // Step 3: Retrieve results for the original query
+            var originalIssues = await _ragService.SearchIssuesAsync(issueIndexName, issueSemanticName, issueFieldName, query, top, scoreThreshold, labels);
+            var originalDocs = await _ragService.SearchDocumentsAsync(documentIndexName, documentSemanticName, documentFieldName, query, top, scoreThreshold, labels);
+
+
+            // Step 4: Deduplicate original query results as they are added
+            foreach (var originalIssue in originalIssues)
+                uniqueIssues.Add(originalIssue);
+            foreach (var originalDoc in originalDocs)
+                uniqueDocs.Add(originalDoc);
```
jsquire, line 87:
```suggestion
            foreach (var originalIssue in originalIssues)
            {
                uniqueIssues.Add(originalIssue);
            }
            
            foreach (var originalDoc in originalDocs)
            {
                uniqueDocs.Add(originalDoc);
            }
```

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -110,10 +142,87 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
+                    uniqueIssues.Add(subqueryIssue);
+                foreach (var subqueryDoc in subqueryDocs)
+                    uniqueDocs.Add(subqueryDoc);
+            });
+            await Task.WhenAll(subqueryTasks);
+        }
+
+        private Dictionary<string, string> BuildUserPromptData(
+            List<Issue> allIssues,
+            List<Document> allDocs,
```
jsquire, line 196:
```suggestion
            IEnumerable<Issue> allIssues,
            IEnumerable<Document> allDocs,
```

We could also use `ICollection<T>` if we think there's a need to mutate, but I don't see us needing that.

tools/issue-labeler/src/IssueLabelerService/TraigeRag.cs
```diff
@@ -225,6 +225,30 @@ public string LabelsFilter(Dictionary<string, string> labels)
...
+            }
+            
+            _logger.LogInformation($"Instructions for subquery generation: {subqueriesPrompt}");
+
+            string response = await SendMessageQnaAsync(subqueriesPrompt, query, modelName);
+
+            var subqueries = response
+                .Split('\n')
+                .Select(line => line?.Trim())
+                .Where(line => !string.IsNullOrWhiteSpace(line) && line.Length >= 2 && char.IsDigit(line[0]))
```
jsquire, line 245:
```suggestion
                .Where(line => !string.IsNullOrEmpty(line) && line.Length >= 2 && char.IsDigit(line[0]))
```
Since we're trimming first, we'll never get whitespace, just empty.  This form is slightly more efficient.

tools/issue-labeler/src/IssueLabelerService/TraigeRag.cs
```diff
@@ -225,6 +225,30 @@ public string LabelsFilter(Dictionary<string, string> labels)
...
+                .Select(line => line.Substring(2).Trim())
```
jsquire, line 246:
```suggestion
                .Select(line => line.Substring(2).TrimEnd())
```
Very minor efficiency gain.   We've already done a trim, so we know there's no leading space.  There may be a trailing one, since we took a slice.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -37,37 +37,69 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
-            double scoreThreshold = double.Parse(_config.ScoreThreshold);
-            double solutionThreshold = double.Parse(_config.SolutionThreshold);
+            var scoreThreshold = double.Parse(_config.ScoreThreshold);
+            var solutionThreshold = double.Parse(_config.SolutionThreshold);
+
+            //Subquery generation configurations
+            int minSubqueries = 3;
+            int maxSubqueries = 7;
+            int charsPerSubquery = 30;
+            string subqueryCount = Math.Max(minSubqueries, Math.Min(maxSubqueries, query.Length / charsPerSubquery)).ToString();
```
ShivangiReja, line 47: We can simplify this logic by using 
```c#
var subqueryCount = Math.Clamp(query.Length / charsPer, min, max).ToString();
```

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -37,37 +37,69 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
+            };
 
-            var issues = await _ragService.SearchIssuesAsync(issueIndexName, issueSemanticName, issueFieldName, query, top, scoreThreshold, labels);
-            var docs = await _ragService.SearchDocumentsAsync(documentIndexName, documentSemanticName, documentFieldName, query, top, scoreThreshold, labels);
- 
-            if (docs.Count == 0 && issues.Count == 0)
+            var subqueriesPrompt = AzureSdkIssueLabelerService.FormatTemplate(_config.SubqueriesGenerationPrompt, replacementSubqueriesPrompt, _logger);
+
+            //Step 1: Generate sub-queries
+            var subqueries = await _ragService.GenerateSubqueriesAsync(subqueriesPrompt, query, modelName);
```
ShivangiReja, line 57: This logic is good but could be extracted into a dedicated private method for clarity and readability.

```c#
private async Task<IReadOnlyCollection<string>> GenerateSubqueriesAsync(
    string query,
    string modelName)
{
    var min = 3;
    var max = 7;
    const int charsPer = 30;
    var count = Math.Clamp(query.Length / charsPer, min, max).ToString();

    var replacements = new Dictionary<string, string>
    {
        ["subqueryCount"] = count,
        ["query"] = query
    };
    var prompt = AzureSdkIssueLabelerService.FormatTemplate(_config.SubqueriesGenerationPrompt, replacements, _logger);
    _logger.LogDebug("Generating {Count} subqueries for query: {Query}", count, query);

    return await _ragService.GenerateSubqueriesAsync(prompt, query, modelName);
}
```

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -37,37 +37,69 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
+                issueFieldName,
+                documentIndexName,
+                documentSemanticName,
+                documentFieldName,
+                top,
+                scoreThreshold,
+                labels,
+                uniqueIssues,
+                uniqueDocs
+            );
```
ShivangiReja, line 76: Consider reducing the number of parameters passed to RetrieveAndAggregateSubqueryResults. Many of these values (like issueIndexName, documentIndexName, semantic names, field names, etc.) are already available via _config. Pulling them inside the method would improve readability and reduce boilerplate when calling the method.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -37,37 +37,69 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
+                issueSemanticName,
+                issueFieldName,
+                documentIndexName,
+                documentSemanticName,
+                documentFieldName,
+                top,
+                scoreThreshold,
+                labels,
+                uniqueIssues,
+                uniqueDocs
```
ShivangiReja, line 75: Consider returning the unique issues and documents from RetrieveAndAggregateSubqueryResults instead of passing them as mutable sets. This would make the method signature cleaner.

```c#
private async Task<(HashSet<Issue>, HashSet<Document>)> RetrieveResultsAsync(
    IEnumerable<string> subqueries,
    int top,
    double threshold,
    Dictionary<string, string> labels)
```

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
ShivangiReja, line 87: Consider using your helper method AddUniqueItems here to keep deduplication consistent and reusable.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -37,37 +37,69 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
+            );
+
+            // Step 3: Retrieve results for the original query
+            var originalIssues = await _ragService.SearchIssuesAsync(issueIndexName, issueSemanticName, issueFieldName, query, top, scoreThreshold, labels);
+            var originalDocs = await _ragService.SearchDocumentsAsync(documentIndexName, documentSemanticName, documentFieldName, query, top, scoreThreshold, labels);
+
```
ShivangiReja, line 81: Consider extracting each step into a dedicated private method. This would significantly improve the readability and structure of the method. Something like:

```c#
// Step 1: Generate subqueries
var subqueries = await GenerateSubqueriesAsync(query, modelName);

// Step 2: Retrieve and dedupe subquery results
var (uniqueIssues, uniqueDocs) = await RetrieveResultsAsync(subqueries, top, scoreThreshold, labels);

// Step 3: Always include original query results
await AddOriginalQueryResultsAsync(query, top, scoreThreshold, labels, uniqueIssues, uniqueDocs);
```

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -18,70 +19,72 @@ public OpenAiAnswerService(ILogger<AnswerFactory> logger, RepositoryConfiguratio
...
-                instructions = _config.SolutionInstructions;
-                userPrompt = AzureSdkIssueLabelerService.FormatTemplate(_config.SolutionUserPrompt, replacementsUserPrompt, _logger);
+                instructions = _serviceConfiguration.SolutionInstructions;
+                userPrompt = AzureSdkIssueLabelerService.FormatTemplate(_serviceConfiguration.SolutionUserPrompt, replacementsUserPrompt, _logger);
             }
             else
             {
-                instructions = _config.SuggestionInstructions;
-                userPrompt = AzureSdkIssueLabelerService.FormatTemplate(_config.SuggestionUserPrompt, replacementsUserPrompt, _logger);
+                instructions = _serviceConfiguration.SuggestionInstructions;
```
jsquire, line 83:
```csharp
userPrompt = isSolution 
    ? _serviceConfiguration.SolutionUserPrompt, replacementsUserPrompt 
    : _serviceConfiguration.SuggestionUserPrompt, replacementsUserPrompt;

userPrompt = AzureSdkIssueLabelerService.FormatTemplate(userPrompt, replacementsUserPrompt, _logger);
```

Unless I'm misreading, the only difference in the if/else branches is the config parameter.   I think we can simplify a bit.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -110,10 +113,78 @@ public async Task<AnswerOutput> AnswerQuery(IssuePayload issue, Dictionary<strin
...
+                Answer = formatted_response,
+                AnswerType = solution ? "solution" : "suggestion"
             };
         }
+        private async Task RetrieveAndAggregateSubqueryResults(
+            IEnumerable<string> subqueries,
+            Dictionary<string, string> labels,
+            HashSet<Issue> uniqueIssues,
+            HashSet<Document> uniqueDocs
+        )
```
jsquire, line 127:
```suggestion
            HashSet<Document> uniqueDocs)
```

nit:  C# convention is generally to throw the closing parens on the same line as the last parameter.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -10,6 +10,7 @@ namespace IssueLabelerService
 {
     public class OpenAiAnswerService : IAnswerService
     {
+        private readonly ServiceConfiguration _serviceConfiguration;
```
jsquire, line 13:
```suggestion
        private readonly ServiceConfiguration ServiceConfiguration;
```

This one gets inconsistently applied, but the `_name` pattern indicates a class-level variable.   Since this is a member that is intended to be constant after the instance is created, we generally name these like we would a constant.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -18,102 +19,151 @@ public OpenAiAnswerService(ILogger<AnswerFactory> logger, RepositoryConfiguratio
...
+            if (uniqueIssues.Count == 0 && uniqueDocs.Count == 0)
             {
-                instructions = _config.SolutionInstructions;
-                userPrompt = AzureSdkIssueLabelerService.FormatTemplate(_config.SolutionUserPrompt, replacementsUserPrompt, _logger);
-            }
-            else
-            {
-                instructions = _config.SuggestionInstructions;
-                userPrompt = AzureSdkIssueLabelerService.FormatTemplate(_config.SuggestionUserPrompt, replacementsUserPrompt, _logger);
+                throw new Exception($"Not enough relevant sources found for {issue.RepositoryName} using the Complete Triage model for issue #{issue.IssueNumber}. Documents: {uniqueDocs.Count}, Issues: {uniqueIssues.Count}.");
```
jsquire, line 63: No need to change it here but throwing the base `Exception` is generally discouraged and considered an anti-pattern.   For real-world scenarios, we would want to find the most specific built-in error type that we could so that callers have a chance to `catch`, identify, and react to the error.

For our purposes here, this is good enough.   I just wanted to highlight since we're using this PR as an introduction to C# conventions.

tools/issue-labeler/src/IssueLabelerService/Answers/OpenAiAnswerService.cs
```diff
@@ -18,102 +19,151 @@ public OpenAiAnswerService(ILogger<AnswerFactory> logger, RepositoryConfiguratio
...
+            public bool Equals(Issue x, Issue y) => x.Id == y.Id;
+            public int GetHashCode(Issue obj) => obj.Id.GetHashCode();
+        }
+
+        private class DocumentUrlComparer : IEqualityComparer<Document>
+        {
+            public bool Equals(Document x, Document y) => x.Url == y.Url;
+            public int GetHashCode(Document obj) => obj.Url.GetHashCode();
+        }
+        private record ServiceConfiguration(
```
jsquire, line 152:
```suggestion

        private record ServiceConfiguration(
```

big nit:  just normalizing spacing.

tools/issue-labeler/src/IssueLabelerService/TraigeRag.cs
```diff
@@ -22,6 +22,10 @@ public class TriageRag
         private static SearchIndexClient s_searchIndexClient;
         private ILogger<TriageRag> _logger;
 
+        private const int MinSubqueries = 3;
```
jsquire, line 25: nit:  Can we move these to the first class members?   Generally, C# conventions call for ordering by scope/visibility:

- const (which is static)
- static readonly
- static

and each type by visibility

- private
- protected
- internal
- public

jsquire (approved):
Looks good!  Just a couple more C# discussion items and formatting nits.

jsquire:
Active workstream, but not current focus.   Tagging to prevent auto-closure.

PR #10675, opened 2025-05-16
API surface outline

chidozieononiwu:


tjprescott (approved):
Assuming this is the logic that implements what you showed me in chat, it looks good to me.

PR #10679, opened 2025-05-16
Performance improvements for Java's ApiView generator

alzimmermsft:
- Set `ParserConfiguration.languageLevel` to null to prevent `JavaParser` from validating the source JAR code when parsing files. This isn't needed as the code needs to be valid to be able to create a source JAR. This also allows for the removal of having both a Java 8 and Java 11 parser as nothing is being validated (previously with languageLevel = Java 8, processing a module-info file with it would throw an exception).
- Eagerly load the entire file into memory rather than letting JavaParser handle loading data. This will reduce how often data needs to be loaded from disk. Additionally, pre-process the file to change all `\r\n` line endings to `\n` line ending (needed for a later point).
- Change many calls on JavaParser types from `toString()` to other code. `toString()` on many types results in more expensive tree node traversal by JavaParser to create the string. In many cases we just need a simple value that can be accessed through APIs on the node plus some minor string concatenations (such `StringLiteralExpr` being `" + expr.getValue() + "` rather than spinning up node traversal and a type printer).
- Decomposed Javadoc processing to manually handle creating the values for ApiView. This removes many node traversals, string splits and cleanups, and even more string splits and cleanups in our code.
- Moved around some conditional expressions so that faster ones appear first.
- Changed the "No Locales" rule to use a case-insensitive `Pattern`. This aligns with how URLs are case-insensitive and also removes the need to lowercase huge numbers of strings.
- Changed "Upper case naming" to defer creation of the ID until it is known that the rule was broken, as ID creation can be expensive for some JavaParser types.
-


src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -615,7 +732,7 @@ private void visitDefinition(BodyDeclaration<?> definition, ReviewLine parentLin
         } else if (definition instanceof FieldDeclaration) {
             FieldDeclaration fieldDeclaration = (FieldDeclaration) definition;
             id = makeId(fieldDeclaration);
-            name = fieldDeclaration.toString();
+            name = fieldDeclaration.getVariable(0).getNameAsString();
```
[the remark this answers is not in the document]
alzimmermsft: Good catch, I did mean to revert this change but missed it.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/model/ReviewLine.java
```diff
@@ -182,15 +182,9 @@ public ReviewLine hideLine() {
...
-
-        if (lineId != null) {
-            jsonWriter.writeStringField(JSON_LINE_ID, lineId);
-        }
-
-        if (crossLanguageId != null) {
-            jsonWriter.writeStringField(JSON_CROSS_LANGUAGE_ID, crossLanguageId);
-        }
+        jsonWriter.writeStartObject()
+            .writeStringField(JSON_LINE_ID, lineId)
```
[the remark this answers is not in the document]
alzimmermsft: Incorrect, `azure-json` has built-in logic to skip writing JSON fields with null values.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/model/ReviewToken.java
```diff
@@ -141,15 +141,9 @@ public TokenKind getTokenKind() {
...
-
-        if (navigationDisplayName != null) {
-            jsonWriter.writeStringField("NavigationDisplayName", navigationDisplayName);
-        }
-
-        if (navigateToId != null) {
-            jsonWriter.writeStringField("NavigateToId", navigateToId);
-        }
+            .writeStringField(JSON_NAME_VALUE, value)
+            .writeStringField("NavigationDisplayName", navigationDisplayName)
```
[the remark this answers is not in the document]
alzimmermsft: Incorrect, `azure-json` has built-in logic to skip writing JSON fields with null values.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/diagnostics/rules/general/MissingEqualsAndHashCodeDiagnosticRule.java
```diff
@@ -67,16 +68,16 @@ private boolean hasValidHashCodeMethod(TypeDeclaration<?> typeDeclaration) {
 
     private boolean isValidEqualsMethod(MethodDeclaration method) {
         return method.getNameAsString().equals("equals")
-                && method.isAnnotationPresent("Override")
-                && method.getParameters().size() == 1
-                && method.getParameters().get(0).getType().asString().equals("Object")
-                && method.getType().asString().equals("boolean");
+            && method.getParameters().size() == 1
+            && method.isAnnotationPresent("Override")
```
srnagar, line 72: The order is switched because `isAnnotationPresent` check is more expensive than checking the size? If so, we should do this below as well in `isValidHashCodeMethod()`.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -252,6 +341,34 @@ private Optional<ScanElement> scanForTypes(Path path) {
         }
     }
 
+    private static String loadAndPreprocessFile(Path path) throws IOException {
```
srnagar, line 344: Add javadoc here - "Reads the file and replaces `\r\n` with `\n`"

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -1074,11 +1192,11 @@ private void visitExpression(Expression expression, ReviewLine definitionLine, b
 //            node.addToken(TEXT, expression.toString());
         } else if (expression instanceof NameExpr) {
             NameExpr nameExpr = (NameExpr) expression;
-            definitionLine.addToken(TYPE_NAME, nameExpr.toString(), Spacing.NO_SPACE);
+            definitionLine.addToken(TYPE_NAME, nameExpr.getName().getIdentifier(), Spacing.NO_SPACE);
```
srnagar, line 1195: Will this change the token signature? This can cause a lot of false positive diffs requiring approvals when there are no changes.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -1310,7 +1428,7 @@ private void visitClassType(Type type, ReviewLine reviewLine, TokenKind kind) {
         SpacingState spacingBefore = spacingState;
 
         if (type.isPrimitiveType()) {
-            reviewLine.addToken(kind, type.asPrimitiveType().toString(), spacingState.getSpacing());
+            reviewLine.addToken(kind, type.asPrimitiveType().getType().asString(), spacingState.getSpacing());
```
srnagar, line 1431: Same here - we should ensure the json output doesn't change

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
alzimmermsft: This shouldn't, as looking through the code path for `NameExpr.toString()` you get the following:

```java
    @Override
    public void visit(final NameExpr n, final Void arg) {
        printOrphanCommentsBeforeThisChildNode(n);
        printComment(n.getComment(), arg);
        n.getName().accept(this, arg);
        printOrphanCommentsEnding(n);
    }
```

By default, all comment printing is disabled so only `n.getName().accept(this, arg)` ends up running logic. And that calls

```java
    @Override
    public void visit(SimpleName n, Void arg) {
        printOrphanCommentsBeforeThisChildNode(n);
        printComment(n.getComment(), arg);
        printer.print(n.getIdentifier());
    }
```

And again, all comment printing is disabled so we're left with `printer.print(n.getIdentifier())` which should be the same as the change to `nameExpr.getName().getIdentifier()` in the change here.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
alzimmermsft: Same case as above.

srnagar (approved):

PR #10681, opened 2025-05-16
Specify run directory as repo root for mcp runner

benbp:
Specify run directory as repo root for mcp runner

### DIFF
--- a/eng/common/mcp/azure-sdk-mcp.ps1
+++ b/eng/common/mcp/azure-sdk-mcp.ps1
@@ -6,6 +6,7 @@ param(
     [string]$Version, # Default to latest
     [string]$InstallDirectory = (Join-Path $HOME ".azure-sdk-mcp" "azsdk"),
     [string]$Repository = 'Azure/azure-sdk-tools',
+    [string]$RunDirectory = (Resolve-Path (Join-Path $PSScriptRoot .. .. ..)),
     [switch]$Run,
     [switch]$UpdateVsCodeConfig,
     [switch]$Clean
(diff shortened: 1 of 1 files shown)


eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -6,6 +6,7 @@ param(
     [string]$Version, # Default to latest
     [string]$InstallDirectory = (Join-Path $HOME ".azure-sdk-mcp" "azsdk"),
     [string]$Repository = 'Azure/azure-sdk-tools',
+    [string]$RunDirectory = ($PSScriptRoot + "/../../../"),
```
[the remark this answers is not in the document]
scbedd: If you were to go with this I would recommend the much simpler `$RunDirectory = Join-Path -Path $PSScriptRoot ".." ".." ".."` 😆

weshaggard, line 9: FWIW https://github.com/Azure/azure-sdk-tools/blob/d008f93d062cea221e32f2e5d2ac5b4f7feb371b/eng/common/scripts/common.ps1#L3

That is how we define this in common.

weshaggard: Looks like we also resolve a similar path to the mcp.josn file below.

benbp: > FWIW
> 
> https://github.com/Azure/azure-sdk-tools/blob/d008f93d062cea221e32f2e5d2ac5b4f7feb371b/eng/common/scripts/common.ps1#L3
> 
> That is how we define this in common.

This is actually bugged, you need a leading slash. Hence why it's using 4 `..` instead of 3 (the first ends up getting ignored).

benbp: https://github.com/Azure/azure-sdk-tools/pull/10694

PR #10687, opened 2025-05-20
SDK documentation Validation automation project init framework to setup pipeline.

zedy-wj:
We are working on an SDK documentation validation automation project. The part related to the Python SDK documentation has been completed, so we are planning to migrate it to this repository.

Since this tool involves a relatively large amount of content, we need to ensure the migration process goes smoothly. To do this, we will create a pipeline to test the contents of the pull request.

**This PR mainly includes the entry point (`eng/pipelines/python/start.yml`) for setting up the pipeline**. Additionally, it contains some descriptor files to outline the overall structure of the project.

The origin repo of the project: https://github.com/zedy-wj/content-validation.

After this PR is merged, we will create a new PR that includes all the code from the tool to complete the full migration.

@xiangyan99, @msyyc for notification.


tools/content-validation/eng/pipelines/python/start.yml
```diff
@@ -0,0 +1,132 @@
+name: Run switched packages content validation for python
```
weshaggard, line 1: Is this specific to python?

I assume this pipeline is expected to be created in our DevOps org. Can you please name the yml file closer to what it is about? start.yml isn't very descriptive perhaps something like docs-content-validation.yml.

tools/content-validation/eng/pipelines/python/start.yml
```diff
@@ -0,0 +1,132 @@
...
+  
+        echo "##vso[task.setvariable variable=pipelineUrl;]$pipelineUrl"
+      displayName: Get the current pipeline url
+
+    - task: DotNetCoreCLI@2
+      displayName: Summary Total Issues of all packages
+      env:
+        PackageName: ${{ parameters.packages }}
+        PipelineRunLink: $(pipelineUrl)
+        GitHubPat: $(GITHUB_PAT)
```
weshaggard, line 99: WEe should define these as pipeline variables in this yml and then for the PAT we will hook up the variable group to the pipeline once we create the pipeline.

tools/content-validation/eng/pipelines/common/content-validation-for-python.yml
```diff
@@ -0,0 +1,191 @@
...
+      inputs:
+        command: custom
+        custom: 'tool '
+        arguments: install Microsoft.Playwright.CLI
+
+    - task: DotNetCoreCLI@2
+      displayName: Build all projects and libraries
+      inputs:
+        command: 'build'
+        projects: 'content-validation.sln'
```
weshaggard, line 67: Are you missing this code in the PR?

tools/content-validation/eng/pipelines/python/start.yml
zedy-wj: Yes, it is specific to python, in our current design, each language should have its own pipeline.

Updated to `python-sdk-docs-content-validation.yml`

tools/content-validation/eng/pipelines/python/start.yml
zedy-wj: The current code is used in our test pipeline. After you create the pipeline, we will adjust the code accordingly based on the location of the PAT.

zedy-wj:
@weshaggard - 
As you mentioned, the current PR is indeed incomplete. However, its primary purpose is to introduce an entry file `python-sdk-doc-content-validation.yml` for setting up a pipeline. We understand that this pipeline will not be runnable immediately, as the functional code will be provided in a follow-up PR.

The main goal at this stage is to establish the pipeline infrastructure so that future PRs, which will include the actual implementation, can be properly tested. We appreciate your understanding and are happy to adjust the approach if needed.

tools/content-validation/eng/pipelines/common/content-validation-for-python.yml
zedy-wj: These files are not currently included in this PR plan.

tools/content-validation/eng/pipelines/python/start.yml
```diff
@@ -0,0 +1,132 @@
...
+  - JavaScript
+  - DotNet
+  - Go
+
+stages:
+- stage: ContentValidationTesting
+  jobs:
+  - ${{ each packageName in split(parameters.packages, ',') }}:
+    - ${{ if ne(parameters.packages, 'all') }}:
+      - template: /eng/pipelines/python/package/${{ packageName }}.yml
```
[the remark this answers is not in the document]
zedy-wj: These files are not currently included in this PR plan.

weshaggard:
After chatting with @xiangyan99 you should first get the code into the repo and a ci.yml file which is a pipeline just to build and test that code. After that is all working then you can do a PR to add another pipeline which can be used to run the automation on some schedule. It does no good to add the broken pipeline into the repo currently.

zedy-wj:
Hi, @weshaggard - The prerequisite PR https://github.com/Azure/azure-sdk-tools/pull/10738 has been merged. In this PR, we will complete Step2: `adding the pipeline configuration`. Currently, two features of this tool are disabled:

1. **Automatic issue creation**: we need to discuss with @xiangyan99 where these issues should be created.
2. **Generating the status markdown file**: in our personal repository, we update the status markdown file in a dedicated branch. We're not sure if this approach is appropriate for this repository. @xiangyan99 Do you have any ideas?

tools/content-validation/eng/pipelines/dotnet/package/azure-ai-agents-persistent.yml
```diff
@@ -0,0 +1,31 @@
+
```
weshaggard: We should not have a pipeline per package. That is way too much overhead to maintain. We should have one pipeline that will validate all the packages.

tools/content-validation/eng/pipelines/common/content-validation-for-dotnet.yml
```diff
@@ -0,0 +1,151 @@
...
+- name: GitHubPat
+  type: string
+  default: ''
+- name: GitHubOwner
+  type: string
+  default: ''
+- name: GitHubRepo
+  type: string
+  default: ''
+- name: AzureDevOpsPat
```
weshaggard: What is this needed for?

tools/content-validation/eng/pipelines/dotnet/package/azure-ai-agents-persistent.yml
zedy-wj: I think you're right. Our current pipeline works in the same way — one pipeline handles the validation for all packages of a given language. You can check their format via https://dev.azure.com/v-wenjyu/content-validation-automation/_build/results?buildId=54&view=results. (I've tried to invite you; you can accept it via the email to gain access.)

As for the “per pipeline per package” approach you mentioned, I guess it's because you saw many YAML files for different packages. In reality, during execution, each package is assigned to a separate job to run, and those YAML files simply configure specific parameters for them. When creating pipelines, we only need to create four pipelines — one each for Python, Java, JS, and .NET. These correspond to the `eng/pipelines/language/xxx-sdk-docs-content-validation.yml` files.

tools/content-validation/eng/pipelines/common/content-validation-for-dotnet.yml
zedy-wj: We previously had a feature: when automatically creating GitHub issues, we would list only part of the errors in the issue body, and then append a link to the pipeline artifacts at the end. Through that link, users could download an Excel file to view all the issues. However, this feature required an Azure DevOps PAT, and our PAT tokens are only valid for seven days. During the testing phase, frequent updates were needed, so we disabled this feature.

Currently, we’ve adopted an alternative approach proposed by @xiangyan99: per package per rule one issue. With this, individual issue bodies are no longer too long and thus avoid errors.

@msyyc do you think the artifact link feature is still necessary? If not, we’ll remove all PAT-related parameters, since as of now, they don’t seem to be used.

msyyc: It is Ok for me to remove the artifact link if it is not used.

zedy-wj: Removed.

tools/content-validation/eng/pipelines/dotnet/package/azure-ai-agents-persistent.yml
weshaggard: From past experience I would suggest you make the package discovery more dynamic instead of needing to maintain a yml template for every package. I would also not do one job per package as that will quickly blow up our machine pool. 

Which yml file is the entry point for the pipeline?

I assume you are using your own devops for testing purposes but this will end up in the azure-sdk devops instance once it is ready?

zedy-wj: Yes, this will end up in the Azure SDK DevOps instance once it is ready. Currently, there is a separate pipeline for each of the four languages:

- eng/pipelines/dotnet/dotnet-sdk-docs-content-validation.yml

- eng/pipelines/dotnet/java-sdk-docs-content-validation.yml

- eng/pipelines/dotnet/javascript-sdk-docs-content-validation.yml

- eng/pipelines/dotnet/python-sdk-docs-content-validation.yml

Regarding the "one job per package" model — we are trying to switch to a model where one job validates all packages of a single language. One thing we’d like to point out is: although the previous approach had many jobs, they ran in parallel, and each pipeline typically finished within 2–3 hours. With the new design where a single job validates all packages, the total runtime of a pipeline could extend to 15–20 hours. If this runtime is acceptable, we will proceed with updating the code logic.

weshaggard: What is this tool doing that it takes 15-20 hours to validate doc files? That should be looked at more closely to see if that can be improved.

One pipeline per language seems reasonable but I would not expect this pipeline to run for more than 1-2 hours so please investigate why that is taking so long.

zedy-wj: Due to the large volume of data, taking `Python` as an example (which is actually one of the smaller ones among the four languages):

1. There are a total of 62 packages.
2. Across all packages, there are approximately 4,600 doc links that need to be validated.
3. Each doc link needs to be validated against 6 rules.

Based on these numbers, we can estimate that if a single job is used to test all packages for a language, it would result in around `4,600 × 6 = 27,600` test cases. This would take approximately 10 hours to complete.

In our current test framework, test cases are already executed in parallel. We've also observed during local runs that, on average, each test case only takes a few seconds to execute.

If we stick to the current "one job per package" model, the entire pipeline can finish in 1~3 hours, since different packages can also be validated concurrently.

weshaggard: What sort of checks are we doing for the links? I wouldn't expect it to take a few seconds per link that is pretty slow. Are you actually invoking each link? If so, we should try to cache that data because doing all that work every time shouldn't be necessary. For example, we have a daily pipeline https://github.com/Azure/azure-sdk-tools/blob/main/eng/pipelines/githubio-linkcheck.yml#L139 that will cache all the links that we visit from a number of our sites. You might be able to reuse that cache or create a similar cache in order to avoid a lot of this cost.

zedy-wj: Sorry, I may have expressed myself unclearly and caused some misunderstanding. When I mentioned the `4,600` "doc links" that need to be validated, I wasn’t referring to checking whether the links themselves are valid or accessible, but rather to validating the content within each linked page.

You might be missing a bit of context, so let me briefly explain the actual purpose of this tool. Using `Python` as an example: the tool is designed to validate all packages listed on this page: https://learn.microsoft.com/en-us/python/api/overview/azure/?view=azure-python. So, when I mentioned “doc links,” I meant that there are approximately 4,600 documentation pages across the Python SDK that require content validation.

For Python, we’ve defined `6` validation rules, which means that each documentation page (or doc link) will generate 6 test cases. Let me illustrate with one specific rule: say we are testing for the `“Missing Content”` issue in the Python SDK documentation, for each doc link, the tool will detect and extract all such issues (like the ones shown in the screenshot), and automatically create GitHub issues to track them. This constitutes a single test case.
image

Therefore, we believe that spending a few seconds per test case is still quite reasonable. And since there are around `27,600` such test cases for the Python SDK alone, that explains why the overall execution takes a relatively long time.

As for your other question: “Are you actually invoking each link?” — the answer is yes. Before running the tests, we first retrieve the list of doc links whose content needs to be validated. The caching suggestion you mentioned is a good one. In fact, we implemented a similar mechanism in a recent PR, where all the doc links are saved into a [`DataFolder`](https://github.com/Azure/azure-sdk-tools/tree/main/tools/content-validation/DataSource/DataFolder/python).

That said, we're currently unsure how frequently these links or their content might change, so we’ve been cautious about using the cached data as a full replacement for fresh data collection each time. Based on our testing, even with local caching in place, we would likely only save about 1 hour of runtime, for example, reducing total runtime from 15 hours to 14 hours. It doesn’t significantly change the overall testing duration.

weshaggard: Instead of pulling all these files via web crawling wouldn't it be more efficient to validate them from the docs repo directly? https://github.com/MicrosoftDocs/azure-docs-sdk-python

cc @danieljurek @xiangyan99

zedy-wj: The current tool mainly verifies the rendered page, and verifying from the source code would miss some rendering-related issues.

zedy-wj: @weshaggard 
We’ve removed a large portion of the configuration from `packages.yml` files and moved it into the `config.json` of the `DataSource` project, allowing the required parameters to be automatically retrieved at runtime based on mapping relationships.

On top of this, we introduced a `metrics mechanism`, enabling you to customize the maximum level of parallelism when running a pipeline. You can decide how many agents to use based on current resource availability. By default, each pipeline is configured with 2 agents.

Below are some performance results when using different numbers of agents across various pipelines. Since the number of packages/docs being validated differs by language (pipeline), the execution time does not scale linearly with the number of machines (of course, within each pipeline, efficiency will increase linearly as more agents are added.):

- Python: 2 agents — runtime ~5 hours

- Java: 8 agents — runtime ~2 hours

- JavaScript: 5 agents — runtime ~3 hours

- .NET: 5 agents — runtime ~6 hours

If you feel the default agent allocation for any language should be adjusted, or if you have any other suggestions, please feel free to let me know. Thank you!

tools/content-validation/eng/pipelines/python/run-all-packages.yml
```diff
@@ -0,0 +1,88 @@
+parameters:
+- name: packages
+  type: object
+  displayName: |
+    A list of package names to validate.
+  default:
+    - azure-ai-contentsafety
```
weshaggard: While reducing the packages to just a list here is a step in the right direction I think we can still do better. Can we figure out a way to dynamically find the list of packages so that we don't have to maintain this and we don't have to worry about missing any?

tools/content-validation/eng/pipelines/dotnet/package/azure-ai-agents-persistent.yml
weshaggard: This is moving in the right direction. Can we figure out a way to dynamically find the list of packages so we don't have to hard-code and keep that list updated?

tools/content-validation/eng/pipelines/python/run-all-packages.yml
zedy-wj: One way we can think of at the moment is to access this [page](https://learn.microsoft.com/en-us/python/api/overview/azure/?view=azure-python-preview) through playwright and get all the packages according to the service. However, there is a problem, not all packages in our test plan need to be tested, such as management-related packages, abandoned packages, and some packages that do not need to be tested. We need to filter these out. This means that even if we automatically get a list of packages that need to be tested, we need to configure the list of packages that need to be filtered. cc @xiangyan99 Any ideas?

weshaggard: What are the criteria for filtering out? Can it be checked in code without needing a hard-coded list?

zedy-wj: Due to changes in the test scope (the scope of “all” now refers to the packages released last month), we’ve updated the configuration here. It now uses dynamic fetching, and the hard-coded list has been removed.

zedy-wj:
@weshaggard - Sorry for the delay. We’ve removed a large number of configuration files and also replaced the hard-coded list with dynamically fetching the list of packages that need to be tested each month. Please re-review it, thanks!

tools/content-validation/eng/pipelines/python/python-sdk-docs-content-validation.yml
```diff
@@ -0,0 +1,229 @@
...
+            templateParameters = @{
+              packages    = $env:MATRIX_JSON
+              maxParallel = $env:MAX_PARALLEL
+            }
+          } | ConvertTo-Json -Depth 10
+
+          Write-Host ">> Triggering pipeline #$pipelineId in $proj with:"
+          Write-Host $body
+
+          Invoke-RestMethod `
```
weshaggard: You shouldn't need to trigger the pipeline again to get the dynamic matrix. You can dynamically create the matrix like you are and then it will define the set of jobs in the same pipeline run.

weshaggard: You can use https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/jobs/archetype-sdk-tests-generate.yml as an example of where we do this.

tools/content-validation/PendingTestingPackagesThisMonth/Program.cs
```diff
@@ -0,0 +1,262 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using System.Text.Json;
+
+namespace PendingTestingPackagesThisMonth
+{
+    public class PendingTestingPackagesThisMonth
+    {
+        private static readonly string RELEASE_PACKAGES_URL_PREFIX = "https://github.com/Azure/azure-sdk/tree/main/_data/releases/";
+        private static readonly string PENDING_TEST_PACKAGES_LIST_LOCATION = "../eng/pipelines/packages.json";
```
weshaggard: In general, we should try to avoid putting files in the source tree and use some temp directory.

tools/content-validation/eng/pipelines/dotnet/dotnet-sdk-docs-content-validation.yml
```diff
@@ -0,0 +1,229 @@
...
+        custom: 'tool '
+        arguments: run playwright install
+
+    - task: DotNetCoreCLI@2
+      displayName: Running project PendingTestingPackagesThisMonth
+      env:
+        Language: 'DotNet'
+      inputs:
+        command: 'run'
+        workingDirectory: '$(System.DefaultWorkingDirectory)/PendingTestingPackagesThisMonth'
```
weshaggard, line 84: I would suggest passing in the file path of the packages list so the tool can write to that path instead of hardcoding it in the tool and the pipeline.

tools/content-validation/PendingTestingPackagesThisMonth/Program.cs
```diff
@@ -0,0 +1,262 @@
...
+            var configPath = Path.Combine(Directory.GetCurrentDirectory(), FILTER_PACKAGES_FOR_PYTHON_PATH);
+
+            // Resolve filter packages in filter list.
+            if (File.Exists(configPath))
+            {
+                var configContent = await File.ReadAllTextAsync(configPath);
+                var configJson = JsonSerializer.Deserialize<List<Dictionary<string, List<string>>>>(configContent);
+                var packages = configJson?.FirstOrDefault()?["packages"] ?? new List<string>();
+
+                if (packages.Contains("azure-mgmt-*"))
```
weshaggard: Why are you doing filtering in code as well in the filter config files? I would expect to filter in one or the other.

tools/content-validation/eng/pipelines/python/python-sdk-docs-content-validation.yml
zedy-wj: Thank you very much for providing the example, which has been very helpful to us. We had previously attempted to use this approach but were not successful. We are pleased to inform you that it is now working properly.

tools/content-validation/PendingTestingPackagesThisMonth/Program.cs
zedy-wj: The logic here has been updated. We will now uniformly filter the packages with identifiable patterns (such as mgmt packages) directly in the code. The filter config files will still be kept to make it convenient for developers to add any packages they don’t want included in the test list later on.

zedy-wj:
@weshaggard - We have updated according to your comments, please re-review it, thanks.

weshaggard:
> @weshaggard - We have updated according to your comments, please re-review it, thanks.

I'm good with the changes once you get the pipeline passing.

zedy-wj:
@weshaggard - CI all checks have passed.

weshaggard (approved):

PR #10686, opened 2025-05-20
Remove the source code of mock-service-host

lirenhe:
1. The component is removed from language pipelines.
2. it targets swagger and caused a lot of CVE issues

Remove this component to clean up the source.

PR #10690, opened 2025-05-20
Update CODEOWNERS

weshaggard:
Code removed with https://github.com/Azure/azure-sdk-tools/pull/10686 so removing codeowner entry as well


weshaggard:
FYI @lirenhe

PR #10735, opened 2025-05-23
Add common npm publishing template

weshaggard:


eng/common/pipelines/templates/jobs/npm-publish.yml
```diff
@@ -0,0 +1,73 @@
...
+    isProduction: ${{ eq(parameters.Registry, 'https://registry.npmjs.org/') }}
+    inputs:
+    - input: pipelineArtifact
+      artifactName: ${{ parameters.ArtifactName }}
+      itemPattern: '**/*.tgz'
+      targetPath: $(ArtifactPath)
+
+  pool:
+    name: azsdk-pool
+    image: ubuntu-24.04
```
benbp, line 28: Given this will run in a context that includes the image.yml template, do we want to use those vars here, or have this be entirely self contained?

eng/common/pipelines/templates/jobs/npm-publish.yml
```diff
@@ -0,0 +1,73 @@
...
+              Usemanagedidentity: true
+              KeyVaultName: 'kv-azuresdk-codesign'
+              SignCertName: 'azure-sdk-esrp-release-certificate'
+              Intent: 'PackageDistribution'
+              ContentType: 'npm'
+              FolderLocation: $(ArtifactPath)
+              Owners: ${{ coalesce(variables['Build.RequestedForEmail'], 'azuresdk@microsoft.com') }}
+              Approvers: ${{ coalesce(variables['Build.RequestedForEmail'], 'azuresdk@microsoft.com') }}
+              ServiceEndpointUrl: 'https://api.esrp.microsoft.com'
+              MainPublisher: 'ESRPRELPACMANTEST'
```
benbp, line 57: Is this right?

eng/common/pipelines/templates/jobs/npm-publish.yml
weshaggard: I want it to be self-contained because I expect a number of repos, think autorest.* repos, to consume this directly as a template instead of pulling in all the files from the repos. However, you do make a good point we might be able to consume it via path from the tools repo. I'll think about that one.

eng/common/pipelines/templates/jobs/npm-publish.yml
weshaggard: I did not double check these parameters I used what was there. However, I will find where these are defined and see what is actually required.

PR #10802, opened 2025-06-03
Move Publish-To-Net template out of `azure-sdk-build-tools`

scbedd:
I will file a follow-up issue that will tackle updating `eng/common/scripts/create-tag-and-git-release` to have the features that are located in the build-tools create-tag-and-git-release. Given that's an eng/common update I will complete it in a follow-up.

### DIFF
--- a/tools/test-proxy/ci.yml
+++ b/tools/test-proxy/ci.yml
@@ -26,7 +26,6 @@ extends:
   parameters:
     ToolDirectory: tools/test-proxy
     ReleaseBinaries: true
-    Use1ESOfficial: true
     StandaloneExeMatrix:
     - rid: osx-arm64
       framework: net8.0
--- a/eng/pipelines/templates/stages/archetype-sdk-tool-dotnet.yml
+++ b/eng/pipelines/templates/stages/archetype-sdk-tool-dotnet.yml
@@ -40,7 +40,7 @@ parameters:
     default: true
   - name: Use1ESOfficial
     type: boolean
-    default: false
+    default: true
   - name: TestMatrix
     type: object
     default:
--- a/tools/azsdk-cli/ci.yml
+++ b/tools/azsdk-cli/ci.yml
@@ -26,7 +26,6 @@ extends:
   parameters:
     ToolDirectory: tools/azsdk-cli/Azure.Sdk.Tools.Cli
     ReleaseBinaries: true
-    Use1ESOfficial: true
     TestDirectory: tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests
     StandaloneExeMatrix:
     - rid: osx-x64


eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
```diff
@@ -0,0 +1,202 @@
...
+    pool:
+      name: ${{ parameters.SigningPool.name }}
+      ${{ if eq(parameters.OSName, 'macOS') }}:
+        vmImage: ${{ parameters.SigningPool.image }}
+      ${{ else }}:
+        image: ${{ parameters.SigningPool.image }}
+      os: ${{ parameters.SigningPool.os }}
+    steps:
+      - checkout: azure-sdk-build-tools
+        path: _azure-sdk-build-tools
```
benbp, line 154: So we need 2 copies?

scbedd: There is a single checkout for azure-sdk-build-tools, and one for azure-sdk-tools, which is `self`. I'm confused at the implication. The `signing` referenced scripts use azure-sdk-build-tools power shell, if you're wondering why we keep this one on the job above.

This PR will get one last commit that moves `post-files-to-git-release` TO the azure-sdk-tools directory. At which point I'll remove this checkout: azure-sdk-buildtools in favor of using the local script.

tools/test-proxy/ci.yml
```diff
@@ -30,18 +30,12 @@ extends:
     - rid: osx-x64
       framework: net8.0
       assembly: Azure.Sdk.Tools.TestProxy
-    - rid: osx-arm64
```
scbedd, line 33: This is just to allow easy testing and will be reverted.

eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
scbedd: Turns ou `publish-symbols` does need a local azure-sdk-build-tools script. So still leaving it, but left a comment as to why.

eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
```diff
@@ -0,0 +1,204 @@
...
+    os: windows
+
+# Needs Github Release variable group linked.
+# Needs AzureSDK Nuget Release Pipelines Secrets variable group linked.
+
+stages:
+- stage: Release
+  dependsOn: ${{ parameters.dependsOn }}
+  variables:
+    Artifacts: $(System.DefaultWorkingDirectory)/_artifacts/packages
```
benbp, line 25: Do we want these in `$(Build.ArtifactStagingDirectory)` or is there any benefit/drawback to having them alongside source?

scbedd: I don't believe there is any benefit. I'll fix this in a follow-up PR that I address the create-tags-and-git-release in.

eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
```diff
@@ -0,0 +1,204 @@
...
+      os: ${{ parameters.SigningPool.os }}
+
+    templateContext:
+      type: releaseJob
+      isProduction: false
+    strategy:
+      runOnce:
+        deploy:
+          steps:
+            - ${{ if eq(parameters.ShouldSign, 'true') }}:
```
weshaggard, line 114: It seems like we need to remove this condition as if ShouldSign is ever false we will not have any artifact downloaded.

eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
```diff
@@ -0,0 +1,204 @@
...
+              path: $(Binaries)
+        - template: ../steps/platform-specific-signing.yml
+          parameters:
+            BinariesPath: $(Binaries)
+            BuildScriptsPath: $(BuildToolScripts)
+            ExeMatrix: ${{ parameters.ExeMatrix }}
+            Use1ESTemplate: ${{ parameters.Use1ESTemplate }}
+
+        - task: PowerShell@2
+          displayName: 'Post to github release'
```
weshaggard, line 190: Do we do this for our normal sdk releases? I'm wondering if this is something we need any more or if we should at least make it optional.

weshaggard: Or is this just for the executables? Taking a closer look it at the indention it seems like it is.

PR #10834, opened 2025-06-09
MCP server changes to remove language constraints

praveenkuttappan:
Changes:
1. Added a check in MCP tool that links namespace approval to ensure the release plan is for management plane.
2. Removed list of languages constraint when updating SDK details and keep it as per the input which is generated from tspconfig.yaml config.
3. Modified namespace validation regex to compile time instead of run time.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlanTool/ReleasePlanTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlanTool/ReleasePlanTool.cs
@@ -10,7 +10,6 @@ using Azure.Sdk.Tools.Cli.Contract;
 using Azure.Sdk.Tools.Cli.Helpers;
 using Azure.Sdk.Tools.Cli.Models;
 using Azure.Sdk.Tools.Cli.Services;
-using Microsoft.Extensions.Primitives;
 using ModelContextProtocol.Server;
 using Octokit;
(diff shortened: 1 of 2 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlanTool/ReleasePlanTool.cs
```diff
@@ -10,15 +10,14 @@
...
-using Microsoft.Extensions.Primitives;
 using ModelContextProtocol.Server;
 using Octokit;
 
 namespace Azure.Sdk.Tools.Cli.Tools
 {
     [Description("Release Plan Tool type that contains tools to connect to Azure DevOps to get release plan work item")]
     [McpServerToolType]
-    public class ReleasePlanTool(IDevOpsService devOpsService, ITypeSpecHelper typeSpecHelper, ILogger<ReleasePlanTool> logger, IOutputService output, IUserHelper userHelper, IGitHubService githubService) : MCPTool
+    public partial class ReleasePlanTool(IDevOpsService devOpsService, ITypeSpecHelper typeSpecHelper, ILogger<ReleasePlanTool> logger, IOutputService output, IUserHelper userHelper, IGitHubService githubService) : MCPTool
```
scbedd, line 20: I may revert this `partial` when I split up release plantool into individual CLI actions, but we will see.

PR #10843, opened 2025-06-10
Rename methods to include 'Async' suffix

maririos:
This PR renames asynchronous methods to include the 'Async' suffix, following C# naming conventions.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlanTool/ReleasePlanTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlanTool/ReleasePlanTool.cs
@@ -51,7 +51,7 @@ namespace Azure.Sdk.Tools.Cli.Tools
 
 
         [McpServerTool, Description("Get release plan for API spec pull request. This tool should be used only if work item Id is unknown.")]
-        public async Task<string> GetReleasePlanForPullRequest(string pullRequestLink)
+        public async Task<string> GetReleasePlanForPullRequestAsync(string pullRequestLink)
         {
             try
             {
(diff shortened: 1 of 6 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlanTool/ReleasePlanTool.cs
```diff
@@ -51,12 +51,12 @@ public partial class ReleasePlanTool(IDevOpsService devOpsService, ITypeSpecHelp
 
 
         [McpServerTool, Description("Get release plan for API spec pull request. This tool should be used only if work item Id is unknown.")]
-        public async Task<string> GetReleasePlanForPullRequest(string pullRequestLink)
+        public async Task<string> GetReleasePlanForPullRequestAsync(string pullRequestLink)
```
praveenkuttappan: This method name is shown as MCP tool name and a change here will cause an issue if any prompt directly uses MCP tool (which I think we do). Can we avoid MCP tool name change in this PR? MCP tool attribute has name property to set a different name that method name as MCP tool name but I haven't tried that yet.

praveenkuttappan: This is true for all methods marked as `McpServerTool`

praveenkuttappan: `McpServerTool(Name = "GetReleasePlanForPullRequest")`

praveenkuttappan: `McpServerTool(Name = "GetReleasePlanForPullRequest")`

Issue #10821, opened 2025-06-05
[tsp-client] allow passing in additional emitter option of object


jeremymeng:
Currently there's no way to customize typescript emitter options of `object` type, specifically the `package-details` option which is an object with `name` and `version` and other properties.

When passing something like this to tsp-client

```bash
execCommand(
  `tsp-client update -d -o src/generated --emitter-options="generate-metadata=false;generate-test=false;package-details={\\"name\\":\\"@azure/eventgrid-namespaces\\",\\"version\\":\\"1.0.1\\"}"`,
);
```

an error is reported:

>/home/meng/git/jsmain/sdk/eventgrid/eventgrid-namespaces/src/generated/TempTypeSpecFiles/Azure.Messaging.EventGrid/tspconfig.yaml:51:5 - error invalid-schema: Schema violation: must be object (/package-details)

Looks that we only split and pass the string key/value pair:

https://github.com/Azure/azure-sdk-tools/blob/580e523566d1941e582bfa1e762f11c953add40b/tools/tsp-client/src/typespec.ts#L108

This is blocking TypeScript developers who wants to customize the package details via the command-line. Currently they either have to add the package details in the specs repo's tspconfig.yaml, which is a heavy process; or they have to manually update the generated code.

I found a work-around by changing the above code,

```ts
  if (key && value) {
    if (value.startsWith("{") && value.endsWith("}")) {
      emitterOverrideOptions[key] = JSON.parse(value);
    } else {
      emitterOverrideOptions[key] = value;
    }
  }

```

but it feels a bit hacky. Are there any better approach?

jeremymeng:
Any thoughts @catalinaperalta or anyone else?

catalinaperalta:
I think checking for the `{}` is a bit hacky, but also the fastest way to check if one of the options is an object...The other option would be to have a try/catch checking the results of json.parse, but it's a shame we'd have to call it for every value in the options since this is something only used in ts:
```
 try {
    const parsed = JSON.parse(str);
    return typeof parsed === 'object' && parsed !== null && !Array.isArray(parsed);
  } catch {
    return false;
  }
```

catalinaperalta:
You can open a PR with the contribution you shared above and I'm happy to review...we can add that into the tool for now to unblock ts folks

jeremymeng:
> The other option would be to have a try/catch checking the results of json.parse, but it's a shame we'd have to call it for every value in the options since this is something only used in ts:

[...]

PR #10851, opened 2025-06-10
[tsp-client] Allow emitter option of object type 

jeremymeng:
tsp support options of object type but currently tsp-client only pass key/value string pairs as additional emitter options to tsp.  This PR tries to parse the value string first. If the result is a non-null and non-array object, pass the parsed object; otherwise pass the original string.

Resolves https://github.com/Azure/azure-sdk-tools/issues/10821

### DIFF
--- a/tools/tsp-client/CHANGELOG.md
+++ b/tools/tsp-client/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release
 
-## 2025-06-12 - 0.22.0
+## 2025-06-13 - 0.22.0
 
 - Add support for emitter options of `object` type.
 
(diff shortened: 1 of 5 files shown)


jeremymeng:
> Do you mind adding an entry in the changelog as well?

@catalinaperalta Done. I added an entry with today's date.  Let me know if a different release date should be used.

tools/tsp-client/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release
 
+## 2025-06-12 - 0.22.0
```
catalinaperalta: Ok since I just saw this we'll probably release early tomorrow. Can you change this:
```suggestion
## 2025-06-13 - 0.22.0
```
Also to make things faster then can you also please update the package.json + lock file version so that we can release this tomorrow morning after merging this in?

jeremymeng: oh right I forgot to bump the version.  Done now.

PR #10913, opened 2025-06-17
Use Path.Combine for create directory

weshaggard:
Path.Combine will allow for fully qualified paths to override the combination. For example `Path.Combine("a","b","c:\test")` will resolve to `c:\test'. We have depended on such behavior in a few places like https://github.com/Azure/azure-sdk-for-python/blob/d66b5160f2fb9a3ca03e833a65b1c429993b2bc2/eng/pipelines/templates/steps/smoke-test-steps.yml#L99 so I'm reverting back to Path.Combine.

### DIFF
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -163,7 +163,7 @@ try {
     $root = $repositoryRoot = "$PSScriptRoot/../../.." | Resolve-Path
 
     if($ServiceDirectory) {
-        $root = "$repositoryRoot/sdk/$ServiceDirectory" | Resolve-Path
+        $root = [System.IO.Path]::Combine($repositoryRoot, "sdk", $ServiceDirectory) | Resolve-Path
     }
 
     if ($TestResourcesDirectory) {


weshaggard:
This will fix errors like https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4980447&view=logs&j=c6f9cd5b-5bdf-5b25-bbd1-1024e3d5d5f3&t=dc0e108f-9a61-5adb-6d1b-370d9ae40dc8

```
Resolve-Path: /mnt/vss/_work/1/s/eng/common/TestResources/New-TestResources.ps1:166
Line |
 166 |  …        $root = "$repositoryRoot/sdk/$ServiceDirectory" | Resolve-Path
     |                                                             ~~~~~~~~~~~~
     | Cannot find path
     | '/mnt/vss/_work/1/s/sdk/mnt/vss/_work/1/s/common/smoketest/' because it
     | does not exist.
```

benbp:
Seems to me like we would want to fail fast in these scenarios and make the caller fix the path they are passing in to not be absolute, if we are going to prefix it under the hood.

weshaggard:
@benbp I don't know how we would currently do the smoke-test example where I need to point it at another different directory. We could perhaps use the TestResourceDirectory parameter but that isn't currently exposed in our template. If you think that is a better approach I could move in that direction.

benbp:
I'm fine with this going in, but yeah it's probably a good idea for us to expose TestResourcesDirectory anyways

scbedd (approved):
I had the mentioned failure on my list of items to investigate this afternoon 👍 

Thanks for this Wes.

PR #10927, opened 2025-06-18
Convert azure-sdk-tools to use `ubuntu-24.04` instead of `20.04`

scbedd:
As well as swapping on to `azsdk-pool` wherever I made this update.

Manually invoked:

- [x] [Inventory](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4989177&view=results)
  - ~~Discovered a failure in `windows` invocation. `Get-WmiObject` is crashing on `windows-2022` for some reason.~~
  - ~~It's not assigning to a windows machine. Trying to understand why.~~
  - Solved the windows machine conflict, now the same `demands` methodology that works for `windows` isn't properly assigning `ubuntu 22.04` instead of `ubuntu 24.04`.
- [x] [Common Tests](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4989172&view=results)
- [x] [Test proxy invocation to check a standard build w/ signing etc on the new image version](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4989210&view=results) (running)

As well as running a test run for `test-proxy` that should exercise the new platform.

### DIFF
--- a/eng/pipelines/agent-software-inventory.yml
+++ b/eng/pipelines/agent-software-inventory.yml
@@ -106,7 +106,9 @@ jobs:
   - job: assemble
     dependsOn:
       - inventory_azure_pipelines
-      - inventory_custom_agents_linux
+      - inventory_custom_agents_linux_22_hardcoded
+      - inventory_custom_agents_linux_22_hardcoded_variable
+      - inventory_custom_agents_linux_22_strategy_variable
       - inventory_custom_agents_windows
     condition: succeededOrFailed()
     pool:
--- a/eng/pipelines/agent-software-inventory.yml
+++ b/eng/pipelines/agent-software-inventory.yml
@@ -52,16 +52,10 @@ jobs:
     - template: /eng/pipelines/templates/steps/inventory-steps.yml
 
   - job: inventory_custom_agents_windows
-    strategy:
-      matrix:
-        CustomWindows:
-          Pool: azsdk-pool
-          OSVmImage: windows-2022
-          Host: Windows
 
     pool:
-      name: $(Pool)
-      image: $(OSVmImage)
+      name: azsdk-pool
+      image: windows-2022
       os: windows
 
     steps:
--- a/eng/pipelines/agent-software-inventory.yml
+++ b/eng/pipelines/agent-software-inventory.yml
@@ -47,7 +47,6 @@ jobs:
 
     pool:
       name: $(Pool)
-      image: $(OSVmImage)
       demands: $(Demands)
       os: linux
 
--- a/eng/pipelines/agent-software-inventory.yml
+++ b/eng/pipelines/agent-software-inventory.yml
@@ -39,15 +39,15 @@ jobs:
           OSVmImage: ubuntu-22.04
           Demands: ImageOverride -equals ubuntu-22.04
           Host: Ubuntu
-        CustomLinux24:
-          Pool: azsdk-pool
-          OSVmImage: ubuntu-24.04
-          Demands: ImageOverride -equals ubuntu-24.04
-          Host: Ubuntu
+        # CustomLinux24:
+        #   Pool: azsdk-pool
+        #   OSVmImage: ubuntu-24.04
+        #   Demands: ImageOverride -equals ubuntu-24.04
+        #   Host: Ubuntu
 
     pool:
       name: $(Pool)
-      demands: $(Demands)
+      demands: ImageOverride -equals ubuntu-22.04
       os: linux
 
     steps:
--- a/eng/pipelines/agent-software-inventory.yml
+++ b/eng/pipelines/agent-software-inventory.yml
@@ -22,17 +22,17 @@ jobs:
         Pool: Azure Pipelines
         OSVmImage: macOS-latest
         Host: MacOS
-      CustomLinux20:
-        Pool: azsdk-pool-mms-ubuntu-2004-general
-        OSVmImage: MMSUbuntu20.04
-        Host: Ubuntu
       CustomLinux22:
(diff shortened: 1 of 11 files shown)


eng/pipelines/templates/variables/image.yml
```diff
@@ -2,20 +2,20 @@
...
+    value: azsdk-pool
   - name: WINDOWSPOOL
-    value: azsdk-pool-mms-win-2022-general
+    value: azsdk-pool
   - name: MACPOOL
     value: Azure Pipelines
 
   - name: LINUXVMIMAGE
-    value: azsdk-pool-mms-ubuntu-2004-1espt
+    value: ubuntu-24.04
```
scbedd, line 12: In the interests of updates, I figure we'll add ubuntu-2604 as linuxnextimage eventually. I am ok with backing it down to "standard is 22.04" but the rest of the language repos went to `2404` so I'm just maintaining that standard.

scbedd:
@benbp Hey ben you got any ideas why it refuses to assign `ubuntu 22.04` on this agent software inventory? I've tried what feels like every variation of 

`image` vs `vmimage`. I know it's related to not actually running a `1es` job which is why I'm running into difficulty.

With the `demands` update I was able to make `windows-2022` work, so it's just down to `ubuntu-2204` never getting mapped.

[Latest build run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4989433&view=results)

~~It's also possible that the pool doesn't _have_ 22.04 I'm checking on that next.~~

Nope: 
image

eng/pipelines/templates/variables/image.yml
weshaggard: I think you can just remove LINUXNECTVMIMAGE as it probably isn't used. I removed it in .NET

eng/pipelines/agent-software-inventory.yml
```diff
@@ -7,183 +7,92 @@ pr:
...
+    strategy:
+      matrix:
+        CustomLinux22:
+          Pool: azsdk-pool
+          OSVmImage: ubuntu-22.04
+          Demands: ImageOverride -equals ubuntu-22.04
+          Host: Ubuntu
+        CustomLinux24:
+          Pool: azsdk-pool
+          OSVmImage: ubuntu-24.04
```
weshaggard, line 43: I guess given the nature of this pipeline you are just hardcoding instead of using the image variables. I'm fine with that.

eng/pipelines/agent-software-inventory.yml
```diff
@@ -7,183 +7,92 @@ pr:
...
+          Demands: ImageOverride -equals ubuntu-24.04
+          Host: Ubuntu
+
+    pool:
+      name: $(Pool)
+      image: $(OSVmImage)
+      os: linux
```
weshaggard: Isn't this only interesting for 1es pipelines?

weshaggard (approved):

eng/pipelines/agent-software-inventory.yml
```diff
@@ -7,183 +7,92 @@ pr:
...
+          Host: Ubuntu
+        CustomLinux24:
+          Pool: azsdk-pool
+          OSVmImage: ubuntu-24.04
+          Demands: ImageOverride -equals ubuntu-24.04
+          Host: Ubuntu
+
+    pool:
+      name: $(Pool)
+      image: $(OSVmImage)
```
benbp: @scbedd re: your question, you would need `demands` here, right?

benbp:
```suggestion
      demands: $(Demands)
```

scbedd: I tried `demands`. My latest commit just went back to solely image. Links inc.

eng/pipelines/agent-software-inventory.yml
scbedd: I'm trying to make this thing select the right ubuntu version. Anything you see is just experimentation to make this work >.>.

eng/pipelines/agent-software-inventory.yml
scbedd: [This build](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4989433&view=results) was solely demands.

[the commit](https://github.com/Azure/azure-sdk-tools/commit/fc7d127011968debc22537ced712edd77a656b61)

scbedd:
image

The fully rendered yaml from the latest commit. Demands _should_ be coming through. I don't think it's a quotation issue because this is being honored just fine:

image

eng/pipelines/templates/variables/image.yml
scbedd: I'll do a ctrl-f and then remove it if it's unused.

scbedd:
The trick to make `demands` work from a matrix is to:

```
    strategy:
      matrix:
        CustomLinux22:
          Pool: azsdk-pool
          OSVmImage: ubuntu-22.04
          Host: Ubuntu

    pool:
      name: $(Pool)
      demands: ImageOverride -equals $(OSVmImage)
```

And having the demands hardcoded to `ImageOverride` is good enough to honor the variable.

Defining a `demands` in the pool config and then referencing solely that doesn't resolve. ie:

```
    strategy:
      matrix:
        CustomLinux22:
          Pool: azsdk-pool
          OSVmImage: ubuntu-22.04
          Demands: ImageOverride -equals ubuntu-22.04
          Host: Ubuntu

    pool:
      name: $(Pool)
      demands: $(Demands)
```

weshaggard (approved):

PR #10936, opened 2025-06-19
Add log analysis tool

benbp:
This adds a local file keyword search tool for finding errors, primarily from azure pipelines logs.

The simple approach, which I realized after adding a larger keyword list, is to scan for ANSI color codes to find stderr messages in azure pipelines logs. This doesn't necessarily always find errors, so there is a full search mode which may find more, but also has higher risk of false positives.

As a follow-on to this, I will add the functionality into the devops mcp tool.

If people are ok with this approach I'll add another commit to this PR with some testing and hardening for MCP mode.


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/LogAnalysisHelper.cs
```diff
@@ -0,0 +1,141 @@
...
+    private readonly ILogger<LogAnalysisHelper> logger = logger;
+
+    // Built-in error keywords for robust error detection
+    private static readonly HashSet<Keyword> defaultErrorKeywords =
+    [
+        // ANSI color codes (red, etc.)
+        "[31m",
+
+        // custom keyword comparers
+        new("error", (i) => {
```
weshaggard, line 49: Might need a similar setup for fail/ed

PR #10955, opened 2025-06-20
Remove "tsp-client compare"

tjprescott:
Removing this command as `rest-api-diff` is no longer being developed, so partners should not rely on this command.


catalinaperalta (commented):
Lgtm, do you mind adding a changelog entry and updating the package.json + lock version as well?

pshao25 (approved):
LGTM. Do we have any tests in specs repo that rely on tsp-client compare?

PR #10959, opened 2025-06-20
Remove python release/v3 branch

weshaggard:
cc @msyyc as it looks like you removed the release/v3 branch from python repo


msyyc:
> cc @msyyc as it looks like you removed the release/v3 branch from python repo

Yeah, release/v3 is for track1 SDK release which has been deprecated years ago so there is no need to keep the branch anymore. Thanks for the PR!

PR #10964, opened 2025-06-23
Support TestResourcesDirectory in deploy-test-resources.yml

hallipr:
Pass TestResourcesDirectory from deploy-test-resources.yml to New-TestResources.ps1


weshaggard (approved):

weshaggard:
Do remember that if you are doing multiple deployments in one job you will need to set an appropriate name so that the remove script knows which env variable to look for to clean-up.

PR #10990, opened 2025-06-24
Update codeowner linter so it works on ubuntu24 and when there is root global.json

weshaggard:
For this common pipeline we can rely on the version of .NET installed on the agent as we are only installing and running an existing tool. 

There are some repositories that have a global.json file at the root of the repository (tools and .NET) that specifies a version of .NET SDK that conflicts with this command so in order to workaround we are running the install command outside of the source directory.

### DIFF
--- a/eng/common/pipelines/codeowners-linter.yml
+++ b/eng/common/pipelines/codeowners-linter.yml
@@ -45,7 +45,7 @@ stages:
           useGlobalJson: true
           
       - pwsh: |
-          dotnet install --global --add-source "$(DotNetDevOpsFeed)" --version "$(CodeownersLinterVersion)" "Azure.Sdk.Tools.CodeownersLinter"
+          dotnet tool install --global --add-source "$(DotNetDevOpsFeed)" --version "$(CodeownersLinterVersion)" "Azure.Sdk.Tools.CodeownersLinter"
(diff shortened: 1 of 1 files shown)


eng/common/pipelines/codeowners-linter.yml
```diff
@@ -38,13 +38,16 @@ stages:
...
+        retryCountOnTaskFailure: 3
         inputs:
-          command: custom
-          custom: 'tool'
-          arguments: 'install --global --add-source "$(DotNetDevOpsFeed)" --version "$(CodeownersLinterVersion)" "Azure.Sdk.Tools.CodeownersLinter"'
-          workingDirectory: '$(Build.SourcesDirectory)/eng/common'
+          useGlobalJson: true
+          
+      - pwsh: |
+          dotnet install --global --add-source "$(DotNetDevOpsFeed)" --version "$(CodeownersLinterVersion)" "Azure.Sdk.Tools.CodeownersLinter"
```
weshaggard:
```suggestion
          dotnet tool install --global --add-source "$(DotNetDevOpsFeed)" --version "$(CodeownersLinterVersion)" "Azure.Sdk.Tools.CodeownersLinter"
```

PR #11004, opened 2025-06-25
[APIView Copilot] Add Summarization endpoint

tjprescott:


packages/python-packages/apiview-copilot/cli.py
```diff
@@ -350,6 +350,23 @@ def search_knowledge_base(
...
 
 
+def review_summarize(language: str, target: str, base: str = None):
+    """
+    Summarize an API or a diff of two APIs using the deployed API review service.
+    """
+    payload = {"language": language, "target": target}
+    if base:
+        payload["base"] = base
+    APP_NAME = os.getenv("AZURE_APP_NAME")
```
kristapratico, line 363: nit: Maybe use os.environ["AZURE_APP_NAME"] here for a better error message than the exception that will be raised on the next line if this isn't set

packages/python-packages/apiview-copilot/src/_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    }
+    return language_pretty_names.get(language, language.capitalize())
+
+
+def get_architects_for_language(language: str) -> list[str]:
+    """
+    Returns a list of architects for the given language. Pulled from:
+    https://github.com/Azure/azure-sdk/blob/main/.github/CODEOWNERS
+    """
+    language_architects = {
```
kristapratico: I don't think this is the most up-to-date

packages/python-packages/apiview-copilot/src/agent/plugins/_database_plugin.py
```diff
@@ -85,3 +96,13 @@ async def add_guideline_memory(
...
         memory_container.upsert_item(memory.model_dump())
         return {"status": "Memory added successfully"}
+
+
+def get_current_user() -> str:
+    """
+    Placeholder function to get the current user.
+    In a real application, this would retrieve the user from the authentication context.
+    """
+    # For demonstration purposes, we return a hardcoded user.
```
kristapratico: maybe throw in a TODO here too ?

packages/python-packages/apiview-copilot/src/_utils.py
tjprescott: I'm going to remove this from this PR. It's the beginning of a different PR entirely. Also I should probably pull these from an easily updateable list rather than have them buried in code.

packages/python-packages/apiview-copilot/src/agent/plugins/_database_plugin.py
tjprescott: Yeah, I'm gonna just remove it from this PR.

packages/python-packages/apiview-copilot/cli.py
tjprescott: Added a check and ValueError if not set.

packages/python-packages/apiview-copilot/src/_utils.py
tjprescott: Removed

packages/python-packages/apiview-copilot/src/agent/plugins/_database_plugin.py
tjprescott: Removed.

PR #11012, opened 2025-06-26
Build samples required by release

skywing918:


tools/js-sdk-release-tools/src/mlc/clientGenerator/modularClientPackageGenerator.ts
```diff
@@ -47,7 +46,7 @@ export async function generateAzureSDKPackage(options: ModularClientPackageOptio
         // TODO: to be compatible with current tool, input relative generated package dir
         const changelog = await generateChangelogAndBumpVersion(relativePackageDirToSdkRoot, options);
         updateChangelogResult(packageResult, changelog);
-        await tryBuildSamples(packageDirectory, rushxScript, options.sdkRepoRoot);
+        await tryBuildSamples(packageDirectory, rushxScript, options);
```
wanlwanl: why change to pass options, we don't need other properties

PR #11036, opened 2025-06-27
Add sdk details for release plan before sdk generation

smw-ms:
- Added validation login in the GenerateSDK method to check if package name and languages are properly configured in the release plan before proceeding with SDK generation


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -55,7 +55,41 @@ public class SpecWorkflowTool(IGitHubService githubService,
...
+        private async Task<GenericResponse> IsReleasePlanReadyToGenerateSDKAsync(int workItemId, string language)
+        {
+            var response = new GenericResponse()
+            {
+                Status = "Failed"
+            };
+
+            try
+            {
+                var releasePlan = await devopsService.GetReleasePlanAsync(workItemId);
```
praveenkuttappan, line 73: We should check and make sure work item Id is non-zero.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -55,7 +55,41 @@ public class SpecWorkflowTool(IGitHubService githubService,
...
+                var sdkInfo = releasePlan?.SDKInfo.FirstOrDefault(s => string.Equals(s.Language, language, StringComparison.OrdinalIgnoreCase));
```
praveenkuttappan: Add an explicit check here to check and return if SDKInfo list is empty with error "SDK details are not present in the release plan".

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -166,7 +200,7 @@ private async Task<GenericResponse> IsSpecReadyToGenerateSDKAsync(string typeSpe
         }
 
 
-        [McpServerTool, Description("This tool runs pipeline to generate SDK for a TypeSpec project. This tool calls IsSpecReadyForSDKGeneration to make sure Spec is ready to generate SDK.")]
+        [McpServerTool, Description("This tool runs pipeline to generate SDK for a TypeSpec project. This tool calls IsSpecReadyForSDKGeneration and IsReleasePlanReadyToGenerateSDKAsync make sure Spec and Release Plan is ready to generate SDK.")]
```
praveenkuttappan: You should remove `IsReleasePlanReadyToGenerateSDKAsync ` from this description. It's a private method so agent is not aware of that method.

smw-ms: fixed

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
smw-ms: fixed

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
smw-ms: done

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -55,7 +55,56 @@ public class SpecWorkflowTool(IGitHubService githubService,
...
+                    return response;
+                }
+                
+                var releasePlan = await devopsService.GetReleasePlanAsync(workItemId);
+
+                var sdkInfoList = releasePlan?.SDKInfo;
+
+                if (sdkInfoList == null || sdkInfoList.Count == 0)
+                {
+                    response.Details.Add($"SDK details are not present in the release plan.");
```
maririos, line 79: what action should the user take when this happens?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -55,7 +55,56 @@ public class SpecWorkflowTool(IGitHubService githubService,
         private readonly Option<int> releasePlanIdOpt = new(["--release-plan"], "SDK release plan id") { IsRequired = false };
         private readonly Option<int> workItemOptionalIdOpt = new(["--workitem-id"], "Release plan work item id") { IsRequired = false };
 
+        private async Task<GenericResponse> IsReleasePlanReadyToGenerateSDKAsync(int workItemId, string language)
```
maririos, line 58: Generic question, for all the messages added to `response.Details` is the expectation that with that the agent takes action? or the user?
My concern is that we are exposing to the user that there is a work item per Release Plan, but they don't know that and they shouldn't. They just interact with the Release plan. So if the message goes to the users, it might be better to use the release plan ID in the messages instead of the workitem id

praveenkuttappan: It's for the agent to take action and provide correct value back when invoking MCP tool Challenge of using release Plan ID in all MCP tool call is that it's not unique and agent has no idea which environment it needs to connect.  If a release plan is used, it is always checked in production. User can still provide a release plan Id in the prompt and agent find the corresponding work item and uses it in further calls. We don't need user to know the work item Id.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
praveenkuttappan: User does not have to. This response is for agent to find the SDK details and send correct SDK details. Understood your point.
@smw-ms  Can you please update this message to add a line to current message. `Update the SDK details using the information in tspconfig.yaml`

smw-ms: Addressed in this pr: https://github.com/Azure/azure-sdk-tools/pull/11096

PR #11053, opened 2025-06-30
[APIView Copilot] Add @metnion chat endpoint

### LINKED ISSUE #10943: [APIView Copilot] Add endpoint to handle @mentions
The agent needs an endpoint that can receive threads from APIView. If an "@mention" is from an architect and necessitates and update, the agent should clear the Auth Agent, process the update, and reply that its knowledge has been updated. If it fails auth or for whatever reason doesn't make an update, it should still respond.

tjprescott:
Closes #10943


packages/python-packages/apiview-copilot/tsp/main.tsp
```diff
@@ -60,6 +73,9 @@ namespace ApiViewCopilot {
     /** The line number of the comment */
     lineNo: int32;
 
+    /** The offset date and time of the comment */
+    timestamp: offsetDateTime;
```
tjprescott: @AlitzelMendez I think we need to be sending in the time associated with APIView comments, which we aren't currently doing. This will be necessary to ensure they are sequenced properly for context.

packages/python-packages/apiview-copilot/tsp/main.tsp
```diff
@@ -36,6 +36,19 @@ namespace ApiViewCopilot {
   @post
   op agentChat(@body request: AgentChatRequest): AgentChatResponse;
 
+  /** Handles an @mention from ApiView. Requires the bearer token to be set to the GH OAuth2 token. */
+  @route("/api-review/mention")
+  @post
+  op handleMention(
+    @body request: MentionRequest,
+    @header("Authorization") authorization: string,
```
tjprescott, line 44: This requires APIView to send in the GH OAuth2 token, which copilot will then need to authenticate. There are other auth strategies too we could consider (JWT for example).

packages/python-packages/apiview-copilot/tsp/main.tsp
AlitzelMendez: I think you also need the comment text, right? or are you pulling that in another way?

packages/python-packages/apiview-copilot/tsp/main.tsp
```diff
@@ -60,11 +73,23 @@ namespace ApiViewCopilot {
...
+
+    /** The creator of the comment */
+    createdBy: string;
 
     /** The contents of the comment */
-    comment: string;
+    commentText: string;
+
+    /** Whether the comment is marked resolved */
+    isResolved?: boolean;
```
tjprescott, line 101: @AlitzelMendez I added some additional fields we want from the APIView comments, and matched the names used on the C#. The only property you can't get directly is lineNo because that's not how they're stored on the APIView side.

AlitzelMendez: ok great! yes I already have some code for the lineNo so we are good! thanks Travis!

packages/python-packages/apiview-copilot/tsp/main.tsp
```diff
@@ -60,11 +73,23 @@ namespace ApiViewCopilot {
...
+    createdOn: offsetDateTime;
+
+    /** The count of upvotes */
+    upvotes?: int32;
+
+    /** The count of downvotes */
+    downvotes?: int32;
+
+    /** The creator of the comment */
+    createdBy: string;
```
AlitzelMendez, line 95: @tjprescott , this is currently `author` just double checking if changing this name will break anything on the `review` side, I don't know if you are already using this code :D

tjprescott: It won't break anything--I just changed it align with the names used in APIView already

packages/python-packages/apiview-copilot/tsp/main.tsp
tjprescott: Yes we do need that too. commentText should already be a property in the payload (previously called just comment)

packages/python-packages/apiview-copilot/src/agent/_agent.py
```diff
@@ -70,3 +89,83 @@ def create_kernel() -> Kernel:
...
+## CODE
+This {language} code is being discussed for the {package_name} package.
+```{language}
+{code}
+```
+## COMMENTS
+{ '\n\n'.join(comment_lines) }
+
+# INSTRUCTIONS
+- Focus on the comment that mentions @azure-sdk. That person is an Azure SDK architect, and you should assume their feedback is correct.
```
kristapratico, line 133: how does the agent know if the person is an architect?

tjprescott: For this PR it doesn't really matter because the agent only says what it _would do_. However, for this sprint overall, we are relying on APIView itself doing the validation that the person who made the `@mention` is an architect before calling the service. As a follow-up to that, copilot will need to validate the OAuth credentials sent into the request.

PR #11057, opened 2025-07-01
Add "@mention" capability to APIView

### LINKED ISSUE #10942: [APIView] Add "@mention" capability to APIView
This is necessary to organically trigger the feedback cycle on demand through APIView. When an `@azure-sdk` mention is made by an architect, it should trigger the feedback endpoint in the agent.

AlitzelMendez:
Closes #10942 

## UI Impact
`@azure-sdk` is in bold
image

## UX Story
image


src/dotnet/APIView/APIViewWeb/Repositories/CosmosReviewRepository.cs
```diff
@@ -68,6 +68,22 @@ public async Task<IEnumerable<ReviewListItemModel>> GetReviewsAsync(string langu
             return reviews;
         }
 
+        public async Task<string> GetLanguageReview(string reviewId)
```
praveenkuttappan: Why do you need this API? You can get review model using review Id and it has the language property?

src/dotnet/APIView/APIViewWeb/Models/AIReviewModel.cs
```diff
@@ -47,5 +48,7 @@ public class CommentModelForCopilot
         public string Author { get; set; }
         [JsonPropertyName("comment")]
         public string CommentText { get; set; }
+        [JsonPropertyName("timestamp")]
+        public DateTimeOffset Timestamp { get; set; }
```
praveenkuttappan: Can you change this property name to include the use of it? Is it created time stamp or last modified time stamp?

praveenkuttappan: I see above change. So it looks like `CreatedOn`

tjprescott: It should be the created timestamp so we can ensure comments are in their logical order.

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -148,6 +165,98 @@ public async Task<CommentItemModel> UpdateCommentAsync(ClaimsPrincipal user, str
...
+            catch
+            {
+                response = "An error occurred while attempting to contact the agent service.";
+            }
+
+            commentResult.CommentText = response;
+            await _commentsRepository.UpsertCommentAsync(commentResult);
+        }
+
+        private bool IsUserAllowedToChatWithAgent(ClaimsPrincipal name, string language)
```
praveenkuttappan: nit: ClaimsPrincipal param name should be renamed for consistency. I think it is user ( to not it's user claim everywhere)

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -148,6 +165,98 @@ public async Task<CommentItemModel> UpdateCommentAsync(ClaimsPrincipal user, str
...
+        }
+
+        private bool IsUserLanguageArchitect(ClaimsPrincipal user, string language)
+        {
+            string githubUser = user.GetGitHubLogin();
+
+            if (string.IsNullOrEmpty(githubUser) || string.IsNullOrEmpty(language))
+                return false;
+
+            string architects = _configuration[$"Architects:{language}"] ?? "";
```
praveenkuttappan: Are we changing the configuration now to put separate config for each language?

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -148,6 +165,98 @@ public async Task<CommentItemModel> UpdateCommentAsync(ClaimsPrincipal user, str
...
+                CreatedBy = ApiViewConstants.BotName,
+                CreatedOn = DateTime.UtcNow,
+                CommentType = CommentType.SampleRevision
+            };
+
+            if (!IsUserAllowedToChatWithAgent(user, language))
+            {
+                commentResult.CommentText =
+                    "Interaction with the agent is restricted. Only designated language architects are authorized to use this feature.";
+                await _commentsRepository.UpsertCommentAsync(commentResult);
```
praveenkuttappan: This will leave unnecessary trail of comments in the review if non-architects started creating a comment. Can you add a check at the beginning and fail the request with if user does not have permission to tag agent.

praveenkuttappan: This will leave unnecessary trail of comments in the review if non-architects started creating a comment. Can you add a check at the beginning and fail the request with if user does not have permission to tag agent.

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -76,6 +85,14 @@ public async void LoadTaggableUsers()
             TaggableUsers = new HashSet<GithubUser>(TaggableUsers.OrderBy(g => g.Login));
         }
 
+
+        private static readonly Regex tagsRegex = new(@"@([^\s]+)", RegexOptions.Compiled);
+        public bool IsApiViewAgentTagged(CommentItemModel comment)
```
praveenkuttappan: This can be a static method

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
AlitzelMendez: the architects are language specific so we will be saving them that way, but there is just one config that is Architects, and `:` it is just the wat to access them, the config looks this way 
image

src/dotnet/APIView/APIViewWeb/Repositories/CosmosReviewRepository.cs
AlitzelMendez: I didn't want to pull the whole review if one value was needed, but I can total call the other one

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
AlitzelMendez: duplicate message

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -148,6 +165,98 @@ public async Task<CommentItemModel> UpdateCommentAsync(ClaimsPrincipal user, str
...
+                CommentType = CommentType.SampleRevision
+            };
+
+            if (!IsUserAllowedToChatWithAgent(user, language))
+            {
+                commentResult.CommentText =
+                    "Interaction with the agent is restricted. Only designated language architects are authorized to use this feature.";
+                await _commentsRepository.UpsertCommentAsync(commentResult);
+                return;
+            }
```
tjprescott: I like this, though it won't absolve me ultimately of need to validate authorization through the agent.

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
chidozieononiwu: Use SignalR to push a notification to the client instead see https://github.com/Azure/azure-sdk-tools/blob/3149511d827f3ae68519d4edfdc473c5106b9679/src/dotnet/APIView/APIViewWeb/Controllers/CommentsController.cs#L35

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -148,6 +165,98 @@ public async Task<CommentItemModel> UpdateCommentAsync(ClaimsPrincipal user, str
...
+                        Author = threadComment.CreatedBy,
+                        CommentText = threadComment.CommentText,
+                        Timestamp = threadComment.CreatedOn,
+                        LineNumber = lineIdToIndex.TryGetValue(threadComment.ElementId, out var id) ? id : -1
+                    })
+                    .ToList();
+
+                var client = _httpClientFactory.CreateClient();
+                var payload = new Dictionary<string, object> { { "comments", commentsForAgent } };
+                response = await client.PostAsync(getCommentReplyEndpoint, new StringContent(JsonSerializer.Serialize(payload), Encoding.UTF8, "application/json"));
```
tjprescott: Send the authorization header as the Bearer token so I can validate it.

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
tjprescott: You can use the following endpoint in the APIView API to get this info:
GET `/api/Reviews/allowedApprovers`

src/dotnet/APIView/APIViewWeb/Models/AIReviewModel.cs
tjprescott: 👍 for `CreatedOn`. I will update the TypeSpec accordingly in my PR.

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
AlitzelMendez: that is not language specific, is that ok?

basically this endpoint is calling `_configuration["approvers"];` so it is all those that can see the "generate copilot review" button

tjprescott: @chidozieononiwu is there a different endpoint you were referring to?

tjprescott: This? The reviewId would basically encode the language.
GET `/api/Reviews/{reviewId}/preferredApprovers`

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -148,6 +165,98 @@ public async Task<CommentItemModel> UpdateCommentAsync(ClaimsPrincipal user, str
...
+                await _commentsRepository.UpsertCommentAsync(commentResult);
+                return;
+            }
+
+            string response;
+            try
+            {
+                /*
+                var copilotEndpoint = _configuration["CopilotServiceEndpoint"];
+                var getCommentReplyEndpoint = $"{copilotEndpoint}/agent/chat";
```
tjprescott: The endpoint here will be `{copilotEndpoint}/api-review/mention`.

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -76,6 +85,14 @@ public async void LoadTaggableUsers()
             TaggableUsers = new HashSet<GithubUser>(TaggableUsers.OrderBy(g => g.Login));
         }
 
+
+        private static readonly Regex tagsRegex = new(@"@([^\s]+)", RegexOptions.Compiled);
```
tjprescott: So one thing I don't think is addressed in this PR. When we render a comment, if there are `@mention`s we should display them in **bold** as that's the UX that most would expect. We don't need to do anything fancier than that for now.

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
AlitzelMendez: I don't think that is the endpoint, I was looking at the code and what it does is 
- get _configuration["approvers"]; (all the approvers)
- filter them base on your profile (do you remember last release that you were not able to see the button because you had specific some filter in here? it is this code)
image


@chidozieononiwu is there another endpoint? :)

chidozieononiwu: Shouldn't that be sufficient? Architect select the language they care about using that drop down.

praveenkuttappan: Correct. I don't prefer having separate configuration with duplicate info. If we are going to make a change in the config, then it should be changed language specific. But as Dozie mentioned, let's just use preferred approvers for now.
If a user selects a language and preferred review language and then we remove that user from main config, do we still return the user as preferred approver?

src/dotnet/APIView/APIViewWeb/Helpers/ApiViewConstants.cs
```diff
@@ -1,5 +1,6 @@
 namespace APIViewWeb.Helpers;
 
+
```
praveenkuttappan: nit: empty line

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsController.cs
```diff
@@ -132,13 +133,22 @@ public async Task<ActionResult> CreateCommentAsync(
                 CommentType = commentType
             };
 
+            bool isApiViewAgentTagged = _commentsManager.IsApiViewAgentTagged(comment, out string commentTextWithIdentifiedTags);
+            comment.CommentText = commentTextWithIdentifiedTags;
```
praveenkuttappan, line 137: Can you provide a sample Input comment and how it looks on UI after it's replaced by `IsApiViewAgentTagged`

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
tjprescott: Yes, I think for now, that endpoint is "good enough". We can tighten it down later if need be. Since the ability to trigger copilot in the first place is based on this logic, then it kind of makes sense that if you can trigger copilot, you can train it (definitely not long term, but for now!)

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsController.cs
AlitzelMendez: sure!

image

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
AlitzelMendez: I sent a dummy one, working on the real one in a future pr

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -148,6 +172,127 @@ public async Task<CommentItemModel> UpdateCommentAsync(ClaimsPrincipal user, str
...
+                string agentMentionEndPoint = $"{_configuration["CopilotServiceEndpoint"]}/api-review/mention";
+                var request = new HttpRequestMessage(HttpMethod.Post, agentMentionEndPoint)
+                {
+                    Content = new StringContent(
+                        System.Text.Json.JsonSerializer.Serialize(mentionRequest),
+                        Encoding.UTF8,
+                        "application/json")
+                };
+
+                //TODO: Set a real token here
```
tjprescott:
```suggestion
                //TODO: See: https://github.com/Azure/azure-sdk-tools/issues/11128
```

src/dotnet/APIView/APIViewWeb/Models/AgentChatResponse.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Text.Json.Serialization;
+
+namespace APIViewWeb.Models
+{
+    public class AgentChatResponse
+    {
+        [JsonPropertyName("response")]
+        public string Response { get; set; }
+
+        [JsonPropertyName("thread_id")]
```
tjprescott: You should be able to pass "threadId" here. If that doesn't work, let me know. You shouldn't have to send properies in "snake_case".

tjprescott (commented):
Overall LGTM, but I'll defer to @praveenkuttappan and @chidozieononiwu for approval since they are vastly more knowledgeable about the code than me.

src/dotnet/APIView/APIViewWeb/DTOs/SiteNotificationDto.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace APIViewWeb.DTOs;
+
+public class SiteNotificationDto
+{
+    public string ReviewId { get; set; }
+    public string RevisionId { get; set; }
+    public string Title { get; set; }
+    public string Summary { get; set; }
+    public string Message { get; set; }
+    public string Status { get; set; }
```
chidozieononiwu: Do you want to use SiteNotificationStatus here?

src/dotnet/APIView/APIViewWeb/DTOs/SiteNotificationDto.cs
```diff
@@ -0,0 +1,21 @@
+using APIViewWeb.Models;
+
+namespace APIViewWeb.DTOs;
+
+public class SiteNotificationDto
```
chidozieononiwu, line 6: You could potentially merge this with https://github.com/Azure/azure-sdk-tools/blob/b3a27ba626eab78ed0f644d2f1e96e0d9d7791b3/src/dotnet/APIView/APIViewWeb/Models/AIReviewModel.cs#L43
That way there is one object for sending notifications

src/dotnet/APIView/APIViewWeb/Models/AIReviewModel.cs
```diff
@@ -58,14 +59,22 @@ public class AIReviewJobCompletedModel
         public int NoOfGeneratedComment { get; set; }
     }
 
-    public class CommentModelForCopilot
+    public class ApiViewComment
```
chidozieononiwu: Should this have a name that indicates it is information going to the Copilot service.

src/dotnet/APIView/APIViewWeb/Models/AIReviewModel.cs
```diff
@@ -58,14 +59,22 @@ public class AIReviewJobCompletedModel
...
     {
-        [JsonPropertyName("line_no")]
+        [JsonPropertyName("lineNo")]
```
chidozieononiwu: Did the Copilot service get updated?

src/dotnet/APIView/ClientSPA/src/app/_dtos/siteNotificationDto.ts
```diff
@@ -0,0 +1,15 @@
+export enum SiteNotificationStatus {
+    Success = 'success',
+    Info = 'info',
+    Warning = 'warning',
+    Error = 'error'
+}
+
+export interface SiteNotificationDto {
```
chidozieononiwu, line 8: I think you can merge this with https://github.com/Azure/azure-sdk-tools/blob/a398ff6e867233f9f1564b3b1015cf4199a723cf/src/dotnet/APIView/ClientSPA/src/app/_dtos/aiReviewJobCompletedDto.ts#L3

src/dotnet/APIView/ClientSPA/src/app/_services/signal-r/signal-r.service.ts
```diff
@@ -16,6 +17,7 @@ export class SignalRService {
   private aiReviewUpdates: Subject<AIReviewJobCompletedDto> = new Subject<AIReviewJobCompletedDto>();
   private reviewUpdates: Subject<Review> = new Subject<Review>();
   private apiRevisionUpdates: Subject<APIRevision> = new Subject<APIRevision>();
+  private siteNotifications: Subject<SiteNotificationDto> = new Subject<SiteNotificationDto>();
```
chidozieononiwu, line 20: You could potentially use the same observable for `SiteNotificationDto` and `AIReviewJobCompletedDto` since they are both notifications, Perhaps a standard model for everything notifications.

src/dotnet/APIView/APIViewWeb/Models/AIReviewModel.cs
AlitzelMendez: we can change the name of the file so all the models in here are for the Agent, what do you think about that? I am trying to match the defined names on tsp contract for our API, so it is easier for us in the future to understand but pieces and how they communicate

src/dotnet/APIView/APIViewWeb/Models/AIReviewModel.cs
AlitzelMendez: yes :)

src/dotnet/APIView/APIViewWeb/Models/AgentChatResponse.cs
AlitzelMendez: this is a response and I need to deserialize, so I think in this case I need the "snake_case" because the json is in that format

src/dotnet/APIView/APIViewWeb/DTOs/SiteNotificationDto.cs
AlitzelMendez: I don't , as I need this to be a string, I just want to have a define "enum" value to avoid passing a random string, I can add a check.

src/dotnet/APIView/ClientSPA/src/app/_services/signal-r/signal-r.service.ts
AlitzelMendez: let me see how I can combine this too, I think there are some things that makes AIReviewJobCompletedDto very specifc for that scenario as everything is populated on the notification send part

I tried to do SiteNotificationDto none specific for future usage so I will migrate AIReviewJobCompletedDto to this, if we only want one observable

src/dotnet/APIView/APIViewWeb/Models/AIReviewModel.cs
chidozieononiwu: I think either way you go there would be confusion. From APIView perspective this is not an ApiViewComment at least it not the model being used.

src/dotnet/APIView/ClientSPA/src/app/_services/signal-r/signal-r.service.ts
AlitzelMendez: this is a little of a big change in the sense that AIReviewJobCompletedDto  is custom creating the notification afterward very specifically for reviews. 

I will move this logic to do it in advance to avoid doing post processing, but I need to have the host name, probably in a config, as I will not be able to get it from the page

as this involves significant change on the review side I prefer doing this refactoring on a separate pull request, if that is ok :)

src/dotnet/APIView/APIViewWeb/DTOs/SiteNotificationDto.cs
AlitzelMendez: I think I would need to replace this with the new notification system and pass the create the notification in advance

chidozieononiwu (approved):

PR #11066, opened 2025-07-01
MCP Prompt tools

### LINKED ISSUE #11042: Devex release agent: Add support to copy prompt files
Add a setup prompts tool in the MCP server that copies MCP prompt files from GitHub repo path configured in MCP server to .github/prompts directory. Tools should check if this step is required as initial step. This will allow us to easily support Devex release agent in all repos.

jeo02:
## Summary
Two new tools intended to allow users to download prompts for the Azure SDK MCP tools. First tool checks if the user's repository is missing any prompts. If prompts are missing, then we download them. Resolves #11042
MCP Prompt support is closely being looked at as they bring better UI support for IDEs. (main problem being it shows the user the whole prompt in chat for VS code)
### Download Prompts Tool
- Downloads prompt files from GitHub repositories (defaults to Azure/azure-rest-api-specs)
- Can download all files from a directory in a repository OR specific files from a list
- Automatically adds downloaded files to `.gitignore` to prevent accidental commits
### File Validation Tool
- Compares GitHub repo files to the local files at a certain path. (Only files not folder themselves)

One thing to note is I included default values but seems like agents tend to just ignore them and provide their own values.


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/PromptValidationTool/DownloadPromptsTool.cs
```diff
@@ -0,0 +1,297 @@
...
+            try
+            {
+                logger.LogInformation("Starting download from {sourceRepoOwner}/{sourceRepoName}/{sourcePath} to {destinationPath}", 
+                    sourceRepoOwner, sourceRepoName, sourcePath, destinationPath);
+
+                // Create destination directory if it doesn't exist
+                Directory.CreateDirectory(destinationPath);
+
+                // Step 1: Get the list of files to download
+                var filesToDownload = await GetFilesToDownloadAsync(sourceRepoOwner, sourceRepoName, sourcePath, missingFilesList);
```
praveenkuttappan, line 83: How do you handle the case of skipping the files if it already exists or are you force overwriting them?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/PromptValidationTool/DownloadPromptsTool.cs
```diff
@@ -0,0 +1,297 @@
...
+        public async Task<DownloadResponse> DownloadPrompts(
+            string sourceRepoOwner, 
+            string sourceRepoName, 
+            string sourcePath, 
+            string destinationPath,
+            string? missingFilesList = null)
+        {
+            try
+            {
+                logger.LogInformation("Starting download from {sourceRepoOwner}/{sourceRepoName}/{sourcePath} to {destinationPath}",
```
praveenkuttappan, line 76: Let's add a step in the beginning to check if currently working repo is same as the repo it's targeting. We don't want to download prompt files from azure-rest-api-specs to cloned path for azure-rest-api-specs. It should download only if current repo does not match with target repo.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/PromptValidationTool/DownloadPromptsTool.cs
```diff
@@ -0,0 +1,297 @@
...
+                var repoRoot = FindGitRepositoryRoot(destinationPath);
+                if (repoRoot == null)
+                {
+                    logger.LogWarning("Not in a git repository, skipping .gitignore update");
+                    return;
+                }
+
+                // Skip .gitignore update if we're in the azure-rest-api-specs repository
+                var repoName = Path.GetFileName(repoRoot);
+                if (string.Equals(repoName, "azure-rest-api-specs", StringComparison.OrdinalIgnoreCase))
```
praveenkuttappan: I think this will become irrelevant if we short circuit the call in the beginning if repo name in the request is same as repo wea redownloading from.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/PromptValidationTool/DownloadPromptsTool.cs
jeo02: Yea validation would technically only get the ones that don't exist but yes as it is it would overwrite if not given a list of files to download. Added this in so we don't override. [changes](https://github.com/Azure/azure-sdk-tools/pull/11066/files#diff-ef977187e718649249cdbb12d41720a72e187f78689846df02c8c7ab5a2e3292R179-R184)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/PromptValidationTool/DownloadPromptsTool.cs
jeo02: Added in the step [here](https://github.com/Azure/azure-sdk-tools/pull/11066/files#diff-ef977187e718649249cdbb12d41720a72e187f78689846df02c8c7ab5a2e3292R64-R75)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/PromptValidationTool/DownloadPromptsTool.cs
```diff
@@ -0,0 +1,308 @@
...
+            try
+            {
+                // Check if we are already in the source repository
+                var repoRoot = FindGitRepositoryRoot(destinationPath);
+                if (string.Equals(Path.GetFileName(repoRoot), sourceRepoName, StringComparison.OrdinalIgnoreCase))
+                {
+                    return new DownloadResponse
+                    {
+                        Message = $"We are in source repository no need to download files. {sourceRepoName} = {repoRoot}",
+                        Success = false,
```
praveenkuttappan: We should return true to avoid confusing LLM in this case. right? We are saying effectively nothing to do and all look great.

praveenkuttappan (approved):
Looks fine to me except one nit change to return response as true when we don't want to update anything.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/PromptValidationTool/DownloadPromptsTool.cs
jeo02: oh yes true!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/GitHubService.cs
```diff
@@ -274,5 +275,31 @@ public async Task<Issue> GetIssueAsync(string repoOwner, string repoName, int is
...
             return await gitHubClient.Issue.Get(repoOwner, repoName, issueNumber);
         }
+
+        /// <summary>
+        /// Helper method to get contents from a GitHub repository path.
+        /// </summary>
+        /// <param name="owner">Repository owner</param>
+        /// <param name="repoName">Repository name</param>
+        /// <param name="path">Directory or file path</param>
+        /// <param name="expectSingleFile">If true, returns only the first file content; if false, returns all contents</param>
```
maririos, line 285: nit: looks like this parameter no longer exists

PR #11067, opened 2025-07-01
Add additional create matrix logging

weshaggard:
To help debug matrix filters and other generation issues add additional logging.

### DIFF
--- a/eng/common-tests/ci.yml
+++ b/eng/common-tests/ci.yml
@@ -29,5 +29,5 @@ extends:
     TargetDirectory: eng/common-tests
     TargetTags: UnitTest,IntegrationTest
     PreTestSteps:
-      - temppate: /eng/pipelines/templates/steps/install-dotnet.yml
+      - template: /eng/pipelines/templates/steps/install-dotnet.yml
       - template: /eng/common/pipelines/templates/steps/enable-long-path-support.yml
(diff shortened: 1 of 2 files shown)


eng/common-tests/ci.yml
```diff
@@ -29,4 +29,5 @@ extends:
     TargetDirectory: eng/common-tests
     TargetTags: UnitTest,IntegrationTest
     PreTestSteps:
+      - temppate: /eng/pipelines/templates/steps/install-dotnet.yml
```
weshaggard:
```suggestion
      - template: /eng/pipelines/templates/steps/install-dotnet.yml
```

weshaggard:
@benbp I'm really surprised you didn't already have some logging for this as it can be difficult to see all the values by going through the templates.

benbp:
> @benbp I'm really surprised you didn't already have some logging for this as it can be difficult to see all the values by going through the templates.

@weshaggard I've always relied on the `download full yaml` to see exactly what's passed to `Create-JobMatrix.ps1` as it is much faster than running a pipeline to get the values.

Looks like yaml download is also available after the fact now for pipeline runs:

image

weshaggard:
Well now it is also in the pipeline logs :)

PR #11080, opened 2025-07-02
copilot added code

l0lawrence:


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,529 @@
...
+using Azure.Sdk.Tools.Cli.Services;
+using ModelContextProtocol.Server;
+
+namespace Azure.Sdk.Tools.Cli.Tools.TspClientTool
+{
+    [Description("Tools for generating SDKs using tsp-client from TypeSpec projects")]
+    [McpServerToolType]
+    public class TspClientTool(ILogger<TspClientTool> logger, IOutputService output, ITypeSpecHelper typeSpecHelper) : MCPTool
+    {
+        private const string initCommandName = "init";
```
l0lawrence, line 22: I figured these were the most worth it tsp-client commands to have for now

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,529 @@
...
+            bool skipInstall = false,
+            string? emitterOptions = null,
+            string? emitterPackageJsonPath = null,
+            bool saveInputs = false,
+            bool debug = false)
+        {
+            try
+            {
+                // Validate tspconfig path exists
+                if (!File.Exists(tspConfigPath) && !tspConfigPath.StartsWith("http"))
```
l0lawrence: http?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,529 @@
...
+        private readonly Option<bool> noPromptOpt = new(["--no-prompt", "-y"], "Skip any interactive prompts") { };
+
+        // Init command options
+        private readonly Option<string> tspConfigOpt = new(["--tsp-config", "-c"], "Path to tspconfig.yaml") { IsRequired = true };
+        private readonly Option<bool> skipSyncAndGenerateOpt = new(["--skip-sync-and-generate"], "Skip syncing and generating the TypeSpec project") { };
+        private readonly Option<string> localSpecRepoOpt = new(["--local-spec-repo"], "Path to local repository with the TypeSpec project") { };
+        private readonly Option<bool> saveInputsOpt = new(["--save-inputs"], "Don't clean up the temp directory after generation") { };
+        private readonly Option<string> emitterOptionsOpt = new(["--emitter-options"], "The options to pass to the emitter") { };
+        private readonly Option<string> commitOpt = new(["--commit"], "Commit hash to be used") { };
+        private readonly Option<string> repoOpt = new(["--repo"], "Repository where the project is defined") { };
```
catalinaperalta, line 43: Wondering if we need to declare all of this? Seems like this tool will take a pretty heavy dependency on tsp-client and will need to track any future updates to the tool pretty closely. Is there any way to balance having the agent read/understand tsp-client docs and have these tools use the relevant flags that they need? (I say this as I start my review without looking at the specific command configs in this tool :P)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,529 @@
...
+        private readonly Option<bool> skipInstallOpt = new(["--skip-install"], "Skip installing dependencies") { };
```
catalinaperalta, line 44: For example enable this flag might also benefit from having the install-dependencies command then

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,529 @@
...
+            bool saveInputs = false,
+            bool debug = false)
+        {
+            try
+            {
+                // Validate tspconfig path exists
+                if (!File.Exists(tspConfigPath))
+                {
+                    // Check if it's a directory containing tspconfig.yaml
+                    var tspConfigInDir = Path.Combine(tspConfigPath, "tspconfig.yaml");
```
catalinaperalta: I believe this check already exists in tsp-client, let's not duplicate this logic

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,529 @@
...
+                        tspConfigPath = tspConfigInDir;
+                    }
+                    else if (Directory.Exists(tspConfigPath) && typeSpecHelper.IsValidTypeSpecProjectPath(tspConfigPath))
+                    {
+                        // Use the helper to validate the path
+                        logger.LogInformation("Valid TypeSpec project path detected: {Path}", tspConfigPath);
+                    }
+                    else
+                    {
+                        return new DefaultCommandResponse
```
catalinaperalta: Seems like this is performing validations that might not be necessary since tsp-client will also return a similar error response

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,529 @@
...
+                AddOptionalArg(args, "--local-spec-repo", localSpecRepo);
+                AddOptionalArg(args, "--commit", commit);
+                AddOptionalArg(args, "--repo", repo);
+                AddOptionalArg(args, "--emitter-options", emitterOptions);
+                AddOptionalArg(args, "--emitter-package-json-path", emitterPackageJsonPath);
+
+                if (skipSyncAndGenerate) args.Add("--skip-sync-and-generate");
+                if (skipInstall) args.Add("--skip-install");
+                if (saveInputs) args.Add("--save-inputs");
+                if (debug) args.Add("--debug");
```
catalinaperalta, line 175: How do folks set these options when using the mcp server?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,529 @@
...
+            
+            try
+            {
+                logger.LogInformation("Executing npx @azure-tools/typespec-client-generator-cli with args: {Args}", string.Join(" ", args));
+
+                // Check if npx and the package are available
+                if (!await IsTspClientAvailable())
+                {
+                    var commandDuration = (int)(DateTime.UtcNow - startTime).TotalMilliseconds;
+                    return (false, "npx or @azure-tools/typespec-client-generator-cli is not available. Please ensure Node.js and npm are installed.", commandDuration);
```
catalinaperalta: Do we do any setup checks to make sure folks have node and tsp-client in their environment?

catalinaperalta: I saw something below checking if tsp-client is installed, what's the relationship between that check and then running a tsp-client command?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
catalinaperalta: tsp-client supports local paths as well as links to the tspconfig.yamls in the remote repos

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
l0lawrence: it grabs the args from the context

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
l0lawrence: we do this would be used for each lang so might eb worth keeping?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
l0lawrence: since this is theoretically going to be used by every lang I thought it would be ebst to give everyone access to all the command options

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+                output.Output(new DefaultCommandResponse
+                {
+                    ResponseError = $"Error executing tsp-client command: {ex.Message}"
+                });
+            }
+
+            await Task.CompletedTask;
+        }
+
+        [McpServerTool(Name = "tsp-client-init"), Description("Initialize SDK project folder from tspconfig.yaml")]
```
praveenkuttappan, line 147: Description says, initialize SDK project. Is it better to rename tool name to make it more `azsdk.init-sdk-project`
My concern is that tsp-client-init is conveying it's use only for those who know they need to run tsp init to start

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+        public async Task<DefaultCommandResponse> InitializeProject(
+            string tspConfigPath,
+            string? outputDir = null,
+            bool skipSyncAndGenerate = false,
+            string? localSpecRepo = null,
+            string? commit = null,
+            string? repo = null,
+            bool skipInstall = false,
+            string? emitterOptions = null,
```
praveenkuttappan, line 156: What are the sample emitter options for agent to understand what needs to be passed?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+
+        [McpServerTool(Name = "tsp-client-init"), Description("Initialize SDK project folder from tspconfig.yaml")]
+        public async Task<DefaultCommandResponse> InitializeProject(
+            string tspConfigPath,
+            string? outputDir = null,
+            bool skipSyncAndGenerate = false,
+            string? localSpecRepo = null,
+            string? commit = null,
+            string? repo = null,
+            bool skipInstall = false,
```
praveenkuttappan, line 155: Is this Python specific?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+                var exceptionDuration = (int)(DateTime.UtcNow - startTime).TotalMilliseconds;
+                logger.LogError(ex, "Failed to execute npx @azure-tools/typespec-client-generator-cli command");
+                SetFailure(1);
+                return Task.FromResult((false, (string?)ex.Message, exceptionDuration));
+            }
+        }
+
+        private static string RunProcess(string command, string args, string workingDirectory)
+        {
+            var processInfo = new ProcessStartInfo
```
praveenkuttappan, line 438: @benbp I think we should add a process running helper to avoid this process running step in each tool. I will add a helper class.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+        }
+
+        [McpServerTool(Name = "tsp-client-init"), Description("Initialize SDK project folder from tspconfig.yaml")]
+        public async Task<DefaultCommandResponse> InitializeProject(
+            string tspConfigPath,
+            string? outputDir = null,
+            bool skipSyncAndGenerate = false,
+            string? localSpecRepo = null,
+            string? commit = null,
+            string? repo = null,
```
praveenkuttappan, line 154: Is this repo name or Repo root path?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+        public async Task<DefaultCommandResponse> GenerateClientLibrary(
+            string? outputDir = null,
+            string? emitterOptions = null,
+            bool saveInputs = false,
+            bool skipInstall = false,
+            bool debug = false)
+        {
+            try
+            {
+                var args = new List<string> { "generate" };
```
praveenkuttappan, line 250: Can we all tsp init from this call if it's not initialized? This tool will have to take additional param for tspconfig path.  So  user can simply run `tsp-client-generate` with tspconfig path and tool will check if it's initialized or note and run init if required.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+            {
+                logger.LogError(ex, "Error generating client library with tsp-client");
+                return new DefaultCommandResponse
+                {
+                    ResponseError = $"Error generating client library: {ex.Message}"
+                };
+            }
+        }
+
+        [McpServerTool(Name = "tsp-client-sync"), Description("Sync TypeSpec project specified in tsp-location.yaml")]
```
praveenkuttappan, line 277: Can we please add more in description to make it clear when this sync is required to LLM and someone new to tsp? Or is this a service partner is supposed to know already?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+            return await UpdateProject(outputDir, repo, commit, tspConfig, localSpecRepo, emitterOptions, saveInputs, skipInstall, debug);
+        }
+
+        private async Task<DefaultCommandResponse> ExecuteSyncCommand(InvocationContext ctx)
+        {
+            var outputDir = ctx.ParseResult.GetValueForOption(outputDirOpt);
+            var localSpecRepo = ctx.ParseResult.GetValueForOption(localSpecRepoOpt);
+            var debug = ctx.ParseResult.GetValueForOption(debugOpt);
+
+            var args = new List<string> { "sync" };
```
praveenkuttappan, line 348: Why not calling MCP tool method `SyncProject` here like other cli handler?

praveenkuttappan: It will help us when adding telemetry logic if all CLI handlers internally calls the MCP tool method.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+            }
+        }
+
+        [McpServerTool(Name = "tsp-client-update"), Description("Sync and generate from TypeSpec project")]
+        public async Task<DefaultCommandResponse> UpdateProject(
+            string? outputDir = null,
+            string? repo = null,
+            string? commit = null,
+            string? tspConfig = null,
+            string? localSpecRepo = null,
```
praveenkuttappan, line 201: Is there any validation whether these are valid tspconfig path and spec repo path or you expecting that it is handled by tsp client code?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
benbp: @praveenkuttappan was just merged yesterday so we can use it now: https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+    [McpServerToolType]
+    public class TspClientTool(ILogger<TspClientTool> logger, IOutputService output) : MCPTool
+    {
+        private const string initCommandName = "init";
+        private const string updateCommandName = "update";
+        private const string syncCommandName = "sync";
+        private const string generateCommandName = "generate";
+
+        public new CommandGroup[] CommandHierarchy { get; set; } = [
+            new CommandGroup("tsp-client", "TypeSpec client library generation using tsp-client")
```
benbp, line 28: Do we want to have `TspTool` (`tsp` command) and `TspClientTool` (`tsp-client` command) be under the same top level group `typespec` ? Or are they different enough that they should be at the root command group list?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspClientTool/TspClientTool.cs
```diff
@@ -0,0 +1,476 @@
...
+                    updateCommandName => await ExecuteUpdateCommand(ctx),
+                    syncCommandName => await ExecuteSyncCommand(ctx),
+                    generateCommandName => await ExecuteGenerateCommand(ctx),
+                    _ => new DefaultCommandResponse { ResponseError = $"Unknown command: {commandName}" }
+                };
+
+                ctx.ExitCode = ExitCode;
+                output.Output(response);
+            }
+            catch (Exception ex)
```
benbp, line 133: We shouldn't need the try/catch at this level since we enforce it at the tool/command level

l0lawrence:
closing

PR #11081, opened 2025-07-02
[apiview-copilot] update typescript exception list

jeremymeng:


packages/python-packages/apiview-copilot/metadata/typescript/filter.yaml
```diff
@@ -1,3 +1,13 @@
 exceptions: |
-  1. Comment on lack of documentation comments
-  2. Comment on the use of "."
+  1. DO NOT comment on lack of documentation comments
+  2. DO NOT Comment on the use of "."
+  3. DO NOT Comment on syntax errors
+  4. DO NOT Suggest renaming overload methods to distinguish one from another
+  5. DO NOT Suggest consolidating overloads into a single declaration
+  6. DO NOT Suggest consolidating the similar delete option interfaces into one
```
jeremymeng:
```suggestion
  6. DO NOT Suggest consolidating the similar Options interfaces into one
```

deyaaeldeen (commented):
Is it possible to adjust the prompt to put the the model on the right track instead?

tjprescott:
> Is it possible to adjust the prompt to put the model on the right track instead?

@deyaaeldeen yes, to an extent. If there are filters you would rather try that for, we can omit them from the filter. If training doesn't have the desired effect they can then be added to the filter.

tjprescott (approved):

Issue #11078, opened 2025-07-02
Devex agent: Allow running Azure SDK MCP server in more than one vs code instance in parallel


praveenkuttappan:
User has Azure SDK MCP server configured in more than one repo and it fails to start when attempting to start MCP server in repos parallelly because both instances by default uses localhost:5000 instead of dynamic port. We need a change in MCP server to use dynamic port instead of using 5000 always.

PR #11084, opened 2025-07-02
Start MCP server using available port instead of 5000

praveenkuttappan:
Azure SDK MCP server cannot be started on two vs code instances because both instances are trying to use port 5000. If user has Azure SDK MCP configured in more than one repo, then it's throwing exception when starting MCP server.

This change will use available port instead of 5000 to start MCP server.


praveenkuttappan:
Fixes https://github.com/Azure/azure-sdk-tools/issues/11078

jeo02 (approved):
LG

scbedd (approved):

PR #11117, opened 2025-07-07
[apiview-copilot] Update java filter list

samvaity:


packages/python-packages/apiview-copilot/metadata/java/filter.yaml
```diff
@@ -1,2 +1,4 @@
 exceptions: |
-  1. Comment on lack of documentation comments
+  1. Do not comment on lack of documentation comments
```
srnagar, line 2: We can remove this line as we want to have comments on lack of documentation.

tjprescott: This is a standard item in every tuning file. Copilot cannot comment on lines that have no elementId, and documentation, for the few languages that support it like Java, do not provide this.

PR #11142, opened 2025-07-09
Global versioning for azsdk MCP tool

jeo02:
## Summary
- Updated the azure-sdk-mcp script to retrieve version from [azsdk_version tag](https://github.com/Azure/azure-sdk-tools/releases/tag/azsdk_version).
- Now to update the version you need to change the description inside of the azsdk_version tag (will default to latest if invalid).

### DIFF
--- a/eng/common/mcp/azure-sdk-mcp.ps1
+++ b/eng/common/mcp/azure-sdk-mcp.ps1
@@ -21,6 +21,19 @@ if (-not $InstallDirectory)
 }
 . (Join-Path $PSScriptRoot '..' 'scripts' 'Helpers' 'AzSdkTool-Helpers.ps1')
 
+# If version is not provided, look for the tag <package>_Latest and set the version to the tag's description
+if (-not $Version) {
+    try {
+        $versionTag = "${Package}_Latest"
+        $specificReleaseUrl = "https://api.github.com/repos/$Repository/releases/tags/$versionTag"
+        $latestRelease = Invoke-RestMethod -Uri $specificReleaseUrl
+        $Version = $latestRelease.body
+    }
+    catch {
+        Write-Warning "Could not fetch version from ${Package}_Latest tag. Will run the latest version."
+    }
+}
+
 if ($Clean) {
     Clear-Directory -Path $InstallDirectory
 }
--- a/eng/common/mcp/azure-sdk-mcp.ps1
+++ b/eng/common/mcp/azure-sdk-mcp.ps1
@@ -21,19 +21,6 @@ if (-not $InstallDirectory)
 }
 . (Join-Path $PSScriptRoot '..' 'scripts' 'Helpers' 'AzSdkTool-Helpers.ps1')
 
-# If version is not provided, look for the tag <package>_version and set the version to the tag's description
-if (-not $Version) {
-    try {
-        $versionTag = "${Package}_version"
-        $specificReleaseUrl = "https://api.github.com/repos/$Repository/releases/tags/$versionTag"
-        $latestRelease = Invoke-RestMethod -Uri $specificReleaseUrl
-        $Version = $latestRelease.body
-    }
-    catch {
-        Write-Warning "Could not fetch version from ${Package}_version tag. Will run the latest version."
-    }
-}
-
 if ($Clean) {
     Clear-Directory -Path $InstallDirectory
 }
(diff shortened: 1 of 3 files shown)


eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -21,6 +21,19 @@ if (-not $InstallDirectory)
 }
 . (Join-Path $PSScriptRoot '..' 'scripts' 'Helpers' 'AzSdkTool-Helpers.ps1')
 
+# If version is not provided, look for the tag <package>_version and set the version to the tag's description
+if (-not $Version) {
+    try {
+        $versionTag = "${Package}_version"
+        $specificReleaseUrl = "https://api.github.com/repos/$Repository/releases/tags/$versionTag"
```
weshaggard: Do we need to call this explicitly? Can we not just add support for this in the common install function?

jeo02: Yea it can be placed in the install for sure let me make that change

eng/common/scripts/Helpers/AzSdkTool-Helpers.ps1
```diff
@@ -138,17 +138,30 @@ function Install-Standalone-Tool (
...
-        foreach ($release in $releases) {
-            if ($release.tag_name -like "$Package*") {
-                $tag = $release.tag_name
-                $Version = $release.tag_name -replace "${Package}_", ""
-                $found = $true
-                break
-            }
+        # Look for specific version tag first, formatted as <package>_version
+        try{
+            $versionTag = "${Package}_version"
```
weshaggard, line 144: Should we instead pass in the value `$Version = "version"` and let the logic below kick in?

jeo02: Yea that would be cleaner

weshaggard: We should be careful about storing a version number in plain text in the body here.

jeo02: What would be a better alternative?

weshaggard: I was hoping we could just have `azsdk_version` point exactly to `azsdk_<version>` and while we can make the sha the same it doesn't look like we can pull the release based on the alternative tag name. 

@scbedd @benbp what you think is a good way to enable us to pin a version without needing to update the version in all the repos consuming the mcp server? We could consider an aka.ms link or maybe a version text file we pull from tools main.

scbedd: > without needing to update the version in all the repos consuming the mcp server

I think my preferred if we HAD to go to a truly global version.

1. [raw link](https://raw.githubusercontent.com/Azure/azure-sdk-tools/refs/heads/main/eng/common/testproxy/target_version.txt) to `main` version file.(this is just the test-proxy version file to use an example of this)
2. Remove target version from the checked in `repo-specific` mcp locations: https://github.com/Azure/azure-rest-api-specs/blob/bd99a78116dcf58af8356838f10a86de2f8b1ed4/.vscode/mcp.json#L6
3. If a target version isn't provided on the call site, fall back to the release indicated `main` raw link.

This would allow for local modification of mcp.json if they truly wanted a specific version, while unblocking a global pinned version for everyone else.

weshaggard: @scbedd given your experience with test-proxy do you think it would be better to have a version file that we sync to all the repos or just one version file we pull from the tools repo to get the version, assuming folks didn't override it directly in emergencies?

scbedd: I've run into very few issues with the local repo pin. In my experience, the eng/common sync has the following primary issue + benefit:

- Pushing engy/common changes to all the repos can be annoying and/or time consuming. I'll grant I'm a bit more tolerant of this than most engys contributors.
- However forcing changes through `eng/common` ALSO functions as a `integration test` of whatever new version I'm pushing out. This gives a bunch of confidence that at least for standard playback scenarios, a new version of the test-proxy is compatible across all our repos. I've caught a couple regressions due to this.

I think as long as we have the local escape hatch on the startup script, and we have decent integration testing, I think the raw link to `main` target version would be fine.

weshaggard: Thanks for the feedback.

@jeo02 lets move to making a version text file that holds the latest global pinned version of the MCP server and then update these install scripts to read that and pass it along if the version isn't overridden. 

We should also have some tests that trigger on that version update to do some basic validation to ensure we don't break things when updating it.

PR #11168, opened 2025-07-10
[APIView Copilot] Enable Mention Agent action

### LINKED ISSUE #10943: [APIView Copilot] Add endpoint to handle @mentions

### LINKED ISSUE #11158: [APIView Copilot] Force Search Indexer run on DB update?
Currently, indexers are set to run daily, which means that when you add or update something in the database, the search index won't see it for up to 24 hours! Instead, we should probably have database updates manually trigger indexers to run, to ensure that the search index is up-to-date.

The use case is where you get a bad comment, teach it something, and want to verify it learned by re-running the review. This will not work unless you teach it, re-index, and then re-request review.

### LINKED ISSUE #11159: [APIView Copilot] MentionAgent doesn't reliably link items together
The MentionAgent is not reliably creating the links it needs to when invoked. From logs it seems like the reason is because it fails to create IDs that are valid in some cases. We either need to improve that, or break this into a multi-step process where it generates individual Memory and Example payloads and sends them into a function that will generate IDs and link everything together appropriately.

This is complicated by the fact that MentionAgent must be single-request. It cannot come back and ask for clarification or confirmation.

tjprescott:
Closes #10943.
Closes #11158
Closes #11159


tjprescott:
@Copilot

Issue #11211, opened 2025-07-15
[APIView] Summary comment no-longer appearing for Java/Swift


tjprescott:
This is a regression. There was a PR that forced the Copilot Summary comment to appear on line 1 of the APIView even if the token file for that review didn't have an elementId. Something recently broke that logic.

- [ ] Ensure summary appears on line 1 regardless of whether an elementId exists
- [ ] Add test to protect against future regression

PR #11217, opened 2025-07-15
[APIView]  Fix Summary comment no-longer appearing for Java/Swift 

AlitzelMendez:
## Local Validation 
image

## Issue
https://github.com/Azure/azure-sdk-tools/issues/11211

## Work
- Fixed the error by assigning the `FIRST_ROW` elementId to the summary comment
- Added tests in both ends, where the `FIRST_ROW` is added and when is readed

### DIFF
--- a/src/dotnet/APIView/APIViewUnitTests/CopilotPollingBackgroundHostedServiceTests.cs
+++ b/src/dotnet/APIView/APIViewUnitTests/CopilotPollingBackgroundHostedServiceTests.cs
@@ -104,29 +104,42 @@ namespace APIViewUnitTests
         }
 
         [Theory]
-        [InlineData("any")]
-        [InlineData("summary")]
-        public async Task ExecuteAsync_CommentTypes_FormatsCorrectly(string commentType)
+        [InlineData("summary", 1, "FIRST_ROW")]
+        [InlineData("summary", 2, "line-2")]
+        [InlineData("any", 2, "line-2")]
+        public async Task ExecuteAsync_CommentTypes_FormatsCorrectly(string commentType, int lineNo, string expectedElementId)
         {
-            var comment = commentType == "any"
-                ? CreateCommentWithSuggestions()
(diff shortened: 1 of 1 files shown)


tjprescott (commented):
This LGTM. My only question: if a token file _has_ an elementId on the first line (like Python), does this logic overwrite it, or is it just a default fallback?  Might but worth adding a test case for when line 1 has an element ID (and maybe you already do)

AlitzelMendez:
> This LGTM. My only question: if a token file _has_ an elementId on the first line (like Python), does this logic overwrite it, or is it just a default fallback? Might but worth adding a test case for when line 1 has an element ID (and maybe you already do)

that is a great point!!!!!!

src/dotnet/APIView/APIViewWeb/HostedServices/CopilotPollingBackgroundHostedService.cs
```diff
@@ -97,7 +97,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                                     CreatedOn = DateTime.UtcNow,
                                     ReviewId = jobInfo.APIRevision.ReviewId,
                                     APIRevisionId = jobInfo.APIRevision.Id,
-                                    ElementId = comment.Source == SummarySource ? CodeFileHelpers.FirstRowElementId : codeLine.lineId,
+                                    ElementId = codeLine.lineId ?? (comment.Source == SummarySource ? CodeFileHelpers.FirstRowElementId : null),
```
tjprescott, line 100: Sweet. Let's add a test case to make sure this special casing behavior is protected

AlitzelMendez: sure! I added couple of more test to include the scenario in where is null but not summary, summary not null, summary null and standard not summary not null :)

PR #11228, opened 2025-07-16
[azsdk-cli] Use chained token credential. Handle pipeline links for analysis.

benbp:
This PR removes the use of DefaultAzureCredential in favor of ChainedTokenCredential. For local user flows, we pass all the usual user credentials + interactive as a fallback. For github action flows, we pass the AzureCliCredential specifically because we will run it in the context of the azure task with an oidc login token.

This also updates the analyze pipeline command to take a pipeline link as a valid argument for analysis instead of just a build id.


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/AzurePipelines/PipelineAnalysisTool.cs
```diff
@@ -23,7 +22,7 @@
 namespace Azure.Sdk.Tools.Cli.Tools;
 
 [McpServerToolType, Description("Fetches data from an Azure Pipelines run.")]
-public class PipelineAnalysisTool : MCPTool
+public partial class PipelineAnalysisTool : MCPTool
```
praveenkuttappan: Is it required to be partial?

praveenkuttappan (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/AzurePipelines/PipelineAnalysisTool.cs
benbp: :D in no way is it supposed to be partial, copilot trying to sneak it in. Updated

Issue #11114, opened 2025-07-07
[JS] Expected accpet breaking change 2


v-jiaodi:
change log link: https://github.com/Azure/azure-sdk-for-js/pull/35110/files#diff-2b97b7548bc8a5761c87a511a2feb90b2ebdcfaa6f346c3b2f4df4005c1355a8R114

```
Removed Type Alias XXXResponse
```

wanlwanl:
@skywing918 for PR #11237, please ignore the breaking that refer to the ignored type if no other breakings in that type

Issue #11112, opened 2025-07-07
[JS] Expected accpet breaking change


v-jiaodi:
link: https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/avs/arm-avs/CHANGELOG.md

```
xxx no longer has parameter resumeFrom
xxx no longer has parameter $host
xxx no longer has parameter endpoint
```

wanlwanl:
@skywing918 for PR #11237 and #11443, please ignore the breaking that refer to the ignored type if no other breakings in that type

PR #11237, opened 2025-07-17
Enhance breaking change detection logic to ignore specific cases 

skywing918:
https://github.com/Azure/azure-sdk-tools/issues/11112
https://github.com/Azure/azure-sdk-tools/issues/11114


tools/js-sdk-release-tools/src/changelog/v2/DifferenceDetector.ts
```diff
@@ -72,16 +73,71 @@ export class DifferenceDetector {
...
+    if (k.endsWith('Result') && v.some(pair => pair.reasons === DiffReasons.Removed)) {
+      return true;
+    }
+
+    if (k.endsWith('Headers') && v.some(pair => pair.reasons === DiffReasons.Removed)) {
+      return true;
+    }
+
+    // Special case: if interface is ErrorAdditionalInfo and any pair has reasons == 2, ignore it
+    if (k === 'ErrorAdditionalInfo' && v.some(pair => pair.reasons === DiffReasons.TypeChanged)) {
```
wanlwanl: please remove this one, we need to create a new pr to ignore all `any` cases
add a TODO comment instead

tools/js-sdk-release-tools/src/changelog/v2/DifferenceDetector.ts
```diff
@@ -72,16 +73,71 @@ export class DifferenceDetector {
...
+      return true;
+    }    
+
+    // Check if any pair has a target name that should be ignored
+    const ignoreTargets = [
```
wanlwanl, line 141: @v-jiaodi are they comes from interface only?

tools/js-sdk-release-tools/src/changelog/v2/DifferenceDetector.ts
```diff
@@ -72,16 +73,71 @@ export class DifferenceDetector {
...
+      "resumeFrom",
+      "$host",
+      "endpoint"
+    ];
+    return v.some(pair => {
```
wanlwanl, line 146: it means if there is one of "resumeFrom", "$host",  "endpoint", this code will ignore diff of the interface.
This is not what we want. We want to ignore the removed diff of these properties, but if other properties have breaking changes, we should still report them

tools/js-sdk-release-tools/src/changelog/v2/DifferenceDetector.ts
```diff
@@ -72,16 +73,71 @@ export class DifferenceDetector {
...
+
+  private shouldIgnoreTypeAliasBreakingChange(v: DiffPair[], k: string): boolean {
+    if (k.endsWith('Response') && v.some(pair => pair.reasons === DiffReasons.Removed)) {
+      return true;
+    }    
+    return false;
+  }
+
+  private shouldIgnoreFunctionBreakingChange(v: DiffPair[], k: string): boolean {
+    if (k === 'getContinuationToken' && v.some(pair => pair.reasons === DiffReasons.Removed)) {
```
wanlwanl: return k === 'getContinuationToken' && v.some(pair => pair.reasons === DiffReasons.Removed;

tools/js-sdk-release-tools/src/changelog/v2/DifferenceDetector.ts
```diff
@@ -72,16 +73,71 @@ export class DifferenceDetector {
...
+      "endpoint"
+    ];
+    return v.some(pair => {
+      const name = pair.target?.name || pair.source?.name;
+      return name && ignoreTargets.includes(name);
+    });
+  }
+
+  private shouldIgnoreTypeAliasBreakingChange(v: DiffPair[], k: string): boolean {
+    if (k.endsWith('Response') && v.some(pair => pair.reasons === DiffReasons.Removed)) {
```
wanlwanl: return k.endsWith('Response') && v.some(pair => pair.reasons === DiffReasons.Removed)

tools/js-sdk-release-tools/src/changelog/v2/DifferenceDetector.ts
v-jiaodi: Class and type alias are also acceptable.

skywing918: > Class and type alias are also acceptable.

Currently, the type alias define like`export type DataProductStatus = "Active" | "Inactive" | "Pending"`
So how to apply this to type alias?

skywing918: Confirmed, type alias is not for this

tools/js-sdk-release-tools/src/test/changelog/changelogGenerator.test.ts
```diff
@@ -1896,3 +1862,231 @@ describe("Changelog reading", () => {
...
+            {
+                apiView: currentApiView,
+                sdkType: SDKType.ModularClient,
+            },
+        );
+        const items = getItemsByCategory(
+            changelogItems,
+            ChangelogItemCategory.ModelPropertyTypeChanged,
+        );
+        expect(items).toHaveLength(0);
```
wanlwanl: should be no changes in `changelogItems`, check no items in both breakingChange and newFeature

Issue #10983, opened 2025-06-24
[APIView] Allow external sites to access the APIVIew API


tjprescott:
APIView uses cookie authentication, but in order for Copilot and APIView to be able to communicate back and forth through their respective APIs, we need some kind of key authentication in place.

AlitzelMendez:
After some conversations, the plan moving forward is to use Managed Identity. The reasons are the security it provides and that this has been the standard in Microsoft. Managed Identity eliminates the need for storing credentials in code or configuration files, provides automatic credential rotation, and follows Microsoft's Zero Trust security model with built-in Azure AD integration.

Just to clarify, this means that we are going to support the current model (GitHub) and we are going to add to this support by adding Managed Identity.

In order to achieve this, we will leverage our existing app registrations for APIViewStaging and APIView, and we will add the necessary permissions to apiview-gpt (System assigned identity).

Although if the expectation is have bidirectional communication we can you the same App Registration for both

tjprescott:
I'm not sure this is simply a matter of managed identity. Managed Identity would allow access to the webapp _resource_, which we don't need or care about. APIVIew uses cookie authentication for its API and that needs to be addressed in order for copilot to make calls to its APIs.  cc/ @chidozieononiwu 

I believe that what we actually need is for the APIVIew API to rely on token auth (OAuth or JWT) rather than cookie auth which doesn't scale cross-origin.

AlitzelMendez:
The plan is to update this authentication requierement: https://github.com/Azure/azure-sdk-tools/blob/9882dfa0172807a19575dd4a00ee07b236f74887/src/dotnet/APIView/APIViewWeb/LeanControllers/BaseApiController.cs#L6 that is the access to the endpoints to allow not only organization (aka github) but also MI, which looks like we want  but let me talk with @chidozieononiwu  to understand if this is an approach that wouldn't be useful for our scenario. 

bottom line we will still be using JWT but this will be generated with the MI permissions

tjprescott:
Implemented in #11246

PR #11246, opened 2025-07-17
[APIView] Managed identity as an alternative way of authentication for API

AlitzelMendez:
Issue: https://github.com/Azure/azure-sdk-tools/issues/10983

Adds support for Azure Managed Identity authentication alongside existing GitHub organization authentication, enabling both web users (cookies) and automated services (Bearer tokens) to access APIView.

Key Changes
- New Authentication System: Created AuthenticationValidator.cs with centralized validation logic for both GitHub org and Azure Managed Identity
- Dual Authentication Handler: Added CookieOrBearerAuthenticationHandler.cs supporting both cookie and Bearer token authentication
- JWT Bearer Support: Configured proper Azure AD token validation with multi-tenant support in Startup.cs- 
- Authorization Policies: Added RequireOrganizationOrManagedIdentityPolicy for combined access control


packages/python-packages/apiview-copilot/cli.py
```diff
@@ -667,15 +667,23 @@ def get_apiview_comments(review_id: str, environment: str = "production", use_ap
...
             apiview_endpoints = {
                 "production": "https://apiview.dev",
-                "staging": "https://apiviewstaging.azurewebsites.net",
+                "staging": "https://apiviewstagingtest.com",
             }
             endpoint_root = apiview_endpoints.get(environment)
             endpoint = f"{endpoint_root}/api/Comments/{review_id}?commentType=APIRevision&isDeleted=false"
+            apiview_scopes = {
+                "production": "api://apiview/.default",
+                "staging": "api://apiviewstaging/.default",
```
tjprescott, line 661: Do we need a UX test scope as well, or does that just use staging scope?

packages/python-packages/apiview-copilot/cli.py
```diff
@@ -667,15 +667,23 @@ def get_apiview_comments(review_id: str, environment: str = "production", use_ap
                 raise ValueError("When using the API, `--review-id` must be provided.")
             apiview_endpoints = {
                 "production": "https://apiview.dev",
-                "staging": "https://apiviewstaging.azurewebsites.net",
+                "staging": "https://apiviewstagingtest.com",
```
tjprescott, line 655: Do we need UX test here as well?

tjprescott (commented):
Will defer to the other reviewers for the C# side, but should we have an auth test?

AlitzelMendez:
> Will defer to the other reviewers for the C# side, but should we have an auth test?

@tjprescott do you mean something like this: https://github.com/Azure/azure-sdk-tools/blob/2f01cfb9079ca844ee824a3fbb263efc4bb55dbb/src/dotnet/APIView/APIViewWeb/Controllers/AuthTestController.cs ?

packages/python-packages/apiview-copilot/cli.py
AlitzelMendez: this only use staging scope

packages/python-packages/apiview-copilot/cli.py
AlitzelMendez: for localhost? :)

tjprescott:
> @tjprescott do you mean something like this: https://github.com/Azure/azure-sdk-tools/blob/2f01cfb9079ca844ee824a3fbb263efc4bb55dbb/src/dotnet/APIView/APIViewWeb/Controllers/AuthTestController.cs ?

Probably. Why doesn't it show any changes to that file in this PR?  The file itself looks like it already includes managed identity, but wasn't that done in this PR?

AlitzelMendez:
> > @tjprescott do you mean something like this: https://github.com/Azure/azure-sdk-tools/blob/2f01cfb9079ca844ee824a3fbb263efc4bb55dbb/src/dotnet/APIView/APIViewWeb/Controllers/AuthTestController.cs ?
> 
> Probably. Why doesn't it show any changes to that file in this PR? The file itself looks like it already includes managed identity, but wasn't that done in this PR?

Added tests! oh yes! it wasn't there, I was just wondering if that was what you had in mind!  but I just add them :D thanks Travis

packages/python-packages/apiview-copilot/cli.py
tjprescott: This is localhost? These look like the live endpoints. There's also: https://apiviewuxtest.com/   Dunno if the API for UX test is the same as staging (only the UI differs) and thus it wouldn't matter.

src/dotnet/APIView/APIViewWeb/Startup.cs
```diff
@@ -143,15 +144,46 @@ public void ConfigureServices(IServiceCollection services)
             {
                 services.AddAuthentication(options =>
                 {
-                    options.DefaultAuthenticateScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                    options.DefaultAuthenticateScheme = "CookieOrBearerAuthentication";
```
tjprescott, line 147: What options do we have for auth?  CookieAuthentication and CookieOrBearerAuthentication? Should we have "BearerAuthentication"? Ideally, we'd transition away from CookieAuth completely (eventually).

src/dotnet/APIView/APIViewWeb/LeanControllers/BaseApiController.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace APIViewWeb.LeanControllers
 {
-    [Authorize("RequireOrganization")]
+    [Authorize("RequireOrganizationOrManagedIdentity")]
```
tjprescott, line 6: Does "Organization" mean "Cookie"?

AlitzelMendez: yes, organization means that you are part of github organization that is the info that is saved on the cookie

tjprescott (commented):
Based on our discussion I'm happy with merging this, but will let @praveenkuttappan provide the approval.

packages/python-packages/apiview-copilot/cli.py
tjprescott: Well, I guess since environment only accepts product and staging it doesn't make sense to add ux_test for now. :P We can add it later easily enough if needed.

src/dotnet/APIView/APIViewWeb/Controllers/AuthTestController.cs
```diff
@@ -0,0 +1,150 @@
...
+            Message = "Bearer token required for development testing",
+            Hint = "Add 'Authorization: Bearer <any-token>' header"
+        });
+    }
+
+    /// <summary>
+    ///     Debug endpoint to test JWT token parsing - no authorization required
+    /// </summary>
+    [HttpGet("jwt-debug")]
+    public IActionResult JwtDebug()
```
praveenkuttappan, line 119: Can we disable this code block in no debug env in addition to Dev environment just to be on safer side from security point?
`#if DEBUG  .... 
#endif`

praveenkuttappan (approved):
One nit about debug api for token. Otherwise, it looks fine.

Issue #11249, opened 2025-07-17
CopilotPollingBackgroundHostedService tests are flaky due to multi-threaded BackgroundService architecture


AlitzelMendez:
## Problem
Tests for `CopilotPollingBackgroundHostedService` are flaky due to the service's complex threading architecture that combines:
- BackgroundService infinite loop
- Task.Run for concurrent job processing  
- Multiple async external dependencies (HTTP, database, SignalR)
- Non-deterministic execution order
- Built-in timing delays

## Root Causes
1. **Multi-threaded execution**: BackgroundService + Task.Run + async operations
2. **Race conditions**: Multiple threads accessing shared state
3. **Timing dependencies**: Fixed delays and polling intervals

## Proposed Solutions
- [ ] Extract business logic from BackgroundService for unit testing
- [ ] Create integration tests with TestContainers
- [ ] Redesign service for better testability

## Importance of These Tests
These tests are critical for the new Copilot integration effort. Since we'll be actively developing this code and the actual process involves expensive Copilot service calls, comprehensive unit test coverage is essential for ensuring reliability and catching regressions during development.

tjprescott:
Fixed in #11570

PR #11250, opened 2025-07-17
Disabaling CopilotPollingBackgroundHostedService tests due to flakiness 

AlitzelMendez:
This is the whole issue explaining why these tests are particularly challenging but also why I really want to have them:
- https://github.com/Azure/azure-sdk-tools/issues/11249

I will be OOF next week, so I don't want to just let these flaky tests in.

If anyone feel strongly about having skipping test and prefer that I delete them in the meantime, please let me know.


tjprescott (approved):
Thanks for creating the tracking issue to re-enable these :)

PR #11281, opened 2025-07-21
[aztsp-cli] Adds a new azure typespec mcp server

chrisradek:
This MCP server currently exposes 2 tools:
- quickstart
- convert_swagger

`quickstart` is used to initialize a new TypeSpec project, and `convert_swagger` is used to convert an existing swagger to a TypeSpec project.

My first c sharp experience, so don't hold back :smile: Used copilot heavily for creating test cases, and used `typespec-mcp` for generating the MCP server scaffolding.

### DIFF
--- a/tools/aztsp-cli/Azure.Tsp.Tools.Mcp.Tests/Azure.Tsp.Tools.Mcp.Tests.csproj
+++ b/tools/aztsp-cli/Azure.Tsp.Tools.Mcp.Tests/Azure.Tsp.Tools.Mcp.Tests.csproj
@@ -1,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="coverlet.collector" Version="6.0.2" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
+    <PackageReference Include="Moq" Version="4.20.72" />
+    <PackageReference Include="NUnit" Version="3.14.0" />
+    <PackageReference Include="NUnit.Analyzers" Version="3.9.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
+  </ItemGroup>
+
+
+  <ItemGroup>
+    <Using Include="NUnit.Framework" />
+  </ItemGroup>
+
+
+  <ItemGroup>
+    <ProjectReference Include="..\Azure.Tsp.Tools.Mcp\Azure.Tsp.Tools.Mcp.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/tools/aztsp-cli/Azure.Tsp.Tools.Mcp.Tests/InitImplTests.cs
+++ b/tools/aztsp-cli/Azure.Tsp.Tools.Mcp.Tests/InitImplTests.cs
@@ -1,0 +1,769 @@
+using Azure.Tsp.Tools.Mcp.Tools;
+
+namespace Azure.Tsp.Tools.Mcp.Tests;
+
+[TestFixture]
+public class InitImplTests
+{
+    private InitImpl _initImpl;
+    private string _tempDirectory;
+
+    [SetUp]
+    public void Setup()
+    {
+        _initImpl = new InitImpl();
+        _tempDirectory = Path.Combine(Path.GetTempPath(), "InitImplTests_" + Guid.NewGuid().ToString("N")[..8]);
+        Directory.CreateDirectory(_tempDirectory);
+    }
+
+    [TearDown]
+    public void TearDown()
+    {
+        // Clean up temp directory after each test
+        if (Directory.Exists(_tempDirectory))
+        {
+            try
+            {
+                Directory.Delete(_tempDirectory, recursive: true);
+            }
+            catch
+            {
+                // Best effort cleanup - ignore errors
+            }
+        }
+    }
+
+    #region QuickstartAsync Validation Tests
+
+    [TestCase("")]
+    [TestCase("   ")]
+    public async Task QuickstartAsync_WithInvalidTemplate_ReturnsFailureMessage(string template)
+    {
+        // Arrange
+        string serviceNamespace = "TestService";
+        string outputDir = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Does.StartWith("Failed: template must be one of:"));
+        Assert.That(result, Does.Contain("azure-core, azure-arm"));
+    }
+
+    [Test]
+    public async Task QuickstartAsync_WithNullTemplate_ReturnsFailureMessage()
+    {
+        // Arrange
+        string template = null!;
+        string serviceNamespace = "TestService";
+        string outputDir = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Does.StartWith("Failed: template must be one of:"));
+        Assert.That(result, Does.Contain("azure-core, azure-arm"));
+    }
+
+    [TestCase("invalid-template")]
+    [TestCase("Azure-Core")] // Case sensitive
+    [TestCase("azure_core")] // Wrong separator
+    public async Task QuickstartAsync_WithUnsupportedTemplate_ReturnsFailureMessage(string template)
+    {
+        // Arrange
+        string serviceNamespace = "TestService";
+        string outputDir = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Does.StartWith("Failed: template must be one of:"));
+        Assert.That(result, Does.Contain($"but was '{template}'"));
+    }
+
+    [TestCase("")]
+    [TestCase("   ")]
+    public async Task QuickstartAsync_WithInvalidServiceNamespace_ReturnsFailureMessage(string serviceNamespace)
+    {
+        // Arrange
+        string template = "azure-core";
+        string outputDir = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Is.EqualTo("Failed: serviceNamespace must be provided and cannot be empty."));
+    }
+
+    [Test]
+    public async Task QuickstartAsync_WithNullServiceNamespace_ReturnsFailureMessage()
+    {
+        // Arrange
+        string template = "azure-core";
+        string serviceNamespace = null!;
+        string outputDir = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Is.EqualTo("Failed: serviceNamespace must be provided and cannot be empty."));
+    }
+
+    [TestCase("")]
+    [TestCase("   ")]
+    public async Task QuickstartAsync_WithInvalidOutputDir_ReturnsFailureMessage(string outputDir)
+    {
+        // Arrange
+        string template = "azure-core";
+        string serviceNamespace = "TestService";
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Is.EqualTo("Failed: outputDirectory must be provided"));
+    }
+
+    [Test]
+    public async Task QuickstartAsync_WithNullOutputDir_ReturnsFailureMessage()
+    {
+        // Arrange
+        string template = "azure-core";
+        string serviceNamespace = "TestService";
+        string outputDir = null!;
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Is.EqualTo("Failed: outputDirectory must be provided"));
+    }
+
+    [Test]
+    public async Task QuickstartAsync_WithNonExistentOutputDir_ReturnsFailureMessage()
+    {
+        // Arrange
+        string template = "azure-core";
+        string serviceNamespace = "TestService";
+        string outputDir = Path.Combine(_tempDirectory, "nonexistent");
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Does.StartWith("Failed: Full output directory"));
+        Assert.That(result, Does.EndWith("does not exist."));
+    }
+
+    [Test]
+    public async Task QuickstartAsync_WithNonEmptyOutputDir_ReturnsFailureMessage()
+    {
+        // Arrange
+        string template = "azure-core";
+        string serviceNamespace = "TestService";
+        string outputDir = _tempDirectory;
+
+        // Create a file in the output directory to make it non-empty
+        File.WriteAllText(Path.Combine(outputDir, "existing-file.txt"), "content");
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Does.StartWith("Failed: The full output directory"));
+        Assert.That(result, Does.EndWith("points to a non-empty directory."));
+    }
+
+    #endregion
+
+    #region QuickstartAsync Valid Input Tests
+
+    [TestCase("azure-core")]
+    [TestCase("azure-arm")]
+    public async Task QuickstartAsync_WithValidInputs_CallsProcessCorrectly(string template)
+    {
+        // Arrange
+        string serviceNamespace = "TestService";
+        string outputDir = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        // Since we can't easily mock the ProcessHelper.RunProcess call without refactoring,
+        // we test that the method doesn't fail on validation and attempts to run the process
+        // The actual process execution might fail (which is expected in a test environment without tsp installed)
+        Assert.That(result, Is.Not.Null);
+        Assert.That(result, Is.Not.Empty);
+    }
+
+    [Test]
+    public async Task QuickstartAsync_WithPascalCaseServiceNamespace_AcceptsInput()
+    {
+        // Arrange
+        string template = "azure-core";
+        string serviceNamespace = "MyTestService"; // Pascal case as recommended
+        string outputDir = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, outputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Is.Not.Null);
+        // The method should pass validation and attempt to run the process
+    }
+
+    [Test]
+    public async Task QuickstartAsync_WithRelativePath_ConvertsToAbsolutePath()
+    {
+        // Arrange
+        string template = "azure-core";
+        string serviceNamespace = "TestService";
+        string relativeOutputDir = Path.GetRelativePath(Environment.CurrentDirectory, _tempDirectory);
+
+        // Act
+        var result = await _initImpl.QuickstartAsync(template, serviceNamespace, relativeOutputDir, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Is.Not.Null);
+        // Should handle relative paths correctly by converting to absolute path
+    }
+
+    #endregion
+
+    #region ConvertSwaggerAsync Validation Tests
+
+    [TestCase("")]
+    [TestCase("   ")]
+    public async Task ConvertSwaggerAsync_WithInvalidPathToSwaggerReadme_ReturnsFailureMessage(string pathToSwaggerReadme)
+    {
+        // Arrange
+        string outputDirectory = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.ConvertSwaggerAsync(pathToSwaggerReadme, outputDirectory, null, null, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Is.EqualTo("Failed: pathToSwaggerReadme must be a valid Markdown file."));
+    }
+
+    [Test]
+    public async Task ConvertSwaggerAsync_WithNullPathToSwaggerReadme_ReturnsFailureMessage()
+    {
+        // Arrange
+        string pathToSwaggerReadme = null!;
+        string outputDirectory = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.ConvertSwaggerAsync(pathToSwaggerReadme, outputDirectory, null, null, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Is.EqualTo("Failed: pathToSwaggerReadme must be a valid Markdown file."));
+    }
+
+    [TestCase("readme.txt")]
+    [TestCase("readme")]
+    [TestCase("readme.MD")] // Should work - case insensitive
+    [TestCase("readme.markdown")]
+    public async Task ConvertSwaggerAsync_WithNonMarkdownFile_ReturnsFailureMessage(string fileName)
+    {
+        // Arrange
+        string pathToSwaggerReadme = fileName == "readme.MD" ? fileName : fileName; // Special case for testing
+        string outputDirectory = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.ConvertSwaggerAsync(pathToSwaggerReadme, outputDirectory, null, null, CancellationToken.None);
+
+        // Assert
+        if (fileName == "readme.MD")
+        {
+            // This should pass the extension check but fail on file existence
+            Assert.That(result, Does.StartWith("Failed: pathToSwaggerReadme"));
+            Assert.That(result, Does.EndWith("does not exist."));
+        }
+        else
+        {
+            Assert.That(result, Is.EqualTo("Failed: pathToSwaggerReadme must be a valid Markdown file."));
+        }
+    }
+
+    [Test]
+    public async Task ConvertSwaggerAsync_WithNonExistentMarkdownFile_ReturnsFailureMessage()
+    {
+        // Arrange
+        string pathToSwaggerReadme = Path.Combine(_tempDirectory, "nonexistent-readme.md");
+        string outputDirectory = _tempDirectory;
+
+        // Act
+        var result = await _initImpl.ConvertSwaggerAsync(pathToSwaggerReadme, outputDirectory, null, null, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Does.StartWith("Failed: pathToSwaggerReadme"));
+        Assert.That(result, Does.EndWith("does not exist."));
+    }
+
+    [TestCase("")]
+    [TestCase("   ")]
+    public async Task ConvertSwaggerAsync_WithInvalidOutputDirectory_ReturnsFailureMessage(string outputDirectory)
+    {
+        // Arrange
+        string readmePath = Path.Combine(_tempDirectory, "readme.md");
+        File.WriteAllText(readmePath, "# Swagger README");
+
+        // Act
+        var result = await _initImpl.ConvertSwaggerAsync(readmePath, outputDirectory, null, null, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Is.EqualTo("Failed: outputDirectory must be provided"));
+    }
+
+    [Test]
+    public async Task ConvertSwaggerAsync_WithNullOutputDirectory_ReturnsFailureMessage()
+    {
+        // Arrange
+        string readmePath = Path.Combine(_tempDirectory, "readme.md");
+        File.WriteAllText(readmePath, "# Swagger README");
+        string outputDirectory = null!;
+
+        // Act
+        var result = await _initImpl.ConvertSwaggerAsync(readmePath, outputDirectory, null, null, CancellationToken.None);
+
+        // Assert
+        Assert.That(result, Is.EqualTo("Failed: outputDirectory must be provided"));
+    }
+
+    [Test]
+    public async Task ConvertSwaggerAsync_WithNonExistentOutputDirectory_ReturnsFailureMessage()
+    {
+        // Arrange
+        string readmePath = Path.Combine(_tempDirectory, "readme.md");
+        File.WriteAllText(readmePath, "# Swagger README");
+        string outputDirectory = Path.Combine(_tempDirectory, "nonexistent");
+
+        // Act
+        var result = await _initImpl.ConvertSwaggerAsync(readmePath, outputDirectory, null, null, CancellationToken.None);
+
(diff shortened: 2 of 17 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/TspToolResponse.cs
```diff
@@ -0,0 +1,38 @@
...
+        public string ErrorMessage { get; set; } = string.Empty;
+
+        public override string ToString()
+        {
+            StringBuilder output = new StringBuilder();
+
+            output.AppendLine($"### Is Successful: {IsSuccessful}");
+            if (IsSuccessful)
+            {
+                output.AppendLine($"### TypeSpec Project Path: {TypeSpecProjectPath}");
```
benbp: Intended to be empty for now?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ProcessHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+// Licensed under the MIT License.
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Azure.Sdk.Tools.Cli.Helpers
+{
+    public class ProcessHelper
+    {
+        public static ProcessResult RunProcess(string command, string[] args, string workingDirectory)
```
benbp: We needed this! Thanks for adding

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ProcessHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+                };
+                process.ErrorDataReceived += (sender, e) =>
+                {
+                    if (e.Data != null)
+                    {
+                        output.AppendLine(e.Data);
+                    }
+                };
+
+                Console.Error.WriteLine($"Running command: {command} {string.Join(" ", args)} in {workingDirectory}");
```
benbp: We should dependency inject `ILogger` instead of using console. Lmk if you need an example.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/TspToolResponse.cs
chrisradek: Not sure I follow - `TypeSpecProjectPath` should be filled in if `IsSuccessful` is true - though I'm not sure whether it's better to conditionally include this line or not.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ProcessHelper.cs
chrisradek: Logger injected! The process helper is also registered for dependency injection now too.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ProcessHelper.cs
```diff
@@ -0,0 +1,107 @@
...
+                if (process.ExitCode != 0)
+                {
+                    output.Append($"{Environment.NewLine}Process failed.");
+                }
+            }
+
+            return new ProcessResult { Output = output.ToString(), ExitCode = exitCode };
+        }
+
+        public ProcessResult RunNpx(List<string> args, string workingDirectory)
```
benbp: This is fine to start but I think we should split out the core process helper and put common commands into their own helper classes (so the process part is hidden from the user).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/TspToolResponse.cs
benbp: The commit I commented on had no code in the block (or there was some UI weirdness) :D

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,342 @@
...
+        private async Task HandleInitCommand(InvocationContext ctx, CancellationToken ct)
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
+        {
+            var template = ctx.ParseResult.GetValueForArgument(templateArg);
+            var serviceNamespace = ctx.ParseResult.GetValueForArgument(serviceNamespaceArg);
+            var outputDirectory = ctx.ParseResult.GetValueForArgument(outputDirectoryArg);
+
+            var result = Init(template, serviceNamespace, outputDirectory);
+            ctx.ExitCode = ExitCode;
+            output.Output(result.ToString());
```
benbp:
```suggestion
            output.Output(result.ToString());
            return Task.CompletedTask;
```

I think you can do something like this to avoid the annoying pragma warning disable statements

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,342 @@
...
+        private readonly Argument<string> outputDirectoryArg = new("--output-directory", "The output directory for the generated TypeSpec project. This directory must already exist and be empty.");
+
+        private readonly Argument<string> swaggerReadmeArg = new("--swagger-readme", "The path or URL to an Azure swagger README file.");
+        private readonly Option<bool> isArmOption = new("--arm", "Whether the generated TypeSpec project is for an Azure Resource Management (ARM) API. This should be true if the swagger's path contains 'resource-manager'.");
+        private readonly Option<bool> fullyCompatibleOption = new("--fully-compatible", "Whether to generate a TypeSpec project that is fully compatible with the swagger. It is recommended not to set this to true so that the converted TypeSpec project leverages TypeSpec built-in libraries with standard patterns and templates.");
+
+        public override Command GetCommand()
+        {
+            var tspCommand = new Command("tsp", "Tools for initializing TypeSpec projects and converting existing Azure service swagger definitions to TypeSpec projects");
```
benbp, line 33: We have some abstraction to enable sub-commands (https://github.com/Azure/azure-sdk-tools/blob/a2fe9df6c47d068dac9cc2cef88f97f7cbc645af/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/EngSys/Cleanup/CleanupTool.cs#L30-L43, https://github.com/Azure/azure-sdk-tools/blob/a2fe9df6c47d068dac9cc2cef88f97f7cbc645af/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedCommandGroups.cs#L8) that you can use. I need to come up with an example for you though.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ProcessHelper.cs
chrisradek: I'm happy to break the `npx` command into its own helper - still ok if it returns `ProcessResult` or prefer `NpxResult` that extends `ProcessResult`? (Even though it won't have anything added yet)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Ah - I think I get it - I can create a `tsp` shared command, and that's the entry I'd have in my `CommandHierarchy`. So then I wouldn't have to do the `tspCommand.AddCommand` call.

Is that the main advantage? Would we want all commands that have more than 1 subcommand to be in `SharedCommandGroups`, even if all subcommands are in one file?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ProcessHelper.cs
chrisradek: Created new `NpxHelper`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
benbp: The general idea with the command hierarchy is that we can know with a glance what exact command a tool class corresponds to. Additionally when we do have sub-commands across classes we can share options and arguments vs. having to re-define them in each class.

I'm debating whether we should mandate that each sub-command go in its own class, though it seems tedious. But trying to figure out whether there's a middle ground here. @scbedd any thoughts?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/NpxHelper.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    public class NpxHelper(IProcessHelper processHelper) : INpxHelper
+    {
+        public ProcessResult RunNpx(List<string> args, string workingDirectory)
+        {
+            var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            if (isWindows)
+            {
+                return processHelper.RunProcess(
+                    "cmd.exe",
```
benbp: Is the cmd call necessary?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/NpxHelper.cs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
```
benbp: Should have thought of this on last review, but maybe we should put this and process helper into `Helpers/Process/<files>`?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,340 @@
...
+    /// </summary>
+    [McpServerToolType, Description("Tools for initializing TypeSpec projects and converting existing Azure service swagger definitions to TypeSpec projects.")]
+    public class TypeSpecTool(INpxHelper npxHelper, ILogger<TypeSpecTool> logger, IOutputService output) : MCPTool
+    {
+
+        // commands
+        private const string initCommandName = "init";
+        private const string convertSwaggerCommandName = "convert-swagger";
+
+        private readonly Argument<string> templateArg = new("--template", "The template to use for the TypeSpec project. Valid values are: azure-core (for data-plane services), azure-arm (for resource-manager services)");
```
benbp: I think these should be defined as [Option](https://learn.microsoft.com/en-us/dotnet/api/system.commandline.option?view=net-9.0-pp) types not Argument types. Arguments are more for positional values vs. options which are flags followed by values in any order. 

Example: https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/AzurePipelines/PipelineAnalysisTool.cs#L60-L65

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
scbedd: The goal of the `command hierarchy` is so that you can get grouped subcommands across various command classes without having to do a bunch of work to ensure that those two subcommands belong to the same parent, which then gets picked up by the automation creating the CLI entrypoint for the server.

IF you're just adding the subcommands to the parent command that gets returned by `GetCommand()`, I don't see where this would break the [standard command hierarchy creation.](https://github.com/Azure/azure-sdk-tools/blob/c2862713609a6dbe51aa65f555ffe07832bb3707/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/CommandFactory.cs#L46).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Ah yes, you're right - I want these to be options not arguments.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/NpxHelper.cs
scbedd: It shouldn't be. I'd prefer that we don't include it if we can.

scbedd: Then again I haven't attempted to invoke `npx` directly from a .NET so there may be restrictions.

chrisradek: This is because `npx` isn't an exe on windows - it's either a `.cmd` or `.ps1`. Trying to invoke it directly or as `npx.cmd/npx.ps1` doesn't work. Only ways I've found to get it to work on Windows is to either call `cmd.exe /C` or `powershell.exe -Command`.

I mostly based this on an existing example: https://github.com/Azure/azure-sdk-tools/blob/c2862713609a6dbe51aa65f555ffe07832bb3707/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecValidationTool/SpecValidationTool.cs#L99-L104

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,220 @@
...
+                var fullPathToSwaggerReadme = Path.GetFullPath(pathToSwaggerReadme.Trim());
+                if (!File.Exists(fullPathToSwaggerReadme))
+                {
+                    SetFailure(1);
+                    return new TspToolResponse
+                    {
+                        IsSuccessful = false,
+                        ErrorMessage = $"Failed: pathToSwaggerReadme '{fullPathToSwaggerReadme}' does not exist."
+                    };
+                }
```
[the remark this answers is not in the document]
chrisradek: pathToSwaggerReadme updated to only be path, not URL

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/NpxHelper.cs
scbedd: And there is the context that gives the reasoning to it!👍 Glad I left myself an out on that answer ahaha

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,219 @@
...
+                    pathToSwaggerReadme, outputDirectory, isAzureResourceManagement, fullyCompatible);
+
+                // Validate pathToSwaggerReadme
+                if (string.IsNullOrWhiteSpace(pathToSwaggerReadme) || !pathToSwaggerReadme.EndsWith(".md", StringComparison.OrdinalIgnoreCase))
+                {
+                    SetFailure(1);
+                    return new TspToolResponse
+                    {
+                        IsSuccessful = false,
+                        ErrorMessage = "Failed: pathToSwaggerReadme must be a valid Markdown file."
```
praveenkuttappan: Should we say valid path to swagger readme? Copilot will provide path to any md file in retry the call unless we explicitly mention it as readme md file.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,219 @@
...
+                logger.LogInformation("Converting swagger to TypeSpec: {pathToSwaggerReadme}, output: {outputDirectory}, isArm: {isArm}, fullyCompatible: {fullyCompatible}",
+                    pathToSwaggerReadme, outputDirectory, isAzureResourceManagement, fullyCompatible);
+
+                // Validate pathToSwaggerReadme
+                if (string.IsNullOrWhiteSpace(pathToSwaggerReadme) || !pathToSwaggerReadme.EndsWith(".md", StringComparison.OrdinalIgnoreCase))
+                {
+                    SetFailure(1);
+                    return new TspToolResponse
+                    {
+                        IsSuccessful = false,
```
praveenkuttappan: This is not required since default is false.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Updated!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Ah, my JS is showing - updated!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/NpxHelper.cs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using System.Runtime.InteropServices;
+
+namespace Azure.Sdk.Tools.Cli.Helpers
```
AlitzelMendez, line 5: can you change the namespace to `Azure.Sdk.Tools.Cli.Helpers.Process` so it matches the actual location? :)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,214 @@
...
+    /// <summary>
+    /// This tool provides functionality for initializing TypeSpec projects and converting existing Azure service swagger definitions to TypeSpec projects.
+    /// Use this tool to onboard to TypeSpec for new services or convert existing services.
+    /// </summary>
+    [McpServerToolType, Description("Tools for initializing TypeSpec projects and converting existing Azure service swagger definitions to TypeSpec projects.")]
+    public class TypeSpecTool(INpxHelper npxHelper, ILogger<TypeSpecTool> logger, IOutputService output) : MCPTool
+    {
+
+        // commands
+        private const string convertSwaggerCommandName = "convert-swagger";
```
AlitzelMendez: just for naming convention change to ConvertSwaggerCommandName

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        private Task HandleConvertCommand(InvocationContext ctx, CancellationToken ct)
+        {
+            var swaggerReadme = ctx.ParseResult.GetValueForOption(swaggerReadmeArg);
+            var outputDirectory = ctx.ParseResult.GetValueForOption(outputDirectoryArg);
+            var isArm = ctx.ParseResult.GetValueForOption(isArmOption);
+            var fullyCompatible = ctx.ParseResult.GetValueForOption(fullyCompatibleOption);
+
+            var result = ConvertSwagger(swaggerReadme, outputDirectory, isArm, fullyCompatible);
```
AlitzelMendez: usually the convention for var is that if it clear what the type is
for examples 
`var x = new List<string>();`
we can use var, but if it is not clear, it is better to specify the type so is easier to read
so in this case I would change to 
            TspToolResponse result = ConvertSwagger(swaggerReadme, outputDirectory, isArm, fullyCompatible);

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,214 @@
...
+        {
+            try
+            {
+                logger.LogInformation("Converting swagger to TypeSpec: {pathToSwaggerReadme}, output: {outputDirectory}, isArm: {isArm}, fullyCompatible: {fullyCompatible}",
+                    pathToSwaggerReadme, outputDirectory, isAzureResourceManagement, fullyCompatible);
+
+                // Validate pathToSwaggerReadme
+                if (string.IsNullOrWhiteSpace(pathToSwaggerReadme) || !pathToSwaggerReadme.EndsWith(".md", StringComparison.OrdinalIgnoreCase))
+                {
+                    SetFailure(1);
```
AlitzelMendez: I think the `1` is redundant because it is already your default

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,214 @@
...
+                    return new TspToolResponse
+                    {
+                        ErrorMessage = "Failed: pathToSwaggerReadme must be a valid path to a swagger README.md file."
```
AlitzelMendez: the only problem that I see with this message is that we are not actually checking if the path is valid, just checking if it is null/empty/whitespaces

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,214 @@
...
+                {
+                    SetFailure(1);
+                    return new TspToolResponse
+                    {
+                        ErrorMessage = $"Failed: pathToSwaggerReadme '{fullPathToSwaggerReadme}' does not exist."
+                    };
+                }
+
+                // validate outputDirectory using the extracted method
+                var validationResult = ValidateOutputDirectory(outputDirectory);
```
AlitzelMendez: I like what you did in here with ValidateOutputDirectory, I think it would be nice having something pretty similar for ValidateSwaggerReadme

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Ah, I actually did camelCase instead of PascalCase because all the other tools were using camelCase for the command names 😅

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Hmmm, I could update this to say `pathToSwaggerReadme must be provided and non-empty` - I was thinking of it as "if the path is not provided and is whitespace, it's not a valid path". I suspect both will work for MCP, though the former might be clearer for the CLI use case

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,230 @@
...
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using System.ComponentModel;
+using Azure.Sdk.Tools.Cli.Services;
+using Azure.Sdk.Tools.Cli.Contract;
+using Azure.Sdk.Tools.Cli.Models.Responses;
+using ModelContextProtocol.Server;
+using Azure.Sdk.Tools.Cli.Helpers.Process;
+
+namespace Azure.Sdk.Tools.Cli.Tools.TspTool
```
benbp:
```suggestion
namespace Azure.Sdk.Tools.Cli.Tools
```

We're just using the one Tools namespace for these. I'm going to fix the editor warnings on this.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedOptions.cs
```diff
@@ -6,6 +6,7 @@
 using Azure.Sdk.Tools.Cli.Tools.HostServer;
 using Azure.Sdk.Tools.Cli.Tools.SdkRelease;
 using Azure.Sdk.Tools.Cli.Tools.ReleaseReadiness;
+using Azure.Sdk.Tools.Cli.Tools.TspTool;
```
benbp:
```suggestion
```

See my below comment.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
benbp: I'm going to normalize everything to hyphenated rather soon.

benbp: That's our bad :) I don't think they are const everywhere either, some are just `private string` hence the lowercase.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,230 @@
...
+                    pathToSwaggerReadme, outputDirectory, isAzureResourceManagement, fullyCompatible);
+
+                // Validate pathToSwaggerReadme
+                string? readmeValidationResult = ValidateSwaggerReadme(pathToSwaggerReadme);
+                if (readmeValidationResult != null)
+                {
+                    SetFailure();
+                    return new TspToolResponse
+                    {
+                        ErrorMessage = readmeValidationResult
```
benbp: You can use the `ResponseError` or `ResponseErrors` property from the base `Response` class for this. We already have base formatting/behavior for `ResponseError` that gets called automatically in failure cases.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,230 @@
...
+            if (!File.Exists(fullPathToSwaggerReadme))
+            {
+
+                return $"Failed: pathToSwaggerReadme '{fullPathToSwaggerReadme}' does not exist.";
+            }
+
+            return null; // Validation passed
+        }
+
+        private static string? ValidateOutputDirectory(string outputDir)
```
benbp: We could probably move these into a `FileHelper` if you're willing.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,230 @@
...
+            }
+
+            return null; // Validation passed
+        }
+
+        private TspToolResponse RunTspClient(string pathToSwaggerReadme, string outputDirectory, bool isAzureResourceManagement, bool fullyCompatible)
+        {
+            var argsList = new List<string>
+            {
+                "--package=@azure-tools/typespec-client-generator-cli",
```
benbp: If npx helper gets more use, we may consider something like:

```
var cmd = npxHelper.CreateCommand();
cmd.Package = "@azure-tools/typespec"
cmd.AddArgs([...]);
cmd.AddArgs("--arm");
cmd.Cwd = Environment.CurrentDirectory;
cmd.Run();
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/NpxHelper.cs
benbp: Early on we decided on the convention of a flat namespace for Helpers, Tools, etc.

chrisradek: Switching back

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedOptions.cs
chrisradek: Updating! Figures I followed one of the few tools doing this 😅

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Added!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Switched to using `ResponseError`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Created a `FileHelper` with `ValidateEmptyDirectory`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/FileHelper.cs
```diff
@@ -0,0 +1,42 @@
...
+    public static class FileHelper
+    {
+        /// <summary>
+        /// Validates that the specified directory exists and is empty.
+        /// </summary>
+        /// <param name="dir">The directory path to validate.</param>
+        /// <returns>An error message if validation fails, or null if validation passes.</returns>
+        public static string? ValidateEmptyDirectory(string dir)
+        {
+            if (string.IsNullOrWhiteSpace(dir))
```
benbp, line 18: The error case being a non-empty string seems a bit off here. Do we want to throw an exception instead with a void return type, or is it not really an error case per se? If not, maybe better represented as an enum vs. a string so the caller doesn't have to parse the reason.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,193 @@
...
+            var isArm = ctx.ParseResult.GetValueForOption(isArmOption);
+            var fullyCompatible = ctx.ParseResult.GetValueForOption(fullyCompatibleOption);
+
+            TspToolResponse result = ConvertSwagger(swaggerReadme, outputDirectory, isArm, fullyCompatible);
+            ctx.ExitCode = ExitCode;
+            output.Output(result);
+            return Task.CompletedTask;
+        }
+
+        [McpServerTool(Name = "ConvertSwaggerToTypeSpec"), Description(@"Converts an existing Azure service swagger definition to a TypeSpec project.
```
benbp: We're going to start requiring mcp tool names be snake_case

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/FileHelper.cs
benbp: Ah I see we're just passing the validation message along as we compose the error response above. That's probably less tediuos than a try/catch any time this is called.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,193 @@
...
+        {
+            var swaggerReadme = ctx.ParseResult.GetValueForOption(swaggerReadmeArg);
+            var outputDirectory = ctx.ParseResult.GetValueForOption(outputDirectoryArg);
+            var isArm = ctx.ParseResult.GetValueForOption(isArmOption);
+            var fullyCompatible = ctx.ParseResult.GetValueForOption(fullyCompatibleOption);
+
+            TspToolResponse result = ConvertSwagger(swaggerReadme, outputDirectory, isArm, fullyCompatible);
+            ctx.ExitCode = ExitCode;
+            output.Output(result);
+            return Task.CompletedTask;
```
benbp: This whole method can be made sync (`private void`)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -0,0 +1,193 @@
...
+            var command = ctx.ParseResult.CommandResult.Command.Name;
+
+            switch (command)
+            {
+                case ConvertSwaggerCommandName:
+                    await HandleConvertCommand(ctx, ct);
+                    return ctx.ExitCode;
+                default:
+                    logger.LogError($"Unknown command: {command}");
+                    return 1;
```
benbp: You can remove the `int` return type and just return Task

PR #11284, opened 2025-07-21
Update azsdk readme

benbp:
Improving the landing readme for the azsdk cli

### DIFF
--- a/tools/azsdk-cli/README.md
+++ b/tools/azsdk-cli/README.md
@@ -100,7 +100,8 @@ Tool classes are the core of the azure sdk app and implement CLI commands or MCP
 - Reference or copy from the [hello world tool](Azure.Sdk.Tools.Cli/Tools/HelloWorldTool/HelloWorldTool.cs) for an example.
 - See [Tools/README.md](./Azure.Sdk.Tools.Cli/Tools/README.md) for more details
 
-Each tool must implement a `GetCommand()` method and a `HandleCommand(...)` method. These allow the tool to be called from CLI mode.
+Each tool class should implement a `GetCommand()` method and a `HandleCommand(...)` method. These allow the tools to be called from CLI mode.
+Some exceptions may apply on a case by case basis for tool classes where we implement MCP mode but not CLI mode and vice versa.
 
 ```csharp
 public override Command GetCommand()


tools/azsdk-cli/README.md
```diff
@@ -1,56 +1,138 @@
...
+
+## Adding a New Tool
+
+Tool classes are the core of the azure sdk app and implement CLI commands or MCP server tools. To add a new tool start with the following:
+
+- Determine the right directory and tool class name under `Azure.Sdk.Tools.Cli/Tools`
+- Reference or copy from the [hello world tool](Azure.Sdk.Tools.Cli/Tools/HelloWorldTool/HelloWorldTool.cs) for an example.
+- See [Tools/README.md](./Azure.Sdk.Tools.Cli/Tools/README.md) for more details
+
+Each tool must implement a `GetCommand()` method and a `HandleCommand(...)` method. These allow the tool to be called from CLI mode.
```
weshaggard: I assume GetCommand is for the CLI portion, correct? Is it a must to implement if you only need the MCP Functionality?

tools/azsdk-cli/README.md
```diff
@@ -1,56 +1,138 @@
...
+    var result = await SomeMethod();
+    ctx.ExitCode = ExitCode;
+    output.Output(result);
+}
+```
+
+Additionally, the tool class and any methods that will be included in the MCP server tools list must have the MCP attributes:
+
+```csharp
+[McpServerToolType, Description("Example tool")]
```
weshaggard, line 125: One thing I would like us to investigate is how closely this matches what we are doing in the azure-mcp repo. I would like the registration process to be as close as possible, and we might even start depending on the core library from there.

weshaggard (approved):

tools/azsdk-cli/README.md
```diff
@@ -1,56 +1,138 @@
...
 
-Concept of `chaining` of various MCP servers. Can this hub support taking the output of one spoke and shoving it to another? We probably should right?
+### Pipelines
 
+Public CI - https://dev.azure.com/azure-sdk/public/_build?definitionId=7677
 
-## Some random DOs and DON'Ts of this server
+Release - https://dev.azure.com/azure-sdk/internal/_build?definitionId=7684 
+
+## Adding a New Tool
```
weshaggard, line 95: It might be worth talking about the relationship between the CLI vs MCP command and how they relate.

tools/azsdk-cli/README.md
benbp: Are there any MCP tool areas we have right now that we wouldn't also want to expose in the CLI?

Right now we expose some tools to the MCP server and not the CLI, but they still live within a classes that have functionality exposed to both. We could split this up but we would need to define another base class that doesn't require it.

FYI @scbedd

tools/azsdk-cli/README.md
benbp: The azure-mcp repo has more abstraction/complexity around this based on the nature of their app that we decided we did not need in this codebase. But for the most part we follow the same general pattern - every method can plug into a cli command or use an mcp attribute.

tools/azsdk-cli/README.md
benbp: I don't see a compelling reason (yet) that 99% of new commands we add can't be 1:1 in the cli or mcp.

I'll add some more detail on what you suggest, but I think that once the command structure and response schema has been laid out, most developers should not have to think too deeply about the caller of their tool (cli vs. mcp).

tools/azsdk-cli/README.md
scbedd: Yeah, we would simply create a more base base-class that didn't implement `GetCommand()`. We will still discover _those_ via our static analyzer to ensure it's added to the tools list, but it wouldn't force a CLI portion.

I do think it's quite valuable to force the CLI side (pit of success) but I won't die on that hill 👍

weshaggard: Ok let's have the default be add both CLI and MCP tool together, and if we find any cases where that isn't feasible or realistic then we can consider our options.

maririos: Can we reword this to match what Wes said? by default we expect this behavior, but if it doesn't apply, make sure to communicate the reasons

PR #11287, opened 2025-07-22
Add approval publish environment to azsdk release pipeline

benbp:
Add approval publish environment to azsdk release pipeline

### DIFF
--- a/eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
+++ b/eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
@@ -4,6 +4,7 @@ parameters:
   Feed: 'public/azure-sdk-for-net'
   ShouldPublishExecutables: false
   RequireStrongNames: true
+  PublishEnvironment: ''
   ExeMatrix: []
 
 # Needs Github Release variable group linked.
@@ -33,6 +34,8 @@ stages:
     parameters:
       ArtifactName: packages
       Feed: ${{ parameters.Feed }}
+      ${{ if parameters.PublishEnvironment }}:
+        PublishEnvironment: ${{ parameters.PublishEnvironment }}
 
   - job: PostPackagePublish
     displayName: Post Package Publish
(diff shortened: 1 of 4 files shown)


tools/azsdk-cli/ci.yml
```diff
@@ -27,6 +27,7 @@ extends:
     ToolDirectory: tools/azsdk-cli/Azure.Sdk.Tools.Cli
     ReleaseBinaries: true
     TestDirectory: tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests
+    PublishEnvironment: 'tools'  # require engsys team approval for release
```
weshaggard: Let's use the general 'publish-package" environment we don't need to create another for just this.

eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
```diff
@@ -33,6 +34,8 @@ stages:
     parameters:
```
weshaggard, line 34: Let's condition this publish cli publishing to only happen from main.

tools/azsdk-cli/ci.yml
benbp: @weshaggard `publish-package` is permissioned for azure sdk partners. I assumed we wanted publish permissions for this package to be a bit more limited?

eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
benbp: @weshaggard were you thinking this for all net packages in tools or just azsdk-cli? CC @scbedd

tools/azsdk-cli/ci.yml
weshaggard: I don't think we need to limit it we aren't trying to lock it down but just trying to put a manual approval.

eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
weshaggard: I think for the cli only.

weshaggard (approved):

PR #11289, opened 2025-07-22
Add shell script for spell check

fanyang-mono:
Add a shell script for non-windows developers and documentation for the scripts.

### DIFF
--- a/eng/common/spelling/README.md
+++ b/eng/common/spelling/README.md
@@ -1,0 +1,109 @@
+# Spelling Check Scripts
+
+This directory contains scripts to run cspell (Code Spell Checker) on the repository using the dependencies defined in the adjacent `package*.json` files.
+
+## Adding Legitimate Words
+
+If the spell checker flags legitimate words as misspelled, you can add them to the dictionary configuration file located at `.vscode/cspell.json`.
+
+### Where to Add Words
+
+There are two main places to add legitimate words. Maintain alphabetical order when adding words to keep the dictionary organized:
+
+1. **Root-level words array**: Add words to the `"words"` array at the root level of the configuration file. This is the preferred location for project-specific terms, technical vocabulary, and commonly used words.
+
+2. **Baseline dictionary**: Add words to the `"baseline"` dictionary under `"dictionaryDefinitions"`. This is typically used for words that were already present in the codebase when the spell checker was first introduced.
+
+
+### Example
+
+To add new words, edit `.vscode/cspell.json` and add them to the `"words"` array:
+
+```json
+{
+    "words": [
+        "myprojectname",
+        "customterm",
+        "technicalword"
+    ]
+}
+```
+
+### Guidelines
+
+- Use lowercase for words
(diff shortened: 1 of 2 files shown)


weshaggard:
@fanyang-mono what is the context for this? In what context can you not run the existing ps script? We generally try to avoid maintaining copies of these scripts.

fanyang-mono:
Hi @weshaggard, I usually develop on mac and would like to have a shell script to run spell checks. Developers working on non-windows platforms usually don't have powershell installed in their system.

Regarding to maintenance, with the help of copilot, I think the human effort needed here can be very minimal. For example, someone make a change to one of the script, they could simply prompt to copilot "apply the same change to \<the other scrip\>".

benbp:
I appreciate the effort to improve the cross platform dev experience! Most of our engsys/developer automation requires powershell core as a base dependency, precisely because it allows us to write scripts with cross platform support instead of duplicating logic. [Powershell](https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell-on-macos?view=powershell-7.5) is a very easy tool to install on any platform. I have been gradually making our powershell scripts executable from a bash shell by running `chmod +x <script>` and adding a shebang to the first line: `#! /bin/env pwsh` ([example](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/stress-testing/deploy-stress-tests.ps1#L1)). That way you can just do `./eng/common/spelling/Invoke-Cspell.ps1` without having to drop into `pwsh` from bash. I think it would be better to go with that approach here as well.

fanyang-mono:
@benbp Thanks for acknowledge my intention here. I appreciate that you are aware of the non-windows dev experience. I agree that it is easy to install powershell. However, I can't help but wonder why installing powershell when there is built-in bash and if powershell is the best choice for cross-platform scripting.

benbp:
@fanyang-mono it's a fair question, and a debate we've had quite a bit on the team. The large majority of azure sdk contributors work on windows, with a fair amount on linux/wsl and a few on mac. Powershell has tended to be a better common denominator than bash for this (windows installations are typically Git Bash which is a poor substitute), and it's been an easier scripting language to handle at scale than bash (easily 10k+ lines across the repos). There's also the consideration that common functions in powershell are going to be more consistent in implementation across platforms and we can enforce a minimum runtime version vs. calling out to core utils (gnu on linux vs. freebsd on mac or whatever version of bash or gnu coreutils the user may have installed).

We've also considered/debated python, js and other languages to handle this but have come back to powershell mainly because it's simpler to shell out and easier for simple ad-hoc scripting.

I acknowledge that it's not necessarily the most popular choice, but it's the one we've made and thus a convention we try to stick to. Arguably it's better to have one consistent automation language, even if inferior in some ways, than a proliferation of different ones as we are a small team that has to maintain the whole surface area at the end of the day.

fanyang-mono:
By no means that I intend to make it harder for your team. And I respect your team's decision. In terms of which programming language is easier to work with, arguably, it comes down to a personal choice. With that, I am going to withdraw my PR.

PR #11291, opened 2025-07-22
[perf] Always install .NET before running perf automation

mikeharder:
- May not be pre-installed on ubuntu-24

### DIFF
--- a/eng/common/pipelines/templates/jobs/perf.yml
+++ b/eng/common/pipelines/templates/jobs/perf.yml
@@ -113,6 +113,8 @@ jobs:
     parameters:
       AgentImage: $(OSVmImage)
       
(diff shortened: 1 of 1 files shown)


benbp (approved):

eng/common/pipelines/templates/jobs/perf.yml
```diff
@@ -113,6 +113,8 @@ jobs:
     parameters:
       AgentImage: $(OSVmImage)
       
+  - template: /eng/pipelines/templates/steps/install-dotnet.yml
```
weshaggard, line 116: This isn't going to work in repos where this template doesn't exist. I think it only exists in the tools and .NET repo.

PR #11319, opened 2025-07-23
Update `Test-Proxy` devcert

scbedd:
Version in `main` expires `7/27`, so I'm getting this update out now.


weshaggard:
This file currently isn't in the specs repo which is what is causing the sync to fail. I don't remember exactly the issue there but was flagging some sort of security thing if IIRC. It might have been CredScan but I've disabled that in the spec repo recently so it might be worth trying to add this file back.

PR #11321, opened 2025-07-23
local and spec repo sdk generation flow

radhgupta:
## Summary

This PR introduces command-line tool that enables developers to generate Azure SDKs locally from either local TypeSpec projects or GitHub commits. The tool integrates with Azure AI Agents for automated Analyzer or Typespec error fixing during the generation process.

### 🚀 **Dual Generation Modes**
- **Local TypeSpec**: Generate SDK from a local TypeSpec project directory
- **GitHub Commit**: Generate SDK from a specific commit in the Azure REST API specs repository

#### **Generate SDK from Local TypeSpec Project**
```bash
 dotnet run -- -t "C:\Users\radhgupta\Desktop\azure-rest-api-specs\specification\testservice\TestService" -o "C:\Users\radhgupta\Desktop\azure-sdk-for-net\sdk\testservice\Azure.Data.TestService"
```
 

**Generate SDK from specs repository**
 ```bash
 dotnet run -- -c "a1ad5540e8b9d32653a5a0e477213a499ae533a7" -t "specification/testservice/TestService" -o "C:\Users\radhgupta\Desktop\azure-sdk-for-net\sdk\testservice\Azure.Data.TestService"
```

### CODE REVIEW

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+
+            Option<string?> typespecPathOption = new(
+                "--typespec-path",
+                "Path to the local TypeSpec project directory");
+
+            Option<string?> commitIdOption = new(
+                "--commit-id",
+                "GitHub commit ID to generate SDK from");
+
+            Option<string?> typespecSpecDirectoryOption = new(
```
jorgerangel-msft: should we shortname this option with an alias? Something like `-t` or some other alias.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <summary>
+        /// Creates and configures the root command for the application.
+        /// </summary>
+        /// <param name="handler">The handler function to execute when the command is invoked.</param>
+        /// <returns>The configured root command.</returns>
+        public RootCommand CreateRootCommand(Func<string?, string?, string?, string, Task<int>> handler)
+        {
+            RootCommand rootCommand = new("Azure SDK Generator Agent");
+
+            Option<string?> typespecPathOption = new(
```
jorgerangel-msft: This is more of a general design question, but have we considered just having a single option for the path to the typespec that handles both local and remote specs? We could always check to see if a spec exists on disk first, and if it doesn't fallback to the specs repo. If the spec doesn't exist in either location, then the command could fail. Though there are some considerations for how that would work with the commit id but it could simply the usage.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+                "--typespec-path",
+                "Path to the local TypeSpec project directory");
+
+            Option<string?> commitIdOption = new(
```
jorgerangel-msft, line 39: should we shortname this option with an alias? Something like `-c `?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+                "--commit-id",
+                "GitHub commit ID to generate SDK from");
+
+            Option<string?> typespecSpecDirectoryOption = new(
+                "--typespec-spec-directory",
+                "TypeSpec specification directory (e.g., specification/testservice/TestService). Required when using --commit-id.");
+
+            Option<string> sdkOutputPathOption = new(
```
jorgerangel-msft: should we just call this output-directory or `-o` for short?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            AppSettings = appSettings;
+            Logger = logger;
+            ProcessExecutor = processExecutor;
+            CommitId = commitId;
+            TypespecSpecDirectory = typespecSpecDirectory;
+            SdkOutputDirectory = sdkOutputDirectory;
+        }
+
+        public async Task<bool> CompileTypeSpecAsync(CancellationToken cancellationToken = default)
```
jorgerangel-msft: should we just have this method return the Task and let the caller handle the exception?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,175 @@
...
+            {
+                Logger.LogError("Source directory not found: {SrcDirectory}", srcDirectory);
+                return false;
+            }
+
+            Logger.LogInformation("Executing dotnet build in directory: {Directory}", srcDirectory);
+
+            (bool success, string output, string error) = await ProcessExecutor.ExecuteAsync(
+                AppSettings.CommandExecutor,
+                $"{AppSettings.CommandPrefix} dotnet build /t:generateCode",
```
jorgerangel-msft: Just something to consider for a follow-up or sometime in the future. There may be cases where some additional debug output may be useful to the user. We could optionally accept additional options so the build command is ran with `/p:Debug=True`.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,286 @@
...
+            }
+        }
+
+        protected virtual async Task<bool> InstallTypeSpecDependencies(CancellationToken cancellationToken)
+        {
+            Logger.LogInformation("Installing TypeSpec dependencies in: {ProjectPath}", TypeSpecSourcePath);
+
+            (bool success, string output, string error) = await ProcessExecutor.ExecuteAsync(
+                AppSettings.CommandExecutor,
+                $"{AppSettings.CommandPrefix} npm install --save-dev {AppSettings.TypespecEmitterPackage}",
```
jorgerangel-msft: Is this supposed to be installing the emitter? If so, I think we should install it at the global level and not as a dev dependency: `npm install -g {AppSettings.TypespecEmitterPackage}`

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,286 @@
...
+
+        protected virtual async Task<bool> CompileTypeSpec(CancellationToken cancellationToken)
+        {
+            Logger.LogInformation("Compiling TypeSpec project");
+
+            string tspOutputPath = Path.Combine(SdkOutputDirectory, AppSettings.TspOutputDirectoryName);
+
+            (bool success, string output, string error) = await ProcessExecutor.ExecuteAsync(
+                AppSettings.CommandExecutor,
+                $"{AppSettings.CommandPrefix} npx tsp compile . --emit {AppSettings.TypespecEmitterPackage} --output-dir \"{tspOutputPath}\"",
```
jorgerangel-msft: Similar to another question I posted. More of an FYI and personal opinion: We should keep in mind future improvements to possibly allow passing in additional options.

jorgerangel-msft: the first positional arg in the compile command is assuming the typespec file exists in the current directory this command is being executed from. Is it guaranteed that the spec will exist there when this is executed?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,286 @@
...
+                string tspOutputPath = Path.Combine(SdkOutputDirectory, AppSettings.TspOutputDirectoryName);
+                string generatedSourcePath = Path.Combine(tspOutputPath, AppSettings.TypeSpecDirectoryName, AppSettings.HttpClientCSharpDirectoryName);
+
+                if (!DirectoryExists(generatedSourcePath))
+                {
+                    Logger.LogError("Generated source directory not found: {GeneratedSourcePath}", generatedSourcePath);
+                    return false;
+                }
+
+                Logger.LogInformation("Moving files from {SourcePath} to {TargetPath}", generatedSourcePath, SdkOutputDirectory);
```
jorgerangel-msft: Maybe I'm missing something here, but why do we need to move the files after generating? Could we not just specify the target output directory when emitting?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Azure.Tools.GeneratorAgent.csproj
```diff
@@ -6,6 +6,9 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <SignAssembly>false</SignAssembly>
+    <PackAsTool>true</PackAsTool>
+    <ToolCommandName>typespec-generator</ToolCommandName>
```
jsquire: That's a long command.    Maybe we can trim this down to `generate` and see if there's any conflict.   It's an internal-facing tool, so we don't have to be guarded against a giant set of edge cases.

That would give us:  `dotnet tool run generate`

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs
jsquire: It's a good suggestion to start with.   If we need to differentiate because we have to do PRs for client.tsp updates or something, we can always split later.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
jsquire: +1 to that thought.   I think we'll need to shift to that pattern eventually so that we have the error context to pass to the Fixer + Agent to try and remediate.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,175 @@
...
+                return true;
+            }
+            catch (Exception ex)
+            {
+                Logger.LogError(ex, "Unexpected error during GitHub-based TypeSpec compilation");
+                return false;
+            }
+        }
+
+        protected virtual string ExtractAzureSdkPath()
```
jsquire: This one confuses me.   Help me understand what we need this for?   What's the end-to-end scenario of what we're triyng to do?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            string scriptPath = Path.Combine(azureSdkPath, AppSettings.PowerShellScriptPath);
+
+            if (!FileExists(scriptPath))
+            {
+                Logger.LogError("PowerShell script not found: {ScriptPath}", scriptPath);
+                return false;
+            }
+
+            string arguments = $"/c pwsh \"{scriptPath}\" " +
```
jsquire: We have to be very careful here and make sure that we're validating anything that someone passes in.   Unless we form paths with zero settings/user input, they're untrusted.

We should always do a `Directory.Exists` or `File.Exists` or similar.   That guarantees us that it is a valid path, performs the basic validations for us.

The reason that it is important in this context is because we're executing a process that runs a script.  We need to be sure that nobody can potentially inject something unexpected here and compromise the environment.    We should probably talk through approach and see if there any things we overlooked.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
jsquire: That'd be helpful for any Fixer + Agent too!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,175 @@
...
+                    Logger.LogError("dotnet build standard output: {Output}", output);
+                }
+                return false;
+            }
+
+            Logger.LogInformation("dotnet build /t:generateCode completed successfully");
+            return true;
+        }
+
+        protected virtual bool FileExists(string path) => File.Exists(path);
```
jsquire: Thinking testability, I like it!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,286 @@
...
+                throw new ArgumentException("TypeSpec source path cannot be null or whitespace", nameof(typeSpecSourcePath));
+            }
+
+            if (string.IsNullOrWhiteSpace(SdkOutputDirectory))
+            {
+                throw new ArgumentException("SDK output directory cannot be null or whitespace", nameof(sdkOutputDirectory));
+            }
+        }
+
+        protected virtual bool DirectoryExists(string path) => Directory.Exists(path);
```
jsquire: You can define these common things as a default implementation in the interface or shift to a base class so that you're not writing the same thing over and over.

It may even make sense to extract a "FilerOperations" class for this.   (there are test shim packages that do a nice job here, but none are on our approved list, sadly.)

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
jsquire: @jorgerangel-msft:  Is this a normal part of the generation workflow?   I'm on the fence for whether we should assume we can/should do installs here or move this up to the interactive space.    But, I don't know what's normal.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutor.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// <exception cref="ArgumentException">Thrown when command is null or whitespace.</exception>
+        public virtual async Task<(bool Success, string Output, string Error)> ExecuteAsync(
+            string command,
+            string arguments,
+            string? workingDirectory,
+            CancellationToken cancellationToken,
+            TimeSpan? timeout = null)
+        {
+            if (string.IsNullOrWhiteSpace(command))
+                throw new ArgumentException("Command cannot be null or empty", nameof(command));
```
jsquire:
```suggestion
            if (string.IsNullOrWhiteSpace(command))
            {
                throw new ArgumentException("Command cannot be null or empty", nameof(command));
            }
```

Always braces, please.   That one is important because it removes the potential for a pretty nasty bug to creep in.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutor.cs
```diff
@@ -0,0 +1,206 @@
...
+
+            arguments ??= string.Empty;
+
+            try
+            {
+                StringBuilder outputBuilder = new StringBuilder(capacity: 4096);
```
jsquire: Unless we have some data around these numbers being helpful for memory management, we should just use the default constructor.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutor.cs
```diff
@@ -0,0 +1,206 @@
...
+                StringBuilder errorBuilder = new StringBuilder(capacity: 1024);
+                object outputLock = new object();
+                object errorLock = new object();
```
jsquire:
```suggestion
                object outputLock = new();
                object errorLock = new();
```

Since you're not a `var` fan, I'll mention this is something that you can also do to safe keystrokes.   Your call, just mentioning for awareness.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutor.cs
```diff
@@ -0,0 +1,206 @@
...
+                    return (false, string.Empty, ex.Message);
+                }
+                catch (InvalidOperationException ex)
+                {
+                    Logger.LogError(ex, "Invalid operation starting process: {Command}", command);
+                    return (false, string.Empty, ex.Message);
+                }
+
+                string output, error;
+                lock (outputLock)
```
jsquire: Good threading awareness.   We should take a deeper look at the lock strategies and such here when we do performance tuning.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs
radhgupta: Sounds good, let me implement it.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
radhgupta: Sure, the flow for using github commit id to generate sdk is:

1. `pwsh eng/scripts/automation/Invoke-TypeSpecDataPlaneGenerateSDKPackage.ps1 ` -sdkFolder "c:\Users\radhgupta\Desktop\azure-sdk-for-net\sdk\testservice\Azure.Data.TestService" ` -typespecSpecDirectory "specification/testservice/TestService" ` -commit "a1ad5540e8b9d32653a5a0e477213a499ae533a7" ` -repo "Azure/azure-rest-api-specs"`
2. `cd  .testservice\Azure.Data.TestService\src\`
 
3. `dotnet build /t:generateCode --verbosity diagnostic`
 
 We need `ExtractAzureSdkPath` because the first command runs at the root folder level and generates a template for sdk. We assume the user starts the tool with giving us the exact  folder of the sdk e.g. "c:\Users\radhgupta\Desktop\azure-sdk-for-net\sdk\testservice\Azure.Data.TestService"

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
radhgupta: @jorgerangel-msft yes, it is suposse to install the emitter, I will do it at global level, i agree it better!!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
radhgupta: Completely agree, I am gonna take a look at it.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
radhgupta: I attempted that setup, but it still creates a subfolder—named after the emitter—inside the specified output directory. Within that emitter-named folder is the “generated” folder.

This structure exists so we can clearly identify which emitter produced each piece of code. By default, the system always creates a subfolder using the emitter’s name.

JoshLove-msft: Can you share the command you were using? Using the [emitter-output-dir](https://github.com/Azure/azure-sdk-for-net/tree/main/eng/packages/http-client-csharp#emitter-output-dir) _should_ work.

radhgupta: npx tsp compile . --emit "@azure-typespec/http-client-csharp"  --output-dir \"{tspOutputPath}\""

I think I missed using  "emitter-output-dir",  I am going to give it a try and fix it.

Thanks for sharing this!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
JoshLove-msft: If this is meant to be used mostly for Azure use cases, should we consider using tsp-client? Even if it is meant to be more generic I wonder if we should add features to tsp-client rather than adding a new system for generating based off of a commit.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
JoshLove-msft: The full syntax would be:
```npx tsp compile . --emit "@azure-typespec/http-client-csharp" --option "@azure-typespec/http-client-csharp.emitter-ouput-dir={tspOutputPath}""```

But I am curious why we are not using tsp-client as it has a lot of this stuff built-in.

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs (hunk 7 above)
Comment (radhgupta) [reply src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs]: @JoshLove-msft, as you mentioned tsp-client can be used only if we have the commit id however, given the condition that the typespec files are present locally and yet to be pushed to the spec repo I think tsp-client won't be much of some help.

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs (hunk 9 above)
Comment (JoshLove-msft) [reply src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs]: Nevermind - this is for the case of using a local tsp file.

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs (hunk 7 above)
Comment (JoshLove-msft) [reply src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs]: Yep, makes sense - thanks for clarifying.

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs [hunk 19]
```diff
@@ -0,0 +1,175 @@
...
+            // Extract azure-sdk-for-net path from SDK output directory
+            string azureSdkPath = SdkOutputDirectory;
+            while (!string.IsNullOrEmpty(azureSdkPath) && !Path.GetFileName(azureSdkPath).Equals(AppSettings.AzureSdkDirectoryName, StringComparison.OrdinalIgnoreCase))
+            {
+                string? parentPath = Path.GetDirectoryName(azureSdkPath);
+                if (parentPath == null)
+                {
+                    break;
+                }
+                azureSdkPath = parentPath;
```
christothes: nit: I think you could simplify this to:
```suggestion
                azureSdkPath = Path.GetDirectoryName(azureSdkPath);
```

The while loop will terminate naturally when `azureSdkPath` is null or empty and the check below will still handle it as before.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,286 @@
...
+            {
+                if (!DirectoryExists(TypeSpecSourcePath))
+                {
+                    throw new DirectoryNotFoundException($"TypeSpec project directory not found: {TypeSpecSourcePath}");
+                }
+
+                Logger.LogInformation("Starting TypeSpec compilation for project: {ProjectPath}", TypeSpecSourcePath);
+                Logger.LogInformation("Output SDK path: {OutputPath}", SdkOutputDirectory);
+
+                if (!await InstallTypeSpecDependencies(cancellationToken))
```
christothes: Just a small note on style - typically methods that indicate success or failure via a bool result would be called `Try*`

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,286 @@
...
+                    else
+                    {
+                        MoveFile(sourceFile, destFile);
+                    }
+                    
+                    return;
+                }
+                catch (IOException) when (attempt < maxRetries)
+                {
+                    Thread.Sleep(retryDelayMs * attempt);
```
christothes: Just curious - under what conditions would we expect to fail here and recover on a retry?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutor.cs
```diff
@@ -0,0 +1,206 @@
...
+            }
+            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            {
+                Logger.LogInformation("Command execution was cancelled: {Command}", command);
+                return (false, string.Empty, "Operation was cancelled");
+            }
+            catch (Exception ex)
+            {
+                Logger.LogError(ex, "Unexpected error executing command: {Command}", command);
+                return (false, string.Empty, ex.Message);
```
christothes, line 159: It is possible that `output` and/or `error` have data in them here?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
radhgupta: I was thinking of a condition where the file gets locked because some other process might be accessing the same file. In that condition we usually get a prompt " this file is currently opened...please retry", thinking about that scenario I assumed it could be better if we retry.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutor.cs
radhgupta: Yeah, you are right, there could be a scenario where the command fails however we still have output and/or error. I will work on it. Thanks for highlighting this

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
jorgerangel-msft: > @jorgerangel-msft: Is this a normal part of the generation workflow? I'm on the fence for whether we should assume we can/should do installs here or move this up to the interactive space. But, I don't know what's normal.

@jsquire - I think the 95% case is azure SDK users are going to be running the `dotnet build` command which then calls tsp-client. AFAIK, tsp-client handles installing all the dependencies for X emitter. In this case, I can think of 2 pre-requisites needed to run the csharp emitter directly:

1. [installing typespec](https://typespec.io/docs/#install-typespec)
2. installing the emitter

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
jorgerangel-msft: > I attempted that setup, but it still creates a subfolder—named after the emitter—inside the specified output directory. Within that emitter-named folder is the “generated” folder.
> 
> This structure exists so we can clearly identify which emitter produced each piece of code. By default, the system always creates a subfolder using the emitter’s name.

can you share what the output folder structure is even when you specify the output directory? is it `http-client-csharp/src/generated`?

radhgupta: @jorgerangel-msft  on running: ` npx tsp compile .  --emit "@azure-typespec/http-client-csharp" --output-dir "C:\Users\radhgupta\Desktop\azure-sdk-for-net\sdk\testservice\Azure.Data.TestService"`

I got 
image

radhgupta: Currently, we are assuming that the user would already have their tspconfig.yaml. is it fair to assume that the user would have emitter-output-dir already in their tspconfig.yaml??

jorgerangel-msft: > --option "@azure-typespec/http-client-csharp.emitter-ouput-dir={tspOutputPath}""

ah okay, so it looks like typespec adds the emitter name to the output directory by default. Can you try Josh's suggestion and supply the emitter specific option and remove the `output-dir` flag? `--option "@azure-typespec/http-client-csharp.emitter-output-dir={absolute_path}`

jorgerangel-msft: > Currently, we are assuming that the user would already have their tspconfig.yaml. is it fair to assume that the user would have emitter-output-dir already in their tspconfig.yaml??

I'm not sure we can safely assume that.

radhgupta: @jorgerangel-msft @JoshLove-msft  thanks for the feedback, I was not aware that we can set the emitter-output-dir using --option "@azure-typespec/http-client-csharp.emitter-output-dir={absolute_path}. This worked!!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
jsquire: Ok.   I think we can optimize this a bit and cheat:

- Execute a process in the `SdkOutputDirectory`
- Run `git rev-parse --show-toplevel`, which will return the fully qualified root of the repository

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+            if (string.IsNullOrEmpty(azureSdkPath))
+            {
+                throw new InvalidOperationException("Could not locate azure-sdk-for-net directory from SDK output path");
+            }
+
+            return azureSdkPath;
```
jsquire: For this to be the valid repository root, there would need to exist a `.git` directory here.   If that's not true, we're not running in an active clone of the repo.

Something similar to the following would work:

```csharp

static string FindGitRoot(string startDir)
{
    var dir = new DirectoryInfo(startDir);

    while (dir != null)
    {
        if (Directory.Exists(Path.Combine(dir.FullName, ".git")))
        {
            return dir.FullName;
        }

        dir = dir.Parent;
    }

    return null;
}
```

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
```diff
@@ -15,5 +15,16 @@ public AppSettings(IConfiguration configuration)
         public string Model => Configuration.GetSection("AzureSettings:Model").Value ?? "gpt-4o";
         public string AgentName => Configuration.GetSection("AzureSettings:AgentName").Value ?? "AZC Fixer";
         public string AgentInstructions => Configuration.GetSection("AzureSettings:AgentInstructions").Value ?? "";
+        public string TypespecEmitterPackage => Configuration.GetSection("AzureSettings:TypespecEmitterPackage").Value ?? "@typespec/http-client-csharp";
+        public string TspOutputDirectoryName => Configuration.GetSection("AzureSettings:TspOutputDirectoryName").Value ?? "tsp-output";
+        public string TypeSpecDirectoryName => Configuration.GetSection("AzureSettings:TypeSpecDirectoryName").Value ?? "@typespec";
+        public string HttpClientCSharpDirectoryName => Configuration.GetSection("AzureSettings:HttpClientCSharpDirectoryName").Value ?? "http-client-csharp";
+        public string CommandExecutor => Configuration.GetSection("AzureSettings:CommandExecutor").Value ?? "cmd.exe";
```
jsquire: Be very careful here.  Anything in this file is untrusted and could be compromised.   For something like `cmd,exe` that is valid only on Windows, which we can't assume.   This is also one of the things that I'd say would be better to hardcode than allow configurability on for security reasons.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
```diff
@@ -15,5 +15,16 @@ public AppSettings(IConfiguration configuration)
...
+        public string CommandPrefix => Configuration.GetSection("AzureSettings:CommandPrefix").Value ?? "/c";
+        public string AzureSpecRepository => Configuration.GetSection("AzureSettings:AzureSpecRepository").Value ?? "azure-rest-api-specs";
+        public string PowerShellScriptPath => Configuration.GetSection("AzureSettings:PowerShellScriptPath").Value ?? "eng/scripts/automation/Invoke-TypeSpecDataPlaneGenerateSDKPackage.ps1";
```
jsquire: Assuming data plane is fine in the short term, but we'll need to allow for management as well as "data plane using SCM"

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+            string arguments = $"/c pwsh \"{scriptPath}\" " +
+                           $"-sdkFolder \"{SdkOutputDirectory}\" " +
+                           $"-typespecSpecDirectory \"{TypespecSpecDirectory}\" " +
+                           $"-commit \"{CommitId}\" " +
+                           $"-repo \"{AppSettings.AzureSpecRepository}\"";
+
+            (bool success, string output, string error) = await ProcessExecutor.ExecuteAsync(
+                AppSettings.CommandExecutor,
```
jsquire: We definitely don't want to be passing `cmd.exe` in here.   That'll be a big problem across platforms.    We should be able to just run `pwsh` with the args.   The subprocess isn't creating a window and you're redirecting the output.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutor.cs
```diff
@@ -0,0 +1,206 @@
...
+            try
+            {
+                StringBuilder outputBuilder = new StringBuilder(capacity: 4096);
+                StringBuilder errorBuilder = new StringBuilder(capacity: 1024);
+                object outputLock = new object();
+                object errorLock = new object();
+
+                using Process process = CreateProcess(command, arguments, workingDirectory);
+
+                process.OutputDataReceived += (_, e) =>
```
jsquire: I don't think we need these.   We should be able to just slurp everything up after, I think?

```csharp

var startInfo = new ProcessStartInfo
{
    FileName = "<< WHATEVER >>",
    Arguments = arguments,
    RedirectStandardOutput = true,
    RedirectStandardError = true,
    UseShellExecute = false,
    CreateNoWindow = true
};

using var process = new Process { StartInfo = startInfo };
process.Start();

var output = process.StandardOutput.ReadToEnd();
var error = process.StandardError.ReadToEnd();

// But with your cancellation token.  Removed here
// for simplicity.
process.WaitForExit();
```

I think that's the right order.  We may need read after the exit, but I don't think so.   That would remove all the synchronization and such.   We don't need the intermittent form, right?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutor.cs
```diff
@@ -0,0 +1,206 @@
...
+            };
+        }
+
+        /// <summary>
+        /// Gets the current directory. Virtual for testing.
+        /// </summary>
+        /// <returns>The current directory path.</returns>
+        protected virtual string GetCurrentDirectory()
+        {
+            return Directory.GetCurrentDirectory();
```
jsquire: I don't believe that this does what you're trying to do.   I believe we'd want:
`Environment.CurrentDirectory`

That will set the working directory to where the generator tool was launched.

src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/ProcessExecutorTests.cs
```diff
@@ -0,0 +1,302 @@
...
+using NUnit.Framework;
+using System.ComponentModel;
+using System.Diagnostics;
+
+namespace Azure.Tools.GeneratorAgent.Tests
+{
+    [TestFixture]
+    public class ProcessExecutorTests
+    {
+        private const string ValidCommand = "cmd.exe";
```
jsquire: I'd expect the pipeline stages for Linux and macOS to bomb on this.

src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/ProcessExecutorTests.cs
```diff
@@ -0,0 +1,302 @@
...
+        private const string EchoArguments = "/c echo test message";
+        private const string FailArguments = "/c exit 1";
+        private const string ValidWorkingDirectory = @"C:\Windows\System32";
```
jsquire: ... except when not running on Windows.

src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/ProcessExecutorTests.cs
```diff
@@ -0,0 +1,302 @@
...
+        private const int TimeoutMilliseconds = 5000;
+
+        private Mock<ILogger<ProcessExecutor>> mockLogger = null!;
+        private ProcessExecutor executor = null!;
+        private CancellationTokenSource cancellationTokenSource = null!;
+
+        [SetUp]
+        public void SetUp()
+        {
+            mockLogger = new Mock<ILogger<ProcessExecutor>>();
```
jsquire: Do we need class-level state?   Better to keep tests fully isolated and create/destroy everything in the test itself.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Program.cs
```diff
@@ -9,52 +10,130 @@ namespace Azure.Tools.GeneratorAgent
...
+            }
+            catch (Exception ex)
+            {
+                Logger.LogError(ex, "Unexpected error occurred during command execution");
+                return ExitCodeFailure;
+            }
+        }
 
+        private async Task<int> ExecuteGenerationAsync(
+            string? typespecPath,
```
jsquire: Each parameter here should be treated as untrusted.   Validate to the best of our ability.   The TypeSpec path should have an expected `.tsp` extension, for example.

Kicking off a process is very dangerous if we accept any user input related to it.   Be paranoid, and don't trust anything.   Better to get a bug because we're too restrictive than one that we're vulnerable to executing arbitrary commands.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
radhgupta: I see, in that case, I need to explore for management and data plane using SCM.
Can we plan on integrating that in the next PR? Maybe it could be too much for this PR?

src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/ProcessExecutorTests.cs
radhgupta: looking into it.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
jsquire: Yup.   We can get an end-to-end running MVP for data plane and then circle back around.   I just wanted to draw attention to it.  😄

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
jsquire: I was able to validate this on WSL, a stand-alone distro, and Windows.   I'm assuming macOS will work too, but I can't validate locally.

```csharp

var scriptPath = "/path/to/your/script.ps1"; 
var arguments = $"-File \"{scriptPath}\""

var startInfo = new ProcessStartInfo
{
    FileName = "pwsh", // PowerShell Core
    Arguments = arguments,
    RedirectStandardOutput = true,
    RedirectStandardError = true,
    UseShellExecute = false,
    CreateNoWindow = true
 };

using var process = new Process { StartInfo = startInfo };
process.Start();

var output = process.StandardOutput.ReadToEnd();
var error = process.StandardError.ReadToEnd();

process.WaitForExit();

Console.WriteLine("Output:");
Console.WriteLine(output);

if (!string.IsNullOrWhiteSpace(error))
{
    Console.WriteLine("Errors:");
    Console.WriteLine(error);
}

File: src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/ProcessExecutorTests.cs (hunk 29 above)
Comment (jsquire) [reply src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/ProcessExecutorTests.cs]: I just left an example, because I was curious and wanted to test a theory.  😄

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs (hunk 2 above)
Comment (radhgupta) [reply src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs]: Fixed!

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs (hunk 4 above)
Comment (radhgupta) [reply src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs]: fixed

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs (hunk 12 above)
Comment (radhgupta) [reply src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs]: Fixed using a validation class!!

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs (hunk 8 above)
Comment (radhgupta) [reply src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs]: I have added a validation check in InputValidator.cs to ensure there is atleast one tsp file

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs [hunk 33]
```diff
@@ -0,0 +1,87 @@
...
+            {
+                IsRequired = true
+            };
+
+            Option<string?> commitIdOption = new(
+                new[] { "--commit-id", "-c" },
+                "GitHub commit ID to generate SDK from (optional, used with --typespec-dir for GitHub generation)");
+
+            Option<string> sdkDirOption = new(
+                new[] { "--sdk-dir", "-o" },
```
jorgerangel-msft: nit: This is entirely up to you, but the shorthand option and the full option naming could be slightly confusing. I'm wondering if we should just make this option `--output-dir` ?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,354 @@
...
+            Logger.LogInformation("Extracting Azure SDK directory using git rev-parse --show-toplevel");
+
+            try
+            {
+                // Use git to find the repository root directory
+                var result = ProcessExecutor.ExecuteAsync(
+                    SecureProcessConfiguration.GitExecutable,
+                    "rev-parse --show-toplevel",
+                    SdkOutputDir,
+                    CancellationToken.None).GetAwaiter().GetResult();
```
jorgerangel-msft: any reason we are not using async here?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,354 @@
...
+            if (!string.IsNullOrEmpty(azureSdkDirByName))
+            {
+                Logger.LogInformation("Found Azure SDK directory by name: {AzureSdkDir}", azureSdkDirByName);
+                return azureSdkDirByName;
+            }
+
+            throw new InvalidOperationException("Could not locate azure-sdk-for-net directory from SDK output path");
+        }
+
+        protected virtual async Task<bool> RunPowerShellGenerationScript(string azureSdkPath, CancellationToken cancellationToken)
```
jorgerangel-msft: is this used anywhere? If so, this looks very similar to `RunPowerShellGenerationScriptThrows`. If they are intended to have the same core functionality, with the exception of one throwing, should we consolidate them?

jorgerangel-msft (commented):
Overall, the generator invocation and execution LGTM. I just had a few minor comments left.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
radhgupta: My bad, I was trying to change the return type from Task<bool> to Task and ended up pushing both the function. Removed the duplication.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs
radhgupta: Agreed, fixed!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/CommandLineConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+                
+                Logger.LogInformation("All input validation completed successfully");
+                return ExitCodeSuccess;
+            }
+            catch (ArgumentException ex)
+            {
+                Logger.LogError("Input validation failed: {Error}", ex.Message);
+                return ExitCodeFailure;
+            }
+            catch (InvalidOperationException ex)
```
jsquire, line 75: Since you handle the success return case on L68, you could remove all the error returns from the catch block and just do it once after the try/catch closes.   Not really impactful  and doesn't change behavior, but saves on some minor duplication.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
```diff
@@ -1,19 +1,45 @@
...
         }
 
-        public string ProjectEndpoint => Configuration.GetSection("AzureSettings:ProjectEndpoint").Value ?? "";
+        // Azure AI Settings
+        public string ProjectEndpoint => GetRequiredSetting("AzureSettings:ProjectEndpoint");
         public string Model => Configuration.GetSection("AzureSettings:Model").Value ?? "gpt-4o";
         public string AgentName => Configuration.GetSection("AzureSettings:AgentName").Value ?? "AZC Fixer";
         public string AgentInstructions => Configuration.GetSection("AzureSettings:AgentInstructions").Value ?? "";
+        
+        public string TypespecEmitterPackage => "@typespec/http-client-csharp";
```
jsquire, line 27: I think this is a great compromise to get started.    Its already hanging around in configuration, so the impact to change if we need to make it actually configurable is tiny - just here.    Love it!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/ToolConfiguration.cs
```diff
@@ -31,7 +31,9 @@ public virtual ILoggerFactory CreateLoggerFactory()
 
     public virtual AppSettings CreateAppSettings()
     {
-        return new AppSettings(_configuration);
+        using var loggerFactory = CreateLoggerFactory();
```
jsquire: Should we maybe take in an `ILogger` instance as a parameter rather than creating a new factory and spawning one?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        private string ExtractAzureSdkDirFallback()
+        {
+            Logger.LogInformation("Using directory traversal to find Azure SDK directory");
+            
+            string? currentDir = SdkOutputDir;
+            string? azureSdkDirByName = null;
+            
+            while (!string.IsNullOrEmpty(currentDir))
```
jsquire: Give yourself an iteration limit here too.  You don't want to fall victim to a symlink that puts you in an endless recursion.   For example, if we hit 64 levels of depth, I think that's already beyond resonable.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,263 @@
...
+
+        private async Task InstallTypeSpecDependenciesThrows(CancellationToken cancellationToken)
+        {
+            Logger.LogInformation("Installing TypeSpec dependencies globally");
+
+            string arguments;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                // On Windows, use PowerShell to execute npm reliably
+                arguments = $"-Command \"npm install --global {AppSettings.TypespecEmitterPackage}\"";
```
jsquire, line 106: This is interesting.   I would have expected that we could just call directly.   Curious about what you ran into here.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,263 @@
...
+            }
+            
+            ValidationResult argValidation = InputValidator.ValidateProcessArguments(arguments);
+            if (!argValidation.IsValid)
+            {
+                Logger.LogError("Process arguments validation failed: {Error}", argValidation.ErrorMessage);
+                throw new InvalidOperationException($"Process arguments validation failed: {argValidation.ErrorMessage}");
+            }
+
+            string workingDirectory = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
```
jsquire: I'd go with `Path.GetTempPath()` here.   You really don't want to be dumping junk in the system area or directly in the user profile.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/SdkGenerationServiceFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+namespace Azure.Tools.GeneratorAgent
+{
+    /// <summary>
+    /// Factory for creating SDK generation service instances.
+    /// </summary>
+    internal static class SdkGenerationServiceFactory
+    {
+        private static void ValidateCommonParameters(AppSettings appSettings, ILoggerFactory loggerFactory, ProcessExecutor processExecutor)
+        {
+            ArgumentNullException.ThrowIfNull(appSettings);
```
jsquire, line 13: Very minor styling nit:   We're using a bunch of different forms of arg validation in different places.   It would be nice to align the style so that all are consistent.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
```diff
@@ -0,0 +1,281 @@
...
+    /// </summary>
+    internal static class InputValidator
+    {
+        private static readonly Regex CommitIdRegex = new(@"^[a-fA-F0-9]{6,40}$", RegexOptions.Compiled);
+        private static readonly Regex TypeSpecFileRegex = new(@"\.(tsp|yaml)$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+        
+        /// <summary>
+        /// Injection patterns that are considered dangerous in process arguments and configuration values.
+        /// </summary>
+        private static readonly string[] DangerousPatterns =
```
jsquire: While I love the thought, any attempt to deny list patterns is doomed to fail.   For example, I can create a symlink called `happypuppies.exe` that aliases `cmd.exe`.   

You're doing a good job in the code of ensuring that only the app sets the base command and everything else is passed as part of the arguments.    I'd concentrate on that approach - only allow things that you set as a command.

Mainly our goal is to prevent running user input that would allow someone to bypass normal security.   If someone has access to the local system and can change the PS script contents or hacked NPM to upload a malicious package, there's really nothing that we can do.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// <param name="pathType">Description of the path type for logging</param>
+        /// <returns>ValidationResult indicating if path is safe</returns>
+        public static ValidationResult ValidateDirTraversal(string? path, string pathType = "path")
+        {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                return ValidationResult.Invalid($"{pathType} cannot be null or empty");
+            }
+
+            if (path.Contains("..") || path.Contains("~"))
```
jsquire: What if I have my source in `~/src/microsoft/some-project`?   That's a common pattern.

jsquire: Likewise, what happens if I'm in `c:\users\jsquire\src\microsoft\azure-sdk-dotnet\sdk\blah\MyProject` and I want to point you at `..\tsp.config`?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
```diff
@@ -0,0 +1,281 @@
...
+                    
+                    if (!Directory.Exists(fullPath))
+                    {
+                        return ValidationResult.Invalid($"TypeSpec directory not found: {fullPath}");
+                    }
+                    
+                    string[] allFiles = Directory.GetFiles(fullPath, "*.*", SearchOption.TopDirectoryOnly);
+                    string[] typeSpecFiles = allFiles
+                        .Where(file => TypeSpecFileRegex.IsMatch(file))
+                        .ToArray();
```
jsquire:
```suggestion
                    bool areTypeSpecFiles = allFiles
                        .Any(file => TypeSpecFileRegex.IsMatch(file));
```

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
```diff
@@ -0,0 +1,281 @@
...
+                        
+                    if (typeSpecFiles.Length == 0)
+                    {
+                        return ValidationResult.Invalid($"No .tsp or .yaml files found in directory: {fullPath}");
+                    }
+                    
+                    string[] nonTypeSpecFiles = allFiles
```
jsquire: Likewise here.   Any time you call `ToArray`, you're allocating a new array and making a full eager copy of the sequence.   Prefer `Any` to  `Length` for these, as will just enumerate the set and exit when an item is found (or at the end) without new allocation and copying.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/SecureProcessConfiguration.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// <summary>
+    /// Provides secure, hardcoded process execution configuration for cross-platform support.
+    /// No user-configurable values to prevent security vulnerabilities.
+    /// </summary>
+    internal static class SecureProcessConfiguration
+    {
+        /// <summary>
+        /// Gets the PowerShell Core executable name (cross-platform).
+        /// </summary>
+        public static string PowerShellExecutable => "pwsh";
```
jsquire, line 16: This is beautiful.   The allow-list approach of setting the executable directly is the safest way to do this.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
radhgupta: On Windows, npm is not directly executable - it's actually npm.cmd (a batch file). When you try to execute npm directly, the system can't find it because npm is installed as npm.cmd on Windows I think ProcessExecutor is likely looking for an .exe file therefore The PATH resolution isn't working the same way as it would in a shell.

jsquire: interesting.   I wonder if we can set the process as `npm.cmd` and then pass `install --global` as the arguments.   If we can, that may be helpful, but it, then this is a clever fallback.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
radhgupta: Agreed!
I am planning on  simplifying the validation from a complex denylist to just basic command separators: &, |, ;, &&, || and  removing the impossible-to-maintain patterns like cmd, powershell, bash, etc

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
radhgupta: e.g. it blocks: install @typespec/compiler; rm -rf /abc/data

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
radhgupta: e.g it still blocks:  install @typespec/compiler; rm -rf /abc/data

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
radhgupta: From a security perspective, what's your take? Are we being overly cautious about ~, or is there a real attack vector I should be worried about?

jsquire: I'd loosen a bit.   If someone gives us a path to something that passes a quick sniff test for "does it exist and does it have the extension that I think it should" then I think we've done our diligence.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
jsquire: Makes sense to me.   From what Larry told us, Process does some complex guarding against these, but it certainly can't hurt to have extra coverage there.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,303 @@
...
+                    SdkOutputDir,
+                    CancellationToken.None).ConfigureAwait(false);
+
+                if (!result.Success)
+                {
+                    Logger.LogWarning("Git command failed, falling back to directory traversal. Error: {Error}", result.Error);
+                    return ExtractAzureSdkDirFallback();
+                }
+
+                string gitRoot = result.Output.Trim();
```
jsquire:
```suggestion
                string gitRoot = result.Output?.Trim();
```

That'll guard against a potential `null` here and only apply the trim if its safe.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutor.cs
```diff
@@ -0,0 +1,186 @@
...
+                    {
+                        await process.WaitForExitAsync(cts.Token).ConfigureAwait(false);
+                    }
+                    catch (OperationCanceledException) when (cts.Token.IsCancellationRequested && !cancellationToken.IsCancellationRequested)
+                    {
+                        Logger.LogError("Process timed out after {Timeout}ms: {Command}", timeout.Value.TotalMilliseconds, command);
+
+                        try
+                        {
+                            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
jsquire: May be helpful to have a comment here as to why we're doing a platform sniff.   I think you could rewrite to simplify

```csharp
process.Kill(entireProcessTree: RuntimeInformation.IsOSPlatform(OSPlatform.Windows));
```

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
```diff
@@ -0,0 +1,274 @@
...
+                    if (!hasTypeSpecFiles)
+                    {
+                        return ValidationResult.Invalid($"No .tsp or .yaml files found in directory: {fullPath}");
+                    }
+                    
+                    if (allFiles.Any(file => !TypeSpecFileRegex.IsMatch(file)))
+                    {
+                        string[] nonTypeSpecFiles = allFiles
+                            .Where(file => !TypeSpecFileRegex.IsMatch(file))
+                            .ToArray();
```
jsquire:
```suggestion
```

No need (or value) in doing an array allocation here.   The `string.Join` that you're calling has an `IEnumerable<string>` overload and will walk the set without you having to materialize it first.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// </summary>
+        public static ValidationResult ValidatePowerShellScriptPath(string scriptPath, string azureSdkPath)
+        {
+            if (string.IsNullOrWhiteSpace(scriptPath))
+            {
+                return ValidationResult.Invalid("PowerShell script path cannot be null or empty");
+            }
+
+            // Ensure it's a PowerShell script
+            if (!scriptPath.EndsWith(".ps1", StringComparison.OrdinalIgnoreCase))
```
jsquire:
```suggestion
            if (!string.Equals(Path.GetExtension(scriptPath), ".ps1", StringComparison.OrdinalIgnoreCase))
```

You'll want to use the `Path` built-ins where you can.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/SecureProcessConfiguration.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <summary>
+        /// Gets the Node.js executable name.
+        /// </summary>
+        public static string NodeExecutable => "node";
+        
+        /// <summary>
+        /// Gets the NPM executable name.
+        /// On Windows, use PowerShell to execute npm for reliable execution. On other platforms, use npm.
+        /// </summary>
+        public static string NpmExecutable => RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? PowerShellExecutable : "npm";
```
jsquire: This form will run every time, even though platform won't change.

```csharp
public static string NpmExecutable { get; } =  RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? PowerShellExecutable : "npm";
```

That form will execute once to initialize and then use the cached value.   Since you can't change platforms while running, I'd suggest converting them over.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/ValidationContext.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public static ValidationContext ValidateAndCreate(
+            string? typespecPath, 
+            string? commitId, 
+            string sdkOutputPath, 
+            ILogger logger)
+        {
+            logger.LogInformation("Starting input validation...");
+
```
jsquire, line 31:
```suggestion
```

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -10,16 +10,20 @@ internal class ErrorFixerAgent : IAsyncDisposable
         private readonly ILogger<ErrorFixerAgent> Logger;
         private readonly PersistentAgentsAdministrationClient AdminClient;
         private readonly Lazy<PersistentAgent> Agent;
-        private bool _disposed = false;
+        private bool Disposed = false;
```
jsquire, line 13:
```suggestion
        private volatile bool _disposed = false;
```

This one isn't `readonly` and does get mutated.   We'll want to use the standard variable format.   The `volatile` keyword here makes sure that there's a read/write barrier and threads don't interact with a cached value.   Not strictly needed here, since we're not doing any threading - but its a good habit to get into for things like this.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -0,0 +1,303 @@
...
+            string scriptPath = Path.Combine(azureSdkPath, AppSettings.PowerShellScriptPath);
+
+            ValidationResult scriptValidation = InputValidator.ValidatePowerShellScriptPath(AppSettings.PowerShellScriptPath, azureSdkPath);
+            if (!scriptValidation.IsValid)
+            {
+                Logger.LogError("PowerShell script validation failed: {Error}", scriptValidation.ErrorMessage);
+                throw new InvalidOperationException($"PowerShell script validation failed: {scriptValidation.ErrorMessage}");
+            }
+
+            List<string> arguments = new List<string>
```
jsquire, line 159:
```suggestion
            string[] arguments = new[]
```

No value in an expandable set here, since you've got a static initialization and are immediately turning into a single string.

src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/AppSettingsTests.cs
```diff
@@ -22,175 +23,141 @@ private static Mock<IConfigurationSection> CreateConfigurationSectionMock(string
...
+            return new AppSettings(configuration ?? CreateConfigurationMock().Object, mockLogger.Object);
         }
 
         [Test]
         public void Constructor_WithNullConfiguration_ThrowsArgumentNullException()
         {
-            // Act & Assert
-            var ex = Assert.Throws<ArgumentNullException>(() => new AppSettings(null!));
-            Assert.That(ex.ParamName, Is.EqualTo("configuration"));
+            var mockLogger = new Mock<ILogger<AppSettings>>();
```
jsquire, line 33:
```suggestion
            var mockLogger = Mock.Of<ILogger<AppSettings>>();
```
No real difference, other than you don't have to `.Object` this.   Since you're not setting up behavior expectations, this a simpler form.    Your call on whether to use or not, but figured that I'd mention it being available.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
radhgupta: Fixing it in next PR.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
radhgupta: Fixing it in next PR.

src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/AppSettingsTests.cs
radhgupta: fixing it in the next PR.

PR #11326, opened 2025-07-24
Pinned version azsdk

jeo02:
## Summary
- Added in a pinned version of the azsdk
- Added testing for changes made to the script/version. 
- Used discussion from here: https://github.com/Azure/azure-sdk-tools/pull/11142#discussion_r2196232535


eng/common/mcp/test-azure-sdk-mcp.ps1
```diff
@@ -0,0 +1,163 @@
+<#
```
weshaggard, line 1: We put eng/common tests under the eng/common-test directory to avoid syncing them to all the repos.

eng/common/mcp/test-azure-sdk-mcp.yml
```diff
@@ -0,0 +1,35 @@
+trigger:
```
weshaggard, line 1: We shouldn't need this as we already have an eng/common test pipeline

eng/common/scripts/Helpers/AzSdkTool-Helpers.ps1
```diff
@@ -136,6 +136,16 @@ function Install-Standalone-Tool (
 
     $tag = "${Package}_${Version}"
 
+    # Check for pinned version for azsdk package
+    if (!$Version -or $Version -eq "*") {
```
weshaggard, line 140: @jeo02 @benbp I thought we decided to not pin this tool and only pull the latest. If ever needed to pint the version we would add a default version inline in the azure-sdk-mcp.ps1 script itself. Was there another conversation about this that I'm not aware of?

eng/common/mcp/test-azure-sdk-mcp.yml
jeo02: Ah okay I see that makes a lot more sense I was wondering how test-proxy was triggered 🤔

weshaggard (changes requested):
I'm fine with the tests once you move to the eng/common-test folder but I want to understand the pinning plans before we do that part of this PR.

eng/common/scripts/Helpers/AzSdkTool-Helpers.ps1
weshaggard: Reading over the comments again at https://github.com/Azure/azure-sdk-tools/pull/11142#discussion_r2196232535 I think forgot to put the final update in that comment.

Here was our latest chat:
image

jeo02: Yup last we had talked we had agreed to use latest. However wouldn't it be safer to have a pinned version? 

I did see that Ben made a PR recently for a publishing pipeline as well so they are not auto released. Personally I like the idea also of having integration tests for having updating to a new version just so we make sure the new release is bare minimum functioning. 

We can also have something to the extent of the version inside the azure-sdk-mcp script but theres not to much of a difference against having it in a txt file how we did here?

@praveenkuttappan

jeo02:
Closing after talking to praveen.

PR #11330, opened 2025-07-24
Add shebang and make cspell script executable for bash invocation

benbp:
Add shebang and make cspell script executable for bash invocation


weshaggard (approved):

eng/common/spelling/Invoke-Cspell.ps1
```diff
@@ -1,3 +1,5 @@
+#!/bin/env pwsh
```
benbp:
```suggestion
#!/usr/bin/env pwsh
```

PR #11333, opened 2025-07-24
Install azsdk cli to a place in PATH

### LINKED ISSUE #11348: [azsdk-cli] Add PATH setup to install script
Make it easier to get started using the CLI by auto-updating the user's shell profile to add the tool installation to PATH

benbp:
This way the tool is usable immediately on its own as a CLI tool after running the install script. The installer script can still be used with `-Run` for mcp server mode.

Also adding `IncludeNativeLibrariesForSelfExtract` to bundle the libgit2sharp .so file so we can be a truly single file executable. This can be removed after #11344 is implemented.

Resolves #11348

### DIFF
--- a/eng/common/mcp/azure-sdk-mcp.ps1
+++ b/eng/common/mcp/azure-sdk-mcp.ps1
@@ -65,16 +65,12 @@ Copy-Item -Path $tempExe -Destination $toolInstallDirectory -Force
 $exeName = Split-Path $tempExe -Leaf
 $exe = Join-Path $toolInstallDirectory $exeName
 
-if (!$IsLinux) {
-    Write-Host "Package $package is installed at $exe"
-    if (!$UpdatePathInProfile) {
-        Write-Warning "To add the tool to PATH for new shell sessions, re-run with -UpdatePathInProfile to modify the shell profile file."
-    } else {
-        Add-InstallDirectoryToPathInProfile -InstallDirectory $toolInstallDirectory
-        Write-Warning "'$exeName' will be available in PATH for new shell sessions."
-    }
+Write-Host "Package $package is installed at $exe"
+if (!$UpdatePathInProfile) {
+    Write-Warning "To add the tool to PATH for new shell sessions, re-run with -UpdatePathInProfile to modify the shell profile file."
 } else {
-    Write-Host "Package '$package' is installed at '$exe' and can be run directly by typing '$exeName'"
+    Add-InstallDirectoryToPathInProfile -InstallDirectory $toolInstallDirectory
+    Write-Warning "'$exeName' will be available in PATH for new shell sessions."
 }
 
 if ($Run) {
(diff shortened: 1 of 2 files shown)


eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -8,23 +11,61 @@ param(
...
-if (-not $InstallDirectory)
-{
-    $homeDir = if ($env:HOME) { $env:HOME } else { $env:USERPROFILE }
-    $InstallDirectory = (Join-Path $homeDir ".azure-sdk-mcp" "azsdk")
-}
 . (Join-Path $PSScriptRoot '..' 'scripts' 'Helpers' 'AzSdkTool-Helpers.ps1')
 
-if ($Clean) {
-    Clear-Directory -Path $InstallDirectory
+function getInstallDirectory {
```
weshaggard:
```suggestion
function Get-InstallDirectory {
```

benbp: I named it this way to kind of mark it as private vs. the powershell external cmdlet convention.

benbp: but actually this could be useful elsewhere. Maybe I should move this to common

eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -8,23 +11,61 @@ param(
     [string]$Repository = 'Azure/azure-sdk-tools',
     [string]$RunDirectory = (Resolve-Path (Join-Path $PSScriptRoot .. .. ..)),
     [switch]$Run,
-    [switch]$UpdateVsCodeConfig,
-    [switch]$Clean
+    [switch]$UpdateVsCodeConfig
```
weshaggard: Let's just change this to "AddToPath" and it will only persist for until the shell is closed. I think that is better than trying to mutate the PS Profile script.

eng/common/scripts/Helpers/AzSdkTool-Helpers.ps1
```diff
@@ -192,3 +192,64 @@ function Install-Standalone-Tool (
...
+
+    # Update PATH in current session
+    if (-not ($env:PATH -like "*$InstallDirectory*")) {
+        $env:PATH += ";$InstallDirectory"
+    }
+
+    return $installDirectory
+}
+
+function Add-InstallDirectoryToPathInProfile(
```
weshaggard, line 210: I'm personally would not have this but I'm fine keeping given it is opt-in.

eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -54,13 +49,30 @@ if ($UpdateVsCodeConfig) {
...
+    -Directory $tempInstallDirectory `
     -Repository $Repository
 
+Copy-Item -Path $tempExe -Destination $toolInstallDirectory -Force
+$exeName = Split-Path $tempExe -Leaf
+$exe = Join-Path $toolInstallDirectory $exeName
+
+Write-Host "Package $package is installed at $exe"
+if (!$UpdatePathInProfile) {
+    Write-Warning "To add the tool to PATH for new shell sessions, re-run with -UpdatePathInProfile to modify the shell profile file."
```
weshaggard, line 70: You are outputting a warning in both cases.

PR #11336, opened 2025-07-24
[perf.yaml] Install .NET from tools repo global.json

mikeharder:


eng/common/pipelines/templates/jobs/perf.yml
```diff
@@ -112,8 +112,29 @@ jobs:
...
+
+  # Copied from eng/pipelines/templates/steps/install-dotnet.yml, but changed workingDirectory
+  - task: UseDotNet@2 # About UseDotNet@2 task: https://learn.microsoft.com/azure/devops/pipelines/tasks/reference/use-dotnet-v2?view=azure-pipelines
+    displayName: "Use .NET SDK from global.json"
+    retryCountOnTaskFailure: 3
+    inputs:
+      useGlobalJson: true
+      workingDirectory: azure-sdk-tools
+
+  # Copied from eng/pipelines/templates/steps/install-dotnet.yml, but changed workingDirectory
```
weshaggard, line 124: I don't expect this should be needed as long as the tool works on .NET 9 which is what is in global.json

mikeharder: Without this, the tool failed because it wants .NET 8.  I can upgrade the tool to .NET 9 if it would help.

mikeharder: PerfAutomation is still targeting .NET 8.  Should I update it?  I thought .NET tools in the tools repo were updated centrally.

https://github.com/Azure/azure-sdk-tools/blob/d5037b7f2f46dd17ce6d3e8ecf02057f89364307/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Azure.Sdk.Tools.PerfAutomation.csproj#L4

mikeharder: @weshaggard: In this run, you can see the Net job passed (because the Net repo itself installs .NET 8), but the other langs did not.

```
You must install or update .NET to run this application.

App: /mnt/vss/_work/1/s/azure-sdk-tools/artifacts/bin/Azure.Sdk.Tools.PerfAutomation/Debug/net8.0/Azure.Sdk.Tools.PerfAutomation
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '8.0.0' (x64)
.NET location: /opt/hostedtoolcache/dotnet

The following frameworks were found:
  9.0.1 at [/opt/hostedtoolcache/dotnet/shared/Microsoft.NETCore.App]
```

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5136452&view=logs&j=dc42670d-1473-57cd-faea-36a5bc05eab2&t=112ca725-a100-5ce3-182a-396c0d166ac3&l=18

If we can update PerfAutomation to run on net9.0, I have no problem doing so, and removing the net8.0 install.

mikeharder: I do think the .NET team may want to run perf tests on both .NET 8 and 9, but the automation runs separate dotnet processes for each perf test, so it shouldn't matter whether the automation itself runs on 8 or 9.

weshaggard: .NET 8.0 is the current LTS so I think it is best to stick with that. However, in order to use new tool features we updated the SDK in global.json to be 9.x. In order to run .NET 8.0 apps you will need the .NET 8 runtime which is what this second step is using or you could set the [roll-forward option](https://learn.microsoft.com/en-us/dotnet/core/runtime-discovery/troubleshoot-app-launch?pivots=os-windows#configure-roll-forward-behavior). I'm good with either approach.

PR #11340, opened 2025-07-25
library to fix analyzer, general errors

radhgupta:


src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/ErrorAnalyzerService.cs
```diff
@@ -0,0 +1,106 @@
...
+                catch (Exception)
+                {
+                    continue;
+                }
+            }
+
+            return allAnalyzers.AsReadOnly();
+        }
+
+        private static readonly ConcurrentBag<IAnalyzerProvider> AnalyzerProviders = new ConcurrentBag<IAnalyzerProvider>();
```
radhgupta: nit: Should be initialized at the top

src/dotnet/ErrorAnalyzers/Client/src/Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers/ClientRuleAnalyzerRegistry.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// Provider for client rule analyzers.
+    /// This is the primary way to access client-specific rule analyzers.
+    /// </summary>
+    public sealed class ClientAnalyzerProvider : IAnalyzerProvider
+    {
+        private static readonly IReadOnlyList<AgentRuleAnalyzer> clientAnalyzers = new AgentRuleAnalyzer[]
+        {
+            new AZC0012RuleAnalyzer(),
```
jsquire: Not necessary for this set, but we should discover these rather than forcing developers to keep the list up to date.   Because - they won't.   😄 

If we're going to defer, can we please add an issue to circle back?   I don't want us to forget.

src/dotnet/ErrorAnalyzers/Client/src/Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers/ErrorAnalyzers/AZC0012RuleAnalyzer.cs
```diff
@@ -0,0 +1,53 @@
...
+using Azure.Tools.ErrorAnalyzers;
+
+namespace Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers
+{
+    /// <summary>
+    /// Handles AZC0012: “Type name '{0}' is too generic… Consider using a more descriptive multi-word name, such as '{1}'.”
+    /// </summary>
+    internal sealed class AZC0012RuleAnalyzer : AgentRuleAnalyzer
+    {
+        private static readonly Regex ErrorMessageRegex = new Regex(
```
jsquire: This scares me, as it's brittle and will stop working if the rule text changes.    Ideally, we'd analyze this in a way that the intent can be interpreted in real time, such as by LLM.    At worst, we should anchor on a set of conventions for renames where we document/update the analyzer messages to have a known pattern for automation.   Maybe a suffix of something like:

`RENAME:  '{0]'  to  '{1}'`, then the `CanFix` check  for both the AZC code AND the "RENAME:" text?

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/ErrorAnalyzerService.cs
```diff
@@ -0,0 +1,106 @@
...
+
+namespace Azure.Tools.ErrorAnalyzers
+{
+    /// <summary>
+    /// Main entry point for the Error Analyzer library.
+    /// Provides a unified interface for analyzing errors and generating fixes.
+    /// </summary>
+    public static class ErrorAnalyzerService
+    {
+
```
jsquire, line 18:
```suggestion
```

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/ErrorAnalyzerService.cs
```diff
@@ -0,0 +1,106 @@
...
+        private static IReadOnlyList<AgentRuleAnalyzer> CreateAllAnalyzers()
+        {
+            var allAnalyzers = new List<AgentRuleAnalyzer>();
+
+            foreach (IAnalyzerProvider provider in AnalyzerProviders)
+            {
+                try
+                {
+                    var providedAnalyzers = provider.GetAnalyzers().ToList();
```
jsquire:
```suggestion
                    var providedAnalyzers = provider.GetAnalyzers();
```

`AddRange` is going to enumerate the set.  There's no need to `ToList` here, as that will allocate a list that you're going to just throw away.

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/ErrorAnalyzerService.cs
```diff
@@ -0,0 +1,106 @@
...
+        public static void RegisterProvider(IAnalyzerProvider provider)
+        {
+            ArgumentNullException.ThrowIfNull(provider);
+            AnalyzerProviders.Add(provider);
+        }
+
+        /// <summary>
+        /// Checks if any analyzer can handle the specified error type.
+        /// </summary>
+        public static bool CanHandle(string errorType)
```
jsquire: What do we see as the use case for this?   The `Any` call is fairly expensive and its hiding behind a simple interface.   It may be better to do something like `GetAnalyzersFor(errorType)` and return it as an `IEnumerable<AgentRuleAnalyzer>` so that you can execute lazily.

src/dotnet/ErrorAnalyzers/Management/tests/Azure.Tools.ErrorAnalyzers.ManagementRuleAnalyzers.Tests/Azure.Tools.ErrorAnalyzers.ManagementRuleAnalyzers.Tests.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <TargetFramework>net9.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
+    <PackageReference Include="xunit" Version="2.9.2" />
```
jsquire: Please move over to Nunit.   The xUnit runner exists only for legacy compat.

src/dotnet/ErrorAnalyzers/samples/basic-usage/AdvancedExample.cs
```diff
@@ -0,0 +1,160 @@
+using Azure.Tools.ErrorAnalyzers;
+using Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers;
+using Azure.Tools.ErrorAnalyzers.GeneralRuleAnalyzers;
+using Azure.Tools.ErrorAnalyzers.ManagementRuleAnalyzers;
+
+namespace BasicUsage;
```
jsquire: Nice idea to include!

src/dotnet/ErrorAnalyzers/Client/src/Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers/ClientRuleAnalyzerRegistry.cs
radhgupta: I think, we should fix this right away!!

src/dotnet/ErrorAnalyzers/Client/src/Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers/ErrorAnalyzers/AZC0012RuleAnalyzer.cs
```diff
@@ -0,0 +1,57 @@
...
+        public override Fix? GetFix(RuleError error)
+        {
+            if (error?.message == null)
+                throw new ArgumentException("Error message cannot be null.", nameof(error));
+            if (!CanFix(error))
+            {
+                return null;
+            }
+
+            // TODO: figure out a way to move the hard coded prompts and context into a configurable setup.
```
radhgupta, line 32: I am thinking of having a json file to store instruction, example, background etc. And then use a class to parse the json and generate the prompt.
This way it becomes maintainable in case someone needs to change the prompt or add new instruction, example, background

jsquire: That's a good approach to start.   We could create an appSettings file for prompts and just to an .AddConfig when creating `IConfiguration`

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/FixAction.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// <summary>
+        /// Instruct the AI agent with a prompt snippet.
+        /// </summary>
+        AgentPrompt,
+
+        /// <summary>
+        /// A simple in‐code rename, containing Context = from, Action = to.
+        /// Currently not in use, maybe used later as the project grows
+        /// </summary>
+        Rename,
```
jsquire: We probably want to pull this out and just have `AgentPrompt` for now.

src/dotnet/ErrorAnalyzers/Core/tests/Azure.Tools.ErrorAnalyzers.Tests/Azure.Tools.ErrorAnalyzers.Tests.csproj
```diff
@@ -0,0 +1,26 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.6.0" />
+    <PackageReference Include="NUnit" Version="3.14.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="9.0.0" />
+    <PackageReference Include="coverlet.collector" Version="6.0.0">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
```
jsquire:
```suggestion
```

I don't think we need code coverage here.

src/dotnet/ErrorAnalyzers/Management/tests/Azure.Tools.ErrorAnalyzers.ManagementRuleAnalyzers.Tests/Azure.Tools.ErrorAnalyzers.ManagementRuleAnalyzers.Tests.csproj
```diff
@@ -0,0 +1,24 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
+    <PackageReference Include="NUnit" Version="4.0.1" />
+    <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
+    <PackageReference Include="coverlet.collector" Version="6.0.2">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
```
jsquire:
```suggestion
```

jsquire (approved):

PR #11347, opened 2025-07-25
Added generateFromTypeSpec to execution report

raych1:
This information will be used to determine if the generation is based on TypeSpec then set the required settings for .NET.


scbedd (approved):
Never been a fan of the `initiate object with variable that is also the name of the map item` but that's not a blocker 😄

Issue #10929, opened 2025-06-18
[tsp-client] Make 'init' command support existing package scenario


raych1:
Here's the latest discussion result:

Update the tsp-client update command to initialize the sdk if it hasn't been already created already, otherwise use the existing tsp-location values (this includes the alternate emitter package json path variable), if a `tspconfig.yaml` is passed in with a certain commit the commit will be updated.

catalinaperalta:
@raych1 when we added support for an alternate emitter-package-json in the tool we also added a flag in the `init` command so that folks could pass in an alternate path to a emitter-package.json file, this value is also saved to tsp-location.yaml. I'm not clear on what else you need here? Also, whichever emitter is configured in the target emitter-package.json, then the options in tspconfig.yaml will be retrieved for that specific emitter.

JoshLove-msft:
We would like for the new emitter to be used for new data plane packages, i.e. if there is no existing tsp-location.yaml file. @catalinaperalta, is there a command to check to see if the tsp-location.yaml exists? 

For packages that do have a tsp-location.yaml, we want the emitterPackageJsonPath value to be honored from there. It seems that calling Init will wipe out the emitterPackageJsonPath value in tsp-location.  I think we need for Init to retain the emitterPackageJsonPath when writing tsp-location.yaml. Alternatively, if there is a way to look up the tsp-location using tsp-client, we can just parse it ourselves then pass that as an argument to Init. Since we need to do this anyways to figure out if it is a new project or not, this might be the best path forward.

catalinaperalta:
> We would like for the new emitter to be used for new data plane packages, i.e. if there is no existing tsp-location.yaml file. @catalinaperalta, is there a command to check to see if the tsp-location.yaml exists?

There is no command to check if tsp-location.yaml exists, but wouldnt you know what a new package is if it doesnt exist in the repo yet? Therefore, you'd run tsp-client init for any new packages?

Also, it sounds like this new emitter should be the default one? And the older one should be the alternate emitter? If you have a list of the older packages that still rely on the older emitter, cant you pass that list into your script so you pass in the alternate emitter-package.json file path to the init command?

> Since we need to do this anyways to figure out if it is a new project or not, this might be the best path forward.

[...]

PR #11349, opened 2025-07-25
[tsp-client] Support init or update

catalinaperalta:
Fixes https://github.com/Azure/azure-sdk-tools/issues/10929

Add `--update-if-exists` flag to `tsp-client init`

### DIFF
--- a/tools/tsp-client/CHANGELOG.md
+++ b/tools/tsp-client/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2025-07-29 - 0.26.1
 
-- Support an `--update-if-exists` flag when using `tsp-client init`. This flag will update the tspconfig.yaml file based on new inputs to the command and keep any other existing config in the file.
+- Support an `--update-if-exists` flag when using `tsp-client init`. This flag will update the tsp-location.yaml file based on new inputs to the command and keep any other existing config in the file.
 
 ## 2025-07-21 - 0.26.0
 


tools/tsp-client/src/utils.ts
```diff
@@ -102,3 +102,43 @@ export async function writeTspLocationYaml(
...
+      'entrypointFile',
+      'additionalDirectories',
+      'emitterPackageJsonPath'
+    ];
+
+    // Update each property if it has a valid value
+    for (const property of updatableProperties) {
+      const value = tspLocationData[property];
+      if (shouldUpdate(value)) {
+        (updatedTspLocation as any)[property] = value;
```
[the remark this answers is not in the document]
catalinaperalta: Without this ts raises an error because it believes value might be undefined although we check for that in the shouldUpdate call

timotheeguerin, line 130: You can update should update return type to be : T. Which will tell typescript you checked the value is defined

catalinaperalta: I moved the check inline after looking at the code again. `shouldUpdate()` wasnt really adding anything. So the undefined check is inline now, but there's still an issue where we need to use the `as any` for updatedTspLocation. This is the error ts raises if I remove the as any: Type 'string | string[]' is not assignable to type 'string & string[]'.
  Type 'string' is not assignable to type 'string & string[]'.
    Type 'string' is not assignable to type 'string[]'.ts(2322)

timotheeguerin (approved):

PR #11351, opened 2025-07-25
Add emitterPackageJsonPath to tsp-location docs

JoshLove-msft:


weshaggard (approved):

weshaggard:
@catalinaperalta @raych1 is this the best place for this documentation?

I know we plan to remove the associated scripts of the same name. That said I don't have any issue adding it here in the meantime.

PR #11353, opened 2025-07-25
[azsdk-cli] Fixes for mac install

benbp:
It seems like `/usr/bin/env` is more conventional than `/bin/env` and not all users have the latter.

### DIFF
--- a/eng/common/mcp/azure-sdk-mcp.ps1
+++ b/eng/common/mcp/azure-sdk-mcp.ps1
@@ -1,4 +1,4 @@
-#!/bin/env pwsh
+#!/usr/bin/env pwsh
 
 #Requires -Version 7.0
 #Requires -PSEdition Core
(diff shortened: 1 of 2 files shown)


eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -1,4 +1,4 @@
-#!/bin/env pwsh
+#!/usr/bin/env pwsh
```
scbedd, line 1: Do we want to change this generally?

scbedd (approved):

eng/common/mcp/azure-sdk-mcp.ps1
benbp: @scbedd yeah I'll put up another PR

praveenkuttappan (approved):

weshaggard (approved):

PR #11354, opened 2025-07-25
Change instances of /bin/env shebang to /usr/bin/env

benbp:
Change instances of /bin/env shebang to /usr/bin/env


scbedd (approved):

richardpark-msft:
This is when I find out that there ever is a /bin/env. I'm glad I've only ever used the "correct" one, I suppose (by luck)

PR #11363, opened 2025-07-28
[tspconfig] Enable inheritance in tspconfig

VeryEarly:


tools/tsp-client/src/utils.ts
```diff
@@ -48,6 +52,314 @@ export function getServiceDir(configYaml: any, emitter: string): string {
...
+  }
+  
+  // Fallback - return as-is if we can't determine the context
+  return dirPath;
+}
+
+/**
+ * Loads and merges TypeSpec configuration with inheritance support
+ */
+export async function loadTspConfig<T>(
```
timotheeguerin: Why are we reinventing the parsing of the tspconfig.yaml. This will inevitably be out of sync and create unfixable divergence in how things behave. 

Why can't we use the compiler api to resolve the config

tools/tsp-client/src/commands.ts
timotheeguerin, line 1: Can you also explain a bit more what is the purpose of this change

VeryEarly: Hi Timothee, I have added you in an internal email thread, we can continue the conversation over there

tools/tsp-client/src/utils.ts
VeryEarly: sry I'm not aware of such API, do you mind point me the source code?

tools/tsp-client/src/commands.ts
timotheeguerin: I was forwarded the email, and as cat replied there is not enough info on the email.
Can you show a full flow of what you expect to happen where files are, how they are written. 
How this is different from the compiler existing config extends

tools/tsp-client/src/utils.ts
timotheeguerin: https://typespec.io/docs/standard-library/reference/js-api/functions/resolvecompileroptions/#_top

PR #11382, opened 2025-07-29
Perf regression agent

sa7936:
--A tool used to find performance regressions in Azure SDK libraries.
--Cherry-picked to get rid of random edits to files unrelated to the agent.


tools/perf-regression-finder/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Walkthrough
+1. Run `dotnet run` in the `perf-regression-finder` directory.
+```
+> dotnet run
+Welcome to the Performance Regression Finder Agent! Enter a blank line to send
+```
+2. Enter your performance regression scenario or question:
+```
+User > when sending a byte array to 'GetMessageContentAsync' speciically argetting OpenAi Gpt 4.1. i can see the memory usage spike by 80x the original byte arrays size.
```
m-redding:
```suggestion
User > when sending a byte array to 'GetMessageContentAsync' specifically targetting OpenAi Gpt 4.1, I can see the memory usage spike by 80x the original byte arrays size. 
```
nit

tools/perf-regression-finder/README.md
```diff
@@ -0,0 +1,171 @@
...
+## Walkthrough
+1. Run `dotnet run` in the `perf-regression-finder` directory.
+```
+> dotnet run
+Welcome to the Performance Regression Finder Agent! Enter a blank line to send
+```
+2. Enter your performance regression scenario or question:
+```
+User > when sending a byte array to 'GetMessageContentAsync' speciically argetting OpenAi Gpt 4.1. i can see the memory usage spike by 80x the original byte arrays size. 
+propose a root cuase by searching through the azure.core and system.clientmodel repos as well as a call stack
```
m-redding:
```suggestion
propose a root cause by searching through the Azure.Core and System.Clientmodel source code, as well as a call stack
```
nit

tools/perf-regression-finder/src/SemanticAzureAi.cs
```diff
@@ -0,0 +1,163 @@
...
+        {
+            // Load configuration from appsettings.json
+            config = _config;
+        }
+        public async Task<AzureAIAgent> CreateAgentAsync()
+        {
+            var deployment = config["ModelDeployment"];
+            var endpoint = config["AzureAIFoundryProjectEndpoint"];
+            var agentId = config["AzureAIFoundryAgentId"];
+            PersistentAgentsClient client = AzureAIAgent.CreateAgentsClient(endpoint!, new DefaultAzureCredential());
```
m-redding: nit but might be worth considering throwing with a specific error message that "AzureAIFoundryProjectEndpoint can't be null" or something along those lines, instead of using the null forgiving `!`

m-redding: unless it's ok if it's null / CreateAgentsClient does that for you

tools/perf-regression-finder/src/SemanticAzureAi.cs
```diff
@@ -0,0 +1,163 @@
...
+                    4. If direct fixes are not possible, report the issues to the maintainers of `Microsoft.SemanticKernel` and `Azure.Core` to address inefficiencies with their serialization approach.
+
+
+                    """
+                );
+                // Update the appsettings.json with the new agent ID
+                UpdateAppSettings("./bin/Debug/net8.0/appsettings.json", "AzureAIFoundryAgentId", definition.Id);
+            }
+            // 2. Create plugins for the agent (Replace with your actual local paths)
+            var azureCore = new RepoAccessPlugin(@"C:\Users\t-akpoyibos\source\repos\azure-sdk-for-net\sdk\core\Azure.Core\src");
```
m-redding, line 105: Could the agent prompt for this at startup rather than it needing to be hardcoded?

tools/perf-regression-finder/src/SemanticAzureAi.cs
```diff
@@ -0,0 +1,163 @@
...
+            var clientModel = new RepoAccessPlugin(@"C:\Users\t-akpoyibos\source\repos\azure-sdk-for-net\sdk\core\System.ClientModel\src");
```
m-redding: Could the agent prompt for these at startup rather than it needing to be hardcoded?

tools/perf-regression-finder/src/SemanticAzureAi.cs
sa7936: No, there should be a value - I edited this in new commit.

tools/perf-regression-finder/src/SemanticAzureAi.cs
sa7936: Yup, this can be done - I'll edit this

tools/perf-regression-finder/src/SemanticAzureAi.cs
```diff
@@ -0,0 +1,180 @@
...
+            KernelPlugin azureCorePlugin = KernelPluginFactory.CreateFromObject(azureCore, "azure_core");
+            KernelPlugin clientModelPlugin = KernelPluginFactory.CreateFromObject(clientModel, "system_clientmodel");
+            AzureAIAgent agent = new(definition, client, plugins: [azureCorePlugin, clientModelPlugin]);
+            return agent;
+        }
+
+        public async Task RunChatLoopAsync()
+        {
+            // Prompt for repo paths at startup
+            Console.Write("Enter the path to the Azure.Core src: ");
```
m-redding: Nit - it might be good to more explicit, I've seen confusion around stuff like this before where you need a really specifically formatted input, idk maybe something like "Enter the absolute path to your local copy of Azure.Core's src directory, for example, \'C:\\<path to repo root\>\azure-sdk-for-net\sdk\core\Azure.Core\src\'"

tools/perf-regression-finder/README.md
```diff
@@ -0,0 +1,170 @@
...
+## Features
+
+- Diagnoses speed and memory regressions in `Azure.Core` and `System.ClientModel`
+- Traces issues to root causes with annotated call stacks
+- Provides plain text output
+- Supports interactive, multi-line user input
+
+## Getting Started
+
+1. **Clone the repository:**
```
m-redding, line 17: Since cloning the azure-sdk-for-net repo is also required it might be helpful to note that here in the getting started section

tools/perf-regression-finder/src/SemanticAzureAi.cs
sa7936: Sounds good

maririos:
@sa7936 cool!!! could you expand more on how this will be used by the team? When/where/ etc

sa7936:
@maririos It is meant to help with scenarios like https://github.com/microsoft/semantic-kernel/issues/12550. Here, developers spent a lot of time trying to figure out the source of the regression in semantic kernel when the issue was really in System.ClientModel. The agent is meant to help cut down the time for exploration and give you areas you can start your search from if the regression is related to Azure.Core or System.ClientModel. The idea was if you have to call stack already, you would be able to explore all the paths that led to the call being reported for a regression.

m-redding (approved):
LGTM!!

PR #11441, opened 2025-07-31
[tsp-client] Add prettier check

catalinaperalta:


tools/tsp-client/package.json
```diff
@@ -21,7 +21,9 @@
     "test": "vitest run",
     "test:commands": "npx tsx ./test/run_commands.ts",
     "tsp-client": "node cmd/tsp-client.js",
-    "watch": "tsc -p . --watch"
+    "watch": "tsc -p . --watch",
+    "format": "prettier . --write",
+    "format:check": "prettier . --check"
```
timotheeguerin, line 26: Might also want to register the typespec petttier plugin

catalinaperalta: We dont really have tsp files related to core tsp-client functionality that we need to run formatting on 🤔 I think we could wait on this and add it later if it becomes relevant. I dont feel strongly if you think I should just add it now.

timotheeguerin: Oh if you don't have any then not needed, just assumed you did

scbedd (approved):

PR #11443, opened 2025-07-31
ignore Concret Type To Any

skywing918:
ignore all any cases
https://github.com/Azure/azure-sdk-tools/issues/11112


wanlwanl:

PR #11459, opened 2025-08-01
[azsdk-cli] Readme generation using OpenAI.

richardpark-msft:
This is a simple CLI tool to use OpenAI, and a markdown template, to generate a readme.

For now, it's just a CLI command, under the `generators` verb, to keep things neat, and not an MCP tool since we're still debating what lives there.

I also added in a 'service' so everyone can just declare IAzureOpenAiClient as a parameter in their command and have it available (it uses the defacto "AZURE_OPENAI_ENDPOINT" variable - if it's configured, we can connect). It also assumes DefaultAzureCredential (for now)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Azure.Sdk.Tools.Cli.csproj
```diff
@@ -14,10 +14,12 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.AI.Agents.Persistent" Version="1.0.0" />
+    <PackageReference Include="Azure.AI.OpenAI" Version="2.1.0" />
     <PackageReference Include="Microsoft.Graph" Version="5.79.0" />
+    <PackageReference Include="OpenAI" Version="2.2.0" />
```
jsquire: You get this dependency transitively via `Azure.AI.OpenAI`, adding a stand-alone dependency is going to hoist the version that AOAI uses.   There's a very strong dependency between them.   Normally, it shouldn't be a problem - but OpenAI evolves very, very differently than Azure services.

We sure that's what we want here?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedOptions.cs
```diff
@@ -19,14 +20,15 @@ public static class SharedOptions
...
+            typeof(ReadMeGeneratorTool),
             typeof(ReleasePlanTool),
             typeof(ReleaseReadinessTool),
             typeof(SdkReleaseTool),
             typeof(SpecCommonTools),
             typeof(SpecPullRequestTools),
             typeof(SpecWorkflowTool),
             typeof(SpecValidationTools),
-            typeof(TestAnalysisTool),
+            typeof(TestAnalysisTool),
```
jsquire:
```suggestion
            typeof(TestAnalysisTool),
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
```diff
@@ -75,6 +75,7 @@ public static WebApplicationBuilder CreateAppBuilder(string[] args)
             throw new ArgumentException($"Invalid output format '{outputFormat}'. Supported formats are: plain, json");
         }
 
+        builder.Services.AddSingleton<IAzureOpenAIClient>(new AzureOpenAIClientService());
```
jsquire: You'll want to use the `Microsoft.Extensions.Azure` package and the Azure Client Builder to manage Azure Clients.  This ensures the proper lifetime, cleanup, options configuration, and auth token.

see: [Dependency Injection with the Azure SDK for .NET](https://learn.microsoft.com/dotnet/azure/sdk/dependency-injection?tabs=host-builder).

Copying your patterns, it should look something like:
```csharp
services.AddAzureClients(clientBuilder =>
{
    // Do not use DefaultAzureCredential.  Use a custom chained token
    // for development scenarios.
    TokenCredential credential = useDevelopmentCredentials switch
    {
        true => new ChainedTokenCredential(
            new AzureCliCredential(),
            new AzurePowerShellCredential(),
            new VisualStudioCodeCredential(),
            new VisualStudioCredential()),

        _ => new AzurePipelinesCredential(
            Environment.GetEnvironmentVariable("AZURE_TENANT_ID"),
            Environment.GetEnvironmentVariable("AZURE_CLIENT_ID"),
            Environment.GetEnvironmentVariable("AZURE_SERVICE_CONNECTION_ID"),
            Environment.GetEnvironmentVariable("AZURE_WES_CREDIT_CARD_NUMBER"))
     };
    
    // Assign this token to all Azure clients created, unless
    // specifically overridden.
    clientBuilder.UseCredential(credential);
    
    // Azure OpenAI client does not, for some reason, have an
    // in-package facade for this, so register manually.
    clientBuilder.AddClient<AzureOpenAIClient, AzureOpenAIClientOptions>(
        (options, credential, _) =>
        {
            var ep = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT");

            if (string.IsNullOrEmpty(ep))
            {
                throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set, OpenAI related commands will not be available");
            }
            
            return new AzureOpenAIClient(new Uri(ep), credential, options);
        });
});
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/OpenAIService.cs
```diff
@@ -0,0 +1,61 @@
...
+using Azure.Identity;
+using Azure.Sdk.Tools.Cli.Models;
+using OpenAI.Chat;
+
+namespace Azure.Sdk.Tools.Cli.Services;
+
+/// <summary>
+/// Interface for Azure OpenAI client to enable dependency injection and testing
+/// </summary>
+public interface IAzureOpenAIClient
```
jsquire: This file should be deleted.   You shouldn't be adding an interface facade over Azure clients, nor building/configuring them separately. 

Please use the Azure client builder, as mentioned in `Program.cs`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/OpenAIService.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <param name="model">The model deployment name</param>
+    /// <returns>A chat client instance</returns>
+    ChatClient GetChatClient(string model);
+}
+
+
+/// <summary>
+/// Wrapper implementation of IAzureOpenAIClient that delegates to the actual AzureOpenAIClient
+/// </summary>
+public class AzureOpenAIClientService : IAzureOpenAIClient
```
jsquire: By convention, all C# types should have their own file.   You don't want to piggy-back this way.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,215 @@
...
+        private Option<string> templatePathOption;
+        private Option<string> serviceDocumentationOption;
+        private Option<string> modelOption;
+
+        public ReadMeGeneratorTool(ILogger<ReadMeGeneratorTool> logger, IOutputService output, IAzureOpenAIClient openAiClient)
+        {
+            ArgumentNullException.ThrowIfNull(logger);
+            ArgumentNullException.ThrowIfNull(output);
+
+            ArgumentNullException.ThrowIfNull(openAiClient);
```
jsquire:
```suggestion
            ArgumentNullException.ThrowIfNull(logger, nameof(logger));
            ArgumentNullException.ThrowIfNull(output, nameof(output));

            // This should be removed, as it won't be possible to have `null` here when the
            // patterns for DI are fixed.
            ArgumentNullException.ThrowIfNull(openAiClient, nameof(openAiClient));
```

Always include the parameter name, or you'll never be able to debug this.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,215 @@
...
+            ArgumentException.ThrowIfNullOrEmpty(templatePath);
+            ArgumentNullException.ThrowIfNull(serviceDocumentation);
+            ArgumentException.ThrowIfNullOrEmpty(outputPath);
+            ArgumentException.ThrowIfNullOrEmpty(engPath);
+
+            var readmeText = await File.ReadAllTextAsync(templatePath);
+
+            var messages = new ChatMessage[]{
+                new SystemChatMessage(
+                    string.Join("\n",
```
jsquire: This is not a pattern that you want.   You're allocating like crazy here.   You can avoid this by just using a literal string.

```csharp
new SystemChatMessage("""
  Hello, my name is Richard and I'm still sad that we 
  never built the receiver redirect feature because it 
  is my friend Chris' favorite thing in the world
  and we let him down.
""");
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,215 @@
...
+                        "The parameters are:",
+                        "* A URL that contains service documentation, to be used when creating key concepts, an introduction blurb and any other places where conceptual docs are needed",
+                        "* A package path that we can use to generate documentation links",
+                        "Here are some more rules to follow:",
+                        "- Do not touch the following sections, or its subsections: Contributing.",
+                        "- Do not generate sample code.",
+                        "- Rules for proper readmes can be found here: https://github.com/Azure/azure-sdk/blob/main/docs/policies/README-TEMPLATE.md"
+                    )
+                ),
+                new SystemChatMessage("The readme template is this: \"" + readmeText + "\" which we fill in with the user parameters, which follow:"),
```
jsquire:
```suggestion
                new SystemChatMessage($"The readme template is this: \"{readmeText}\" which we fill in with the user parameters, which follow:"),
```

No reason to use a different pattern here.  The bling string is just fine.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        private async Task Validate(string engPath, string readmePath)
+        {
+            output.Output($"Running {engPath}/common/scripts/Verify-Links.ps1 \"{readmePath}\"");
+
+            // TODO: I think chradek has a replacement for this.
+            var process = Process.Start(new ProcessStartInfo()
+            {
+                FileName = Environment.OSVersion.Platform == PlatformID.Win32NT ? "powershell.exe" : "pwsh",
```
jsquire: Change this over to [RuntimeInformation.IsOSPlatform](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.runtimeinformation.isosplatform?view=net-9.0).   It's a lot more reliable and consistent.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,215 @@
...
+                Arguments = $"\"{Path.Join(engPath, "common", "scripts", "Verify-Links.ps1")}\" \"{readmePath}\"",
+                RedirectStandardError = true,
+                RedirectStandardOutput = true
+            });
+
+            ArgumentNullException.ThrowIfNull(process);
```
jsquire:
```suggestion
```

This can't happen.  If the process cannot be created, it'll throw.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,215 @@
...
+
+            await process.WaitForExitAsync();
```
jsquire: You want to include a timeout threshold here via cancellation token.   You must protect against hangs.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,215 @@
...
+
+            if (process.ExitCode != 0)
+            {
+                //var stderr = await process.StandardError.ReadToEndAsync();
+                var stdout = await process.StandardOutput.ReadToEndAsync();
+                var tabified = string.Join("\n  ", stdout.Split("\n"));
```
jsquire:
```suggestion
                var tabified = stdout.Replace("\n", "\n  ");
```

Split/join is going to create a ton of unnecessary allocations.  Doing an inline replace will be much more efficient here.

jsquire (dismissed):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/OpenAIService.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <param name="endpoint">The Azure OpenAI endpoint URI</param>
+    /// <param name="credential">The token credential for authentication</param>
+    public AzureOpenAIClientService()
+    {
+        var ep = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT");
+
+        if (!string.IsNullOrEmpty(ep))
+        {
+            // TODO: there's a pipeline credential that we might also need to support?
+            _client = new AzureOpenAIClient(new Uri(ep), new DefaultAzureCredential());
```
jsquire: We should never be using `DefaultAzureCredential`, per current SFI guidance.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I feel ashamed on this one.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
richardpark-msft: I'll check this out. 

My bigger question now is - how the heck is anyone supposed to know that?!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I added that to the prompt.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Azure.Sdk.Tools.Cli.csproj
richardpark-msft: Let me see if I can remove it. We're using such basic functionality that any version of the package (provided it doesn't have bugs around chat completions) would work just fine.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/OpenAIService.cs
richardpark-msft: I'll change it, but you're making me the odd-one-out in this folder. :)

jsquire: Then we should change the folder, and people should feel bad about what they did.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: Process.Start is declared with `Process?` as it's return type. I'm okay with using process!.WaitForExitAsync, but I have to do something.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/OpenAIService.cs
richardpark-msft: Switched over to using the `AzureService` that was already in there, which understands how to pick up the right credential.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I did it! It was interesting, and I hope I did that right - more lines of code than I was expecting.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: !'d.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: Done!

I would be interested in what makes the other one unreliable. Maybe tricky because you could get down into comparing windows versions, which I never want to do again.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: Bling'd

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/OpenAIService.cs
richardpark-msft: I am not avoiding this responsibility, but I ended up not needing the file so I deleted it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/OpenAIService.cs
richardpark-msft: Deleted!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
richardpark-msft: I did this, with a slight variation because azsdk-cli already has a TokenCredential injector in place.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: It's beautiful now.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Azure.Sdk.Tools.Cli.csproj
richardpark-msft: I removed it, no issues.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
```diff
@@ -75,6 +77,23 @@ public static WebApplicationBuilder CreateAppBuilder(string[] args)
             throw new ArgumentException($"Invalid output format '{outputFormat}'. Supported formats are: plain, json");
         }
 
+        builder.Services.AddAzureClients(clientBuilder =>
+        {
+            // Azure OpenAI client does not, for some reason, have an
+            // in-package facade for this, so register manually.
+            clientBuilder.AddClient<AzureOpenAIClient, AzureOpenAIClientOptions>(
```
praveenkuttappan: This will force configuring OPENAI endpoint by everyone who wants to use our MCP server. right? Can we have a custom service class that abstracts this OpenAI client? Service class can look for the env variable only if they are invoking OpenAI functionality.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
```diff
@@ -75,6 +77,23 @@ public static WebApplicationBuilder CreateAppBuilder(string[] args)
...
+                (options, credential, _) =>
+                {
+                    var ep = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? "https://<need to know how to have a default>";
+
+                    if (string.IsNullOrEmpty(ep))
```
praveenkuttappan: nit: ep is not empty based on null handling in ln 87. And also, the Invalid Operation Exception will fail MCP server start.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
jsquire: We should absolutely NOT be wrapping the Azure SDK clients.    This only registers the factory, the actual resolving of the client does not occur until the client is injected.    If you don't ask for the `OpenAIClient`, the environment variable will never be read.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
jsquire: L87 only guarantees `not null`.    Personally, I think that's plenty for a sanity check, but it does not cover the empty case here, so it's not equivalent.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,219 @@
...
+            templatePathOption.SetDefaultValue("Templates/ReadMeGenerator/README-template.go.md");
+
+            serviceDocumentationOption = new Option<string>(
+                "--service-documentation")
+            {
+                Description = "URL to the service documentation",
+                IsRequired = true
+            };
+
+            serviceDocumentationOption.SetDefaultValue("https://learn.microsoft.com/en-us/azure/service-bus-messaging/");
```
praveenkuttappan: Option is marked as required. So why are setting a default value when field is required?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,219 @@
...
+
+            packagePathOption = new Option<string>(
+                "--package-path")
+            {
+                Description = "Path, underneath the 'sdk' folder (ex: devcenter/devbox)",
+                IsRequired = true,
+            };
+
+            engPathOption = new Option<string>(
+                "--eng-path")
```
praveenkuttappan: Can you please change this to repo root path and append "eng"? "eng" part is constant in all scenarios including pipeline runs.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
```diff
@@ -75,6 +77,23 @@ public static WebApplicationBuilder CreateAppBuilder(string[] args)
             throw new ArgumentException($"Invalid output format '{outputFormat}'. Supported formats are: plain, json");
         }
 
+        builder.Services.AddAzureClients(clientBuilder =>
```
jsquire: @praveenkuttappan:  Your credential registration for Azure clients should also go into the `AzureClientBuilder` and call `builder.UseCredential` so that it is the default for all Azure clients.

Any wrappers that you have, or other DI registration should be moved over to using the builder.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,219 @@
...
+            {
+                Description = "Path to an 'eng' folder within an SDK repo (ex: /home/user/src/azure-sdk-for-go/eng)",
+                IsRequired = true,
+            };
+
+            outputPathOption = new Option<string>(
+                "--output-path")
+            {
+                Description = "Path to write the generated README contents",
+                IsRequired = true,
```
praveenkuttappan: Is this required or optional? Can this path be defaulted to package root path?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,219 @@
...
+            });
+
+
+            using var verifyLinksCt = CancellationTokenSource.CreateLinkedTokenSource(ct);
+            verifyLinksCt.CancelAfter(TimeSpan.FromSeconds(30)); // Set your time limit here
+            await process!.WaitForExitAsync(verifyLinksCt.Token);
+
+            if (process.ExitCode != 0)
+            {
+                var stderr = await process.StandardError.ReadToEndAsync();
```
praveenkuttappan: I have seen this hanging and not returning from process when output size is large. Better approach to handle if you expect the output to be larger is to subscribe async events to output stream.
E.g. https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecValidationTool/SpecValidationTool.cs#L141

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,219 @@
...
+                    - Do not generate sample code.
+                    - Rules for proper readmes can be found here: https://github.com/Azure/azure-sdk/blob/main/docs/policies/README-TEMPLATE.md
+                    """
+                ),
+                new SystemChatMessage($"The readme template is this: {readmeText} which we fill in with the user parameters, which follow:"),
+                new UserChatMessage($"Service URL: {serviceDocumentation}"),
+                new UserChatMessage($"Package path: {packagePath}")
+            };
+
+            var response = await chatClient.CompleteChatAsync(messages);
```
praveenkuttappan, line 169: Exception handling for this OpenAI call?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I want to default this to something so badly, but there's no guarantee of what location they're in when they run this (or any) command.

I could try to default it to the current folder, and then fail if it turns out I can't derive a package path from it, however. What do you think of that?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I will check that out. I think the output is basically proportional to the # of links in the readme, but it's good to defend against this. 

If I can, I'll turn this into something we can both use.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: If we can't call into OpenAI the entire command is busted (this assumes we've also blown past the actual retry mechanism in the SDK). So I think I'm good with letting this fly out, unless you're saying we have a different pattern? (ie, are commands not allowed to throw exceptions?)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: Oh, yes, I can - I wasn't sure if there was already a convention here.

Do we have a convention on the parameter name? Should it be --repo?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
richardpark-msft: Agreed with @jsquire, and just want to mention that we do need to make sure that we don't ever end up doing what @praveenkuttappan is worried about. The user should be able to use any other command in here, provided they don't need a dependency.

(and in this case, we're safe since the client is created on demand)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
richardpark-msft: I updated it - it was kind of silly to have any cases where they wouldn't end up with some string value.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I fixed this up a bit - the command now infers the repo path from the package path you give it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I removed the repo part, it just figures it out from the package path.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: @benbp - offhand, are commands intended to handle their own exceptions when they're fatal, or is it okay to let them fly out, with reckless abandon?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: Ah, okay, it looks like @chrisradek already has a nicer process helper in his pending PR that uses this technique. He might want to merge with yours? (for now I'll just move on, and we'll merge later)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I removed it. I was thinking of it as an example but yeah, it shouldn't be the default as it doesn't even work for other packages at all :) I did move it as an example, into the description for the option.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedOptions.cs
```diff
@@ -6,6 +6,7 @@
 using Azure.Sdk.Tools.Cli.Tools.HostServer;
 using Azure.Sdk.Tools.Cli.Tools.SdkRelease;
 using Azure.Sdk.Tools.Cli.Tools.ReleaseReadiness;
+using Azure.Sdk.Tools.Cli.Tools.Generators;
```
benbp:
```suggestion
```

We're trying to just use the Tools namespace for everything but a few have snuck in, which I'm going to fix.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,285 @@
...
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Text.RegularExpressions;
+using Azure.AI.OpenAI;
+using Azure.Sdk.Tools.Cli.Commands;
+using Azure.Sdk.Tools.Cli.Contract;
+using Azure.Sdk.Tools.Cli.Services;
+using OpenAI.Chat;
+
+namespace Azure.Sdk.Tools.Cli.Tools.Generators
```
benbp:
```suggestion
namespace Azure.Sdk.Tools.Cli.Tools
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,285 @@
...
+        private Option<string> packagePathOption;
+        private Option<string> outputPathOption;
+        private Option<string> templatePathOption;
+        private Option<string> serviceDocumentationOption;
+        private Option<string> modelOption;
+
+        public ReadMeGeneratorTool(ILogger<ReadMeGeneratorTool> logger, IOutputService output, AzureOpenAIClient openAiClient)
+        {
+            ArgumentNullException.ThrowIfNull(logger, nameof(logger));
+            ArgumentNullException.ThrowIfNull(output, nameof(output));
```
benbp: I don't think these are necessary, you can trust the dependency injector, if they're null we'll have bigger problems. Or is this a check for test instantiation?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,285 @@
...
+                IsRequired = true,
+                Description = "Path to the README template file (ie: Templates/README-template.go.md)",
+            };
+
+            templatePathOption.SetDefaultValue("Templates/ReadMeGenerator/README-template.go.md");
+
+            serviceDocumentationOption = new Option<string>(
+                "--service-url")
+            {
+                Description = "URL to the service documentation (ex: https://learn.microsoft.com/en-us/azure/service-bus-messaging)",
```
benbp:
```suggestion
                Description = "URL to the service documentation (ex: https://learn.microsoft.com/azure/service-bus-messaging)",
```

Lest the link checker flag us.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,285 @@
...
+            this.logger = logger;
+            this.output = output;
+            this.openAiClient = openAiClient;
+
+            this.CommandHierarchy = [SharedCommandGroups.Generators];
+        }
+
+        public override Command GetCommand()
+        {
+            templatePathOption = new Option<string>("--template-path")
```
benbp: I have a preference for assigning these at declaration instead of within `GetCommand`, just to keep the method shorter, but up to you.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: That's a good point, I probably don't need those - I believe they got added "automatically".

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,285 @@
...
+
+            outputPathOption.SetDefaultValue("README.output.md");
+
+            modelOption = new Option<string>("--model")
+            {
+                Description = "The OpenAI model to use when generating the readme. Note, this will match the name of your Azure OpenAI model deployment.",
+                IsRequired = true,
+            };
+
+            modelOption.SetDefaultValue("gpt-4.1");
```
benbp:
```suggestion
            modelOption = new Option<string>("--model", () => "gpt-4.1")
            {
                Description = "The OpenAI model to use when generating the readme. Note, this will match the name of your Azure OpenAI model deployment.",
                IsRequired = true,
            };
```

Another option, though arguably less descriptive. But lets you define it as a literal if you move it up to where the class variables are defined.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
benbp: Commands should not throw exceptions. If this method was had an `McpTool` attribute we would throw a compile warning from our analyzer. Reason being we still want to format output nicely for the user in current and future cases.

In CLI it's not as big a deal, but with MCP we still need to be able to wrap the exception in the protocol response, otherwise it gets dumped to the mcp client logs in a way that the user cannot read it clearly. Due to the way the MCP SDK works with asp.net, we are not able to register our own exception handling middleware, so we have to handle it at the tool level, unfortunately.

For simple return types like this you can use `DefaultCommandResponse`, see an example for failure handling: https://github.com/Azure/azure-sdk-tools/blob/1f807aa40f4ed94fef202e29e53860f6da2e06fb/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/HelloWorldTool/HelloWorldTool.cs#L55-L58

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: Yeah, actually, I was just looking at that. The "newer" version of System.CommandLine forces a function to set default values, so while it looks a little weird, it's more proper for the future.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,285 @@
...
+        {
+            var valid = true;
+
+            try
+            {
+                var result = await VerifyLinks(repoPath, readmePath, ct);
+
+                if (result != null)
+                {
+                    output.OutputError($"[FAIL] Verify-Links.ps1 failed: {result}");
```
benbp: This is perhaps a poorly named class and some missing docs on my part. Output service is more for final response formatting than for intermediate logging. You should use `ILogger/this.logger` here, and only use output service in `HandleCommand()`.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// <returns>A string, if the verification did not pass, or null if it did.</returns>
+        /// <exception cref="Exception"></exception>
+        private async Task<string> VerifyLinks(string repoPath, string readmePath, CancellationToken ct)
+        {
+            var verifyLinksPs1 = Path.Join(repoPath, "eng", "common", "scripts", "Verify-Links.ps1");
+            var errors = new List<string>();
+
+            output.Output($"Running {verifyLinksPs1} {readmePath}");
+
+            var process = Process.Start(new ProcessStartInfo()
```
benbp, line 219: We can refactor this later into the process service that @chrisradek has in PR right now.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,285 @@
...
+            verifyLinksCt.CancelAfter(TimeSpan.FromSeconds(30)); // Set your time limit here
+
+            await process!.WaitForExitAsync(verifyLinksCt.Token);
+
+            if (process.ExitCode != 0)
+            {
+                var stderr = await process.StandardError.ReadToEndAsync(ct);
+                var stdout = await process.StandardOutput.ReadToEndAsync(ct);
+
+                return $"Verify-Links.ps1 check did not pass.\nStdout: {stdout.Replace("\n", "\n  ")}\nStderr: {stderr.Replace("\n", "\n  ")}";
```
benbp, line 237: Replace `\n` with `Environment.NewLine`, despite how annoyingly verbose it is.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,285 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Text.RegularExpressions;
+using Azure.AI.OpenAI;
```
benbp, line 8: The PR description mentions this being moved into a service? Would be nice to avoid dependencies on the openai packages in this class for testing/coupling purposes.

richardpark-msft: The .NET SDKs have a nice way of doing mocks, even when using the concrete types (ie, no wrappers are necessary): https://learn.microsoft.com/en-us/dotnet/azure/sdk/unit-testing-mocking?tabs=csharp

The client itself is GA and we can mock it, so I figured it was fine to depend on it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// <param name="packagePath">The package path, under 'sdk', that houses your package.</param>
+        /// <param name="outputPath">Path to write the generated README contents.</param>
+        /// <param name="repoPath">The root path of the repository containing the package.</param>
+        /// <param name="ct">Cancellation token for the operation.</param>
+        /// <returns>
+        /// True if a valid README was generated and passed validation checks; 
+        /// false if the README did not pass validation (e.g., due to dead or invalid links).
+        /// </returns>
+        /// <returns>true if a valid readme was generated, false is the readme did not pass validation checks. For instance, dead/invalid links.</returns>
+        async Task<bool> Generate(ChatClient chatClient, string templatePath, Uri serviceDocumentation, string packagePath, string outputPath, string repoPath, CancellationToken ct)
```
benbp, line 142: I know you mention keeping things simple/neat to start, but I don't think it would be so bad to enable this for MCP at the start. For azure agents I have the service default to our infra ([config](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Azure/Agent/AzureAgentService.cs#L17), [usage](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/AzurePipelines/PipelineAnalysisTool.cs#L113)) so that it's easy to use with MCP without requiring any config setup, assuming one has access. We could do the same for the IAzureOpenAiClient/Service.

Getting repo path and sub package parameters would probably be trivial for the mcp client to figure out based on the user's workspace.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/.gitignore
```diff
@@ -0,0 +1,3 @@
+# generated by the ReadMeGeneratorTool, so you'll see it a lot...
+README.output.md
```
benbp: I think this go in the repo root, the tool may be run from different directories in usage/test.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
benbp: Similar to what Praveen said in the below comment, we don't build any services here because we want them all to be lazy loaded to avoid both configuration overload and to reduce startup time per CLI invocation.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Templates/ReadMeGenerator/README-template.go.md
```diff
@@ -0,0 +1,106 @@
+# Azure Template Package client library for Go
+
+<Describe the service, and what kinds of features it provides - this should be a short paragraph>
+
+Use the <package> client module `github.com/Azure/azure-sdk-for-go/sdk/<package path>` in your application to:
+
+<give a bullet point list of activities you can do with this client>
```
benbp: I know it's just a template, but for anyone who might read it rendered anything in `<>` disappears in markdown since it becomes html.


```suggestion
[give a bullet point list of activities you can do with this client]
```

or I think this? (or %3C %3E)


```suggestion
\<give a bullet point list of activities you can do with this client\>
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Templates/ReadMeGenerator/README-template.go.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Prerequisites
+
+- [Supported](https://aka.ms/azsdk/go/supported-versions) version of Go - [Install Go](https://go.dev/doc/install)
+- Azure subscription - [Create a free account](https://azure.microsoft.com/free/)
+- <Include examples on how to deploy this resource using the Azure CLI, and via the Azure Portal>
+
+### Install the package
+
+Install the Azure <TODO: package name goes here> client module for Go with [go get][goget]:
```
benbp: I think the `goget` link is missing at the end of the file

benbp:
```suggestion
Install the Azure <TODO: package name goes here> client module for Go with [go get](goget):
```

benbp: Also same comment about the `<>`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Templates/ReadMeGenerator/README-template.go.md
```diff
@@ -0,0 +1,106 @@
...
+
+```
+go get github.com/Azure/azure-sdk-for-go/sdk/<package path>
+```
+
+# Key concepts
+
+<Bullet point list of your library's main concepts.>
```
benbp:
```suggestion
[Bullet point list of your library's main concepts.]
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
benbp: We did disable the nullable/non-nullable compile errors which is maybe why these were added.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
benbp: Since we have a default we should probably remove `IsRequired` as well

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
benbp: I'd love for us to be at a point where the only things necessary to mock for testing tools are the dependencies that get injected. And similarly for tool classes to not take any third party dependencies directly. I think we'll be better off in the long run as far as testing and maintenance if we gravitate towards that approach.

jsquire: @benbp:  I agree with the thought.   We should, however, be following best practices for mocking Azure clients - which is to mock them directly or via a mocking library (Moq, NSubstitute, etc).    We should not be writing wrappers for them.    Registration should be conditional based on the scenario (live vs testing) and the Azure client factory should be used as intended to manage client registrations.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Templates/ReadMeGenerator/README-template.go.md
richardpark-msft: It probably doesn't matter - I'm betting any convention that delineates some sort of replaceable text will work. 

I'll try a few variations.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: Done.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: They're still required - a user could use the option, and set a "", which will get flagged properly.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Templates/ReadMeGenerator/README-template.go.md
richardpark-msft: Done. Used ().

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Templates/ReadMeGenerator/README-template.go.md
richardpark-msft: I just removed the link. The command example is right below it, and there's literally nothing for the user to install.

richardpark-msft: Fixed. Used ()

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: We'll have further discussion about this. 

Our "default export as MCP tool" thing eats up some valuable real estate for people that might be using any other MCP in their workspace. Adding the attribute in a later PR is always an option - we don't have to resolve it all now.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/.gitignore
richardpark-msft: Done.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: Ah, I wondered what the intended difference was there.

We might be better off saying "this is what you return from your function" instead. Your other comment about the different path for MCP makes me think it'd be good to talk about how to unify these - it is a bit confusing.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: Let's have this discussion in future PRs. I'll rollback/modify whatever I need to in my code (and others) to get them to the standard we agree upon at a later date.

This is really just enabling people to start writing OpenAI based tools, which is happening now.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
richardpark-msft: We talked about this offline - this construct, as it is, is lazy, so we don't have to worry about non-OpenAI commands having to deal with failures because OpenAI is not configured.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I removed them, and nothing exploded so I think we're good.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: Fixed. :)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I swapped it so output was only called once (at the end), and logger was used for the other calls.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: We talked about this offline. 

We're going to go with "Azure clients can be used directly, but 3rd party clients need to be wrapped" but _only_ because Azure SDK clients make it simple to test them without having to introduce interfaces or wrappers.

Our original decision was to use PersistentAgents as the standard, but that involves more state tracking that we want to look into, so this PR will stick with OpenAIClient.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I promise, next PR.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs
richardpark-msft: I added in exception handling. :)

PR #11468, opened 2025-08-01
[tsp-client] Search for tsp-location.yaml in cwd

maorleger:
## What

When locating the tsp-location.yaml file, start by searching in the cwd. If not
found, fallback to the previous behavior of searching in the output directory.

## Why

Our documentation specifies that the various commands will look for a
tsp-location.yaml file in the current directory; however, when specifying an
output directory tsp-client will expect to find tsp-location.yaml in the output
directory. That may not always be the desired behavior and does not allow
scenarios such as:

```
- sdk/
  - keyvault/
    - keyvault-keys/
      - src/
        - tsp-location.yaml
        - generated/

keyvault-keys $ tsp-client update -o generated/
```

As a result, various workarounds have been spotted:

1. Placing tsp-location.yaml inside the generated folder: https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/keyvault/keyvault-keys/src/generated/tsp-location.yaml
2. Adding logic to move tsp-location to the right folder, then codegen, then
remove it: https://github.com/Azure/azure-sdk-for-js/blob/4a53654f0da085a1cf2c5f855a55fcea67497786/sdk/ai/ai-agents/package.json#L101

Etc.

Leading to confusion and inconsistency. Ideally we would like to see tsp-location placed in the same location
(sdk/<service>/<package-name>/tsp-location.yaml) regardless of the output type.

### DIFF
--- a/tools/tsp-client/src/fs.ts
+++ b/tools/tsp-client/src/fs.ts
@@ -19,12 +19,20 @@ export async function createTempDirectory(outputDir: string): Promise<string> {
   return tempRoot;
 }
 
-export async function readTspLocation(rootDir: string): Promise<TspLocation> {
+/**
+ * Reads the tsp-location.yaml file from the expected directory.
+ *
+ * By default, it searches in the current working directory. If not found,
+ * it searches in the provided fallback directory which is likely the output directory.
+ *
+ * @param fallbackDir Path to the fallback directory to search for tsp-location.yaml.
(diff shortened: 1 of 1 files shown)


maorleger:
Chatted for a bit offline and agree it's acceptable to have tsp-location.yaml live where the generated code is outputted to. Closing for now, but we can revisit if needed later

Issue #11282, opened 2025-07-21
[Devex Agent] Common Instructions


jeo02:
An easier approach to copilot instructions for the Devex agent would be to have a prompt.md file in eng sys common that can be referenced by #file:path/to/file. This way when the instructions change we don't have to go to every repo to change them.

jeo02:
Thanks to Richard for pointing this out :) https://github.com/Azure/azure-sdk-for-go/pull/24919

jeo02:
After discussion with Praveen we will be syncing all the prompts inside of the eng/common under a new folder. copilot/azsdk and put it there.

jeo02:
https://github.com/Azure/azure-rest-api-specs/pull/36744

PR #11477, opened 2025-08-01
Move prompts to common

jeo02:
## Summary
- Moving prompts from the `azure-rest-api-specs` to common so it can be used across repositories. Whether this is for just file referencing in the repository copilot instructions or creating a MCP cookbook. 
- Will also be removing the prompts from the specs repository. 
- Chose the file structure to be ...Common/instruction/azsdk-tools just so for clarity in case others outside the azsdk-tools wish to place there instructions/prompts in here. Welcome feedback on naming structure though.

### DIFF
--- a/eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
+++ b/eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
@@ -35,11 +35,11 @@ If any details are missing, prompt the user accordingly:
 - Provide this resource: [Release Plan Creation Guide](https://eng.ms/docs/products/azure-developer-experience/plan/release-plan-create)
 - Once all information is gathered, use `CreateReleasePlan` to create the release plan
 - Display the newly created release plan details to the user for confirmation
-- Refer [sdk details in release plan](sdk-details-in-release-plan.instructions.md) to identify languages configured in the TypeSpec project and add them to the release plan
+- Refer to #file:sdk-details-in-release-plan.instructions.md to identify languages configured in the TypeSpec project and add them to the release plan
 
 ## Step 5: Update SDK Details in Release Plan
-- Refer [sdk details in release plan](sdk-details-in-release-plan.instructions.md) to add languages and package names to the release plan
-- If the TypeSpec project is for a management plane, run [verify namespace approval](verify-namespace-approval.instructions.md) if this is first release of SDK.
+- Refer to #file:sdk-details-in-release-plan.instructions.md to add languages and package names to the release plan
+- If the TypeSpec project is for a management plane, refer to #file:verify-namespace-approval.instructions.md if this is first release of SDK.
 
 ## Step 6: Link SDK Pull Requests (if applicable)
 - Ask the user if they have already created SDK pull requests locally for any programming language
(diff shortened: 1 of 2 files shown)


eng/common/instructions/azsdk-tools/check-api-readiness.prompt.md
```diff
@@ -0,0 +1,7 @@
+---
```
weshaggard: Are these prompts or instructions? IIRC prompts cannot be outside of the well-defined prompts folder but we can import instructions.

jeo02: Really they are instructions since the manner they are being used is through file reference inside of the copilot instructions. I can change the name for consistency sake, to .instruction.md

eng/common/instructions/azsdk-tools/check-api-readiness.instructions.md
```diff
@@ -0,0 +1,7 @@
+---
+mode: 'agent'
+tools: ['CheckApiReadyForSDKGeneration', 'codebase', 'GetPullRequest', 'GetGitHubUserDetails', 'GetPullRequestForCurrentBranch']
```
praveenkuttappan: Do we still need to mention the tools list here since it's an instruction file?

jeo02: Yea I believe we don't have to

praveenkuttappan (approved):

jeo02:
Added a fix to prompts saying run x prompt. If prompts name is the same as a mcp tool it will actually understand run the tool instead.

eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
```diff
@@ -35,11 +35,11 @@ If any details are missing, prompt the user accordingly:
 - Provide this resource: [Release Plan Creation Guide](https://eng.ms/docs/products/azure-developer-experience/plan/release-plan-create)
 - Once all information is gathered, use `CreateReleasePlan` to create the release plan
 - Display the newly created release plan details to the user for confirmation
-- Refer [sdk details in release plan](sdk-details-in-release-plan.instructions.md) to identify languages configured in the TypeSpec project and add them to the release plan
+- Refer to #file:sdk-details-in-release-plan.instructions.md to identify languages configured in the TypeSpec project and add them to the release plan
```
weshaggard, line 38: Wouldn't this still need to be in a link format?

eng/common/instructions/azsdk-tools/check-package-readiness.instructions.md
```diff
@@ -0,0 +1,34 @@
+---
+description: 'This prompt is designed to check the release readiness of a SDK package.'
```
maririos, line 2:
```suggestion
description: 'This instruction is designed to check the release readiness of a SDK package.'
```

eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
maririos: for it to be human readable, yes. But for instructions looks like this should work: https://learn.microsoft.com/en-us/visualstudio/ide/copilot-chat-context-references?view=vs-2022#reference-a-file

(Juan told me about this support yesterday. I haven't tried it yet though #11485 )

jeo02: I believe since it's in the same folder it works takes it as ./sdk-details-in-release-plan.instructions.md
Worked when I tested it out in the rest-api-specs repo

eng/common/instructions/azsdk-tools/check-package-readiness.instructions.md
jeo02: I'll go around with copilot making sure everything in the instructions is correct. Make a new issue for that and follow up with this change

jeo02:
Resolves #11282

PR #11486, opened 2025-08-02
[tsp-client] support multi-emitter repos

catalinaperalta:


tools/tsp-client/src/commands.ts
```diff
@@ -259,7 +316,7 @@ export async function syncCommand(argv: any) {
   }
 
   try {
-    let emitterLockPath = getEmitterLockPath(getEmitterPackageJsonPath(repoRoot, tspLocation));
+    let emitterLockPath = getEmitterLockPath(emitterPackageJsonPath);
```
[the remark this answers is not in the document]
catalinaperalta: It's defined at the top of the command, this was an improvement to not process this twice

tools/tsp-client/src/commands.ts
```diff
@@ -31,6 +33,56 @@ import { sortOpenAPIDocument } from "@azure-tools/typespec-autorest";
 
 const defaultRelativeEmitterPackageJsonPath = joinPaths("eng", "emitter-package.json");
 
+// This function resolves for the correct emitter to use in a language repository based on whether
+// the user has provided an alternate dependency file path or if there are existing configuration
+// files in the repository. Supported configuration files are located at
+// <repo root>/eng/emitter-package.json or <repo root>/eng/tspclientconfig.yaml.
+// The function will return the emitter name and the path to the emitter package.json file relative
+// to repo root. If the default emitter-package.json is used, the path will be undefined.
```
JoshLove-msft: nit: should we use the standard multiline docs for function docs?

tools/tsp-client/src/commands.ts
```diff
@@ -31,6 +33,56 @@ import { sortOpenAPIDocument } from "@azure-tools/typespec-autorest";
...
+      path: relative(repoRoot, emitterPackageJsonOverride),
+    };
+  }
+
+  // If a global config file exists with supportedEmitters configured, use it to
+  // find the right emitter. The list of supported emitters will be checked in order, stopping
+  // at the first match in tspconfig.yaml.
+  if (globalConfigFile && globalConfigFile.supportedEmitters) {
+    for (const supportedEmitter of globalConfigFile.supportedEmitters) {
+      for (const configEmitter of Object.keys(tspConfigData.options) ?? []) {
```
JoshLove-msft: nit: is it worth parsing each config into dictionaries to avoid the O(N^2) loop?

tools/tsp-client/test/commands.spec.ts
```diff
@@ -372,6 +374,36 @@ describe.sequential("Verify commands", () => {
     }
   });
 
+  it("Init with global tspclientconfig.yaml", async () => {
```
JoshLove-msft, line 379: Can we add test cases for no matches between config and supportedEmitters and multiple matches?

tools/tsp-client/src/commands.ts
catalinaperalta: Done

tools/tsp-client/src/commands.ts
catalinaperalta: Done

tools/tsp-client/test/commands.spec.ts
catalinaperalta: Added a no matches test and added another entry into the global example file that would catch any issue with multiple matches, it would still be the same as the first test I wrote.

scbedd (approved):

PR #11493, opened 2025-08-04
Removed broken link from README

vcolin7:
The SSE protocol was removed from the MCP spec and it seems the link referring to it has also been removed from the spec docs.

### DIFF
--- a/eng/common/mcp/README.md
+++ b/eng/common/mcp/README.md
@@ -32,7 +32,5 @@ When running in copilot the default is stdio mode, but SSE is useful to support 
 
 See the [C# MCP SDK](https://github.com/modelcontextprotocol/csharp-sdk)
 
-Add an [SSE transport](https://github.com/modelcontextprotocol/csharp-sdk/tree/main/samples/AspNetCoreSseServer)
-
 TODO: Add the azsdk-cli project to pull in MCP server dependencies from the repo
 


eng/common/mcp/README.md
```diff
@@ -32,7 +32,5 @@ When running in copilot the default is stdio mode, but SSE is useful to support
 
 See the [C# MCP SDK](https://github.com/modelcontextprotocol/csharp-sdk)
 
-Add an [SSE transport](https://github.com/modelcontextprotocol/csharp-sdk/tree/main/samples/AspNetCoreSseServer)
```
benbp, line 35: Will you also update line 27?

```
Azure SDK MCP servers should support [stdio and streamable http transports](https://modelcontextprotocol.io/docs/concepts/transports).
```

benbp (approved):

eng/common/mcp/README.md
vcolin7: Good catch!

eng/common/mcp/README.md
```diff
@@ -24,15 +24,13 @@ The script will install the latest version of the azsdk cli executable from [too
 
 Azure SDK MCP server code is in [azure-sdk-tools/tools/azsdk-cli/Azure.Sdk.Tools.Cli](https://github.com/Azure/azure-sdk-tools/tree/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli).
 
-Azure SDK MCP servers should support [stdio and sse transports](https://modelcontextprotocol.io/docs/concepts/transports#server-sent-events-sse).
+Azure SDK MCP servers should support [stdio and streamable HTTP transports](https://modelcontextprotocol.io/docs/concepts/transports).
```
joshfree: MUST support stdio and SHOULD support StreamableHttp using latest OAuth 2.1 Best Current Practices (BCPs)

joshfree (changes requested):

eng/common/mcp/README.md
benbp: @joshfree do you mean to say if we are adding http it must require auth?

I originally had this sse/http line in here to ensure that servers would be debuggable via inspector, IDE breakpoints, etc.

PR #11494, opened 2025-08-04
[Python APIView] remove inherited Model base class methods

### LINKED ISSUE #11240: [Python APIView] don't show inherited methods from DPG Model base class
Per Anna:
   - Don't show inherited methods from DPG Model base class. Only show methods if they've been patched and keep the constructor. Keep ivars + properties.

swathipil:
Fixes #11240


packages/python-packages/apiview-stub-generator/apistub/nodes/_class_node.py
```diff
@@ -121,7 +121,7 @@ def _should_include_function(self, func_obj):
             return (
                 function_module
                 and function_module.startswith(self.pkg_root_namespace)
-                and not function_module.endswith("_model_base")
+                and not (function_module.endswith("_model_base") or function_module.endswith("model_base"))
```
tjprescott, line 124: Is there an example of what a "function_module" looks like that you're trying to omit?

swathipil: yeah it's the new file name for the current codegen. azure-healthinsights-radiologyinsights has recently regenerated and uses this module, so I added that into the tests.

tjprescott: No I mean like, what would a string that reaches this look like? Obviously it will end with "._model_base" or ".model_base" but do you have an example string that would be filtered out? How robust is this logic? Does model_base come from Azure core or is it a convention that service teams would need to purposefully follow?

swathipil: https://github.com/Azure/azure-sdk-for-python/blob/62ab7624e2331013f7a18fc4588cca9b90408e0c/sdk/ai/azure-ai-projects/azure/ai/projects/_utils/model_base.py#L537

tjprescott: Is `model_base.py` and auto-generated file?

packages/python-packages/apiview-stub-generator/tests/azure_tests/azure_test.py
```diff
@@ -75,8 +78,25 @@ def _get_src(src_dir, subdirectory, package_name, version):
     
     _copy_directory_from_github(src_dir, repo_url, directory, tag)
     print(f"Directory copied to: {src_dir}")
-    return src_dir
+    # Check if apiview-properties.json exists in the source directory
+    mapping_file_src = os.path.join(src_dir, "apiview-properties.json")
+    mapping_file = mapping_file_src if os.path.exists(mapping_file_src) else None
+    return src_dir, mapping_file
 
+def _get_mapping_file(dest_dir, subdirectory, package_name, version):
```
tjprescott, line 86: Is this relevant to the PR, or something extra?

swathipil: Something extra but will need to be added soon anyway b/c we're going to pass the mapping file in the CI for auto generated APIViews, to mirror that behavior.

PR #11531, opened 2025-08-05
Source user object id directly instead of with graph

benbp:
Source user object id directly instead of with graph

### DIFF
--- a/eng/common/TestResources/New-TestResources.ps1
+++ b/eng/common/TestResources/New-TestResources.ps1
@@ -353,15 +353,12 @@ try {
     # Make sure the provisioner OID is set so we can pass it through to the deployment.
     if (!$ProvisionerApplicationId -and !$ProvisionerApplicationOid) {
         if ($context.Account.Type -eq 'User') {
-            # Support corp tenant and TME tenant user id lookups
-            $user = Get-AzADUser -Mail $context.Account.Id
-            if ($null -eq $user -or !$user.Id) {
-                $user = Get-AzADUser -UserPrincipalName $context.Account.Id
-            }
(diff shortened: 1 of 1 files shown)


eng/common/TestResources/New-TestResources.ps1
```diff
@@ -353,15 +353,11 @@ try {
     # Make sure the provisioner OID is set so we can pass it through to the deployment.
     if (!$ProvisionerApplicationId -and !$ProvisionerApplicationOid) {
         if ($context.Account.Type -eq 'User') {
-            # Support corp tenant and TME tenant user id lookups
-            $user = Get-AzADUser -Mail $context.Account.Id
-            if ($null -eq $user -or !$user.Id) {
-                $user = Get-AzADUser -UserPrincipalName $context.Account.Id
-            }
-            if ($null -eq $user -or !$user.Id) {
+            $userAccountId = (Get-AzContext).Account.ExtendedProperties.HomeAccountId.Split('.')[0]
```
weshaggard, line 357: Is this documented to always be present? What does this id look like that you are splitting?

benbp: The format is `<object id>.<tenant id>`. We've been having new issues with calls to the graph API based on conditional access policy changes at the corp tenant level, so I'm trying to remove it. I'm not sure how reliable this property is, but I'm pretty sure it's more reliable than all the breaking changes in the graph module and graph api permissions we've had to deal with over the years.

It's certainly possible we will find some users that for some reason don't have this populated, but the graph piece seems to be broken for many right now.

weshaggard (approved):

PR #11547, opened 2025-08-05
Throw error if deprecated parameter -UserAuth is passed

benbp:
Throw error if deprecated parameter -UserAuth is passed


benbp:
@weshaggard the parameter doesn't exist it was removed last year, powershell just lets you pass undefined parameters silently.

PR #11557, opened 2025-08-06
Integrate ErrorAnalyzer library and added logic to upload files

radhgupta:
The PR focuses on:

1. Integrating Error Analyzer library with Generator Agent.
2. Added Logic to Generator Agent for uploading typespec files, generating vector and store.
3. Refactored code to remove duplicate error handling/logs, maintain consistency by using Result Pattern. 

There is no critical logic changes involved in Refactoring code. The core files added are:

- src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
- src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
- src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Result.cs
- src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/SdkBuildService.cs

### DIFF
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
@@ -98,12 +98,13 @@ namespace Azure.Tools.GeneratorAgent
                 return new List<RuleError>();
             }
 
-            List<RuleError> errors = new List<RuleError>();
-            
             MatchCollection matches = ErrorRegex.Matches(combinedOutput);
             Logger.LogDebug("Found {MatchCount} potential error matches using pattern", matches.Count);
             
             // TODO: Implement LLM-based error message parsing as a fallback
+
+            HashSet<(string type, string message)> seenErrors = new HashSet<(string type, string message)>();
+            List<RuleError> errors = new List<RuleError>();
 
             foreach (Match match in matches)
             {
@@ -112,11 +113,17 @@ namespace Azure.Tools.GeneratorAgent
                     string errorType = match.Groups[1].Value.Trim();
                     string errorMessage = match.Groups[2].Value.Trim();
 
-                    // Skip if we couldn't extract meaningful data
                     if (string.IsNullOrWhiteSpace(errorType) || string.IsNullOrWhiteSpace(errorMessage))
                     {
                         continue;
                     }
+
+                    (string type, string message) errorKey = (errorType, errorMessage);
+                    if (seenErrors.Contains(errorKey))
+                    {
+                        continue;
+                    }
+                    seenErrors.Add(errorKey);
 
                     try
                     {
@@ -131,23 +138,20 @@ namespace Azure.Tools.GeneratorAgent
                 }
             }
 
-            // Remove duplicates based on both type and message
-            return errors
-                .GroupBy(e => new { e.type, e.message })
-                .Select(g => g.First())
-                .ToList();
+            return errors;
         }
         
         /// <summary>
-        /// Takes List&lt;RuleError&gt; and calls ErrorAnalyzers.GetFixes method
+        /// Takes IEnumerable&lt;RuleError&gt; and calls ErrorAnalyzers.GetFixes method
         /// </summary>
-        public IEnumerable<Fix> GetFixes(List<RuleError> errors)
+        public IEnumerable<Fix> GetFixes(IEnumerable<RuleError> errors)
         {
             ArgumentNullException.ThrowIfNull(errors);
             try
             {
-                IEnumerable<Fix> fixes = ErrorAnalyzerService.GetFixes(errors);
-                Logger.LogInformation("Generated fixes for {ErrorCount} errors", errors.Count);
+                List<RuleError> errorList = errors as List<RuleError> ?? errors.ToList();
+                IEnumerable<Fix> fixes = ErrorAnalyzerService.GetFixes(errorList);
+                Logger.LogInformation("Generated fixes for {ErrorCount} errors", errorList.Count);
                 return fixes;
             }
             catch (Exception ex)
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
@@ -111,7 +111,7 @@ namespace Azure.Tools.GeneratorAgent
 
                 // Read the .tsp file content and upload as .txt since .tsp is not supported
                 string content = await File.ReadAllTextAsync(filePath, ct);
-                using var contentStream = new MemoryStream(Encoding.UTF8.GetBytes(content));
+                using MemoryStream contentStream = new MemoryStream(Encoding.UTF8.GetBytes(content));
                 
                 Response<PersistentAgentFileInfo>? uploaded = await Client.Files.UploadFileAsync(
                     contentStream,
@@ -151,19 +151,19 @@ namespace Azure.Tools.GeneratorAgent
             const int batchSize = 10;
             while (stopwatch.Elapsed < maxWaitTime)
             {
-                var results = new List<(string FileId, PersistentAgentFileInfo? File)>();
+                List<(string FileId, PersistentAgentFileInfo? File)> results = new List<(string FileId, PersistentAgentFileInfo? File)>();
                 
                 for (int i = 0; i < uploadedFilesIds.Count; i += batchSize)
                 {
-                    var batch = uploadedFilesIds.Skip(i).Take(batchSize);
-                    var checkTasks = batch.Select(fileId => CheckFileStatusAsync(fileId, ct));
-                    var batchResults = await Task.WhenAll(checkTasks);
+                    IEnumerable<string> batch = uploadedFilesIds.Skip(i).Take(batchSize);
+                    IEnumerable<Task<(string FileId, PersistentAgentFileInfo? File)>> checkTasks = batch.Select(fileId => CheckFileStatusAsync(fileId, ct));
+                    (string FileId, PersistentAgentFileInfo? File)[] batchResults = await Task.WhenAll(checkTasks);
                     results.AddRange(batchResults);
                 }
                 
                 bool allIndexed = true;
                 List<string>? pendingFiles = Logger.IsEnabled(LogLevel.Debug) ? new List<string>() : null;
-                var currentStatusCounts = Logger.IsEnabled(LogLevel.Information) ? new Dictionary<string, int>() : null;
+                Dictionary<string, int>? currentStatusCounts = Logger.IsEnabled(LogLevel.Information) ? new Dictionary<string, int>() : null;
                 
                 foreach ((string FileId, PersistentAgentFileInfo? File) result in results)
                 {


src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -0,0 +1,135 @@
...
+            ArgumentNullException.ThrowIfNull(logger);
+            Logger = logger;
+            
+            InitializeErrorAnalyzers();
+        }
+
+        /// <summary>
+        /// Initializes the ErrorAnalyzer providers. Thread-safe singleton initialization.
+        /// </summary>
+        private static void InitializeErrorAnalyzers()
```
m-redding: Could this be done in a static ctor instead?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,274 @@ internal virtual PersistentAgent CreateAgent()
...
-            Logger.LogInformation("✅ Agent created successfully: {Name} ({Id})", agent.Name, agent.Id);
+            Logger.LogInformation("Agent created successfully: {Name} ({Id})", agent.Name, agent.Id);
             return agent;
         }
 
+        public async Task FixCodeAsync(CancellationToken ct)
+        {
+            PersistentAgent agent = Agent.Value;
+
+            // TODO: Implement the code fixing logic here
```
m-redding, line 61: Is this an intentional TODO?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,274 @@ internal virtual PersistentAgent CreateAgent()
...
+            IEnumerable<Task<string?>> uploadTasks = tspFiles.Select(async file =>
+            {
+                await ConcurrencyLimiter.WaitAsync(ct);
+                try
+                {
+                    return await UploadSingleFileAsync(file, ct);
+                }
+                finally
+                {
+                    ConcurrencyLimiter.Release();
```
m-redding, line 94: curious is this to help avoid rate limits?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,274 @@ internal virtual PersistentAgent CreateAgent()
...
+            System.Diagnostics.Stopwatch stopwatch = System.Diagnostics.Stopwatch.StartNew();
+
+            Logger.LogInformation("Waiting for {Count} files to be indexed... (timeout: {Timeout}s)", 
+                uploadedFilesIds.Count, maxWaitTime.TotalSeconds);
+
+            Dictionary<string, int> statusCounts = new Dictionary<string, int>();
+
+            while (stopwatch.Elapsed < maxWaitTime)
+            {
+                List<Task<(string FileId, PersistentAgentFileInfo? File)>> checkTasks = new List<Task<(string FileId, PersistentAgentFileInfo? File)>>();
```
m-redding: This is just for tsp files, right? I can't imagine there ever being too many tsp files but I would just start to worry about resource exhaustion if there were like 100s of tasks getting spun up

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,274 @@ internal virtual PersistentAgent CreateAgent()
...
+                        if (!status.Equals("Processed", StringComparison.OrdinalIgnoreCase))
+                        {
+                            allIndexed = false;
+                            pendingFiles.Add($"{result.File?.Filename}({status})");
+                        }
+                    }
+                }
+
+                string statusSummary = string.Join(", ", statusCounts.Select(kvp => $"{kvp.Key}: {kvp.Value}"));
+                Logger.LogInformation("Indexing status summary: {StatusSummary} | Elapsed: {Elapsed:F1}s",
```
m-redding: nit but if the status counts and pending files list are only for informational logging I would probably try to guard the work to format all the strings and build the dictionary behind an IsEnabled check on ILogger to make sure that informational logging is enabled before doing it.

Unless informational logs are always enabled?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,274 @@ internal virtual PersistentAgent CreateAgent()
...
+
+                if (pendingFiles.Count <= 3)
+                {
+                    Logger.LogDebug("Still waiting for: {PendingFiles}", string.Join(", ", pendingFiles));
+                }
+
+                await Task.Delay(pollingInterval, ct);
+            }
+
+            (string FileId, PersistentAgentFileInfo? File)[] finalResults = await Task.WhenAll(uploadedFilesIds.Select(id => CheckFileStatusAsync(id, ct)));
```
m-redding, line 224: It seems like if you get to this point you've waited another polling interval and checked the status again, could it be possible that all the tasks are now done? Should that be double checked before throwing in 235?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
```diff
@@ -41,65 +41,65 @@ public GitHubTypeSpecSdkGenerationService(
...
+            if (powerShellResult.IsFailure)
+            {
+                throw new InvalidOperationException($"PowerShell generation script failed: {powerShellResult.Error}");
+            }
+
+            // Step 2: Run dotnet build generateCode
+            Result buildResult = await RunDotNetBuildGenerateCodeAsync(cancellationToken);
+            if (buildResult.IsFailure)
+            {
+                // TODO: Send buildResult.Error to AI for analysis
```
m-redding: Assuming this one is also an intentional TODO?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
radhgupta: Yes, I intent to implement it in the next PR to avoid creating too big PRs

m-redding: Makes sense! Just wanted to double check

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            if (IsInitialized) return;
+
+            lock (InitializationLock)
+            {
+                if (IsInitialized) return;
+
+                try
+                {
+                    ErrorAnalyzerService.RegisterProvider(new ClientAnalyzerProvider());
```
jsquire: If you made the provider registry backed by a `ConcurrentDictionary`, you could just do a `TryAdd`  there instead and remove this.

Since it's singleton that you're going for, you could move this to a static constructor and you'd have a guarantee of running only once.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -0,0 +1,135 @@
...
+                    ErrorAnalyzerService.RegisterProvider(new GeneralAnalyzerProvider());
+                    ErrorAnalyzerService.RegisterProvider(new ManagementAnalyzerProvider());
+                    IsInitialized = true;
+                }
+                catch (Exception ex)
+                {
+                    // Log the error but don't throw - we can still continue without error analysis
+                    System.Diagnostics.Debug.WriteLine($"Failed to initialize ErrorAnalyzers: {ex.Message}");
```
jsquire: Shouldn't we be logging here?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
radhgupta: yes it is used as a throttling mechanism to handle scenarios where a directory contains a large number of TypeSpec files. Without this limiter, uploading hundreds or thousands of files simultaneously could overwhelm the Azure service endpoints. By limiting concurrent uploads to match the number of processor cores, we ensure efficient throughput while maintaining system stability.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -0,0 +1,135 @@
...
+        public List<RuleError> ParseBuildOutput(string buildOutput)
+        {
+            ArgumentException.ThrowIfNullOrWhiteSpace(buildOutput);
+
+            List<RuleError> errors = new List<RuleError>();
+            
+            // Regular expression to match error patterns like:
+            // error AZC0012: Single word class names are too generic...
+            string errorPattern = @"error\s+([A-Z]+\d+):\s*(.+?)(?=\s*\[|$)";
+            Regex regex = new Regex(errorPattern, RegexOptions.IgnoreCase | RegexOptions.Multiline);
```
jsquire: Hoist this to the static class level and use the `Compiled` option.   RegEx is expensive and you want to pay down what you can earlier.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -0,0 +1,135 @@
...
+        /// Takes raw build output string and generates List&lt;RuleError&gt;
+        /// </summary>
+        public List<RuleError> ParseBuildOutput(string buildOutput)
+        {
+            ArgumentException.ThrowIfNullOrWhiteSpace(buildOutput);
+
+            List<RuleError> errors = new List<RuleError>();
+            
+            // Regular expression to match error patterns like:
+            // error AZC0012: Single word class names are too generic...
```
jsquire: This makes an assumption that all analyzers follow our patterns for messaging.  I'm not sure that all the system analyzers do.    I wonder if we should use an LLM to decompose the message.   if the RegEx fails to get a match.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubTypeSpecSdkGenerationService.cs
radhgupta: Yup, this one is intentional too!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,274 @@ internal virtual PersistentAgent CreateAgent()
...
+            return thread.Id;
+        }
+
+        private async Task<List<string>> UploadTspAsync(string typeSpecDir, CancellationToken ct)
+        {
+            string[] tspFiles = Directory.GetFiles(typeSpecDir, "*.tsp", SearchOption.AllDirectories);
+
+            IEnumerable<Task<string?>> uploadTasks = tspFiles.Select(async file =>
+            {
+                await ConcurrencyLimiter.WaitAsync(ct);
```
jsquire, line 87: Why do we need to serialize the loads?  It kind of makes the approach meaningless, as you would have less context switching to just to them in a local `foreach` loop.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,274 @@ internal virtual PersistentAgent CreateAgent()
...
+                    return await UploadSingleFileAsync(file, ct);
+                }
+                finally
+                {
+                    ConcurrencyLimiter.Release();
+                }
+            });
+
+            string?[] results = await Task.WhenAll(uploadTasks);
+            List<string> uploadedFilesIds = results.Where(id => !string.IsNullOrEmpty(id)).Cast<string>().ToList();
```
jsquire: `Results` is already a string.   I don't think you need the `Cast` here.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,274 @@ internal virtual PersistentAgent CreateAgent()
...
+        }
+
+        private async Task<string?> UploadSingleFileAsync(string filePath, CancellationToken ct)
+        {
+            try
+            {
+                string fileName = Path.GetFileNameWithoutExtension(filePath);
+
+                // Read the .tsp file content and upload as .txt since .tsp is not supported
+                string content = await File.ReadAllTextAsync(filePath, ct);
```
jsquire, line 113: We shouldn't need to write a temp file just to give it a filename, since you're passing that in L125.   We should be able to just open the original file stream and upload.    Am I missing something?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,274 @@ internal virtual PersistentAgent CreateAgent()
...
+            catch (Exception ex)
+            {
+                Logger.LogError(ex, "Error uploading file: {FilePath}", filePath);
+                return null;
+            }
+        }
+
+        private async Task WaitForIndexingAsync(List<string> uploadedFilesIds, CancellationToken ct)
+        {
+            TimeSpan maxWaitTime = TimeSpan.FromSeconds(180);
```
jsquire: These seem like something we should be driving from conifg.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,274 @@ internal virtual PersistentAgent CreateAgent()
...
+            
+            string finalStatusSummary = string.Join(", ", finalSummary.Select(kvp => $"{kvp.Key}: {kvp.Value}"));
+            
+            throw new TimeoutException(
+                $"Timeout after {maxWaitTime.TotalSeconds}s while waiting for file indexing to complete. " +
+                $"Final status: {finalStatusSummary}. " +
+                $"This may indicate the Azure AI service is experiencing delays or the files require more processing time.");
+        }
+
+        private async Task<(string FileId, PersistentAgentFileInfo? File)> CheckFileStatusAsync(string fileId, CancellationToken ct)
```
jsquire, line 247: The file isn't available after upload, so we have to poll to see when it is available.   Seems like there should be an LRO or other way to do this where we don't need to poll ourselves.

@ShivangiReja | @m-redding :  Anything that you've come across for this?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
```diff
@@ -38,122 +34,77 @@ public LocalTypeSpecSdkGenerationService(
...
-
-                Logger.LogInformation("TypeSpec dependencies installed globally successfully: {Output}", output);
-                return true;
-            }
-            catch (OperationCanceledException)
-            {
-                // Let cancellation exceptions bubble up for this protected method too
-                throw;
+                // TODO: Send compileResult.Error to AI for analysis
+                throw new InvalidOperationException($"Failed to compile TypeSpec: {compileResult.Error}");
```
jsquire: nit:  Is `InvalidOperationException` what we want here?   That indicates "you can't do this operation" usually because doing so is invalid for the current state.   This is more "OperationFailed" which doesn't quite fit.

Not a huge deal because it's all internal, but we may want to look over the list and see if there's anything that's a better match.  

Maybe:  `System.ClientModel.ClientResultException`?    If we're going to catch this, it may be worth creating our own type to make sure that we only catch this scenario.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
radhgupta: Thanks for suggesting, I agree, its a better way of doing it! Implementing it!!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Result.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        // Implicit conversions for convenience
+        public static implicit operator Result<T>(T value) => Success(value);
+        public static implicit operator Result<T>(string error) => Failure(error);
+    }
+
+    /// <summary>
+    /// Non-generic result for operations that don't return a value.
+    /// </summary>
+    public class Result
```
jsquire: One class per file, please.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Result.cs
```diff
@@ -0,0 +1,78 @@
+namespace Azure.Tools.GeneratorAgent
+{
+    /// <summary>
+    /// Represents the result of an operation that can either succeed or fail.
+    /// </summary>
+    /// <typeparam name="T">The type of the success value.</typeparam>
+    public class Result<T>
```
jsquire: Consider making this a `record` type.   You get structural equality and such for free.

jsquire:
```suggestion
    internal class Result<T>
```

I don't think we want this exposed/used outside of our scope.   If so, we need to make the name more explicit.   "Result" is very generic.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
radhgupta: I'll modify the code to process files in batches of 10 to prevent resource exhaustion while still maintaining parallel processing benefits.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Result.cs
```diff
@@ -0,0 +1,78 @@
...
+                if (Exception != null)
+                {
+                    throw Exception;
+                }
+                throw new InvalidOperationException(Error);
+            }
+        }
+
+        // Implicit conversion for convenience
+        public static implicit operator Result(string error) => Failure(error);
```
jsquire: This seems dangerous.  What's the use case?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/SdkBuildService.cs
```diff
@@ -0,0 +1,182 @@
...
+using System.Threading.Tasks;
+using Azure.Tools.GeneratorAgent.Security;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Tools.GeneratorAgent
+{
+    /// <summary>
+    /// Service responsible for building the generated SDK and capturing build logs.
+    /// </summary>
+    internal class SdkBuildService
```
jsquire, line 14: At some point, we should move over from `Sdk` to `Library` in all the types.   "Azure SDK` is a name for the collective set of Azure libraries.   Individually, they're "libraries".

Very low on the priority list, but worth mentioning and tracking.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
radhgupta: Good catch! You're right, the files could complete indexing during that final check. I'll add a check to verify if all files are processed before throwing the timeout exception.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
radhgupta: Oh yeah, I was planning on passing all the error which doesn't match the regex directly to the agent, I forgot to add a TODO there. I will add that.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
radhgupta: This makes me wonder that given that we need our Agent to handle two types of error:
1. dotnet build on generated code
2. typespec compilation error
Would it be a good idea to have custom exception?? this way throughout the code, if we see any of these custom exception, we can trigger the agent?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
radhgupta: After the polling loop times out, it performs one final check of all files. If this final check shows that all files are now processed, it returns successfully with a log message indicating completion 'during final check'. The timeout exception is only thrown if files are still not processed after this final verification. This  way we make sure, we don't throw a timeout unnecessarily if the files complete processing right at the end of the waiting period

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+                    }
+                    catch (Exception ex)
+                    {
+                        Logger.LogWarning(ex, "Failed to create RuleError for type='{ErrorType}', message='{ErrorMessage}'", errorType, errorMessage);
+                    }
+                }
+            }
+
+            // Remove duplicates based on both type and message
+            List<RuleError> uniqueErrors = errors
```
jsquire: Does it need to be a list here?   Can we pass back `IEnumerable<RuleError>` and defer the materialization of the set until we actually need it?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <summary>
+        /// Takes List&lt;RuleError&gt; and calls ErrorAnalyzers.GetFixes method
+        /// </summary>
+        public IEnumerable<Fix> GetFixes(List<RuleError> errors)
+        {
+            ArgumentNullException.ThrowIfNull(errors);
+            try
+            {
+                IEnumerable<Fix> fixes = ErrorAnalyzerService.GetFixes(errors);
+                List<Fix> fixList = fixes.ToList();
```
jsquire: You definitely don't need to `ToList` here.   You're returning `IEnumerable` - save the allocation.   😄

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/LocalTypeSpecSdkGenerationService.cs
jsquire: That's an interesting thought.   `RecoverableError` or something like that?   That would let us clearly mark "we think this can be fixed because it's a known thing" versus "OMG, what just happened?"

radhgupta: Yup! I will implement it in this PR.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            HashSet<(string type, string message)> seenErrors = new HashSet<(string type, string message)>();
+            List<RuleError> errors = new List<RuleError>();
+
+            foreach (Match match in matches)
+            {
+                if (match.Groups.Count >= 3)
+                {
+                    string errorType = match.Groups[1].Value.Trim();
+                    string errorMessage = match.Groups[2].Value.Trim();
```
m-redding, line 114: Just curious but what is this leaving out? From the regex it looks like probably just the filename/line number at the end?

radhgupta: Yes, you are right, I used 2 strings here because the error analyzer library we have takes 2 sperate input i.e. error message and error type

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
m-redding: I agree that there should be an LRO if that's the case. I'm not familiar with everything available in the package but from the samples they only say to "wait it for it to finish indexing" but there isn't any info about what that means...

m-redding: I wonder if this is a bit of a bug in the API?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -0,0 +1,164 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Tools.GeneratorAgent
+{
+    /// <summary>
+    /// Service for analyzing build errors and generating AI prompts for fixes.
+    /// </summary>
+    internal class BuildErrorAnalyzer
+    {
+        // Regular expression to match error patterns.
```
jsquire, line 19: nit:  It may be helpful to just mention the pattern that you're looking to match in the comment.   That would help identify potential "it's not matching" bugs without digging deeper into the murky world of RegEx.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -0,0 +1,164 @@
...
+        
+        /// <summary>
+        /// Takes IEnumerable&lt;RuleError&gt; and calls ErrorAnalyzers.GetFixes method
+        /// </summary>
+        public IEnumerable<Fix> GetFixes(IEnumerable<RuleError> errors)
+        {
+            ArgumentNullException.ThrowIfNull(errors);
+            try
+            {
+                List<RuleError> errorList = errors as List<RuleError> ?? errors.ToList();
```
jsquire, line 152: Why do we need to force a list here?   Can't `GetFixes` just take an enumerable and walk the set?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -1,45 +1,44 @@
 using Azure.AI.Agents.Persistent;
 using Azure.Tools.GeneratorAgent.Configuration;
 using Microsoft.Extensions.Logging;
+using System.Text.Json;
```
jsquire, line 4: nit:  Please use the "Remove and Sort" refactoring to order system usings first.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,18 +50,280 @@ internal virtual PersistentAgent CreateAgent()
...
+                    return await UploadSingleFileAsync(file, ct);
+                }
+                finally
+                {
+                    ConcurrencyLimiter.Release();
+                }
+            });
+
+            string?[] results = await Task.WhenAll(uploadTasks);
+            List<string> uploadedFilesIds = results.Where(id => !string.IsNullOrEmpty(id)).Select(id => id!).ToList();
```
jsquire, line 99: Do we need to materialize the list here?  Can we just pass back IEnumerable instead?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Exceptions/ProcessExecutionException.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+
+namespace Azure.Tools.GeneratorAgent.Exceptions
+{
+    /// <summary>
+    /// Base exception class for process execution related errors.
+    /// Contains common properties for all process execution failures.
+    /// </summary>
+    public abstract class ProcessExecutionException : Exception
```
jsquire, line 9:
```suggestion
    internal abstract class ProcessExecutionException : Exception
```
Does this need to be `public`?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Exceptions/GeneralProcessExecutionException.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+
+namespace Azure.Tools.GeneratorAgent.Exceptions
+{
+    /// <summary>
+    /// General exception thrown when any process execution fails.
+    /// This exception type is suitable for AI analysis and automated fixing.
+    /// </summary>
+    internal class GeneralProcessExecutionException : ProcessExecutionException
```
jsquire, line 9: The question that we generally ask for a custom execption type is "what steps would you take to handle these?"   If you have a need to handle the specific scenario, it's a good path.   If not, then you really don't need a custom exception.

m-redding (approved):
All my feedback is addressed, this looks good to me!

jsquire (approved):

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
radhgupta: SOunds good, fixing it in next PR.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
radhgupta: fixing it in next PR.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
radhgupta: fixing it in next PR.

PR #11559, opened 2025-08-06
Add a new script to mark package as released from release pipeline

praveenkuttappan:
This script will be plugged in to the release stage of the pipeline post publishing steps to identify any release plans created for releasing this SDK and mark package release status as released in the release plan work item.

### DIFF
--- a/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
+++ b/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
@@ -1015,3 +1015,66 @@ function UpdateValidationStatus($pkgvalidationDetails, $BuildDefinition, $Pipeli
     Write-Host "[$($workItem.id)]$LanguageDisplayName - $pkgName($versionMajorMinor) - Updated"
     return $true
 }
+
+
+function Get-LanguageDevOpsName
+{
+    param([string]$LanguageShort)
+
+    switch ($LanguageShort.ToLower()) 
+    {
+        "net" { return "Dotnet" }
+        "js" { return "JavaScript" }
+        "java" { return "Java" }
+        "go" { return "Go" }
+        "python" { return "Python" }
+        "rust" { return "Rust" }
+        "cpp" { return "Cpp" }
+        "c" { return "C" }
+        default { return $null }
+    }
+}
+
+function Get-ReleasePlanForPullRequest($prLink)
+{
+  $devopsFieldLanguage = Get-LanguageDevOpsName -LanguageShort $LanguageShort
+  if (!$devopsFieldLanguage)
+  {
+    Write-Host "Unsupported language to check release plans, language [$LanguageShort]"
+    return $null
+  }
+
+  $prFieldName = "SDKPullRequestFor$($devopsFieldLanguage)"
+  $releaseStatusField = "ReleaseStatusFor$($devopsFieldLanguage)"
+  $fields = @()
+  $fields += "System.ID"
+  $fields += "System.State"
+  $fields += "System.AssignedTo"
+  $fields += "System.Parent"
+  $fields += "System.Tags"
+
+  $fieldList = ($fields | ForEach-Object { "[$_]"}) -join ", "
+  $query = "SELECT ${fieldList} FROM WorkItems WHERE [Work Item Type] = 'Release Plan' AND [$prFieldName] = '${prLink}'"
+  $query += " AND [$releaseStatusField] <> 'Released'"
+  $query += " AND [System.State] NOT IN ('Finished', 'Abandoned')"
+  $workItems = Invoke-Query $fields $query
+  return $workItems
+}
+
+function Update-ReleaseStatusInReleasePlan($releasePlanWorkItemId, $status, $version)
+{  
+  $devopsFieldLanguage = Get-LanguageDevOpsName -LanguageShort $LanguageShort
+  if (!$devopsFieldLanguage)
+  {
+    Write-Host "Unsupported language to check release plans, language [$LanguageShort]"
+    return $null
+  }
+
+  $fields = @()
+  $fields += "`"ReleaseStatusFor$($devopsFieldLanguage)=$status`""
+  $fields += "`"ReleasedVersionFor$($devopsFieldLanguage)=$version`""
+
+  Write-Host "Updating Release Plan [$releasePlanWorkItemId] with status [$status] for language [$LanguageShort]."
+  $workItem = UpdateWorkItem -id $releasePlanWorkItemId -fields $fields
+  Write-Host "Updated release status for [$LanguageShort] in Release Plan [$releasePlanWorkItemId]"
+}
--- a/eng/common/scripts/Mark-ReleasePlanCompletion.ps1
+++ b/eng/common/scripts/Mark-ReleasePlanCompletion.ps1
@@ -1,0 +1,139 @@
+param(
+    [Parameter(Mandatory = $true)]
+    [string]$Path,
+    
+    [Parameter(Mandatory = $true)]
+    [string]$PackageName,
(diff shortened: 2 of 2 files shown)


eng/common/scripts/Mark-ReleasePlanCompletion.ps1
```diff
@@ -0,0 +1,139 @@
...
+Write-Host "Finding most recent merged pull requests for path: $Path"
+
+$pullRequests = Get-MergedPullRequests
+if (!$pullRequests)
+{
+    Write-Host "No valid pull request commits found."
+    exit 0
+}
+# Check Azure DevOps Release Plan work items if LanguageShort is available
+Write-Host "Checking Azure DevOps Release Plan work items for language: $LanguageShort"
```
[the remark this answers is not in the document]
praveenkuttappan: It's defined in Common.ps1.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -1015,3 +1015,66 @@ function UpdateValidationStatus($pkgvalidationDetails, $BuildDefinition, $Pipeli
...
+        "python" { return "Python" }
+        "rust" { return "Rust" }
+        "cpp" { return "Cpp" }
+        "c" { return "C" }
+        default { return $null }
+    }
+}
+
+function Get-ReleasePlanForPullRequest($prLink)
+{
```
benbp, line 1034: One function is 2 spaced indentation the other is 4 spaced.

praveenkuttappan: Resolved

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -1015,3 +1015,66 @@ function UpdateValidationStatus($pkgvalidationDetails, $BuildDefinition, $Pipeli
...
+        "go" { return "Go" }
+        "python" { return "Python" }
+        "rust" { return "Rust" }
+        "cpp" { return "Cpp" }
+        "c" { return "C" }
+        default { return $null }
+    }
+}
+
+function Get-ReleasePlanForPullRequest($prLink)
```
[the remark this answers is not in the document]
benbp: This seems valid where is LanguageShort defined?

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -1015,3 +1015,64 @@ function UpdateValidationStatus($pkgvalidationDetails, $BuildDefinition, $Pipeli
...
+    switch ($LanguageShort.ToLower()) 
+    {
+        "net" { return "Dotnet" }
+        "js" { return "JavaScript" }
+        "java" { return "Java" }
+        "go" { return "Go" }
+        "python" { return "Python" }
+        "rust" { return "Rust" }
+        "cpp" { return "Cpp" }
+        "c" { return "C" }
```
maririos: we don't create release plans for these languages. Would it be better to avoid running `Get-ReleasePlanForPullRequest` for them?

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -1015,3 +1015,64 @@ function UpdateValidationStatus($pkgvalidationDetails, $BuildDefinition, $Pipeli
...
+    $fields = @()
+    $fields += "System.ID"
+    $fields += "System.State"
+    $fields += "System.AssignedTo"
+    $fields += "System.Parent"
+    $fields += "System.Tags"
+
+    $fieldList = ($fields | ForEach-Object { "[$_]"}) -join ", "
+    $query = "SELECT ${fieldList} FROM WorkItems WHERE [Work Item Type] = 'Release Plan' AND [$prFieldName] = '${prLink}'"
+    $query += " AND [$releaseStatusField] <> 'Released'"
```
maririos: I am a little confused about this condition. Why is this important at this point when you are only getting the release plans for a specific PR?
I understand we want only the active ones, but idk why not the ones that are not in released state.
Also, what does `release` state means?

weshaggard (changes requested):
As we chatted lets not use the PRs for this lets use the package name and PR status from the release plant

maririos (commented):
as we talked offline, leaving the feedback here. but I will check once there is an update

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
praveenkuttappan: Script doesn't run for them. But I didn't want to remove them from helper mapping method class if we ever decides to add them.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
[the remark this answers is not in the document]
praveenkuttappan: Common.ps1

eng/common/scripts/Mark-ReleasePlanCompletion.ps1
```diff
@@ -0,0 +1,55 @@
...
+{
+    Write-Host "No active release plans found for package name: $PackageName."
+    exit 0
+}
+
+$activeReleasePlan = $workItems
+if($workItems.Count -gt 1 -and ($workItems -is [System.Array]))
+{
+    $activeReleasePlan = $workItems[0]
+    Write-Warning "Multiple active release plans found for package name: $PackageName. Using the first release plan to update release status."
```
weshaggard: Just to help with debugging I would list the ids for the workitems found and the one you are choosing.

eng/common/scripts/Mark-ReleasePlanCompletion.ps1
```diff
@@ -0,0 +1,55 @@
+param(    
+    [Parameter(Mandatory = $true)]
+    [string]$PackageName,
+
+    [Parameter(Mandatory = $true)]
+    [string]$PackageVersion,
```
weshaggard, line 6: I assume you will get this version from the PackageInfo when you make the call to it in the pipelines.

weshaggard (approved):
One minor suggestion but otherwise looks good.

eng/common/scripts/Mark-ReleasePlanCompletion.ps1
maririos: I am not sure if we should go ahead with marking the package as released without checking for more information to make sure the release plan is the correct one.
If we have the package version from the pipeline, can we validate it with the version on the release plan? (is that information available?)
Or would there be another way to make a more informed decision?
I think worst case scenario I would prefer to not mark anything as released and add the warning, than marking incorrectly a package as released and marking a release plan as completed when that is not the case

weshaggard: We don't have a great way to match up the version. We are assuming that if you have a release plan and you have merged your PR that you are ready to release and so the next package released would be the one associated with that release plan. It isn't 100% but it is about as close as we can get with the data in hand.

praveenkuttappan: Yes, We don't have the version info that we can fully trust because package work item retains only latest version from CI pipeline and this will get mixed up if a team runs CI pipeline from different branches with different versions. IIRC, Storage and ML team does this now.

eng/common/scripts/Mark-ReleasePlanCompletion.ps1
praveenkuttappan: Correct

PR #11572, opened 2025-08-06
Updating MCP for new parameter

maririos:
With PR https://github.com/Azure/azure-rest-api-specs/pull/36502 we need to update our code too. 

FRI @raych1 to keep us in the loop of when the change happens so the MCP is not broken


maririos:
closing in favor of https://github.com/Azure/azure-sdk-tools/pull/11583

Issue #11167, opened 2025-07-10
[APIView] Send diagnostics to Copilot to include in de-duplication


tjprescott:
Currently, when we request a review, we send any existing comments to Copilot so that its de-duplication stages ensure that comments are duplicated with multiple runs. However, this can result in Copilot making a comment where APIView already displays a diagnostic, which is also redundant.

We should consider sending APIView Diagnostics to copilot as well so we can include that info in the de-duplication stage.

AlitzelMendez:
@tjprescott do you need a specific format/structure for those?

tjprescott:
I believe we just use the JSON of whatever "existing comments" are sent in. If the keys are similiar that's probably good enough.

tjprescott:
However, this issue also isn't high priority right now.

PR #11573, opened 2025-08-07
Send diagnostics to copilot agent

AlitzelMendez:
Fixes: https://github.com/Azure/azure-sdk-tools/issues/11167

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
+++ b/src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
@@ -12,6 +12,7 @@ using System.Text;
 using System.Text.Json;
 using System.Threading.Tasks;
 using ApiView;
+using APIView;
 using APIViewWeb.Helpers;
 using APIViewWeb.Hubs;
 using APIViewWeb.LeanModels;
@@ -345,6 +346,7 @@ namespace APIViewWeb.Managers
             List<ApiViewAgentComment> existingCommentInfo = AgentHelpers.BuildCommentsForAgent(comments: reviewComments, codeFile: activeCodeFile);
             var activeCodeLines = activeCodeFile.CodeFile.GetApiLines(skipDocs: true);
             var activeApiOutline = activeCodeFile.CodeFile.GetApiOutlineText();
+            List<CodeDiagnostic> diagnostics = activeCodeFile.CodeFile.Diagnostics.ToList();
 
             var copilotEndpoint = _configuration["CopilotServiceEndpoint"];
             var startUrl = $"{copilotEndpoint}/api-review/start";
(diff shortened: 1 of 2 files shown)


src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
```diff
@@ -12,6 +12,7 @@
 using System.Text.Json;
 using System.Threading.Tasks;
 using ApiView;
+using APIView;
```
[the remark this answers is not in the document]
AlitzelMendez: they are different, upper case, lower case, pointing to different things :/

src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
```diff
@@ -354,7 +356,8 @@ public async Task GenerateAIReview(ClaimsPrincipal user, string reviewId, string
                 { "language", LanguageServiceHelpers.GetLanguageAliasForCopilotService(activeApiRevision.Language) },
                 { "target", String.Join("\\n", activeCodeLines.Select(item => item.lineText.Trim())) },
                 { "outline", activeApiOutline },
-                { "comments", existingCommentInfo }
+                { "comments", existingCommentInfo },
+                { "diagnostics", diagnostics }
```
tjprescott, line 360: So I don't think we can just dump the diagnostics like this, because they will include stuff Copilot doesn't know or care about like lineId. We would need to pass things in similar to existingCommentInfo, where the diagnostic is associated with a line NUMBER. That's the minimal transformation we would need.

tjprescott (commented):
Also, https://github.com/Azure/azure-sdk-tools/issues/11167 isn't scheduled for this sprint, so if you feel like the work involved will start getting lengthy, we can always defer it. If you think it's low hanging fruit we can easily get it, then let's continue

src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
tjprescott: Better yet, if we could map a Diagnostic into the existingCommentInfo format and just pass it in as a comment, that would "just work". Copilot won't really care if the contents were a "comment" or a "diagnostic". It only cares what meaning they convey, so from that perspective they really are equivalent.

AlitzelMendez: Addressed. The main difference is that these comments will only include:
- LineNumber = elementIdToLineNumber.TryGetValue(diagnostic.TargetId, out int id) ? id : -1;
- CommentText = diagnostic.Text;

tjprescott: If the diagnostic doesn't map to a line number, we should just filter it out instead of passing it with a -1 line number because copilot will never be able to use that data so it will just be needless junk on the wire.

PR #11597, opened 2025-08-08
[sdk-sample-generator] Add support for user prompts

deyaaeldeen:
## Summary

Adds support for supplying a user-authored prompt from a Markdown file to drive sample generation directly, bypassing the “idea generation” step when provided.

## Usage

- With client API docs (Storage example):
```bash
npx genaiscript run generate --apply-edits --vars \
  client-api=.artifacts/azure-sdk-for-js/sdk/storage/storage-blob/review \
  language=TypeScript \
  user-prompt=.artifacts/test/test-user-prompt.md \
  out=.artifacts/test/ts-output \
  skip-review=true
```

### DIFF
--- a/tools/sdk-sample-generator/genaisrc/src/parseUserPrompt.ts
+++ b/tools/sdk-sample-generator/genaisrc/src/parseUserPrompt.ts
@@ -1,0 +1,51 @@
+import * as fs from "node:fs/promises";
+import type { SampleIdea } from "./types.ts";
+
+/**
+ * Parses a user prompt from a markdown file and converts it to a SampleIdea
+ * @param promptPath Path to the markdown file containing the user prompt
+ * @returns A SampleIdea object based on the user prompt
+ */
+export async function parseUserPrompt(promptPath: string): Promise<SampleIdea> {
+    try {
+        const content = await fs.readFile(promptPath, "utf-8");
+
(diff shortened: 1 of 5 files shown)


tools/sdk-sample-generator/genaisrc/src/parseUserPrompt.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export async function parseUserPrompt(promptPath: string): Promise<SampleIdea> {
+    try {
+        const content = await fs.readFile(promptPath, "utf-8");
+
+        // Extract title from first # heading or use filename
+        const titleMatch = content.match(/^#\s+(.+)/m);
+        const title = titleMatch
+            ? titleMatch[1].trim()
+            : "User Provided Sample";
```
richardpark-msft, line 17: Perhaps it's better to just throw if things aren't formatted properly, just to make sure our starting conditions are correct.

Or, if you think they aren't important pieces of data, perhaps we don't care how the user idea is formatted?

deyaaeldeen: It is not critical to the process so I would feel bad if I error here. I want to keep the prompt freeform. We might opt later to remove this sample idea structure from this path, since a prompt is unstructured by definition.

PR #11605, opened 2025-08-08
Mgmt namespace enforcement 

helen229:
Currently pipeline checks for APIView and namespace status only for all data plane releases and only .NET management plane release. We need to change this to check ONLY Namespace approval status for the management plane for Java, JS, Python and Go in addition to current logic. It should not check for APIView approval and check for namespace approval only.

### DIFF
--- a/eng/common/scripts/Create-APIReview.ps1
+++ b/eng/common/scripts/Create-APIReview.ps1
@@ -241,6 +241,27 @@ function ProcessPackage($packageName)
                 }
                 Process-ReviewStatusCode $respCode $packageName $apiStatus $pkgNameStatus
 
+                # Management plane namespace approval only gating for Java, JS, Python, Go.
+                $isMgmtNamespaceOnly = (
+                  $pkgInfo.SdkType -eq 'mgmt' -and
+                  ($LanguageShort -in @('java','js','python','go'))
+                )
+                if ($isMgmtNamespaceOnly)
+                {
+                    if (!$pkgInfo.ReleaseStatus -or $pkgInfo.ReleaseStatus -eq 'Unreleased')
+                    {
+                        Write-Host "Management plane package $packageName ($LanguageShort) is Unreleased; skipping namespace approval enforcement for now."
+                        return 0
+                    }
+                    if (-not $pkgNameStatus.IsApproved)
+                    {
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Create-APIReview.ps1
```diff
@@ -241,6 +241,27 @@ function ProcessPackage($packageName)
                 }
                 Process-ReviewStatusCode $respCode $packageName $apiStatus $pkgNameStatus
 
+                # Management plane namespace approval only gating for Java, JS, Python, Go.
```
maririos: looking at the structure of this code, the enforcement shouldn't be here. There is already a function `IsApiviewStatusCheckRequired` that is doing that. 
That would also avoid the code duplication

https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Create-APIReview.ps1#L169

praveenkuttappan: Correct. It should be changed in IsApiviewStatusCheckRequired

praveenkuttappan (changes requested):
Change is required in IsApiviewStatusCheckRequired only to make the check required for both `client` and `mgmt`

maririos (approved):

eng/common/scripts/Create-APIReview.ps1
```diff
@@ -166,7 +166,8 @@ function IsApiviewStatusCheckRequired($packageInfo)
         return &$IsApiviewStatusCheckRequiredFn $packageInfo
     }
 
-    if ($packageInfo.SdkType -eq "client" -and $packageInfo.IsNewSdk) {
+    # API review is required for both client and management plane packages
+    if (($packageInfo.SdkType -eq "client" -or $packageInfo.SdkType -eq "mgmt") -and $packageInfo.IsNewSdk) {
```
praveenkuttappan, line 170: This will still be an issue. We don't need API view approval for GA release. What we need to block is the preview release only for mgmt plane when namespace is not approved. Current change will block GA release if entire API review is not approved for mgmt.

praveenkuttappan: You should add a separate method for namespace check required and that should be called from https://github.com/Azure/azure-sdk-tools/pull/11605/files#diff-ee60aabbce995095134759c343bf23ef121e2ff90c9ab61f38051d7f989f86b3L255 to avoid blocking mgmt GA release.

PR #11606, opened 2025-08-08
[Devex Agent] Additional Contextual links

### LINKED ISSUE #11446: Enhance AzSDK MCP server tool to add contextual links in agentic SDK pull requests
## Summary

Currently, SDK pull requests generated via the AzSDK MCP server tool using agentic flow do not include sufficient contextual information. Specifically, they are missing:
- A link to the associated TypeSpec API specification pull request
- A link to the release plan
- A reference to the release plan work item

### Proposal

Update the MCP tool to automatically identify generated SDK pull requests and update their summaries to include the following additional context:
- Release plan link
- Release plan work item
- TypeSpec API pull request link

This will improve traceability and ensure that reviewers have all relevant context when reviewing SDK pull requests generated via the agentic flow.

### Impact
- Enhances the MCP server tool functionality
- Improves developer and reviewer experience
- Increases traceability for SDK releases

jeo02:
## Summary
- Added additional context to SDK generations PR's.
- Release plan link
- Release plan work item
- TypeSpec API pull request link
- API version
- Resolves #11446


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/GitHubService.cs
```diff
@@ -79,6 +79,8 @@ public interface IGitHubService
         public Task<PullRequest?> GetPullRequestForBranchAsync(string repoOwner, string repoName, string remoteBranch);
         public Task<Issue> GetIssueAsync(string repoOwner, string repoName, int issueNumber);
         public Task<IReadOnlyList<RepositoryContent>?> GetContentsAsync(string owner, string repoName, string path);
+        public Task UpdatePullRequestAsync(string repoOwner, string repoName, int pullRequestNumber, string title, string body, ItemState state);
+        public Task AddReleasePlanInfoInSdkAsync(string repoOwner, string repoName, int prNumber, string releasePlanLink, string specPrLink, string workItemLink, string apiVersion);
```
praveenkuttappan: This method should not be in GitHub service. GitHub service should have a method to update pull request description. APIs in GitHub service is just helper to send request to GitHub.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/GitHubService.cs
```diff
@@ -311,5 +324,49 @@ public async Task<Issue> GetIssueAsync(string repoOwner, string repoName, int is
...
+            {
+                links += $"\nSpec Pull Request: {specPrLink}";
+            }
+            if (!string.IsNullOrEmpty(apiVersion))
+            {
+                links += $"\nSpec API version: {apiVersion}";
+            }
+
+            // Check if the PR body already contains the release plan/spec PR info
+            if (!string.IsNullOrEmpty(pr.Body) && pr.Body.Contains(links))
```
praveenkuttappan: alternate approach check this is to add header in the PR body like `## Release plan details` and just check if PR body contains this header instead of checking entire content.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -389,8 +389,19 @@ public async Task<string> GetSDKPullRequestDetails(string language, int workItem
                     return $"SDK generation pipeline did not succeed. Status: {pipeline.Result?.ToString()}. For more details: {DevOpsService.GetPipelineUrl(buildId)}";
                 }
 
-                var data = await devopsService.GetSDKPullRequestFromPipelineRunAsync(buildId, language, workItemId);
-                return data;
+                var pr = await devopsService.GetSDKPullRequestFromPipelineRunAsync(buildId, language, workItemId);
+
+                if (gitHelper.IsPullRequestUrl(pr))
```
praveenkuttappan: You can avoid this check by using one regex to parse out all 3 components instead of having a helper in git. If regex parser fails, skip updating the info.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/MockServices/MockGitHubService.cs
```diff
@@ -59,6 +59,11 @@ public Task<Issue> GetIssueAsync(string repoOwner, string repoName, int issueNum
             return Task.FromResult(issue);
         }
 
+        public Task UpdatePullRequestAsync(string repoOwner, string repoName, int pullRequestNumber, string title, string body, ItemState state)
+        {
+            throw new NotImplementedException();
```
maririos: is there a reason for the tests not to be added in this PR?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -452,14 +466,54 @@ public async Task<string> LinkSdkPullRequestToReleasePlan(string language, strin
...
                 await devopsService.AddSdkInfoInReleasePlanAsync(releasePlan.WorkItemId, language, "", parsedLink);
-                return $"Successfully linked pull request to release plan {releasePlan.ReleasePlanId}, work item id {releasePlan.WorkItemId}";
+
+                // Parse PR URL components and update PR description if valid
+                var prUrlMatch = GITHUB_PR_URL_REGEX.Match(parsedLink);
+                if (prUrlMatch.Success)
+                {
+                    var repoOwner = prUrlMatch.Groups[1].Value;
+                    var prNumber = int.Parse(prUrlMatch.Groups[3].Value);
+                    await AddReleasePlanInfoInSdkAsync(repoOwner, repoName, prNumber, releasePlan.ReleasePlanLink, releasePlan.ActiveSpecPullRequest, releasePlan.WorkItemUrl, releasePlan.SpecAPIVersion);
```
maririos: if `await devopsService.AddSdkInfoInReleasePlanAsync(releasePlan.WorkItemId, language, "", parsedLink);` fails, does that mean that the release plan info won't be added to the SDK PR?
Is this dependency intentional?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -452,14 +466,54 @@ public async Task<string> LinkSdkPullRequestToReleasePlan(string language, strin
                 }
 
                 await devopsService.AddSdkInfoInReleasePlanAsync(releasePlan.WorkItemId, language, "", parsedLink);
-                return $"Successfully linked pull request to release plan {releasePlan.ReleasePlanId}, work item id {releasePlan.WorkItemId}";
+
+                // Parse PR URL components and update PR description if valid
+                var prUrlMatch = GITHUB_PR_URL_REGEX.Match(parsedLink);
```
maririos: `parsedLink` was already a url that was parsed. Do you need to parse it again?

maririos (commented):
Looks like the description of the original issue got updated and `Namespace GH issue` for management plane is now missing.

Could you add that property too?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -452,14 +466,54 @@ public async Task<string> LinkSdkPullRequestToReleasePlan(string language, strin
...
+            }
+            
+            // Check if the PR body already contains the release plan link (main indicator)
+            var header = "## Release Plan Details";
+            if (!string.IsNullOrEmpty(pr.Body) && pr.Body.Contains(header, StringComparison.OrdinalIgnoreCase))
+            {
+                return $"PR {repoOwner}/{repoName}#{prNumber} already contains release plan info. Skipping update.";
+            }
+
+            var links = $"{header}";
```
benbp: I would use `StringBuilder` here with `.AppendLine` that way newlines are cross platform (\n vs \r\n). Also just a bit easier to construct/read.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/MockServices/MockGitHubService.cs
jeo02: Oh woops put that there so I could build then forgot about it. I'll implement it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
jeo02: for the matching being successful itself not really since as you said we already know there is a link. But I do need the 3 groups which are not extracted from currently.

jeo02: https://github.com/Azure/azure-sdk-tools/blob/f191a018e8e0ab6f4380cffd8609daa42ada404a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs#L617-L622

Just get's the value vs I want the repo owner, repo name, and pr number.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -465,16 +457,10 @@ public async Task<string> LinkSdkPullRequestToReleasePlan(string language, strin
                     return $"Release plan with ID {releasePlanId} or work item ID {workItemId} is not found.";
                 }
 
-                await devopsService.AddSdkInfoInReleasePlanAsync(releasePlan.WorkItemId, language, "", parsedLink);
+                var sdkInfoInRelease = devopsService.AddSdkInfoInReleasePlanAsync(releasePlan.WorkItemId, language, "", parsedLink);
```
richardpark-msft: (outsider comment)

It's so weird, based on the names of these methods, that they don't have sort of ordering dependency.

jeo02: I guess the thought being neither methods are dependent on each others success. 
But I also see the point that if one method fails and we probably will need to rerun the mcp tool then why run the second part.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
maririos: 2 regex for the same thing feels like an overkill though. But I would leave it to you

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
maririos: Yes. I think they shouldn't depend on each other. They use similar data but they are not the same. If one fails, I would expect the tool to retry it. I won't expect the tool to fail for both.
I don't think one has priority over the other one either

maririos: also, maririos lol not marianan

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -455,8 +461,12 @@ public async Task<string> LinkSdkPullRequestToReleasePlan(string language, strin
                     return $"Release plan with ID {releasePlanId} or work item ID {workItemId} is not found.";
                 }
 
-                await devopsService.AddSdkInfoInReleasePlanAsync(releasePlan.WorkItemId, language, "", parsedLink);
-                return $"Successfully linked pull request to release plan {releasePlan.ReleasePlanId}, work item id {releasePlan.WorkItemId}";
+                var sdkInfoInRelease = devopsService.AddSdkInfoInReleasePlanAsync(releasePlan.WorkItemId, language, "", parsedLink);
+                var releaseInfoInSdk = AddReleasePlanInfoInSdkAsync(parsedLink, releasePlan);
```
maririos: I don't think this should be here. If this method is only to link a PR to a release plan, I don't see why we need to also add the release plan to the SDK PR.
That is already done by the `GetSDKPullRequestDetails`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -455,8 +461,12 @@ public async Task<string> LinkSdkPullRequestToReleasePlan(string language, strin
...
+
+                await Task.WhenAll(sdkInfoInRelease, releaseInfoInSdk);
```
maririos, line 466: it would also eliminate the need to run this

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -393,8 +395,12 @@ public async Task<string> GetSDKPullRequestDetails(string language, int workItem
                     return $"SDK generation pipeline did not succeed. Status: {pipeline.Result?.ToString()}. For more details: {DevOpsService.GetPipelineUrl(buildId)}";
                 }
 
-                var data = await devopsService.GetSDKPullRequestFromPipelineRunAsync(buildId, language, workItemId);
-                return data;
+                var pr = await devopsService.GetSDKPullRequestFromPipelineRunAsync(buildId, language, workItemId);
+
+                var rp = await devopsService.GetReleasePlanForWorkItemAsync(workItemId);
+                await AddReleasePlanInfoInSdkAsync(pr, rp);
```
maririos: given that this method is getting the pull request and adding context to it, we should consider renaming it and maybe the description? although copilot shouldn't care if we add the link to the SDK PR....

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
```diff
@@ -465,6 +475,48 @@ public async Task<string> LinkSdkPullRequestToReleasePlan(string language, strin
             }
         }
 
+        public async Task AddReleasePlanInfoInSdkAsync(string pullRequestUrl, ReleasePlan releasePlan)
```
maririos: should this be private instead? I don't think people directly should call this

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
jeo02: It would be helpful from the dev standpoint to add it maybe to the xml summary but I don't think it matters for copilot itself.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
jeo02: The reason it was added here was for the case where the SDK generation is done before creation of release plan. This MCP tool is the one called to then link the release plan and sdk pr in that scenario. 

The GetSDKPullRequestDetails wouldn't be able to link them in that scenario.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
maririos: let's rename it

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
maririos: Ohh interesting. Could you clarify more how this flow works?
i.e
- When is the release plane required to be created? or prompt for it to be created?
- Someone generates the SDKs, at what point to they know they need to create a release plan and associate the PR links to that release plan?
- What is keeping track of all the PRs created curing the SDK generation phase if there was no release plan?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
richardpark-msft: Also, if you're doing this all the time you could just make a small type and pass that around, instead of the string URL.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecWorkflowTool/SpecWorkFlowTool.cs
jeo02: I believe one of the big reasons was local generation. Here is a breakdown from Copilot.
1. [create-release-plan.prompt.md](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/prompts/create-release-plan.prompt.md#step-6-link-sdk-pull-requests-if-applicable) - Step 6:
When: After a release plan is created or found
Context: If the user has already created SDK pull requests locally for any programming language
Purpose: To link existing SDK pull requests to the newly created or existing release plan
Process:
Ask user if they have SDK pull requests
Collect pull request links from user
Use LinkSdkPullRequestToReleasePlan to link each SDK PR to the release plan
2. [run-sdk-gen-pipeline.prompt.md](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/prompts/run-sdk-gen-pipeline.prompt.md#step-1-check-for-existing-sdk-pull-requests) - Step 1:
When: At the beginning of SDK generation pipeline process
Context: If SDK pull requests already exist from local SDK generation
Purpose: To link existing SDK pull requests to the release plan before generating new ones
Process:
Check if SDK pull requests exist from local SDK generation
If SDK PR exists for a language, skip SDK generation for that language
Link existing SDK pull request to release plan
Summary: The tool is called to link pre-existing SDK pull requests (created locally or previously) to a release plan, either during release plan creation or before running the SDK generation pipeline to avoid duplicate SDK generation.

jeo02: Last question I assume either user needs to provide all the information to create the release plan or the copilot chat history would have everything needed to create the release plan.

maririos: Then disregard my comment :)

maririos (approved):

PR #11611, opened 2025-08-08
Test recording/agent-in-agent workflow proof of concept

timovv:
Hey folks, this draft PR demonstrates a proof-of-concept for how we might create our own smaller agents which work together as part of a larger workflow (I called them "microagents" in this PR, and I wrote up some thoughts on the problem the approach is trying to solve a few weeks ago in [this doc](https://microsoft.sharepoint.com/:fl:/g/contentstorage/x8FNO-xtskuCRX2_fMTHLdjRwgl20jZAsg1yEsU4OFI/EaZ1IxkfbdFMmsKqXx_KHBUBICbQ0IbjjGKj5a9uiktqvA?e=dInNBv&nav=cz0lMkZjb250ZW50c3RvcmFnZSUyRng4Rk5PLXh0c2t1Q1JYMl9mTVRITGRqUndnbDIwalpBc2cxeUVzVTRPRkkmZD1iJTIxSzFaVzYtdDRxMG0tUlA1bS1VbVlLZzhZNVprRlVkMUpzdG44QWRtN081VGo2alhlU2tvV1JaTUFZM0dLMHJBYSZmPTAxQUFTWFE1NUdPVVJSU0gzTjJGR0pWUVZLTDRQNFVIQVYmYz0lMkYmYT1Mb29wQXBwJnA9JTQwZmx1aWR4JTJGbG9vcC1wYWdlLWNvbnRhaW5lciZ4PSU3QiUyMnclMjIlM0ElMjJUMFJUVUh4dGFXTnliM052Wm5RdWMyaGhjbVZ3YjJsdWRDNWpiMjE4WWlGTE1WcFhOaTEwTkhFd2JTMVNVRFZ0TFZWdFdVdG5PRmsxV210R1ZXUXhTbk4wYmpoQlpHMDNUelZVYWpacVdHVlRhMjlYVWxwTlFWa3pSMHN3Y2tGaGZEQXhRVUZUV0ZFMVdVRlFWVlZSU0RkVFRGTktSMXBSV0VFM04wNU1Xa1pKU0ZnJTNEJTIyJTJDJTIyaSUyMiUzQSUyMmZlZDMxNTdhLTZlMjItNDJjMi05OGU5LTMxNjQ0ODEzYjdiZCUyMiU3RA%3D%3D)). Having a group of agents with a small context window and toolset and feeding the results from one agent to the next programmatically has two benefits: increased reliability of the agent's output due to the reduced context, and greater determinism, since the tool program maintains stays in charge of the control flow. This is in contrast to the alternative, less reliable, approach where the agent itself is responsible for high-level control flow. Both approaches have their pros and cons.

I'm not sure that we do want to go ahead with this approach of having mini-agents in the CLI or not, but I did want to at least demonstrate here that it (a) is possible and (b) doesn't require _that_ much code. The idea is that our MCP tools/CLI commands will be able to invoke an agent for tasks using a shared set of common tools that we write once and can use multiple times.

The example I implemented that uses the pattern is a tool which helps you run your recorded tests in a loop. Different agents are responsible for a few different things, such as:
- Extracting test failure messages from the test run output
- Identifying and extracting the implementation of failing test cases from the test code
- Providing suggestions and possible diagnoses for test failures taking into account the extracted test implementation, failure message, and test recording content.

At the moment, the tool does not actually make updates to the test files. It merely provides suggestions. But this is something I want to implement before merging.

Apologies for any poor C# in this PR, it is not my first language :). The PR still needs validation, and tests, but I wanted to get something out there to demonstrate the approach early on before I continue down this rabbit hole a bit further. Some to-dos:

- [ ] Write tests for the tools and the agent host.
- [ ] Implement tools for updating files and better code search (e.g fuzzy find) like VS code has.
- [ ] Add a mechanism for tool call approval, for destructive tools such as writing to a file and 
- [ ] Harden existing tools, e.g., to ensure the agent can't break out of the base directory using `../../../` in a tool call
- [ ] Allow the recorded test helper to actually make changes to test files to fix them through an "update file" tool
- [ ] Improve logging and output for the recorded test helper
- [ ] Make the recorded test helper an MCP tool as well, at the moment it's just a CLI tool for simplicity's sake.

<details>
<summary><h2>Copilot-generated summary</h2></summary>
This pull request introduces a new microagent framework to the CLI, enabling the creation and orchestration of lightweight, task-focused agents powered by LLMs and tool calls. It adds core abstractions, supporting infrastructure, and two sample tools for file operations. Additionally, it introduces a test runner service and registers these new services for dependency injection.

**Microagent Framework Introduction:**

* Added a new `Microagent<TResult>` record to represent agent definitions, including system prompt, tools, model, and iteration limits (`Microagent.cs`).
* Implemented `IMicroagentHostService` and `MicroagentHostService`, which manage agent execution loops, tool invocation, and result handling (`MicroagentHostService.cs`).
* Defined the `IAgentTool` interface and an abstract `AgentTool<TInput, TOutput>` base class for strongly-typed tool implementations, supporting JSON (de)serialization (`MicroagentTool.cs`).
* Added `ToolHelpers` utility for extracting JSON schema representations from tool input/output types, leveraging type and property descriptions (`ToolHelpers.cs`).

**Sample Tools:**

* Implemented `ListFilesTool` for listing files in a directory, with support for recursion and glob patterns (`ListFilesTool.cs`).
* Implemented `ReadFileTool` for reading file contents by relative path (`ReadFileTool.cs`).

**Test Runner Service:**

* Introduced `ITestRunnerService` and a `JavaScriptTestRunnerService` implementation for running JavaScript tests via `pnpm`, supporting test modes (record, playback, live) (`TestRunnerService.cs`).

**Dependency Injection and Tool Registration:**

* Registered `IMicroagentHostService`, `ITestRunnerService`, and related services in the DI container (`ServiceRegistrations.cs`). [[1]](diffhunk://#diff-94e7bc208d2b8202471670e0e657d96742b82220d745042c2210d14749d97b60R5) [[2]](diffhunk://#diff-94e7bc208d2b8202471670e0e657d96742b82220d745042c2210d14749d97b60R33-R35)
* Minor update to tool registration in `SharedOptions.cs` (no functional impact). [[1]](diffhunk://#diff-6a9640332d11c1d0c2a2f505907e2df0dd66b529fc6e3751555b7e7ee4803f45L7-R8) [[2]](diffhunk://#diff-6a9640332d11c1d0c2a2f505907e2df0dd66b529fc6e3751555b7e7ee4803f45R32)
</details>

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedOptions.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedOptions.cs
@@ -4,8 +4,8 @@ using System.IO.Enumeration;
 using Azure.Sdk.Tools.Cli.Tools;
 using Azure.Sdk.Tools.Cli.Tools.HelloWorldTool;
 using Azure.Sdk.Tools.Cli.Tools.HostServer;
-using Azure.Sdk.Tools.Cli.Tools.SdkRelease;
 using Azure.Sdk.Tools.Cli.Tools.ReleaseReadiness;
+using Azure.Sdk.Tools.Cli.Tools.SdkRelease;
 
 namespace Azure.Sdk.Tools.Cli.Commands
 {
@@ -29,6 +29,7 @@ namespace Azure.Sdk.Tools.Cli.Commands
             typeof(SpecWorkflowTool),
             typeof(SpecValidationTools),
             typeof(TestAnalysisTool),
+            typeof(RecordTestsTool),
             #if DEBUG
             // only add this tool in debug mode
             typeof(HelloWorldTool),
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Microagent.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Microagent.cs
@@ -1,0 +1,20 @@
+namespace Azure.Sdk.Tools.Cli.Microagents;
+
+/// <summary>
+/// A definition of a microagent, including a system prompt, a list of tools, and a model.
+/// 
+/// A microagent is a lightweight, focused agent designed to perform a specific task or set of tasks within a larger system.
+/// It operates an agent loop (LLM <-> tool call) until the LLM determines that its task is complete, at which point
+/// it calls the special sentinel "exit" tool with the result.
+/// </summary>
+/// <typeparam name="TResult">A type representing the schema of the result of the agent's execution</typeparam>
+/// <param name="SystemPrompt">The system prompt provided to the agent.</param>
+/// <param name="Tools">A list of tools that are made available to the agent. For a simple LLM call without tools, this can be left empty.</param>
+/// <param name="Model">The model that this microagent is designed for.</param>
+/// <param name="MaxIterations">The maximum number of iterations in the agent loop. If this is exceeded, the microagent run is deemed to have failed.</param>
+public record Microagent<TResult>(
+    string SystemPrompt,
+    IEnumerable<IAgentTool> Tools = null,
+    string Model = "gpt-4.1",
+    int MaxIterations = 1_000
+);
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/MicroagentHostService.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/MicroagentHostService.cs
@@ -1,0 +1,106 @@
+
+using System.ComponentModel;
+using Azure.AI.OpenAI;
+using Microsoft.TeamFoundation.WorkItemTracking.WebApi;
+using OpenAI.Chat;
+
+namespace Azure.Sdk.Tools.Cli.Microagents;
+
+public interface IMicroagentHostService
+{
+    Task<TResult> RunAgentToCompletionAsync<TResult>(Microagent<TResult> agentDefinition, CancellationToken ct = default);
+}
+
+public class MicroagentHostService(AzureOpenAIClient openAI, ILogger<MicroagentHostService> logger) : IMicroagentHostService
+{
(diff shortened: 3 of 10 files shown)


benbp:
There are also some orchestration options and attributes to populate the tool list schema available in [semantic kernel](https://learn.microsoft.com/en-us/semantic-kernel/frameworks/agent/agent-orchestration/?pivots=programming-language-csharp). Though I like that this is small and self contained, easily understandable.

How much we want to bake agents into the CLI or MCP server has been an open question for me, but I don't think it's something we'll be able to completely avoid. While I love the simplicity of the flat mcp client -> tool call structure, I think that sometimes we will need control over model, specialized prompts, chat history and other infra for things like RAG.

An open question I have in this area is how the tool gets limited in portability the more agent to agent dependencies we create, as the CLI may end up having a lot of infra/auth dependencies. If we're just doing it for model selection and prompts, I'm more tempted to invest effort in making sure we have reliability cross model (or can define some sort of minimum mcp client model assertion per tool). For flows like test proxy recording and interpretation, how much of that is really only possible/optimal with agent orchestration vs. leveraging an mcp client (or an agent calling the CLI in the shell).

Overall I like this idea and it would be interesting to have the cli be a home for it, though I'm still not sure where I land on how we balance which scenarios support mcp vs. agent flows vs. both. Then there are maintenance toil costs to consider like model hosting and upgrades (e.g. gpt-5 deprecating everything).

haolingdong-msft:
@timovv, Thanks a lot for the PoC! It helps me a lot to learn more on the microagent idea!
I like the idea of having a small agent specialized at specific tasks, with minimum needed tools and context. Maybe in the future, we can have 'test micro agent', 'customized code micro agent', 'readme agent' etc.

timovv:
Closing since we merged #11660. I'll follow up with the recorded test stuff in another PR

PR #11614, opened 2025-08-08
Make process helper async and improve timeout and cross platform utility

benbp:
Make process helper async and improve timeout and cross platform utility

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
@@ -1,37 +1,71 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 using System.Text;
+using Microsoft.Graph.Drives.Item.Items.Item.GetActivitiesByInterval;
+using Microsoft.Graph.Models.TermStore;
 
 namespace Azure.Sdk.Tools.Cli.Helpers
 {
     public interface IProcessHelper
     {
-        public ProcessResult RunProcess(string command, string[] args, string workingDirectory);
-        public ProcessResult RunProcess(string command, string[] args, string workingDirectory, int processTimeout);
+        public IProcessHelper CreateForCrossPlatform(string linuxCommand, string[] linuxArgs, string windowsCommand, string[] windowsArgs, string workingDirectory);
+        public Task<ProcessResult> RunProcess(CancellationToken ct = default);
+        public Task<ProcessResult> RunProcess(TimeSpan timeout, CancellationToken ct = default);
+        public Task<ProcessResult> RunProcess(string command, string[] args, string workingDirectory, CancellationToken ct = default);
+        public Task<ProcessResult> RunProcess(string command, string[] args, string workingDirectory, TimeSpan timeout, CancellationToken ct = default);
     }
 
     public class ProcessHelper(ILogger<ProcessHelper> logger) : IProcessHelper
     {
 
-        private readonly int DefaultProcessTimeoutMs = 120_000; // Default timeout of 2 minutes
+        private const int DEFAULT_PROCESS_TIMEOUT_SECONDS = 120;  // Default timeout of 2 minutes
 
         private readonly ILogger<ProcessHelper> logger = logger;
 
-        /// <summary>
-        /// Runs a process with the specified command and arguments in the given working directory.
-        /// </summary>
-        /// <param name="command">The command to run.</param>
-        /// <param name="args">The arguments to pass to the command.</param>
-        /// <param name="workingDirectory">The working directory for the process.</param>
-        /// <returns>A <see cref="ProcessResult"/> containing the output and exit code of the process.</returns>
-        /// <remarks>
-        /// The default timeout for the process is set to 120 seconds (2 minutes).
-        /// If the process does not complete within this time, it will be terminated.
-        /// </remarks>
-        public ProcessResult RunProcess(string command, string[] args, string workingDirectory)
+        private bool _initializedForCrossPlatform { get; set; } = false;
+        private string _unixCommand { get; set; }
+        private string[] _unixArgs { get; set; }
+        private string _windowsCommand { get; set; }
+        private string[] _windowsArgs { get; set; }
+        private string _workingDirectory { get; set; }
+
+        public IProcessHelper CreateForCrossPlatform(string unixCommand, string[] unixArgs, string windowsCommand, string[] windowsArgs, string workingDirectory)
         {
-            return RunProcess(command, args, workingDirectory, DefaultProcessTimeoutMs);
+            return new ProcessHelper(logger)
+            {
+                _initializedForCrossPlatform = true,
+                _unixCommand = unixCommand,
+                _unixArgs = unixArgs,
+                _windowsCommand = windowsCommand,
+                _windowsArgs = windowsArgs,
+                _workingDirectory = workingDirectory
+            };
+        }
+
+        public async Task<ProcessResult> RunProcess(CancellationToken ct)
+        {
+            var timeout = TimeSpan.FromSeconds(DEFAULT_PROCESS_TIMEOUT_SECONDS);
+            return await RunProcess(timeout, ct);
+        }
+
+        public async Task<ProcessResult> RunProcess(TimeSpan timeout, CancellationToken ct)
+        {
+            if (!_initializedForCrossPlatform)
+            {
+                throw new InvalidOperationException("ProcessHelper is not properly initialized. Call CreateForCrossPlatform() first.");
+            }
+
+            var command = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? _windowsCommand : _unixCommand;
+            var args = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? _windowsArgs : _unixArgs;
+            return await RunProcess(command, args, _workingDirectory, timeout, ct);
+        }
+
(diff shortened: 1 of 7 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
```diff
@@ -84,22 +129,39 @@ public ProcessResult RunProcess(string command, string[] args, string workingDir
...
+                try
+                {
+                    await process.WaitForExitAsync(linkedCts.Token);
+                }
+                catch (OperationCanceledException) when (timeoutCts.IsCancellationRequested)
+                {
+                    output.AppendLine();
+                    output.AppendLine($"Process '{command}' timed out after {timeout.TotalMilliseconds}ms");
+                    logger.LogError("Process '{command}' timed out after {timeout}ms", command, timeout.TotalMilliseconds);
+                    return new ProcessResult { Output = output.ToString(), ExitCode = 124 };
```
richardpark-msft: We should avoid overloading ExitCode like this. If our worry is that it's difficult to tell if a ProcessResult represents a failure we can add a method, like `bool Failed()`, so people aren't making assumptions based on the exit code.

The other option is just to let the exception out. We could create a subclass and just add in Output as one of the fields.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
```diff
@@ -84,22 +129,39 @@ public ProcessResult RunProcess(string command, string[] args, string workingDir
...
+                }
+                catch (Exception ex)
+                {
+                    output.AppendLine();
+                    output.AppendLine($"Process '{command}' failed with exception: {ex.Message}");
```
richardpark-msft: We should document that we're adding in this output - ie, the user shouldn't expect the command's output to be _only_ what the shell command outputted, on failure.

benbp: Yeah I was hesitant about this as well. Maybe a prefix or just omit it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
benbp: There is no original process exit code in this timeout case. Or are you saying we should just always have exit code be 1?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
```diff
@@ -1,37 +1,71 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 using System.Text;
+using Microsoft.Graph.Drives.Item.Items.Item.GetActivitiesByInterval;
+using Microsoft.Graph.Models.TermStore;
```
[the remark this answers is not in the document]
richardpark-msft: @benbp, part of me wonders what code you were going to add to this, that you deleted, based on this dependency.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
```diff
@@ -40,13 +74,24 @@ public ProcessResult RunProcess(string command, string[] args, string workingDir
...
         /// <param name="workingDirectory">The working directory for the process.</param>
-        /// <param name="processTimeout">The timeout in milliseconds for the process to complete.</param>
+        /// <param name="timeout">The timeout for the process to complete.</param>
+        /// <param name="ct">The cancellation token to monitor for cancellation requests.</param>
         /// <returns>A <see cref="ProcessResult"/> containing the output and exit code of the process.</returns>
         /// <remarks>
         /// If the process does not complete within the specified timeout, it will be terminated.
         /// </remarks>
-        public ProcessResult RunProcess(string command, string[] args, string workingDirectory, int processTimeout)
+        public async Task<ProcessResult> RunProcess(string command, string[] args, string workingDirectory, TimeSpan timeout, CancellationToken ct)
```
richardpark-msft: I have a variation on this that I'd like to get your take on. There were a couple of changes:

1. It just took a ProcessStartInfo() directly, rather than trying to split out parameters. This also made it so you could configure other things within the process, and use ArgumentList if you preferred to avoid having to handle quoting yourself.
2. It automatically added streaming output (similar to what you're doing) with one big change - it used locks, and put stderr and stdout into the same stream. I think you need this, as a human, to really understand what broke because the stdout and stderr often complement each other.

I don't think we need to replace this one, but I'd like to add a variant that works like that.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
```diff
@@ -84,22 +129,39 @@ public ProcessResult RunProcess(string command, string[] args, string workingDir
...
+                    output.AppendLine($"Process '{command}' failed with exception: {ex.Message}");
+                    logger.LogError(ex, "Process '{command}' failed", command);
+                    return new ProcessResult { Output = output.ToString(), ExitCode = 1 };
+                }
 
+                exitCode = process.ExitCode;
                 if (process.ExitCode != 0)
                 {
-                    output.Append($"{Environment.NewLine}Process failed.");
+                    output.AppendLine($"Process '{command}' failed with exit code {process.ExitCode}");
```
richardpark-msft: I think 'command' is just 'cmd.exe', right? We should add in 'command and args', otherwise we're not getting as much info as we want.

richardpark-msft:
@benbp, reviewing some more.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
```diff
@@ -64,7 +109,7 @@ public ProcessResult RunProcess(string command, string[] args, string workingDir
             }
 
             var output = new StringBuilder();
-            int exitCode = -1;
+            int exitCode = 1;
 
             using (var process = new Process())
             {
```
richardpark-msft: I can't see any docs that guarantee that OutputDataReceived and ErrorDataReceived won't be called simultaneously so we should probably `lock(output)` in the two callsbacks before we do the AppendLine().

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
richardpark-msft: No, I was trying to think about how we might have two types of output if we still want to just return in this case. One that has an ExitCode, and another indicating that it was cancelled. Just to make the calling code not have to be too weird.

However, it's still weird, so I think we're just going with throwing an exception with extra data.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/ProcessHelper.cs
benbp: We put stdout and stderr into the same stream currently, but I'm wondering if we should improve it so that when we print it to the user we can preserve both the message ordering and the stdout/stderr differences vs. just converting it all to stdout. We could either prefix each line, or better just have a list of message tuples like `[(stdout/stderr, message)]`

richardpark-msft: Maybe just two functions - one that blends stdout/stderr into a beautiful stew, and one that gives you them separate. That way the return value can be different and (hopefully) still be clear.

We're basically creating an overload that works like Go's .CombinedOutput() and one that doesn't do that.

benbp: @richardpark-msft take another look when you get the chance. I overhauled the process helpers to take different options types. Now there is only one actual core implementation of `Run`, and the option classes handle setting their commands/args/etc. based on platform and other custom logic.

```
// Simple case
var options = new ProcessOptions("sleep", ["1"]);
await processHelper.Run(options, ct);

// Handle cases where commands/args may differ across unix and windows
var result = await processHelper.Run(new(
                "goimports", ["-w", "."],
                "gofmt.exe", ["-w", "."],
                workingDirectory: packagePath
             ), ct);

// Customized options class, first arg is 'package' which gets turned into npx under the hood
var npxOptions = new NpxOptions(
    "@azure-tools/typespec-client-generator-cli",
    ["tsp-client", "convert", "--swagger-readme", pathToSwaggerReadme, "--output-dir", outputDirectory]);
await npxHelper.Run(npxOptions, ct);
```

We can expose the other `ProcessStartInfo` properties too but I don't think they are necessary at the moment.

The process helper also streams live output to the console according to stdout/stderr, prefixed by the "short name", e.g.:

```bash
# Process output from the npx helper
[@azure-tools/typespec-client-generator-cli] Using output directory '/tmp/tsptest/'

# Process output from the powershell helper
[azsdk_example_03f98c.ps1] 1: foobar
[azsdk_example_03f98c.ps1] 2: foobar
[azsdk_example_03f98c.ps1] Write-Error: Test error message, no failure
[azsdk_example_03f98c.ps1] 3: foobar
```

FYI @chrisradek I also updated the tsp tool to handle a new case where we are in CLI mode but have already streamed the output to console (the tool was logging error output additionally on failure).

l0lawrence:
Should new_tool.md be updated to reflect these changes?

PR #11626, opened 2025-08-11
[tsp-client] Support emitter-output-dir

### LINKED ISSUE #10823: tsp-client should attempt to reuse emitter-output-dir rather than package-dir
package-dir is not supported by most emitters, but it is required to be included in the tspconfig.yaml for the tsp-client init command. This will cause problems if attempting to compile directly using `tsp compile`. We can instead leverage the `service-dir` parameter along with the TypeSpec defined emitter-output-dir property, e.g.:

```emitter-output-dir: "{service-dir}/azure-eventgrid"```

### PR DESCRIPTION
**Pull request by catalinaperalta**
- Refactor init code to support `emitter-output-dir` and use a function for shared processing between the url and local spec scenario
- Add tests
- Some other minor refactoring/improvements to variable names. 

Fixes #10823

### CODE REVIEW

File: tools/tsp-client/src/commands.ts [hunk 1]
```diff
@@ -33,6 +39,106 @@ import { sortOpenAPIDocument } from "@azure-tools/typespec-autorest";
...
+  }
+
+  // Check for relevant package path variables and resolve
+  let emitterOutputDir = tspConfigData?.options?.[emitterData.emitter]?.["emitter-output-dir"];
+  const packageDir = tspConfigData?.options?.[emitterData.emitter]?.["package-dir"];
+  let newPackageDir = outputDir;
+  if (emitterOutputDir) {
+    const [options, _] = await resolveCompilerOptions(NodeHost, {
+      cwd: process.cwd(),
+      entrypoint: "main.tsp",
```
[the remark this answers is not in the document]
catalinaperalta: This is a dummy value, since we're only trying to resolve the interpolated variables in tspconfig.yaml

tools/tsp-client/src/commands.ts
```diff
@@ -33,6 +39,106 @@ import { sortOpenAPIDocument } from "@azure-tools/typespec-autorest";
...
+    emitterOutputDir = options.options?.[emitterData.emitter]?.["emitter-output-dir"];
+    Logger.info(`Resolved emitter-output-dir for ${emitterData.emitter}: ${emitterOutputDir}`);
+    newPackageDir = resolve(emitterOutputDir);
+  } else if (packageDir) {
+    // Warn that this behavior is deprecated
+    Logger.warn(
+      `Please update your tspconfig.yaml to include the "emitter-output-dir" option under the "${emitterData.emitter}" emitter options. "package-dir" support is deprecated and will be removed in future versions.`,
+    );
+    // If no emitter-output-dir is specified, fall back to the legacy package-dir path resolution for the new package directory
+    newPackageDir = resolve(
```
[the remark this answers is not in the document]
catalinaperalta: We're widely using typespec/compiler path functions, one day we can migrate over to node path if we deem it necessary

tools/tsp-client/src/commands.ts
```diff
@@ -33,6 +39,106 @@ import { sortOpenAPIDocument } from "@azure-tools/typespec-autorest";
...
+      cwd: process.cwd(),
+      entrypoint: "main.tsp",
+      configPath: tspConfigPath,
+      overrides: {
+        outputDir: outputDir,
+      },
+    });
+    emitterOutputDir = options.options?.[emitterData.emitter]?.["emitter-output-dir"];
+    Logger.info(`Resolved emitter-output-dir for ${emitterData.emitter}: ${emitterOutputDir}`);
+    newPackageDir = resolve(emitterOutputDir);
```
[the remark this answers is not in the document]
catalinaperalta: same as other response

tools/tsp-client/src/commands.ts
```diff
@@ -33,6 +39,106 @@ import { sortOpenAPIDocument } from "@azure-tools/typespec-autorest";
...
+    tspclientGlobalConfigData,
+    emitterPackageOverride,
+  );
+  if (emitterData.path) {
+    // store relative path to repo root
+    tspLocationData.emitterPackageJsonPath = emitterData.path;
+  }
+
+  // Check for relevant package path variables and resolve
+  let emitterOutputDir = tspConfigData?.options?.[emitterData.emitter]?.["emitter-output-dir"];
```
jorgerangel-msft, line 100: @JoshLove-msft - I think this means we'll need to add this property to all the tspconfig.yaml files for all the .net libraries that support the new emitters? Looks like this property is just being parsed directly from the yaml.

JoshLove-msft: I think this change will have to happen for all tspconfig entries - not just the new emitter. Essentially package-dir is being replaced by emitter-ouput-dir.

catalinaperalta: this is a wide-ranging change that will be configured for all emitters

tools/tsp-client/test/examples/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml
```diff
@@ -13,30 +13,30 @@ options:
     emitter-output-dir: "{project-root}/.."
     output-file: "{azure-resource-provider-folder}/{service-name}/{version-status}/{version}/widgets.json"
   "@azure-tools/typespec-python":
-    package-dir: "azure-contoso-widgetmanager"
+    emitter-output-dir: "{output-dir}/{service-dir}/azure-contoso-widgetmanager"
```
JoshLove-msft, line 16: Where does output-dir get specified?

tools/tsp-client/src/commands.ts
```diff
@@ -33,6 +39,105 @@ import { sortOpenAPIDocument } from "@azure-tools/typespec-autorest";
...
+      cwd: process.cwd(),
+      entrypoint: "main.tsp",
+      configPath: tspConfigPath,
+      overrides: {
+        outputDir: outputDir,
+      },
+    });
+    emitterOutputDir = options.options?.[emitterData.emitter]?.["emitter-output-dir"];
+    Logger.info(`Resolved emitter-output-dir for ${emitterData.emitter}: ${emitterOutputDir}`);
+    newPackageDir = resolve(emitterOutputDir);
```
JoshLove-msft, line 113: Should we log newPackageDir as this will be the resolved path?

tools/tsp-client/src/commands.ts
```diff
@@ -370,6 +403,19 @@ export async function generateCommand(argv: any) {
...
+  let tspConfigData;
+  try {
+    tspConfigData = parseYaml(await readFile(tspConfigPath, "utf8"));
+  } catch (err) {
+    Logger.warn(
+      `Could not read tspconfig.yaml at ${tspConfigPath}. Will use the repo root as the target output directory during compilation, this may cause unexpected behavior. Error: ${err}`,
+    );
+  }
+
+  const legacyPathResolution = !tspConfigData?.options?.[emitter]?.["emitter-output-dir"];
```
JoshLove-msft, line 418: Do we need the flag for this? Can we just have the fallback implicitly?

tools/tsp-client/test/examples/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml
catalinaperalta: Configuring an output-dir flag is supported by both tsp-client and the tsp cli. The output-dir variable in tspconfig.yaml gets resolved when we call the resolveCompilerOptions method from the typespec/compiler library. The behavior for using an output-dir is unchanged for both init and update commands, so folks are still expected to run the commands from the appropriate directory to get the expected output path resolved, or alternately pass in a custom output dir value which would prefix the rest of the path, similar behavior for new and legacy path resolution strategies here, with the exception that with the new strategy the typespec/compiler handles that for us :)

tools/tsp-client/src/commands.ts
catalinaperalta: Sure, I'll move the log message lower in the function until after we resolve it

tools/tsp-client/src/commands.ts
catalinaperalta: This flag just lets us know in the compileTsp function that we need to maintain legacy behavior when we compile. The check would still have to happen though since in the legacy scenario we need to explicitly specify the emitter-output-dir and that doesnt get resolved until we process the override options before calling compile

jsquire (approved):

Issue #11332, opened 2025-07-24
[azsdk tsp] Re-add `tsp init` subcommand to azsdk cli


chrisradek:
Revert this commit once TypeSpec 1.3.0 is released:
https://github.com/Azure/azure-sdk-tools/pull/11281/commits/411b7c4254dc58e36faca8cfc8182a011b257d39

The next TypeSpec release will support calling `tsp init` with the additional CLI flags the MCP tool needs.

PR #11628, opened 2025-08-11
[azsdk cli] add tsp init tool

chrisradek:
Fixes https://github.com/Azure/azure-sdk-tools/issues/11332

This adds the `tsp init` subcommand to the azsdk cli. This is exposed both at the CLI and MCP layers.

It ends up calling the typespec compiler's `tsp init` tool under the hood.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspConvertTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspConvertTool.cs
@@ -71,19 +71,26 @@ namespace Azure.Sdk.Tools.Cli.Tools.TypeSpec
             output.Output(result);
         }
 
-        [McpServerTool(Name = "azsdk_convert_swagger_to_typespec"), Description(@"Converts an existing Azure service swagger definition to a TypeSpec project.
-        Pass in the `pathToSwaggerReadme` which is the path to the swagger README file.
-        Pass in the `outputDirectory` where the TypeSpec project should be created. This must be an existing empty directory.
-        Pass in `isAzureResourceManagement` to indicate whether the swagger is for an Azure Resource Management (ARM) API.
-        This should be true if the swagger's path contains `resource-manager`.
-        Pass in `fullyCompatible` to indicate whether the generated TypeSpec project should be fully compatible with the swagger.
-        It is recommended not to set this to `true` so that the converted TypeSpec project
-        leverages TypeSpec built-in libraries with standard patterns and templates.
-        Returns path to the created project.")]
+        [
+            McpServerTool(Name = "azsdk_convert_swagger_to_typespec"),
+            Description("Converts an existing Azure service swagger definition to a TypeSpec project. Returns path to the created project.")
+        ]
         public async Task<TspToolResponse> ConvertSwaggerAsync(
+            [Description("Path to the swagger README file.")]
             string pathToSwaggerReadme,
+            [Description("The output directory for the generated TypeSpec project. This must be an existing empty directory.")]
             string outputDirectory,
+            [Description(@"
+                Indicates whether the swagger is for an Azure Resource Management (ARM) API.
+                Should be true if the swagger's path contains `resource-manager`.
+                ")
+            ]
             bool? isAzureResourceManagement,
+            [Description(@"
+                Indicates whether the generated TypeSpec project should be fully compatible with the swagger.
+                It is recommended to set this to `false` so that the generated project leverages TypeSpec built-in libraries with standard patterns and templates.
+                ")
+            ]
             bool? fullyCompatible,
             bool isCli,
             CancellationToken ct


tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/TspToolTests.cs
```diff
@@ -63,5 +69,57 @@ public void ConvertSwagger_WithNonExistentFile_ShouldReturnError()
...
+
+        [Test]
+        public void Init_WithInvalidServiceNamespace_ShouldReturnError()
+        {
+            var npxHelper = new Mock<INpxHelper>().Object;
+            var logger = new Mock<ILogger<TypeSpecTool>>().Object;
+            var outputService = new Mock<IOutputService>().Object;
+            var tool = new TypeSpecTool(npxHelper, logger, outputService);
+
+            var result = tool.InitTypeSpecProject(outputDirectory: @"C:\temp", template: "azure-core", serviceNamespace: "");
```
richardpark-msft:
```suggestion
            var result = tool.InitTypeSpecProject(outputDirectory: @"not-used-at-all", template: "azure-core", serviceNamespace: "");
```

richardpark-msft: I like using this pattern to make it clear to readers that the parameter isn't used at all for the test. Otherwise it's kind of weird to have a parameter that's a Windows path when it should run everywhere.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/TspToolTests.cs
```diff
@@ -63,5 +69,57 @@ public void ConvertSwagger_WithNonExistentFile_ShouldReturnError()
...
+
+        [Test]
+        public void Init_WithNonExistentDirectory_ShouldReturnError()
+        {
+            var npxHelper = new Mock<INpxHelper>().Object;
+            var logger = new Mock<ILogger<TypeSpecTool>>().Object;
+            var outputService = new Mock<IOutputService>().Object;
+            var tool = new TypeSpecTool(npxHelper, logger, outputService);
+
+            var result = tool.InitTypeSpecProject(outputDirectory: @"C:\nonexistent", template: "azure-core", serviceNamespace: "MyService");
```
richardpark-msft: Windows vs Linux paths?

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/TspToolTests.cs
```diff
@@ -63,5 +69,57 @@ public void ConvertSwagger_WithNonExistentFile_ShouldReturnError()
...
+
+            Assert.Multiple(() =>
+            {
+                Assert.That(result.IsSuccessful, Is.False);
+                Assert.That(result.ResponseError, Does.Contain("Invalid --output-directory"));
```
richardpark-msft: If these errors are just coming from `tsp` I don't think we need to test them here.

Specifically, you're just testing your integration calling into `tsp`, and not `tsp` itself.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -133,6 +161,64 @@ public TspToolResponse ConvertSwagger(string pathToSwaggerReadme, string outputD
...
+        public TspToolResponse InitTypeSpecProject(string outputDirectory, string template, string serviceNamespace)
+        {
+            try
+            {
+                logger.LogInformation("Initializing TypeSpec project: {outputDirectory}, template: {template}, serviceNamespace: {serviceNamespace}",
+                    outputDirectory, template, serviceNamespace);
+
+                // Validate template
+                var validTemplates = new[] { "azure-core", "azure-arm" };
+                if (string.IsNullOrWhiteSpace(template) || !validTemplates.Contains(template))
```
richardpark-msft: There's some shorthand for these kinds of checks:

[Argument.](https://github.com/Azure/azure-sdk-tools/blob/6135bcb69219964de571be1630aec34153115400/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Generators/ReadMeGeneratorTool.cs#L144)

Then you could just centralize our exception handling in the caller to this function, and shorten it a bit.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/TspToolTests.cs
chrisradek: This is actually checking that the azsdk tool throws a validation error - not that the underlying `tsp` cli throws that error - if that makes sense?

richardpark-msft: On that note does the error message that comes back actually just mention that the directory doesn't exist?

richardpark-msft: With enough info I think GHCP could solve it.

chrisradek: Yeah - the full error from a test run would be:
> Failed: Invalid --output-directory, Directory '/tmp/test-nonexistent-382c38a3-776b-4131-8212-e143a257af84' does not exist.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -19,11 +19,21 @@ namespace Azure.Sdk.Tools.Cli.Tools
...
         // commands
         private const string ConvertSwaggerCommandName = "convert-swagger";
+        private const string InitCommandName = "init";
 
+        // Shared command options
         private readonly Option<string> outputDirectoryArg = new("--output-directory", "The output directory for the generated TypeSpec project. This directory must already exist and be empty.") { IsRequired = true };
 
+        // Init command options
+        private readonly Option<string> templateArg = new("--template", "The template to use for the TypeSpec project. Valid values are: azure-core (for data-plane services), azure-arm (for resource-manager services)") { IsRequired = true };
+        private readonly Option<string> serviceNamespaceArg = new("--service-namespace", "The namespace of the service you are creating. This should be in Pascal case and represent the service's namespace.") { IsRequired = true };
```
benbp: I think you can add a validation function via System.CommandLine if you want to fail fast at invocation time.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -19,11 +19,21 @@ namespace Azure.Sdk.Tools.Cli.Tools
...
+
         // commands
         private const string ConvertSwaggerCommandName = "convert-swagger";
+        private const string InitCommandName = "init";
 
+        // Shared command options
         private readonly Option<string> outputDirectoryArg = new("--output-directory", "The output directory for the generated TypeSpec project. This directory must already exist and be empty.") { IsRequired = true };
 
+        // Init command options
+        private readonly Option<string> templateArg = new("--template", "The template to use for the TypeSpec project. Valid values are: azure-core (for data-plane services), azure-arm (for resource-manager services)") { IsRequired = true };
```
benbp: Using an enum type here instead of `string` would validate inclusion within the limited options.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
benbp: @richardpark-msft we can just use an enum type for the cli option.

benbp: I guess this doesn't help with mcp mode :( though the mcp client will always try to fill in values for these if they aren't optional parameters.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -79,6 +97,17 @@ private void HandleConvertCommand(InvocationContext ctx, CancellationToken ct)
             output.Output(result);
         }
 
+        private void HandleInitCommand(InvocationContext ctx, CancellationToken ct)
+        {
+            var outputDirectory = ctx.ParseResult.GetValueForOption(outputDirectoryArg);
+            var template = ctx.ParseResult.GetValueForOption(templateArg);
+            var serviceNamespace = ctx.ParseResult.GetValueForOption(serviceNamespaceArg);
+
+            TspToolResponse result = InitTypeSpecProject(outputDirectory: outputDirectory, template: template, serviceNamespace: serviceNamespace);
```
benbp: I think it's useful to pass the cancellation token down so it eventually gets used by the process helper and we can ctrl-c.

benbp: Then if we need to set global timeouts later on it's also supported.

benbp:
Looks pretty good just a couple minor comments.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Yeah Richard and I were just discussing this yesterday. I put validation in the MCP path to be as helpful as possible to the calling AI agent. I could _also_ put more on the CLI path, but didn't put too much there since the CLI path calls the MCP method anyway.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: @benbp I updated process helper and npx helper to be async so that they accept cancellation tokens now.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Updated to an enum, though I'm a little sad I still have to use a string with the choices listed out for the MCP path 😢

AlitzelMendez (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
benbp: @chrisradek ah my bad I should have mentioned I have a PR up for process helper async (https://github.com/Azure/azure-sdk-tools/pull/11614) but I still have some work to do. I forgot that it's still sync in main.

chrisradek: Oh cool! Is that close to getting merged? I can back this change out or just wait for yours to get merged and resolve conflicts here, depending on timelines.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
benbp: Yeah I guess I didn't realize the enum custom string shortcomings with c#. Does this mean we can't pass `--template azure-core` and it has to be `--template azurecore` now?

chrisradek: That's right - no dashes going the enum route

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
benbp: I'm aiming to get it in tomorrow. Otherwise I'm happy to just try to rebase against your changes, don't want to block you.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
benbp: Ahh, in that case maybe let's not do enums. Sorry to send you down this path.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -88,7 +167,7 @@ Pass in `fullyCompatible` to indicate whether the generated TypeSpec project sho
         It is recommended not to set this to `true` so that the converted TypeSpec project
         leverages TypeSpec built-in libraries with standard patterns and templates.
         Returns path to the created project.")]
-        public TspToolResponse ConvertSwagger(string pathToSwaggerReadme, string outputDirectory, bool? isAzureResourceManagement = null, bool? fullyCompatible = null)
+        public async Task<TspToolResponse> ConvertSwagger(string pathToSwaggerReadme, string outputDirectory, bool? isAzureResourceManagement = null, bool? fullyCompatible = null, CancellationToken ct = default)
```
praveenkuttappan:
```suggestion
        public async Task<TspToolResponse> ConvertSwaggerAsync(string pathToSwaggerReadme, string outputDirectory, bool? isAzureResourceManagement = null, bool? fullyCompatible = null, CancellationToken ct = default)
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -133,6 +212,68 @@ public TspToolResponse ConvertSwagger(string pathToSwaggerReadme, string outputD
             }
         }
 
+        [McpServerTool(Name = "azsdk_init_typespec_project"), Description("Use this tool to initialize a new TypeSpec project. Returns the path to the created project.")]
+        public async Task<TspToolResponse> InitTypeSpecProject(
```
praveenkuttappan:
```suggestion
        public async Task<TspToolResponse> InitTypeSpecProjectAsync(
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -133,6 +212,68 @@ public TspToolResponse ConvertSwagger(string pathToSwaggerReadme, string outputD
...
+            [Description("Pass in the output directory where the project should be created. Must be an existing empty directory.")]
+            string outputDirectory,
+            [Description("`azure-core` for data-plane services, or `azure-arm` for resource-manager services.")]
+            string template,
```
praveenkuttappan: One another option is to take a param whether it's mgmt or data and then use the template based on the type?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -133,6 +212,68 @@ public TspToolResponse ConvertSwagger(string pathToSwaggerReadme, string outputD
...
+                {
+                    SetFailure();
+                    return new TspToolResponse
+                    {
+                        ResponseError = $"Failed: Invalid --template, '{template}'. Must be one of: {string.Join(", ", validTemplates)}."
+                    };
+                }
+
+                // Validate service namespace
+                if (string.IsNullOrWhiteSpace(serviceNamespace))
```
praveenkuttappan: Description says "Microsoft." prefix should be removed from service name when tool is called. Is it better to enforce it and remove it ourselves if it is present in the value.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Do you mean just renaming these values? Or changing to a boolean like `isArm` and deciding based on that?

praveenkuttappan: Changing to bool and decide based on that. But if you intention is to use any other project as a template in the future then let's go with current approach.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
```diff
@@ -11,25 +11,76 @@
...
+
+
     /// <summary>
     /// This tool provides functionality for initializing TypeSpec projects and converting existing Azure service swagger definitions to TypeSpec projects.
     /// Use this tool to onboard to TypeSpec for new services or convert existing services.
     /// </summary>
     [McpServerToolType, Description("Tools for initializing TypeSpec projects and converting existing Azure service swagger definitions to TypeSpec projects.")]
     public class TypeSpecTool(INpxHelper npxHelper, ILogger<TypeSpecTool> logger, IOutputService output) : MCPTool
     {
+        /// <summary>
```
praveenkuttappan: You will need to rebase with latest and move this class into TypeSpec namespace. I would also like to see this new tool added in its own MCPTool class. We are trying to get to a stage one tool per class to enforce some cli hierarchy and as part of that work, we will split current class with multiple tools into its own class per tool call.

richardpark-msft: @jsquire will be happy to read this ;)

jsquire: Yay!

chrisradek: Thanks - was waiting on the processhelper changes to rebase - see those are in now! Will split this into its own class.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Switched back to not enums

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Yes - I do imagine adding other templates in the future as well.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Now removing prefix for ARM specs

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TspTool/TspTool.cs
chrisradek: Updated!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspInitTool.cs
```diff
@@ -0,0 +1,226 @@
...
+using Azure.Sdk.Tools.Cli.Services;
+using Azure.Sdk.Tools.Cli.Contract;
+using Azure.Sdk.Tools.Cli.Models.Responses;
+using ModelContextProtocol.Server;
+using Azure.Sdk.Tools.Cli.Helpers;
+using Azure.Sdk.Tools.Cli.Commands;
+
+namespace Azure.Sdk.Tools.Cli.Tools.TypeSpec
+{
+
```
richardpark-msft: This excessive whitespace is just throwing away money.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedCommandGroups.cs
```diff
@@ -41,12 +41,18 @@ public static class SharedCommandGroups
             Options: []
         );
 
-        #if DEBUG
+        public static readonly CommandGroup TypeSpec = new(
+            Verb: "tsp",
+            Description: "Tools for setting up TypeSpec projects",
```
richardpark-msft:
```suggestion
            Description: "Tools for setting up or working with TypeSpec projects",
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspConvertTool.cs
```diff
@@ -8,47 +8,53 @@
 using Azure.Sdk.Tools.Cli.Models.Responses;
 using ModelContextProtocol.Server;
 using Azure.Sdk.Tools.Cli.Helpers;
+using Azure.Sdk.Tools.Cli.Commands;
 
 namespace Azure.Sdk.Tools.Cli.Tools.TypeSpec
 {
+
```
richardpark-msft: Seriously, do you think we're made of newlines here?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspConvertTool.cs
```diff
@@ -8,47 +8,53 @@
...
-    /// This tool provides functionality for initializing TypeSpec projects and converting existing Azure service swagger definitions to TypeSpec projects.
-    /// Use this tool to onboard to TypeSpec for new services or convert existing services.
+    /// This tool provides functionality for converting existing Azure service swagger definitions to TypeSpec projects.
+    /// Use this tool to convert existing services to TypeSpec.
     /// </summary>
-    [McpServerToolType, Description("Tools for initializing TypeSpec projects and converting existing Azure service swagger definitions to TypeSpec projects.")]
-    public class TypeSpecTool(INpxHelper npxHelper, ILogger<TypeSpecTool> logger, IOutputHelper output) : MCPTool
+    [McpServerToolType, Description("Tools for converting existing Azure service swagger definitions to TypeSpec projects.")]
+    public class TypeSpecConvertTool : MCPTool
     {
```
richardpark-msft: I'm going to take away your ENTER key

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspConvertTool.cs
```diff
@@ -8,47 +8,53 @@
...
+        // command options
         private readonly Option<string> outputDirectoryArg = new("--output-directory", "The output directory for the generated TypeSpec project. This directory must already exist and be empty.") { IsRequired = true };
-
         private readonly Option<string> swaggerReadmeArg = new("--swagger-readme", "The path or URL to an Azure swagger README file.") { IsRequired = true };
         private readonly Option<bool> isArmOption = new("--arm", "Whether the generated TypeSpec project is for an Azure Resource Management (ARM) API. This should be true if the swagger's path contains 'resource-manager'.");
         private readonly Option<bool> fullyCompatibleOption = new("--fully-compatible", "Whether to generate a TypeSpec project that is fully compatible with the swagger. It is recommended not to set this to true so that the converted TypeSpec project leverages TypeSpec built-in libraries with standard patterns and templates.");
 
         public override Command GetCommand()
         {
-            var tspCommand = new Command("tsp", "Tools for initializing TypeSpec projects and converting existing Azure service swagger definitions to TypeSpec projects");
```
richardpark-msft, line 45: Well, well, well.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspConvertTool.cs
```diff
@@ -87,7 +93,7 @@ Pass in `fullyCompatible` to indicate whether the generated TypeSpec project sho
         It is recommended not to set this to `true` so that the converted TypeSpec project
         leverages TypeSpec built-in libraries with standard patterns and templates.
         Returns path to the created project.")]
-        public async Task<TspToolResponse> ConvertSwagger(
+        public async Task<TspToolResponse> ConvertSwaggerAsync(
```
richardpark-msft, line 78: Hey, a much better programmer already did this much better:

https://github.com/chrisradek/azure-sdk-tools/blob/b899390e4fae1bb92a255dd184560f739d7e515b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspInitTool.cs#L118

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspConvertTool.cs
```diff
@@ -162,7 +168,7 @@ CancellationToken ct
             return null; // Validation passed
```
richardpark-msft, line 161: I can't leave a comment on the right spot, but above this you've just started your newline additions again. AGAIN! :)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspInitTool.cs
```diff
@@ -0,0 +1,226 @@
...
+                serviceNamespaceArg
+            };
+            command.SetHandler(async ctx => { await HandleCommand(ctx, ctx.GetCancellationToken()); });
+
+            return command;
+        }
+
+        public override async Task HandleCommand(InvocationContext ctx, CancellationToken ct)
+        {
+            var command = ctx.ParseResult.CommandResult.Command.Name;
```
richardpark-msft: I think you can just remove this boilerplate, and call into `HandleInitCommandAsync` directly, since there's only one command implemented here.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspConvertTool.cs
```diff
@@ -59,17 +56,7 @@ public override Command GetCommand()
...
-            {
-                case ConvertSwaggerCommandName:
-                    await HandleConvertCommandAsync(ctx, ct);
-                    return;
-                default:
-                    SetFailure();
-                    output.Output($"Unknown command: {command}");
-                    return;
-            }
+            await HandleConvertCommandAsync(ctx, ct);
```
richardpark-msft, line 59: NIT: and now you don't even need this sub-function anymore.

PR #11629, opened 2025-08-11
[API View]  Add authentication policies

AlitzelMendez:
## Authentication workflow
This PR enables GitHub token authentication for APIView and creates different authorization policies to establish authentication patterns for future endpoints.

### Authorization Profiles: 
- RequireCookieAuthentication - Cookie (browser) only for GitHub OAuth users
- RequireTokenAuthentication - Token-only supporting GitHub tokens and Managed Identity
- RequireTokenOrCookieAuthentication - Hybrid authentication accepting cookies or tokens

### Changes
- Added GitHub token support with API validation
- Added HybridApiController supporting RequireTokenOrCookieAuthentication
- RequireCookieAuthentication remains the default behavior
- New authentication policies for flexible endpoint authorization

### Backward Compatibility
- Existing cookie-based authentication unchanged

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Account/AuthenticationValidator.cs
+++ b/src/dotnet/APIView/APIViewWeb/Account/AuthenticationValidator.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Net.Http;
 using System.Security.Claims;
@@ -7,21 +8,22 @@ using System.Text.Json;
 using System.Threading.Tasks;
 using APIView.Identity;
 using APIViewWeb.Models;
+using Microsoft.Extensions.Logging;
 
 namespace APIViewWeb.Account
 {
     public static class AuthenticationValidator
     {
-        #region Managed Identity Validation
+        #region Azure Identity Validation
 
-        public static bool IsValidManagedIdentity(ClaimsPrincipal user)
+        public static bool IsValidAzureIdentity(ClaimsPrincipal user)
         {
             if (!IsAuthenticated(user))
             {
                 return false;
             }
 
-            return HasRequiredManagedIdentityClaims(user) && IsValidAzureAuthentication(user);
+            return HasRequiredAzureIdentityClaims(user) && IsValidAzureAuthentication(user);
         }
 
         private static bool IsValidAzureAuthentication(ClaimsPrincipal user)
@@ -29,7 +31,7 @@ namespace APIViewWeb.Account
             return HasValidAuthenticationType(user) || HasValidIdentityProvider(user);
         }
 
-        private static bool HasRequiredManagedIdentityClaims(ClaimsPrincipal user)
+        private static bool HasRequiredAzureIdentityClaims(ClaimsPrincipal user)
         {
             Claim oidClaim = user.FindFirst("oid") ?? user.FindFirst("http://schemas.microsoft.com/identity/claims/objectidentifier");
             return oidClaim != null;
@@ -83,19 +85,19 @@ namespace APIViewWeb.Account
 
         #region Combined Authentication Methods
         
-        public static bool HasOrganizationOrManagedIdentityAccess(ClaimsPrincipal user, string[] requiredOrganizations)
+        public static bool HasOrganizationOrAzureAuthenticationAccess(ClaimsPrincipal user, string[] requiredOrganizations)
         {
             if (!IsAuthenticated(user))
             {
                 return false;
             }
 
-            return IsValidManagedIdentity(user) || HasOrganizationAccess(user, requiredOrganizations);
+            return IsValidAzureIdentity(user) || HasOrganizationAccess(user, requiredOrganizations);
         }
         
         public static string GetAuthenticationMethod(ClaimsPrincipal user)
         {
-            if (IsValidManagedIdentity(user))
+            if (IsValidAzureIdentity(user))
             {
                 return "ManagedIdentity";
             }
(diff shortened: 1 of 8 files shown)


tjprescott:
So recapping from our discussion, we want three kinds of auth:

- CookieOnly (GH cookie) - this is for endpoints that only support the website internals
- CookieOrToken (GH cookie or GH/AD JWT) - for endpoints used by the website but also useful externally (like get comments)
- TokenOnly (GH/AD JWT) - for endpoints that are only useful externally

Separately, the token support should also include GH JWT (currently I believe it is just AD)

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsController.cs
```diff
@@ -148,7 +135,7 @@ public async Task<ActionResult> CreateCommentAsync(
                 await _commentsManager.RequestAgentReply(User, comment, apiRevisionId);
             }
 
-            return new LeanJsonResult(comment, StatusCodes.Status201Created, Url.Action("GetComments", new { reviewId = reviewId }));
+            return new LeanJsonResult(comment, StatusCodes.Status201Created, Url.Link("GetComments", new { reviewId = reviewId }));
```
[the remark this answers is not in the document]
AlitzelMendez: I validated this but I will double check

src/dotnet/APIView/APIViewWeb/Account/AuthenticationValidator.cs
```diff
@@ -104,6 +109,94 @@ public static string GetAuthenticationMethod(ClaimsPrincipal user)
...
+            httpClient.DefaultRequestHeaders.Clear();
+            httpClient.DefaultRequestHeaders.Authorization = new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", token);
+            httpClient.DefaultRequestHeaders.Add("User-Agent", "APIView-GitHub-Token-Validator");
+            httpClient.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json");
+        }
+
+        private static async Task<GithubUser> GetGitHubUserAsync(HttpClient httpClient)
+        {
+            var response = await httpClient.GetAsync("https://api.github.com/user");
+            if (!response.IsSuccessStatusCode) { return null; }
```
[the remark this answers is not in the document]
tjprescott: This copilot comment makes sense to me. Do we have existing logging in APIView? If so, is there a reason not to hook into it on failure?

src/dotnet/APIView/APIViewWeb/Account/AuthenticationValidator.cs
```diff
@@ -104,6 +109,94 @@ public static string GetAuthenticationMethod(ClaimsPrincipal user)
...
+
+            string json = await response.Content.ReadAsStringAsync();
+            return JsonSerializer.Deserialize<GithubUser>(json, jsonOptions);
+        }
+
+        private static async Task<string[]> GetUserOrganizationsAsync(HttpClient httpClient)
+        {
+            var response = await httpClient.GetAsync("https://api.github.com/user/orgs");
+            if (!response.IsSuccessStatusCode) { return []; }
```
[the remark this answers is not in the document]
tjprescott: Same here.

src/dotnet/APIView/APIViewWeb/Account/ConfigureTokenAuthenticationPolicy.cs
```diff
@@ -0,0 +1,37 @@
...
+            policy.RequireAssertion(context =>
+            {
+                if (context.User.Identity?.AuthenticationType == "Cookies")
+                {
+                    return false;
+                }
+
+                return AuthenticationValidator.HasOrganizationOrManagedIdentityAccess(context.User,
+                    _options.Value.RequiredOrganization);
+            });
```
tjprescott: Just in terms of the name, I don't think a user principal (AD token) is a ManagedIdentity--that's a separate thing. If this is really saying "HasOrganizationOrTokenAccess" then maybe we rename it.

src/dotnet/APIView/APIViewWeb/Controllers/AuthTestController.cs
```diff
@@ -42,6 +42,41 @@ public IActionResult TestAuth()
         });
     }
 
+    /// <summary>
+    ///     Test endpoint that only allows cookie-based authentication (GitHub OAuth via browser)
+    /// </summary>
+    [HttpGet("cookie-only")]
+    [Authorize("RequireCookieAuthentication")]
+    public IActionResult TestCookieOnly()
```
tjprescott, line 50: It would probably be good as well to verify that "RequireCookieAuthentication" fails with a token, and vice versa. Also a test that "RequiredCookieOrToken" works with either.

src/dotnet/APIView/APIViewWeb/Account/ConfigureTokenAuthenticationPolicy.cs
AlitzelMendez: let me think in another name, is not checking `HasOrganizationOrTokenAccess` because the github token should be in the organization, but the AD token should be just a valida AD token

tjprescott: ManagedIdentity uses AD tokens, but when _you_ personally log in that is also an AD token that is _not_ ManagedIdentity. That's UserPrincipal identity. Hence, this makes it sound like it would work for managed identities only, which I suspect is not the case.

tjprescott (approved):

PR #11635, opened 2025-08-12
added upload file service to the Agent

radhgupta:
This PR introduces an unified access to TypeSpec files from both local directories and GitHub repositories, with significant performance optimizations for GitHub operations.

- TypeSpecFileService: Unified interface that automatically routes to local or GitHub sources based on context
- GitHubFilesRetreiveService: Specialized service for fast GitHub file retrieval using REST API

### DIFF
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
@@ -48,22 +48,7 @@ namespace Azure.Tools.GeneratorAgent
                 hasAuthHeader ? "Configured" : "Not configured - using rate-limited access");
         }
 
-        public async Task<Result<Dictionary<string, string>>> GetTypeSpecFilesAsync(CancellationToken cancellationToken = default)
-        {
-            try
-            {
-                Result<Dictionary<string, string>> apiResult = await FetchTypeSpecFilesViaApiAsync(cancellationToken).ConfigureAwait(false);
-                return apiResult;
-            }
-            catch (Exception ex) when (ex is not OperationCanceledException)
-            {
-                Logger.LogCritical(ex, "Unexpected error while fetching TypeSpec files from GitHub repository {Repository}", 
-                    AppSettings.AzureSpecRepository);
-                throw;
-            }
-        }
-
-        private async Task<Result<Dictionary<string, string>>> FetchTypeSpecFilesViaApiAsync(CancellationToken cancellationToken)
+        public async Task<Dictionary<string, string>> GetTypeSpecFilesAsync(CancellationToken cancellationToken = default)
         {
             try
             {
@@ -84,7 +69,11 @@ namespace Azure.Tools.GeneratorAgent
 
                 if (contents == null)
                 {
-                    throw new InvalidOperationException("Failed to deserialize GitHub API response");
+                    Logger.LogCritical("Failed to deserialize GitHub API response from {ApiUrl}. " +
+                        "Response content: {JsonContent}", apiUrl, jsonContent);
+                    throw new InvalidOperationException($"Failed to deserialize GitHub API response for " +
+                        $"'{AppSettings.AzureSpecRepository}/{TypespecSpecDir}' at commit '{CommitId}'. " +
+                        $"Please verify the repository path and commit ID are correct.");
                 }
 
                 Dictionary<string, string> typeSpecFiles = new(contents.Length);
@@ -96,12 +85,6 @@ namespace Azure.Tools.GeneratorAgent
 
                 IEnumerable<Task<(string FileName, string Content)>> downloadTasks = tspFiles
                     .Select(tspFile => DownloadFileContentAsync(tspFile.Name, tspFile.DownloadUrl, cancellationToken));
-
-                if (!downloadTasks.Any())
-                {
-                    Logger.LogWarning("No valid .tsp files found with download URLs");
-                    throw new InvalidOperationException("No valid .tsp files found with download URLs");
-                }
 
                 (string FileName, string Content)[] downloadedFiles = await Task.WhenAll(downloadTasks).ConfigureAwait(false);
 
@@ -118,7 +101,7 @@ namespace Azure.Tools.GeneratorAgent
                     }
                 }
 
-                return Result<Dictionary<string, string>>.Success(typeSpecFiles);
+                return typeSpecFiles;
             }
             catch (Exception ex)
             {
@@ -131,28 +114,18 @@ namespace Azure.Tools.GeneratorAgent
         {
             if (string.IsNullOrEmpty(downloadUrl))
             {
-                Logger.LogWarning("Download URL is null or empty for file {FileName}", fileName);
-                return (fileName, string.Empty);
+                throw new InvalidOperationException($"Download URL is null or empty for file {fileName}");
             }
 
-            try
-            {
-                HttpResponseMessage response = await HttpClient.GetAsync(downloadUrl, cancellationToken).ConfigureAwait(false);
-                if (response.IsSuccessStatusCode)
-                {
-                    string content = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
-                    return (fileName, content);
-                }
+            HttpResponseMessage response = await HttpClient.GetAsync(downloadUrl, cancellationToken).ConfigureAwait(false);
 
-                Logger.LogWarning("Failed to download file {FileName} from {Url}: {StatusCode} {ReasonPhrase}", 
-                    fileName, downloadUrl, response.StatusCode, response.ReasonPhrase);
-                return (fileName, string.Empty);
-            }
-            catch (Exception ex) when (ex is not OperationCanceledException)
+            if (!response.IsSuccessStatusCode)
             {
-                Logger.LogCritical(ex, "Error downloading file {FileName} from {Url}", fileName, downloadUrl);
-                return (fileName, string.Empty);
+                throw new HttpRequestException($"Failed to download file {fileName}: {response.StatusCode} {response.ReasonPhrase}");
             }
+
+            string content = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
+            return (fileName, content);
         }
     }
 }
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Program.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Program.cs
@@ -93,16 +93,7 @@ namespace Azure.Tools.GeneratorAgent
                 var fileServiceFactory = ServiceProvider.GetRequiredService<Func<ValidationContext, TypeSpecFileService>>();
                 TypeSpecFileService fileService = fileServiceFactory(validationContext);
 
-                Result<Dictionary<string, string>> fileResult = await fileService.GetTypeSpecFilesAsync(cancellationToken);
-
-                if (fileResult.IsFailure)
-                {
-                    Logger.LogError("Failed to retrieve TypeSpec files: {Error}", 
-                        fileResult.Exception?.Message ?? fileResult.ProcessException?.Message ?? "Unknown error");
-                    return ExitCodeFailure;
-                }
-
-                Dictionary<string, string> typeSpecFiles = fileResult.Value!;
+                Dictionary<string, string> typeSpecFiles = await fileService.GetTypeSpecFilesAsync(cancellationToken);
 
                 // Step 3: Initialize ErrorFixer Agent with files in memory (singleton)
                 ErrorFixerAgent agent = ServiceProvider.GetRequiredService<ErrorFixerAgent>();
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecFileService.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecFileService.cs
@@ -39,7 +39,7 @@ namespace Azure.Tools.GeneratorAgent
         /// Gets TypeSpec files from either local directory or GitHub repository.
         /// Returns a dictionary where key is filename and value is file content.
         /// </summary>
-        public async Task<Result<Dictionary<string, string>>> GetTypeSpecFilesAsync(
+        public async Task<Dictionary<string, string>> GetTypeSpecFilesAsync(
             CancellationToken cancellationToken = default)
         {
             if (string.IsNullOrWhiteSpace(ValidationContext.ValidatedCommitId))
@@ -54,7 +54,7 @@ namespace Azure.Tools.GeneratorAgent
             }
         }
 
-        private async Task<Result<Dictionary<string, string>>> GetLocalTypeSpecFilesAsync(
+        private async Task<Dictionary<string, string>> GetLocalTypeSpecFilesAsync(
             string typeSpecDir, 
             CancellationToken cancellationToken)
         {
@@ -75,7 +75,7 @@ namespace Azure.Tools.GeneratorAgent
                 }
 
                 Logger.LogInformation("Successfully read {Count} TypeSpec files from local directory", typeSpecFiles.Count);
-                return Result<Dictionary<string, string>>.Success(typeSpecFiles);
+                return typeSpecFiles;
             }
             catch (Exception ex) when (ex is not OperationCanceledException)
             {
@@ -84,7 +84,7 @@ namespace Azure.Tools.GeneratorAgent
             }
         }
 
-        private async Task<Result<Dictionary<string, string>>> GetGitHubTypeSpecFilesAsync(
+        private async Task<Dictionary<string, string>> GetGitHubTypeSpecFilesAsync(
             CancellationToken cancellationToken)
         {
             Logger.LogInformation("Fetching TypeSpec files from GitHub: {TypeSpecDir} at commit {CommitId}", 
@@ -94,12 +94,9 @@ namespace Azure.Tools.GeneratorAgent
             {
                 GitHubService = GitHubServiceFactory(ValidationContext);
 
-                Result<Dictionary<string, string>> result = await GitHubService.GetTypeSpecFilesAsync(cancellationToken);
+                Dictionary<string, string> result = await GitHubService.GetTypeSpecFilesAsync(cancellationToken);
 
-                if (result.IsSuccess)
-                {
-                    Logger.LogInformation("Successfully fetched {Count} TypeSpec files from GitHub", result.Value!.Count);
-                }
+                Logger.LogInformation("Successfully fetched {Count} TypeSpec files from GitHub", result.Count);
 
                 return result;
             }
--- a/src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/GitHubFileServiceTests.cs
+++ b/src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/GitHubFileServiceTests.cs
@@ -361,10 +361,9 @@ namespace Azure.Tools.GeneratorAgent.Tests
             // Assert
             Assert.Multiple(() =>
             {
-                Assert.That(result.IsSuccess, Is.True);
-                Assert.That(result.Value, Is.Not.Null);
-                Assert.That(result.Value!.Count, Is.EqualTo(1));
-                Assert.That(result.Value["test.tsp"], Is.EqualTo("model Test {}"));
+                Assert.That(result, Is.Not.Null);
+                Assert.That(result.Count, Is.EqualTo(1));
+                Assert.That(result["test.tsp"], Is.EqualTo("model Test {}"));
             });
 
 
@@ -407,7 +406,7 @@ namespace Azure.Tools.GeneratorAgent.Tests
         }
 
         [Test]
-        public void GetTypeSpecFilesAsync_WithNoTspFiles_ShouldThrowInvalidOperationException()
+        public async Task GetTypeSpecFilesAsync_WithNoTspFiles_ShouldReturnEmptyDictionary()
         {
             // Arrange
             var mockHandler = CreateMockHttpMessageHandler();
@@ -416,12 +415,15 @@ namespace Azure.Tools.GeneratorAgent.Tests
 
             SetupEmptyApiResponse(mockHandler);
 
-            // Act & Assert
-            var exception = Assert.ThrowsAsync<InvalidOperationException>(async () => 
-                await service.GetTypeSpecFilesAsync());
-            Assert.That(exception!.Message, Does.Contain("No valid .tsp files found with download URLs"));
+            // Act
+            var result = await service.GetTypeSpecFilesAsync();
 
-
+            // Assert
+            Assert.Multiple(() =>
+            {
+                Assert.That(result, Is.Not.Null);
+                Assert.That(result.Count, Is.EqualTo(0));
+            });
         }
 
         [Test]
@@ -442,10 +444,10 @@ namespace Azure.Tools.GeneratorAgent.Tests
             // Assert
             Assert.Multiple(() =>
             {
-                Assert.That(result.IsSuccess, Is.True);
-                Assert.That(result.Value!.Count, Is.EqualTo(2));
-                Assert.That(result.Value.ContainsKey("file1.tsp"), Is.True);
-                Assert.That(result.Value.ContainsKey("file2.tsp"), Is.True);
+                Assert.That(result, Is.Not.Null);
+                Assert.That(result.Count, Is.EqualTo(2));
+                Assert.That(result.ContainsKey("file1.tsp"), Is.True);
+                Assert.That(result.ContainsKey("file2.tsp"), Is.True);
             });
 
 
@@ -494,17 +496,17 @@ namespace Azure.Tools.GeneratorAgent.Tests
             // Assert
             Assert.Multiple(() =>
             {
-                Assert.That(result.IsSuccess, Is.True);
-                Assert.That(result.Value!.Count, Is.EqualTo(1));
-                Assert.That(result.Value.ContainsKey("test.tsp"), Is.True);
-                Assert.That(result.Value.ContainsKey("readme.md"), Is.False);
+                Assert.That(result, Is.Not.Null);
+                Assert.That(result.Count, Is.EqualTo(1));
+                Assert.That(result.ContainsKey("test.tsp"), Is.True);
+                Assert.That(result.ContainsKey("readme.md"), Is.False);
             });
 
 
         }
 
         [Test]
-        public async Task GetTypeSpecFilesAsync_WithFailedFileDownload_ShouldSkipFailedFiles()
+        public void GetTypeSpecFilesAsync_WithFailedFileDownload_ShouldThrowHttpRequestException()
         {
             // Arrange
             var mockHandler = CreateMockHttpMessageHandler();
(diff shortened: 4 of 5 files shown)


src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -17,8 +14,9 @@ namespace Azure.Tools.GeneratorAgent
     internal class BuildErrorAnalyzer
     {
         // Regular expression to match error patterns.
+        // Example: "error CS0103: The name 'InvalidVariable' does not exist in the current context [/path/to/file.cs]"
```
jsquire: :heart

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -62,14 +59,18 @@ public async Task FixCodeAsync(CancellationToken ct)
...
+            CancellationToken ct = default)
         {
-            List<string> uploadedFilesIds = await UploadTspAsync(typeSpecDir, ct);
-            if (uploadedFilesIds.Count == 0)
-                throw new InvalidOperationException("No TypeSpec files (*.tsp) found in the directory. Cannot proceed with AZC error fixing.");
+            IEnumerable<string> uploadedFilesIds = await UploadTspFromMemoryAsync(typeSpecFiles, ct);
+            List<string> uploadedFilesList = uploadedFilesIds.ToList(); // Materialize once for multiple usage
+            
+            if (uploadedFilesList.Count == 0)
+                throw new InvalidOperationException("No TypeSpec files provided. Cannot proceed with AZC error fixing.");
```
jsquire:
```suggestion
            if (uploadedFilesList.Count == 0)
            {
                throw new InvalidOperationException("No TypeSpec files provided. Cannot proceed with AZC error fixing.");
            }
```

Braces always, please.  This keeps us from having some subtle errors are the code changes over time.  The one exception that we allow is single-line use.  Even that, I'd like to discourage.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -78,16 +79,23 @@ public async Task<string> InitializeAgentEnvironmentAsync(string typeSpecDir, Ca
...
+                kvp.Key.EndsWith(".yml", StringComparison.OrdinalIgnoreCase) ||
+                kvp.Key.EndsWith(".json", StringComparison.OrdinalIgnoreCase));
+
+            IEnumerable<Task<string?>> uploadTasks = relevantFiles.Select(async file =>
             {
                 await ConcurrencyLimiter.WaitAsync(ct);
                 try
                 {
-                    return await UploadSingleFileAsync(file, ct);
+                    return await UploadSingleFileFromMemoryAsync(file.Key, file.Value, ct);
```
jsquire:
```suggestion
                    return UploadSingleFileFromMemoryAsync(file.Key, file.Value, ct);
```

We can elide the state machine by just returning here instead of awaiting.   The only difference is what appears in the stack trace.   Since the source is a lambda expression, there's no real readable value to having it.   (You'll have to remove `async` from the signature too)

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -96,21 +104,22 @@ private async Task<List<string>> UploadTspAsync(string typeSpecDir, Cancellation
...
+            string fileName, 
+            string content, 
+            CancellationToken ct)
         {
             try
             {
-                string fileName = Path.GetFileNameWithoutExtension(filePath);
-                string txtFileName = $"{fileName}.txt";
+                string fileNameWithoutExt = Path.GetFileNameWithoutExtension(fileName);
+                string txtFileName = $"{fileNameWithoutExt}.txt";
```
jsquire:
```suggestion
                string txtFileName = Path.ChangeExtension(fileName, "txt");
```

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GithubFilesRetreiveService.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Text.Json.Serialization;
+using Azure.Tools.GeneratorAgent.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Tools.GeneratorAgent
+{
+    /// <summary>
+    /// TypeSpec file service for GitHub repository access using GitHub API for fast file retrieval.
+    /// </summary>
+    internal class GitHubFilesRetreiveService : IDisposable
```
jsquire:
```suggestion
    internal class GitHubFileRetrievalService : IDisposable
```

nit:  Read a bit awkard.   Maybe something simpler like "GitHubFileService" would work?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GithubFilesRetreiveService.cs
```diff
@@ -0,0 +1,152 @@
...
+            ArgumentNullException.ThrowIfNull(validationContext);
+
+            AppSettings = appSettings;
+            Logger = logger;
+
+            CommitId = validationContext.ValidatedCommitId;
+            TypespecSpecDir = validationContext.ValidatedTypeSpecDir;
+
+            // Initialize HttpClient with User-Agent header and timeout for GitHub API
+            HttpClient = new HttpClient
```
jsquire: We should look at injecting this via DI so that the lifetime is managed automatically via use of `HttpClientFactory`.   Assuming you take the short name, it would look like:

```csharp
services.AddHttpClient<GitHubFileService>(client =>
{
    client.BaseAddress = new Uri("https://api.github.com");
    client.Timeout = TimeSpan.FromMinutes(2);
    client.DefaultRequestHeaders.Add("User-Agent", "AzureSDK-TypeSpecGenerator/1.0");
});
```

You'll need to add a reference to `Microsoft.Extensions.Http` for this extension.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
m-redding:
```suggestion
        // Matches strings like "error CS0103: The name 'InvalidVariable' does not exist in the current context [/path/to/file.cs]" into groups: "error (CS0103): (The name 'InvalidVariable' does not exist in the current context) [/path/to/file.cs]"
```
nit just a suggestion for consideration

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GithubFilesRetreiveService.cs
```diff
@@ -0,0 +1,152 @@
...
+
+                IEnumerable<GitHubContent> tspFiles = contents.Where(c => 
+                    string.Equals(c.Type, "file", StringComparison.Ordinal) && 
+                    c.Name.EndsWith(".tsp", StringComparison.OrdinalIgnoreCase));
+
+                // Download each .tsp file content in parallel
+                Task<(string FileName, string Content)>[] downloadTasks = tspFiles
+                    .Where(tspFile => !string.IsNullOrEmpty(tspFile.DownloadUrl))
+                    .Select(tspFile => DownloadFileContentAsync(tspFile.Name, tspFile.DownloadUrl!, cancellationToken))
+                    .ToArray();
```
jsquire: You don't need to materialize the list.   `Task.WhenAll` takes an `IEnumerable`

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GithubFilesRetreiveService.cs
```diff
@@ -0,0 +1,152 @@
...
+                return (fileName, string.Empty);
+            }
+            catch (Exception ex)
+            {
+                Logger.LogCritical(ex, "Error downloading file {FileName}", fileName);
+                return (fileName, string.Empty);
+            }
+        }
+
+        public void Dispose()
```
jsquire: Let's move this to injected via DI and then you can remove the need for disposal here.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Program.cs
```diff
@@ -90,33 +89,49 @@ private async Task<int> ExecuteGenerationAsync(string? typespecdir, string? comm
                 AppSettings appSettings = ToolConfig.CreateAppSettings(appSettingsLogger);
                 ProcessExecutor processExecutor = new ProcessExecutor(LoggerFactory.CreateLogger<ProcessExecutor>());
 
-                // Step 2: Create and Initialize ErrorFixer Agent
+                // Step 2: Get TypeSpec files (from local or GitHub)
+                using TypeSpecFileService fileService = new TypeSpecFileService(
```
jsquire: You really don't want to do a 1-use `HttpClient` if we're going to be iterating.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
radhgupta: Sorry, I always tend to forget it! I need to fix my memory.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GithubFilesRetrieveService.cs
```diff
@@ -0,0 +1,152 @@
...
+                Timeout = TimeSpan.FromMinutes(2) // Reasonable timeout for file downloads
+            };
+            HttpClient.DefaultRequestHeaders.Add("User-Agent", "AzureSDK-TypeSpecGenerator/1.0");
+        }
+
+        public async Task<Result<Dictionary<string, string>>> GetTypeSpecFilesAsync(CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                Result<Dictionary<string, string>> apiResult = await FetchTypeSpecFilesViaApiAsync(cancellationToken);
```
m-redding:
```suggestion
                Result<Dictionary<string, string>> apiResult = await FetchTypeSpecFilesViaApiAsync(cancellationToken).ConfigureAwait(false);
```
nit

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -96,21 +104,22 @@ private async Task<List<string>> UploadTspAsync(string typeSpecDir, Cancellation
             });
 
             string?[] results = await Task.WhenAll(uploadTasks);
-            List<string> uploadedFilesIds = results.Where(id => !string.IsNullOrEmpty(id)).Select(id => id!).ToList();
+            IEnumerable<string> uploadedFilesIds = results.Where(id => !string.IsNullOrEmpty(id)).Select(id => id!);
 
-            Logger.LogInformation("Successfully uploaded {Count}/{Total} TypeSpec files as text files", uploadedFilesIds.Count, tspFiles.Length);
+            Logger.LogInformation("Successfully uploaded TypeSpec files from memory");
```
m-redding, line 98: This is super nitpicky but if you're ever feeling like doing a little logging refactoring, there's a big perf savings for using source generation to define logger messages at compile time by using the LoggerMessage pattern and avoiding string interpolation - https://learn.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging

I did this in SCM for example - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/System.ClientModel/src/Internal/Logging/PipelineMessageLogger.cs#L53

This might not be as realistic in your agent app because there are so many different messages, but if there are some places where you're logging in a loop and you're logging the same message a bunch of times, or maybe you could consolidate some of the error logging, then using the LoggerMessage pattern instead of LogInformation, LogWarning, etc could be a big perf boost

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GithubFilesRetrieveService.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        private async Task<Result<Dictionary<string, string>>> FetchTypeSpecFilesViaApiAsync(CancellationToken cancellationToken)
+        {
+            try
+            {
+                // GitHub API endpoint for directory contents
+                string apiUrl = $"https://api.github.com/repos/{AppSettings.AzureSpecRepository}/contents/{TypespecSpecDir}?ref={CommitId}";
+
+                HttpResponseMessage response = await HttpClient.GetAsync(apiUrl, cancellationToken).ConfigureAwait(false);
```
m-redding: Not very familiar with this API but is there supposed to be an accept header?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GithubFilesRetrieveService.cs
```diff
@@ -0,0 +1,152 @@
...
+                return Result<Dictionary<string, string>>.Success(typeSpecFiles);
+            }
+            catch (Exception ex)
+            {
+                Logger.LogCritical(ex, "Error in GitHub API TypeSpec files fetch");
+                throw;
+            }
+        }
+
+        private async Task<(string FileName, string Content)> DownloadFileContentAsync(string fileName, string downloadUrl, CancellationToken cancellationToken)
```
m-redding: nit but if it's ok for downloadUrl to be null it's nice to be more explicit and have this be string? and then you can remove the null forgiving operator in L96

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GithubFilesRetrieveService.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            CommitId = validationContext.ValidatedCommitId;
+            TypespecSpecDir = validationContext.ValidatedTypeSpecDir;
+
+            // Initialize HttpClient with User-Agent header and timeout for GitHub API
+            HttpClient = new HttpClient
+            {
+                Timeout = TimeSpan.FromMinutes(2) // Reasonable timeout for file downloads
+            };
+            HttpClient.DefaultRequestHeaders.Add("User-Agent", "AzureSDK-TypeSpecGenerator/1.0");
```
m-redding: Per Jesse's comment about not making a new HttpClient each time, you could make this static shared like in https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/System.ClientModel/src/Pipeline/HttpClientPipelineTransport.cs#L19

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecFileService.cs
```diff
@@ -0,0 +1,118 @@
...
+            string typeSpecDir, 
+            CancellationToken cancellationToken)
+        {
+            Logger.LogInformation("Reading TypeSpec files from local directory: {TypeSpecDir}", typeSpecDir);
+
+            try
+            {
+                Dictionary<string, string> typeSpecFiles = new Dictionary<string, string>();
+                
+                string[] allFiles = Directory.GetFiles(typeSpecDir, "*.tsp", SearchOption.AllDirectories);
```
m-redding, line 67: If I'm following the flow correctly, it seems like the local files are only pulling .tsp and the GitHub download is including TypeSpec related files too? Or am I misunderstanding?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
m-redding: Also this is just a thought to think about in general, definitely not blocking or necessary in this PR

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GithubFilesRetrieveService.cs
m-redding: NVM I didn't see his other suggestion for fixing this

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
radhgupta: We genuinely need a log reactoring because, we need to have 2 types of logs. One for the debugging purposes, one that would be shown to the user who uses the tool. Its already in my TODO. I really appreciate this insight.  I will keep it in my mind.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GithubFilesRetrieveService.cs
radhgupta: Added it in the DI!!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecFileService.cs
radhgupta: you are correct! the code was not consistent, I have fixed it now! thanks for letting me know :)
For consistency, purposes now  we only have a .tsp check for both github and local

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,215 @@
...
+                
+                if (!string.IsNullOrEmpty(githubToken))
+                {
+                    client.DefaultRequestHeaders.Authorization = 
+                        new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", githubToken);
+                }
+            })
+            .ConfigurePrimaryHttpMessageHandler(() => new HttpClientHandler());
+
+            services.AddHttpClient();
```
jsquire: Is this line a typo?   We shouldn't need it.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,215 @@
...
+
+            return services;
+        }
+
+        /// <summary>
+        /// Adds application-specific services to the service collection.
+        /// </summary>
+        private static IServiceCollection AddApplicationServices(this IServiceCollection services)
+        {
+            services.AddScoped<ErrorFixerAgent>();
```
jsquire: Why are we using `Scoped` rather than a singleton instance?    Do we have concurrency or stateful reasons that we'd need to kill/refresh them?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFilesService.cs
```diff
@@ -0,0 +1,159 @@
...
+using System.Text.Json.Serialization;
+using Azure.Tools.GeneratorAgent.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Tools.GeneratorAgent
+{
+    /// <summary>
+    /// TypeSpec file service for GitHub repository access using GitHub API for fast file retrieval.
+    /// </summary>
+    internal class GitHubFilesService
```
jsquire:
```suggestion
    internal class GitHubFileService
```

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFilesService.cs
```diff
@@ -0,0 +1,159 @@
...
+            try
+            {
+                // GitHub API endpoint for directory contents
+                string apiUrl = $"https://api.github.com/repos/{AppSettings.AzureSpecRepository}/contents/{TypespecSpecDir}?ref={CommitId}";
+
+                HttpResponseMessage response = await HttpClient.GetAsync(apiUrl, cancellationToken).ConfigureAwait(false);
+                
+                if (!response.IsSuccessStatusCode)
+                {
+                    string errorContent = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
```
jsquire: What do we do with the content from this call?   I'm wondering if we can avoid reading it into memory and stream it to a file inline, if that's what we do.

If we're using the content directly, then its probably fine, as we're screening to just TSP files and they shouldn't be huge.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
radhgupta: I got confused here, I read about scoped vs singleton and I think it should have been singleton. Thanks Jesse, for highlighting this!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
```diff
@@ -0,0 +1,158 @@
...
+            bool hasAuthHeader = httpClient.DefaultRequestHeaders.Authorization != null;
+            Logger.LogInformation("GitHub API client initialized. Authentication: {AuthStatus}", 
+                hasAuthHeader ? "Configured" : "Not configured - using rate-limited access");
+        }
+
+        public async Task<Result<Dictionary<string, string>>> GetTypeSpecFilesAsync(CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                Result<Dictionary<string, string>> apiResult = await FetchTypeSpecFilesViaApiAsync(cancellationToken).ConfigureAwait(false);
```
m-redding: Does this need to be in a helper? It just seems like an unnecessary extra call and try/catch since this method doesn't do any additional work

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
```diff
@@ -0,0 +1,158 @@
...
+            }
+            catch (Exception ex) when (ex is not OperationCanceledException)
+            {
+                Logger.LogCritical(ex, "Unexpected error while fetching TypeSpec files from GitHub repository {Repository}", 
+                    AppSettings.AzureSpecRepository);
+                throw;
+            }
+        }
+
+        private async Task<Result<Dictionary<string, string>>> FetchTypeSpecFilesViaApiAsync(CancellationToken cancellationToken)
```
m-redding: I feel like the error handling in this method is a bit inconsistent. i.e. 
- My get directory contents fails -> immediate and specific failure
- All of my DownloadFileContentAsync tasks threw an exception -> success and return an empty dictionary

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
```diff
@@ -0,0 +1,158 @@
...
+                    throw new HttpRequestException($"GitHub API request failed: {response.StatusCode} {response.ReasonPhrase}");
+                }
+
+                string jsonContent = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
+                GitHubContent[]? contents = JsonSerializer.Deserialize<GitHubContent[]>(jsonContent, JsonOptions);
+
+                if (contents == null)
+                {
+                    throw new InvalidOperationException("Failed to deserialize GitHub API response");
+                }
```
m-redding: nit but it might be nice to make this exception message a little more actionable

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
```diff
@@ -0,0 +1,158 @@
...
+                HttpResponseMessage response = await HttpClient.GetAsync(downloadUrl, cancellationToken).ConfigureAwait(false);
+                if (response.IsSuccessStatusCode)
+                {
+                    string content = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
+                    return (fileName, content);
+                }
+
+                Logger.LogWarning("Failed to download file {FileName} from {Url}: {StatusCode} {ReasonPhrase}", 
+                    fileName, downloadUrl, response.StatusCode, response.ReasonPhrase);
+                return (fileName, string.Empty);
```
m-redding: Curious what kinds of errors are getting swallowed here?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Program.cs
```diff
@@ -81,42 +82,48 @@ internal async Task<int> HandleCommandAsync(string? typespecPath, string? commit
...
+                // Step 1: Get services from DI container
+                AppSettings appSettings = ServiceProvider.GetRequiredService<AppSettings>();
+                
+                // Step 2: Get TypeSpec files (from local or GitHub)
+                var fileServiceFactory = ServiceProvider.GetRequiredService<Func<ValidationContext, TypeSpecFileService>>();
+                TypeSpecFileService fileService = fileServiceFactory(validationContext);
+
+                Result<Dictionary<string, string>> fileResult = await fileService.GetTypeSpecFilesAsync(cancellationToken);
+
+                if (fileResult.IsFailure)
```
m-redding: It seems like GetTypeSpecFilesAsync only ever throws or returns a success result, maybe using the Result abstraction is unnecessary in this case?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
radhgupta: I wanted to create this thin wrapper around the `FetchTypeSpecFilesViaApiAsync` because, it helps me in having a consistent log pattern (logCritical) around failure occuring at each step. Also, for testing purposes, we only need to mock the public `GetTypeSpecFilesAsync`. It makes our tests much simpler because we don't have to deal with all the HTTP and JSON complexity inside.

I may be wrong here, please correct me if I am wrong.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
radhgupta: I agree, I fixed it. Now in case of even partial downloadable files fail, it will directly throw error rather than passing empty dictionary.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Program.cs
radhgupta: I agree with your point! The Result abstraction was introduced to maintain consistency with other methods called in Program.cs. However, I also see validity in your concern that this might be overengineering for this specific case.
Could you help me understand the trade-offs here? Should we prioritize consistency across the codebase (using Result everywhere) or focus on simplicity and avoid potential overengineering? I'd value your perspective on which approach would be better for long-term maintainability

m-redding: That makes sense about the consistency and I'm fine if you do want to leave it so everything is aligned.
Looking at the other method calls here tho my perspective is that this case seems to match up with the initialize agent environment rather than the compile/build outputs. For the compile/build the fail result is important for the flow of the app, whereas here the fail result means the app can't run properly and should quit.

radhgupta: Oh yes!! it makes sense, I think we should avoid this over engineering!!!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
m-redding: Could the log critical in the catch of FetchTypeSpecFilesViaApiAsync be combined with the one in this catch to cover the logging base? Or is there a certain pattern you're looking to log? And for mocking you should still be able to mock GetTypeSpecFilesAsync the same way without the ~inner helper~ thin wrapper

(edited to be more specific - because adding helpers can be helpful for mocking but since the inner helper isn't isolating certain behavior that can be mocked, it shouldn't matter in this case)

PR #11650, opened 2025-08-13
[azsdk-cli] Adding in tests for the ReadMeGeneratorTool

richardpark-msft:
Adding in some simple testing for ReadMeGenerator.

One is a unit test - pretty simple, mocks OpenAIClient and returns a faux response.

The second one is a live test, which will only run if the proper environment variables are defined. We need to get a pipeline going before that'll run anywhere else but on my local machine.

Both of these tests use a ServiceCollection and invoke the command through the command line framework. So it's as real as I can get it.

I also added in some other utilities:
- MockOutputService, if you want to just accumulate all the Output/OutputError calls into a single list and assert on it
- OpenAIMockHelper, just to make it simple to create and wire up a chat client.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Azure.Sdk.Tools.Cli.Tests.csproj
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Azure.Sdk.Tools.Cli.Tests.csproj
@@ -35,6 +35,12 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Include="..\Azure.Sdk.Tools.Cli\Templates\ReadMeGenerator\README-template.go.md" Link="TestAssets\README-template.go.md">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
+
+  <ItemGroup>
     <None Update="TypeSpecTestData\specification\testcontoso\Contoso.Management\employee.tsp">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
@@ -1,0 +1,162 @@
+using System.ClientModel;
+using System.ClientModel.Primitives;
+using System.CommandLine;
+using System.CommandLine.Parsing;
+using Azure.AI.OpenAI;
+using Azure.Sdk.Tools.Cli.Services;
+using Azure.Sdk.Tools.Cli.Tests.MockServices;
+using Azure.Sdk.Tools.Cli.Tests.TestHelpers;
+using Azure.Sdk.Tools.Cli.Tools;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Moq;
+using OpenAI.Chat;
+
+namespace Azure.Sdk.Tools.Cli.Tests.Tools.Generators
+{
+    internal class ReadMeGeneratorToolTests
+    {
+        [Test]
+        public async Task TestReadmeGeneratorTool()
+        {
+            var testClients = SetupOpenAIMocks();
+            (DirectoryInfo root, string packagePath) = await CreateFakeLanguageRepo();
+
+            var readmeOutputPath = Path.GetTempFileName();
+            var readmeTemplatePath = Path.Combine(AppContext.BaseDirectory, "TestAssets", "README-template.go.md");
+
+            try
+            {
+                var tool = ActivatorUtilities.CreateInstance<ReadMeGeneratorTool>(testClients.ServiceProvider);
+                var command = tool.GetCommand();
+
+                int exitCode = command.Invoke($"--output-path \"{readmeOutputPath}\" --service-url \"https://learn.microsoft.com/azure/service-bus-messaging\" --template-path {readmeTemplatePath} --package-path {packagePath}");
+
+                Assert.Multiple(() =>
+                {
+                    Assert.That(exitCode, Is.EqualTo(0), "Command should execute successfully");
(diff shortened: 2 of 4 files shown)


richardpark-msft:
@benbp, I think this is what we were thinking of, but let me know.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Azure.Sdk.Tools.Cli.Tests.csproj
```diff
@@ -34,6 +34,12 @@
     <None Include="TestAssets\**\*" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Include="..\Azure.Sdk.Tools.Cli\Templates\ReadMeGenerator\README-template.go.md" Link="TestAssets\README-template.go.md">
```
benbp, line 38: Should we combine all these copy to directives within one item group, or is the separate organization by test area useful?

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            try
+            {
+                var tool = ActivatorUtilities.CreateInstance<ReadMeGeneratorTool>(testClients.ServiceProvider);
+                var command = tool.GetCommand();
+
+                int exitCode = command.Invoke($"--output-path \"{readmeOutputPath}\" --service-url \"https://learn.microsoft.com/azure/service-bus-messaging\" --template-path {readmeTemplatePath} --package-path {packagePath}");
+
+                Assert.Multiple(() =>
+                {
+                    Assert.That(exitCode, Is.EqualTo(0), "Command should execute successfully");
```
benbp, line 37: Is it worth testing string formatting here? The other option is to add `-o json` and then try to cast the CLI response back to the response type and check against properties. That or sniff on what `output.Format()` is called with and check the type properties.

To me string result testing is a balance between incompleteness (partial matching) and brittleness (full matching). If we want to test string output then we could have tests for the tool's response class instead, and just do property assertions here. That kind of breaks the black box testing mode though. Curious for your thoughts.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            finally
+            {
+                Directory.Delete(root.FullName, true);
+            }
+        }
+
+        [Test]
+        public void TestReadmeGeneratorToolLive()
+        {
+            var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT");
```
benbp, line 53: Fine for now but I'm taking a note to prioritize a higher level configuration class that can source env vars + other sources.

benbp (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Azure.Sdk.Tools.Cli.Tests.csproj
richardpark-msft: Oh, I'm happy to combine them. I'm not an msbuild guy, but since this is grabbing it from a link to a file will that still work?

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
richardpark-msft: I think there's two parts to this - I agree with you most of the time but I think the free-form output from our commands (to some extent) is a contract that we should adhere to, since we want to produce consistent behavior with LLMs that might read it.

However, I think I can codify that a bit more by making them constants, or something similar. I'd still check the literal string here so we know this is intentionally "static".

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
richardpark-msft: Yeah, this is a stop-gap measure. An injectable test object would be preferrable.

I do setup the service collection below so that's easy to do.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
richardpark-msft: I'll do something locally here, we can discuss it more in the larger group.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Azure.Sdk.Tools.Cli.Tests.csproj
richardpark-msft: Ah, never mind, I think I'm stuck because the file is a linked file, and not actually present in the folder. 

(it's the "production" readme template)

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
richardpark-msft: Let me hold this one for a separate PR - thinking about it more I think it would be better to standardize this kind of "Output".

benbp: @richardpark-msft currently we actually force json output in MCP mode for this kind of reason. So testing string output is just exercising the CLI path. But I agree with your point about the output being a contract due to LLM usage (which eventually can be tested with foundry evals).

PR #11659, opened 2025-08-13
[azsdk-cli] make package check an enum

l0lawrence:


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/CheckAllTool/PackageCheckTool.cs
```diff
@@ -37,9 +37,9 @@ public override Command GetCommand()
             Command command = new("run-checks", "Run validation checks for SDK packages");
             command.AddOption(SharedOptions.PackagePath);
             
-            var checkTypeOption = new Option<string>(
+            var checkTypeOption = new Option<PackageCheckType>(
```
l0lawrence: re- Sameeksha's comment, do we want a default value here?

timovv: I guess if it defaults to all that could make sense. Can't think of any other sane default value, but also running all checks by default doesn't feel great since that could take a long time (so perhaps should be opt-in).

Generally I think of flags starting with `--` as optional; having one as required feels weird. If we require people specify the check I would rather see it as a subcommand e.g. `azsdk check changelog`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/PackageCheckType.cs
```diff
@@ -0,0 +1,22 @@
+namespace Azure.Sdk.Tools.Cli.Models;
+
+/// <summary>
+/// Defines the types of validation checks that can be run on SDK packages.
+/// </summary>
+public enum PackageCheckType
```
timovv:
```suggestion
public enum PackageCheckName
```

Nit, maybe `PackageCheckName` would be a little clearer? At first glance, this name made me think that each enum value represented a category of checks and that there would be multiple checks of each type. Might just be me though

timovv (approved):

richardpark-msft:
@l0lawrence, I can approve it but you have to move it out of draft and stand by your work (and signal to me that it's really ready)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/CheckAllTool/PackageCheckTool.cs
l0lawrence: yah thats fair, we have another issue task for the command naming

timovv (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/PackageCheckName.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// </summary>
+    All,
+    
+    /// <summary>
+    /// Run changelog validation check.
+    /// </summary>
+    Changelog,
+    
+    /// <summary>
+    /// Run dependency analysis check.
```
richardpark-msft, line 19: Not for this PR, but I'd like to get a better idea of what this means. I don't think it translates 1:1 with other languages - perhaps it's very clear in Python what that means.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/PackageCheckName.cs
```diff
@@ -0,0 +1,22 @@
+namespace Azure.Sdk.Tools.Cli.Models;
+
+/// <summary>
+/// Defines the types of validation checks that can be run on SDK packages.
+/// </summary>
+public enum PackageCheckName
+{
```
richardpark-msft, line 7:
```suggestion
public enum PackageCheckType
{
```

richardpark-msft: (it's not a _name_, that I can tell?)

richardpark-msft (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/PackageCheckName.cs
l0lawrence: It is different per lang we have a work item from the design meeting for each lang to figure out what it means for them -- we can update the description as we implement

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/PackageCheckName.cs
l0lawrence: I was leaning name because each checkName could contain different check types depending on the language (like formatCode is black for python but could be eslint or something for js) that might be overthinking it will circle back as I add more checks

PR #11660, opened 2025-08-13
Microagent framework v1

timovv:
#11611 but without the test recorder stuff so we can iterate on this separately

### To do

- [x] Write tests for the tools and the agent host.
- [x] ~~Implement tools for updating files and better code search (e.g fuzzy find) like VS code has.~~ (out of scope for this PR, let's add these on an as-needed basis)
- [x] ~~Add a mechanism for tool call approval, for destructive tools such as writing to a file~~ (out of scope for this PR)
- [x] Harden existing tools, e.g., to ensure the agent can't break out of the base directory using `../../../` in a tool call
- [x] How do we handle dependency injection in tools (@richardpark-msft) - thinking about this some more I think we wouldn't really be in the middle here. You'd just instantiate your instance using [ActivatorUtilities](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities.createinstance?view=net-9.0-pp) or similar. The LanguageHelperFactory already does.


richardpark-msft:
@timovv, do you mind updating the PR description to list the things you know you wanted TODO? I don't mind helping out, but just want to "claim" things so we don't bump into each other.

richardpark-msft:
@timovv, I'll take 'How do we handle dependency injection in tools'.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/MicroagentHostService.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            string toolResult;
+            try
+            {
+                toolResult = await tool.InvokeAsync(toolCall.FunctionArguments.ToString(), ct);
+            }
+            catch (Exception e)
+            {
+                // TODO: is this the best way to communicate an error to the LLM?
+                toolResult = $"Error: {e.Message}";
```
haolingdong-msft, line 99: Or we can define a error structure/model and return the structured data to LLM?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/MicroagentHostService.cs
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /// <summary>
+    /// Wrapper object for all tool call results so that we are able to make calls that e.g. just return a string
+    /// (OpenAI expects an object at the top level)
+    /// </summary>
+    /// <typeparam name="T">Type of the result</typeparam>
+    private class ToolCallResult<T>
+    {
+        [Description("The result of the agent run. Output the result requested exactly, without additional padding, explanation, or code fences unless requested.")]
```
haolingdong-msft, line 117:
```suggestion
        [Description("The result of the agent run. Output the result requested exactly, without additional padding, explaination, or code fences unless requested.")]
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
```diff
@@ -31,6 +32,8 @@ public static void RegisterCommonServices(IServiceCollection services)
             services.AddSingleton<ISpecPullRequestHelper, SpecPullRequestHelper>();
             services.AddSingleton<IUserHelper, UserHelper>();
 
+            services.AddSingleton<IMicroagentHostService, MicroagentHostService>();
```
haolingdong-msft, line 49: For discussion - Do we always have one micro agent instance, or we would have multiple instances which are good at specific tasks. e.g. sample agent, test agent, readme agent, update customized code agent etc.

From my understanding, if we have only one microagent instance, from OpenAI side, it is corresponding to one chat session, while if we have multiple micro agents, there would be multiple chat sessions. I think one benifit of having multiple chat sessions is: it would be easier to manage the LLM context.

Happy to be corrected if I understand wronly and would like to know your thoughts.

haolingdong-msft:
@timovv & @richardpark-msft, Thanks a lot for the PoC! It helps me a lot to learn more on the microagent idea!
I like the idea of having a small agent specialized at specific tasks, with minimum needed tools and context. Maybe in the future, we can have 'test micro agent', 'customized code micro agent', 'readme agent' etc.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
timovv: Here we have one instance of the _host service_, which is responsible for running multiple microagents. Like you're saying, supporting multiple chat sessions is important. The current design does allow for this -- each call to `MicroagentHostService.RunAgentToCompletionAsync` creates an entirely new conversation so we get a fresh context each time.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/MicroagentHostService.cs
timovv: It really is spelled explanation, for some reason.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/MicroagentHostService.cs
timovv: Yeah, I think we need some way for tools to provide more error information than just an exception message.

timovv:
Copy-pasting @benbp's comment from the other PR here:

> There are also some orchestration options and attributes to populate the tool list schema available in [semantic kernel](https://learn.microsoft.com/en-us/semantic-kernel/frameworks/agent/agent-orchestration/?pivots=programming-language-csharp). Though I like that this is small and self contained, easily understandable.
> 
> How much we want to bake agents into the CLI or MCP server has been an open question for me, but I don't think it's something we'll be able to completely avoid. While I love the simplicity of the flat mcp client -> tool call structure, I think that sometimes we will need control over model, specialized prompts, chat history and other infra for things like RAG.
> 
> An open question I have in this area is how the tool gets limited in portability the more agent to agent dependencies we create, as the CLI may end up having a lot of infra/auth dependencies. If we're just doing it for model selection and prompts, I'm more tempted to invest effort in making sure we have reliability cross model (or can define some sort of minimum mcp client model assertion per tool). For flows like test proxy recording and interpretation, how much of that is really only possible/optimal with agent orchestration vs. leveraging an mcp client (or an agent calling the CLI in the shell).
> 
> Overall I like this idea and it would be interesting to have the cli be a home for it, though I'm still not sure where I land on how we balance which scenarios support mcp vs. agent flows vs. both. Then there are maintenance toil costs to consider like model hosting and upgrades (e.g. gpt-5 deprecating everything).

The MCP Client -> tool -> return to client flow is something that I think works well when each tool is standalone and makes no assumption about the presence or absence of other tools. However, we're seeing more and more examples in our workflow where different tools have interdependencies. For example:
- In the release agent workflow, we expect the LLM to call tools in a certain order
- In @samvaity's customization workflow, there is an AI-orchestrated series of tools which need to be followed
- The recorded test workflow I was working on in #11611 has a similar idea: deterministic workflow, with some steps where AI can help
- Sample and test generation also have multiple AI steps which feed into each other like in Deyaa's prototype.

To use the "flat MCP client approach" in workflows like the above the flow is something like
1. MCP client calls the first tool
1. Tool does some computation, and then **returns instructions for the client**, including next tool to call and how
1. Client _hopefully_ follows the instructions and then calls the next tool. Rinse and repeat.

I appreciate the simplicity of the approach, and the fact that we don't have to make our own LLM calls or manage LLM resources is definitely a plus. That said, behind the apparent simplicity, there is, I think, a deceptive amount of complexity:
- Testability. It's really hard to test this. How do we verify that the correct tool calls are made beyond spinning up an actual instance of VS Code/whatever MCP client?
- No guarantee of determinism. We can't ever be 100% sure that the MCP client will follow our instructions and call the next tool. In contrast, if the orchestration is in code, determinism is guaranteed for everything outside of the LLM call itself.
- Tight coupling to the MCP client implementation. When tools are standalone, it doesn't matter which MCP client is calling them -- the tools are simply made available, it's up to the client to decide when to call them, and we don't care what the client does with the result. Once we add interdependencies between tool calls, however, we do start to care about whether the client follows the instructions provided in the result. And while we may test our tools with a certain client using a certain model, as soon as the client changes the system prompt, the toolset, or the model changes, there is a risk that we get broken. I worry that this will make our tooling brittle and not portable despite our best efforts.
- Prompt injection. Perhaps this isn't a huge concern, since in theory MCP servers should be trusted, but the whole idea of 'return instructions in the tool call result' looks like a potential attack vector to me. A malicious tool could return malicious instructions. Even though the way we are using this technique is benign, I wouldn't be surprised if we see MCP clients make efforts to safeguard against these prompt injection attacks, for example by adding guidance in the system prompt to not follow instructions returned by tool calls. If such a mitigation was attempted, our cookbooks would be impacted as collateral damage.

Essentially, MCP tool chaining adds a hidden dependency on the MCP client's implementation details I think we should try and avoid if we can (granted, this won't always be feasible -- like I don't see how a long running flow like the release agent could use a different approach as is). The microagent model, where we take a direct dependency on the LLM, makes the dependency explicit and gives us full control over the toolset and the context window. Sure, we may need to make changes every now and then like for when certain models get retired, but we stay in control of when and how we make the change. This does come at a cost in terms of having to maintain more code and an LLM resource, but I suspect tool chaining has a lot of hidden costs that we would be able to avoid by doing this.

To be clear, I still definitely see the value of exposing our tooling as MCP tools, from a discoverability point of view and as to allow Copilot to use it. But keeping each tool standalone is, I think, a good thing where possible, and the microagent approach allows us to do that. We are moving forward with this approach in the inner loop effort and from what I can see so far, there is promise.

richardpark-msft:
@timovv:

>  Add a mechanism for tool call approval, for destructive tools such as writing to a file and

I think we should remove this. We already have a constraint on the boundaries of where a file is written to and the entire point of this is to be automatic, without user input. We can reconsider this for later.

richardpark-msft:
> @timovv & @richardpark-msft, Thanks a lot for the PoC! It helps me a lot to learn more on the microagent idea! 

(To be clear, this is ALL @timovv, I'm just excited about his work on this) :)

timovv:
> I think we should remove this. We already have a constraint on the boundaries of where a file is written to and the entire point of this is to be automatic, without user input. We can reconsider this for later.

I agree it shouldn't block merging what we have but I do think that having some way to allow for verification is important for some cases. It can be opt-in. But if I want the agent to make whatever fixes to my tests it deems necessary, for example, then I would have to give it permission to update all my test files, which still gives it quite a lot of room to mess things up.

richardpark-msft:
> > I think we should remove this. We already have a constraint on the boundaries of where a file is written to and the entire point of this is to be automatic, without user input. We can reconsider this for later.
> 
> I agree it shouldn't block merging what we have but I do think that having some way to allow for verification is important for some cases. It can be opt-in. But if I want the agent to make whatever fixes to my tests it deems necessary, for example, then I would have to give it permission to update all my test files, which still gives it quite a lot of room to mess things up.

Perhaps that would be an agreed-upon parameter, passed in, rather than an interactive prompt.

Similar to say, --force, or -Confirm style switches for most CLI commands.

timovv:
Right, I guess that's what I was trying to get at with "opt-in". If a user is running it and wants to check the agent's work interactively, we should allow that, but we can have a flag to bypass for non-interactive environments.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Microagent.cs
```diff
@@ -0,0 +1,32 @@
...
+/// It operates an agent loop (LLM <-> tool call) until the LLM determines that its task is complete, at which point
+/// it calls the special sentinel "exit" tool with the result.
+/// </summary>
+/// <typeparam name="TResult">A type representing the schema of the result of the agent's execution</typeparam>
+public class Microagent<TResult>
+{
+    /// <summary>
+    /// The system prompt provided to the agent.
+    /// </summary>
+    public required string SystemPrompt { get; init; }
```
timovv: The OpenAI Responses API calls this property [`instructions`](https://platform.openai.com/docs/api-reference/responses/create#responses_create-instructions). Any opinions? I lean towards instructions since we don't have the distinction of user vs. system prompt here

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/AgentTool.cs
```diff
@@ -0,0 +1,27 @@
+
+using System.Text.Json;
+
+namespace Azure.Sdk.Tools.Cli.Microagents;
```
praveenkuttappan, line 4: MicroAgents?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/ReadFileTool.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            throw new ArgumentException("The provided path is invalid or outside the allowed base directory.");
+        }
+        if (!File.Exists(path))
+        {
+            throw new ArgumentException($"{path} does not exist");
+        }
+
+        // Read the file content
+        var content = File.ReadAllText(path);
```
richardpark-msft:
```suggestion
        var content = await File.ReadAllTextAsync(path);
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/ReadFileTool.cs
```diff
@@ -0,0 +1,34 @@
...
+
+public record ReadFileInput([property: Description("Relative path of the file to read")] string FilePath);
+public record ReadFileOutput([property: Description("The content of the file")] string FileContent);
+
+public class ReadFileTool(string baseDir) : AgentTool<ReadFileInput, ReadFileOutput>
+{
+    public override string Name { get; init; } = "ReadFile";
+    public override string Description { get; init; } = "Read the contents of a file";
+
+    public override Task<ReadFileOutput> Invoke(ReadFileInput input, CancellationToken ct)
```
richardpark-msft:
```suggestion
    public override async Task<ReadFileOutput> Invoke(ReadFileInput input, CancellationToken ct)
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/ReadFileTool.cs
```diff
@@ -0,0 +1,34 @@
...
+            throw new ArgumentException("The provided path is invalid or outside the allowed base directory.");
+        }
+        if (!File.Exists(path))
+        {
+            throw new ArgumentException($"{path} does not exist");
+        }
+
+        // Read the file content
+        var content = File.ReadAllText(path);
+        return Task.FromResult(new ReadFileOutput(content));
```
richardpark-msft:
```suggestion
        return new ReadFileOutput(content);
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/ToolHelpers.cs
```diff
@@ -0,0 +1,108 @@
...
+        },
+        TreatNullObliviousAsNonNullable = true,
+    };
+
+    /// <summary>
+    /// Try to resolve a provided path against a base directory, ensuring the result stays within the base directory.
+    /// </summary>
+    /// <param name="baseDirectory">The base directory that bounds all operations.</param>
+    /// <param name="relativePath">A relative path provided by the user or caller.</param>
+    /// <param name="fullPath">Outputs the resolved full path if successful; otherwise an empty string.</param>
```
richardpark-msft:
```suggestion
    /// <param name="fullPath">Outputs the resolved full path if successful; otherwise it's null.</param>
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/ListFilesTool.cs
```diff
@@ -0,0 +1,49 @@
...
+public record ListFilesOutput(IList<string> fileNames);
+
+public class ListFilesTool(string baseDirectory) : AgentTool<ListFilesInput, ListFilesOutput>
+{
+    public override string Name { get; init; } = "ListFiles";
+    public override string Description { get; init; } = "List files in the given directory";
+
+    private readonly string baseDirectory = baseDirectory;
+
+    public override Task<ListFilesOutput> Invoke(ListFilesInput input, CancellationToken ct)
```
praveenkuttappan, line 21: This method is not async. Any specific reason to return Task instead of `ListFilesOutput`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/AgentTool.cs
```diff
@@ -0,0 +1,27 @@
...
+/// <typeparam name="TInput">Input schema type</typeparam>
+/// <typeparam name="TOutput">Output schema type</typeparam>
+public abstract class AgentTool<TInput, TOutput> : IAgentTool
+{
+    public string InputSchema { get; } = ToolHelpers.GetJsonSchemaRepresentation(typeof(TInput));
+
+    public abstract string Name { get; init; }
+    public abstract string Description { get; init; }
+
+    public async Task<string> Invoke(string input, CancellationToken ct = default)
```
praveenkuttappan, line 19:
```suggestion
    public async Task<string> InvokeAsync(string input, CancellationToken ct = default)
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/MicroagentHostService.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        var chatCompletionOptions = new ChatCompletionOptions
+        {
+            // Force the LLM to choose a tool on every turn.
+            ToolChoice = ChatToolChoice.CreateRequiredChoice(),
+            // For simplicity, only allow for one tool to be called at once.
+            AllowParallelToolCalls = false,
+        };
+
+        // Add the special-case "exit" tool
```
richardpark-msft, line 41: Can you add a bit describing why this Exit tool exists? (the reasoning is good from our offline discussion, makes sense for it to be here)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Microagent.cs
richardpark-msft: Sounds good.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/ListFilesTool.cs
timovv: The method it overrides returns `Task` so our hands are tied.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/AgentTool.cs
timovv: Micro- is a prefix, so I think of this as one word. Like I wouldn't capitalize the s in microservices, I also don't think we should capitalize the A in microagent.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Microagent.cs
timovv: I updated it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/MicroagentHostService.cs
timovv: Let's revisit this in a future PR.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/MicroagentHostService.cs
```diff
@@ -0,0 +1,118 @@
...
+        chatCompletionOptions.Tools.Add(
+            ChatTool.CreateFunctionTool(ExitToolName,
+                "Call this tool when you are finished with the work or are otherwise unable to continue.",
```
praveenkuttappan, line 46: How is the criteria passed for OpenAI to figure out whether it finished the task?

timovv: It's up to the LLM to decide. Either it eventually thinks that it's done with the task and calls the Exit tool, or we hit the max number of tool calls before that happens and so we exit the loop and throw.

benbp:
> The microagent model, where we take a direct dependency on the LLM, makes the dependency explicit and gives us full control over the toolset and the context window.

Thanks for typing up the extended rationale, I have a better understanding of the problem and goals now. I think this will be a good approach to experiment with.

Some other questions:

- I assume these flows can potentially take a while and state tracking is an unsolved problem here. Would it be useful to consider running the microagent host in a daemon mode that can handle the LLM interchange? That way the user can async poll with the CLI, or ctrl-c and re-enter, etc. and we would have fewer occurrences where we have to cold start a flow that just needs to resume.
- Re: the destructive/force behavior with @richardpark-msft I've seen that azure-mcp has some concept of destructive/idempotent/read-only attributes (https://github.com/Azure/azure-mcp/blob/379b5571e5bdb681801d985837cf1b5d4191ae29/eng/tools/ToolDescriptionEvaluator/Models/McpModels.cs#L65-L81) but I'm not quite sure how they are utilized. Possibly some useful patterns there.
- There still seems like a lot of overlap with this and semantic kernel, which lets you define a set of kernel functions, a prompt, tool calling behavior, and then execute the set. Is it more helpful to roll our own here vs. using that framework which has a lot more interopability should we need it? I assume there is some other work in the roadmap that may diverge from it or become more complex?

PR #11663, opened 2025-08-13
Fix 11009

scbedd:
I decided to do some due diligence and also confirm that array is returning proper error. I actually discovered a _new_ bug for array comparison because of that. If the `request` body has extra entries we don't catch it rn. This fixes that. Will do a FULL recorded test run on all languages before I merge this.


scbedd:
When I was fixing the actual target of this problem, I discovered that we weren't comparing the shapes of arrays within json appropriately. We were only examining the first one. This could be a breaking change which is why this hasn't moved. I need to publish a proxy version and try some regression tests :/

PR #11672, opened 2025-08-14
[sdk-sample-generator] Update the Go side of the fence to match expectations

richardpark-msft:
- generateSamples needs to throw an exception if one comes back, otherwise you can run this program a bunch of times and not realize what's failing. Err...so I've heard.
- Updating the typecheck part to do checks the way we do in the pipelines (at least, some of it)
- adding some stuff to the system prompt for Go. I'm unsure if it's overriding the standard declaration that's already in the genaiscript package or not though.

### DIFF
--- a/tools/sdk-sample-generator/genaisrc/src/go/typecheck.ts
+++ b/tools/sdk-sample-generator/genaisrc/src/go/typecheck.ts
@@ -1,6 +1,6 @@
+import path from "node:path";
 import type { TypeCheckParameters, TypeCheckResult } from "../types.ts";
 import { getUniqueDirName } from "../utils.ts";
-import path from "node:path";
 
 const fileName = "temp.go";
 let container: ContainerHost | undefined = undefined;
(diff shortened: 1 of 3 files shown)


tools/sdk-sample-generator/genaisrc/src/go/typecheck.ts
```diff
@@ -40,39 +40,35 @@ export async function typecheckGo(
             );
         }
 
-        // Tidy up the Go module
-        const tidyResult = await container.exec("go", ["mod", "tidy"], {
-            cwd: projectDir,
-        });
+        const commands = [
+            ["go", "install", "github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.3.1"],
```
deyaaeldeen, line 45: can the LLM do anything if this command failed? If not, perhaps we can run it before the feedback loop.

tools/sdk-sample-generator/genaisrc/src/go/typecheck.ts
```diff
@@ -40,39 +40,35 @@ export async function typecheckGo(
...
-            cwd: projectDir,
-        });
+        for (let cmd of commands) {
+            const result = await container.exec(cmd[0], cmd.slice(1), {
+                cwd: projectDir,
+            });
+
+            const output = `${cmd.join(" ")} output:\n${result.stdout ?? ""}, ${result.stderr ?? ""}`;
+
+            if (result.exitCode === 0 || result.failed) {
```
deyaaeldeen: I think 0 means success, no?
```suggestion
            if (result.exitCode !== 0 || result.failed) {
```

tools/sdk-sample-generator/genaisrc/src/go/typecheck.ts
```diff
@@ -40,39 +40,35 @@ export async function typecheckGo(
...
         }
 
-        // Tidy up the Go module
-        const tidyResult = await container.exec("go", ["mod", "tidy"], {
-            cwd: projectDir,
-        });
+        const commands = [
+            ["go", "install", "github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.3.1"],
+            ["go", "build"],
+            ["go", "test", "-c"],
```
deyaaeldeen, line 47: Does this run the code? I don't think we're ready for that yet.

richardpark-msft: No, `-c` means "just compile". `go build` is only going to build the non-test files, so it'll skip over example files.

tools/sdk-sample-generator/genaisrc/src/go/typecheck.ts
richardpark-msft: It can - `go install` actually downloads and compiles the source code, so if there are weird errors there's a possibility of troubleshooting. 

Though, TBH, I wasn't thinking about it like that, I was just trying to arrange all the commands together...

deyaaeldeen (approved):

PR #11673, opened 2025-08-14
[Evals] Default to multiple evaluation runs

mccoyp:
While investigating our variance on evaluation runs, I noticed that the script was only running evals once in pipelines. It looks like the default number of runs was changed to 1, possibly for local testing, when it was originally `NUM_RUNS`. This reverts the change.


kristapratico (approved):

laiapat:
Closing because of refactors to the eval structure

PR #11685, opened 2025-08-14
[azsdkcli] add checks for readme and cspell

l0lawrence:
Adding more common checks - waiting for new process helper to merge + will refactor


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
```diff
@@ -168,4 +193,168 @@ protected async Task<CLICheckResponse> ValidateChangelogCommonAsync(string packa
...
+            }
+        }
+        catch (Exception ex)
+        {
+            stopwatch.Stop();
+            return new CLICheckResponse(1, "", $"Unhandled exception: {ex.Message}");
+        }
+        finally
+        {
+            await Task.CompletedTask; // Make this async for consistency
```
l0lawrence: this doesnt seem like this does anything at all

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
```diff
@@ -168,4 +193,168 @@ protected async Task<CLICheckResponse> ValidateChangelogCommonAsync(string packa
...
+            if (!Directory.Exists(packagePath))
+            {
+                return new CLICheckResponse(1, "", $"Package path does not exist: {packagePath}");
+            }
+
+            // Find the SDK repository root by looking for common repository indicators
+            var packageRepoRoot = _gitHelper.DiscoverRepoRoot(packagePath);
+            if (string.IsNullOrEmpty(packageRepoRoot))
+            {
+
```
benbp:
```suggestion
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
```diff
@@ -168,4 +193,168 @@ protected async Task<CLICheckResponse> ValidateChangelogCommonAsync(string packa
...
+            }
+
+            // Get the target branch (usually main or master) for the repository
+            var targetBranch = _gitHelper.GetBranchName(packageRepoRoot);
+            if (string.IsNullOrEmpty(targetBranch))
+            {
+                targetBranch = "main"; // fallback to main if we can't determine the default branch
+            }
+
+            var command = "pwsh";
```
benbp: I'm thinking I should add a powershell helper that extends proc helper

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/CheckAllToolTests.cs
```diff
@@ -78,14 +78,38 @@ public async Task RunPackageCheck_WithChangelogCheck_ReturnsResult()
...
+        [Test]
+        public async Task RunPackageCheck_WithReadmeCheck_ReturnsResult()
+        {
+            // Act
+            var result = await _packageCheckTool.RunPackageCheck(_testProjectPath, PackageCheckType.Readme);
+
+            // Assert
+            Assert.IsNotNull(result);
+            // README check may succeed or fail depending on directory contents
+            Assert.That(result.ExitCode, Is.GreaterThanOrEqualTo(0));
```
benbp: Does this test anything? Copilot loves these "pass if we didn't completely crash the prcoess" type tests :D

l0lawrence: yah it does not

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/CheckAllToolTests.cs
```diff
@@ -78,14 +78,34 @@ public async Task RunPackageCheck_WithChangelogCheck_ReturnsResult()
...
+        }
+
+        [Test]
+        public async Task RunPackageCheck_WithSpellingCheck_ReturnsResult()
+        {
+            // Act
+            var result = await _packageCheckTool.RunPackageCheck(_testProjectPath, PackageCheckType.Cspell);
+
+            // Assert
+            Assert.IsNotNull(result);
```
l0lawrence, line 107: I feel like it could be worth keeping these for future unit testing updates

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageRepoService.cs
```diff
@@ -10,8 +10,8 @@ namespace Azure.Sdk.Tools.Cli.Services;
 /// </summary>
 public class JavaScriptLanguageRepoService : LanguageRepoService
 {
-    public JavaScriptLanguageRepoService(IProcessHelper processHelper, IGitHelper gitHelper) 
-        : base(processHelper, gitHelper)
+    public JavaScriptLanguageRepoService(IProcessHelper processHelper, IGitHelper gitHelper, INpxHelper npxHelper)
```
richardpark-msft: Perhaps we should look into this object hierarchy here. Why should child implementations be forced to inject a dependency they don't use?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
```diff
@@ -56,11 +70,13 @@ public class LanguageRepoService : ILanguageRepoService
 {
     protected readonly IProcessHelper _processHelper;
     protected readonly IGitHelper _gitHelper;
+    protected readonly INpxHelper _npxHelper;
```
richardpark-msft, line 81: These shouldn't be protected. If child classes need it they can declare it in their constructor. Otherwise it's a hidden dependency and that'll make testing less fun.

richardpark-msft: (Ignore this comment, I want to look a bit more to see what can even be done)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageRepoService.cs
timovv: Yeah it feels like inheritance might not be the best pattern here. Maybe LanguageRepoService should be an interface, and only contain functions that are truly language-specific. Any concrete code in LanguageRepoService could be extracted into another service for common cross-language operations. That way each LanguageRepoService and the common service can each request the dependencies they actually need.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/PackageCheckType.cs
```diff
@@ -3,7 +3,7 @@ namespace Azure.Sdk.Tools.Cli.Models;
 /// <summary>
 /// Defines the types of validation checks that can be run on SDK packages.
 /// </summary>
-public enum PackageCheckName
+public enum PackageCheckType
```
timovv, line 6: Are we going back to this name or was this just Copilot doing Copilot things

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageRepoService.cs
richardpark-msft: This'll involve some cleanup, but it seems best. We could even split out what's in LanguageServer into separate smaller services as well. 

The extra cost (minimal) is that you'd have to make the call to the contained class, yourself, but that's a small price to pay.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
richardpark-msft: Okay, I looked into it, and I think we'll need to go with something like what @timovv said here: https://github.com/Azure/azure-sdk-tools/pull/11685#discussion_r2283587147

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/PackageCheckType.cs
l0lawrence: Going back to this name re feedback on last pr

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageRepoService.cs
timovv: I think we can address this in a follow up PR

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
```diff
@@ -153,23 +176,117 @@ protected async Task<CLICheckResponse> ValidateChangelogCommonAsync(string packa
...
+
+            var command = "pwsh";
+            var args = new[] { 
+                "-File", scriptPath, 
+                "-SettingsPath", settingsPath,
+                "-ScanPaths", packagePath,
+            };
+
+            // Use a longer timeout for README validation - 10 minutes should be sufficient as it may need to install doc-warden
+            var timeout = TimeSpan.FromMilliseconds(600_000); // 10 minutes
```
richardpark-msft:
```suggestion
            var timeout = TimeSpan.FromMinutes(10);
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
```diff
@@ -153,23 +176,117 @@ protected async Task<CLICheckResponse> ValidateChangelogCommonAsync(string packa
...
+                return new CLICheckResponse(1, "", $"Package path does not exist: {packagePath}");
+            }
+
+            // Find the SDK repository root by looking for common repository indicators
+            var packageRepoRoot = _gitHelper.DiscoverRepoRoot(packagePath);
+            if (string.IsNullOrEmpty(packageRepoRoot))
+            {
+                return new CLICheckResponse(1, "", $"Could not find repository root from package path: {packagePath}");
+            }
+
```
richardpark-msft: Might as well make this a function - you're doing similar logic in a couple of spots.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
```diff
@@ -153,23 +176,117 @@ protected async Task<CLICheckResponse> ValidateChangelogCommonAsync(string packa
...
+            // Convert absolute path to relative path from repo root
+            var relativePath = Path.GetRelativePath(packageRepoRoot, packagePath);
+
+            // Build arguments for npx cspell lint
+            var args = new[] {
+                "cspell", "lint",
+                "--config", cspellConfigPath,
+                "--root", packageRepoRoot,
+                $"./{relativePath}/**"
```
richardpark-msft: Are those '/''s path separators? If so, you'll need to use Path.DirectorySeparatorChar.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
```diff
@@ -153,23 +176,117 @@ protected async Task<CLICheckResponse> ValidateChangelogCommonAsync(string packa
...
+            };
+
+            var processResult = await _processHelper.Run(new("npx", args, workingDirectory: packageRepoRoot), ct: default);
```
richardpark-msft: @timovv, for JS based commands, like this one, should we just recommend `npm install`ing this thing as a prerequisite or is 'npx' the right fit here? (I have no preference, just never occurred to me how often we use it)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
```diff
@@ -153,23 +176,117 @@ protected async Task<CLICheckResponse> ValidateChangelogCommonAsync(string packa
...
+            return CreateResponseFromProcessResult(processResult);
         }
         catch (Exception ex)
         {
-            stopwatch.Stop();
             return new CLICheckResponse(1, "", $"Unhandled exception: {ex.Message}");
```
richardpark-msft, line 299: Sometimes the stack trace is useful, especially if it's something we did that's broken. Do you mind _logger.LogError(ex) the full exception, just so we have something to use for diagnostics?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
```diff
@@ -153,23 +176,117 @@ protected async Task<CLICheckResponse> ValidateChangelogCommonAsync(string packa
...
         }
         finally
```
richardpark-msft: I can't seem to leave a comment on the correct line, but what does this line mean?

```csharp
finally {
  await Task.CompletedTask; // Make this async for consistency
}
```

Should this finally even exist? Isn't this just the default behavior?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
timovv: `npx` is fine. Under the hood it just installs the package if it isn't installed yet. That only happens the first time.

The only caveat is that if it does need to install something, npx asks interactively whether it should install the package. I don't think it's smart enough to figure out it's running in a non-interactive subprocess so we might need to change this to `npx -y ...` which bypasses that check.

l0lawrence: oh so true let me add that

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
l0lawrence: it should be removed this was pre- processHelper being an async class

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageRepoService.cs
richardpark-msft: Perhaps we should make the helper do that, that's a pitfall just waiting to happen.

(scratch this for now, I swore we had an NpxHelper class already)

timovv: I think the npxhelper was in and got backed out somehow, either that or we're both hallucinating

richardpark-msft: For a bit @l0lawrence was randomly deleting code, but I think we should bring that one back to avoid issues like this one.

richardpark-msft: @timovv, @l0lawrence: as usual, @chrisradek has really good code tucked away in his own PR that he refuses to share: https://github.com/Azure/azure-sdk-tools/pull/11628/files#diff-cfa866013827d7d2e5ff30443f91da9593664d510fc02005f3f84cd86cd28cb6

But once that's done, we should definitely kill our NPX related code in favor of that helper.

chrisradek: @richardpark-msft You can already use the npx helper:
https://github.com/Azure/azure-sdk-tools/blob/b7f5662c46b340dc57421be587eb59e0a5f56723/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/TspTool.cs#L174-L190

richardpark-msft: @chrisradek, I apologize, you're absolutely right. Let me rewrite my comment to take your feedback into account!

@l0lawrence, you might as well just change your code over now.

richardpark-msft (approved):

PR #11688, opened 2025-08-15
[tsp-client] Fix emitter-output-dir support bug

catalinaperalta:
always pass repo root when using emitter-output-dir. All specs that use the output-dir variable to build the absolute path with the emitter-output-dir variable will do so based off of the repo root.

### DIFF
--- a/tools/tsp-client/src/commands.ts
+++ b/tools/tsp-client/src/commands.ts
@@ -445,7 +445,7 @@ export async function generateCommand(argv: any) {
   }
   const result = await compileTsp({
     emitterPackage: emitter,
-    outputPath: outputDir,
+    outputPath: legacyPathResolution ? outputDir : await getRepoRoot(outputDir), // always use repo root when using emitter-output-dir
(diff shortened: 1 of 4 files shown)


tools/tsp-client/src/commands.ts
```diff
@@ -445,7 +445,7 @@ export async function generateCommand(argv: any) {
   }
   const result = await compileTsp({
     emitterPackage: emitter,
-    outputPath: outputDir,
+    outputPath: legacyPathResolution ? outputDir : await getRepoRoot(outputDir), // always use repo root when using emitter-output-dir
```
JoshLove-msft: nit: can we cache the result getRepoRoot? I see it is called [here](https://github.com/Azure/azure-sdk-tools/pull/11688/files#diff-dd1c27612030a2ec40031d890bfbe89776b0025252f7d881306a4fd42136fa68R400) as well.

JoshLove-msft (approved):

Issue #11129, opened 2025-07-08
[APIView Copilot] Copilot needs separate staging and production resources


tjprescott:
Currently copilot is just a single service, and all flavors of APIView (production, staging, UX test) depend on it. This creates problems when something on APIView staging breaks and API contract and Copilot is updated to match, because not APIView production will be impacted!

We should follow the best practice to have separate resources. It's fine if UX test and staging use the same resources, but production should be a separate set of resources. 

Related to: https://github.com/Azure/azure-sdk-tools/issues/10592

tjprescott:
Complete!

Issue #10592, opened 2025-05-09
[APIView Copilot] Migrate resources from Playground to Engineering Systems


tjprescott:
Currently, all resources are located in the Azure SDK Playground subscription. They need to be migrated to the "Azure SDK Engineering System" subscription.

Additionally, we should create separate production and staging resources, as described in https://github.com/Azure/azure-sdk-tools/issues/11129

I need to add the following to existing scripts:
- [x] Azure AI Foundry
- [x] Azure AI Foundry Project

And need to integrate the following to follow best practices:
- [x] KeyVault
- [x] App Configuration

tjprescott:
Complete!

PR #11693, opened 2025-08-15
[AVC] Update to use AppConfiguration and KeyVault instead of webapp settings

tjprescott:
Related to #11129 and #10592.

To make copilot follow best practices for production deployment, we need to use AppConfiguration and KeyVault to store configuration over simple app settings (and .env files). The PR facilitates that by creating a `SettingsManager` singleton that can query AppConfiguration, and, if needed, KeyVault. Retrieves settings are cached and the cache can be invalidated to force refresh from AppConfig at runtime.


packages/python-packages/apiview-copilot/evals/README.md
```diff
@@ -6,16 +6,7 @@ This directory contains the evaluation testing for APIView Copilot.
...
-  ```
-  AZURE_OPENAI_ENDPOINT=<your-endpoint>
-  AZURE_OPENAI_API_KEY=<your-key>
-  AZURE_SUBSCRIPTION_ID=<playground-subscription-id>
-  AZURE_FOUNDRY_RESOURCE_GROUP=openai-shared
-  AZURE_FOUNDRY_PROJECT_NAME=apiview-ai
-  ```
-
-> Note: the API key is needed for AI-assisted evaluations. The service does not support Entra ID yet.
+- Environment variables set up in a `.env` file as shown in the ApiView Copilot README.
```
laiapat, line 9: The ApiView Copilot README only seems to mention these environment variables:

- AZURE_APP_CONFIG_ENDPOINT
- ENVIRONMENT_NAME
- OPENAI_ENDPOINT

But evals additionally need:

- OPENAI_API_KEY
- EVALS_SUBSCRIPTION
- EVALS_RG
- EVALS_PROJECT_NAME

packages/python-packages/apiview-copilot/evals/run.py
```diff
@@ -1,37 +1,35 @@
...
+import sys
+from typing import Any, Set, Tuple
+
 import prompty
 import prompty.azure_beta
-import copy
-import sys
 import yaml
 
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
laiapat: Could we slightly simplify this?
```suggestion
sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
```

packages/python-packages/apiview-copilot/evals/run.py
```diff
@@ -45,12 +43,23 @@ def in_ci():
     return os.getenv("TF_BUILD", False)
 
 
+def get_model_config():
```
laiapat: Could we use a global `SettingsManager` instance and model config instead of this getter, like is done in `apiview-copilot/app.py`?

packages/python-packages/apiview-copilot/src/_settings.py
```diff
@@ -0,0 +1,92 @@
...
+        else:
+            cache_key = (key, self.label)
+            self._cache.pop(cache_key, None)
+
+    def _get_secret_from_keyvault(self, secret_uri):
+        # Parse the Key Vault URI
+        # Example: https://<vault-name>.vault.azure.net/secrets/<secret-name>/<version>
+        from urllib.parse import urlparse
+
+        parsed = urlparse(secret_uri)
```
laiapat: There's a [KeyVaultSecretIdentifier](https://github.com/Azure/azure-sdk-for-python/blob/de7adba1a14362ec582cc02ea4830c3d4ecaf4ab/sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py#L251) class that can help with this parsing

packages/python-packages/apiview-copilot/evals/README.md
tjprescott: These are all set in the AppConfiguration (or KeyVault), so the only environment variables you need are those two.

packages/python-packages/apiview-copilot/src/_settings.py
tjprescott: Nice! Thank you!

tjprescott: Updated

packages/python-packages/apiview-copilot/evals/run.py
tjprescott: Updated

packages/python-packages/apiview-copilot/evals/run.py
tjprescott: Updated

PR #11698, opened 2025-08-15
[PerfAutomation] Add advanced statistics to perf automation

jalauzon-msft:
This change adds a new flag `--advanced-stats` to PerfAutomation which triggers a workflow to track more advanced latency statistics for perf test runs. If enabled, this flag will instruct the perf tests it is invoking to track latencies and size (if present) for each individual operation and write those results to a temporary file allocated by PerfAutomation. PerfAutomation will then gather those results from all iterations and generate latency metrics for each test and add them to the final `results.json` output.

The per-language tests must be updated in order to work with the new `--advanced-stats` flag. For now, updating .NET and Python with this support here:
.NET - https://github.com/Azure/azure-sdk-for-net/pull/52038
Python - https://github.com/Azure/azure-sdk-for-python/pull/42554

### DIFF
--- a/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Program.cs
+++ b/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Program.cs
@@ -419,6 +419,8 @@ namespace Azure.Sdk.Tools.PerfAutomation
                                 }
 
                                 result.Iterations.Add(iterationResult);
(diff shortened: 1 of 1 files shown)


jalauzon-msft:
PerfAutomation run with Python and `--advanced-stats` passed:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5231751&view=results

mikeharder (approved):

PR #11717, opened 2025-08-18
Speed up azsdk-cli CI checks by only testing linux and not producing executables

benbp:
Speed up azsdk-cli CI checks by only testing linux and not producing executables. This brings CI times down from ~5-10 minutes to ~2-3 minutes.


richardpark-msft:
@benbp, the most common platform will probably end up being Windows, and I've already seen some "used the wrong path sep" bugs in some other programs in our ecosystem. So we should probably still leave in a Windows platform, or even just test on Windows since we have a lot of developers running Linux.

benbp:
Going to move CI to github actions so we can keep testing windows but have quick turnaround.

PR #11728, opened 2025-08-19
added logic for the Agent to fix error

radhgupta:
1. Added `ErrorFixerAgent.FixCodeAsync `

- Sequential Fix Processing: Processes a list of Azure SDK code fixes one by one, maintaining conversation state so each fix builds upon previous changes
- Converts each fix to a prompt, sends it to the Azure AI agent, waits for completion, and extracts the updated TypeSpec content from the response
- Uses semaphore-based concurrency control and configurable delays between fixes to prevent API throttling
- Implements fail-fast strategy with comprehensive logging, stopping on first failure while tracking progress ({Current}/{Total}) (TODO)
- Returns the final updated client.tsp content after successfully applying all fixes in the conversation thread

2. Added `AgentResponseParser.cs`

- Uses pre-compiled regex patterns to extract TypeSpec content from various code block formats (typespec, tsp, typescript, generic) in priority order. we use typescript here because I have noticed that sometimes (rarely) agent response with typescript code block
- Selects the last matching code block when multiple are present, ensuring the most recent/final content is captured from AI agent responses

### DIFF
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/AgentResponseParser.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/AgentResponseParser.cs
@@ -1,0 +1,93 @@
+using System.Text.RegularExpressions;
+using Azure.Tools.GeneratorAgent.Models;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Tools.GeneratorAgent
+{
+    /// <summary>
+    /// Service for parsing and processing AI agent responses from TypeSpec code blocks
+    /// </summary>
+    internal class AgentResponseParser
+    {
+        private readonly ILogger<AgentResponseParser> Logger;
+        private static readonly Regex TypeSpecCodeBlockRegex = new(@"```typespec\s*\n(.*?)\n```", RegexOptions.Singleline | RegexOptions.IgnoreCase | RegexOptions.Compiled);
+        private static readonly Regex TspCodeBlockRegex = new(@"```tsp\s*\n(.*?)\n```", RegexOptions.Singleline | RegexOptions.IgnoreCase | RegexOptions.Compiled);
+        private static readonly Regex TypeScriptCodeBlockRegex = new(@"```typescript\s*\n(.*?)\n```", RegexOptions.Singleline | RegexOptions.IgnoreCase | RegexOptions.Compiled);
+        private static readonly Regex GenericCodeBlockRegex = new(@"```\s*\n(.*?)\n```", RegexOptions.Singleline | RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+        public AgentResponseParser(ILogger<AgentResponseParser> logger)
+        {
+            ArgumentNullException.ThrowIfNull(logger);
+            Logger = logger;
+        }
+
+        /// <summary>
+        /// Parses the agent's response and extracts the updated client.tsp content from TypeSpec code blocks
+        /// </summary>
+        public AgentResponse ParseResponse(string rawResponse)
+        {
+            try
+            {
+                string updatedContent = ExtractTypeSpecCodeBlock(rawResponse);
+                
+                Logger.LogInformation("Successfully parsed agent response with {Length} characters of client.tsp content", 
+                    updatedContent.Length);
+                
+                return new AgentResponse
+                {
+                    UpdatedFileContent = updatedContent.Trim(),
+                    HasValidContent = true
+                };
+            }
+            catch (Exception ex) when (!(ex is OperationCanceledException))
+            {
+                Logger.LogError(ex, "Unexpected error while parsing agent response. Response length: {Length}", 
+                    rawResponse?.Length ?? 0);
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// Extracts TypeSpec content from code blocks in the response
+        /// </summary>
+        private string ExtractTypeSpecCodeBlock(string response)
+        {
+            try
+            {
+                // Try pre-compiled regex patterns in order of preference
+                (Regex, string)[] regexPatterns = new[]
+                {
+                    (TypeSpecCodeBlockRegex, "typespec"),        // ```typespec ... ``` (primary - correct format)
+                    (TspCodeBlockRegex, "tsp"),                  // ```tsp ... ``` (file extension)
+                    (TypeScriptCodeBlockRegex, "typescript"),    // ```typescript ... ``` (legacy fallback)
+                    (GenericCodeBlockRegex, "generic")           // ``` ... ``` (no language)
+                };
+
+                foreach ((Regex regex, string patternName) in regexPatterns)
+                {
+                    MatchCollection matches = regex.Matches(response);
+                    
+                    if (matches.Count > 0)
+                    {
+                        Match lastMatch = matches[matches.Count - 1];
+                        string extractedContent = lastMatch.Groups[1].Value.Trim();
+                        
+                        if (!string.IsNullOrWhiteSpace(extractedContent))
+                        {
+                            Logger.LogDebug("Successfully extracted content using {PatternName} pattern", patternName);
+                            return extractedContent.Trim();
+                        }
+                    }
+                }
+
+                Logger.LogWarning("No code blocks found in response using any pattern");
+                throw new InvalidOperationException("No valid TypeSpec code blocks found in agent response");
+            }
+            catch (Exception ex) when (!(ex is OperationCanceledException))
+            {
+                Logger.LogError(ex, "Error extracting TypeSpec code block from response");
+                throw;
+            }
+        }
+    }
+}
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
@@ -15,8 +15,7 @@ namespace Azure.Tools.GeneratorAgent
     {
         // Regular expression to match error patterns.
         // Matches strings like "error CS0103: The name 'InvalidVariable' does not exist in the current context [/path/to/file.cs]" into groups: "error (CS0103): (The name 'InvalidVariable' does not exist in the current context) [/path/to/file.cs]"
-
-        private static readonly Regex ErrorRegex = new(@"error\s+([A-Z]+\d+):\s*(.+?)(?=\s*\[|$)", RegexOptions.IgnoreCase | RegexOptions.Multiline | RegexOptions.Compiled);
+        private static readonly Regex ErrorRegex = new(@"error\s+([A-Z]+\d+):\s*(.+?)(?=\s*\[|$)", RegexOptions.IgnoreCase | RegexOptions.Multiline | RegexOptions.Compiled, TimeSpan.FromSeconds(1));
 
         private readonly ILogger<BuildErrorAnalyzer> Logger;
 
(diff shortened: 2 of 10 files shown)


src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Models/AgentResponse.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        /// <summary>
+        /// The complete updated client.tsp content extracted from TypeSpec code blocks
+        /// </summary>
+        public string UpdatedFileContent { get; set; } = string.Empty;
+        
+        /// <summary>
+        /// Indicates whether the response contained valid TypeSpec content
+        /// </summary>
+        public bool HasValidContent { get; set; } = false;
```
radhgupta: TODO: Add logic to check if the Updated client.tsp is valid and set this variable.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/AgentResponseParser.cs
```diff
@@ -0,0 +1,93 @@
...
+{
+    /// <summary>
+    /// Service for parsing and processing AI agent responses from TypeSpec code blocks
+    /// </summary>
+    internal class AgentResponseParser
+    {
+        private readonly ILogger<AgentResponseParser> Logger;
+        private static readonly Regex TypeSpecCodeBlockRegex = new(@"```typespec\s*\n(.*?)\n```", RegexOptions.Singleline | RegexOptions.IgnoreCase | RegexOptions.Compiled);
+        private static readonly Regex TspCodeBlockRegex = new(@"```tsp\s*\n(.*?)\n```", RegexOptions.Singleline | RegexOptions.IgnoreCase | RegexOptions.Compiled);
+        private static readonly Regex TypeScriptCodeBlockRegex = new(@"```typescript\s*\n(.*?)\n```", RegexOptions.Singleline | RegexOptions.IgnoreCase | RegexOptions.Compiled);
```
radhgupta: While testing I have noticed that there are certain RARE instances where the agent response has a typescript code block with client.tsp content.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/AgentResponseParser.cs
```diff
@@ -0,0 +1,93 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Tools.GeneratorAgent
+{
+    /// <summary>
+    /// Service for parsing and processing AI agent responses from TypeSpec code blocks
+    /// </summary>
+    internal class AgentResponseParser
+    {
+        private readonly ILogger<AgentResponseParser> Logger;
```
jsquire, line 12: If we have to RegEx agent responses, then I think we should consider refactoring the prompt and telling it how to create structured output.   We do not want to be in the business of doing text parsing - that's why we're using an LLM to start with.   😄

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildErrorAnalyzer.cs
```diff
@@ -112,17 +111,16 @@ public List<RuleError> ParseBuildOutput(ProcessExecutionException processExcepti
                     string errorType = match.Groups[1].Value.Trim();
                     string errorMessage = match.Groups[2].Value.Trim();
 
-                    if (string.IsNullOrWhiteSpace(errorType) || string.IsNullOrWhiteSpace(errorMessage))
+                    if (errorType.AsSpan().IsWhiteSpace() || errorMessage.AsSpan().IsWhiteSpace())
```
jsquire, line 114: That's a neat efficiency.   I didn't realize that was a thing.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -53,9 +53,9 @@ private static IServiceCollection AddCredentialServices(this IServiceCollection
             services.AddSingleton<CredentialFactory>();
             services.AddSingleton<TokenCredential>(provider =>
             {
-                var credentialFactory = provider.GetRequiredService<CredentialFactory>();
```
jsquire: image

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,20 +60,171 @@ internal virtual PersistentAgent CreateAgent()
...
+        public async Task<string> FixCodeAsync(List<Fix> fixes, string threadId, CancellationToken cancellationToken = default)
+        {            
+            Logger.LogInformation("Starting code fix process with {Count} fixes using thread {ThreadId}", fixes.Count, threadId);
+
+            string? finalUpdatedContent = null;
+            int processedCount = 0;
+
+            try
+            {
+                await ConcurrencyLimiter.WaitAsync(cancellationToken).ConfigureAwait(false);
```
jsquire: We could probably use some comments around the approach that you're using here.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Models/AgentResponse.cs
```diff
@@ -0,0 +1,22 @@
+using System.Text.Json.Serialization;
+
+namespace Azure.Tools.GeneratorAgent.Models
+{
+    /// <summary>
+    /// Represents the parsed response from the AI agent containing updated TypeSpec content
+    /// </summary>
+    public class AgentResponse
```
m-redding:
```suggestion
    public class AgentTypeSpecResponse
```
nit but just an idea

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -74,7 +73,7 @@ private static IServiceCollection AddHttpClientServices(this IServiceCollection
                 client.DefaultRequestHeaders.UserAgent.Clear();
                 client.DefaultRequestHeaders.UserAgent.ParseAdd("AzureSDK-TypeSpecGenerator/1.0");
 
-                string? githubToken = EnvironmentVariables.GitHubToken;
+                var githubToken = EnvironmentVariables.GitHubToken;
```
m-redding, line 76: Curious why change all these to var?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
```diff
@@ -44,6 +44,18 @@ public AppSettings(IConfiguration configuration, ILogger<AppSettings> logger)
         public TimeSpan VectorStoreReadyWaitTime => TimeSpan.FromMilliseconds(
             int.Parse(Configuration.GetSection("AzureSettings:VectorStoreReadyWaitTimeMs").Value ?? "5000"));
         
+        // Agent run settings
+        public TimeSpan? AgentRunMaxWaitTime => Configuration.GetSection("AzureSettings:AgentRunMaxWaitTimeMinutes").Value != null
```
m-redding: Why not add defaults for these ones?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
```diff
@@ -51,31 +60,205 @@ internal virtual PersistentAgent CreateAgent()
...
+        {            
+            Logger.LogInformation("Starting code fix process with {Count} fixes using thread {ThreadId}", fixes.Count, threadId);
+
+            var finalUpdatedContent = (string?)null;
+            var processedCount = 0;
+
+            try
+            {
+                // Ensure only one fix process runs at a time to maintain thread safety
+                await ConcurrencyLimiter.WaitAsync(cancellationToken).ConfigureAwait(false);
```
m-redding: Is this just a precaution? It doesn't seem like this code is ever run in parallel since each call is awaited in a for loop. It seems like it might be overkill if it's never called in parallel? Unless I'm misunderstanding something

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
jsquire: all-the-things-meme

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
radhgupta: Var it is!! Yeahhyy!!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorFixerAgent.cs
radhgupta: It was an overkill :( I removed it!
Thanks for highlighting it :)

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
m-redding: Anne always made me un var all the things so I never know 😆

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/FixPromptService.cs
```diff
@@ -0,0 +1,81 @@
...
+            ArgumentNullException.ThrowIfNull(fix);
+
+            if (fix is AgentPromptFix promptFix)
+            {
+                return BuildAgentPromptFixMessage(promptFix);
+            }
+            else
+            {
+                return BuildGenericFixMessage(fix);
+            }
```
m-redding, line 48: nit but I think this could be simplified quite a bit, since both fixes are the same format just with different context and fixInstruction. I feel like the helpers aren't needed here

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/FixPromptService.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        /// <summary>
+        /// Builds a formatted message for generic Fix types
+        /// </summary>
+        private string BuildGenericFixMessage(Fix fix)
+        {
+            string fixTypeName = fix switch
+            {
+                AgentPromptFix => nameof(AgentPromptFix),
```
m-redding, line 71: Aren't the AgentPromptFixes getting built by BuildAgentPromptFixMessage?

m-redding (approved):

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/FixPromptService.cs
radhgupta: Fixing it in the upcoming PR.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/FixPromptService.cs
radhgupta: Yes, I removed the switch. I will update it in the next PR.

PR #11733, opened 2025-08-19
Generalize spec pull request tool

### LINKED ISSUE #11447: Restructure CreatePullRequest MCP tool for repo-agnostic use
The CreatePullRequest MCP tool is currently tightly coupled to the azure-rest-api-specs repository via the SpecPullRequestTools type. This limits its ability to be reused for other repositories or general pull request automation workflows.

**Proposal:**
- Refactor the SpecPullRequestTools type and related logic to be more generic, adopting a structure similar to PullRequestTools that is not specific to the spec repo.
- Consider any additional changes needed to make the tool fully independent for use across different repositories and scenarios.

jeo02:
## Summary
- Generalized spec pull request by removing hard coded specs repo information.
- Removed repo owner and repo name params as they can be retrieved internally instead.
- Moved the `CheckIfSpecInPublicRepo` to the spec validation tool as it is used to see if we are in the spec repo which is more in line with validation than pull request. 
- Resolves #11447


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/SpecValidationTool/SpecValidationTool.cs
```diff
@@ -75,6 +77,22 @@ public IList<string> RunTypeSpecValidation(string typeSpecProjectRootPath)
             }
         }
 
+        [McpServerTool(Name = "azsdk_check_typespec_project_in_public_repo"), Description("Check if TypeSpec project is in public spec repo. Provide absolute path to TypeSpec project root as param.")]
+        public string CheckTypeSpecProjectInPublicRepo(string typeSpecProjectPath)
```
praveenkuttappan: Can you put this MCP tool in a new class called SpecRepoValidationTool.cs? We are trying to change the structure where only one MCP tool resides in one class.

PR #11745, opened 2025-08-20
Fix wiql issue in getting release plan work items

praveenkuttappan:
Fix the issue in select statement when fetching release plans.


maririos (approved):
oops. is there any way to avoid these problems in the future? a test

weshaggard (approved):

Issue #11644, opened 2025-08-12
azsdkcli cli-checks updates needed


l0lawrence:
- [ ] Figure out unit testing/best practices: https://github.com/Azure/azure-sdk-tools/pull/11432/files#r2263695203
- [ ] Documentation on response parameters/Discussing parsing tool output and best ways to output it - do we want tools to just be a black box for the user?: https://github.com/Azure/azure-sdk-tools/pull/11432/files#r2270766511
- [ ] Also related to above, for parsing tool output: https://github.com/Azure/azure-sdk-tools/pull/11432/files#r2271140956
- [ ] Nit Go Abstract method: https://github.com/Azure/azure-sdk-tools/pull/11432/files#r2270153425
- [ ] How will multiple workflows tie into the LanguageRepoService - do we want to differentiate?:  https://github.com/Azure/azure-sdk-tools/pull/11432/files#r2268196257
- [ ] On a similar note for the languageFactory -- https://github.com/Azure/azure-sdk-tools/pull/11432/files#r2270733786
- [x] Nit - make check type an enum: https://github.com/Azure/azure-sdk-tools/pull/11432/files#r2271133859
- [x] Do we make commands have a default value?: https://github.com/Azure/azure-sdk-tools/pull/11432/files#r2270743695
- [ ] Do we want to use DefaultCommandResponse? https://github.com/Azure/azure-sdk-tools/pull/11432/files#r2270758074
- [ ] What should we name the subcommand and how should we allow people to specify which checks they want to run? (`azsdk check links` vs `azsdk azsdk package run-checks --check-type checkLinks`)
- [x] https://github.com/Azure/azure-sdk-tools/pull/11685#discussion_r2283587147
- [ ] https://github.com/Azure/azure-sdk-tools/pull/11747#discussion_r2298642467
- [ ] Next steps working examples needed:   https://github.com/Azure/azure-sdk-tools/pull/11747#discussion_r2298650990

PR #11747, opened 2025-08-20
[azsdkcli] add next steps

l0lawrence:


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
```diff
@@ -15,26 +15,35 @@ public class Response
...
+        foreach (var nextStep in NextSteps ?? [])
+        {
+            messages.Add("[NEXT STEPS] " + nextStep);
         }
 
-        if (errors.Count > 0)
+        if (messages.Count > 0)
         {
-            value = string.Join(Environment.NewLine, errors);
+            value = string.Join(Environment.NewLine, messages);
```
l0lawrence, line 59: do we think the next steps may always be populated, we just only want it to output when the failure happens? In which case this would need to change

benbp: I think in both success and failure cases we should output next steps if it's populated, that way each sub-class doesn't have to stringify next steps.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
```diff
@@ -15,26 +15,35 @@ public class Response
...
         }
         foreach (var error in ResponseErrors ?? [])
         {
-            errors.Add("[ERROR] " + error);
+            messages.Add("[ERROR] " + error);
+        }
+
+        foreach (var nextStep in NextSteps ?? [])
+        {
+            messages.Add("[NEXT STEPS] " + nextStep);
```
praveenkuttappan: Can this be changed to output like: `NEXT STEPS]: <concatenated steps>

Currently it will look like below and that's confusing.
[NEXT STEPS]: <step 1>
[NEXT STEPS]: <step 2>

praveenkuttappan (approved):
Looks fine to me except a nit suggestion about NEXT STEPS output.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
l0lawrence: ah I can also make it a string and not a list?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
```diff
@@ -15,26 +15,35 @@ public class Response
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public List<string> ResponseErrors { get; set; }
 
+    [JsonPropertyName("next_steps")]
```
samvaity, line 27:
```suggestion
    [JsonPropertyName("nextStep")]
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
```diff
@@ -15,26 +15,35 @@ public class Response
...
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? NextSteps { get; set; }
```
samvaity: Why do we need it to be plural?
It is just going to be a blob of string.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
```diff
@@ -15,26 +15,35 @@ public class Response
...
+
     protected string ToString(StringBuilder value)
     {
         return ToString(value.ToString());
     }
 
     protected string ToString(string value)
     {
-        List<string> errors = [];
+        List<string> messages = [];
```
samvaity, line 38: nit: I think "errors" would be clearer here.

timovv (approved):
Looks good to me, I think this idea of having actionable steps to take is a good one and will help people through the process

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
timovv: Snake case is consistent with the other properties in this class, do we have a convention for camel case elsewhere?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
l0lawrence: nextSteps isnt necessarily only if there is an error

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
benbp: Our mcp tool names are snake case so I'm inclined to say all json properties also be snake case

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
benbp: Some of the more complex tool implementations do a lot of string building at various points. It seems easier for this to be a list rather than making the caller have to worry about string formatting when they append next steps multiple times?

benbp: @l0lawrence I think re: Praveen's earlier comment I had thought it would end up being something like:

```
if (NextSteps?.Count > 0)
{
    messages.Add("[NEXT STEPS]");
    foreach (var step in NextSteps)
    {
        messages.Add(step);
    }
}
```

maririos:
Understanding the flow: does this means that each tool will be responsible for managing the next steps?
or is there a "source of truth" on the steps that each tool references?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
richardpark-msft: What are next steps for then? We wouldn't drive a normal deterministic workflow using it, right?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
richardpark-msft: I think we need some doc comments on this before it gets checked in. Leaving this ambiguous just means people will fill it in, incorrectly.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
```diff
@@ -15,26 +15,39 @@ public class Response
...
+            {
+                messages.Add(step);
+            }
         }
 
-        if (errors.Count > 0)
+        if (messages.Count > 0)
         {
-            value = string.Join(Environment.NewLine, errors);
+            value = string.Join(Environment.NewLine, messages);
```
richardpark-msft, line 59: Let's change this over to use a StringBuilder instead.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
```diff
@@ -15,26 +15,39 @@ public class Response
...
+            messages.Add("[ERROR] " + error);
+        }
+
+        if (NextSteps?.Count > 0)
+{
+            messages.Add("[NEXT STEPS]");
+            foreach (var step in NextSteps)
+            {
+                messages.Add(step);
+            }
```
samvaity, line 54: I do have concerns about this. I think we need semantics to be explicit so the MCP can decide whether something is just a suggestion or a deterministic action to run. Overloading the string structure might cause confusions with the tooling hints.

In my opinion, we can enhance the structure to include more details if we to use structured `NextStep.Deterministic == true` for deterministic internal flows (the MCP client can auto-run those small commands).
or `Deterministic == false (or absent)` for suggestions the LLM/agent can evaluate further.

richardpark-msft:
I'm approving, but @samvaity's comments point to us wanting to have some more workable examples to look at. 

Given the examples from the other chat, it looks like this is primarily about enhancing error messages, or (sometimes) trying to get the user to a next step that has docs and (probably) stuff they need to read. From talking with @l0lawrence, I think the main goal is that we blanket the user with all the info they need to get from one step to another, and stuff like this can keep them in the chat session (if that's where they started), or the CLI, if they work there.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
l0lawrence: next pr

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Response.cs
l0lawrence: linked https://github.com/Azure/azure-sdk-tools/issues/11644#issue-3316095866

PR #11779, opened 2025-08-21
refactor --check-type and --package-path and service initialization for language repos

l0lawrence:


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/ILanguageSpecificChecks.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Gets the language this implementation supports.
+    /// </summary>
+    string SupportedLanguage { get; }
+
+    /// <summary>
+    /// Determines if this implementation can handle the given package path.
+    /// </summary>
+    /// <param name="packagePath">Path to the package directory</param>
+    /// <returns>True if this implementation can handle the package, false otherwise</returns>
+    bool CanHandle(string packagePath);
```
timovv: I wonder if this should be the responsibility of this interface or not. If we create other language-specific interfaces (e.g. for tests?) we'd have to implement `CanHandle` in each one, and the implementation would look roughly the same.

Maybe this check could be at a higher level like in the resolver?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        var repositoryPath = _gitHelper.DiscoverRepoRoot(packagePath);
+
+        // Get the repository name from the directory path
+        var repoName = Path.GetFileName(repositoryPath?.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar))?.ToLowerInvariant() ?? "";
+
+        _logger.LogInformation($"Repository name: {repoName}");
+
+        // Extract the language from the repository name
+        if (repoName.Contains("azure-sdk-for-java"))
```
timovv: I think we might be doing this elsewhere too, but getting the repo name from the repo path is not a reliable way of detecting the language. The folder name could be different. For example, when creating a Codespace the repo is cloned into a folder called `workspace` regardless of the name of the repo the Codespace is created from.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageSpecificCheckResolver.cs
```diff
@@ -0,0 +1,61 @@
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Sdk.Tools.Cli.Services;
+
+/// <summary>
+/// Resolves the appropriate language-specific check implementation based on package contents.
+/// Uses composition pattern instead of inheritance.
+/// </summary>
+public class LanguageSpecificCheckResolver
```
timovv: Should this service have an interface too like the other services do so we can mock it

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
```diff
@@ -6,108 +6,31 @@
...
 {
     protected readonly IProcessHelper _processHelper;
     protected readonly INpxHelper _npxHelper;
     protected readonly IGitHelper _gitHelper;
-    protected readonly ILogger<LanguageRepoService> _logger;
+    protected readonly ILogger<LanguageChecks> _logger;
+    protected readonly LanguageSpecificCheckResolver _languageSpecificCheckResolver;
 
-    public LanguageRepoService(IProcessHelper processHelper, INpxHelper npxHelper, IGitHelper gitHelper, ILogger<LanguageRepoService> logger)
+    public LanguageChecks(IProcessHelper processHelper, INpxHelper npxHelper, IGitHelper gitHelper, ILogger<LanguageChecks> logger, LanguageSpecificCheckResolver languageSpecificCheckResolver)
```
timovv: Maybe this should have an interface too

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ChecksHelper.cs
```diff
@@ -0,0 +1,19 @@
...
+/// Helper class for check operations.
+/// </summary>
+public static class ChecksHelper
+{
+    /// <summary>
+    /// Creates a CLICheckResponse from a process result.
+    /// </summary>
+    /// <param name="processResult">The process result to convert</param>
+    /// <returns>CLI check response</returns>
+    public static CLICheckResponse CreateResponseFromProcessResult(ProcessResult processResult)
```
timovv: Maybe this method could be on `CLICheckResponse` so that we don't need a separate class here

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        var repositoryPath = _gitHelper.DiscoverRepoRoot(packagePath);
+
+        // Get the repository name from the directory path
+        var repoName = Path.GetFileName(repositoryPath?.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar))?.ToLowerInvariant() ?? "";
+
+        _logger.LogInformation($"Repository name: {repoName}");
+
+        // Extract the language from the repository name
+        if (repoName.Contains("azure-sdk-for-dotnet"))
```
timovv: `azure-sdk-for-net`?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
```diff
@@ -6,108 +6,31 @@
...
-/// </summary>
-public class LanguageRepoService : ILanguageRepoService
+public class LanguageChecks 
 {
     protected readonly IProcessHelper _processHelper;
     protected readonly INpxHelper _npxHelper;
     protected readonly IGitHelper _gitHelper;
-    protected readonly ILogger<LanguageRepoService> _logger;
+    protected readonly ILogger<LanguageChecks> _logger;
+    protected readonly LanguageSpecificCheckResolver _languageSpecificCheckResolver;
```
timovv: Can these be private now?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        _processHelper = processHelper;
+        _npxHelper = npxHelper;
+        _gitHelper = gitHelper;
+        _logger = logger;
+    }
+
+    public string SupportedLanguage => "Java";
+
+    public bool CanHandle(string packagePath)
```
timovv: This logic seems to be repeated, perhaps we could extract it into a common helper?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
l0lawrence: good point, I had previously done language_settings and then the readme at the root.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
l0lawrence: Why?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
l0lawrence: Let me try

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
l0lawrence: ah for mocking

timovv: Also for consistency since all the other services have interfaces. It's not super high priority, can always just add the interface when we get around to writing tests that need it, it's just doing that gets more annoying the more things we have depending on it

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedOptions.cs
```diff
@@ -63,9 +64,9 @@ public static class SharedOptions
             IsRequired = false,
         };
 
-        public static Option<string> PackagePath = new(["--package-path", "-p"], "Path to the package directory to check")
+        public static Option<string> PackagePath = new(["--package-path", "-p"], () => Environment.CurrentDirectory, "Path to the package directory to check")
```
timovv:
```suggestion
        public static Option<string> PackagePath = new(["--package-path", "-p"], () => Environment.CurrentDirectory, "Path to the package directory to check. Defaults to the current working directory")
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedOptions.cs
```diff
@@ -63,9 +64,9 @@ public static class SharedOptions
...
         {
-            IsRequired = true
+            IsRequired = false
```
timovv, line 69: I think `IsRequired` defaults to false, if so this could be removed

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageSpecificCheckResolver.cs
```diff
@@ -0,0 +1,155 @@
...
+            return null;
+        }
+    }
+
+    private async Task<string?> ExtractLanguageFromFileAsync(string filePath)
+    {
+        try
+        {
+            var options = new PowershellOptions([". '" + filePath.Replace("'", "''") + "'; Write-Output $Language"], 
+                workingDirectory: Path.GetDirectoryName(filePath));
```
l0lawrence, line 115: I am going full circle back to to this because imo this is the most reliable way to get the language and we are already calling pwsh scripts elsewhere

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
```diff
@@ -0,0 +1,38 @@
...
+        IGitHelper gitHelper,
+        ILogger<DotNetLanguageSpecificChecks> logger)
+    {
+        _processHelper = processHelper;
+        _npxHelper = npxHelper;
+        _gitHelper = gitHelper;
+        _logger = logger;
+    }
+
+    public string SupportedLanguage => "Dotnet";
```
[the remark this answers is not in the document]
timovv: It's `dotnet`: https://github.com/Azure/azure-sdk-for-net/blob/main/eng/scripts/Language-Settings.ps1#L1

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedOptions.cs
l0lawrence: leaving for rn as its in the other options

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CLICheckResponse.cs
```diff
@@ -25,6 +26,16 @@ public CLICheckResponse(int exitCode, string checkStatusDetails, string error =
         }
     }
 
+    /// <summary>
+    /// Creates a CLICheckResponse from a process result.
+    /// </summary>
+    /// <param name="processResult">The process result to convert</param>
+    /// <returns>CLI check response</returns>
+    public static CLICheckResponse CreateResponseFromProcessResult(ProcessResult processResult)
```
benbp: This could just be a constructor overload instead of a static method.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageSpecificChecks.cs
```diff
@@ -0,0 +1,163 @@
...
+        try
+        {
+            var compilerExists = (await _processHelper.Run(new ProcessOptions(compilerName, ["version"], compilerNameWindows, ["version"]), ct)).ExitCode == 0;
+            var linterExists = (await _processHelper.Run(new ProcessOptions(linterName, ["--version"], linterNameWindows, ["--version"]), ct)).ExitCode == 0;
+            var formatterExists = (await _processHelper.Run(new ProcessOptions("echo", ["package main", "|", formatterName]), ct)).ExitCode == 0;
+            return compilerExists && linterExists && formatterExists;
+        }
+        catch (Exception)
+        {
+            return false;
```
benbp, line 53: I assume there are some cases where the failure may be notable (e.g. a timeout), do we want to at least log debug?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageSpecificChecks.cs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public string GetSDKPackagePath(string repo, string packagePath)
+    {
+        if (!repo.EndsWith(Path.DirectorySeparatorChar))
+        {
+            repo += Path.DirectorySeparatorChar;
+        }
+
+        // ex: sdk/messaging/azservicebus
```
benbp:
```suggestion
        // ex: sdk/messaging/azservicebus/
```

Given the above wouldn't it be this?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageSpecificCheckResolver.cs
```diff
@@ -0,0 +1,155 @@
...
+        IEnumerable<ILanguageSpecificChecks> languageChecks,
+        IGitHelper gitHelper,
+        IPowershellHelper powershellHelper,
+        ILogger<LanguageSpecificCheckResolver> logger)
+    {
+        _languageChecks = languageChecks ?? throw new ArgumentNullException(nameof(languageChecks));
+        _gitHelper = gitHelper ?? throw new ArgumentNullException(nameof(gitHelper));
+        _powershellHelper = powershellHelper ?? throw new ArgumentNullException(nameof(powershellHelper));
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
```
benbp: Could be simplified a little to use primary constructor, and we don't need the null checks we can rely on DI.

```suggestion
public class LanguageSpecificCheckResolver(
    IEnumerable<ILanguageSpecificChecks> _languageChecks,
    IGitHelper _gitHelper,
    IPowershellHelper _powershellHelper,
    ILogger<LanguageSpecificCheckResolver> _logger
) : ILanguageSpecificCheckResolver
{
```

timovv (approved):

PR #11780, opened 2025-08-21
[AVC] Migrate create_resources.py script from private repo

tjprescott:
This adds the `create_resources.py` script that was used to stand up all Azure resources for APIView Copilot. This script was previously stored in a private repo.

### DIFF
--- a/packages/python-packages/apiview-copilot/scripts/infra/_variables.py
+++ b/packages/python-packages/apiview-copilot/scripts/infra/_variables.py
@@ -6,69 +6,61 @@ from yaml import safe_load
 dotenv.load_dotenv(override=True)
 
 
+# List of variables that require the '-staging' suffix in the staging environment
+VARIABLES_WITH_STAGING_SUFFIX = {
+    "APP_CONFIGURATION_NAME",
+    "APP_SERVICE_PLAN_NAME",
+    "COGNITIVE_SERVICES_NAME",
+    "COSMOS_ACCOUNT_NAME",
+    "KEYVAULT_NAME",
+    "RG_NAME",
+    "SEARCH_INDEX_NAME",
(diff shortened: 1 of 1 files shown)


packages/python-packages/apiview-copilot/scripts/infra/_variables.py
```diff
@@ -0,0 +1,90 @@
...
+
+dotenv.load_dotenv(override=True)
+
+
+class Variables:
+    def __init__(self, *, is_staging: bool, path: str = "variables.yaml"):
+        # verify that the file exists
+        if not os.path.isfile(path):
+            raise FileNotFoundError(f"Configuration file not found: {path}")
+        with open(path, "r", encoding="utf-8") as stream:
```
laiapat: Nit: it will incur another indentation, but I would do a `try/except` here since it would error straight away and be faster on the happy path
```
        try:
           with open(path, "r", encoding="utf-8") as stream:
                ...
        except FileNotFoundError as ex:
            raise FileNotFoundError(f"Configuration file not found: {path}") from ex
```

packages/python-packages/apiview-copilot/scripts/infra/_variables.py
```diff
@@ -0,0 +1,90 @@
...
+            data = safe_load(stream)
+            variables = [
```
laiapat: Nit: it would be cleaner and easier to maintain if we define these (and the set of variables that shouldn't get the `"-staging"` modification) in a global variable

packages/python-packages/apiview-copilot/scripts/infra/_variables.py
tjprescott: Updated

PR #11796, opened 2025-08-22
Changing the readme tool to use Timo's MicroAgent framework

richardpark-msft:
Also, some refactors to cleanup the internals of the generator. There's more work to do to allow it to support multiple languages, but one big change at a time. :)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
```diff
@@ -151,57 +170,95 @@ async Task<bool> Generate(ChatClient chatClient, string templatePath, Uri servic
...
+                    FuncAdapterTool<ReadmeContents, CheckReadmeResult>.Create("check_readme_tool", "Checks a readme to make sure that all the required values have been replaced", CheckReadme),
+                ]
+            }, ct);
+
+            // to guard against hallucinations, make sure this readme _truly_ is correct.
+            var checkReadmeResult = await CheckReadme(result, ct);
+
+            if (checkReadmeResult.Suggestions.Any())
+            {
+                throw new Exception(string.Join(Environment.NewLine, checkReadmeResult.Suggestions));
```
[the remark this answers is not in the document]
richardpark-msft: In this particular case there's no value in having a specific exception (nobody would handle it, and it's treated generically when handled and it doesn't "escape" outside of the boundaries we expect).

richardpark-msft:
@timovv, I nominate you to review this particular one, as it hits a few areas that I know you care about:

1. Making the MCPTool type have less code in it
2. Using Micro**A**gents
3. I added a small "adapter" so you can create a tool that's just a function. It's a small little wrapper, but I'm hoping to reduce boilerplate/excess classes if possible. I use it in my readme generator to allow my check function to run.
4. I believe I've broken down my prompt and the return values the way you were hoping it'd work, where I have a function returning suggestions back to the LLM, etc.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/FuncAdapterTool.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <param name="description">The description of your tool.</param>
+    /// <param name="invoke">The function that will be invoked, when this tool is used.</param>
+    public class FuncAdapterTool<ToolInputT, ToolOutputT>(string name, string description, Func<ToolInputT, CancellationToken, Task<ToolOutputT>> invoke) : AgentTool<ToolInputT, ToolOutputT>
+    {
+        private readonly Func<ToolInputT, CancellationToken, Task<ToolOutputT>> invoke = invoke;
+
+        public override string Name { get; init; } = name;
+        public override string Description { get; init; } = description;
+
+        public static AgentTool<ToolInputT, ToolOutputT> Create(string name, string description, Func<ToolInputT, CancellationToken, Task<ToolOutputT>> invokeHandler)
```
timovv: What's the point of the factory function if it's just calling the public constructor? Should the constructor be private?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
```diff
@@ -151,57 +162,95 @@ async Task<bool> Generate(ChatClient chatClient, string templatePath, Uri servic
...
+            }
+
+            if (parameters.Contents.Contains("(package path)"))
+            {
+                suggestions.Add("The readme contains '(package path)', which is a placeholder that should be removed and replaced with the proper package path");
+            }
+
+            if (parameters.Contents.Contains("en-us"))
+            {
+                suggestions.Add("The readme contains links with en-us. Keep the link, but remove en-us.");
```
benbp: I wouldn't put it past the model to include other languages like `/en-de` maybe we want to check for the regex `/en-../` instead?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/FuncAdapterTool.cs
```diff
@@ -0,0 +1,28 @@
+namespace Azure.Sdk.Tools.Cli.Microagents.Tools
```
benbp: Nice

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
```diff
@@ -123,26 +114,46 @@ public override async Task HandleCommand(InvocationContext ctx, CancellationToke
                 ctx.ExitCode = 1;
             }
         }
+    }
+
+    public partial class ReadmeGenerator
```
timovv, line 122: Why `partial`? Is this for testing?

timovv: Do we really need this class? Seems like all we do with it is `new` it up and then immediately call `Generate`, and it's immutable -- could we just put the method in the tool or even inline it?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/FuncAdapterTool.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Let's you create a tool that's just a function or method.
+    /// </summary>
+    /// <typeparam name="ToolInputT">The input fields for your tool.</typeparam>
+    /// <typeparam name="ToolOutputT">The output fields for your tool.</typeparam>
+    /// <param name="name">The name of your tool.</param>
+    /// <param name="description">The description of your tool.</param>
+    /// <param name="invoke">The function that will be invoked, when this tool is used.</param>
+    public class FuncAdapterTool<ToolInputT, ToolOutputT>(string name, string description, Func<ToolInputT, CancellationToken, Task<ToolOutputT>> invoke) : AgentTool<ToolInputT, ToolOutputT>
```
timovv: I like this idea, but as a nitpick, do you think it would be nicer if we put the static factory method on `AgentTool<T, U>` itself? We could make this class private in `AgentTool` as well since it's really an implementation detail at that point.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
timovv: IIRC the format is `<language>-<country>`, so we'd see things like `de-de`, `fr-ca`, etc., which makes it a bit broader. Would the model be smart enough if we just say "the readme contains links with a language code please remove them from the links" and provide an example of what doing that looks like?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
```diff
@@ -151,57 +162,95 @@ async Task<bool> Generate(ChatClient chatClient, string templatePath, Uri servic
...
-                    logger.LogInformation($"[PASS] Verify-Links.ps1 passed");
+                    suggestions.Add($"Some links were broken and should be replaced: {output}");
                 }
             }
-            catch (Exception ex)
+            finally
             {
-                logger.LogError(ex, "[ERROR] Verify-Links.ps1 threw an exception");
-                valid = false;
+                try { File.Delete(tempFile); } catch (Exception) { }
```
timovv: nit

```suggestion
                try { File.Delete(tempFile); } catch { }
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
```diff
@@ -151,57 +162,95 @@ async Task<bool> Generate(ChatClient chatClient, string templatePath, Uri servic
...
-                    valid = false;
-                }
-                else
+                if (output != null)
                 {
-                    logger.LogInformation($"[PASS] Verify-Links.ps1 passed");
+                    suggestions.Add($"Some links were broken and should be replaced: {output}");
                 }
             }
-            catch (Exception ex)
```
timovv, line 198: Do we still want to catch in addition to the `finally`?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
timovv: oh, I see, it's partial for the generated Regex magic thing. Is this why it's a separate class as well?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
```diff
@@ -151,57 +162,95 @@ async Task<bool> Generate(ChatClient chatClient, string templatePath, Uri servic
...
+            {
+                suggestions.Add("The readme contains '(package path)', which is a placeholder that should be removed and replaced with the proper package path");
+            }
+
+            if (parameters.Contents.Contains("en-us"))
+            {
+                suggestions.Add("The readme contains links with en-us. Keep the link, but remove en-us.");
+            }
+
+            if (parameters.Contents.Contains("aztemplate"))
```
praveenkuttappan: This is Go specific template package name. How about other language? Is there any template package for arm?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
richardpark-msft: Two reasons:

* It separates the "CLI + MCP boilerplate" from the "normal C# class" code
* Using the class+singlefunc means I can split out the single function call into a bunch of helpers without having to re-pass a bunch of parameters. This one's mostly style, but I prefer it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
richardpark-msft: Yeah, I have a next step to split this out so the validation logic is language specific.

richardpark-msft: (in the next PR)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
richardpark-msft: I think it's a simple regex. I can add that.

I could also say "constrain your output to English".

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/FuncAdapterTool.cs
richardpark-msft: Oh yeah, I can remove one of them. I was hoping to take advantage of some type inference to remove some more boilerplate, but I never got there. :)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/FuncAdapterTool.cs
richardpark-msft: Oh, I'd love that. Didn't want to mess with the core bits for this, but it makes a lot of sense.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
richardpark-msft: How does the MicroAgent handle a tool throwing an exception? Now I'm wondering...

timovv: At the moment any exception gets caught and the exception message is sent back to the LLM.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/FuncAdapterTool.cs
richardpark-msft: Done.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs
richardpark-msft: Okay, I think that's actually okay. The only kind of exception that would get thrown here would be a mistake on my part.

PR #11799, opened 2025-08-22
move test proxy package files to test-proxy directory

m-redding:


scbedd (approved):
I see no issue with this! Thanks for verifying @m-redding

PR #11803, opened 2025-08-22
added client rule analyzer and moved prompts to json

radhgupta:
## Client rules

| Rule Code | Analyzer                 | Scenario                   | Error Message                                                                                                                                                                                                  |
|----------:|--------------------------|---------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| AZC0012   | TypeNameAnalyzer         | Single word type names    | "Type name '{0}' is too generic and has high chance of collision with BCL types or types from other libraries. Consider using a more descriptive multi-word name, such as '{1}'."                                             |
| AZC0030   | GeneralSuffixAnalyzer    | Model naming suffix issues | "Model name '{0}' ends with '{1}'. We suggest renaming it to '{2}' or another name with this suffix."                                                                                                                  |
| AZC0034   | DuplicateTypeNameAnalyzer| Type name conflicts      | "Type name '{0}' conflicts with '{1}'. Consider renaming to '{2}' to avoid confusion."                                                                                                                                  |
| AZC0035   | ModelFactoryAnalyzer     | Missing model factory methods | "Output model type '{0}' should have a corresponding method in a model factory class. Add a static method that returns '{0}' to a class ending with 'ModelFactory'." 

- added the Prompt and Context for all Client Rule.
- This PR also moves the previously hardcoded prompt and context to JSON file.


src/dotnet/ErrorAnalyzers/Client/src/Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers/ErrorAnalyzers/AZC0030RuleAnalyzer.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers
+{
+    /// <summary>
+    /// Handles AZC0030: Model naming suffix issues.
+    /// </summary>
+    [DiscoverableAnalyzer]
+    internal sealed class AZC0030RuleAnalyzer : BasePromptAnalyzer
+    {
+        public AZC0030RuleAnalyzer() : base("AZC0030") { }
```
m-redding: Why do you need a separate class for each rule if the only difference is the rule id string? Would it work to do something like an enum with all the supported rules? Something like
```csharp
public enum AgentRuleAnalyzers
{
    AZC0012,
    AZC0030,
    AZC0034,
    AZC0035
}
```
And then one base class that manages the supported rules and prompts? I don't see where these are being used so I'm not totally sure if it would work, it just seems like a lot of code/maintenance to do it this way. The json prompts you implemented here are cool and could really simplify this part I think

radhgupta: good point, fixing it

src/dotnet/ErrorAnalyzers/Client/src/Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers/ClientRuleAnalyzerRegistry.cs
```diff
@@ -1,16 +1,44 @@
...
+            }
+
+            return analyzers;
+        }
+
+        /// <summary>
+        /// Creates a new analyzer instance for the specified rule type.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static AgentRuleAnalyzer CreateAnalyzer(AzcRuleType ruleType)
```
m-redding: Why create a helper for this one?

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/AnalyzerProviderBase.cs
```diff
@@ -15,18 +15,32 @@ public abstract class AnalyzerProviderBase : IAnalyzerProvider
...
+        /// Initializes a new instance of the <see cref="AnalyzerProviderBase"/> class.
+        /// </summary>
         protected AnalyzerProviderBase()
         {
             LazyAnalyzers = new Lazy<IReadOnlyList<AgentRuleAnalyzer>>(DiscoverAnalyzers);
         }
 
-        public IEnumerable<AgentRuleAnalyzer> GetAnalyzers() => LazyAnalyzers.Value;
+        /// <summary>
+        /// Gets the collection of analyzers provided by this provider.
```
m-redding: I'm a little confused about some of the structure here, because the actual analyzers that are looking at the code are defined elsewhere, and these aren't really analyzers right? They're more just like types to match a prompt for certain error messages?

I'm just struggling to wrap my head around the value of having all the different types - AgentRuleAnalyzer, AnalyzerProviderBase, IAnalyzerProvider, BastPromptAnalyzer, etc. Right now, if I understand correctly, they all work to be able to map an error code to a prompt to send the LLM? Is there a larger long-term goal than that?

src/dotnet/ErrorAnalyzers/Client/src/Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers/ClientRuleAnalyzerRegistry.cs
```diff
@@ -1,16 +1,45 @@
...
+        public static void RegisterClientRules()
+        {
+            ErrorAnalyzerService.RegisterRules(SupportedRules);
+        }
+
+        /// <summary>
+        /// Gets all supported client rule IDs.
+        /// Returns a cached array for optimal performance.
+        /// </summary>
+        public static ReadOnlySpan<string> GetSupportedRules() => SupportedRules;
```
jsquire: What's our goal here?   This seems like an aggressive optimization to `ReadOnlySpan<string>` which is going to be fairly limited due to the restrictions on ROS.    Should we just create one `string[]` property and initialize it to simplify?

Callers can always cast or `.AsSpan` when needed.   The only real benefit is eliding bounds checking, since the memory has already been allocated.

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/AgentPromptFix.cs
```diff
@@ -15,7 +13,7 @@ public AgentPromptFix(string prompt, string? context = null)
         {
             ArgumentException.ThrowIfNullOrWhiteSpace(prompt);
             Prompt = prompt;
-            Context = context;
+            Context = context ?? string.Empty;
```
jsquire: Why empty?   We generally want to use `null` to communicate "there is no value here" - in this case, that's the scenario, right?   There just isn't a context?

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/AnalyzerPromptProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        /// <summary>
+        /// Tries to get the prompt associated with the specified rule ID.
+        /// </summary>
+        internal static bool TryGetPrompt(string ruleId, out string prompt)
+        {
+            prompt = string.Empty;
+
+            if (string.IsNullOrWhiteSpace(ruleId))
```
jsquire:
```suggestion
            if (ruleId is { Length: 0 })
```
That'll do a "null or empty" check, which is all we need.   You could even choose to go just `ruleId is not null` and be fine.  The whitespace check is pretty heavy and you'd already get a miss on `TryGetValue` for a whitespace-only value.

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/AnalyzerPromptProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            return false;
+        }
+
+        /// <summary>
+        /// Tries to get the formatted context for the specified rule ID and error message.
+        /// </summary>
+        internal static bool TryGetContext(string ruleId, string errorMessage, out string context)
+        {
+            context = string.Empty;
```
jsquire: Convention says that this would be `null` if not found.   We generally don't see "" used as a "no value" marker outside of strange scenarios.

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/AnalyzerPrompts.json
```diff
@@ -0,0 +1,18 @@
+{
```
jsquire: We should consider moving this to a dictionary in the config.    You could do a partial class to keep the organization clean and avoid a giant file.

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/Azure.Tools.ErrorAnalyzers.csproj
```diff
@@ -5,9 +5,16 @@
...
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <SignAssembly>false</SignAssembly>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="9.0.0" />
   </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Include="AnalyzerPrompts.json">
```
jsquire: This is going to be pretty heavy.   If we can move this to settings, it'll be a much more efficient approach.   The benefits of using an external text file are lost when you embed it, since you cannot update it on the fly because the asset is baked into the assembly - just like code would be.

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/ErrorAnalyzerService.cs
```diff
@@ -14,80 +15,103 @@ namespace Azure.Tools.ErrorAnalyzers
...
+        private static volatile FrozenDictionary<string, RuleAnalyzer>? FrozenAnalyzers;
+        private static readonly object FreezeLock = new();
 
         /// <summary>
-        /// Attempts to generate fixes for multiple errors.
+        /// Registers analyzers for the specified rule IDs.
+        /// Thread-safe and optimized for high-performance lookups after registration.
         /// </summary>
-        public static IEnumerable<Fix> GetFixes(IEnumerable<RuleError> errors)
+        public static void RegisterRules(params string[] ruleIds)
```
jsquire: Unless you expect to call this with loose arguments like:

`ErrorAnalyzerService.RegisterRules("AZC001", "AZC002")`  

there really isn't a reason to use `params`.    That is just a bit of sugar that allows you to pass individual values in lieu of a collection.

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/ErrorAnalyzerService.cs
```diff
@@ -14,80 +15,103 @@ namespace Azure.Tools.ErrorAnalyzers
...
-            ArgumentNullException.ThrowIfNull(provider);
-            AnalyzerProviders.Add(provider);
+            // Fast path: return frozen dictionary if available
+            if (FrozenAnalyzers is not null)
+            {
+                return FrozenAnalyzers;
+            }
+
+            // Slow path: create frozen dictionary under lock
+            lock (FreezeLock)
```
jsquire: Is there any way that we can move this to the point of initialization so that we have a determinisic point?   Generally, you'd want this to be in the constructor so that the object is fully initialized upfront.

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Runtime.CompilerServices;
+
+// Allow internal types to be visible to Client, General, and Management analyzer projects
+[assembly: InternalsVisibleTo("Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers")]
```
jsquire: This tells us that we have things `internal` that should not be.   The only things that should need internal access are the tests and the dynamic proxy.

We should review the architectural patterns and see if we should make adjustments or make things `public`.

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/RuleAnalyzer.cs
```diff
@@ -0,0 +1,47 @@
...
+            ArgumentException.ThrowIfNullOrWhiteSpace(ruleId);
+            RuleId = ruleId;
+        }
+
+        public string RuleId { get; }
+
+        public bool CanFix(RuleError error)
+        {
+            ArgumentNullException.ThrowIfNull(error);
+            return string.Equals(error.type, RuleId, StringComparison.OrdinalIgnoreCase);
```
jsquire: This looks like it should be `error.Type`, I thinik.

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/RuleAnalyzer.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        public Fix? GetFix(RuleError error)
+        {
+            if (!CanFix(error))
+                return null;
+
+            return AnalyzerPromptProvider.TryGetPrompt(RuleId, out var prompt) &&
```
jsquire: It feels like these should be local to each other and not two separate calls.  We should talk over the design here.

src/dotnet/ErrorAnalyzers/Client/src/Azure.Tools.ErrorAnalyzers.ClientRuleAnalyzers/ClientRuleAnalyzerRegistry.cs
radhgupta: You're absolutely right! I initially chose ReadOnlySpan<string> for the theoretical performance benefits and immutability guarantees, but the practical limitations (can't store in fields, can't use in async methods, can't use with most LINQ) definitely outweigh the minimal gains for this Azure library - I've simplified to string[] and callers can always call .AsSpan() when they need span operations

src/dotnet/ErrorAnalyzers/Core/src/Azure.Tools.ErrorAnalyzers/AnalyzerPromptProvider.cs
radhgupta: fixed.

src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/AnalyzerPrompts.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// Adds general rule prompts to the builder.
+        /// Implemented in AnalyzerPrompts.General.cs
+        /// </summary>
+        static partial void AddGeneralPrompts(Dictionary<string, AgentPromptFix> builder);
+
+        /// <summary>
+        /// Adds management rule prompts to the builder.
+        /// Implemented in AnalyzerPrompts.Management.cs
+        /// </summary>
+        static partial void AddManagementPrompts(Dictionary<string, AgentPromptFix> builder);
```
m-redding: nit - I think the compiler should be able to call these methods from this file without needing the partial signatures unless I'm misunderstanding, I think you shoudl be able to remove these

src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/AnalyzerPrompts.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Tries to get the formatted context for the specified rule ID and error message.
+        /// Provided for backward compatibility with tests.
+        /// </summary>
+        internal static bool TryGetContext(string ruleId, string errorMessage, out string? context)
```
m-redding: It seems like this isn't being used anymore in favor of ErrorAnalyzerService.GetFix below? Or maybe the error analyzer service should delegate the context formatting to this method?

src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/AnalyzerPrompts.cs
radhgupta: Fixed!!

src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/AnalyzerPrompts.cs
radhgupta: Fixed!

PR #11818, opened 2025-08-25
[azsdk-cli] Prototype instrumentation for MCP tool calls

benbp:
This is a prototype of an approach to add some static instrumentation wrapper methods to the base `MCPTool` class, and replaces the mcp tool registration logic to call a telemetry wrapper service.

I haven't worked this into CLI mode yet because I want to do some higher level refactoring to simplify `GetCommand()` and `HandleCommand()` so we only have to wrap them once at the base.

In the future we will switch to the built-in middleware of the mcp c# sdk, but it is currently in development.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/HostServer/TelemetryWrapper.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/HostServer/TelemetryWrapper.cs
@@ -1,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using ModelContextProtocol.Protocol;
+using ModelContextProtocol.Server;
+using Azure.Sdk.Tools.Cli.Services;
+
+namespace Azure.Sdk.Tools.Cli.Tools;
+
+public class InstrumentedTool(ILogger logger, McpServerTool innerTool, string toolName) : DelegatingMcpServerTool(innerTool)
+{
+    private const int MaxInstrumentationUploadTime = 5;
+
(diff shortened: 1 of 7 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/HostServer/TelemetryWrapper.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using ModelContextProtocol.Protocol;
+using ModelContextProtocol.Server;
+using Azure.Sdk.Tools.Cli.Services;
+
+namespace Azure.Sdk.Tools.Cli.Tools;
+
+public class InstrumentedTool(ILogger logger, McpServerTool innerTool, string toolName) : DelegatingMcpServerTool(innerTool)
```
praveenkuttappan, line 9: Is this the middleware that when MCP tool is called? This is great.

praveenkuttappan (approved):
Thanks for doing this change

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/HostServer/TelemetryWrapper.cs
benbp: They don't have the middleware yet, but I found this class which is a convenient helper for building pass-through mcp tools that we can register manually with the service provider.

benbp: The real meat is in the service registration, i.e. we don't really need this class it just saves a little boilerplate.

PR #11839, opened 2025-08-26
Update Instruction tool names to snake case

jeo02:
## Summary
- Changed the tool names to the new snake case naming convention


weshaggard:
For my context is there a reason for the tool name convention change?

praveenkuttappan (approved):

jeo02:
> For my context is there a reason for the tool name convention change?
In this PR https://github.com/Azure/azure-sdk-tools/pull/11619 we changed all the names to the tools. Subsequently it changes how the tools are referred to by copilot. To make sure we are referring to the correct tool and not just copilots guess of what tool sounds similar enough we need to change the naming in the instructions as well.

jeo02:
> For my context is there a reason for the tool name convention change?

In this PR https://github.com/Azure/azure-sdk-tools/pull/11619 we changed all the names to the tools. Subsequently it changes how the tools are referred to by copilot. To make sure we are referring to the correct tool and not just copilots guess of what tool sounds similar enough we need to change the naming in the instructions as well.

maririos (approved):

benbp:
@maririos @jeo02 is it worth thinking about generating these instructions from templates, that way we can inject things that need to reference named properties? Or we can plug it into the tool selection tests once those get spun up.

Issue #11838, opened 2025-08-26
[APIView] Special handling for JS `@azure-rest` libraries


tjprescott:
Javascript's `@azure-rest` libraries are REST-level clients (RLC) that are thin wrappers around the REST API. As such, we don't review them against the JS guidelines because any change demanded by the guidelines would require a change to the REST API, which is not allowed.

Based on this:
1. The "Approve" button should _always_ be enabled for JS libraries that are in the `@azure-rest` namespace.
2. The "Generate Copilot Review" button should always be disabled for these same packages.

AlitzelMendez:
this also should be the case for management plane - dotnet (if it is easy to identify the management plane revisions)

PR #11852, opened 2025-08-26
[APIView] Special handling for JS @azure-rest libraries

AlitzelMendez:
closes: https://github.com/Azure/azure-sdk-tools/issues/11838


AlitzelMendez:
> Looks good to me, but curious why this logic is all in the JS SPA but the last PR seemed to be mostly in the C#?

last pull requests were like half and half, because we needed to get the values from the App Configuration so for that we used the server in C#, but still the filtering was happening in here. (https://github.com/Azure/azure-sdk-tools/pull/11808/files / https://github.com/Azure/azure-sdk-tools/pull/11627/files )

it looks like more changes in ts this time because I refactor the code a little bit so it looks more verbose now but I think is more clear and is going to be easier to adapt if we need to do the data plane changes. 

also the biggest thing is that I didn't add test for this part of the code in my previous pull request, I was adding more test in the logic of, is correctly returning that was already reviewed?) or more on that area, but now this logic is becoming more complex so I also added tests on this side :)

tjprescott (approved):

Issue #11178, opened 2025-07-11
[JS] Apply latest condition to tell if spec is model-only or not


wanlwanl:
https://github.com/Azure/azure-sdk-tools/blob/main/tools/js-sdk-release-tools/src/mlc/apiVersion/apiVersionTypeExtractor.ts#L42
We should add check if `src/api` (from github, not npm code) exists. If so, the client has operations; otherwise, it's model only.

FYI. refector and reuse https://github.com/Azure/azure-sdk-tools/blob/7ed72483c1c7117ffb9a4e7aab58a7a0701c136d/tools/js-sdk-release-tools/src/common/npmUtils.ts#L60

PR #11907, opened 2025-09-01
Apply latest condition to tell if spec is model-only or not

skywing918:
https://github.com/Azure/azure-sdk-tools/issues/11178

test pipeline:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5300355&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692


tools/js-sdk-release-tools/src/mlc/apiVersion/apiVersionTypeExtractor.ts
```diff
@@ -38,15 +41,39 @@ export const getApiVersionType: IApiVersionTypeExtractor = async (
...
+            packageRoot,
+            "src/api",
+            packageName,
+            stableVersion
+        );
+
+        if (!hasOperationsInGithub) {
+            logger.warn(`No parameters.ts found in ${packageRoot} and no src/api directory found in GitHub for ${packageName}, this is a model-only service`);
+            return true;
+        }
```
v-jiaodi, line 74: Since a condition has been added here, it can now cover the modular case and correctly determine whether it is modelOnly.

Issue #11656, opened 2025-08-13
Release plan should alway be communicated in terms of the PowerApp


maririos:
User shouldn't know that the release plan uses an Azure DevOps work item on the background. It shouldn't give the user the link or tell the user to go and modify it manually.
Example:

<img width="435" height="969" alt="Image" src="https://github.com/user-attachments/assets/cd2c2e31-786e-468d-b7b6-e76a5b3a8378" />

maririos:
@jeo02 I know you started to look into this a bit. Is there more work remaining?

jeo02:
@maririos Ah no it was a different problem all together. My problem was more on the side of the actual information the devops api was giving me in one of it's fields which lead to the api JSON. 

For this I assume it is grabbing/creating the link from the work item we return when creating a release plan since I believe we return the whole work item.

smw-ms:
<img width="413" height="840" alt="Image" src="https://github.com/user-attachments/assets/2957610e-b69c-45bb-aaf4-01e925bf15c4" />

Result after fix

smw-ms:
<img width="414" height="833" alt="Image" src="https://github.com/user-attachments/assets/a7583a14-cdba-49fb-816e-aae2f4db5028" />

maririos:
For the last image: https://github.com/Azure/azure-sdk-tools/issues/11656#issuecomment-3276653824

What is the value of:
`Release Plan Link: Open Release Plan`

maririos:
For the first image: https://github.com/Azure/azure-sdk-tools/issues/11656#issuecomment-3276387499
This is wrong :( I think Copilot might be getting this info from what we have in the MCP. 

The answer should be: `Open your release plan (link of release plan) and follow the guidance given there`

Could this be added to PR https://github.com/Azure/azure-sdk-tools/pull/11931?

smw-ms:
> For the last image: [#11656 (comment)](https://github.com/Azure/azure-sdk-tools/issues/11656#issuecomment-3276653824)
> 
> What is the value of: `Release Plan Link: Open Release Plan`

https://apps.powerapps.com/play/e/ca95f761-477a-4e98-b34e-dfd5dd03a770/a/3a732482-e479-4ec1-99c8-359f2c2d3f53?release-plan-id=8cafdd31-6b63-f011-bec3-000d3a5b5a24&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47

smw-ms:
[...]

PR #11931, opened 2025-09-03
Change create release plan method to hide work item details

smw-ms:


eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
```diff
@@ -1,5 +1,10 @@
 # Release Plan Creation Process
 You goal is to create a valid release plan. You must prompt user to provide all required information and all input must match the format and requirement mentioned in step 3 below.
+
+> **Important Note**
+> Do **not** ask the user to manually edit the DevOps work item.
+> To update a release plan manually, the user must use the [Release Planner Tool](https://apps.powerapps.com/play/e/ed2ffefd-774d-40dd-ab23-7fff01aeec9f/a/821ab569-ae60-420d-8264-d7b5d5ef734c?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47).
```
maririos:
```suggestion
> To update a release plan manually, the user must use the [Release Planner Tool](https://aka.ms/sdk-release-planner).
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -272,7 +272,22 @@ public async Task<string> CreateReleasePlan(string typeSpecProjectPath, string t
                 }
                 else
                 {
-                    return output.Format(workItem);
+                    if (workItem.Id is int workItemId)
```
maririos, line 284: the method description should be changed to remove the work item part. (I couldn't comment on the specific line)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -272,7 +272,22 @@ public async Task<string> CreateReleasePlan(string typeSpecProjectPath, string t
...
+                    {
+                        releasePlan.ReleasePlanId = releasePlanId;
+                    }
+
+                    if (workItem.Fields.TryGetValue("Custom.ReleasePlanLink", out value) && value is string releasePlanLink)
+                    {
+                        releasePlan.ReleasePlanLink = releasePlanLink;
+                    }
+
+                    return output.Format(releasePlan);
```
maririos: what is this returning? I see the method returns a `string` but `releasePlan` is an object and there is no overwrite to the `ToString`. or the LLM can get the object and from it use the property that matches?
are there other methods that were expecting the `workitemId` and now are going to be broken?

eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
```diff
@@ -1,5 +1,10 @@
 # Release Plan Creation Process
 You goal is to create a valid release plan. You must prompt user to provide all required information and all input must match the format and requirement mentioned in step 3 below.
+
+> **Important Note**
```
maririos: would it be worth to add something on the main instructions files about `when talking to a user always refer to the release plan with the release plan link and release plan ID. If it is for the LLM then use the release plan work item id`
but better?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
smw-ms: I doubt that any methods would be broken because the prior code only returned the string "Microsoft.TeamFoundation.WorkItemTracking.WebApi.Models.WorkItem" because like you said the ToString was not overwritten. This method with the change now returns "Azure.Sdk.Tools.Cli.Models.ReleasePlanDetails" for the same reasons. I can add an override method for ReleasePlanDetails in the next commit

eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
```diff
@@ -1,5 +1,13 @@
 # Release Plan Creation Process
 You goal is to create a valid release plan. You must prompt user to provide all required information and all input must match the format and requirement mentioned in step 3 below.
```
benbp:
```suggestion
Your goal is to create a valid release plan. You must prompt user to provide all required information and all input must match the format and requirement mentioned in step 3 below.
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/ReleasePlanDetails.cs
```diff
@@ -33,6 +33,46 @@ public class ReleasePlanDetails
...
+        {
+            var sdkInfoFormatted = SDKInfo != null && SDKInfo.Any()
+                ? string.Join("\n    ", SDKInfo.Select(s => s.ToString()))
+                : "None";
+
+            var specPullRequestsFormatted = SpecPullRequests != null && SpecPullRequests.Any()
+                ? string.Join(", ", SpecPullRequests)
+                : "None";
+
+            return $@"ReleasePlanDetails:
```
benbp: It's probably better to just return json at this point, that way we don't get potential drift if new properties are added to the class but not to `ToString()`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
benbp: All these methods should eventually be refactored to return object types implementing `Response` instead of string. `output.Format` just ends up serializing the object to json, but shouldn't be used directly in tools anymore.

smw-ms:
> Could you provide screenshots of this on the works? how is the output now? For example, before this change when the user requested the creation of the release plan, the answer from the agent was: image

https://github.com/Azure/azure-sdk-tools/issues/11656#issuecomment-3276387499
https://github.com/Azure/azure-sdk-tools/issues/11656#issuecomment-3276653824

eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
```diff
@@ -4,9 +4,7 @@ Your goal is to create a valid release plan. You must prompt user to provide all
 > **Important Note**
 > Do **not** ask the user to manually edit the DevOps work item.
 > To update a release plan manually, the user must use the [Release Planner Tool](https://aka.ms/sdk-release-planner).
-> When referring to a release plan in conversation with a user:
```
maririos: oh no. I am sorry. I didn't read we already had this. What you deleted is correct.

maririos: unless this works better

PR #11942, opened 2025-09-03
Fallback Mechanism for Unknown Error Types

radhgupta:
This PR implements a comprehensive fallback mechanism that ensures all errors receive AI-powered assistance, even when specific rule handlers don't exist. 

The implementation adds a universal __FALLBACK__ prompt in `AnalyzerPrompts.General.cs` that provides intelligent analysis for various error sources including TypeSpec compiler errors, .NET compiler errors custom analyzer errors. The fallback uses pattern recognition to identify error sources and provides targeted solutions with step-by-step instructions.

### DIFF
--- a/src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/AnalyzerPrompts.General.cs
+++ b/src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/AnalyzerPrompts.General.cs
@@ -15,22 +15,22 @@ namespace Azure.Tools.ErrorAnalyzers
         static void AddGeneralPrompts(Dictionary<string, AgentPromptFix> builder)
         {
             // Fallback prompt for unknown/unhandled errors from any source
-            builder["__FALLBACK__"] = new AgentPromptFix(
+            builder[FallbackRuleId] = new AgentPromptFix(
                 prompt: """
                 TASK: Analyze and Fix Unknown Error
 
                 You are analyzing an error that doesn't have a specific handler. This could be from various sources including:
                 • TypeSpec compiler errors
-                • .NET compiler errors (C#/VB.NET)
-                • Analyzer errors
+                • .NET compiler errors
+                • Azure Analyzer errors
                 • MSBuild errors
 
                 ANALYSIS APPROACH:
                 1. IDENTIFY ERROR SOURCE:
                    - Look at the error code pattern to determine the source:
                      * TypeSpec: Usually starts with letters (e.g., "invalid-template", "duplicate-declaration")
-                     * C# Compiler: Usually CS#### format (e.g., "CS0103", "CS1002")
-                     * Analyzer: Usually custom prefix + numbers (e.g., "AZC####")
+                     * .NET Compiler: Usually CS#### format (e.g., "CS0103", "CS1002")
+                     * Azure Analyzer: Usually custom prefix + numbers (e.g., "AZC####")
 
                 2. UNDERSTAND THE ERROR:
                    - Read the error message carefully to understand what's wrong
@@ -44,13 +44,13 @@ namespace Azure.Tools.ErrorAnalyzers
                    - Ensure proper imports and namespaces
                    - Check for TypeSpec-specific naming conventions
 
-                   For C# Compiler Errors:
+                   For .NET Compiler Errors:
                    - Check for missing using statements
                    - Verify syntax correctness
                    - Look for missing references or assemblies
                    - Check for type mismatches or accessibility issues
 
-                   For Analyzer Errors:
+                   For Azure Analyzer Errors:
                    - Follow the specific rule guidelines mentioned in the error
                    - Check for code style or best practice violations
                    - Look for Azure SDK guideline compliance issues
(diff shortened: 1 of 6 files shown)


src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/AnalyzerPrompts.General.cs
```diff
@@ -10,11 +10,72 @@ internal static partial class AnalyzerPrompts
...
         {
-            // TODO: Add general rule prompts here when they are defined
+            // Fallback prompt for unknown/unhandled errors from any source
+            builder["__FALLBACK__"] = new AgentPromptFix(
+                prompt: """
+                TASK: Analyze and Fix Unknown Error
+
+                You are analyzing an error that doesn't have a specific handler. This could be from various sources including:
+                • TypeSpec compiler errors
+                • .NET compiler errors (C#/VB.NET)
```
jsquire: You call it .NET compiler here and C# compiler on L32.   We probably want to normalize to ".NET compiler" and remove references to VB.   We don't emit any non-C# code.

src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/AnalyzerPrompts.General.cs
```diff
@@ -10,11 +10,72 @@ internal static partial class AnalyzerPrompts
...
+                • Analyzer errors
+                • MSBuild errors
+
+                ANALYSIS APPROACH:
+                1. IDENTIFY ERROR SOURCE:
+                   - Look at the error code pattern to determine the source:
+                     * TypeSpec: Usually starts with letters (e.g., "invalid-template", "duplicate-declaration")
+                     * C# Compiler: Usually CS#### format (e.g., "CS0103", "CS1002")
+                     * Analyzer: Usually custom prefix + numbers (e.g., "AZC####")
```
jsquire:
```suggestion
                     * Azure analyzer: Usually custom prefix + numbers (e.g., "AZC####")
```

Not sure how important the distinction is here, but there are built-in .NET analyzers that also run.  The AZC codes are specific to us.

src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/AnalyzerPrompts.cs
```diff
@@ -21,6 +21,12 @@ internal static partial class AnalyzerPrompts
         /// </summary>
         internal static bool TryGetPromptFix(string ruleId, out AgentPromptFix? fix)
         {
+            if (string.IsNullOrWhiteSpace(ruleId))
```
jsquire: Do we really need `IsWhitespace` here or can we just do the standard empty check and if you have a bunch of spaces in your ruleId, we just blow up because that's not a reasonable thing to do?

src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/ErrorAnalyzerService.cs
```diff
@@ -15,14 +15,22 @@ public static class ErrorAnalyzerService
...
 
-            if (!AnalyzerPrompts.TryGetPromptFix(error.type, out var fix) || fix is null)
+            AgentPromptFix? fix;
+
+            // Try to get specific fix for known rules
+            if (!AnalyzerPrompts.TryGetPromptFix(error.type, out fix) || fix is null)
             {
-                return null;
+                // Fall back to general error analysis for unknown types
+                if (!AnalyzerPrompts.TryGetPromptFix("__FALLBACK__", out fix) || fix is null)
```
jsquire: We should have the magic string abstracted somewhere so that we don't risk drift across the different uses.

src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/ErrorAnalyzerService.cs
```diff
@@ -68,7 +76,10 @@ public static IEnumerable<Fix> GetFixes(IEnumerable<RuleError> errors)
         public static IReadOnlyCollection<string> GetRegisteredRules()
         {
             // Return all available rule IDs from the compile-time dictionary
-            return AnalyzerPrompts.GetAllRuleIds();
+            // Exclude the fallback mechanism from the public API
+            return AnalyzerPrompts.GetAllRuleIds()
```
jsquire, line 67: Can we make the return `IEnumerable<string>` here and avoid the allocation?   Since we have a local data source and caller (nobody is crossing network boundaries here) and we cannot infer what the caller will do with the data, better to have the loose enumerable return and not force a new allocation on each call. 

If we really need a fully materialized set, we should cache it for future use.   (but without concurrency guards, because the race is benign and its less costly to create it twice than guard on each access)

src/dotnet/ErrorAnalyzers/src/Azure.Tools.ErrorAnalyzers/ErrorAnalyzerService.cs
```diff
@@ -15,14 +15,22 @@ public static class ErrorAnalyzerService
...
 
-            if (!AnalyzerPrompts.TryGetPromptFix(error.type, out var fix) || fix is null)
+            AgentPromptFix? fix;
+
+            // Try to get specific fix for known rules
+            if (!AnalyzerPrompts.TryGetPromptFix(error.type, out fix) || fix is null)
             {
-                return null;
+                // Fall back to general error analysis for unknown types
+                if (!AnalyzerPrompts.TryGetPromptFix(AnalyzerPrompts.FallbackRuleId, out fix) || fix is null)
```
m-redding: Instead of calling this twice what do you think about changing TryGetPromptFix to GetPromptFix and doing the fallback there instead?

radhgupta: Fixed!!

m-redding (approved):
Nice!

PR #11957, opened 2025-09-03
Fix json schema for update sdk details input so serialization does no…

praveenkuttappan:
Incoming JSON schema from GitHub copilot has `language` and `packageName` as properties and this causes failure when deserializing it and language and package name are empty for all objects.


scbedd (approved):

scbedd (approved):
Much better 👍

PR #11961, opened 2025-09-04
Improve userAgentInfo update

### LINKED ISSUE #11825: [JS] isModelOnly will disable the userAgentInfo update
## Affected Component 
 Bump Version for userAgentInfo in XXXContext.ts 
  

## Bug Description 
changelog tool failed to bump the version for  userAgentInfo in XXXContext.ts 
  

## Impact Scope 
all modular packages

## Expected Behavior 
update the version
  
## Root Cause Analysis
for modular to update UserAgent, there's a check to check if the library is a model only library
Image
but the check condition `src/rest/parameters.ts` will not be generated in modular
Image

so the check is always true

v-jiaodi:
fixes #11825

### DIFF
--- a/tools/js-sdk-release-tools/src/mlc/apiVersion/apiVersionTypeExtractor.ts
+++ b/tools/js-sdk-release-tools/src/mlc/apiVersion/apiVersionTypeExtractor.ts
@@ -39,14 +39,14 @@ export const getApiVersionType: IApiVersionTypeExtractor = async (
 
 export const isModelOnly: IModelOnlyChecker = async (packageRoot: string): Promise<boolean> => {
     // For MLC, simply check for parameters.ts - its absence indicates a model-only service
-    const parametersPath = join(packageRoot, "src/rest/parameters.ts");
+    const parametersPath = join(packageRoot, "src/api/index.ts");
     const isParametersExists = await exists(parametersPath);
     
     if (!isParametersExists) {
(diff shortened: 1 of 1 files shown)


v-jiaodi:
test pass log: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5299470&view=results
sdk pr https://github.com/Azure/azure-sdk-for-js/pull/35843
image

tools/js-sdk-release-tools/src/mlc/apiVersion/apiVersionTypeExtractor.ts
```diff
@@ -39,14 +39,14 @@ export const getApiVersionType: IApiVersionTypeExtractor = async (
 
 export const isModelOnly: IModelOnlyChecker = async (packageRoot: string): Promise<boolean> => {
     // For MLC, simply check for parameters.ts - its absence indicates a model-only service
-    const parametersPath = join(packageRoot, "src/rest/parameters.ts");
+    const parametersPath = join(packageRoot, "src/api/index.ts");
```
wanlwanl, line 42: I remember we only need to check if the folder exists? @kazrael2119  @MaryGao

v-jiaodi: For the modular case, the logic is the same here regardless of whether it is a folder or a file

PR #11973, opened 2025-09-04
Fail early on sandboxing pipeline when no APIView token file is generated

chidozieononiwu:
- this change makes it easier to identify where a failure occurs. It also makes it so the pipeline step that call APIView service is skipped when there is no artifact generated; this avoids a failure in APIView which is more difficult to diagnose.

### DIFF
--- a/eng/scripts/Create-Apiview-Token-Files.ps1
+++ b/eng/scripts/Create-Apiview-Token-Files.ps1
@@ -17,6 +17,7 @@ param (
 Write-Host "Review Details Json: $($ReviewDetailsJson)"
 $reviews = ConvertFrom-Json $ReviewDetailsJson
 Write-Host $reviews
+$wasApiviewTokenFileCreated = $false
 if ($reviews -ne $null)
 {
     foreach($r in $reviews)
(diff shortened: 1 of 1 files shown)


eng/scripts/Create-Apiview-Token-Files.ps1
```diff
@@ -31,20 +32,30 @@ if ($reviews -ne $null)
...
+
+        if ((Get-ChildItem -Path $CodeFileDirectory -Recurse | Measure-Object).Count -gt 0)
+        {
+            $wasApiviewTokenFileCreated = $true
+        }
+    }
+    if (-not $wasApiviewTokenFileCreated)
+    {
+        Write-Host "No Apiview Token files were created"
+        exit 1;
```
praveenkuttappan: This will be an issue when it processes more than one review. It will avoid processing further reviews in the request. This same script is used for batch update when Python API reviews are upgraded.

chidozieononiwu: I have changed it to devOps SucceededWithIssues, that way it can at least indicate where the initial error occurred.

PR #11999, opened 2025-09-08
[AVC] Add permission grant/revoke commands

### LINKED ISSUE #11995: [AVC] Add command to grant necessary permissions to logged-in user
At a minimum, the logged in user will need access to AppConfig to run various AVC commands, but may also need some permissions for CosmosDB and Search.

tjprescott:
Closes #11995.

### DIFF
--- a/packages/python-packages/apiview-copilot/src/_permissions.py
+++ b/packages/python-packages/apiview-copilot/src/_permissions.py
@@ -1,0 +1,215 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+"""
+Module for assigning Azure RBAC roles and permissions.
+"""
+import sys
+from typing import List, Literal, Optional
+from uuid import uuid4
+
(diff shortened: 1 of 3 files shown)


packages/python-packages/apiview-copilot/src/_permissions.py
```diff
@@ -0,0 +1,215 @@
...
+    cosmos_account_name: str,
+):
+    """Assigns special built-in roles for Cosmos DB."""
+    credential = get_credential()
+    cosmos_mgmt_client = CosmosDBManagementClient(credential, subscription_id)
+    cosmos_account = cosmos_mgmt_client.database_accounts.get(rg_name, cosmos_account_name)
+    role_name = None
+    role_id = None
+    # Assign well-known Role ID for "Cosmos DB Built-in Data Contributor"
+    if kind == "readWrite":
```
AlitzelMendez, line 125: do we want to being able to assign readWrite to users?

tjprescott: I think the helper method should enable it. In this PR I stopped short of actually allowing this for the CLI method. It simply uses readOnly permission.

AlitzelMendez (approved):

Issue #11954, opened 2025-09-03
[APIView] Add/adapt service API endpoints


AlitzelMendez:
This is a follow up of the issue https://github.com/Azure/azure-sdk-tools/issues/10945 and the pull request https://github.com/Azure/azure-sdk-tools/pull/11567

After analyzing this, there is the need to be able to call all the endpoints from the revision id, so we will do some modifications

### GET /getRevisionText?revisionId=
**Usage:** GET /getRevisionText?apiRevisionId=1234
**Parameters:**
  - reviewId (optional) - The review ID
  - apiRevisionId (optional) - Specific revision ID
  - selectionType (optional) - 
     -  Specific (default, requieres apiRevisionId)
     -  Latest
     -  LatestApproved
     - LatestManual

### GET /getComments?revisionId=
For getcomments we will probably need another endpoint, as there is a functionality relying in what is returning right now 
Additionally the functionality of the getComments is not what we actually need, we need to get all the "visible get comments" which means: unresolved comments of the elementIds present

PR #12012, opened 2025-09-09
 [APIView] Add/adapt service API endpoints

AlitzelMendez:
closes : https://github.com/Azure/azure-sdk-tools/issues/11954


src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsController.cs
```diff
@@ -81,6 +81,23 @@ public async Task<ActionResult<IEnumerable<CommentItemModel>>> GetConversationIn
...
+        /// <param name="commentType"></param>
+        /// <returns></returns>
+        [HttpGet("{reviewId}", Name = "GetComments")]
+        public async Task<ActionResult<IEnumerable<CommentItemModel>>> GetCommentsAsync(string reviewId,
+            bool isDeleted = false, CommentType? commentType = null)
+        {
+            IEnumerable<CommentItemModel> comments =
+                await _commentsManager.GetCommentsAsync(reviewId, isDeleted, commentType);
+            return new LeanJsonResult(comments, StatusCodes.Status200OK);
+        }
```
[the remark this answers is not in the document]
AlitzelMendez: this is true, that it is the exact duplication because I moved the code from the hybrid to internal as this specific functionality, we don't want it on a shared space :)

src/dotnet/APIView/APIViewWeb/LeanControllers/APIRevisionsTokenAuthController.cs
```diff
@@ -56,23 +56,28 @@ public async Task<ActionResult<APIRevisionListItemModel>> GetOutlineAPIRevisionA
     /// <param name="apiRevisionId">The specific API revision ID (required when selectionType is Specific)</param>
     /// <param name="selectionType">How to select the API revision</param>
     /// <returns>Plain text representation of the API review</returns>
-    [HttpGet("{reviewId}/getRevisionText", Name = "GetAPIRevisionText")]
+    [HttpGet("getRevisionText", Name = "GetAPIRevisionText")]
     public async Task<ActionResult<string>> GetAPIRevisionTextAsync(
-        string reviewId,
         [FromQuery] string apiRevisionId = null,
+        [FromQuery] string reviewId = null,
```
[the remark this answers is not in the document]
AlitzelMendez: this was decided by design but open to discuss it, we can have two endpoints for this if that makes things cleaner and works for everyone? @tjprescott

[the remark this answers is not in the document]
tjprescott: I'm not really sure what "selectionType" is but if revision ID is globally unique, then reviewID should never be needed. The only reason reviewId would be needed is if you wanted to specify the revision through some "tag" rather than it's ID (for example: last approved GA, last approved preview, latest, etc.)

tjprescott (changes requested):
Can you include the updated TypeSpec for these endpoints?

src/dotnet/APIView/APIViewUnitTests/APIRevisionsControllerTests.cs
```diff
@@ -71,9 +71,7 @@ public async Task GetAPIRevisionTextAsync_SpecificType_WithValidId_ReturnsRevisi
             .Setup(x => x.GetApiRevisionText(expectedRevision))
             .ReturnsAsync(expectedText);
 
-        ActionResult<string> result = await _controller.GetAPIRevisionTextAsync(
-            reviewId,
-            apiRevisionId);
+        ActionResult<string> result = await _controller.GetAPIRevisionTextAsync(apiRevisionId, reviewId);
```
tjprescott, line 74: My understanding is that revision id is globally unique. If that's true, review id should not be required.

src/dotnet/APIView/APIViewWeb/LeanControllers/APIRevisionsTokenAuthController.cs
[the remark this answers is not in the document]
AlitzelMendez: yes, the selectionType is for last approved GA, latest, latest manual, those scenarios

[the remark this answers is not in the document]
tjprescott: Yeah, I think this is fine.

src/dotnet/APIView/APIViewUnitTests/APIRevisionsControllerTests.cs
AlitzelMendez: it is not required, it is optional for scenarios like latest 

default is http://localhost:5000/api/apirevisions/getRevisionText?apiRevisionId=2668702652604dad92d75feb1321c7a4


the reviewId is to get 
 Latest
LatestApproved
LatestManual

on the original defintion this was described but is this really needed? or do we want just to get rid of that?

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsController.cs
```diff
@@ -81,6 +81,23 @@ public async Task<ActionResult<IEnumerable<CommentItemModel>>> GetConversationIn
...
+        /// <summary>
+        ///     Retrieve comments for a review.
+        /// </summary>
+        /// <param name="reviewId"></param>
+        /// <param name="isDeleted"></param>
+        /// <param name="commentType"></param>
+        /// <returns></returns>
+        [HttpGet("{reviewId}", Name = "GetComments")]
+        public async Task<ActionResult<IEnumerable<CommentItemModel>>> GetCommentsAsync(string reviewId,
+            bool isDeleted = false, CommentType? commentType = null)
```
tjprescott, line 94: This should also point to a revision, not a review id. Conceptually it should surface any comments from the review that are "visible" to the target revision (i.e. there's a matching line Id in the target revision).

AlitzelMendez: this one is the one that has always existed, I don't want to change this one because other things are probably relying on it already, for our request I created a new one and returned this to their old original place

the one for the agent copilot is GetVisibleComments

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsTokenAuthController.cs
```diff
@@ -0,0 +1,49 @@
...
+    private readonly IBlobCodeFileRepository _codeFileRepository;
+
+    public CommentsTokenAuthController(ICommentsManager commentsManager,
+        IAPIRevisionsManager apiRevisionsManager,
+        IBlobCodeFileRepository codeFileRepository)
+    {
+        _commentsManager = commentsManager;
+        _apiRevisionsManager = apiRevisionsManager;
+        _codeFileRepository = codeFileRepository;
+    }
```
tjprescott, line 32: Why does Comments need a special token auth?

src/dotnet/APIView/APIViewUnitTests/APIRevisionsControllerTests.cs
```diff
@@ -86,8 +84,7 @@ public async Task GetAPIRevisionTextAsync_SpecificType_WithValidId_ReturnsRevisi
     public async Task GetAPIRevisionTextAsync_SpecificType_WithoutId_ReturnsBadRequest()
     {
         string reviewId = "review123";
-        ActionResult<string> result = await _controller.GetAPIRevisionTextAsync(
-            reviewId);
+        ActionResult<string> result = await _controller.GetAPIRevisionTextAsync(null, reviewId);
 
         BadRequestObjectResult badRequestResult = Assert.IsType<BadRequestObjectResult>(result.Result);
         Assert.Equal("apiRevisionId is required when selectionType is Specific",
```
tjprescott: This seems off. There are two kinds of ids you can provide: review id and revision id. Specific type should be paired with REVIEW id. If you specify revision id, specifying a "SpecificType" should error (arguably unless that revision happens to BE the requested type).

tjprescott: It with help if the unit test include the kind of id. 
`GetAPIRevisionTextAsync_SpecificType_WithoutReviewId_ReturnsBadRequest`

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsTokenAuthController.cs
AlitzelMendez: do you mean why is not public? or why is not on the cookies authentication policy?

src/dotnet/APIView/APIViewUnitTests/APIRevisionsControllerTests.cs
tjprescott: If we let people specify _both_ we need to validate that they agree. If you supply the revision ID AND the review ID, they should match and if they don't we should error out. So we wouldn't want logic that is like:

```
if revisionId {
  # ignore reviewId
} else if reviewId {
  # do whatever
}
```

The logical use cases here are:
1. specific revision id
2. review id + some SpecificType

Weird but valid cases
3. specific revision id + matching review id 
4. specific revision id + matching review id + matching SpecificType

Any other case would be some kind of error.

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsTokenAuthController.cs
tjprescott: Github just highlighted it all in green, which made me think that this is brand new.

AlitzelMendez: this is brand new and this file only has one entry point that is getVisibleComments.
this has the token policy as the expectation is to use it for the mpc/agent !  


just the explain around this
- on commnents (cookie policy) we will keep the regular getComments that returns all comments no matter if they are visible, this is an existing endpoint so altering his behavior does seem like the right choice
- because of this I created a different endpoint for our purpose  getVisibleComments with the intention of being used by agent  so I returned the comments in the way copilot is expecting, this has a different policy, so it is in a different file :)

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsController.cs
praveenkuttappan: I think this is used in Conversation tab which is supposed to show all comments in the review irrespective of revision.

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsTokenAuthController.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    /// <summary>
+    ///     Retrieve visible comments for a revision.
+    /// </summary>
+    /// <param name="apiRevisionId"></param>
+    /// <returns></returns>
+    [HttpGet("getVisibleComments", Name = "GetVisibleComments")]
```
praveenkuttappan: is it better to call it getRevisionComments?

praveenkuttappan (approved):
Looks fine. One suggestion is about end point name

src/dotnet/APIView/APIViewUnitTests/APIRevisionsControllerTests.cs
AlitzelMendez: updated :)

src/dotnet/APIView/APIViewUnitTests/APIRevisionsControllerTests.cs
```diff
@@ -83,11 +81,10 @@ public async Task GetAPIRevisionTextAsync_SpecificType_WithValidId_ReturnsRevisi
     }
 
     [Fact]
-    public async Task GetAPIRevisionTextAsync_SpecificType_WithoutId_ReturnsBadRequest()
+    public async Task GetAPIRevisionTextAsync_SpecificType_WithoutRevisionId_ReturnsBadRequest()
     {
         string reviewId = "review123";
-        ActionResult<string> result = await _controller.GetAPIRevisionTextAsync(
-            reviewId);
+        ActionResult<string> result = await _controller.GetAPIRevisionTextAsync(null, reviewId, APIRevisionSelectionType.Specific);
```
tjprescott: This should not return bad request--this should pass!

It is specific type without _review_ id that should fail. 

To keep the logic simpler, I would recommend just validating that either the user supplies "specificType" AND "reviewId" or ONLY "revisionId". That avoids all the annoying edge cases.

AlitzelMendez: why this should pass?!
 
the revision id is not provided and it is requesting an specific revisionId, the only thing provided is the reviewId so with only the reviewId we can't know which specific revisionId is requested 

"specificType" AND "reviewId"  => should fail
"specificType" AND "revisionId" => should pass
only "revisionId" => should pass

PR #12017, opened 2025-09-09
[azsdk-cli] Use scoped tools and helpers, add micro agent example

benbp:
Making all the MCP tools injected via `AddScoped` instead of as singletons. This will allow us to inject other classes that can manage state for the request scope. In this PR the token usage helper is made scoped so that we can track agent token usage across the code for an entire request, and then later consume it from multiple places like our telemetry service (FYI @praveenkuttappan).

Also updating the backend foundry project endpoint for the agent service, since the legacy project has started breaking for API calls.

To play around with the scoped token usage helper, I added an example tool/cli command for the micro agent service which computes fibonacci.

Finally, this fixes an issue from last month where logs coming out of any `Azure.Sdk.Tools.*` namespace were not printed to console.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TokenUsageHelper.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TokenUsageHelper.cs
@@ -2,24 +2,8 @@ namespace Azure.Sdk.Tools.Cli.Helpers;
 
 public class TokenUsageHelper(IOutputHelper outputHelper)
 {
-    private static Dictionary<string, (double inputPrice, double outputPrice)> modelPrices = new()
-    {
-        // Global pricing via https://platform.openai.com/docs/pricing
-        { "gpt-5", (1.25, 10) },
-        { "gpt-5-mini", (0.25, 2) },
-        { "gpt-4.1", (2, 8) },
-        { "gpt-4.1-mini", (0.40, 1.60) },
-        { "gpt-4.1-nano", (0.10, 0.40) },
-        { "gpt-4o", (2.50, 10) },
-        { "gpt-4o-mini", (0.15, 0.60) },
-        { "o3-mini", (1.10, 4.40) },
-    };
-
     protected double PromptTokens { get; set; } = 0;
     protected double CompletionTokens { get; set; } = 0;
-    protected double InputCost { get; set; } = 0;
-    protected double OutputCost { get; set; } = 0;
-    protected double TotalCost { get; set; } = 0;
     protected IEnumerable<string> ModelsUsed { get; set; } = [];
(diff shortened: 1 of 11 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TokenUsageHelper.cs
```diff
@@ -1,80 +1,58 @@
...
-    protected double PromptTokens { get; set; }
-    protected double CompletionTokens { get; set; }
-    protected double InputCost { get; set; }
-    protected double OutputCost { get; set; }
-    protected double TotalCost { get; set; }
-    public List<string> Models { get; set; } = [];
+    private static Dictionary<string, (double inputPrice, double outputPrice)> modelPrices = new()
+    {
+        // Global pricing via https://platform.openai.com/docs/pricing
+        { "gpt-5", (1.25, 10) },
```
richardpark-msft: This can be tricky because deployment names (in Azure OpenAI) don't have to match model names. Does Azure OpenAI provide query APIs so we can get this from the server-side instead?

The other option is to create separate model deployments for each "tool" instead, and track it there.

richardpark-msft: (also, prices do change, not sure if we want to really have to update the tool each time just so the cosmetics are right)

richardpark-msft: My suggestion - let's just report token values, and calculate cost out-of-band instead.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Example/ExampleTool.cs
```diff
@@ -481,5 +503,64 @@ public async Task<ExampleServiceResponse> DemonstratePowershellExecution(string
...
+
+    public record Fibonacci
+    {
+        public int Index { get; set; }
+        public int Previous { get; set; }
+        public int Current { get; set; }
+    }
+
+    [McpServerTool(Name = "azsdk_example_microagent_fibonacci"), Description("Demonstrates micro-agent computing Nth Fibonacci number via iterative tool calls")]
+    public async Task<DefaultCommandResponse> DemonstrateMicroagentFibonacci(int n, CancellationToken ct = default)
```
benbp, line 515: FYI @timovv I added this example tool and cli command to run the micro agent service in case it's useful for ad-hoc testing.

```
azsdk example demo microagent --fibonacci 7
```

praveenkuttappan (approved):
Looks fine to me.

PR #12021, opened 2025-09-09
Simplify readme generator test services and remove noisy test logging

benbp:
This updates the readme generator tests to not use the service provider for setup, which was causing a lot of logs to get printed out when tests are passing. Instead we now inject `TestLogger` manually.

FYI @richardpark-msft

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
@@ -1,25 +1,40 @@
 using System.CommandLine;
 using System.CommandLine.Parsing;
-using Azure.AI.OpenAI;
-using Azure.Sdk.Tools.Cli.Helpers;
+using Moq;
 using Azure.Sdk.Tools.Cli.Microagents;
-using Azure.Sdk.Tools.Cli.Services;
 using Azure.Sdk.Tools.Cli.Tests.Mocks.Helpers;
 using Azure.Sdk.Tools.Cli.Tests.TestHelpers;
 using Azure.Sdk.Tools.Cli.Tools.Package;
-using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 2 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/Generators/ReadMeGeneratorToolTests.cs
```diff
@@ -128,45 +141,6 @@ public async Task TestBadReadmeContent(string readmeContent, string expectedFeed
             }
         }
 
-        /// <summary>
-        /// Creates a service provider in the same way as the normal program. Can be used to instantiate real clients.
-        /// </summary>
-        /// <param name="customizeServices">Optional Action you can use to register your own client instances</param>
-        /// <returns></returns>
-        private static (ServiceProvider, MockOutputHelper) CreateServiceProvider(Action<ServiceCollection>? customizeServices = default)
```
richardpark-msft, line 136: I had this in here specifically to use the service collection. I'm not sure why we should remove it if all we're trying to do is use a fake logger.

Can we split this out from the changes for this particular PR?

benbp: I think it only needed the service collection so the azure openai sdk could be overridden. However the code is just using the micro agent service now so we can mock that directly.

benbp: If we want to use the service collection more formally in test I think that should be common. I prefer not using it if we can avoid it, as it's easier to define mocks inline to the test where they are used as well as to understand what test dependencies are being used.

richardpark-msft: We can just remove it, just wanted to make the point that it wasn't an accident. I do think it's nice to create the object the same way it's created in "prod".

richardpark-msft (approved):

PR #12051, opened 2025-09-11
Create shared helper module for tsp-client commands

samvaity:
Created shared helpers for tsp-client commands  (common utility class used by multiple tools) so they provide single implementation of core logic 
- Looser coupling (for multiple tool calls to use) 
- Clear separation of “UI” (CLI command) vs logic 

> Will add update workflow in a follow-up PR, which will include details with the TspClientUpdate Tooling (it was increasing the size of PR)


samvaity:
Did a poor rebase

PR #12057, opened 2025-09-11
Create shared helper module for tsp-client commands

samvaity:
Created shared helpers for tsp-client commands  (common utility class used by multiple tools) so they provide single implementation of core logic 
- Looser coupling (for multiple tool calls to use) 
- Clear separation of “UI” (CLI command) vs logic 

> Will add update workflow in a follow-up PR, which will include details with the TspClientUpdate Tooling (it was increasing the size of PR)

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TspClientHelper.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TspClientHelper.cs
@@ -1,0 +1,57 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using Azure.Sdk.Tools.Cli.Models.Responses;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Sdk.Tools.Cli.Helpers;
+
+/// <inheritdoc />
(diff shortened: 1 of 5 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TspClientHelper.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        this.npxHelper = npxHelper;
+        this.logger = logger;
+    }
+
+    /// <inheritdoc />
+    public async Task<TspToolResponse> ConvertSwaggerAsync(string swaggerReadmePath, string outputDirectory, bool isArm, bool fullyCompatible, bool isCli, CancellationToken ct)
+    {
+        logger.LogInformation("tsp-client convert: {readme} -> {out} (arm={arm}, fullyCompatible={fc})", swaggerReadmePath, outputDirectory, isArm, fullyCompatible);
+        var npxOptions = new NpxOptions(
```
raych1, line 24: We will need to install `tsp-client` from common `package-lock` file to avoid vulnerable security issues. But it can be done in a separate PR.

PR #12060, opened 2025-09-11
Pin tsp-client under eng/common/tools

### LINKED ISSUE #11950: Introduce a package-lock file to pin the 'tsp-client' version in eng/common folder
The best practice of using an NPM package in automation to guarantee repeatable result is: 
 - introduce `package.json` and `package-lock.json` to lock the version of this package and all the transit dependencies
 - run `npm ci` command based on this `package-lock.json` file.

The idea is to introduce these two files and update the version when `tsp-client` releases a new version. We should add a basic test for each SDK language to validate the integration part. 

With this, when these files get merged to `azure-sdk-tools` repo, they will be flowed into each SDK language repos. 

CC: @weshaggard  @catalinaperalta

catalinaperalta:
Fixes #11950

### DIFF
--- a/eng/common/tsp-client/README.md
+++ b/eng/common/tsp-client/README.md
@@ -21,6 +21,7 @@ cd eng/common/tsp-client
 
 # Install dependencies
 npm ci
+```
 
 ## Usage
 
@@ -65,7 +66,7 @@ If you need to manually update the tsp-client version:
    ```json
    {
      "dependencies": {
-       "@azure-tools/typespec-client-generator-cli": "0.8.0"
+       "@azure-tools/typespec-client-generator-cli": "0.28.1"
      }
    }
    ```


eng/common/tools/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@azure-tools/typespec-client-generator-cli": "0.28.1"
```
catalinaperalta: Is this enough or do you want to pin anything else? If you look at the lock file all the deps are pinned to a specific version.

raych1: As of now, this is the only library we would like to lock.

raych1:
Looks good to me to keep the name as `package.json` and `package-lock.json` without mentioning 'tsp-client', since we may add more deps to them in the future. 
+@weshaggard @mikeharder to have a look.

weshaggard:
Do we want any wrappers or at least some instructions on how to consume this? i.e. I assume one would likely `npm ci` and then `npx --no --package @azure-tools/typespec-client-generator-cli` from the context of this directory?

As for the directory I'm curious about @mikeharder's take on this to see if we should put it at the root of the `eng\common` directory or have a npm subdirectory we use or what. I know we talked about having other tools in here but if we are following the cspell approach this would go under a tsp-client folder and we would have a wrapper script that would install and execute it.

mikeharder:
> Do we want any wrappers or at least some instructions on how to consume this? i.e. I assume one would likely `npm ci` and then `npx --no --package @azure-tools/typespec-client-generator-cli` from the context of this directory?
> 
> As for the directory I'm curious about @mikeharder's take on this to see if we should put it at the root of the `eng\common` directory or have a npm subdirectory we use or what. I know we talked about having other tools in here but if we are following the cspell approach this would go under a tsp-client folder and we would have a wrapper script that would install and execute it.

High-level, I agree with Wes.  I think we should:

1. Copy/paste from `eng/common/spelling` to `eng/common/tsp-client`, including the pwsh wrapper.
2. Consider refactoring the pwsh wrapper, to reduce duplicate code between `spelling` and `tsp-client`.  Maybe a file like `eng/common/npm-exec-wrapper.ps1`?
3. Update consumers to call `eng/common/tsp-client/Invoke-TspClient.ps1`
    1. I prefer a simpler name like `tsp-client.ps1` (or even `tsp-client.[cmd|sh]`, which can be a tiny wrapper around the pwsh, to give us the flexibility to remove pwsh without a breaking change)

If we ever decide to merge JS tools into a single lockfile, it might look like this:

```
eng/common/node-tools/package[-lock].json (all node-based tools)
eng/common/spelling/Invoke-CSpell.ps1 (installs and runs under node-tools)
eng/common/tsp-client/Invoke-TspClient.ps1 (installs and runs under node-tools)
```

Advantages:
- We only need a single JS lockfile in eng/common
- If callers run both tools on a single machine, install perf will be much better

Disadvantages:
- If callers only run a single tool, install perf may be worse, since you're installing unused packages

@weshaggard: Do you have a preference?  Since cspell and tsp-client are totally unrelated, and we only have two node-based tools, i'd probably keep them separate for now.  But, we can always change this later on, and it won't even be a breaking change, if we don't move or break the PWSH wrappers.

CC: @danieljurek, @benbp

raych1:
I also prefer to keep `spell check` and `tsp-client` as separate since they're totally un-related. 

Is it too much to wrap the execution of `tsp-client`? Are we going to handle all the commands `tsp-client` provide? Is there any concern to integrate with [.NET build target command](https://github.com/Azure/azure-sdk-for-net/blob/6e2a7fcfcdaf9cd3c44b1fa712bd384c74e7f7ed/eng/CodeGeneration.targets#L5)? 

Should we start with minimum version containing `tsp-client` installation only, then iterate on the improvement?

weshaggard:
I think having them separate would be good as well.

As for the wrapper it is more of a guide but folks can use it if they like. The [.NET command]( https://github.com/Azure/azure-sdk-for-net/blob/6e2a7fcfcdaf9cd3c44b1fa712bd384c74e7f7ed/eng/CodeGeneration.targets#L5) will work but someone still needs to run `npm ci` before that will succeed. IIRC .NET team did that because they run it a bunch of times in parallel.

mikeharder:
> I also prefer to keep `spell check` and `tsp-client` as separate since they're totally un-related.

Fine with me, at least for now.  It's a little more work for engsys, to update two lockfiles in eng/common.

> Is it too much to wrap the execution of `tsp-client`? Are we going to handle all the commands `tsp-client` provide? Is there any concern to integrate with [.NET build target command](https://github.com/Azure/azure-sdk-for-net/blob/6e2a7fcfcdaf9cd3c44b1fa712bd384c74e7f7ed/eng/CodeGeneration.targets#L5)?

The pwsh script should just delegate all CLI arguments to tsp-client.  The pwsh script is just responsible for installing the package (if not already installed), then calling `npm exec` with the same CLI arguments.

For perf, the pwsh script should not run `npm ci` before every `npm exec`, since this is too slow.  Something like `npm i --ignore-scripts --no-audit` might be fast enough (should no-op if already up-to-date).  Since we control both the `package.json` and the `package-lock.json`, and will ensure they stay in sync, in theory it should be sufficient to always use `npm i --ignore-scripts --no-audit`, even for the first install.

mikeharder:
> As for the wrapper it is more of a guide but folks can use it if they like. The [.NET command](https://github.com/Azure/azure-sdk-for-net/blob/6e2a7fcfcdaf9cd3c44b1fa712bd384c74e7f7ed/eng/CodeGeneration.targets#L5) will work but someone still needs to run `npm ci` before that will succeed. IIRC .NET team did that because they run it a bunch of times in parallel.

The main complaint I have heard from language repo owners, is they don't want to deal with the nuances of commands like `npm`.  So, I think we should encourage language repos (perhaps except for JS) to use the pwsh wrapper.  It ensures they are installing and running the package in a safe, reliable, and consistent way.

raych1:
> The [.NET command](https://github.com/Azure/azure-sdk-for-net/blob/6e2a7fcfcdaf9cd3c44b1fa712bd384c74e7f7ed/eng/CodeGeneration.targets#L5) will work but someone still needs to run `npm ci` before that will succeed. IIRC .NET team did that because they run it a bunch of times in parallel.

The wrapped powershell script includes the installation part. I expect that invoking the powershell script will work for .NET case.

eng/common/tsp-client/Invoke-TspClient.ps1
```diff
@@ -0,0 +1,99 @@
...
+.EXAMPLE
+./eng/common/tsp-client/Invoke-TspClient.ps1 -TspClientOptions "version"
+
+This will run tsp-client version
+
+#>
+[CmdletBinding()]
+param(
+  [Parameter(Mandatory = $true)]
+  [string] $TspClientOptions,
```
mikeharder:
```suggestion
  [Parameter(ValueFromRemainingArguments = $true)]
  [string[]] $TspClientOptions,
```

I think this is better, since callers can now use just this:

```
./eng/common/tsp-client/Invoke-TspClient.ps1 generate --output-dir ./generated

./eng/common/tsp-client/Invoke-TspClient.ps1 version
```

raych1: I thought about this, but it will be little bit confusing when callers specify script parameters along with the options of `tsp-client`. I intended to take current approach to make it clearer between script parameters and `tsp-client` parameters.
e.g.,
```
./eng/common/tsp-client/Invoke-TspClient.ps1 -PackageInstallCache "~/sdk-repo-root" generate --output-dir ./generated

Comment (mikeharder) [reply eng/common/tsp-client/Invoke-TspClient.ps1]: I think the script parameters are rarely used, and we should optimize for the common case.  I'd rather move the script parameters to env vars, than force callers to wrap all the args they care about into a string.

File: eng/common/tsp-client/Invoke-TspClient.ps1 [hunk 3]
```diff
@@ -0,0 +1,99 @@
...
+
+  [Parameter()]
+  [string] $PackageInstallCache = (Join-Path ([System.IO.Path]::GetTempPath()) "tsp-client-tool-path"),
+
+  [Parameter()]
+  [switch] $LeavePackageInstallCache
```
mikeharder: I think this should default to "true", to optimize for the common case where you might be running this script multiple times, and don't want re-install every time.

eng/common/tsp-client/Invoke-TspClient.ps1
```diff
@@ -0,0 +1,99 @@
...
+if (!(Test-Path "$PackageInstallCache/package-lock.json")) {
+  Copy-Item "$PSScriptRoot/package-lock.json" $PackageInstallCache
+}
+
+$originalLocation = Get-Location
+
+try {
+  Set-Location $PackageInstallCache
+  Write-Host "Installing dependencies in $PackageInstallCache"
+  npm i --ignore-scripts --no-audit | Write-Host
```
weshaggard: Is this thread-safe? I know at least for .NET they were running things in parallel and if we tell them to start using the script we should make it thread-safe in some way.

mikeharder: Good catch, but if this script needs to be threadsafe, this might start getting more complex than i'd like to support.  Hard to know what to do now.  This is why trying to write any sort of "wrapper" around npm is a slippery slope.

I'm guessing this isn't an issue for spellcheck, since nobody cares about running it in parallel?

Maybe we go back to removing the wrapper, and just have callers exec "npm ci" at the right time and place?  Or we could add two wrappers, one for "install" and one for "run", but this might be getting ridiculous.

eng/common/tsp-client/Invoke-TspClient.ps1
raych1: It does need to specify the cache folder in case of SDK validation pipeline. It needs to install 'tsp-client' close to the sdk code folder to avoid using the global cache which might has another version of 'tsp-client' installed from spec repo config.

eng/common/tsp-client/Invoke-TspClient.ps1
raych1: It's thread safe as long as the caller specify the different cache path. Given current .NET only wraps `tsp-client` call within `msbuild target` but leave its installation to external script, how about providing options in this script to the caller to ignore the installation or ignore the 'tsp-client' run?

mikeharder: The goal of the wrapper script, was to prevent callers from trying to install and run the NPM package directly, and doing it unsafely or incorrectly (e.g. `npx tsp-client@latest`).  If the wrapper script isn't a complete replacement for both install and run, I don't think it's worth maintaining at all.  Better to just call "npm ci" and "npm exec" directly, and ensure through code review that all consumers use it correctly.

raych1: Does it make sense to provide this script to other languages but leave .NET to run `npm ci` based on these `package-lock.json` and run `npm exec` separately?

weshaggard: I think we can keep the wrapper but have a parameter that enables folks to disable the install step if they know they have already installed.

mikeharder: @weshaggard: My opinion, the wrapper is becoming more trouble than it's worth at this point.

We're just trading "learn how to use npm" with "learn how to use our custom script, which sometimes is simpler than npm, but sometimes you still need to install with npm before running our script".  And if something goes wrong (like a stale cache), now you need to debug our custom code, instead of relying on the standard behavior of npm.  We tried building something like this (custom package manager) in `autorest`, and it continues to be a support burden today.

But if you feel the script still adds value, I will let you design it.

weshaggard: I'm indifferent about whether or not we have a script. I was hoping the script would act as shared code as well as documentation on how we expect folks to consume this. So, if we don't go for a script, I suggest at least a README that outlines how we expect folks to consume this lock file.

raych1: @mikeharder @weshaggard I removed the wrapper script and added the README.md.

eng/common/tsp-client/README.md
```diff
@@ -0,0 +1,101 @@
...
+- **Reliability**: Reduces potential points of failure during installation
+
+## Usage
+
+After installation, you can run tsp-client using one of these approaches:
+
+### Option 1: Change directory approach (recommended)
+
+```bash
+cd eng/common/tsp-client
```
weshaggard, line 31: Does it require changing the current working directory? @mikeharder is there an option or some way to run this without switching working directories?

weshaggard: Is that the prefix option below?

raych1: It has another usage with `npx`, I don't suggest it since it needs to type the package name. Do you have any suggestions, @mikeharder 
```
npx --package-root eng/common/tsp-client @azure-tools/typespec-client-generator-cli init
```

eng/common/tsp-client/README.md
```diff
@@ -0,0 +1,101 @@
+# TypeSpec Client Generator CLI
+
+This directory contains npm package definitions for `@azure-tools/typespec-client-generator-cli` (tsp-client) with pinned versions to ensure reproducible builds across environments and enhance security by preventing supply chain attacks.
```
mikeharder:
```suggestion
This directory contains npm package definitions for `@azure-tools/typespec-client-generator-cli` (tsp-client) with pinned versions to ensure reproducible builds across environments.
```

eng/common/tsp-client/README.md
```diff
@@ -0,0 +1,101 @@
...
+# Install dependencies (recommended for CI/security)
+npm i --ignore-scripts --no-audit
+```
+
+The `--ignore-scripts --no-audit` flags provide:
+
+- **Security**: Prevents potentially malicious install scripts from running
+- **Performance**: Skips vulnerability audit for faster installs
+- **Reliability**: Reduces potential points of failure during installation
+
```
mikeharder:
```suggestion
npm ci
```

If consumer is installing manually (once), just use `npm ci` for simplicity.

eng/common/tsp-client/README.md
mikeharder: According to this, they recommend changing the working directory, rather than using the `--prefix` flag:

https://github.com/npm/cli/issues/1368#issuecomment-1241076700

I'll keep looking, but unless I can find something, we'll need to keep the guidance to use the working directory.

eng/common/tsp-client/README.md
```diff
@@ -0,0 +1,93 @@
...
+```bash
+cd eng/common/tsp-client
+npm i --ignore-scripts --no-audit
+npm exec --no -- tsp-client generate --output-dir ./output
+```
+
+## Package Management
+
+### Updating tsp-client Version
```
mikeharder: I think we should configure dependabot to update this, rather than relying on remembering to manually update.

Also, we should periodically run `npm update` (e.g. once a month), which is required to update all transitive deps.  We could maybe use dependabot for this as well (or write a scheduled GH action).

eng/common/tsp-client/README.md
raych1: Yes, I have the option called out as recommended approach. For the scenario where cannot change the folder, use the option 2.

eng/common/tsp-client/README.md
raych1: Added it to `dependabot.yml`, and created an [issue](https://github.com/Azure/azure-sdk-tools/issues/12120) for `npm update` work.

weshaggard: I would prefer to use dependabot for all the dependency updates here.

.github/dependabot.yml
```diff
@@ -7,3 +7,9 @@ updates:
     ignore:
       - dependency-name: "*"
         update-types: ["version-update:semver-patch"]
+  
+  - package-ecosystem: "npm"
+    directory: "/eng/common/tsp-client/"
```
weshaggard:
```suggestion
    directory: "/eng/common/"
```

Let's do it for everything under `/eng/common`, we should probably do that for everything in the repo also but that can be separate.

eng/common/tsp-client/README.md
raych1: Updated.

raych1:
@mikeharder can you approve this PR if there're no requested changes?

.github/dependabot.yml
```diff
@@ -7,3 +7,9 @@ updates:
     ignore:
       - dependency-name: "*"
         update-types: ["version-update:semver-patch"]
+  
+  - package-ecosystem: "npm"
+    directory: "/eng/common/"
```
mikeharder, line 12:
```suggestion
    directory: "/eng/common/tsp-client/"
```

Do you want to update everything under eng/common, or just tsp-client?

eng/common/tsp-client/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Install dependencies
+
+```bash
+# Navigate to this directory
+cd eng/common/tsp-client
+
+# Install dependencies (for development/manual use)
+npm ci
+## Usage
```
mikeharder:
```suggestion

## Usage
```

.github/dependabot.yml
raych1: Yes, Wes suggested to do that. https://github.com/Azure/azure-sdk-tools/pull/12060#discussion_r2355925829

eng/common/tsp-client/README.md
mikeharder: Per my comment above, I don't think we should recommend `--prefix` at all.  Only the option to change the working directory.

eng/common/tsp-client/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+# Initialize a new project
+npm exec --prefix "${REPO_ROOT}/eng/common/tsp-client" --no -- tsp-client init --config ./tspconfig.yaml
+```
+
+## CI/CD Best Practices
+
+```bash
+cd eng/common/tsp-client
+npm i --ignore-scripts --no-audit
```
mikeharder:
```suggestion
npm ci
```

mikeharder (approved):
approved pending fixes to latest comments

eng/common/tsp-client/README.md
raych1: This is for pipeline, why not using 'npm i'?

.github/dependabot.yml
raych1: Will update it if it won't work.

eng/common/tsp-client/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **Node.js** (with npm) - Required to install and run tsp-client
+
+## Installation
+
+### Install dependencies
+
+```bash
+# Navigate to this directory
+cd eng/common/tsp-client
```
catalinaperalta, line 20: @raych1 is this guidance only for pipelines? I dont think switching to this directory is very user-friendly for regular developers in the language repos.

raych1: @catalinaperalta  per the best practices NPM recommends, we only suggest user to go to the `eng/common/tsp-client` folder then run the command from there. Refer to [this comment](https://github.com/Azure/azure-sdk-tools/pull/12060#discussion_r2361167482)

PR #11958, opened 2025-09-04
Support `standard` proxy mode in test-proxy

scbedd:
First step of #6177

todo:
- [x] Successfully get a recording using the new standard proxy mode
- [x] Ensure existing standard mode fixed everywhere
- [x] Add a couple of tests checking new functionality
- [x] Readme updates
- [x] Respond to PR feedback
- [x] REmove `stopUniversal` or `startUniversal`. They're unnecessary given that we have the static configuration accessible from `startup.cs`
- [x] Fix failing recorded tests

PR #12061, opened 2025-09-12
Deploy Test-Proxy w/ `standard` proxy + multipart/* body boundaryless body matching

scbedd:
Includes:

- #11958
- #12179


scbedd:
Looking quite positive. I'm waiting for the last run of .NET before updating this PR with a version of test-proxy released from `main`.

weshaggard (approved):

PR #12065, opened 2025-09-12
Change branch name from 'master' to 'main'

maririos:


JiaqiZhang-Dev:
Those knowledge source are related to **azure-rest-api-specs.wiki** and **azure-sdk-for-python.wiki**. For github wiki, it seems the default branch is `master`, not `main`.
image

maririos:
> Those knowledge source are related to **azure-rest-api-specs.wiki** and **azure-sdk-for-python.wiki**. For github wiki, it seems the default branch is `master`, not `mainain`. 

How interesting, I learned that wiki's are like different repositories with their own branch and everything. 
Thank you @JiaqiZhang-Dev

PR #12067, opened 2025-09-12
[azsdk-cli] Unified the path parameter

raych1:
Unified the `path` parameter of all the functions in `GitHelper`:
- accept any path within a repo folder
- calls `DiscoverRepoRoot()` internally to find the repo root
- use consistent naming

@benbp @praveenkuttappan could you review this PR?

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
@@ -22,9 +22,6 @@ namespace Azure.Sdk.Tools.Cli.Helpers
     {
         private readonly ILogger<GitHelper> logger = logger;
         private readonly IGitHubService gitHubService = gitHubService;
-        
-        // Cache for repository root discovery to avoid redundant calls
-        private readonly Dictionary<string, string> _repoRootCache = new();
 
         /// <summary>
         /// Gets the SHA of the merge base (common ancestor) between the current branch and the target branch.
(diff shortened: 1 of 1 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
```diff
@@ -122,52 +152,106 @@ public async Task<string> GetRepoOwnerNameAsync(string path, bool findUpstreamPa
...
         }
 
-        public string DiscoverRepoRoot(string path)
+        /// <summary>
+        /// Internal method to get repository root with caching to avoid redundant Repository.Discover calls.
+        /// </summary>
+        /// <param name="pathInRepo">Any path within the git repository (file or directory)</param>
+        /// <returns>The absolute path to the repository root directory</returns>
+        /// <exception cref="InvalidOperationException">Thrown when no git repository is found at or above the specified path</exception>
+        private string GetRepoRootCached(string pathInRepo)
```
benbp: Do the `Repository.Discover` calls add any non-trivial latency? I have hesitated to add caching to `GetRepoRoot` because I don't want to worry about cache invalidation if someone has been running the MCP server for a while and they move their repo path.

raych1: That's a good point. The latency is trivial to discover the repo root. I removed the cache.

raych1: @benbp can you have one more look?

praveenkuttappan (approved):

benbp (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
benbp: If we find a need for it in the future, I could see us having a more centralized filesystem cache service adding a background watcher on directories to invalidate.

PR #12074, opened 2025-09-15
[azsdk-cli] Revert to using azp for PR CI

benbp:
The github action has slowed down a bit with newer changes so we're going to revert to focusing on performance improvements in the azure pipelines CI instead.


weshaggard:
Let's talk about this a little more tomorrow in our 1:1. While I appreciate GH actions might spin up faster I worry that things will get out-of-sync with what we ship and what we validate.

praveenkuttappan (approved):

PR #12086, opened 2025-09-15
Touch a proxy file, still erroring on restore?

scbedd:
Just checking to see if the same thing happens here as in #11958


scbedd:
It absolutely does. This is a pool update most likely.

PR #12087, opened 2025-09-16
[tsp-client] Make emitter-output-dir support less breaking

catalinaperalta:
Go and Rust have historically recommended folks add an emitter-output-dir variable in their tspconfig.yaml files for their libraries, but relied on tsp-client ignoring emitter-output-dir when generating client libraries. The 0.28.0 version of tsp-client which added support for emitter-output-dir broke these languages because we give preference to emitter-output-dir over package-dir in tspconfig.yaml. Having both `package-dir` and `emitter-output-dir` should be an error going forward (tracking issue TBD), we have also merged updates to the tspconfig.yaml files for Go and Rust libraries so the latest versions of all tsp projects are clean and indeed only have emitter-output-dir. However, since Go has many libraries that will eventually need to be updated to the latest commit in the specs repo, tsp-client is switching the order in which it checks for package-dir or emitter-output-dir for folks referencing older commits of a given tspconfig.yaml.

### DIFF
--- a/tools/tsp-client/src/commands.ts
+++ b/tools/tsp-client/src/commands.ts
@@ -97,10 +97,10 @@ async function initProcessDataAndWriteTspLocation(
   }
 
   // Check for relevant package path variables and resolve
-  let emitterOutputDir = tspConfigData?.options?.[emitterData.emitter]?.["emitter-output-dir"];
+  let emitterOutputDir = tspConfigData?.options?.[emitterData.emitter]?.["package-dir"];
   const packageDir = tspConfigData?.options?.[emitterData.emitter]?.["package-dir"];
   let newPackageDir;
-  if (emitterOutputDir) {
+  if (!packageDir && emitterOutputDir) {
     const [options, _] = await resolveCompilerOptions(NodeHost, {
       cwd: process.cwd(),
       entrypoint: "main.tsp",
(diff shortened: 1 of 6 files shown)


catalinaperalta:
A specific test with a tspconfig.yaml that has both is still tbd, but wanted to confirm that this change will have no impact on C# or other emitter configs @JoshLove-msft @raych1 ? I believe Go and Rust were the ONLY languages to have both package-dir and emitter-output-dir at the same time

JoshLove-msft:
> A specific test with a tspconfig.yaml that has both is still tbd, but wanted to confirm that this change will have no impact on C# or other emitter configs @JoshLove-msft @raych1 ? I believe Go and Rust were the ONLY languages to have both package-dir and emitter-output-dir at the same time

I think this should be fine for C#.

raych1:
> @JoshLove-msft @raych1 ? I believe Go and Rust were the ONLY languages to have both package-dir and emitter-output-dir at the same time

Yes, other languages don't have them at the same time.

timotheeguerin (approved):

tools/tsp-client/src/commands.ts
```diff
@@ -415,7 +415,9 @@ export async function generateCommand(argv: any) {
     );
   }
 
-  const legacyPathResolution = !tspConfigData?.options?.[emitter]?.["emitter-output-dir"];
+  // Give preference to package-dir if both are specified
+  // This is to avoid breaking existing behavior
+  const legacyPathResolution = tspConfigData?.options?.[emitter]?.["package-dir"];
```
JoshLove-msft, line 420: This breaks libraries that are generating from a spec that has neither options.

PR #12089, opened 2025-09-16
improved exception and logging

radhgupta:
## PR Summary & Focus

This PR introduces a loop over the process: Compile TypeSpec, Generate Library,  Capture errors , Find fixes ,  Apply fixes , Repeat  

While testing, I found that applying fixes one-at-a-time causes confusion about which version of `client.tsp` the agent should be working with.  Two approaches were considered:

1. Repeatedly upload the updated `client.tsp` after each fix (which requires additional API calls)  
2. Use a **single prompt** where the agent applies all necessary fixes in one go

I’ve implemented the second approach in this PR, as it appears more effective so far.

---

## Additional Improvements: Logging, Exception Handling & Performance

While doing this work I also applied some learnings around logging, exception handling, and performance:

1. **Avoiding unnecessary cost from logs**  
   Logs can impose overhead even when the level is off, since formatting or argument evaluation may still execute. To reduce that cost, I added `logger.IsEnabled(...)` checks around expensive operations.

2. **Using `Result<T>` vs exceptions**  
   - For *expected failures* (e.g. validation errors), `Result<T>` is faster and cleaner, since using exceptions incurs stack unwinding cost. :contentReference[oaicite:0]{index=0}  
   - For *unexpected failures* (e.g. network timeouts, service errors), exceptions still make sense, since their overhead is small relative to the underlying I/O and their semantics are clearer in those cases.  

3. **Single‐boundary exception handling**  
   I’ve tried to push exception handling up to a single boundary (e.g. in `Program.cs`), so that business logic isn’t littered with try/catch and we have a clear place for logging and translating exceptions.

---

## Feedback Needed

There are still areas I believe can be improved. I’m particularly uncertain about:

- Whether I’ve **differentiated correctly** between when to use `Result<T>` vs when to use exceptions  
- If any exception handling is redundant, misplaced, or overly broad  
- Whether there are places where logging is duplicated, too verbose, or missing needed context  

@jsquire @m-redding I’d appreciate your thoughts especially on the `Result<T>` vs exception usage. Any suggestions for refactoring or optimizing the flow would be great.

### DIFF
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentFileManager.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentFileManager.cs
@@ -9,7 +9,7 @@ namespace Azure.Tools.GeneratorAgent.Agent
     {
         private const string UnknownStatus = "Unknown";
         private const string ProcessedStatus = "Processed";
-        
+
         private readonly PersistentAgentsClient Client;
         private readonly ILogger<AgentFileManager> Logger;
         private readonly AppSettings AppSettings;
@@ -19,7 +19,7 @@ namespace Azure.Tools.GeneratorAgent.Agent
             ArgumentNullException.ThrowIfNull(client);
             ArgumentNullException.ThrowIfNull(logger);
             ArgumentNullException.ThrowIfNull(appSettings);
-            
+
             Client = client;
             Logger = logger;
             AppSettings = appSettings;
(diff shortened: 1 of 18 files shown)


src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutionService.cs
```diff
@@ -194,7 +194,7 @@ private void LogExecutionResult(bool success, int exitCode, string command, stri
             else
             {
                 Logger.LogError("Command {Command} failed with exit code {ExitCode}. Error output: {ErrorOutput}",
-                    command, exitCode, error);
+                    command, exitCode, output);
```
m-redding: Curious - would it make sense to write both the error and the output here or is the output more informative?

m-redding (approved):
This seems reasonable to me! I think it makes sense for the agent to throw exceptions in one root place.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentFileManager.cs
```diff
@@ -33,48 +33,48 @@ public async Task<string> UploadFilesAsync(Dictionary<string, string> files, Can
 
             var results = await Task.WhenAll(uploadTasks).ConfigureAwait(false);
 
-            var uploadedFileIds = results
-                .Where(id => !string.IsNullOrEmpty(id))
-                .Select(id => id!);
+            var failures = results.Where(r => r.IsFailure).ToList();
```
jsquire:
```suggestion
            var failures = results.Where(r => r.IsFailure)
            var firstFailure = failures.FirstOrDefault();;
```

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentFileManager.cs
```diff
@@ -33,48 +33,48 @@ public async Task<string> UploadFilesAsync(Dictionary<string, string> files, Can
...
+            if (failures.Any())
```
jsquire:
```suggestion
            if (firstFailure != null)
```

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentFileManager.cs
jsquire: Since we own the set, we can make the assumption that it is safe to enumerate multiple times and avoid the `ToList` just to pull the first member.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/FixGeneratorService.cs
```diff
@@ -35,14 +34,32 @@ public async Task<List<Fix>> AnalyzeAndGetFixesAsync(Result<object>? compileResu
...
+                Logger.LogDebug("No errors to analyze");
+
+                return Result<List<Fix>>.Success(new List<Fix>());
+            }
+
+            // Process all tasks and handle failures
             var allFixResults = await Task.WhenAll(fixTasks).ConfigureAwait(false);
+            
+            // Check for any failures
+            var failures = allFixResults.Where(r => r.IsFailure).ToList();
```
jsquire: Same thought here around tweaking the pattern.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Program.cs
```diff
@@ -20,46 +20,46 @@ public class Program
...
             ServiceProvider = serviceProvider;
             Logger = ServiceProvider.GetRequiredService<ILogger<Program>>();
             CommandLineConfig = new CommandLineConfiguration(ServiceProvider.GetRequiredService<ILogger<CommandLineConfiguration>>());
         }
 
         public static async Task<int> Main(string[] args)
         {
-            ToolConfiguration toolConfig = new ToolConfiguration();
-            using ILoggerFactory loggerFactory = toolConfig.CreateLoggerFactory();
+            var toolConfig = new ToolConfiguration();
```
jsquire, line 31: image

PR #12126, opened 2025-09-17
Implement extensible evaluators for APIView Copilot

marcarro:
# Refactor evaluation system to support multiple evaluator types via YAML configuration
This PR replaces the hard-coded evaluation approach with a config system that supports different types of evaluators through YAML workflow files.

## Changes
- Added `BaseEvaluator` abstract interface defining the contract for all evaluators
- Implemented registry pattern in `EVALUATORS` dict for dynamic evaluator discovery
- Created YAML workflow configuration loader with validation for workflow definitions
- Refactored `EvalRunner` to instantiate evaluators dynamically based on workflow configuration
- Added concrete evaluator implementations: `MentionActionEvaluator` and `ThreadResolutionActionEvaluator` for prompt-based evaluations
- Updated CLI to accept YAML files
- Created workflow definitions for `apiview`, `mention-action`, and `thread-resolution-action` evaluation types

## Benefits
- New evaluator types can be added by implementing `BaseEvaluator` and updating the registry
- Evaluation configurations are now declarative and stored in YAML files

- Clear separation between evaluation logic and configuration
- The system now supports both the original APIView evaluations and new prompt-based evaluations through a unified interface.

# Usage
```bash
# Run the mention action workflow
python run.py -t workflows/mention-action.yaml

# Run the thread resolution action workflow
python run.py -t workflows/thread-resolution-action.yaml
```


packages/python-packages/apiview-copilot/evals/run.py
```diff
@@ -14,6 +14,13 @@
         default="python",
         help="The language to run evals for. Defaults to python.",
     )
+    parser.add_argument(
+        "--evaluator-type",
+        "-e",
+        type=str,
+        default="apiview",
+        help="The type of evaluator to use. Defaults to apiview.",
+    )
```
tjprescott: So my thought here is that the test you target shoudl really determine this because the format of the test is closely tied to the kind of evaluator needed. It wouldn't make sense to run the "reviews.jsonl" against a prompt evaluator, for example. Also, that would require you to have "mention evaluator", and "thread resolution evaluator" etc. which is not going to be scalable. Is there some way we can unify these concepts?

packages/python-packages/apiview-copilot/evals/_runner.py
```diff
@@ -8,47 +8,35 @@
...
-            "similarity_weight": 0.1,  # Similarity between expected and actual
-            "false_positive_penalty": 0.3,  # Penalty for false positives
-            "fuzzy_match_bonus": 0.2,  # Bonus for fuzzy match (right rule, wrong line)
-        }
 
     def run(self):
         """Run the evaluation."""
-        custom_eval = CustomAPIViewEvaluator()
+        evaluator_class = EVALUATORS[self.evaluator_type]
+        custom_eval = evaluator_class()
```
tjprescott: Two things.
1) refactor `custom_eval` to maybe just `evaluator`.
2) we should have some way so that we don't need to specify `evaluator_type`. Perhaps that is something that can be stored in the JSONL?  Or, we could create something like a YAML format that links the JSONL file to its proper class. Let's discuss at our sync today.

tjprescott (changes requested):
Overall, I like the direction this is going. It addresses a lot of the FIXMEs that were in the code after the last refactor. I do want us to discuss how a test case is linked to the necessary evaluator. Mix-n-match only opens a pit of failure since logically a given test case is tightly bound to the necessary evaluator class.

packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -201,3 +310,250 @@ def __call__(self, *, response: str, query: str, language: str, actual: str, tes
...
+
+    def show_results(self, processed_results: dict) -> None:
+        """Display APIView results in table format."""
+        import json
+        import pathlib
+        from tabulate import tabulate
+        
+        for name, test_results in processed_results.items():
+            baseline_results = {}
+            baseline_path = pathlib.Path(__file__).parent / "results" / "python" / name[:-1]  # TODO: make language dynamic
```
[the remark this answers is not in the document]
tjprescott: Well, this particular test _is_ for Python so this is probably fine here. However, if the number of classes is going to grow, it will likely be prudent to refactor these so they aren't all in the same file.

packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -3,23 +3,131 @@
...
+    from src._apiview_reviewer import ApiViewReview
+    reviewer = ApiViewReview(target=query, language=language, base=None)
+    review = reviewer.run()
+    reviewer.close()
+    return {"actual": review.model_dump_json()}
+
+
+# TODO: Since these are global functions passed to the eval, the prompty path is built from relative paths. 
+# But the prompty path is already included in the YAML, so there's redundancy there. Perhaps the workflow config could somehow be accessed here.
+def _mention_action_workflow(**kwargs):
```
tjprescott: I agree with this TODO. If you look at `_mention_action_workflow` and `_thread_resolution_action_workflow`, they are identical except for he prompty file, which, as you point out, is in the config. What we want, ideally, is a single `_run_prompty` callable that is configued via the config yaml as needed. Something like:

```python
def build_prompty_callable(self, prompt_file: str):
    def _run_prompty(**kwargs):
       prompty_path = Path(__file__).parent.parent / "prompts" / "mention" / prompty_file
    prompty_kwargs = {k: v for k, v in kwargs.items() 
                     if k not in {"response", "testcase"}}
    result = prompty.execute(prompty_path, inputs=prompty_kwargs)
    return {"actual": result}
  return _run_prompty
```

packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -201,3 +328,547 @@ def __call__(self, *, response: str, query: str, language: str, actual: str, tes
...
+                    print(f"     Expected: {test['expected_action']}")
+                    print(f"     Actual:   {test['actual_action']}")
+                    print()
+
+
+# Registry of available evaluators
+EVALUATORS: Dict[str, BaseEvaluator] = {
+    "apiview": CustomAPIViewEvaluator,
+    "mention-action": MentionActionEvaluator,
+    "thread-resolution-action": ThreadResolutionActionEvaluator
```
tjprescott: What are the actual differences between these two evaluators? It seems like they would be effectively or entirely the same. In this instance the goal would be to reuse the same class (`PromptyEvaluator` for example) for both of these, with any differences expressed via the config.

packages/python-packages/apiview-copilot/evals/run.py
```diff
@@ -19,15 +18,19 @@
...
+        help="The number of runs to perform, with the median of results kept. Defaults to 1.",
     )
     parser.add_argument(
         "--test-file",
         "-t",
         type=str,
-        default="reviews.jsonl",
-        help="Only run a particular jsonl test file, takes the name or path to the file. Defaults to all.",
+        required=True,
+        help="Path to workflow YAML or JSONL file name.",
```
tjprescott: How does targeting the JSONL work? Does it look for a matching YAML file, or something else?

tjprescott (changes requested):
Outstanding TODO should be resolved, and a few other opportunities for reuse (hopefully!)

packages/python-packages/apiview-copilot/evals/run.py
marcarro: I initially intented to support both YAML and JSONL file names but figured it was redundant, so I went for yaml-only functionality. Modified the incorrect documentation

packages/python-packages/apiview-copilot/evals/_custom.py
marcarro: You're right, the two classes have the same logic and the only difference is the fields they query. Now, the workflow configuration also includes the comparison fields, which can then be accessed by the unified `PromptWorkflowEvaluator`

packages/python-packages/apiview-copilot/evals/_custom.py
marcarro: I have tried implementing a closure as a target over the past few days (like your example) but Azure kept complaining about it not being a global function or module. I even tried doing something like `globals()[closure] = closurefunc()` but that didn't work either. So I stuck with the approach you see. Any other ideas?

packages/python-packages/apiview-copilot/evals/workflows/mention-action.yaml
```diff
@@ -1,4 +1,10 @@
 name: mention-action
 kind: prompt
 tests: ../tests/python/mention-action.jsonl
-prompty: ../../prompts/mention/parse_conversation_action.prompty
\ No newline at end of file
+prompty: ../../prompts/mention/parse_conversation_action.prompty
+evaluation_config:
+  comparison_field: recommendation
```
tjprescott, line 6: Are you on the latest branch? These should both be "action" fields on the latest.

packages/python-packages/apiview-copilot/evals/_config_loader.py
```diff
@@ -96,12 +104,34 @@ def load_workflow_config(path: str | os.PathLike) -> WorkflowConfig:
     if not isinstance(runs, int) or runs < 1:
         _fail(f"runs must be positive integer (got: {runs!r})")
 
+    evaluation_config = None
+    if kind == "prompt":
+        eval_config_raw = raw.get("evaluation_config")
```
tjprescott: Seems like a code smell that we have this very specific logic that only applies to the `PromptEvaluator` here rather than in the `PromptEvaluator` class itself. Could this be extracted to some kind of method in the ABC that Evaluators can optionally implement?

tjprescott (commented):
A couple comments left, but overall, love the number of red lines compared to the number of green!

packages/python-packages/apiview-copilot/evals/workflows/mention-action.yaml
marcarro: Seems like it's still `recommendation` https://github.com/Azure/azure-sdk-tools/blob/02e3e21aed63b351600b4c6e66268f36c4d09089/packages/python-packages/apiview-copilot/prompts/mention/parse_conversation_action.prompty#L59

tjprescott: Oof, so it is. Okay, we can merge this and I'll refactor as needed. These should use basically the same schema.

tjprescott (approved):
This LGTM. Would like a review by @kristapratico  or @mccoyp  before we merge.

packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -3,23 +3,128 @@
...
 from src._utils import get_prompt_path
 
+from _config_loader import WorkflowConfigError, EvaluationConfig
 
-class CustomAPIViewEvaluator:
+class PromptWorkflowTarget:
+    def __init__(self, prompty_subpath, prompty_filename):
+        self.prompty_path = Path(__file__).parent.parent / "prompts" / prompty_subpath / prompty_filename
+
+    def __call__(self, **kwargs):
```
kristapratico: When trying to run this `python run.py -t workflows/mention-action.yaml` locally I'm getting an error about no data to process / target function not producing results. I suspect this is because the evaluation library doesn't support kwargs on the target function and wants the data from the jsonl file defined as typed parameters on the target function like 

`def __call__(self, testcase: str, response: str, ...)`

Can you double-check?

packages/python-packages/apiview-copilot/evals/_runner.py
```diff
@@ -1,28 +1,19 @@
...
-DEFAULT_NUM_RUNS: int = 1
-
-
-def _review_apiview(query: str, language: str):
-    """Internal global callable for evals framework."""
-    reviewer = ApiViewReview(target=query, language=language, base=None)
-    review = reviewer.run()
-    reviewer.close()
-    return {"actual": review.model_dump_json()}
+from _config_loader import load_workflow_config, WorkflowConfigError, get_evaluator_class
```
kristapratico: When running from CLI it won't find the module unless we add

```suggestion
from evals._config_loader import load_workflow_config, WorkflowConfigError, get_evaluator_class
```

packages/python-packages/apiview-copilot/evals/_runner.py
```diff
@@ -1,28 +1,19 @@
...
+import _custom
```
kristapratico: `import _custom` looks unused

marcarro: At the very end of `_custom.py`, the prompt and apiview evaluators are registered, and `runner.py` seemed not to find those when run, until I added the import statement. I'm not sure why this is.

kristapratico (approved):
Nice job, approved! 🚀

Issue #12083, opened 2025-09-15
Failed to get pull request details from SDK generation pipeline


maririos:
user testing the MCP got this on their logs. Is this because there was no release plan created?
We should fail better :)

> "response": [
        {
          "value": "I see that Go and .NET SDK generation failed, but Python, JavaScript, and Java succeeded. This is common during development - let's check the successful ones first and then address the failures.\n\nLet me get the SDK pull request links for the successful generations: ",
          "supportThemeIcons": false,
          "supportHtml": false,
          "baseUri": {
            "$mid": 1,
            "path": "/c:/Users/mattge/source/repos/testagent/azure-rest-api-specs/",
            "scheme": "file"
          }
        },
        {
          "kind": "prepareToolInvocation",
          "toolName": "mcp_azure-sdk-mcp_azsdk_get_sdk_pull_request_link"
        },
        {
          "kind": "toolInvocationSerialized",
          "invocationMessage": {
            "value": "Running `azsdk_get_sdk_pull_request_link`",
            "isTrusted": false,
            "supportThemeIcons": false,
            "supportHtml": false
          },
          "pastTenseMessage": {
            "value": "Ran `azsdk_get_sdk_pull_request_link` ",
            "isTrusted": false,
            "supportThemeIcons": false,
            "supportHtml": false
          },
          "originMessage": "azure-sdk-mcp (MCP Server)",
          "isConfirmed": {
            "type": 4
          },
          "isComplete": true,
          "source": {
            "type": "mcp",
            "serverLabel": "azsdk",
            "label": "azure-sdk-mcp",
            "collectionId": "mcp.config.ws0",
            "definitionId": "mcp.config.ws0.azure-sdk-mcp"
          },
          "resultDetails": {
            "input": "{\n  \"buildId\": 5332924,\n  \"language\": \"Python\",\n  \"workItemId\": 0\n}",
            "output": [
              {
                "type": "embed",
                "isText": true,
                "value": "Failed to get pull request details from SDK generation pipeline, Error: The controller for path '/_apis/wit/workItems/0' was not found or does not implement IController."
              }
            ],
            "isError": false
          },
          "toolSpecificData": {
            "kind": "input",
            "rawInput": {
              "buildId": 5332924,
              "language": "Python",
              "workItemId": 0
            }
          },
          "toolCallId": "c7b56f3e-e284-429e-bfaa-d19907c040a8",
[...]

PR #12127, opened 2025-09-17
Add validation to get_sdk_pull_request_link for work item id

smw-ms:
https://github.com/Azure/azure-sdk-tools/issues/12083

Need a valid work item id for the code to work with or without a `buildId`


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/SpecWorkFlowTool.cs
```diff
@@ -329,9 +329,9 @@ public async Task<string> GetSDKPullRequestDetails(string language, int workItem
                     return $"Unsupported language to get pull request details. Supported languages: {string.Join(", ", SUPPORTED_LANGUAGES)}";
                 }
 
-                if (buildId == 0 && workItemId == 0)
+                if (workItemId == 0)
```
maririos: why is the workItem ID required if we have the buildId?

praveenkuttappan: We should not do this change. It will effectively enforce a release plan as mandatory to run SDK generation and we don't want to do that. We need to check why did the tool failed to get SDK details from build ID. WorkItemId is optional.

PR #12137, opened 2025-09-17
modelOnly check for Hlc To Modular

skywing918:
… paths and update isModelOnly to check for both "src/api" and "src/operations" directories in GitHub.


kazrael2119:
please use `netapp` to re-run pipeline to test

Issue #11215, opened 2025-07-15
Devex Agent: AI Evaluations


jeo02:
- Start looking at evaluations and how we can utilize in the Devex Agent

jeo02:
# MCP Agent Evaluation follow up

## Overview

The primary objective is to develop an evaluation framework for our MCP agent. My idea is to create **Scenario** objects that can be played against changes to copilot instructions / any instruction. 

## Scenario Components

Each Scenario comprises three main elements:

### 1. Copilot Chat History
A transcript of a copilot session, ideally selected from chats that demonstrate effective interaction. This history serves as the context for evaluation and is used to simulate the ideal scenario for agent behavior.

### 2. Expected Outcome
Focused primarily on the agent's tool usage rather than the textual responses. Key criteria include:
- Correct tools invoked by the agent
- Proper order of tool usage (where sequencing matters)
- Accuracy of tool input parameters

### 3. AzureFoundry Evaluation Boolean
A flag indicating whether to execute Azure Foundry evaluations for the scenario. 

## Tasks

### 1. Extracting Chat History
Identify a well-performing copilot session, particularly at pivotal interaction steps (e.g., SDK generation up to the release planner). Capture the chat history leading to a key agent action.
- For now working with a basic chat history testing the Liftrarize sdk generation but testing `Step 2: Validate TypeSpec Specification` where the expected output would have the correct path to the `...azure-rest-api-specs\\specification\\liftrarize\\ArizeAi.ObservabilityEval.Management`. 
- As this project grows the hope is to have more scenarios. 

### 2. Formatting for OpenAI
Convert the copilot chat log into an OpenAI-compatible format. Two approaches are considered:
- Intercepting the payload via a proxy to directly obtain OpenAI-formatted data
- Programmatically transforming copilot chat data to mirror OpenAI's expected input structure

### 3. AI Evals Integration 
Utilize evaluation methodologies developed by Krista and Azure Foundry's built-in functions for agentic models (which Krista has also worked on). These provide quantitative scores for agent performance.
- Hope to get Kristas feedback on whatever work is done here this week by next week. 

### 4. Infrastructure 
Setting up azure resources and correct permissions. **Currently wrapping this part up**
Currently using the shared resources. Resources needed:
- Web App used to host the mcp server remotely. (Needed to connect to Azure Foundry)
- Azure Foundry

## Goal and Impact

[...]

PR #12147, opened 2025-09-17
MCP Evaluations for Copilot Instructions

### LINKED ISSUE #12040: AI Evaluations: Check in Code
Tracking when current state will be checked in.

jeo02:
- Custom Evaluation using the Microsoft.Extensions.AI Library.
- Main idea comes from this issue: #11215 

## Quick Overview
- When copilot instructions or any instructions are changed, we can quantify/measure the impact through evaluations. The evaluations library also offers many built in evaluations along with the custom one.
- Still work to be done as some missing items but this would be the setup. 

Resolves #12040


tools/ai-evals/azsdk-mcp/azsdk-mcp.csproj
```diff
@@ -0,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <RootNamespace>azsdk_mcp_evals</RootNamespace>
```
benbp: The naming convention we tend to use would be something like `Azure.Sdk.Tools.McpEvals`, for both the root namespace, any child namespaces, and the csproj filename.

tools/ai-evals/azsdk-mcp/TestData/example.json
```diff
@@ -0,0 +1,187 @@
+[
```
benbp, line 1: @weshaggard @maririos @scbedd @kristapratico do we think the mcp eval configs will get pretty large over time? We could consider storing them in the [assets repo](https://github.com/Azure/azure-sdk-assets) instead, as it was built to support general purpose assets beyond test recordings and would be easy to use here.

tools/ai-evals/azsdk-mcp/Evaluators/ExpectedToolInputEvaluator.cs
```diff
@@ -0,0 +1,152 @@
...
+                try
+                {
+                    // Serialize with consistent options for better comparison
+                    var options = new JsonSerializerOptions
+                    {
+                        WriteIndented = false,
+                        PropertyNamingPolicy = JsonNamingPolicy.CamelCase
+                    };
+
+                    var expectedJson = JsonSerializer.Serialize(toolCall.exp.Arguments, options);
```
benbp: Probably a rare case these days, but if you can use a sorted dictionary it will avoid any potential flaky tests as dictionary key ordering is not guaranteed.

tools/ai-evals/azsdk-mcp/azsdk-mcp.csproj
```diff
@@ -0,0 +1,31 @@
...
+    <PackageReference Include="Microsoft.Extensions.AI" Version="9.9.0" />
+    <PackageReference Include="Microsoft.Extensions.AI.Evaluation" Version="9.9.0" />
+    <PackageReference Include="Microsoft.Extensions.AI.Evaluation.Quality" Version="9.9.0" />
+    <PackageReference Include="Microsoft.Extensions.AI.Evaluation.Reporting" Version="9.9.0" />
+    <PackageReference Include="Microsoft.Extensions.AI.OpenAI" Version="9.8.0-preview.1.25412.6" />
+    <PackageReference Include="Microsoft.Extensions.Hosting" Version="9.0.8" />
+    <PackageReference Include="ModelContextProtocol" Version="0.3.0-preview.4" />
+    <PackageReference Include="MSTest.TestFramework" Version="3.7.0" />
+    <PackageReference Include="MSTest.TestAdapter" Version="3.7.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
```
benbp: These appear unused. Also we try to standardize on nunit for tests instead of these and/or `Microsoft.VisualStudio.TestTools.UnitTesting`

tools/ai-evals/azsdk-mcp/TestData/example.json
jeo02: I definitely think that we would have multiple scenarios where the JSON files may get very large. So I definitely see a benefit to storing them there.

tools/ai-evals/azsdk-mcp/azsdk-mcp.csproj
jeo02: I can swap to nunit honestly only used MSTest cause I was following their sample for reporting evaluations but it also supports nunit. And yea woops forgot to take a look at what I wasn't using.

tools/ai-evals/azsdk-mcp/TestData/example.json
benbp: I think it should be relatively simple. We can use the test proxy tool to generate an `assets.json` config and handle pushing recordings. But whatever generates the initial eval json files would need to be able to write to the proxy directory (`azure-sdk-tools/.assets`) in the correct format.

@scbedd I wonder if it's worth putting the `GitStore` into a library that we can reference from this tooling, or if just following the existing path conventions is "good enough."

scbedd: I think needing to call the test-proxy from within a pure .NET environment that is our mcp server seems really awkward.

...That basically tells me that GitStore should be a library.

jeo02: Alright sounds good, I'll reach out offline then as I figure out the recording story. I made this issue to track it for now. #12221

tools/ai-evals/azsdk-mcp/Evaluators/ExpectedToolInputEvaluator.cs
```diff
@@ -0,0 +1,173 @@
+using System.Text.Json;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Evaluation;
+using Azure.Sdk.Tools.McpEvals.Models;
+
+namespace Azure.Sdk.Tools.McpEvals.Evaluators
+{
+   public class ExpectedToolInputEvaluator : IEvaluator
+   {
+        public const string ExpectedToolInputMetricName = "Azure SDK MCP";
```
praveenkuttappan:
```suggestion
        public const string ExpectedToolInputMetricName = "Azure SDK Tools MCP";
```

tools/ai-evals/azsdk-mcp/Evaluators/ExpectedToolInputEvaluator.cs
```diff
@@ -0,0 +1,173 @@
...
+                    EvaluationDiagnostic.Error(
+                        $"No Tool calls detected inside of {nameof(additionalContext)} collection."));
+
+                return result;
+            }
+
+            if (!actualToolCalls.Any())
+            {
+                metric.AddDiagnostics(
+                    EvaluationDiagnostic.Error(
```
praveenkuttappan: Can you include information about expected Tool calls in this error message so it's very clear that actual tool call is empty when the expectation is a number of tool calls.

tools/ai-evals/azsdk-mcp/Evaluators/ExpectedToolInputEvaluator.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            // Equal number of tool calls
+            var expCount = expectedToolCalls.Count();
+            var actCount = actualToolCalls.Count();
+            if (expCount != actCount)
+            {
+                metric.AddDiagnostics(
+                    EvaluationDiagnostic.Error(
+                        $"The LLM Result had more tool calls than expected: Actual calls: {actCount}, Expected calls: {expCount}"));
```
praveenkuttappan: nit: LLM had different set of tool calls than expected.

tools/ai-evals/azsdk-mcp/Evaluators/ExpectedToolInputEvaluator.cs
```diff
@@ -0,0 +1,173 @@
...
+                    EvaluationDiagnostic.Error(
+                        $"Provided LLM Result contained no tool calls."));
+
+                return result;
+            }
+
+            // Equal number of tool calls
+            var expCount = expectedToolCalls.Count();
+            var actCount = actualToolCalls.Count();
+            if (expCount != actCount)
```
praveenkuttappan, line 52: One challenge in this approach is that sometimes LLM will make a tool call with incorrect value and then they auto fix it and call the tool again with correct value. So we will end up having more tool calls with same result.

praveenkuttappan: Do we ignore any tool call with failures to filter out retry attempts?

tools/ai-evals/azsdk-mcp/Evaluators/ExpectedToolInputEvaluator.cs
```diff
@@ -0,0 +1,173 @@
...
+                    EvaluationDiagnostic.Informational(
+                        $"Tool call #{countCalls}, expected the {expToolName} tool, and LLM called the {actToolName} tool."));
+
+                // If the names do not align then the tool calls were made in the wrong order by the LLM
+                if (expToolName != actToolName)
+                {
+                    metric.AddDiagnostics(
+                    EvaluationDiagnostic.Error(
+                        $"Tool call made in the wrong order. Expected the {expToolName} tool but the LLM called {actToolName} tool. This was tool call #{countCalls}"));
+                    return result;
```
praveenkuttappan, line 74: Should we process all calls instead of stopping when the order is incorrect?

tools/ai-evals/azsdk-mcp/Evaluators/ExpectedToolInputEvaluator.cs
jeo02: In this scenario it always calls the same tool back to back right? We can simplify the list of Chat Messages from LLM response to remove tools called back to back.

tools/ai-evals/azsdk-mcp/Evaluators/ExpectedToolInputEvaluator.cs
jeo02: We can, would simply report back all the tool calls that were made in the wrong order. My thought was this may be too much noise?

tools/ai-evals/azsdk-mcp/README.md
```diff
@@ -0,0 +1,91 @@
...
+# Azure OpenAI Configuration
+AZURE_OPENAI_ENDPOINT=https://your-openai-resource.openai.azure.com/
+AZURE_OPENAI_MODEL_DEPLOYMENT_NAME=your-deployment-name
+
+# MCP Server Configuration
+LOCAL_MCP_POWERSHELL_SCRIPT_PATH=path/to/your/mcp-script.ps1
+
+# Instruction File Paths
+COPILOT_INSTRUCTIONS_PATH=path/to/copilot-instructions.md
+AZSDK_TOOLS_INSTRUCTIONS_PATH=path/to/azsdk-tools/instructions
```
praveenkuttappan, line 63: Why is path required? Isn't this referred by copilot instructions?

tools/ai-evals/azsdk-mcp/README.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+## Extending the Framework
+
+### Adding New Scenarios
+
+1. Create scenario data in `TestData/` directory
+2. Implement test method in appropriate scenario class
+3. Define expected outcomes and evaluation criteria
```
praveenkuttappan, line 78: Is this to be done in scenario class?

tools/ai-evals/azsdk-mcp/README.md
```diff
@@ -0,0 +1,91 @@
...
+4. Add any custom evaluators if needed or use built in through Microsoft.Extensions.Ai.Quality
```
praveenkuttappan, line 79: Is there any sample scenario where I need to use custom evaluator

tools/ai-evals/azsdk-mcp/README.md
```diff
@@ -0,0 +1,91 @@
...
+        IEnumerable<ChatMessage> messages,
+        ChatResponse modelResponse,
+        // ... additional parameters
+    )
+    {
+        // Custom evaluation logic
+    }
+}
+```
+
```
praveenkuttappan, line 100: Can you add the instructions about how to run this eval testing?

tools/ai-evals/azsdk-mcp/Evaluators/ExpectedToolInputEvaluator.cs
jeo02: Changed this to `Expected Tool Input` since it should be more reflective of what the metric actually is.

tools/ai-evals/azsdk-mcp/README.md
jeo02: The idea being we would provide the current copilot instructions given that it was changed. Leaving it as local for now but ideally would be pulled from GitHub. 

Not sure what you mean by the tests being referred to by copilot instructions.

tools/ai-evals/azsdk-mcp/README.md
jeo02: `AzsdkTypeSpecGeneration_Step02_TypespecValidation` is the current test Scenario that can be run. But I believe I forgot to update the example data so let me do that.

tools/ai-evals/azsdk-mcp/README.md
jeo02: Sure, I mean all it comes down to is using the dotnet test or VS test interface since it is a test framework, but I can add it in.

tools/ai-evals/azsdk-mcp/README.md
jeo02: Yes since the Scenario class is a `TestFixture` to add new tests you add on to the scenario class.

PR #12111, opened 2025-09-16
Expose APIView functionality to tools/mcp agents

AlitzelMendez:
**Issue:** https://github.com/Azure/azure-sdk-tools/issues/10413
 
I implemented an initial effort to expose APIView functionality through both MCP tools and CLI commands.
 
**For MCP tooling, I exposed:**
- `azsdk_apiview_get_comments` - Get comments for revisions
- `azsdk_apiview_list_review_versions` - List all versions for a review
- `azsdk_apiview_check_authentication` - Check authentication status
- `azsdk_apiview_get_authentication_guidance` - Get setup help for authentication
 
**For CLI access, I additionally exposed:**
- `azsdk_apiview_get_latest_revision` - Get the latest revision content
- `get-token-file-for-revision` - Get revision content with optional diff
 
The last two commands were not exposed as MCP tools because they return large token file data, which doesn't seem particularly beneficial for conversational AI interactions.
 
One topic mentioned in the original issue but not addressed here is diff functionality. I want to discuss what we should return for diffs, as we'll probably need a new endpoint. I've split that effort into a separate task.

I will work on additional validation/unit testing, but other than that this is ready for review.

PR #12149, opened 2025-09-18
[API View] Expose endpoints to be used on mcp and cli 

AlitzelMendez:
effort related to https://github.com/Azure/azure-sdk-tools/pull/12111


src/dotnet/APIView/APIViewWeb/LeanControllers/APIRevisionsTokenAuthController.cs
```diff
@@ -134,4 +134,24 @@ public async Task<ActionResult<string>> GetAPIRevisionTextAsync(
             return StatusCode(StatusCodes.Status500InternalServerError, "Failed to generate review text");
         }
     }
+
+    [HttpGet("{revisionId}/getReviewId", Name = "GetReviewId")]
+    public async Task<ActionResult> GetReviewIdForRevision(string revisionId)
```
praveenkuttappan: What's the use case of this REST API?

praveenkuttappan (approved):

src/dotnet/APIView/APIViewWeb/LeanControllers/ReviewsHybridAuthController.cs
```diff
@@ -0,0 +1,95 @@
...
+    ///Retrieve the Content (codeLines and Navigation) of a review
+    ///</summary>
+    ///<param name="reviewId"></param>
+    ///<param name="activeApiRevisionId"></param>
+    /// <param name="diffApiRevisionId"></param>
+    ///<returns></returns>
+    [Route("{reviewId}/content")]
+    [HttpGet]
+    public async Task<ActionResult<CodePanelData>> GetReviewContentAsync(string reviewId, [FromQuery] string activeApiRevisionId,
+        [FromQuery] string diffApiRevisionId = null)
```
tjprescott: (Copied from other PR)
I'm thinking it would make sense to simply overload the existing endpoint to get the revisionText and provide a query param that allows you to select "text" or "codefile". Just look at this, I would expect you'd need to provide a specific revisionId, not a review Id. I'm not sure what the diffApiRevisionId would mean since we don't diff the codefiles.

tjprescott: (@AlitzelMendez reply from other PR)
when there is a diff, a code file is generated with the tree structure of the diff :) mostly for the rendering.

this make sense :) I don't know if we need an overload, we have two endpoints and we can handle this on the cli level.

just to be clear this is the endpoint that I moved from the other file, this is called from ClientSp, so this is endpoint that is already in used, expecting this token file, also on diff, while we can create other endpoints probably is not necessary if we do the changes on the cli level. but I will take a look

I just wanted to share to make it clear that we need to keep this endpoint, but for sure we can add another parameter if needed, but we can't remove the diffApiRevisionId

tjprescott: I don't agree that we need this endpoint. The logic of retrieving a revision should all be the same as the existing endpoint: revisionId or reviewId + some label. That's different than what is suggested by this endpoint. The only difference between what this should conceptually do and what that endpoint should conceptually do is the output format. One is text and one is the CodeFile. That suggests to me that the functionality of retrieving the Codefile should simply be an output format option for the get content endpoint. We don't want a multitude of endpoints that do logically similar things in slightly different ways.

On the subject of the diff, I think the internal "diff code file" is useless to external users and we should not expose it at all. Our implementation is intended for internal use only and is considered "bugged" for external users (this is why AVC doesn't use it).

tjprescott: > Moved the GetReviewContentAsync endpoint to a new hybrid authentication controller that supports both token and cookie authentication

Okay, got it. Yes, we shouldn't do this. We should leave the internal implementation internal and instead enhance our current third-party-facing endpoint (which currently only returns text) to optionally return the revision as a CodeFile.

AlitzelMendez: ok changed this completely :)

src/dotnet/APIView/APIViewWeb/LeanControllers/APIRevisionsTokenAuthController.cs
AlitzelMendez: sorry Praveen I changed all this code, so resolving this conversation :)

tjprescott (approved):

Issue #12151, opened 2025-09-18
[APIView] TypeSpec should not have the Copilot review button


tjprescott:
Currently the button is showed, and it fails due to TSP having duplicate lineIds. However, even if that weren't the case, TypeSpec should not allow Copilot reviews because the SDK API review logic does not apply to TypeSpec:

1. It has no guidelines
2. "Generic review" means little for TypeSpec which is a new language

We should disable this.

PR #12153, opened 2025-09-18
[APIView] TypeSpec should not have the Copilot review button

AlitzelMendez:
closes: https://github.com/Azure/azure-sdk-tools/issues/12151

### DIFF
--- a/src/dotnet/APIView/ClientSPA/src/app/_components/review-page-options/review-page-options.component.spec.ts
+++ b/src/dotnet/APIView/ClientSPA/src/app/_components/review-page-options/review-page-options.component.spec.ts
@@ -156,6 +156,13 @@ describe('ReviewPageOptionsComponent', () => {
         const result = component['isCopilotReviewSupportedForPackage']();
         expect(result).toBe(true);
       });
+
+      it('should return false for TypeSpec packages', () => {
+        component.review!.packageName = 'some-package';
+        component.review!.language = 'TypeSpec';
+        const result = component['isCopilotReviewSupportedForPackage']();
+        expect(result).toBe(false);
+      });
     });
 
     describe('isCopilotReviewSupported property', () => {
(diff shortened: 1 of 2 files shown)


maririos (approved):

tjprescott (commented):
The button already doesn't appear today for various langauges: Swagger, XML, JSON, etc. I don't see those listed here, so I'm wonder if this is really the right place for this logic.

AlitzelMendez:
> The button already doesn't appear today for various langauges: Swagger, XML, JSON, etc. I don't see those listed here, so I'm wonder if this is really the right place for this logic.

it doesn't appear in those places because they don't have the token file tree which probably doesn't mean much when I say it, but even if you look at the url of those language:
- https://apiview.dev/Assemblies/Review/67884367fda04aaeb3721c92f33ad76c

you can see that it doesn't redirect to our new front end, for reference a python one
- https://spa.apiviewstagingtest.com/review/215e0aa46adc42d484e30f95a44d54aa?activeApiRevisionId=666c97feff2d43d0afee898e159579c4

the front end is different, similar but different, so the `Generate copilot review` is not implemented for that front end, so is not like that language are filtered, they just don't have that implemented.

maririos:
> it doesn't appear in those places because they don't have the token file tree

That is correct. This is the case for C and C++ too.
We have communicated with the leads that if they don't move their parser to use token style, they don't get the new APIView UI components.
Copilot is one of those.

tjprescott:
> > The button already doesn't appear today for various langauges: Swagger, XML, JSON, etc. I don't see those listed here, so I'm wonder if this is really the right place for this logic.
> 
> it doesn't appear in those places because they don't have the token file tree which probably doesn't mean much when I say it, but even if you look at the url of those language:

Ah, okay. Well, even if XML, JSON and Swagger _did_ support the tree format, it wouldn't make sense to expose Copilot for them based on the functionality we have now, so it might make sense to have a more declarative implementation of what should and shouldn't have copilot functionality rather than rely on inferring it from whether they use tree tokens or not.

AlitzelMendez:
> > > The button already doesn't appear today for various langauges: Swagger, XML, JSON, etc. I don't see those listed here, so I'm wonder if this is really the right place for this logic.
> > 
> > 
> > it doesn't appear in those places because they don't have the token file tree which probably doesn't mean much when I say it, but even if you look at the url of those language:
> 
> Ah, okay. Well, even if XML, JSON and Swagger _did_ support the tree format, it wouldn't make sense to expose Copilot for them based on the functionality we have now, so it might make sense to have a more declarative implementation of what should and shouldn't have copilot functionality rather than rely on inferring it from whether they use tree tokens or not.

I agree! I will also handle this so when they start using the new UI the button is not in there. Merging this pr so it is included on Monday deployment

Issue #11901, opened 2025-08-28
Support `multipart/form-data` as  `multipart/mixed`


scbedd:
@m-redding  is working on Azure/azure-sdk-for-net#51611

She discovered two issues:

1. we don't treat `multipart/form-data` the same as `multipart/mixed`. They are the same per spec so I should ensure it gets treated the same as `multipart/mixed` re: deserialization, removal of boundaries, and sanitization of just the body bytes
2. We do not ignore `boundary` values when matching request bodies to a record body. I think it would be perfectly reasonable to ignore `boundary` values and only compare the bytes that are actual "data". The users themselves will need to deal with the different boundary value on their side via sanitization or going by index on their side.

scbedd:
After convocation with @hallipr and @jsquire going to execute on the following:

1. match body bytes MINUS the boundaries
2. Return whatever boundaries exist in the request. Users will have to figure out what to do and/or I'll react to feedback.

scbedd:
@m-nash completed this in the referenced PR. Closing.

PR #12179, opened 2025-09-20
Adds ability to compare multipart bodies while ignoring the boundary

m-nash:
Replaces https://github.com/Azure/azure-sdk-tools/pull/11903.
Contributes to https://github.com/Azure/azure-sdk-tools/issues/11901.


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/MultipartUtilities.cs
```diff
@@ -1,21 +1,21 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
 
 namespace Azure.Sdk.Tools.TestProxy.Common
 {
     public static class MultipartUtilities
     {
-        public static readonly byte[] CrLf = new byte[] { (byte)'\r', (byte)'\n' };
+        public static readonly byte[] CrLf = "\r\n"u8.ToArray();
```
scbedd, line 18: TIL

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordMatcher.cs
```diff
@@ -328,6 +355,21 @@ public virtual int CompareHeaderDictionaries(SortedDictionary<string, string[]>
                         entryHeaderValues = RenormalizeContentHeaders(entryHeaderValues);
                     }
 
+                    if (headerName.Equals("Content-Type", StringComparison.Ordinal))
```
scbedd, line 358: I'm a bit concerned about this line of code here. It seems to be permanently mutating the object, which we definitely want that boundary around for parsing the initial boundary for the body yeah?

PR #12193, opened 2025-09-22
[azsdk-cli] Support custom log analysis prompts

### LINKED ISSUE #11655: Allow Custom Prompts for Log Analysis
Feature request from Waneng Li

There are some mgmt plane scenarios where the current PipelineAnalysisTool falls short, primarily because it [cannot utilize custom queries](https://github.com/Azure/azure-sdk-tools/blob/80fcec9352726b540ba38f4b60fdaeec7fe89ad1/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/AzurePipelines/PipelineAnalysisTool.cs#L403) to analyze the log, significantly limiting its functionality. Would it be possible to add a custom query parameter to make the tool more versatile?
For instance, when updating sdk-suppression files, pipeline logs often include [breaking change items](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5169049&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=6551)
 that are neither classified as errors nor warnings. These items are precisely the ones that should be documented in the [sdk-suppressions.yml](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/chaos/Chaos.Management/sdk-suppressions.yaml) file. Users might request the bot to demonstrate and provide an example of how to add these breaking changes to the sdk-suppressions.yaml file to prevent them from blocking the PR.

benbp:
- Support custom prompts from the command line and MCP clients for log analysis
- Only delete used agent resources
- Bump ai sdk dependencies
- Fix bug where ctrl-c in the agent tool does return exit code 1 (if a tool calls another method that has a try/catch => response error handler then the operation canceled exception will get swallowed).

Resolves #11655

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Microagents/MicroagentHostServiceTests.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Microagents/MicroagentHostServiceTests.cs
@@ -61,7 +61,9 @@ internal class MicroagentHostServiceTests
             BinaryData.FromString("""{"Result":"Success"}""")
         );
 
+#pragma warning disable OPENAI001
         var chatCompletion = OpenAIChatModelFactory.ChatCompletion(role: ChatMessageRole.Assistant, toolCalls: [exitToolCall]);
+#pragma warning restore OPENAI001
 
         var agentDefinition = new Microagent<string>
         {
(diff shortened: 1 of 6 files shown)


scbedd (approved):
👍 Seems reasonable to me. I'm kinda trusting on the chnages within PipelineAnalysisTool, as

PR #12195, opened 2025-09-22
[tsp-client] fix save-inputs bug

catalinaperalta:


tools/tsp-client/src/typespec.ts
```diff
@@ -117,11 +117,20 @@ export async function compileTsp({
   }
   const overrideOptions: Record<string, Record<string, string>> = {
     [emitterPackage]: {
-      "save-inputs": saveInputs ? "true" : "false",
       ...emitterOverrideOptions,
     },
   };
 
+  if (saveInputs) {
```
JoshLove-msft, line 124: Can we add a regression test?

catalinaperalta: Will add tomorrow: https://github.com/Azure/azure-sdk-tools/issues/12196

Issue #11411, opened 2025-07-30
Tool: FixCspell


l0lawrence:
### 🔧 Tool Name

FixCspell

### 📄 What Should This Tool Do?

This tool will take the words that are being incorrectly flagged and add them to the current repos cspell.json which lives under .vscode/cspell.json if it exists

### 📥 Inputs and 📤 Outputs

**Inputs**
- word list of words to add
- repo path

**Outputs**
- None

### 🏗️ Execution Context

MCP only

PR #12202, opened 2025-09-23
[azsdkcli] spelling fix tool

l0lawrence:
#11411 

image

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/CheckAllToolTests.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/CheckAllToolTests.cs
@@ -246,8 +246,7 @@ namespace Azure.Sdk.Tools.Cli.Tests.Tools
 
             // Setup mock microagent service to return a successful spelling fix result
             var mockSpellingFixResult = new LanguageChecks.SpellingFixResult(
-                "Successfully fixed 4 spelling errors and added 0 words to cspell.json",
-                "Fixed 'contians' to 'contains', 'obvioius' to 'obvious', 'speling' to 'spelling', 'erors' to 'errors' in test_fix.md"
+                "Successfully fixed 4 spelling errors and added 0 words to cspell.json. Fixed 'contians' to 'contains', 'obvioius' to 'obvious', 'speling' to 'spelling', 'erors' to 'errors' in test_fix.md"
             );
             _mockMicroagentHostService.Setup(x => x.RunAgentToCompletion(It.IsAny<Microagent<LanguageChecks.SpellingFixResult>>(), It.IsAny<CancellationToken>()))
                                      .ReturnsAsync(mockSpellingFixResult);


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/PackageCheckTool.cs
```diff
@@ -23,12 +23,15 @@ ILanguageChecks languageChecks
         public override CommandGroup[] CommandHierarchy { get; set; } = [SharedCommandGroups.Package];
 
         private const string RunChecksCommandName = "run-checks";
+        
+        private readonly Option<bool> fixOption = new(["--fixCheckErrors"], "Enable fixCheckErrors mode for supported checks (like spelling)") { IsRequired = false };
```
timovv: Could it just be `--fix` since that's less typing

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
```diff
@@ -277,6 +280,46 @@ protected async Task<CLICheckResponse> CheckSpellingCommonAsync(string packagePa
                 logOutputStream: true
             );
             var processResult = await _npxHelper.Run(npxOptions, ct: ct);
+
+            // If fix is requested and there are spelling issues, provide fix guidance
+            if (fixCheckErrors && processResult.ExitCode != 0 && !string.IsNullOrWhiteSpace(processResult.Output))
+            {
+                // Build an LLM prompt that instructs the model to either fix typos or add legitimate terms to cspell.json
+                var sb = new System.Text.StringBuilder();
```
timovv: Could use a raw string literal here instead of `StringBuilder` https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated#interpolated-raw-string-literals

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/PackageCheckTool.cs
```diff
@@ -264,21 +268,34 @@ private async Task<CLICheckResponse> RunReadmeValidation(string packagePath, Can
...
             {
-                result.NextSteps = new List<string>
+                if (fixCheckErrors)
                 {
-                    "Fix spelling errors identified in the package files",
-                    "Add legitimate technical terms to the cspell dictionary if needed",
-                    "Review comments, documentation, and variable names for typos",
-                    "Run cspell locally to identify and fix spelling issues before committing"
-                };
+                    result.NextSteps = new List<string>
```
timovv: If we're setting the next steps here this just overrides the next steps that are created in ChekSpellingAsync right? Maybe we don't need to set the next steps in both places?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
```diff
@@ -277,6 +280,46 @@ protected async Task<CLICheckResponse> CheckSpellingCommonAsync(string packagePa
...
+                sb.AppendLine("       * replacement: (string) present only if action == 'fix' — the corrected token or replacement text");
+                sb.AppendLine("       * justification: (string) a brief one-line justification for the recommendation");
+                sb.AppendLine();
+                sb.AppendLine("2) If you recommend 'ignore', ensure you explain why this word should be kept (e.g. product name, acronym, code identifier, or language-specific term).");
+                sb.AppendLine("3) Keep replacements minimal and preserve code formatting and casing.");
+                sb.AppendLine();
+                sb.AppendLine("cspell lint output:");
+                sb.AppendLine(processResult.Output);
+                sb.AppendLine();
+                sb.AppendLine("Return the JSON array now.");
```
benbp: You can use C# raw strings (`"""`) with `$` which will allow variable interpolation and will set the leading whitespace/indendation based on the indendation level in source. 

```suggestion
                var sb = $"""
                You are an automated spelling helper. You will be provided with the output from cspell lint for a code repository.
                For each reported token, decide whether it is a typo that should be corrected, or a legitimate technical/proper term that should be added to the cspell dictionary (the cspell.json 'words' list).
                
                Requirements:
                1) Output ONLY valid JSON: an array of objects. Each object must contain the following fields:
                   - file: path to the file containing the token (relative to repository root)
                   - line: the line number where the token appears
                   - original: the original token as reported by cspell
                   - recommendation: an object with these fields:
                       * action: one of 'fix' or 'ignore' ('fix' means propose a replacement, 'ignore' means add to cspell.json words)
                       * replacement: (string) present only if action == 'fix' - the corrected token or replacement text
                       * justification: (string) a brief one-line justification for the recommendation
                2) If you recommend 'ignore', ensure you explain why this word should be kept (e.g. product name, acronym, code identifier, or language-specific term).
                3) Keep replacements minimal and preserve code formatting and casing.
                
                cspell lint output:
                {processResult.Output}
                
                Return the JSON array now.
                """;
```

benbp: Also lmao but the LLMs are even putting em dashes in their code too (the string description has an em dash instead of a hyphen).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
```diff
@@ -277,6 +280,46 @@ protected async Task<CLICheckResponse> CheckSpellingCommonAsync(string packagePa
...
+            // If fix is requested and there are spelling issues, provide fix guidance
+            if (fixCheckErrors && processResult.ExitCode != 0 && !string.IsNullOrWhiteSpace(processResult.Output))
+            {
+                // Build an LLM prompt that instructs the model to either fix typos or add legitimate terms to cspell.json
+                var sb = new System.Text.StringBuilder();
+                sb.AppendLine("You are an automated spelling helper. You will be provided with the output from cspell lint for a code repository.");
+                sb.AppendLine("For each reported token, decide whether it is a typo that should be corrected, or a legitimate technical/proper term that should be added to the cspell dictionary (the cspell.json 'words' list).");
+                sb.AppendLine();
+                sb.AppendLine("Requirements:");
+                sb.AppendLine("1) Output ONLY valid JSON: an array of objects. Each object must contain the following fields:");
```
benbp: Word of warning, I have found it tedious to get the LLM to always return a string that can be de-serialized from json (it loves to wrap the json in markdown formatting or add an initial description). I've added some things to another prompt (https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Azure/Agent/AzureAgentService.cs#L43) to help with this and I also strip backticks from the response.

This code is just returning the response back to the user/LLM and not de-serializing, so it's probably not as big of an issue.

benbp: oh nvm this code isn't calling out to chatgpt

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/CheckAllToolTests.cs
```diff
@@ -208,5 +208,30 @@ public async Task RunPackageCheck_EnumValues_WorksCorrectly()
...
+            Assert.IsNotNull(fixResult);
+            Assert.IsNotNull(normalResult.CheckStatusDetails);
+            Assert.IsNotNull(fixResult.CheckStatusDetails);
+            
+            // Both may succeed or partially succeed depending on environment and cspell version
+            Assert.That(normalResult.ExitCode, Is.GreaterThanOrEqualTo(0));
+            Assert.That(fixResult.ExitCode, Is.GreaterThanOrEqualTo(0));
+            
+            // The fix result should have different next steps indicating LLM guidance
+            Assert.IsTrue(fixResult.NextSteps != null && fixResult.NextSteps.Any());
```
benbp: I would do `IsNotEmpty` here, otherwise if it fails the test output will just say "expected 'false' to be true" with no indication of what the assertion was aside from a line number.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/PackageCheckTool.cs
```diff
@@ -300,7 +317,7 @@ private async Task<CLICheckResponse> RunSnippetUpdate(string packagePath, Cancel
         }
 
         // Back-compat overload for callers/tests that don't pass a CancellationToken
-        public Task<CLICheckResponse> RunPackageCheck(string packagePath, PackageCheckType checkType)
-            => RunPackageCheck(packagePath, checkType, ct: default);
+        public Task<CLICheckResponse> RunPackageCheck(string packagePath, PackageCheckType checkType, bool fixCheckErrors = false)
```
benbp: Can we update the `CheckAllToolTests.cs` file to pass `TestContext.CancellationToken` so we can remove these back compat methods?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
```diff
@@ -277,6 +280,45 @@ protected async Task<CLICheckResponse> CheckSpellingCommonAsync(string packagePa
                 logOutputStream: true
             );
             var processResult = await _npxHelper.Run(npxOptions, ct: ct);
+
+            // If fix is requested and there are spelling issues, provide fix guidance
+            if (fixCheckErrors && processResult.ExitCode != 0 && !string.IsNullOrWhiteSpace(processResult.Output))
+            {
+                // Build an LLM prompt that instructs the model to either fix typos or add legitimate terms to cspell.json
+                var prompt = $$"""
```
samvaity: It will be good to move these prompts to some sort of a constants file, for easy update and retrieval. I see this scenario applicable to other places, and it would be good to unify a place for it. Maybe Azure.Sdk.Tools.Cli\Prompts, could really help keep things organized.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/PackageCheckTool.cs
```diff
@@ -50,28 +54,28 @@ public override async Task<CommandResponse> HandleCommand(InvocationContext ctx,
...
             }
             else
             {
                 throw new ArgumentException($"Unknown command: {commandName}");
             }
         }
 
-        [McpServerTool(Name = "azsdk_package_run_check"), Description("Run validation checks for SDK packages. Provide package path and check type (All, Changelog, Dependency, Readme, Cspell, Snippets).")]
-        public async Task<CLICheckResponse> RunPackageCheck(string packagePath, PackageCheckType checkType, CancellationToken ct = default)
+        [McpServerTool(Name = "azsdk_package_run_check"), Description("Run validation checks for SDK packages. Provide package path, check type (All, Changelog, Dependency, Readme, Cspell, Snippets), and whether to fix errors.")]
```
samvaity, line 77: I do worry about "whether" I guess we will know better with evals as to if the MCP client is able to navigate it always for spell fixes. 
Curious question, why don't we want to provide this as standard process to always fix spell checks, anything against that?

l0lawrence: if we want this to run in the CI so that the same command is run everywhere, we need a way to just run the checks without fixing them

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
```diff
@@ -277,6 +281,25 @@ protected async Task<CLICheckResponse> CheckSpellingCommonAsync(string packagePa
...
+            if (fixCheckErrors && processResult.ExitCode != 0 && !string.IsNullOrWhiteSpace(processResult.Output))
+            {
+                // Get the spelling fix recommendation prompt from the prompts library
+                var prompt = ValidationPrompts.GetSpellingFixRecommendationPrompt(processResult.Output);
+                
+
+                var response = new CLICheckResponse(processResult.ExitCode, prompt);
+                response.NextSteps = new List<string>
+                {
+                    "Send the returned JSON prompt to your LLM to obtain fix/ignore recommendations.",
```
samvaity: Im confused, does this tooling not send the prompt internally to get fixes or applied patches?

l0lawrence: no, I was under the impression spelling fix can be a little more arbitrary so I did not see the value in internally looping over the errors vs doing it after the tool outputs. I output the errors and a prompt to be handled so the user has more control over seeing what is decided as an actual error or an ignore word. In the case that it is the cli the next-steps would inform the user what to do with the prompt. Do you think there is a benefit to switch to use the internal loop structure here?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
```diff
@@ -282,22 +288,19 @@ protected async Task<CLICheckResponse> CheckSpellingCommonAsync(string packagePa
...
-                };
-
-                return response;
+                    var fixResult = await RunSpellingFixMicroagent(packageRepoRoot, processResult.Output, ct);
+                    return new CLICheckResponse(0, fixResult.Summary);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "Error running spelling fix microagent");
+                    return new CLICheckResponse(processResult.ExitCode, processResult.Output, $"Spelling fix microagent failed: {ex.Message}");
```
benbp: We only stringify response error/next steps so the process output will get hidden. If it's needed then maybe CLICheckResponse should have another constructor overload that takes an error message list and sets the `ResponseErrors` property instead.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Prompts/ValidationPrompts.cs
```diff
@@ -6,41 +6,39 @@ namespace Azure.Sdk.Tools.Cli.Prompts;
...
-               * action: one of 'fix' or 'ignore' ('fix' means propose a replacement, 'ignore' means add to cspell.json words)
-               * replacement: (string) present only if action == 'fix' — the corrected token or replacement text
-               * justification: (string) a brief one-line justification for the recommendation
+        Tasks:
+        1. Analyze each reported spelling issue from the cspell output
+        2. For each issue, either:
+           - Fix the typo by correcting spelling in the source file
+           - Add legitimate technical terms/product names/proper nouns to cspell.json
+        3. Apply fixes by reading files, making corrections, and writing them back
+        4. Update cspell.json 'words' array with legitimate words (never remove existing words)
```
benbp, line 21: How does it know what/where `cspell.json` is if running in an upstream llm context?

benbp: Or are we passing in/out cspell json content via the tool calls?

l0lawrence: a specific update cspell tool!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
l0lawrence: ok I will look into it :)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
```diff
@@ -277,17 +286,65 @@ protected async Task<CLICheckResponse> CheckSpellingCommonAsync(string packagePa
                 logOutputStream: true
             );
             var processResult = await _npxHelper.Run(npxOptions, ct: ct);
+
+            // If fix is requested and there are spelling issues, use Microagent to automatically apply fixes
+            if (fixCheckErrors && processResult.ExitCode != 0 && !string.IsNullOrWhiteSpace(processResult.Output))
+            {
+                try
+                {
+                    var fixResult = await RunSpellingFixMicroagent(packageRepoRoot, processResult.Output, ct);
```
samvaity, line 295: Should we validate the response from the LLM here? If the summary indicates that no operations were performed, do we return and run the cspell check again, or is there another step we should take?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/UpdateCspellWordsTool.cs
```diff
@@ -0,0 +1,61 @@
...
+public record UpdateCspellWordsOutput([property: Description("Success message with number of words added")] string Message);
+
+public class UpdateCspellWordsTool(string baseDir) : AgentTool<UpdateCspellWordsInput, UpdateCspellWordsOutput>
+{
+    public override string Name { get; init; } = "UpdateCspellWords";
+    public override string Description { get; init; } = "Add words to the cspell.json words list";
+
+    public override async Task<UpdateCspellWordsOutput> Invoke(UpdateCspellWordsInput input, CancellationToken ct)
+    {
+        var cspellRelativePath = Path.Combine(".vscode", "cspell.json");
```
samvaity, line 16: Some of these can be constants.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageChecks.cs
l0lawrence: @timovv is working on a pr for agent validation that will also keep all the conversation history -- I will integrate with that when it comes out! It will make sure the ask of the user is resolved before exiting

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/UpdateCspellWordsTool.cs
l0lawrence: if I start using these values in multiple places will add to constants list :)

PR #12222, opened 2025-09-24
Set pr creator to empty when it's bot

raych1:
Now we have `dependabot` enabled in `azure-sdk-tools` repo, and [the sync PRs](https://github.com/Azure/azure-sdk-tools/pull/12188) failed in assigning to the bot.

The cause could be GitHub doesn't allow to assign an issue to someone which don't have the write permission.

### DIFF
--- a/eng/common/scripts/Get-PullRequestCreator.ps1
+++ b/eng/common/scripts/Get-PullRequestCreator.ps1
@@ -18,15 +18,7 @@ try
 {
     $pullRequest = Get-GithubPullRequest -RepoOwner $RepoOwner -RepoName $RepoName `
     -PullRequestNumber $PullRequestNumber -AuthToken $AuthToken
-
-    $prCreator = $pullRequest.user.login
-
-    # Set the prCreator variable to empty if the PR creator is dependabot or copilot
-    if ($prCreator -ilike "dependabot*" -or $prCreator -ilike "copilot*")
-    {
-        $prCreator = ""
-    }
-    Write-Host "##vso[task.setvariable variable=System.PullRequest.Creator;]$($prCreator)"
+    Write-Host "##vso[task.setvariable variable=System.PullRequest.Creator;]$($pullRequest.user.login)"
 }
 catch
 {
--- a/eng/common/scripts/Get-PullRequestCreator.ps1
+++ b/eng/common/scripts/Get-PullRequestCreator.ps1
@@ -18,7 +18,15 @@ try
 {
     $pullRequest = Get-GithubPullRequest -RepoOwner $RepoOwner -RepoName $RepoName `
     -PullRequestNumber $PullRequestNumber -AuthToken $AuthToken
-    Write-Host "##vso[task.setvariable variable=System.PullRequest.Creator;]$($pullRequest.user.login)"
+
+    $prCreator = $pullRequest.user.login
+
+    # Set the prCreator variable to empty if the PR creator is dependabot or copilot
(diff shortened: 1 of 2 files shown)


eng/common/scripts/Get-PullRequestCreator.ps1
```diff
@@ -18,7 +18,15 @@ try
 {
     $pullRequest = Get-GithubPullRequest -RepoOwner $RepoOwner -RepoName $RepoName `
     -PullRequestNumber $PullRequestNumber -AuthToken $AuthToken
-    Write-Host "##vso[task.setvariable variable=System.PullRequest.Creator;]$($pullRequest.user.login)"
+
+    $prCreator = $pullRequest.user.login
+
+    # Set the prCreator variable to empty if the PR creator is dependabot or copilot
+    if ($prCreator -ilike "dependabot*" -or $prCreator -ilike "copilot*")
```
weshaggard: After thinking about this a little more. I don't think this is the best place to do this check. Instead I think we should do it in https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Submit-PullRequest.ps1#L131, where we are trying to set the review and assignee and just skip those steps if we are given a bot account.

raych1: Updated.

eng/common/scripts/Submit-PullRequest.ps1
```diff
@@ -130,7 +130,8 @@ else {
 
     # ensure that the user that was used to create the PR is not attempted to add as a reviewer
     # we cast to an array to ensure that length-1 arrays actually stay as array values
-    $cleanedUsers = @(SplitParameterArray -members $UserReviewers) | ? { $_ -ne $prOwnerUser -and $null -ne $_ }
+    # we also filter out dependabot and copilot users who don't have write permission to avoid errors
+    $cleanedUsers = @(SplitParameterArray -members $UserReviewers) | ? { $_ -ne $prOwnerUser -and $null -ne $_ -and $_ -inotlike "dependabot*" -and $_ -inotlike "copilot*" }
```
raych1: @weshaggard should we keep `copilot` account here since it can be added as reviewers?

weshaggard: That is fair you can keep it for reviewers, but I don't think it can be added as an assignee.

raych1: Updated to allow copilot assigned as reviewer.

eng/common/scripts/Submit-PullRequest.ps1
```diff
@@ -146,8 +147,13 @@ else {
...
     }
 
+    # Clean assignees - remove null entries and bot accounts
+    $cleanedAssignees = @(SplitParameterArray -members $Assignees) | ? { 
+      $null -ne $_ -and $_ -inotlike "dependabot*" -and $_ -inotlike "copilot*" 
+    }
+
     Update-GitHubIssue -RepoOwner $RepoOwner -RepoName $RepoName -IssueNumber $prNumber `
-    -State $prState -Labels $PRLabels -Assignees $Assignees -AuthToken $AuthToken
+    -State $prState -Labels $PRLabels -Assignees $cleanedAssignees -AuthToken $AuthToken
```
weshaggard, line 156: Does this work correctly if the list is empty.

raych1: Yes, the downstream set parameter function handles empty value case.

Issue #12227, opened 2025-09-24
SdkRepoBranch value passed to DevOps is incorrect


praveenkuttappan:
SDK generation pipeline expects either main or actual branch name to be passed when generating the SDK. Pipeline run is not triggered when empty string is passed as SdkRepoBranch and this causes pipeline validation failures.

error message: "The 'SdkRepoBranch' parameter is not a valid String."

praveenkuttappan:
Pull request to fix this issue https://github.com/Azure/azure-sdk-tools/issues/12227

PR #12228, opened 2025-09-24
Do not pass SdkRepoBranch to pipeline trigger when branch name is not…

praveenkuttappan:
PR has changes to avoid passing empty sdk branch name to queue DevOps pipeline because that causes validation failure on DevOps side. DevOps pipeline has a non empty value as default so we cannot pass empty value.


praveenkuttappan:
Fixes https://github.com/Azure/azure-sdk-tools/issues/12227

maririos (approved):

Issue #11389, opened 2025-07-29
Create an official `release` cadence for `azsdk-cli`


scbedd:
@maririos raised a good point that with the larger onboarding of many teams to the `azsdkcli` tool, we need to lock down a bit harder on what issues are being shipped in each version.

With that being the case, I think that we should begin to manually version, and have an official changelog so we can _track_ what code changes are present in each cli version.

This issue covers the necessary pieces.

1. Define a `contributing.md` for the server that outlines how `main` should be treated (stable only, no test bits)
2. Define a `changelog` that is shipped with the server. Add changelog verification to ci pipeline.
3. Update the tool to manual versioning.

scbedd:
CC @weshaggard @praveenkuttappan @maririos @benbp for FYI.

maririos:
Thanks to @jeo02 for also bringing this to attention :)

benbp:
Great points. I think this will be a good way to handle the large expected volume of contributions across multiple areas.

maririos:
The sooner the better. We had to do a quick fix today for the spec automation pipeline parameters and had to quick the release ... hopefully nothing was half done

scbedd:
It's now versioned with a verified changelog + increment post release. Probably should require changelog updates on the tool as well during PRs, but we'll see.

PR #12237, opened 2025-09-24
Manually Version `azsdk-cli`

scbedd:
See title!

Still need to add the actual changelog verification tho.

Addresses #11389


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Azure.Sdk.Tools.Cli.csproj
```diff
@@ -51,4 +59,10 @@
   <ItemGroup>
     <EmbeddedResource Include="Images/quokka.ansi" />
   </ItemGroup>
+
+  <ItemGroup>
+    <!-- Include changelog and documentation files in the package -->
+    <None Include="CHANGELOG.md" Pack="true" PackagePath="CHANGELOG.md" />
+    <None Include="../README.md" Pack="true" PackagePath="README.md" />
```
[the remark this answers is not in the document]
scbedd: It absolutely exists.

scbedd:
image

Looking good. Ensuring `targets` updates do not [break auto versioning](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5380611&view=results) for other builds.

benbp (approved):

eng/Versioning.targets
```diff
@@ -15,7 +15,7 @@
     - `MINOR` version is in range [0-654]
     - `PATCH` version is in range [0-9999]
   -->
-  <Target Name="_InitializeFileVersion" BeforeTargets="GetAssemblyVersion">
+  <Target Name="_InitializeFileVersion" BeforeTargets="GetAssemblyVersion" Condition="'$(Version)' == ''">
```
weshaggard: I would undo this.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Azure.Sdk.Tools.Cli.csproj
```diff
@@ -11,6 +11,12 @@
     <NoWarn>ASP0000;CS8603;CS8618;CS8625;CS8604</NoWarn>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <AssemblyName>azsdk</AssemblyName>
+    <Version>0.5.0</Version>
```
weshaggard:
```suggestion
    <VersionPrefix>0.5.0</VersionPrefix>
```
I think the versioning targets in this repo use VersionPrefix instead of version. The Version property will end up getting set still from the VersionPrefix so it should accomplish the goal you want.

scbedd: It looks like `0.5.0-202X02103` or whatever if I do this. I did do this once locally.

scbedd: Let me futz around and see what I can try.

scbedd: Yeah. When I use that I get `azsdk.0.5.0-dev.20250924.1.nupkg`

scbedd: image

scbedd: Ok we got something that works.

eng/Versioning.targets
scbedd: I did undo this!

eng/scripts/Verify-Nuget-Artifact-Changelogs.ps1
```diff
@@ -0,0 +1,95 @@
+param(
```
benbp, line 1: Is there a reason we can't checkout and inspect the changelog from the source tree like we do in the language repos? Or would that be even slower since it would need to spin up an independent job and we can't have source alongside publish steps?

scbedd: It'll be roughly the same amount of code. We don't have `packageinfo`, so we will need to resolve the changelog path + the package version, which will require some custom code.

By making it work against the nupkg we can use it without any specific code that's taking the Build Path and searching for a CHANGELOG.md somewhere under it, etc.

PR #12240, opened 2025-09-24
Custom Tools for the Agent

radhgupta:
Removed AgentConversationProcessor and AgentFileManager as we've transitioned to tool-based architecture. 
Introduced a new class for conversation management that supports tool calling. 
All tools are now organized under the tool folder. Merged ErrorParsingService and FixGeneratorService into a unified ErrorAnalyzingService. The next PR will focus on applying these patches and handling rollback scenarios.

### DIFF
--- a/src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/TypeSpecFileServiceTests.cs
+++ b/src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/TypeSpecFileServiceTests.cs
@@ -369,13 +369,9 @@ namespace Azure.Tools.GeneratorAgent.Tests
             
             var validationContext = CreateGitHubValidationContext(typeSpecPath, commitId, outputDir);
             
-            // With the new pattern, we pass GitHubFileService directly instead of using a factory
             var gitHubFileService = CreateGitHubFileService();
             var service = CreateService(gitHubFileService: gitHubFileService);
 
-            // Act & Assert
-            // This test verifies that GetTypeSpecFilesAsync fails when trying to access a GitHub path as local directory
-            // The service should throw DirectoryNotFoundException because the GitHub path doesn't exist as a local directory
             Assert.That(() => service.GetTypeSpecFilesAsync(validationContext, CancellationToken.None),
                 Throws.TypeOf<DirectoryNotFoundException>()
                 .With.Message.Contains("TypeSpec directory not found"));


benbp:
FYI we are also supporting similar flows in the azsdk cli/mcp with [microagents](https://github.com/Azure/azure-sdk-tools/tree/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents) that can run a series of tools in an ongoing loop until the agent decides it's done. Example using an inline func definition: https://github.com/Azure/azure-sdk-tools/blob/779c10ae54ed54207ac7e568fbb24a8f7b568521/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Example/ExampleTool.cs#L433

m-redding:
@benbp Thanks for calling that out, with this change the agent doesn't need as many special features from persistent agents anymore so it might be a good fit, we're discussing

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -4,6 +4,7 @@
 using Azure.Tools.GeneratorAgent.Agent;
 using Azure.Tools.GeneratorAgent.Authentication;
 using Azure.Tools.GeneratorAgent.Configuration;
+using Azure.Tools.GeneratorAgent.Tools;
```
radhgupta, line 7: Just realized, DI seems to be poorly written...Learning and updating it.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ConversationManager.cs
```diff
@@ -0,0 +1,241 @@
...
+        ToolExecutor = toolExecutor ?? throw new ArgumentNullException(nameof(toolExecutor));
+        AppSettings = appSettings ?? throw new ArgumentNullException(nameof(appSettings));
+        Logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        ThreadId = string.Empty; // Will be set when conversation starts
+    }
+
+    /// <summary>
+    /// Sets the agent ID for this conversation manager
+    /// </summary>
+    public void SetAgentId(string agentId)
```
jsquire: Why not make this a setter and do the validation there?

```csharp
public string? AgentId
{
     get;
     set
     {
         ARgumentNullException.ThrowIfNull(value, nameof(AgentId));
         field = value;
      }
  }

Comment (radhgupta) [reply src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ConversationManager.cs]: Impressive!!

File: src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ToolBasedAgent.cs [hunk 3]
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /// <summary>
+    /// Fixes code based on a list of analyzer errors
+    /// </summary>
+    public async Task<Result<string>> FixCodeAsync(List<Fix> fixes, CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            if (fixes == null || fixes.Count == 0)
```
jsquire: Nothing wrong with the form that you're using, but you could shorten this to `fixes?.Count == 0` which would automatically handle the null check.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ToolExecutor.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <summary>
+    /// Executes a tool call by name with JSON arguments
+    /// </summary>
+    public async Task<string> ExecuteToolCallAsync(string toolName, string argumentsJson, CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            return toolName switch
+            {
+                "list_typespec_files" => await ExecuteListTypeSpecFilesAsync(argumentsJson, cancellationToken),
```
jsquire: These seem to be good candidates to hoist into constants.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorAnalysisService.cs
```diff
@@ -0,0 +1,183 @@
...
+            }
+
+            Logger.LogDebug("Total errors found: {TotalErrorCount} (Regex: {RegexCount}, AI: {AICount})", 
+                allErrors.Count, regexErrorsList.Count, allErrors.Count - regexErrorsList.Count);
+
+            return TransformErrorsToFixes(allErrors);
+        }
+
+
+
```
jsquire:
```suggestion

```

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Models/FileHistoryEntry.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// Version number when this content was active
+        /// </summary>
+        public int Version { get; set; }
+
+        /// <summary>
+        /// The file content at this version
+        /// </summary>
+        public string Content { get; set; } = string.Empty;
```
jsquire: Why empty instead of null for these.   The hash, specifically, violates its contract because the value "" is not a valid hash.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ConversationManager.cs
```diff
@@ -0,0 +1,251 @@
...
+        PersistentAgentsClient client,
+        ToolExecutor toolExecutor,
+        AppSettings appSettings,
+        ILogger<ConversationManager> logger)
+    {
+        Client = client ?? throw new ArgumentNullException(nameof(client));
+        ToolExecutor = toolExecutor ?? throw new ArgumentNullException(nameof(toolExecutor));
+        AppSettings = appSettings ?? throw new ArgumentNullException(nameof(appSettings));
+        Logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        ThreadId = string.Empty; // Will be set when conversation starts
```
m-redding: since this class is internal it doesn't really matter but if it doesn't always have a value to me it makes more sense having it null when not set rather than string.Empty

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ConversationManager.cs
```diff
@@ -0,0 +1,251 @@
...
+    
+        var toolOutputsData = toolOutputs.Select(to => new 
+        {
+            tool_call_id = to.ToolCallId,
+            output = to.Output
+        }).ToArray();
+        
+        var requestData = new { tool_outputs = toolOutputsData };
+        
+        await Client.Runs.SubmitToolOutputsToRunAsync(
```
m-redding, line 175: Curious - why use the protocol method for this case?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
```diff
@@ -59,9 +59,14 @@ public AppSettings(IConfiguration configuration, ILogger<AppSettings> logger)
             int.Parse(Configuration.GetSection("AzureSettings:DelayBetweenFixesMs").Value ?? "500");
         
         public string TypespecEmitterPackage => "@typespec/http-client-csharp";
+        public string TypespecCompiler => "@typespec/compiler";
         public string TypeSpecDirectoryName => "@typespec";
         public string HttpClientCSharpDirectoryName => "http-client-csharp";
 
+        // Tool names for Azure AI agent
+        public string ListTypeSpecFilesTool => "list_typespec_files";
```
m-redding: Curious - what's the scenario where it's helpful to have the tool name be customizable?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentResponseParser.cs
```diff
@@ -14,62 +14,29 @@ internal static class AgentResponseParser
             PropertyNameCaseInsensitive = true,
             AllowTrailingCommas = true
         };
-
+
```
jsquire:
```suggestion

```

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ConversationManager.cs
```diff
@@ -0,0 +1,251 @@
...
+    
+    private string? _agentId;
+    
+    public string? AgentId
+    {
+        get => _agentId;
+        set
+        {
+            ArgumentNullException.ThrowIfNull(value, nameof(AgentId));
+            _agentId = value;
```
jsquire, line 26: Ah, they moved `field` to C# 14 (.NET 10).   It was originally in the .NET 9 preview builds.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ConversationManager.cs
```diff
@@ -0,0 +1,251 @@
...
+        Logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        ThreadId = string.Empty; // Will be set when conversation starts
+    }
+
+    /// <summary>
+    /// Sets the validation context on the tool executor
+    /// </summary>
+    public void SetValidationContext(ValidationContext validationContext)
+    {
+        ToolExecutor.SetValidationContext(validationContext);
```
jsquire: Looking down the stack trace, this seems to be an invariant that the tools executor needs to do its job.   Generally, you'd want to pass that at construction time.  The intent being that once constructed, an instance should be usable.

What's the thought process behind pulling this up?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -147,13 +131,27 @@ private static IServiceCollection AddApplicationServices(this IServiceCollection
                     provider.GetRequiredService<Func<ValidationContext, GitHubFileService>>());
             });
 
+            // Note: ITypeSpecToolHandler is registered as a factory below to use proper ValidationContext
+
+            services.AddSingleton<Func<ValidationContext, ITypeSpecToolHandler>>(provider =>
```
jsquire: This is pretty loose for DI.   If we need to singleton it, we'd be better off abstracting into a class/interface.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorAnalysisService.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// Analyzes both compile and build results to generate a unified list of fixes.
+        /// </summary>
+        public async Task<Result<List<Fix>>> GenerateFixesFromResultsAsync(
+            Result<object>? compileResult, 
+            Result<object>? buildResult, 
+            CancellationToken cancellationToken)
+        {
+            var fixTasks = new List<Task<Result<IEnumerable<Fix>>>>();
+            
+            if (compileResult?.IsFailure == true && compileResult?.ProcessException?.Output != null)
```
jsquire, line 40: Is the intent to allow for multiple fix tasks to be created for a single item?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorAnalysisService.cs
```diff
@@ -0,0 +1,181 @@
...
+            }
+
+            if (buildResult?.IsFailure == true && buildResult?.ProcessException?.Output != null)
+            {
+                fixTasks.Add(GenerateFixesFromResultAsync(buildResult, cancellationToken));
+            }
+
+            if (fixTasks.Count == 0)
+            {
+                return Result<List<Fix>>.Success(new List<Fix>());
```
jsquire: nit:  We should hoist `new List<Fix>()` into a `static readonly` member so that we don't have to allocate.   Since we know its empty, we should also ensure we right-size by passing `0` as the initial size.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorAnalysisService.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        /// <summary>
+        /// Analyzes both compile and build results to generate a unified list of fixes.
+        /// </summary>
+        public async Task<Result<List<Fix>>> GenerateFixesFromResultsAsync(
+            Result<object>? compileResult, 
+            Result<object>? buildResult, 
+            CancellationToken cancellationToken)
+        {
+            var fixTasks = new List<Task<Result<IEnumerable<Fix>>>>();
```
jsquire: Do we have a sense of the max size for this upfront?   If we generate any fixes, we await them and then create a new list to return.   I'm wondering if we can use a more efficient local buffer.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ConversationManager.cs
radhgupta: The issue we're facing is that `ITypeSpecToolHandler` actually depends on `ValidationContext` (which contains the TypeSpec directory path and commit ID), but `ValidationContext` is only available at runtime when the CLI executes.

This creates a tension between clean DI (passing dependencies at construction) and our business requirement (TypeSpec operations need specific context).

We have a few options:
1. **Revert to factory pattern** - Keep `Func<ValidationContext, ITypeSpecToolHandler>` (original approach)
2. **Pass ValidationContext through method calls** - `ToolExecutor.ExecuteToolCallAsync(toolName, args, validationContext, cancellationToken)` 
3. **Create a scoped ValidationContext service** - Register ValidationContext as scoped per CLI execution

What's your preference? Option 2 seems cleanest architecturally - it makes the ValidationContext dependency explicit without hiding it in factories.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorAnalysisService.cs
m-redding: I think you could prob just simplify it and do them sequentially. Since only compile or build will have errors to analyze, (if compile fails build isn't run),

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Program.cs
```diff
@@ -117,113 +116,108 @@ private async Task<int> ExecuteGenerationAsync(ValidationContext validationConte
...
+                    // Get patch proposal from agent
+                    Logger.LogInformation("Processing fixes with Agent");
+                    var patchResult = await toolBasedAgent.FixCodeAsync(allFixes, cancellationToken).ConfigureAwait(false);
+                    
+                    if (!patchResult.IsSuccess)
+                    {
+                        Logger.LogError("Failed to get patch from agent: {Error}", 
+                            patchResult.ProcessException?.Message ?? patchResult.Exception?.Message ?? "Unknown error");
+                        break;
+                    }
```
m-redding, line 197: Is the agent currently built with the assumption of just fixing one client.tsp file? I'm just wondering since it seems the loop only allows one patched file change per iteration. I think that's a reasonable assumption for the 99% case just want to make sure I'm understanding.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecFileVersionManager.cs
```diff
@@ -0,0 +1,223 @@
...
+            {
+                FileHistory[fileName] = new List<FileHistoryEntry>();
+            }
+
+            var history = FileHistory[fileName];
+            var existingMetadata = GetFileMetadata(fileName);
+            var newHash = ComputeSha256Hash(content);
+            
+            // Check for duplicate content to avoid storing identical versions
+            var existingEntry = history.FirstOrDefault(h => h.Sha256 == newHash);
```
m-redding, line 132: Should we only not store if the latest version is identical? Maybe I missed this and you guard it in a different place, but I was just wondering about the case where you have:
- Version A
- patch request to Version B
- Version B fails so revert to Version A

We don't want Version B listed as the latest

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecPatchApplicator.cs
```diff
@@ -0,0 +1,324 @@
...
+            {
+                // Single line replacement
+                lines[startIndex] = change.NewContent;
+                Logger.LogDebug("Replaced line {Line}: '{OldContent}' → '{NewContent}'", 
+                    change.StartLine, change.OldContent.Trim(), change.NewContent.Trim());
+            }
+            else
+            {
+                // Multi-line replacement
+                var newLines = change.NewContent.Split(new[] { '\n', '\r' }, StringSplitOptions.RemoveEmptyEntries);
```
m-redding, line 264: Does the RemoveEmptyEntries mean this is removing whitespace?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecFileVersionManager.cs
m-redding: Actually I don't see where the agent uses the version numbers, is this preparing for the rollback change to come later?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ConversationManager.cs
jsquire: 2 would seem the right place to me.   If it's a runtime artifact and it is an invariant to that operation, we should pass to that operation.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorAnalysisService.cs
radhgupta: Yes. The compilationResult is basically a complete build log. The build log might contain multiple errors and for each error we will have one fix.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecFileVersionManager.cs
radhgupta: Oh yes, the version number is useful for rollback scenario, which we are yet to add.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
radhgupta: I agree, I moved it to constants class.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ConversationManager.cs
radhgupta: Azure SDK doesn't provide a typed overload for `SubmitToolOutputsToRunAsync`. The available method signatures is
 `SubmitToolOutputsToRunAsync(string threadId, string runId, RequestContent content, RequestContext context)`
There's no overload like `SubmitToolOutputsToRunAsync(string threadId, string runId, IEnumerable<ToolOutput> outputs, CancellationToken token)`
So the protocol method is the only option available in the current SDK.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Program.cs
radhgupta: Yes, I assumed thar we have only one file that needs to be patched. However, I think given that the patch json has "fileName" in future we can extend it

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecPatchApplicator.cs
radhgupta: It basically removes the empty lines that might be present in the Agents response so that when we split the response we dont have empty lines it the var newLines

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ConversationManager.cs
m-redding: Ohh ok sounds good I misread this overload - https://learn.microsoft.com/en-us/dotnet/api/azure.ai.agents.persistent.threadruns.submittooloutputstorunasync?view=azure-dotnet#azure-ai-agents-persistent-threadruns-submittooloutputstorunasync(azure-ai-agents-persistent-threadrun-system-collections-generic-ienumerable((azure-ai-agents-persistent-tooloutput))-system-threading-cancellationtoken) and thought it was the same thing

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/DependencyInjection/ServiceCollectionExtensions.cs
radhgupta: I have implemented your recommendations - eliminated all Func<ValidationContext, T> factory anti-patterns and moved to clean constructor injection with ValidationContext passed as method parameters.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecPatchApplicator.cs
m-redding: Ok that makes sense! I just remember copilot would remove all the existing whitespace that was already in the file in vs code back when agent mode was new and it could get frustrating. I think for tsp files the whitespace is not nearly as important as in C# files, but it might be something to look at later if there's any feedback about it.

m-redding (approved):

PR #12243, opened 2025-09-24
[azsdkcli] adding lintcode and formatcode infra

l0lawrence:


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/ILanguageSpecificChecks.cs
```diff
@@ -27,4 +27,22 @@ public interface ILanguageSpecificChecks
...
     Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Lints code in the specific package using language-specific tools.
+    /// </summary>
+    /// <param name="packagePath">Path to the package directory</param>
+    /// <param name="fix">Whether to automatically fix linting issues</param>
+    /// <param name="cancellationToken">Cancellation token</param>
+    /// <returns>Result of the code linting operation</returns>
+    Task<CLICheckResponse> LintCodeAsync(string packagePath, bool fix = false, CancellationToken cancellationToken = default);
```
timovv: Just a thought, maybe it would be worth creating [default implementations](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods#concrete-methods-in-interfaces) for these language specific checks that would just default to a response saying that the check is not implemented in the language? Then the language specific check classes wouldn't have to implement stub methods for the things they don't need.

benbp (approved):
LGTM though I agree with the suggestion to add a base/concrete implementation to avoid the "not implemented" duplication

Issue #12052, opened 2025-09-11
Warnings when starting the azure-sdk-mcp server


maririos:
When starting the azure-sdk-mcp server the output shows:

```
2025-09-11 11:01:24.585 [warning] Failed to parse message: "Attempting to find latest version for package 'azsdk'\r\n"
2025-09-11 11:01:25.042 [warning] Failed to parse message: "Installing 'azsdk' '1.0.0-dev.20250909.1' to 'C:\\Users\\mattge\\AppData\\Local\\Temp\\azsdk-install-8948696f-ee1b-4030-97cc-4880fc4b8b34' from [https://github.com/Azure/azure-sdk-tools/releases/download/azsdk_1.0.0-dev.20250909.1/Azure.Sdk.Tools.Cli-standalone-win-x64.zip\r\n](https://github.com/Azure/azure-sdk-tools/releases/download/azsdk_1.0.0-dev.20250909.1/Azure.Sdk.Tools.Cli-standalone-win-x64.zip/r/n)"
2025-09-11 11:01:27.449 [warning] Failed to parse message: "Package azsdk is installed at C:\\Users\\mattge\\bin\\azsdk.exe\r\n"
2025-09-11 11:01:27.468 [warning] Failed to parse message: "WARNING: To add the tool to PATH for new shell sessions, re-run with -UpdatePathInProfile to modify the shell profile file.\r\n"
```

benbp:
@maririos this is a result of the MCP client expecting ALL messages that are sent back to be json-rpc and part of the MCP schema. If we can detect when the script is run from an MCP server config, we can print messages that are wrapped in json rpc so we don't get these red herring type warnings that confuse people.

PR #12246, opened 2025-09-24
[azsdk-cli] Log output from mcp installer script as json-rpc

benbp:
Fixes https://github.com/Azure/azure-sdk-tools/issues/12052

MCP clients expect ALL messages that are sent back over stdio to be json-rpc formatted and using the MCP schema. This change assumes that if someone is passing `-Run` to the installer script they are in MCP client mode, and prints messages wrapped in json rpc so we don't get these red herring type warnings that confuse people.

AFTER:

```
2025-09-24 19:54:54.312 [info] installer: Attempting to find latest version for package 'azsdk'
2
```

BEFORE:

```
2025-09-11 11:01:24.585 [warning] Failed to parse message: "Attempting to find latest version for package 'azsdk'\r\n"
```

@maririos

### DIFF
--- a/eng/common/mcp/azure-sdk-mcp.ps1
+++ b/eng/common/mcp/azure-sdk-mcp.ps1
@@ -20,6 +20,54 @@ $ErrorActionPreference = "Stop"
 
 $toolInstallDirectory = $InstallDirectory ? $InstallDirectory : (Get-CommonInstallDirectory)
 
+$mcpMode = $Run
+
+# Log to console or MCP client json-rpc
+function log([object]$message, [switch]$warn, [switch]$err) {
+    [string]$messageString = $message
+
+    # Assume we are in an MCP client context when `-Run` is specified
+    # otherwise print to console normally
+    if (!$mcpMode) {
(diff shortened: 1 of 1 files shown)


benbp:
FYI @samvaity this will fix the annoying parse error messages in the IntelliJ mcp logs as well.

eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -20,6 +20,51 @@ $ErrorActionPreference = "Stop"
 
 $toolInstallDirectory = $InstallDirectory ? $InstallDirectory : (Get-CommonInstallDirectory)
 
+# Log to console or MCP client json-rpc
+function log([object]$message, [switch]$warn, [switch]$err) {
+    [string]$messageString = $message
+
+    # Assume we are in an MCP client context when `-Run` is specified
+    # otherwise print to console normally
+    if (!$Run) {
```
benbp: One issue I need to fix with this logic ordering in console mode

PR #12252, opened 2025-09-25
[APIView] Identify the package type

### LINKED ISSUE #12242: [APIView] Need helper method to distinguish between data and management plane
We need a helper method in the C# APIView code to distinguish whether a package is for data or management plane. We need this logic for the following languages: C#, Go, Python, Java, JavaScript and TypeSpec.

For all but TypeSpec, we can use the package name as a heuristic to determine the kind.

**C#**
Data:
Management:

helen229:
Closes  #12242


src/dotnet/APIView/APIViewWeb/Helpers/CommonUtilities.cs
```diff
@@ -69,6 +69,126 @@ public static bool IsSDKLanguageOrTypeSpec(string language)
         }
     }
 
+    /// <summary>
+    /// Represents the plane classification of a package
+    /// </summary>
+    public enum PackageType
```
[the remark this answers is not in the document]
tjprescott: Don't rename to PlaneType.

src/dotnet/APIView/APIViewWeb/Helpers/CommonUtilities.cs
```diff
@@ -69,6 +69,126 @@ public static bool IsSDKLanguageOrTypeSpec(string language)
...
+                    // emitter-output-dir pattern: contains /az in path
+                    else if (Regex.IsMatch(packageName, @"(^|.*/)sdk/", RegexOptions.IgnoreCase) ||
+                            Regex.IsMatch(packageName, @"/az[^/]*(/|$)", RegexOptions.IgnoreCase))
+                    {
+                        return PackageType.Data;
+                    }
+                    // Unknown: packages that don't match expected Go patterns
+                    return PackageType.Unknown;
+
+                default:
```
tjprescott: Where's the logic for TypeSpec?

helen229: Added

helen229:
> why are we trying to identify this SDK type in APIView? This logic should not be in APIView. Package type is detected and set by pipeline which originates API view request. We should pass the SDK type from pipeline instead of having a package name based parsing logic here. This logic will break if SDK team decides to add a new format for management or data plane types. So having this logic here is an unknown redundancy.

We discussed the approach of passing type as parameter in APIview endpoint call or editing parser to include type meta data, and decided to handle package type classification in APIview. We can have a quick call to review this with Travis and Ali.

src/dotnet/APIView/APIViewWeb/Helpers/CommonUtilities.cs
```diff
@@ -69,6 +71,127 @@ public static bool IsSDKLanguageOrTypeSpec(string language)
...
+            }
+
+            // Normalize language name for comparison
+            var normalizedLanguage = language.Trim().ToLowerInvariant();
+            
+            switch (normalizedLanguage)
+            {
+                case "c#":
+                    // Management: strings that match regex /^Azure\.ResourceManager\./
+                    if (packageName.StartsWith("Azure.ResourceManager.", StringComparison.Ordinal))
```
praveenkuttappan: We should not do this filtering here. what happens if management plane team decides to add a new prefix in the future?  They won't know that there is a dependency in APIView side for such namespace prefix change. Each package already has such info and that is available as part of package info in the pipeline. more than 95% of our API reviews are generated from pipeline. If this info is missing for manually created reviews, then just tag it as "Unknown" type. But I strongly recommend not doing this in APIView side.

FYI @maririos, @tjprescott, @AlitzelMendez

src/dotnet/APIView/APIViewWeb/Helpers/CommonUtilities.cs
```diff
@@ -69,6 +71,127 @@ public static bool IsSDKLanguageOrTypeSpec(string language)
...
+                        return PackageType.Management;
+                    
+                    // Data: strings that start with "Azure." (but not management plane)
+                    if (packageName.StartsWith("Azure.", StringComparison.Ordinal))
```
praveenkuttappan: This is the exact reason I am against this approach of doing this logic in APIView. There are unbranded libraries that does not have name start with Azure.

tjprescott:
We can probably just close this PR, can't we if we are pursuing passing the SDKType parameter in?

tjprescott:
Closing in favor of https://github.com/Azure/azure-sdk-tools/pull/12398

Issue #12178, opened 2025-09-19
[APIView] Remove RuleIds from AIReviewModel


AlitzelMendez:
After Monday  release, follow up with this issue to remove the property RuleIds from the model AIReviewModel as it will no longer be needed/used.

Keeping for this release to avoid any disruption during the release

PR #12261, opened 2025-09-26
[APIView] Remove RuleIds from AIReviewMode

AlitzelMendez:
closes https://github.com/Azure/azure-sdk-tools/issues/12178


src/dotnet/APIView/APIViewWeb/Models/AIReviewModel.cs
```diff
@@ -10,21 +10,8 @@ namespace APIViewWeb.Models
...
-        }
-        
-        [JsonPropertyName("guideline_ids")]
-        public List<string> GuidelineIds 
-        { 
-            get => _guidelineIds; 
-            set => _guidelineIds = value; 
-        }
+        [JsonPropertyName("guideline_ids")] 
+        public List<string> GuidelineIds { get; set; }
```
tjprescott, line 14: Don't we have comments in the DB that are still stored with `rule_ids`?

AlitzelMendez: do you mean from copilot?

from api view perspective, no, we just started storing these values and it has been stored under "GuidelineIds" even if the value came from "rules_ids"

tjprescott: Ah, how were they stored before? Did we just stuff the guideline link into the text of the comment?

AlitzelMendez: yes, it was just part of the comment :)

AlitzelMendez: for example 
```
 "CommentText": "Async long-running operations should return an AsyncLROPoller instead of the final result type for consistency with guidelines and the synchronous counterpart.\r\n\r\n\r\nSuggestion : `) -> AsyncLROPoller[DeletedSecret]`\r\n\r\n\r\n**Guidelines**\r\n\r\nhttps://azure.github.io/azure-sdk/python_design.html#python-lro-poller\r\n",
```

tjprescott (approved):

PR #12262, opened 2025-09-26
[API View] Identity compliance

AlitzelMendez:
Fixing report by SFI dashboard

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/APIViewWeb.csproj
+++ b/src/dotnet/APIView/APIViewWeb/APIViewWeb.csproj
@@ -39,7 +39,8 @@
     </PackageReference>
     <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.22.0" />
     <PackageReference Include="Microsoft.ApplicationInsights.SnapshotCollector" Version="1.4.5" />
-    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="8.0.4" />
+    <PackageReference Include="Microsoft.Identity.Web" Version="2.16.1" />
+    <PackageReference Include="Microsoft.Identity.Web.MicrosoftGraph" Version="2.16.1" />
     <PackageReference Include="Microsoft.Azure.AppConfiguration.AspNetCore" Version="7.0.0" />
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.40.0" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="7.0.15" />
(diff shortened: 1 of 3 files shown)


benbp:
I don't have much working knowledge of the graph library implementation, but it seems like the old jwt token logic restricted the issuers to a specific tenant. Am I reading that correctly, and if so, does the new logic also do that (and/or is it necessary)?

Possibly via https://learn.microsoft.com/en-us/azure/app-service/scenario-secure-app-access-microsoft-graph-as-user?tabs=azure-resource-explorer#appsettingsjson

AlitzelMendez:
> I don't have much working knowledge of the graph library implementation, but it seems like the old jwt token logic restricted the issuers to a specific tenant. Am I reading that correctly, and if so, does the new logic also do that (and/or is it necessary)?
> 
> Possibly via https://learn.microsoft.com/en-us/azure/app-service/scenario-secure-app-access-microsoft-graph-as-user?tabs=azure-resource-explorer#appsettingsjson

yes, the new logic also does that, using this configuration 
image


that it is what you shared, it is just that the values are on App Configuration

praveenkuttappan (approved):

NuGet.Config
```diff
@@ -3,6 +3,7 @@
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="IDDP" value="https://identitydivision.pkgs.visualstudio.com/_packaging/IDDP/nuget/v3/index.json" />
```
praveenkuttappan, line 6: IIRC this will most likely get flagged as multi source feed security issue which was an SFI issue we resolved in the past. @weshaggard FYI.

AlitzelMendez: @weshaggard  I’d love your feedback here.

This is the guidance for MISE usage:
https://eng.ms/docs/microsoft-security/identity/entra-developer-application-platform/id4s-identity-for-services/authn-middleware-sdk-microsoft-identity-service-essentials/microsoft-identity-service-essentials/articles/migration-guides/jwtbearerhandler/readme

It was added because the code was flagged for not using it during token validation:
https://eng.ms/docs/microsoft-security/identity/entra-developer-application-platform/id4s-identity-for-services/authn-middleware-sdk-microsoft-identity-service-essentials/microsoft-identity-service-essentials/articles/tsgs/mise-compliance-kpi-tsg

Is adding another source feed something we can do? Or do we need to request an exception?

weshaggard: Correct we should not add another feed here for security reasons. I'm actually soon going to be moving away from nuget.org and moving to our devops feed at that point we could put the package into that feed but until then I'm not sure we have a good solution to consuming this. It is strange a security team wants us to consume a package from another feed.

weshaggard: `Response status code does not indicate success: 401 (Unauthorized).` We don't even have permissions to access that feed currently.

AlitzelMendez:
we are trying a different approach for this!

PR #12263, opened 2025-09-27
[tsp-client] Fix init or update scenario

catalinaperalta:
This PR includes the following:
- A fix for the init or update scenario when there's already an existing emitterPackageJsonPath configured in a tsp-location.yaml
- Some test fixes and clean up
- Fix tsp-client init with emitter-output-dir when an alternate output-dir is passed in.

### DIFF
--- a/tools/tsp-client/src/utils.ts
+++ b/tools/tsp-client/src/utils.ts
@@ -109,6 +109,7 @@ export async function writeTspLocationYaml(
 export async function updateExistingTspLocation(
   tspLocationData: TspLocation,
   projectPath: string,
+  emitterPackageJsonOverride?: string,
 ): Promise<TspLocation> {
   try {
     const existingTspLocation = await readTspLocation(projectPath);
@@ -117,13 +118,13 @@ export async function updateExistingTspLocation(
     const updatedTspLocation = { ...existingTspLocation };
 
     // Define the properties that can be updated
+    // NOTE: emitterPackageJsonPath is handled separately below because of the override logic
     const updatableProperties: (keyof TspLocation)[] = [
       "repo",
       "commit",
       "directory",
       "entrypointFile",
       "additionalDirectories",
-      "emitterPackageJsonPath",
     ];
 
     // Update each property if it has a valid value
@@ -134,6 +135,19 @@ export async function updateExistingTspLocation(
       }
     }
 
+    // Only add/replace emitterPackageJsonPath value if an override is provided, otherwise keep existing value or leave undefined
+    if (emitterPackageJsonOverride) {
+      if (existingTspLocation.emitterPackageJsonPath !== undefined) {
+        Logger.debug(
+          `Updating existing emitterPackageJsonPath ${existingTspLocation.emitterPackageJsonPath} with ${emitterPackageJsonOverride} in tsp-location.yaml`,
+        );
+      } else {
+        Logger.debug(
+          `Adding emitterPackageJsonPath ${emitterPackageJsonOverride} to tsp-location.yaml`,
+        );
+      }
+      updatedTspLocation.emitterPackageJsonPath = emitterPackageJsonOverride;
+    }
     return updatedTspLocation;
   } catch (error) {
     Logger.debug(`Will create a new tsp-location.yaml. Error reading tsp-location.yaml: ${error}`);
--- a/tools/tsp-client/src/utils.ts
+++ b/tools/tsp-client/src/utils.ts
@@ -1,12 +1,13 @@
 import { joinPaths, normalizeSlashes } from "@typespec/compiler";
 import { randomUUID } from "node:crypto";
 import { access, constants, mkdir, readFile, writeFile } from "node:fs/promises";
-import { dirname, join } from "node:path";
+import { dirname, join, relative } from "node:path";
 import { fileURLToPath } from "node:url";
 import { Logger } from "./log.js";
 import { TspLocation } from "./typespec.js";
 import { normalizeDirectory, readTspLocation } from "./fs.js";
 import { parse as parseYaml } from "yaml";
+import { getRepoRoot } from "./git.js";
 
 const defaultTspClientConfigPath = joinPaths("eng", "tspclientconfig.yaml");
 
@@ -146,7 +147,9 @@ export async function updateExistingTspLocation(
           `Adding emitterPackageJsonPath ${emitterPackageJsonOverride} to tsp-location.yaml`,
         );
       }
-      updatedTspLocation.emitterPackageJsonPath = emitterPackageJsonOverride;
+      updatedTspLocation.emitterPackageJsonPath = normalizeSlashes(
+        relative(await getRepoRoot(projectPath), emitterPackageJsonOverride),
+      );
     }
     return updatedTspLocation;
   } catch (error) {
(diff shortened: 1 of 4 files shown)


tools/tsp-client/src/utils.ts
```diff
@@ -117,13 +119,13 @@ export async function updateExistingTspLocation(
...
 
     // Define the properties that can be updated
+    // NOTE: emitterPackageJsonPath is handled separately below because of the override logic
     const updatableProperties: (keyof TspLocation)[] = [
       "repo",
       "commit",
       "directory",
       "entrypointFile",
       "additionalDirectories",
-      "emitterPackageJsonPath",
```
JoshLove-msft, line 126: So was the issue that the generated tsp-location would have the emitter specified based on the repo default?

catalinaperalta: The issue was because we follow the default init flow to update these variables (commit, directory, etc). For example, you'll notice that we update the commit based on the new commit in the tspconfig.yaml url or the commit passed in through the flag. So for emitterPackageJsonPath we also followed the regular init flow to find the first C# emitter in the tspconfig.yaml file and get its corresponding path. In your case where you have 2 c# emitters in the tspconfig.yaml, the init flow still needs to find the emitter to use and, as we had all agreed before, tsp-client will stop at the first instance of a language supported emitter it finds, which in this case is the http-client-csharp emitter which was new in the tspconfig.yaml file and it has a different emitter-package.json path which was then updated in the tsp-location.yaml file. That's why last week I was asking if the path to the client library of all the emitters will be the same, because if not we'd need to find some other solution for this problem. For now, (until it becomes a problem) we'll check for the tsp-location.yaml file at the path specified by the first language supported emitter and we dont update the variable unless a path is purposefully passed to the init command with the `emitter-package-json-path` flag when used with the `update-if-exists` flag.

JoshLove-msft: > as we had all agreed before, tsp-client will stop at the first instance of a language supported emitter it finds, which in this case is the http-client-csharp emitter which was new in the tspconfig.yaml file and it has a different emitter-package.json path which was then updated in the tsp-location.yaml file. 

This was only for the init case, i.e. it should not be applied if --update-if-exists was being set. Maybe I'm missing something, but I thought the whole point of this flag was for this very scenario - respecting the existing emitterPackageJsonPath in tspLocation. If that is not the case, why would someone pass this flag?

JoshLove-msft: > For now, (until it becomes a problem) we'll check for the tsp-location.yaml file at the path specified by the first language supported emitter

How is the issue of having multiple emitters for one language in a config made more of a problem by the need to not overwrite the emitterPackageJsonPath? We need to find the existing tsp-location.yaml if someone passes update-if-exists regardless, right?

catalinaperalta: How can we find the client library path if we're not running from the tsp-location.yaml directory? Even with tsp-client update, if you're running outside of the tsp-location.yaml directory you'd have to manually pass the output-dir. For init, we are trying to magically do that all for you under the hood...so we still need to base ourselves off the of the information we get into the command. In this case we only have access to the tspconfig.yaml file, how do we know which emitter to use? Once we know what emitter to use then yes we can work based off of the tsp-location.yaml file because we'll know where it is checked in.

catalinaperalta: Also, the CI pipelines want to test the specs (including existing spec) at a newer commit, so that's why the other variables can be updated in tsp-location.yaml as needed

tools/tsp-client/test/commands.spec.ts
```diff
@@ -400,33 +405,113 @@ describe.sequential("Verify commands", () => {
...
+      const outputDir = await initCommand(args);
+      const tspLocation = await readTspLocation(outputDir);
+      assert.deepEqual(tspLocation, {
+        directory: "specification/contosowidgetmanager/Contoso.WidgetManager",
+        commit: "abc",
+        repo: "Azure/azure-rest-api-specs",
+        additionalDirectories: [
+          "tools/tsp-client/test/examples/specification/contosowidgetmanager/Contoso.WidgetManager.Shared",
+        ],
+        emitterPackageJsonPath: "tools/tsp-client/test/utils/emitter-package.json",
```
JoshLove-msft, line 420: Is this the default emitterPackageJson? If it is, can we add another test case where there is a different existing emitterPackageJsonPath to verify it isn't changed?

tools/tsp-client/src/utils.ts
catalinaperalta: As you saw in the code, we have a list of the variables we allow to be modified when update-if-exists is passed in, we can choose to relax or restrict the ability for these variables to be updated as needed. For now I'd say we leave the others as changeable unless we find a reason to not do that

JoshLove-msft: > Once we know what emitter to use then yes we can work based off of the tsp-location.yaml file because we'll know where it is checked in.

I think this is the crux of the issue - the tsp-location itself communicates which emitter to use. This was the motivation behind --update-if-exists. For every other piece of data, it should be completely overwritten. Though it's possible there will be other properties in the future we don't want to overwrite so having a list of updatable properties makes sense.

tools/tsp-client/test/commands.spec.ts
catalinaperalta: The default emitterPackageJson is the eng/emitter-package.json path and we dont add that to the tsp-location.yaml. I believe this test is checking what you're asking me to check. Line 399-407 we define the "existing" tsp-location.yaml so we have this value defined already and on this line we're verifying that it wasnt changed. Is that what you wanted to test?

tools/tsp-client/test/commands.spec.ts
```diff
@@ -400,33 +405,113 @@ describe.sequential("Verify commands", () => {
...
+      assert.deepEqual(tspLocation, {
+        directory: "specification/contosowidgetmanager/Contoso.WidgetManager",
+        commit: "abc",
+        repo: "Azure/azure-rest-api-specs",
+        additionalDirectories: [
+          "tools/tsp-client/test/examples/specification/contosowidgetmanager/Contoso.WidgetManager.Shared",
+        ],
+        // The emitterPackageJsonPath should be updated to the alternate emitter package path passed through the
+        // emitter-package-json-path argument
+        emitterPackageJsonPath: "tools/tsp-client/test/utils/alternate-emitter-package.json",
```
catalinaperalta, line 468: @JoshLove-msft this is the test for when we do override the value in the file

tools/tsp-client/test/commands.spec.ts
JoshLove-msft: Yep, makes sense, thanks!

tools/tsp-client/test/commands.spec.ts
```diff
@@ -400,33 +405,113 @@ describe.sequential("Verify commands", () => {
...
+      };
+
+      // Add a tsp-location.yaml file to the output directory for the initOrUpdate test to simulate an existing project
+      const existingTspLocation: TspLocation = {
+        directory: "specification/contosowidgetmanager/Contoso.WidgetManager",
+        commit: "45924e49834c4e01c0713e6b7ca21f94be17e396",
+        repo: "Azure/azure-rest-api-specs",
+        additionalDirectories: [
+          "tools/tsp-client/test/examples/specification/contosowidgetmanager/Contoso.WidgetManager.Shared",
+        ],
```
catalinaperalta, line 496: This is the test where the default eng/emitter-package.json is used

tools/tsp-client/src/utils.ts
catalinaperalta: Yeah, so I think this PR is addressing that problem (in part*), so we still need some mechanism to find the tsp-location.yaml file, we stop at the first language supported emitter, look for the client library path and we will not update the emitterPackageJsonPath variable anymore unless the flag is passed.

*in part because if the first emitter changes the client library path, we'll need to find a new way to specify which emitter in tspconfig.yaml to use to get information about the target client library

JoshLove-msft (approved):

Issue #12138, opened 2025-09-17
[Release Plan]:  SDK generation failed for ComputeSchedule SDK


lirenhe:
This is the release planner :  https://apps.powerapps.com/play/e/a49cfcc0-87c2-469b-9516-da85bf6474e6/a/06e7ff11-439f-4a49-84c2-a70dfbecd33c?workflow=d4f5c6f9-b193-f011-b4cc-000d3a9a072d&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47

This is the SDK generation PR:  https://github.com/Azure/azure-rest-api-specs/pull/37523

The below information is quite confused:

<img width="804" height="564" alt="Image" src="https://github.com/user-attachments/assets/59f5a421-de9f-4835-ad86-4813d1cc2e79" />

1.   Why I could not generate SDK PRs for an older api version?  
2. Copilot refer some PRs related to other RPs, not sure the relationship with the current release.

praveenkuttappan:
@lirenhe Copilot failed to download the MCP server from azure-sdk-tools repo so it didn't even trigger SDK generation pipeline. I will add FAQ document to outline the steps a user can take if they run into these types of random issues.

I have retried your request by adding a comment in the pull request created by copilot for the work and tagged copilot. This prompted copilot to rerun the task and generated SDK for 4 out of 5 languages. SDK generation failed for JS and release planner app shows link to the failure details.

You can access the release planner to see these details.


Error from initial run:
```Attempting to find latest version for package 'azsdk'
Invoke-RestMethod: /home/runner/work/azure-rest-api-specs/azure-rest-api-specs/eng/common/scripts/Helpers/AzSdkTool-Helpers.ps1:142
Line |
 142 |          $releases = Invoke-RestMethod -Uri $releasesUrl
     |                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     |  {   "message": "API rate limit exceeded for 172.215.209.69. (But
     | here\u0027s the good news: Authenticated requests get a higher rate
     | limit. Check out the documentation for more details.)",  
     | "documentation_url":```

Comment I used to retrigger SDK gen:  https://github.com/Azure/azure-rest-api-specs/pull/37523#issuecomment-3304376083

praveenkuttappan:
Here are the action items for me:
- [ ] I will make a change to use authenticated requests to avoid rate limit errors.
- [x] I have a issue to generate a FAQ document and link it in the release planner.

[...]

PR #12267, opened 2025-09-29
Use authenticated GitHub request when env variable is present

praveenkuttappan:
Send GitHub bearer header if token env variable is present to avoid rate limit for unauthenticated requests

### DIFF
--- a/eng/common/scripts/Helpers/AzSdkTool-Helpers.ps1
+++ b/eng/common/scripts/Helpers/AzSdkTool-Helpers.ps1
@@ -107,6 +107,7 @@ function isNewVersion(
 
 function Get-GitHubApiHeaders {
     # Use GitHub cli to get an auth token if available
+    $token = ""
     if (Get-Command gh -ErrorAction SilentlyContinue) {
         try
         {


praveenkuttappan:
Fixes #12138

eng/common/scripts/Helpers/AzSdkTool-Helpers.ps1
```diff
@@ -105,6 +105,19 @@ function isNewVersion(
     return $true
 }
 
+function Get-GitHubApiHeaders {
+    $token = $env:GITHUB_TOKEN
```
weshaggard: Where do you expect this to get set?

praveenkuttappan: This will be used when copilot is triggered from GitHub.com and it will use GitHub copilot's built in token. This will be set via mcp configuration in azure-rest-api-specs which already has the env variable for copilot. Currently I have noticed rate limit error when testing code generation using copilot.

benbp: Do we need to set this and have it configured? `gh auth token` should be available for github contexts.

praveenkuttappan: I will have to change the copilot instructions to set the env variable so script can use it.

benbp: Why is the env variable is necessary at all though?

Issue #12259, opened 2025-09-26
`multipart/formdata` update broke `java - openai`


scbedd:
[Failure viewable here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5372221&view=ms.vss-test-web.build-test-results-tab&runId=56989138&resultId=100455&paneView=debug)

Ever since we released on 9/22. The errors are in the code we changed for this release.

Need to identify resolution and re-sync with @alzimmermsft

scbedd:
- The change in the proxy that is responsible for the new break is that we now treat multipart/formdata the same as multipart/mixed. that means we're parsing the content for sanitization when we weren't before.
- The request we're matching against is a big 33MB. and now that we SANITIZE that it's running slooooooooooooooooooooooow. This is where the timeouts are coming from.
- We're blowing up in sanitization due to some segment being longer than 16KB, which is breaking the asp.net multipartreader. this is the source of the 500 errors. This is likely due to a body that has been erroneusly sanitized in such a way that it broke reading. I'm working on understanding where this is going wrong.

scbedd:
Fixed in Azure/azure-sdk-for-java#46886

PR #12273, opened 2025-09-29
Follow-up on form-data adjustments for `java`

scbedd:
Contributes to resolving #12259

Timeline of events:

- We merged a PR that treats `multipart/*` the same as `multipart/mixed`.
- Incoming requests that we match against are slow to sanitize, because they are so big.
- We ignore the body anyway in the configured matcher.

So, this PR includes a couple changes.

1. It will detect a multipart boundary from the body content if the sanitized value if `BOUNDARY` now. We didn't used to. (this was the cause of the failure to parse multipart data we were seeing in java openai tests)
2. Now when we are not comparing the body anyway, we do not sanitize the incoming request body based on the activated sanitizers. There is no reason to given that we are ignoring the body during matching anyway.
3. Request/Response bodies will **still** be sanitized when running `Record/Stop`. This is intentional as we don't know if the request body has anything private in it. From a defensive standpoint this is the right thing to do, and it won't reflect negatively on runtime performance.

Will leave this PR open while I publish a version that can be utilized in java CI and confirm that it all checks out.

### DIFF
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/MultipartUtilities.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/MultipartUtilities.cs
@@ -110,7 +110,7 @@ namespace Azure.Sdk.Tools.TestProxy.Common
         public static string ResolveFirstBoundary(string boundary, byte[] raw)
         {
             // Boundary might have been sanitised to "REDACTED"
-            if (boundary == "REDACTED" || boundary.EndsWith("00000000-0000-0000-0000-000000000000"))
+            if (boundary == "REDACTED" || boundary == "BOUNDARY" || boundary.EndsWith("00000000-0000-0000-0000-000000000000"))
             {
                 ReadOnlySpan<byte> crlf = stackalloc byte[] { 0x0D, 0x0A };
                 int idx = raw.AsSpan().IndexOf(crlf);
--- a/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordMatcher.cs
+++ b/tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordMatcher.cs
@@ -26,17 +26,24 @@ namespace Azure.Sdk.Tools.TestProxy.Common
             "Content-Type"
         };
 
-        private bool _compareBodies;
-        private bool _ignoreQueryOrdering;
+        /// <summary>
+        /// When true, message bodies are compared during matching. In addition, when true incoming request bodies will not be sanitized.
+        /// </summary>
+        public bool ShouldCompareBodies { get; set; }
+
+        /// <summary>
+        /// When true, query parameter ordering is ignored during URI normalization.
+        /// </summary>
+        public bool ShouldIgnoreQueryOrdering { get; set; }
 
         public RecordMatcher(bool compareBodies = true, bool ignoreQueryOrdering = false)
         {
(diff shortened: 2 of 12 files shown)


scbedd:
[Confirmed fixed!](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5399889&view=results)

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordMatcher.cs
```diff
@@ -26,17 +26,24 @@ public class RecordMatcher
...
-        private bool _ignoreQueryOrdering;
+        /// <summary>
+        /// When true, message bodies are compared during matching. In addition, when true incoming request bodies will be sanitized.
+        /// </summary>
+        public bool ShouldCompareBodies { get; set; }
+
+        /// <summary>
+        /// When true, query parameter ordering is ignored during URI normalization.
+        /// </summary>
+        public bool ShouldIgnoreQueryOrdering { get; set; }
```
alzimmermsft: Is there a time when we want these to be settable after initialization?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/MultipartUtilities.cs
```diff
@@ -110,7 +110,7 @@ public static byte[] NormalizeBareLf(byte[] src)
         public static string ResolveFirstBoundary(string boundary, byte[] raw)
         {
             // Boundary might have been sanitised to "REDACTED"
-            if (boundary == "REDACTED" || boundary.EndsWith("00000000-0000-0000-0000-000000000000"))
+            if (boundary == "REDACTED" || boundary == "BOUNDARY" || boundary.EndsWith("00000000-0000-0000-0000-000000000000"))
```
alzimmermsft, line 113: Do all other matching systems work like this where there's a hidden set of allowed redacted values?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordMatcher.cs
scbedd: Possibly, but it's some nebulous "maybe." Realistically swapping the setters to private changes nothing and is more "correct".

```suggestion
        public bool ShouldCompareBodies { get; private set; }

        /// <summary>
        /// When true, query parameter ordering is ignored during URI normalization.
        /// </summary>
        public bool ShouldIgnoreQueryOrdering { get; private set; }
```

scbedd: This responds to your feedback tho 👍

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/MultipartUtilities.cs
scbedd: This is backwards compatibility with all our existing recordings. I simply don't have the time to go back and re-record all of these when I have another looming deadline.

scbedd: The proper solution is to eliminate Content-Type: --boundary=X sanitization, as we modify the Content-Type headers to ignore those boundary values now that we properly compare body bits minus boundaries during matching.

benbp (approved):

PR #12280, opened 2025-09-30
Update csharp parser version

AlitzelMendez:


praveenkuttappan:
I think you will also need to update the version in CSharpLanguageService if current value matches with value in the DB (APIRevisions).

praveenkuttappan (approved):

AlitzelMendez:
> I think you will also need to update the version in CSharpLanguageService if current value matches with value in the DB (APIRevisions).

yes, this is the case, updating! :)
image

Issue #10413, opened 2025-04-23
Expose APIView functionality to tools/mcp agents


maririos:
i.e: Getting the structure of APIs, getting diffs between versions, getting the last published version, list versions, post comment. 

By doing this, different tools can consume the APIs and take advantage of the functionality APIView offers

AlitzelMendez:
There are couple of questions around this.

We have endpoints ready to expose or in where the expected result is clear (last published version, list versions), but for others we need a more thorough analysis and we need to know better the scenarios in where we consider we can use it, for example getting diffs between versions, what is the expectation? just to return a plain text with `+`, `-`? or a token file with the diff information? is just the text enough or line id will become important for the people asking for this? 

so I will start by doing the basic, setting up the known scenarios so we have a starting point and adding more scenarios after that becomes easier. 

After that I will reach out to the teams with specific requests :)

AlitzelMendez:
adding other small requests that can be part of this pr: 
- https://github.com/Azure/azure-sdk-tools/issues/11850
- https://github.com/Azure/azure-sdk-tools/issues/11849

tjprescott:
> adding other small requests that can be part of this pr:
> 
> * [[APIView] CLI + API that returns a token file given the revision ID #11850](https://github.com/Azure/azure-sdk-tools/issues/11850)
> * [[APIView] CLI tool + REST endpoint to get the APIView line associated with the elementId #11849](https://github.com/Azure/azure-sdk-tools/issues/11849)

Not sure these make sense as MCP tools. The second issue is, AFAICT, more related to the output format of the getComments API.

AlitzelMendez:
completed at: https://github.com/Azure/azure-sdk-tools/pull/12111

PR #12288, opened 2025-09-30
[API View]  Add endpoint receiving package name as input for mcp effort

AlitzelMendez:
Part of the work necessary for https://github.com/Azure/azure-sdk-tools/issues/10413 

Users are more familiar with the package name and the language that for a specific revision id.
Adding endpoints the receive this information for:
- get comments (it will exposed on the MCP)
- get content  (it will be only exposed as part of the CLI)

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/LeanControllers/APIRevisionsTokenAuthController.cs
+++ b/src/dotnet/APIView/APIViewWeb/LeanControllers/APIRevisionsTokenAuthController.cs
@@ -76,7 +76,7 @@ public class APIRevisionsTokenAuthController : ControllerBase
     }
 
     /// <summary>
-    ///     Generate review tex or code file for a package by package name, language, and version
+    ///     Generate review text or code file for a package by package name, language, and version
     /// </summary>
     /// <param name="packageName">The package name</param>
     /// <param name="language">The programming language</param>


src/dotnet/APIView/APIViewWeb/Helpers/ControllerHelpers.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        ReviewListItemModel review = await cosmosReviewRepository.GetReviewAsync(language, packageName);
+        if (review == null)
+        {
+            LeanJsonResult errorResult = new($"Package name {packageName} for language {language} not found", StatusCodes.Status404NotFound);
+            return (null, null, errorResult);
+        }
+
+        APIRevisionListItemModel revision = string.IsNullOrEmpty(version)
+            ? await apiRevisionsManager.GetLatestAPIRevisionsAsync(review.Id)
```
praveenkuttappan, line 24: How does this calculate latest revision? Latest automatic?

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsTokenAuthController.cs
```diff
@@ -18,30 +20,90 @@ namespace APIViewWeb.LeanControllers;
...
+        {
+            (ReviewListItemModel _, APIRevisionListItemModel revision, ActionResult errorResult) = await ControllerHelpers.GetReviewAndRevisionAsync(
+                _cosmosReviewRepository, _apiRevisionsManager, packageName, language, version);
+
+            if (errorResult != null)
+            {
+                return errorResult;
+            }
+
+            return await GetRevisionComments(revision);
```
praveenkuttappan, line 92: Why does this returns comments specific to a revision only when request is about the comments for the entire package?

src/dotnet/APIView/APIViewWeb/Helpers/ControllerHelpers.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.AspNetCore.Mvc;
+
+namespace APIViewWeb.Helpers;
+
+public static class ControllerHelpers
+{
+    public static async Task<(ReviewListItemModel review, APIRevisionListItemModel revision, ActionResult errorResult)> GetReviewAndRevisionAsync(ICosmosReviewRepository cosmosReviewRepository, 
+        IAPIRevisionsManager apiRevisionsManager, string packageName, string language, string version = "")
+    {
+        ReviewListItemModel review = await cosmosReviewRepository.GetReviewAsync(language, packageName);
```
tjprescott, line 16: How are we using package name? Is this just a string comparison so that Python needs to exactly match "azure-storage-blob" and some other language needs a different format?  If so, I fear this is going to be a pit of failure, not success.

If we are going to allow "language and package name" it should be fully support fuzzy matching and natural language queries. I should be able to say (in my prompt) "Get me the Python APIView for Storage" and have it work.

To me, that's an enhancement beyond the scope of what we're trying to do by getting our existing endpoints into MCP/CLI.

AlitzelMendez: I don't think it's achievable to fully support natural language. Even if I filter by storage, the question would still be: which storage? Storage Queue? Storage File Share? Storage Blob? Storage Blob Change Feed?

So, we can probably return a list of possible matches, with the expectation that the user will pick one of them for a full match. This way, we can provide the name. Bottom line: we need to know which package name matches. We can always check if there is a close enough match and ask for confirmation, or if there is a full match, just return the result.

The reasoning behind this change is that the expectation is for people to easily interact with the API view. Asking for a review/revision ID wasn't the best approach, as many people are not familiar with those concepts. The midpoint was using the URL, but there was some hesitation around its limitations.

src/dotnet/APIView/APIViewWeb/LeanControllers/CommentsTokenAuthController.cs
AlitzelMendez: Yes, this is not super visible in this particular section of the code, but what is happening is that:

1. We are getting the comments for the entire review on: `IEnumerable<CommentItemModel> comments = await _commentsManager.GetCommentsAsync(apiRevision.ReviewId, false, CommentType.APIRevision);`

2. Then when we are building the comments for the agent, we are sharing these comments from the entire review and the code file. The filtering is just picking the comments that are attached to one line of that revision.

I understand why this is confusing. I will break `AgentHelpers.BuildCommentsForAgent(comments, codeFile);` into different methods to isolate the filtering code and make this clearer.

src/dotnet/APIView/APIViewWeb/Helpers/ControllerHelpers.cs
AlitzelMendez: latest created overall

AlitzelMendez, line 24: small feedback

AlitzelMendez:
no longer neded

PR #12305, opened 2025-10-01
[azsdkcli] Some improvements to microagents

timovv:
A couple of small improvements here:

- As a follow up to #12302, add a `T is notnull` constraint to `MicroagentResult`. This is necessary to make sure the JSON schema exporter marks the result property as not only required, but also as non-nullable (these are distinct, but we need both to ensure the result is populated). 
- Add a `ValidateResult` callback to the microagent defintion. This allows for people to add a pre-exit validation check. If that validation check fails, the existing conversation thread continues, with a response telling the LLM to keep trying and providing a reason for why the validation failed. This allows the agent to keep iterating without losing potentially valuable conversation history. (@radhgupta, this is along the lines of what we were talking about yesterday, so might be helpful if you ever were to use this API. Thanks for the idea!)


radhgupta:
Thanks @timovv!!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Example/ExampleTool.cs
```diff
@@ -464,6 +464,26 @@ public async Task<DefaultCommandResponse> DemonstrateMicroagentFibonacci(int n,
...
                 Tools = [advanceTool],
+                ValidateResult = async result =>
+                {
+                    // Check the result for correctness using fibonacci formula.
+                    var phi = (1 + Math.Sqrt(5)) / 2;
+                    var psi = (1 - Math.Sqrt(5)) / 2;
+                    var expected = (int)Math.Round((Math.Pow(phi, n) - Math.Pow(psi, n)) / Math.Sqrt(5));
+
+                    // Example validation
+                    if (result != expected)
```
l0lawrence, line 477: so this is for deterministic output results (like scripts being run etc.)?

l0lawrence (approved):
lgtm 1 question

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Example/ExampleTool.cs
timovv: Yeah, so taking spelling for example, let's say you are expecting the agent to update the cspell file before it exits. In that case you could add a validate callback which checks the cspell file to make sure that it changed. If it didn't change, then you can provide a reason and it goes back to the agent and it gets to try again.

l0lawrence (approved):

Issue #11864, opened 2025-08-27
[AVC/APIVIew] Bulk resolve similar AVC comments


tjprescott:
Scneario: copilot review generates 10 instances of a comment. The architect reviews the comment and deems it bad. They give it a thumbs-down and issue a correction. Rather than have to handle the other 9 instances, AVC should identify the 9 instances and automatically resolve those comments.

As part of this, we should consider have a "Resolved" status similar to how you close a Github issue (duplicate/won't fix/etc). The reason for this is primarily so that our quality metrics can reason about these duplicate issues and count them as 1 bad comment (or 10--up for debate!). 

This will require that the resolveComment POST operation be accessible by token auth.

tjprescott:
Chatting with @AlitzelMendez it would be helpful if Copilot, when it creates a review, determines which comments are effectively the same thing and assigns them a correlation or batch id. In that way, we can close an entire batch of comments easily in the UI without needing to complicate the logic.

AlitzelMendez:
# Design ideas 

## Side-by-Side Buttons
- Added a "Resolve All" button directly next to the existing "Resolve Conversation" button
- This button should only be present when there are related conversations
<img width="957" height="249" alt="Image" src="https://github.com/user-attachments/assets/50c1205b-e314-4765-a33e-c2e6f7ccd61b" />

## Dropdown Split Button
- Converted the resolve button into a split button with dropdown arrow
- Main button still does "Resolve Conversation", dropdown reveals "Resolve All Related Conversations"
- This dropdown should be only present when there are related conversations

<img width="1001" height="293" alt="Image" src="https://github.com/user-attachments/assets/e60568ba-ab02-475a-a538-505acfc5d942" />

## Modal Preview Dialog 
- Clicking "Resolve Conversation" opens a modal showing all related conversations
- Lists specific lines and comments that would be affected
- Two clear action buttons: "Resolve only current conversation" vs "Resolve all related conversations"
- Only gets triggered when there are related conversations 

<img width="1038" height="277" alt="Image" src="https://github.com/user-attachments/assets/eec8c37d-03f3-4f21-94c7-5ad6da5fc134" />

<img width="788" height="460" alt="Image" src="https://github.com/user-attachments/assets/77ae333c-f225-4faf-88b9-daf3a13d3403" />


@ccbarragan  I would love your feedback and/or ideas on this 😄

ccbarragan:
[...]

PR #12310, opened 2025-10-01
[API VIew] Bulk resolve similar AVC comments 

AlitzelMendez:
closes: https://github.com/Azure/azure-sdk-tools/issues/11864

Added related comments feature for bulk resolution of AI-generated feedback with shared context. Comments are grouped by `correlationId` and displayed in a GitHub-style dialog with code context.

## UI Changes
- Visual Indicator: Badge shows comment count with related issues
- Bulk Resolution Dialog: Select and resolve multiple related comments
- Code Context Display: GitHub-style code snippets for each comment

Correlation Id is already provided by copilot so now we integrated receiving that value 

## Demo 
https://github.com/user-attachments/assets/144cd779-c589-4ffe-8839-577d8543db36


src/dotnet/APIView/ClientSPA/src/app/_components/shared/related-comments-dialog/related-comments-dialog.component.ts
```diff
@@ -0,0 +1,118 @@
...
+  toggleCommentSelection(commentId: string) {
+    if (this.selectedCommentIds.has(commentId)) {
+      this.selectedCommentIds.delete(commentId);
+    } else {
+      this.selectedCommentIds.add(commentId);
+    }
+    this.updateSelectAllState();
+  }
+
+  onSelectAllChange(event: { checked?: boolean }) {
```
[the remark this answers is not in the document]
AlitzelMendez: after some investigation this doesn't seem necessary, it is just closer to prime ng but doesn't deliver any value

src/dotnet/APIView/ClientSPA/src/app/_helpers/comment-relation.helper.ts
```diff
@@ -0,0 +1,91 @@
...
+    
+    return relatedComments.filter(c => visibleNodeIds.has(c.elementId));
+  }
+
+  static getRelatedCommentsCount(comment: CommentItemModel, allComments: CommentItemModel[], codePanelRowData?: CodePanelRowData[]): number {
+    if (!comment.correlationId || !allComments) {
+      return 0;
+    }
+
+    const cacheKey = `${comment.id}_${codePanelRowData ? 'visible' : 'all'}`;
```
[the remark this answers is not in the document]
AlitzelMendez: addressed

src/dotnet/APIView/ClientSPA/src/app/_components/shared/comment-thread/comment-thread.component.ts
```diff
@@ -393,4 +410,52 @@ export class CommentThreadComponent {
...
+    commentIds.forEach(commentId => {
+      const comment = this.relatedComments.find(c => c.id === commentId);
+      if (comment) {
+        const commentCodeRow = this.allCodePanelRowData?.find(row => row.nodeId === comment.elementId);
+        
+        this.commentResolutionActionEmitter.emit({
+          commentThreadUpdateAction: CommentThreadUpdateAction.CommentResolved,
+          elementId: comment.elementId,
+          nodeIdHashed: commentCodeRow?.nodeIdHashed || this.codePanelRowData!.nodeIdHashed,
+          associatedRowPositionInGroup: commentCodeRow?.associatedRowPositionInGroup || this.codePanelRowData!.associatedRowPositionInGroup,
```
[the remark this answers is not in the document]
AlitzelMendez: resolved

src/dotnet/APIView/ClientSPA/src/app/_components/review-page/review-page.component.ts
```diff
@@ -361,10 +363,32 @@ export class ReviewPageComponent implements OnInit {
...
 
+  private processEmbeddedComments() {
+    if (!this.codePanelData || !this.comments) return;
+    
+    Object.values(this.codePanelData.nodeMetaData).forEach(nodeData => {
+      if (nodeData.commentThread) {
+        Object.values(nodeData.commentThread).forEach(commentThreadRow => {
+          if (commentThreadRow.comments) {
+            commentThreadRow.comments.forEach(embeddedComment => {
+              const globalComment = this.comments.find(c => c.id === embeddedComment.id);
```
tjprescott, line 380: What is a global comment or an embedded comment?

AlitzelMendez: Global comments are all the comments returned when the server is called - the complete collection of comments for the review.

Embedded comments are the comments related to the code display, stored in the commentThread as they are associated with specific code nodes.

This architecture allows us to display only the count/related comments for the visible comments.

### Current behavior:
- All comments are loaded and processed. If there isn't a matching code line, you can't see them in the code view, but they aren't filtered by anything. The problem with this approach was that "related" comments were being counted even when not visible.
- Embedded comments are filtered comments that are actually displayed and have a related line shown on the canvas.

### Why not filter them from the server call?
 Great question. I think we need to do this, but right now those comments are visible in the "Comments" tab, so filtering at the server level would break that workflow.

tjprescott: So if they are only visible in conversations but do match a correlation id, shouldn't they still be counted as related?

AlitzelMendez: I don't think they should, but what do you think on these?

My way to see this is that we will be asking you to resolve an issue from a line that is not even there anymore, this could create confusion of: is this line on this revision?

do you think we should show the issues even if they belong to a different revision?

tjprescott: Well, let's do this for now and I'll do some testing to see if there's an issue.

PR #12317, opened 2025-10-02
Add testcase filtering to evals runner

marcarro:
Added -c/--testcase flag to filter JSONL data by testcase field.

## Usage
`python run.py -t workflows/workflow.yaml -c "specific_testcase_name"`

## Changes

- Add --testcase/-c CLI argument
- Filter JSONL data before evaluation using temp files
- No filter = run all 

## Testing
```
# Run specific testcase
python run.py -t workflows/apiview.yaml -c "small_apiview_few_violations"

# Run all (unchanged behavior)
python run.py -t workflows/apiview.yaml
```

### DIFF
--- a/packages/python-packages/apiview-copilot/evals/_runner.py
+++ b/packages/python-packages/apiview-copilot/evals/_runner.py
@@ -142,6 +142,7 @@ class EvalRunner:
                         if data.get('testcase') == self.testcase:
                             filtered_lines.append(line)
                             found_testcase = True
+                            break
                     except json.JSONDecodeError as e:
                         raise ValueError(f"Invalid JSON on line {line_num} in {file_path}: {e}")
             
--- a/packages/python-packages/apiview-copilot/evals/_runner.py
+++ b/packages/python-packages/apiview-copilot/evals/_runner.py
@@ -1,4 +1,6 @@
 import os
+import tempfile
+import json
 import pathlib
 import sys
 from typing import List
@@ -18,9 +20,10 @@ DEFAULT_NUM_RUNS: int = 1
 class EvalRunner:
     """Class to run evals for APIView copilot."""
 
-    def __init__(self, *, language: str, test_path: str, num_runs: int = DEFAULT_NUM_RUNS):
+    def __init__(self, *, language: str, test_path: str, testcase: str, num_runs: int = DEFAULT_NUM_RUNS):
         self.language = language
         self.test_path = test_path
+        self.testcase = testcase
         self.num_runs = num_runs
         self.settings = SettingsManager()
 
@@ -45,6 +48,7 @@ class EvalRunner:
         evaluator = self._evaluator_class(self._workflow_config)
         guideline_ids = set()
         all_run_results = []
+        temp_files = []
 
         azure_ai_project = {
             "subscription_id": self.settings.get("EVALS_SUBSCRIPTION"),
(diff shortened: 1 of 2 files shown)


packages/python-packages/apiview-copilot/evals/_runner.py
```diff
@@ -18,9 +20,10 @@
 class EvalRunner:
     """Class to run evals for APIView copilot."""
 
-    def __init__(self, *, language: str, test_path: str, num_runs: int = DEFAULT_NUM_RUNS):
+    def __init__(self, *, language: str, test_path: str, testcase: str, num_runs: int = DEFAULT_NUM_RUNS):
```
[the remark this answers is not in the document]
tjprescott: Agree. If testcase isn't supplied, it should run all by default.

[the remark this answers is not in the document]
tjprescott: Also, it probably makes sense for this to be of type str or list[str].

packages/python-packages/apiview-copilot/evals/_runner.py
```diff
@@ -102,3 +120,51 @@ def run(self):
...
+                    line = line.strip()
+                    if not line:
+                        continue
+                        
+                    try:
+                        data = json.loads(line)
+                        if data.get('testcase') == self.testcase:
+                            filtered_lines.append(line)
+                            found_testcase = True
+                            break
```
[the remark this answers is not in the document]
tjprescott: Agree, we should support a list of testcases.

packages/python-packages/apiview-copilot/evals/run.py
```diff
@@ -27,10 +27,17 @@
         required=True,
         help="Path to workflow YAML.",
     )
+    parser.add_argument(
```
tjprescott, line 30: I don't recall the syntax for argparse but there should be something like `nargs='+'` which allows you to do something like `--testcase test1 test 2`. We should support that. Also, this should not be required.

tjprescott (changes requested):
A few comments. Also, please update the `avc eval run` command in `cli.py` to include this new parameter. Thanks @marcarro!

marcarro:
Resolved 👍 @tjprescott

packages/python-packages/apiview-copilot/cli.py
```diff
@@ -1108,6 +1108,9 @@ def load_arguments(self, command):
             ac.argument(
                 "num_runs", type=int, options_list=["--num-runs", "-n"], help="Number of times to run the test case."
             )
+            ac.argument(
+                "testcase", type=str, nargs='*', options_list=["--testcase", "-c"], help="List of test cases to run."
```
laiapat: Nit: `eval run` [accepts a `--test-case` option](https://github.com/Azure/azure-sdk-tools/blob/6ebaf43131766d96d84090391c0c6f49e345de36/packages/python-packages/apiview-copilot/cli.py#L1077), so maybe we should align here. Also, if we allow multiple cases, we should maybe make this `--test-cases` (that's a more minor consideration though)
```suggestion
                "test_case", type=str, nargs='*', options_list=["--test-case", "-c"], help="List of test cases to run."
```

packages/python-packages/apiview-copilot/evals/_runner.py
```diff
@@ -68,24 +72,38 @@ def run(self):
...
-                file_run_results.append({file.name: result})
-            all_run_results.extend(file_run_results)
+        try:
+            for file in self._test_files:
+                if not file.exists():
+                    raise ValueError(f"Test file not found: {file}")
+
+                filtered_file_path = self._filter_jsonl_data(str(file))
+                if filtered_file_path != str(file):
+                    temp_files.append(filtered_file_path)
```
laiapat, line 83: I think a comment explaining the logic here could be helpful (I had to dig into `_filter_jsonl_data` a bit to understand the flow/intent)

packages/python-packages/apiview-copilot/cli.py
marcarro: By aligning, do you mean we should only keep one of the two (either the one under `eval` arguments context, or the one under `eval run`), or are you referring to the argument naming? In case of the latter, I changed it to be consistent

laiapat: I meant the latter -- update looks good, thanks!

tjprescott:
Closing this in favor of https://github.com/Azure/azure-sdk-tools/pull/12439

PR #12346, opened 2025-10-03
[azsdkcli] Test runner service + generic language specific resolver

timovv:
- Add SdkLanguage enum to represent languages instead of using strings
- Create generic LanguageSpecificResolver to resolve language-specific services
  - Reuses logic from and replaces LanguageSpecificCheckResolver to reduce duplication. Also replaces ClientUpdateLanguageServiceResolver.
  - To resolve a language-specific service, call Resolve<T> on LanguageSpecificResolver, e.g. `languageSpecificResolver.Resolve<ILanguageSpecificChecks(packagePath)`.
- Add ability to specify environment variables to ProcessHelper. Needed for TEST_MODE environment variable used by JS test framework to specify live vs. recorded tests.
- Implement new RunTestsTool
  - Create language-specific ITestRunner service which also uses LanguageSpecificResolver
  - Other languages should implement ITestRunner for running tests (issues to follow)


tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/TypeSpec/TspClientUpdateToolTests.cs
```diff
@@ -113,11 +114,11 @@ public Task<List<PatchProposal>> ProposeFixesAsync(ClientUpdateSessionState sess
         }
     }
 
-    private class SingleResolver : IClientUpdateLanguageServiceResolver
+    private class SingleResolver : ILanguageSpecificResolver
     {
         private readonly IClientUpdateLanguageService _svc;
         public SingleResolver(IClientUpdateLanguageService svc) { _svc = svc; }
-        public Task<IClientUpdateLanguageService?> ResolveAsync(string? packagePath, CancellationToken ct = default) => Task.FromResult<IClientUpdateLanguageService?>(_svc);
+        public Task<T?> Resolve<T>(string packagePath, CancellationToken ct = default) where T : ILanguageSpecificService => Task.FromResult((T)_svc);
```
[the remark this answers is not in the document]
timovv: This does not work

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/LanguageSpecificResolver.cs
```diff
@@ -0,0 +1,96 @@
...
+    public async Task<TService?> Resolve<TService>(string packagePath, CancellationToken ct = default) where TService : ILanguageSpecificService
+    {
+        var services = _serviceProvider.GetServices<TService>();
+        var language = await DetectLanguageAsync(packagePath, ct);
+        if (language == null)
+        {
+            throw new InvalidOperationException($"Could not detect language for package at path: {packagePath}");
+        }
+
+        return services.FirstOrDefault(s => s.SupportedLanguage == language);
```
benbp: I think this lookup could be done using keyed services (https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services) where the enum is the key, that way we don't have to iterate through all the services.

```csharp
services.AddKeyedSingleton<ILanguageSpecificChecks, JavaScriptLanguageSpecificChecks>(SdkLanguage.JavaScript);
```

That said, I'm wondering if there's a way to avoid having to inject the service provider, possibly by using factory methods in ServiceRegistrations.cs or something along those lines.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/RunTestsTool.cs
```diff
@@ -0,0 +1,77 @@
...
+    [Description("Run tests for SDK packages")]
+    [McpServerToolType]
+    public class RunTestsTool(
+        ILogger<RunTestsTool> logger,
+        ILanguageSpecificResolver languageServiceResolver
+    ) : MCPTool
+    {
+        public override CommandGroup[] CommandHierarchy { get; set; } = [SharedCommandGroups.Package];
+
+        private const string RunTestsCommandName = "run-tests";
```
benbp: Simpler?

```suggestion
        private const string RunTestsCommandName = "test";
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/RunTestsTool.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        protected override Command GetCommand()
+        {
+            var command = new Command(RunTestsCommandName, "Run tests for SDK packages");
+            // Add the package path option to the parent command so it can be used without subcommands
+            command.AddOption(SharedOptions.PackagePath);
+            command.SetHandler(ctx => HandleCommand(ctx, ctx.GetCancellationToken()));
+
+            return command;
+        }
```
benbp: This area has been simplified a bit under the hood, and the `SetHandler` call happens automatically now on CLI setup.

```suggestion
        protected override Command GetCommand() => [
            new(RunTestsCommandName, "Run tests for SDK packages") {
                SharedOptions.PackagePath
            }
        ];
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/RunTestsTool.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                return new DefaultCommandResponse
+                {
+                    Result = $"Tests for package at '{packagePath}' completed successfully."
+                };
+            }
+            catch (Exception ex)
+            {
+                logger.LogError(ex, "Unhandled exception while running package tests");
+                throw;
```
benbp: This should also return `DefaultCommandResponse` instead of a re-throw.

benbp:
It's going to be so nice to have an easy "test this package" entrypoint. I often have to debug pipelines and figure out if it's the engineering system or the tests, but I never have any of the test harness infrastructure set up (or know how to do so).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/RunTestsTool.cs
timovv: My only thought here is that we may add more test-related commands in the future, at which point calling this command `test` would become ambiguous. Maybe e.g. test resource deployment, test recording? not really sure, and some of these could probably be bundled into the test command. But I suppose we can cross that bridge when we get there and rename this command if it's confusing.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/LanguageSpecificResolver.cs
timovv: Keyed services looks promising, I think we can use that for sure. It also feels better to me having the language association at the service provider level just in case e.g. two languages want to share the same implementation for some reason (but unsure when that would happen?).

We probably could avoid pulling in the service provider. One idea I had for that was putting the type parameter at the class level, and then just requesting the IEnumerable in the constructor in the normal. I don't think this would be possible using the keyed services approach though, since we don't know at construction time which language's service we need. Along similar lines, IIUC we also can't resolve the per-language services at the service registration stage, since the package directory we are resolving the service for may change over the lifetime of the process (e.g. if using the MCP server across multiple repositories, we would need to pull in different language services depending on which directory the current tool call is requesting). So some level of indirection is required.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/RunTestsTool.cs
timovv: Renamed it for now, less typing is better. I kept the same name for the MCP tool since I feel like Copilot might appreciate the verbosity (but can change that if we want the consistency)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/LanguageSpecificResolver.cs
timovv: Implemented keyed services, if you have any thoughts on how we could avoid the IServiceProvider let me know. I can't think of anything else clean but am admittedly not a dependency injection wizard :)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Tests/ITestRunner.cs
```diff
@@ -0,0 +1,6 @@
+namespace Azure.Sdk.Tools.Cli.Services.Tests;
+
+public interface ITestRunner
+{
+    public Task RunAllTests(string packagePath, TestMode mode, CancellationToken ct = default);
```
timovv: We don't strictly need the test mode right now, but I anticipate we'll need it soon, so I put it in. If there objections I can take it out.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/LanguageSpecificResolver.cs
benbp: Yeah I'm no wizard either, I've been trying to wrap my head around what's possible and whether we can simplify.

If we inject language services via `AddTransient` (or the change I made to use `AddScoped` may be sufficient) and register a factory method with the service provider then it would be called every time and can check package directory to resolve the right type. That said it also feels a bit weird to put a bunch of I/O into the DI layer (failure stack traces would be monstrous), so I'm not sure whether it's any better than injecting the service provider to the resolver.

Injecting the provider is arguably an anti-pattern but as long as we keep it from being passed around anywhere else I don't think it's so bad.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/RunTestsTool.cs
benbp: @timovv yeah that's a good point. I could see those possibly as flags `azsdk package test --live` `azsdk package test --live --record` as well. I agree verbosity better for mcp tool names and less typing for CLI. I still need to add an auto-completion generator/installer for the CLI as well.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Tests/ITestRunner.cs
benbp: Thoughts about adding `RunTests` and supporting language specific test name filters?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
```diff
@@ -30,20 +32,22 @@ public static void RegisterCommonServices(IServiceCollection services, OutputHel
...
+            services.AddKeyedSingleton<ILanguageSpecificChecks, GoLanguageSpecificChecks>(SdkLanguage.Go);
 
             // Client update language services
-            services.AddSingleton<IClientUpdateLanguageService, JavaUpdateLanguageService>();
-            services.AddSingleton<IClientUpdateLanguageServiceResolver, ClientUpdateLanguageServiceResolver>();
+            services.AddKeyedSingleton<IClientUpdateLanguageService, JavaUpdateLanguageService>(SdkLanguage.Java);
             // Future: services.AddSingleton<IClientUpdateLanguageService, PythonClientUpdateLanguageService>(); etc.
 
+            services.AddKeyedSingleton<ITestRunner, JavaScriptTestRunner>(SdkLanguage.JavaScript);
+
```
benbp: These will need to be `AddKeyedScoped` now due to https://github.com/Azure/azure-sdk-tools/pull/12395, fixing the issue with transitive dependencies on the scoped `TokenUsageHelper`.

```suggestion
            services.AddKeyedScoped<ILanguageSpecificChecks, PythonLanguageSpecificChecks>(SdkLanguage.Python);
            services.AddKeyedScoped<ILanguageSpecificChecks, JavaLanguageSpecificChecks>(SdkLanguage.Java);
            services.AddKeyedScoped<ILanguageSpecificChecks, JavaScriptLanguageSpecificChecks>(SdkLanguage.JavaScript);
            services.AddKeyedScoped<ILanguageSpecificChecks, DotNetLanguageSpecificChecks>(SdkLanguage.DotNet);
            services.AddKeyedScoped<ILanguageSpecificChecks, GoLanguageSpecificChecks>(SdkLanguage.Go);

// Client update language services
            services.AddKeyedScoped<IClientUpdateLanguageService, JavaUpdateLanguageService>(SdkLanguage.Java);
// Future: services.AddSingleton<IClientUpdateLanguageService, PythonClientUpdateLanguageService>(); etc.

            services.AddKeyedScoped<ITestRunner, JavaScriptTestRunner>(SdkLanguage.JavaScript);

```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Tests/ITestRunner.cs
timovv: Yes, this would be useful. But I'd like to avoid the conversation of how the filters look and how they are implemented for now, since I think that could get a little hairy and get in the way of the v1 prototype goal. Definitely something to come back to though.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/LanguageSpecificResolver.cs
timovv: Since the package path is a CLI option I'm not sure if we'll actually know it at DI time anyway?

Tbh, I don't see a way to avoid injecting the service provider. Looking into the reasoning behind it being an antipattern, the key point seems to be that it makes testing more difficult. But by isolating the service provider to this one class I think we're all clear on that front. One thing we could do is move the generic paramter on `Resolve` to the class level (and require one resolver per type). We'd still need the service provider, but it should help mocking since the type of the Resolve method will be specific.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/TestTool.cs
```diff
@@ -0,0 +1,84 @@
...
+        public override async Task<CommandResponse> HandleCommand(InvocationContext ctx, CancellationToken ct)
+        {
+            var packagePath = ctx.ParseResult.GetValueForOption(SharedOptions.PackagePath);
+            var testMode = ctx.ParseResult.GetValueForOption(TestModeOption);
+
+            return await RunPackageTests(packagePath, testMode, ct);
+        }
+
+        [McpServerTool(Name = "azsdk_package_run_tests"), Description("Run tests for SDK packages. Provide package path.")]
+        public async Task<DefaultCommandResponse> RunPackageTests(string packagePath, TestMode testMode = TestMode.Playback, CancellationToken ct = default)
```
l0lawrence: for right now are we supporting all 3 modes or just playback?

timovv: I think if it's not too hard for your language it's a good idea since we will definitely need it soon. For JS it was super easy so I just did it. But if it's involved then feel free to ignore the option for now, it's not in scope.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/TestTool.cs
```diff
@@ -0,0 +1,84 @@
...
+using Azure.Sdk.Tools.Cli.Models;
+using Azure.Sdk.Tools.Cli.Services;
+using Azure.Sdk.Tools.Cli.Services.Tests;
+
+namespace Azure.Sdk.Tools.Cli.Tools.Package
+{
+    /// <summary>
+    /// This tool runs tests for the specified SDK package.
+    /// </summary>
+    [Description("Run tests for SDK packages")]
```
samvaity: Should the description say "runs tests for the specified SDK package."?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/TestTool.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        public override async Task<CommandResponse> HandleCommand(InvocationContext ctx, CancellationToken ct)
+        {
+            var packagePath = ctx.ParseResult.GetValueForOption(SharedOptions.PackagePath);
+            var testMode = ctx.ParseResult.GetValueForOption(TestModeOption);
+
+            return await RunPackageTests(packagePath, testMode, ct);
+        }
+
+        [McpServerTool(Name = "azsdk_package_run_tests"), Description("Run tests for SDK packages. Provide package path.")]
```
samvaity: "runs tests for the specified SDK package."

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/TestTool.cs
samvaity: For TestMode Live how are we accessing the environment vars needed for the test to run are they passed by a configuration file?

samvaity: Another idea, what if we read the test mode from the config too and have playback as default. That way user can just prompt to run tests without having to know the overhead of RECORD or other modes.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
samvaity: @timovv If you want feel free to move language specific resolver to a separate PR, if that helps move faster.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Tests/ITestRunner.cs
benbp: Agreed it would be tricky and worth punting. Off the top of my head I would say take a list of strings or regexes and match them to any strings within the group of test titles / test classes / test directories / etc. Then under the hood we could translate to the more annoying details like c# test filters requiring the full class/namespace. Even typing it out I'm realizing how annoying it's going to be to figure out, but on the other hand it's annoying to figure out the filters already so would be nice to just capture the logic at a lower level.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/TestTool.cs
timovv: for JS these are provided through a .env file which is pre-existing. They could also just already be present in the environment. But how other languages do it I don't know, and how we do it in a cross-language fashion is an interesting question. In the interest of shipping something, I think I will just remove the test mode option for now and we can come back to it later.

timovv: Tbf, the user is going to have to have some idea about the different tests modes since at some point (maybe not in the v1 scenario) they're going to have to record tests.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
timovv: I need the resolver for this PR so it is staying in.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Tests/ITestRunner.cs
timovv: Taking it out

timovv: > more annoying details like c# test filters requiring the full class/namespace

yup, and x5 that since each language is going to be different. Let's punt and revisit at a later time.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/LanguageSpecificResolver.cs
timovv: All right. I have made a few changes here which I think are positive ones:

- Moved the generic parameter up to the class level on the resolver, so each resolver is specific to a service.
- This means that a separate resolver needs to be registered for each language-specific service. To help make sure that happens, I added an `AddPerLanguage` extension method to the service collection so that it's simple to add implementations that vary across languages.
- Since the generic is at the class level now, we know the interface type at the resolver's construction time. That means that we can just have a parameter for each language's implementation instead of pulling in the `IServiceProvider`. The downside of this is that we need to remember to add a parameter there if/when we add more languages, but I'm sure that this isn't too heavy a burden.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/TestTool.cs
timovv: Removing it

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
timovv: This is now in the `AddPerLanguage` extension method which uses `AddKeyedScoped`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/LanguageSpecificResolver.cs
```diff
@@ -0,0 +1,115 @@
...
+    // If adding languages in future, add a corresponding entry here.
+) : ILanguageSpecificResolver<T> where T : class
+{
+    public async Task<T?> Resolve(string packagePath, CancellationToken ct = default)
+    {
+        var language = await DetectLanguageAsync(packagePath, ct);
+
+        return language switch
+        {
+            SdkLanguage.DotNet => dotnetService,
```
deyaaeldeen, line 29: is it possible to pass the packagePath down to the language implementation here? I think it is needed in most cases.

timovv: The current approach is to just accept the package path as a method parameter in the implementation where necessary, which feels fine to me. I'm struggling to come up with a better alternative. Here are a couple of things we could do, but honestly I don't like either:
- Add package path as a constructor parameter to the per language implementations and construct an instance when `Resolve` gets called with the package path. But now we have to manage that new instance's lifetime ourselves and we would have to use reflection in the form of [ActivatorUtilties](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities?view=net-9.0-pp) to actually create the object.
- Add another level of indirection so that the `LanguageSpecificResolver` takes in a factory interface instead of the language specific implementation. That factory would be responsible for constructing the language specific implementation based on the package path. But then we would have two levels of factory...

Ultimately the issue boils down to that we don't know the package path at service resolution/construction time, since it's a CLI argument that only gets parsed after all the services have been resolved. I don't know if there is another clean solution here that still allows the language implementations to get access to services through DI (which I think is important), but if there is something I'm not thinking of I'd be glad to entertain it.

timovv: Resolving so we can get this merged, but happy to address concerns in a follow up PR if necessary.

Issue #12043, opened 2025-09-11
[APIView] APIView cross language changes to support metadata file


praveenkuttappan:
We need changes in APIView and pipeline to upload metadata file along with original source file to support cross language Id in APIView.

Changes required:
-  APIView side changs to support cross language metadata
     - [ ]  Update the Codefile schema to embed the cross language metadata.
     - [ ] In upgrade scripts, ensure that for a revision being upgraded, any cross-language metadata is extracted first and sent into the language-specific parser so that it is preserved in the regenerated token file.

-  Sandboxing(Update) script:
     - [ ] In upgrade script, ensure that for a revision being upgraded, any cross-language metadata is extracted first and passed into the Python APIView parser.

---

Old schema:
```tsp
model CodeFile {
  ...
  LanguageVariant?: "None" | "Spring" | "Android" = "None";
  /** Id to identify Reviews for different languages generated from the same TypeSpec specification. */
  CrossLanguagePackageId?: string;
  ReviewLines: Array<ReviewLine>;
  ...
}
```

New schema:
```tsp
model CodeFile {
  ...
  LanguageVariant?: "None" | "Spring" | "Android" = "None";
  /** Id to identify Reviews for different languages generated from the same TypeSpec specification. */
  CrossLanguageMetadata?: CrossLangaugeMetadata;
  ReviewLines: Array<ReviewLine>;
  ...
}

alias CrossLanguageId = string;
alias LanguageSpecificId = string;

model CrossLanguageMetadata {
  CrossLanguagePackageId: string;
  CrossLanguageDefinitionId: Map<LanguageSpecificId, CrossLanguageId>;
}

```

swathipil:
An example of the current metadata file:
```json
{
    "CrossLanguagePackageId": "AzureHealthInsights",
    "CrossLanguageDefinitionId": {
        "azure.healthinsights.radiologyinsights.models.RadiologyInsightsInference": "AzureHealthInsights.RadiologyInsightsInference",
    }
}
```

Ideally we can update the CrossLanguageMetadata.CrossLanguageIds to CrossLanguageDefinitionId, so that no codegen changes have to be made.

PR #12329, opened 2025-10-02
[Python APIView] add cross language metadata to token file

swathipil:
related to: #12043

Adds CrossLanguageMetadata to token file as per issue.
Token file looks like:
```json
    "ReviewLines": [...],
    "CrossLanguageMetadata": {
        "CrossLanguagePackageId": "SchemaRegistry",
        "CrossLanguageDefinitionId": {
            "azure.schemaregistry.models.SchemaContentTypeValues": "SchemaRegistry.SchemaContentTypeValues"
        }
    },
    "Diagnostics": []
```

PR #12348, opened 2025-10-05
[APIView] Cross language changes to support metadata information

AlitzelMendez:
closes: https://github.com/Azure/azure-sdk-tools/issues/12043

These changes provide support for cross language metadata, which will help with being able to preserve cross language information / relations when there is an update to the parser, as this information is provided just when the revision is generated

part of the process of: https://github.com/Azure/azure-sdk-tools/pull/12329

### DIFF
--- a/src/dotnet/APIView/APIViewUnitTests/AutoReviewControllerTests.cs
+++ b/src/dotnet/APIView/APIViewUnitTests/AutoReviewControllerTests.cs
@@ -6,7 +6,6 @@ using System.Threading.Tasks;
 using ApiView;
 using APIViewWeb;
 using APIViewWeb.Controllers;
-using APIViewWeb.Helpers;
 using APIViewWeb.LeanModels;
 using APIViewWeb.Managers;
 using APIViewWeb.Managers.Interfaces;
@@ -402,7 +401,7 @@ namespace APIViewUnitTests
             public override string[] Extensions => new[] { ".json" };
             public override string VersionString => "1.0";
             public override bool CanUpdate(string versionString) => false;
-            public override Task<CodeFile> GetCodeFileAsync(string originalName, Stream stream, bool runAnalysis) => Task.FromResult<CodeFile>(null);
+            public override Task<CodeFile> GetCodeFileAsync(string originalName, Stream stream, bool runAnalysis, string crossLanguageMetadata = null) => Task.FromResult<CodeFile>(null);
             public override bool UsesTreeStyleParser => _usesTreeStyleParser;
             public override CodeFile GetReviewGenPendingCodeFile(string fileName) => null;
             public override bool GeneratePipelineRunParams(APIRevisionGenerationPipelineParamModel param) => false;


src/dotnet/APIView/APIViewUnitTests/CrossLanguageMetadataTests.cs
```diff
@@ -0,0 +1,118 @@
+using System.Text.Json;
+using APIView.Model;
+using FluentAssertions;
+using Xunit;
+
+namespace APIViewUnitTests;
+
+public class CrossLanguageMetadataTests
```
swathipil, line 9: would be good to add one more test that doesn't include CrossLanguageMetadata at all in the token file, since it's optional

tjprescott (changes requested):
There is TypeSpec that describes the CodeFile schema and related schemas. Please update that accordingly with these changes.

src/dotnet/APIView/APIViewUnitTests/CrossLanguageMetadataTests.cs
AlitzelMendez: added :)

AlitzelMendez:
> There is TypeSpec that describes the CodeFile schema and related schemas. Please update that accordingly with these changes.

done! :)

tools/apiview/parsers/apiview-treestyle-parser-schema/main.tsp
```diff
@@ -97,6 +97,11 @@ model CodeDiagnostic {
   HelpLinkUri?: string;
 }
 
+model CrossLanguageMetadata {
+  CrossLanguagePackageId: string;
+  CrossLanguageDefinitionId: Dictionary<string, string>;
```
tjprescott: Please use aliases to communicate what is the key and what is the value.

```tsp
alias CrossLanguageId = string;
alias LanguageSpecificId = string;

CrossLanguageDefinitionId: Dictionary<LanguageSpecificId, CrossLanguageId>;
```

tools/apiview/parsers/apiview-treestyle-parser-schema/main.tsp
```diff
@@ -134,3 +139,13 @@ model NavigationItem {
...
 }
+
+
+@doc("The generic name value pair.")
+model Dictionary<TName, TValue> {
+  @doc("The name.")
+  name?: TName;
+
+  @doc("The value.")
+  value?: TValue;
```
tjprescott: This can be removed. Just replace Dictionary with whatever the native TSP primitive is (assuming there is one?)

AlitzelMendez: not really, https://typespec.io/docs/language-basics/built-in-types/ 

the closest is `Record<string>` that doesn't really represent this, but could be close enough

AlitzelMendez: or `Record<CrossLanguageId>`

AlitzelMendez: I updated to `Record<CrossLanguageId>` but I don't think is fully clear of what we are representing here

tjprescott: Wow, surprised that TypeSpec doesn't have such a common primitive.

Okay, then maybe instead of Aliases we could get away with something like:

```tsp
model CrossLanguageDefinitionMetadata<TLanguageSpecificId, TCrossLanguageId> {
  key: TLanguageSpecificId;
  value: TCrossLanguageIs;
}
```

tools/apiview/parsers/apiview-treestyle-parser-schema/CodeFile.json
```diff
@@ -211,6 +226,21 @@
...
+                },
+                "value": {
+                    "type": "string"
+                }
+            },
+            "required": [
+                "key",
+                "value"
+            ]
+        },
```
tjprescott: Hmm...this isn't accurate. This would imply an object like `{key: "foo.bar", value: "Foo.bar"}` but the actual structure is `{"foo.bar": "Foo.bar"}`.

tjprescott (approved):

src/dotnet/APIView/APIView/Model/CodeFile.cs
```diff
@@ -41,7 +42,9 @@ public string VersionString
         public string ServiceName { get; set; }
         public string PackageDisplayName { get; set; }
         public string PackageVersion { get; set; }
+        // Deprecated: Use CrossLanguageMetadata.CrossLanguagePackageId instead. Kept for backward compatibility with TypeSpec parsers.
```
maririos, line 45: I think the only parser that has this implemented is Python. If that is true, once Python modifies their parser, then you should be able to remove this

maririos (approved):

src/dotnet/APIView/APIView/Model/CodeFile.cs
AlitzelMendez: yes! it is only Python and TypeSpec

tjprescott: Should we have a migration pattern for reviews pulled out of Cosmos, or just do a one-time migration?

To handle the parsers, would could have logic that simply allows these fields to be extracted from the old schema and pushed into the new schema. That would give us backward compatibility without having to keep this old param around. Thoughts?

tools/apiview/parsers/apiview-treestyle-parser-schema/main.tsp
```diff
@@ -18,8 +18,8 @@ model CodeFile {
   Language: "C"|"C++"|"C#"|"Go"|"Java"|"JavaScript"|"Kotlin"|"Python"|"Swagger"|"Swift"|"TypeSpec";
   /** Language variant is applicable only for java variants*/
   LanguageVariant?: "None" | "Spring" | "Android" = "None";
-  /** Id to identify Reviews for different languages generated from the same TypeSpec specification. */
```
swathipil, line 21: I've actually also updated this file and the CodeFile in my PR: https://github.com/Azure/azure-sdk-tools/pull/12329/files#diff-64ca41dfaef5a9381f9b9c43d455395ccb7b6fd33db84cee3d11d01b2350626e

Do we want to keep changes in your PR and remove mine or do we want to remove yours? Either way works for me, but if we want to keep yours,could we pls copy over the ci.yml Node version bump too? I was seeing dependency conflicts which required the version bump. https://github.com/Azure/azure-sdk-tools/pull/12329/files#diff-0ce325df9a46db00716122c3fb2a2b203bb8b73e7a7aa7a7816366b77250980e

swathipil: OOB: we want to keep the changes here so I'll copy them over the the py parser PR.

src/dotnet/APIView/APIView/Model/CodeFile.cs
AlitzelMendez: I think that is a great idea! :) 
let me create an issue https://github.com/Azure/azure-sdk-tools/issues/12613

PR #12367, opened 2025-10-07
fix handle command bug

l0lawrence:


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/PackageCheckTool.cs
```diff
@@ -42,6 +42,8 @@ protected override List<Command> GetCommands()
                 subCommand.AddOption(SharedOptions.PackagePath);
                 subCommand.AddOption(fixOption);
 
+                subCommand.SetHandler(async ctx => await InstrumentedCommandHandler(subCommand, ctx));
```
benbp: This shouldn't be called at this level (I should change it somehow so it's not visible). The `SetHandler` gets automatically called now via https://github.com/Azure/azure-sdk-tools/blob/6f4dc9c88a1107d31c798ac7ded672ccb7ae05fd/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/MCPMultiCommandTool.cs#L16, so the subCommands need to be returned in the array result along with the parent command (which was changed in https://github.com/Azure/azure-sdk-tools/pull/12243/files#diff-478946e48b6ea49241057d6de1200a76810690d217894cfc49480e90f54aa577L50).

PR #12376, opened 2025-10-07
[AVC] Refactor eval structure

tjprescott:
Refactor's evals so that tests are simply individual JSON files.


packages/python-packages/apiview-copilot/evals/_config_loader.py
```diff
@@ -66,25 +73,40 @@ def get_evaluator_class(kind: str) -> Type:
...
+
+    # look for "test-config.yaml" or "test-config.yml" inside the directory
+    candidates = [base_dir / "test-config.yaml", base_dir / "test-config.yml"]
+    yaml_path = None
+    for c in candidates:
+        if c.exists() and c.is_file():
+            yaml_path = c.resolve()
+            break
+    if not yaml_path:
+        _fail(f"Workflow directory provided but no 'test-config.yaml' found in: {base_dir}")
```
tjprescott, line 109: Simply looks for `test-config.yaml` (or yml) in a folder. Any test cases in that folder are associated with that config (hence why the `tests` key is removed)

packages/python-packages/apiview-copilot/evals/tests/mention_action/update_kb_with_guideline.json
```diff
@@ -0,0 +1,33 @@
+{
```
tjprescott, line 1: The main issue. Having an expanded JSON file per test case is easier to maintain and we can "zip" these up into a JSONL file for execution.

packages/python-packages/apiview-copilot/evals/tests/mention_action/test-config.yaml
```diff
@@ -1,7 +1,6 @@
 name: mention-action
 kind: prompt
-tests: ../tests/python/mention-action.jsonl
-prompty: ../../prompts/mention/parse_conversation_action.prompty
+prompty: ../../../prompts/mention/parse_conversation_action.prompty
```
tjprescott, line 3: Removed the `tests` key because we infer that all tests in the folder where `test-config.yaml` lives are tests.

packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -443,24 +459,22 @@ def show_results(self, processed_results: dict) -> None:
 
             self._output_table(baseline_results, test_results, name)
 
-    def _establish_baseline(self, processed_results: dict, language: str, tests_directory: str) -> None:
+    def _establish_baseline(
```
kristapratico, line 462: I don't know if we need this anymore. Didn't we discuss setting baseline up in the foundry portal instead of locally?

marcarro: Yes, If i remember correctly this is also true of `_calculate_coverage`

tjprescott: I left the code but commented it out for now. Can remove later.

PR #12397, opened 2025-10-07
Fix issue in create release plan tool

praveenkuttappan:
Create release plan tool fails when we try to set the state field at the time of creating a work item. State can only be updated after it's created.

Similarly, we should not throw an exception when an existing release plan is found for the pull request.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
@@ -171,7 +171,7 @@ namespace Azure.Sdk.Tools.Cli.Tools.ReleasePlan
 
         }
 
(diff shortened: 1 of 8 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -297,8 +301,9 @@ public async Task<ObjectCommandResponse> CreateReleasePlan(string typeSpecProjec
 
                     return new ObjectCommandResponse
                     {
-                        Message = "Release plan created:",
-                        Result = releasePlan
+                        Message = "Release plan is being created",
+                        Result = releasePlan,
+                        NextSteps = [$"Get release plan from `workItem`, work item value:{releasePlan.WorkItemId}."]
```
benbp:
```suggestion
                        NextSteps = [$"Get release plan from `workItem`, work item value: {releasePlan.WorkItemId}"]
```

benbp (approved):

praveenkuttappan:
> Can we add unit tests for this?

I will check that in a separate Pr once unblocked.

praveenkuttappan:

PR #12398, opened 2025-10-07
update controller

helen229:


src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
```diff
@@ -187,6 +188,16 @@ public async Task<ActionResult> CreateApiReview(
...
             if (review != null)
             {
+                // Update package type if provided from controller parameter
+                if (!string.IsNullOrEmpty(packageType) && Enum.TryParse<PackageType>(packageType, true, out var parsedPackageType))
+                {
+                    if (review.PackageType != parsedPackageType)
+                    {
+                        review.PackageType = parsedPackageType;
+                        review = await _reviewManager.UpdateReviewAsync(review);
+                    }
```
tjprescott: Hm... not sure this is the right logic. This says "if package type is supplied and different from what's saved, update it."

I think the actual logic should be closer to:
- If the current review has no packageType and packageType is supplied, update it
- If review has a packageType that matches the supplied packageType, do nothing
- If review.packageType is not null and a packageType is supplied, and they DO NOT agree, that would be an error. A management plane cannot suddenly switch to a data-plane, or vice versa.

src/dotnet/APIView/APIViewWeb/Helpers/CommonUtilities.cs
```diff
@@ -69,6 +70,29 @@ public static bool IsSDKLanguageOrTypeSpec(string language)
...
+
+        /// <summary>
+        /// Management plane package (resource management libraries)
+        /// </summary>
+        Management,
+
+        /// <summary>
+        /// Cannot determine plane type from package name
+        /// </summary>
+        Unknown
```
tjprescott: Why this? I thought we default to null if the type can't be parsed?

src/dotnet/APIView/APIViewWeb/LeanModels/ReviewListModels.cs
```diff
@@ -113,6 +113,7 @@ public class ReviewListItemModel : BaseListitemModel
         public List<ReviewAssignmentModel> AssignedReviewers { get; set; } = new List<ReviewAssignmentModel>();
         public bool IsClosed { get; set; }
         public bool IsApproved { get; set; } // TODO: Deprecate in the future - redundant with NamespaceReviewStatus
+        public PackageType? PackageType { get; set; } // Nullable - null means not yet classified
```
tjprescott, line 116: Then why do we have the `Unknown` value?

src/dotnet/APIView/APIViewWeb/Managers/PullRequestManager.cs
```diff
@@ -248,13 +248,20 @@ private async Task ClosePullRequestAPIRevision(PullRequestModel pullRequestModel
...
-            PullRequestModel pullRequestModel, CodeFile baselineCodeFile, MemoryStream baseLineStream, string baselineFileName, CreateAPIRevisionAPIResponse responseContent)
+            PullRequestModel pullRequestModel, CodeFile baselineCodeFile, MemoryStream baseLineStream, string baselineFileName, CreateAPIRevisionAPIResponse responseContent, string packageType = null)
         {
             // fetch review for the package or create brand new review
             var review = await _reviewManager.GetReviewAsync(language: codeFile.Language, packageName: codeFile.PackageName);
             if (review == null)
             {
-                review = await _reviewManager.CreateReviewAsync(language: codeFile.Language, packageName: codeFile.PackageName, isClosed: false);
+                // Parse package type if provided, otherwise pass null for automatic classification
+                APIViewWeb.Helpers.PackageType? parsedPackageType = null;
```
tjprescott: Should this be a helper, or just a model? (do whatever is consistent with existing code-I have no preference)

src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
```diff
@@ -268,6 +270,22 @@ public async Task<ReviewListItemModel> CreateReviewAsync(string packageName, str
             return review;
         }
 
+        /// <summary>
+        /// Update an existing review
+        /// </summary>
+        /// <param name="review">The review to update</param>
+        /// <returns></returns>
+        public async Task<ReviewListItemModel> UpdateReviewAsync(ReviewListItemModel review)
```
tjprescott, line 278: Was nothing about reviews editable before?  If so, perhaps we validate that only the PackageType can be updated. Thoughts?

src/dotnet/APIView/ClientSPA/src/app/_models/review.ts
```diff
@@ -6,6 +6,12 @@ export enum FirstReleaseApproval {
   All
 }
 
+export enum PackageType {
+  Data = 'Data',
+  Management = 'Management',
+  Unknown = 'Unknown'
```
tjprescott: If `null` is unknown, this isn't needed.

src/dotnet/APIView/ClientSPA/src/app/_components/review-page/review-page.component.ts
```diff
@@ -277,13 +277,21 @@ export class ReviewPageComponent implements OnInit {
...
+        error: (error) => {
+          this.loadFailed = true;
+          this.loadFailedMessage = "Failed to load review. Please refresh the page or try again later.";
         }
       });
   }
 
+  updateReview(review: Review) {
+    this.review = review;
+  }
```
tjprescott: Why this change? Seems like it was simpler to just keep the assignments in-place?

src/dotnet/APIView/APIViewWeb/Managers/PullRequestManager.cs
helen229: move to package model

src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
helen229: {
                    if (!review.PackageType.HasValue)
                    {
                        // If the current review has no packageType and packageType is provided, update it
                        review.PackageType = parsedPackageType;
                        review = await _reviewManager.UpdateReviewAsync(review);
                    }
                    
                    this is only place we are calling this function, I think it should be fine as we didn't update other fields

tjprescott (commented):
Overall LGTM, but would like sign-off from @praveenkuttappan since he has more knowledge of the C# side of APIView than me.

src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
```diff
@@ -177,7 +178,7 @@ public async Task<ActionResult> CreateApiReview(
             return StatusCode(statusCode: StatusCodes.Status500InternalServerError);
         }
 
-        private async Task<(ReviewListItemModel review, APIRevisionListItemModel apiRevision)> CreateAutomaticRevisionAsync(CodeFile codeFile, string label, string originalName, MemoryStream memoryStream, bool compareAllRevisions = false)
+        private async Task<(ReviewListItemModel review, APIRevisionListItemModel apiRevision)> CreateAutomaticRevisionAsync(CodeFile codeFile, string label, string originalName, MemoryStream memoryStream, bool compareAllRevisions = false, string packageType = null)
```
praveenkuttappan: nit: Do you need a default for this param? It's a private method and you are passing the value from both calling places.

src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
```diff
@@ -187,6 +188,22 @@ public async Task<ActionResult> CreateApiReview(
 
             if (review != null)
             {
+                // Update package type if provided from controller parameter
```
praveenkuttappan: I don't think anyone will change a client package to mgmt or mgmt to client. So, I think we should update the package type only if it is not present and ignore the param if value is already set for review.

src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
```diff
@@ -239,7 +256,14 @@ public async Task<ActionResult> CreateApiReview(
             }
             else
             {
-                review = await _reviewManager.CreateReviewAsync(packageName: codeFile.PackageName, language: codeFile.Language, isClosed: false);
+                // Parse package type if provided, otherwise pass null for automatic classification
+                PackageType? parsedPackageType = null;
```
praveenkuttappan: Can this be moved outside if to avoid duplicate lines? Line 192 is doing the same.

src/dotnet/APIView/APIViewWeb/Managers/PullRequestManager.cs
```diff
@@ -248,15 +248,35 @@ private async Task ClosePullRequestAPIRevision(PullRequestModel pullRequestModel
...
-            PullRequestModel pullRequestModel, CodeFile baselineCodeFile, MemoryStream baseLineStream, string baselineFileName, CreateAPIRevisionAPIResponse responseContent)
+            PullRequestModel pullRequestModel, CodeFile baselineCodeFile, MemoryStream baseLineStream, string baselineFileName, CreateAPIRevisionAPIResponse responseContent, string packageType = null)
         {
             // fetch review for the package or create brand new review
             var review = await _reviewManager.GetReviewAsync(language: codeFile.Language, packageName: codeFile.PackageName);
             if (review == null)
             {
-                review = await _reviewManager.CreateReviewAsync(language: codeFile.Language, packageName: codeFile.PackageName, isClosed: false);
+                // Parse package type if provided, otherwise pass null
+                Models.PackageType? validPackageType = null;
```
praveenkuttappan: Same here. Remove duplicate lines in if and else.

src/dotnet/APIView/ClientSPA/src/app/_models/review.ts
```diff
@@ -6,6 +6,11 @@ export enum FirstReleaseApproval {
   All
 }
 
+export enum PackageType {
+  Data = 'Data',
```
praveenkuttappan: how does client in server gets mapped to 'Data' in UI side?

src/dotnet/APIView/ClientSPA/src/app/_components/review-page/review-page.component.ts
```diff
@@ -569,7 +572,7 @@ export class ReviewPageComponent implements OnInit {
     if (value) {
       this.reviewsService.toggleReviewApproval(this.reviewId!, this.activeApiRevisionId!, true).pipe(take(1)).subscribe({
         next: (review: Review) => {
-          this.review = review;
+          this.updateReview(review);
```
praveenkuttappan: What is this change for?

src/dotnet/APIView/ClientSPA/src/app/_models/review.ts
helen229: didn't update from previous  enum type

src/dotnet/APIView/APIViewWeb/Helpers/CommonUtilities.cs
```diff
@@ -3,7 +3,8 @@
 
 using System;
 using System.Linq;
-using System.Text.RegularExpressions;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
```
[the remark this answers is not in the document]
maririos: please undo this as it doesn't seem related to the removal of the code `BuysinessDayCalculator`

src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
maririos, line 278: not sure I understand why we need this

src/dotnet/APIView/APIViewWeb/Models/PackageModel.cs
```diff
@@ -1,24 +1,44 @@
...
-        [Name("New")]
-        public bool IsNew { get; set; }
-
-        [Name("GroupId")]
-        public string GroupId { get; set; }
-    }
-}
+using CsvHelper.Configuration.Attributes;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
```
maririos, line 5: you should be able to delete these 2 additions. We want APIView to only use 1 serializer and a while back we decided on not using `Newtonsoft`. If you still find this in the code, it is pending cleanup we have.

src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
helen229: because we need a UpdateReviewAsync method in ReviewManager level to wrap up _reviewsRepository.UpsertReviewAsync(review); so we can use it in Controller, we can't call _reviewsRepository.UpsertReviewAsync directly in controller

src/dotnet/APIView/APIViewWeb/Models/PackageModel.cs
helen229: updated

src/dotnet/APIView/APIViewWeb/Managers/PullRequestManager.cs
```diff
@@ -247,19 +247,25 @@ private async Task ClosePullRequestAPIRevision(PullRequestModel pullRequestModel
             await _pullRequestsRepository.UpsertPullRequestAsync(pullRequestModel);
         }
 
+        private static Models.PackageType? ParsePackageType(string packageType)
```
praveenkuttappan: nit: This can be helper method to avoid having this in two classes.

src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
```diff
@@ -187,6 +200,13 @@ public async Task<ActionResult> CreateApiReview(
 
             if (review != null)
             {
+                // Update package type if provided from controller parameter and not already set
+                if (parsedPackageType.HasValue && !review.PackageType.HasValue)
```
praveenkuttappan, line 195: I just realized. parsedPackageType will throw null exception when type is invalid.

src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
```diff
@@ -177,8 +178,20 @@ public async Task<ActionResult> CreateApiReview(
...
         }
 
-        private async Task<(ReviewListItemModel review, APIRevisionListItemModel apiRevision)> CreateAutomaticRevisionAsync(CodeFile codeFile, string label, string originalName, MemoryStream memoryStream, bool compareAllRevisions = false)
+        private static PackageType? ParsePackageType(string packageType)
         {
+            if (!string.IsNullOrEmpty(packageType) && Enum.TryParse<PackageType>(packageType, true, out var parsedPackageType))
+            {
+                return parsedPackageType;
+            }
+            return null;
```
praveenkuttappan: Can you add "Unknown" as enum value and default this as value in Review Model if value is missing?

tjprescott: Why prefer that over simply interpreting null as unknown? I know we have done that elsewhere in recent PRs

tjprescott: Actually, it was _this_ PR where she initially did have Unknown as a value, but why is that preferable to simply null? Is it a C# convention?

tjprescott: image

src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
```diff
@@ -178,8 +178,20 @@ public async Task<ActionResult> CreateApiReview(
...
+        }
+
+        private async Task<(ReviewListItemModel review, APIRevisionListItemModel apiRevision)> CreateAutomaticRevisionAsync(CodeFile codeFile, string label, string originalName, MemoryStream memoryStream, string packageType, bool compareAllRevisions = false)
+        {
+            // Parse package type once at the beginning
+            var parsedPackageType = ParsePackageType(packageType);
```
praveenkuttappan: Something like.
```suggestion
            var parsedPackageType =  !string.IsNullOrEmpty(pkgtype)? Enum.TryParse<PackageType>(packageType, true, out var parsedPackageType) : null;

File: src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs (hunk 18 above)
Comment (helen229) [reply src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs]: stick to the null.

File: src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs (hunk 17 above)
Comment (helen229) [reply src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs:195]: hasValue will show false if type is null

File: src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs [hunk 20]
```diff
@@ -177,8 +178,11 @@ public async Task<ActionResult> CreateApiReview(
             return StatusCode(statusCode: StatusCodes.Status500InternalServerError);
         }
 
-        private async Task<(ReviewListItemModel review, APIRevisionListItemModel apiRevision)> CreateAutomaticRevisionAsync(CodeFile codeFile, string label, string originalName, MemoryStream memoryStream, bool compareAllRevisions = false)
+        private async Task<(ReviewListItemModel review, APIRevisionListItemModel apiRevision)> CreateAutomaticRevisionAsync(CodeFile codeFile, string label, string originalName, MemoryStream memoryStream, string packageType, bool compareAllRevisions = false)
         {
+            // Parse package type once at the beginning
+            var parsedPackageType = !string.IsNullOrEmpty(packageType) && Enum.TryParse<PackageType>(packageType, true, out var result) ? (PackageType?)result : null;
```
praveenkuttappan, line 184:
```suggestion
            PackageType? parsedPackageType = !string.IsNullOrEmpty(packageType) && Enum.TryParse<PackageType>(packageType, true, out var result) ? (PackageType?)result : null;
```

src/dotnet/APIView/APIViewWeb/Managers/PullRequestManager.cs
```diff
@@ -248,15 +248,28 @@ private async Task ClosePullRequestAPIRevision(PullRequestModel pullRequestModel
         }
 
         private async Task CreateAPIRevisionIfRequired(CodeFile codeFile, string originalFileName, MemoryStream memoryStream,
-            PullRequestModel pullRequestModel, CodeFile baselineCodeFile, MemoryStream baseLineStream, string baselineFileName, CreateAPIRevisionAPIResponse responseContent)
+            PullRequestModel pullRequestModel, CodeFile baselineCodeFile, MemoryStream baseLineStream, string baselineFileName, CreateAPIRevisionAPIResponse responseContent, string packageType = null)
         {
+            var validPackageType = !string.IsNullOrEmpty(packageType) && Enum.TryParse<Models.PackageType>(packageType, true, out var result) ? (Models.PackageType?)result : null;
```
praveenkuttappan, line 259:
```suggestion
            PackageType? validPackageType = !string.IsNullOrEmpty(packageType) && Enum.TryParse<Models.PackageType>(packageType, true, out var result) ? (Models.PackageType?)result : null;
```

praveenkuttappan (approved):

maririos (commented):
Can you ensure that there are tests covering:

- Passing a valid `packageType`
- Passing an invalid `packageType`
- Omitting `packageType`
- Updating an existing review with and without `PackageType` set

PR #12401, opened 2025-10-07
[azsdkcli] Implement lint and format checks for JS

timovv:


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageSpecificChecks.cs
```diff
@@ -45,6 +45,8 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, Canc
             var result = await _processHelper.Run(new(
                     "pnpm",
                     ["run", "update-snippets"],
+                    "pnpm.cmd",
+                    ["run", "update-snippets"],
                     workingDirectory: packagePath
                 ),
```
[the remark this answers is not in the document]
timovv: bro it works

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageSpecificChecks.cs
```diff
@@ -67,4 +69,77 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, Canc
...
+                    workingDirectory: packagePath
+                ),
+                cancellationToken
+            );
+
+            if (result.ExitCode != 0)
+            {
+                _logger.LogError($"'pnpm run {subcommand}' failed with exit code {result.ExitCode}");
+
+                var nextSteps = fix ? "Review the linting errors and fix them manually." : "Run this tool in fix mode to automatically fix some of the errors.";
```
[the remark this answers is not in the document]
timovv: Strategic ambiguity was chosen here since the caller could be either coming from MCP or from the CLI

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageSpecificChecks.cs
```diff
@@ -45,6 +45,8 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, Canc
             var result = await _processHelper.Run(new(
                     "pnpm",
                     ["run", "update-snippets"],
+                    "pnpm.cmd",
```
l0lawrence: why are we adding pnpm.cmd as well?

l0lawrence: does processHelper run "or" the two dependeing on windows or not?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageSpecificChecks.cs
l0lawrence, line 90: Am I understanding this correctly - so if they have chosen to fix the errors and they aren't fixed then it tells them to re-run with -fix. what if they are in mcp and havent passed through fix bc they dont know it exists?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageSpecificChecks.cs
timovv: Yeah, it runs the .cmd in Windows, we need to specify otherwise it doesn't work in Windows I think

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageSpecificChecks.cs
timovv: Copilot should know that the fix option exists right? So it would call it again with fix enabled. At least that's my thought process.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageSpecificChecks.cs
timovv: ok confirmed we don't need it since we put cmd.exe in front for windows, thanks

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageSpecificChecks.cs
benbp: I think one downside of returning `CLICheckResponse` at this level is we have helper functions returning next steps for an unknown caller. If it was the `PackageCheckTool` inspecting the output and generating next steps then the next steps message could be more accurate (e.g. run this with `-f` or run this with `fixCheckErrors: true`). I don't have a better idea though.

Issue #10228, opened 2025-04-07
Java duplicate artifactIds: detect-api-changes.yml and Detect-Api-Changes.ps1


JimSuplizio:
detect-api-changes.yml runs on PullRequests. It loops through all of the PackageInfo files but, does call the FindArtifactForApiReviewFn which is something that Create-APIReview.ps1 also calls. 

Further, the main body of Detect-Api-Changes.ps1 is looping through PackageInfo files and calling Should-Process-Package but instead of passing in the PackageInfo object from the file already loaded, it's passing in the artifactName created with
`$pkgArtifactName = $packageMetadata.ArtifactName ?? $packageMetadata.Name`
and reloading the PackageInfo file with
`$pkgPropPath = Join-Path -Path $configFileDir "$packageName.json"`
instead of just passing in the PackageInfo file. There's nothing else calling the Should-Process-Package except the main body.

PR #12416, opened 2025-10-08
Update detect api changes script

raych1:
Pass `packageInfo` parameter to `Find-Language-Artifacts-For-Apireview` when supported. It will be used in case of Java to fetch the `groupId`.
- Add dynamic parameter detection
- Maintain backward compatibility with existing language implementations

Refer to #10228 for details.

### DIFF
--- a/eng/common/scripts/Detect-Api-Changes.ps1
+++ b/eng/common/scripts/Detect-Api-Changes.ps1
@@ -86,7 +86,7 @@ function Should-Process-Package($packageInfo)
     $modifiedFiles  = @(Get-ChangedFiles -DiffPath "$packagePath/*" -DiffFilterType '')
     $filteredFileCount = $modifiedFiles.Count
     LogInfo "Number of modified files for package: $filteredFileCount"
-    return ($filteredFileCount -gt 0 -and $pkgInfo.IsNewSdk)
+    return ($filteredFileCount -gt 0 -and $packageInfo.IsNewSdk)
 }
 
 function Log-Input-Params()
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Detect-Api-Changes.ps1
```diff
@@ -126,24 +117,42 @@ $responses = @{}
...
-    $packageMetadata = Get-Content $packagePropFile | ConvertFrom-Json
-    $pkgArtifactName = $packageMetadata.ArtifactName ?? $packageMetadata.Name
+    $packageInfo = Get-Content $packageInfoFile | ConvertFrom-Json
+    $pkgArtifactName = $packageInfo.ArtifactName ?? $packageInfo.Name
 
     LogInfo "Processing $($pkgArtifactName)"
 
-    $packages = &$FindArtifactForApiReviewFn $ArtifactPath $pkgArtifactName
+    # Check if the function supports the packageInfo parameter
+    $functionInfo = Get-Command $FindArtifactForApiReviewFn -ErrorAction SilentlyContinue
```
weshaggard, line 133: I like this approach to allow mitigation of the breaking change and allow folks to update incrementally. However, I would like to file an issue to get everyone over to the new method and then remove those once everyone is converted.

raych1: Do we want to have each language implementation to have same signature, like `function Find-java-Artifacts-For-Apireview($artifactDir, $pkgName, $packageInfo = $null)`? It's ok to leave them as is since the new parameter `packageInfo` won't be used in other languages scenarios.

weshaggard: I would like them to all have the new simplified signature.

raych1: Created issue https://github.com/Azure/azure-sdk-tools/issues/12443

PR #12432, opened 2025-10-09
Support v2 Rust performance tests

LarryOsterman:


tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
```diff
@@ -66,8 +79,10 @@ public override async Task<IterationResult> RunAsync(
             string profilerOptions,
             object context)
         {
+            // Set AZURE_TEST_MODE environment variable to "live" before invoking the compiler
+            Environment.SetEnvironmentVariable("AZURE_TEST_MODE", "live");
             // set up the params for cargo
-            string finalParams = $"bench -- \"{testName}\"";
+            string finalParams = $"test --package {primaryPackage} --test performance_tests -- --test-results {_targetResultsDirectory}/{testName}-results.json \"{testName}\" {arguments}";
```
heaths: Does `--test performance_tests` mean that people can only have one test file under their `perf/` directory? Seems we need a way for them to support multiple. I had multiple in KV .NET, so there should be a way to pass in tests or maybe this code needs to discover them. The test name is the module (file name), so a directory listing of `perf/` would suffice.

LarryOsterman: Nope - you can have any number of tests. However, there is only one test executable, named "performance_tests".

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
```diff
@@ -129,21 +144,15 @@ public override async Task<string> CleanupAsync(string project)
...
-            var opsPerSecond = Math.Pow(10, 9) / nanos;
-
-            return opsPerSecond;
+        private double ExtractOpsPerSecond(string testName)
+        {
+            // the results are in the target directory under target/criterion/<testName>/new
+            var results = ParseFromJsonFile(Path.Combine(_targetResultsDirectory, $"{testName}-results.json"));
+
+            // once we have the timing of one operation( in nano seconds) divide one sec in nanos by the time to get ops/sec
+            var opsPerSecond = results.OperationsPerSecond;
```
RickWinter: Why not `return results.OperationsPerSecond;`?

RickWinter (approved):

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
heaths: As discussed offline, to better correlate with the "perf/" directory, could we just call this lone test / drive "perf" e.g., perf/perf.rs? Then everything devs need to worry about is just "perf".

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
LarryOsterman: Because I didn't think of doing that. The original code did a lot more.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
LarryOsterman: Tests renamed to "perf"

LarryOsterman:
Ping - I made some small changes to the test automation.

LarryOsterman:
Ping2 for @heaths and @RickWinter

PR #12436, opened 2025-10-09
Update error message for language exclusion justification tool

praveenkuttappan:
Changes as per additional review comments to update error message

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
@@ -397,7 +397,7 @@ namespace Azure.Sdk.Tools.Cli.Tools.ReleasePlan
                 return new DefaultCommandResponse
                 {
                     Message = sb.ToString(),
-                    NextSteps = excludedLanguages.Any() ? ["Prompt the user for justification for excluded languages and update it in the release plan."] : []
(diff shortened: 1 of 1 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -397,7 +397,7 @@ public async Task<DefaultCommandResponse> UpdateSDKDetailsInReleasePlan(int rele
                 return new DefaultCommandResponse
                 {
                     Message = sb.ToString(),
-                    NextSteps = excludedLanguages.Any() ? ["Prompt the user for justification for excluded languages and update it in the release plan."] : []
+                    NextSteps = excludedLanguages.Any() && string.IsNullOrEmpty(releasePlan.LanguageExclusionRequesterNote) ? ["Prompt the user for justification for excluded languages and update it in the release plan."] : []
```
maririos, line 400: what if the PM didn't provide a note? are we making it a requirement now?
or given we already have the `excludedLanguage` the code can directly look at the state of the exclusion property for that language

praveenkuttappan: This field is for approver to check why did they exclude a language, but that's not mandatory. This note is to guide the LLM to inform the user that they need to provide justification. We have a status at individual language level that's auto updated based on tspconfig.yaml and language selection.

Issue #12424, opened 2025-10-08
[AVC] Allow eval test cases to be in YAML format


tjprescott:
The main reason for this is so that multi-line strings can be included inline:

```yaml
foo: |
  Wow, my multi-line strings
  are so awesome!
```

Instead of:
```json
"foo": "\nWow, my multi-line strings\naresoawesome!"
```

Prompty uses YAML for its sample format, so this would make it easier to develop test cases in prompt and then preserve them in an eval test case.

Issue #12402, opened 2025-10-07
[AVC] Make EvalRunner natively support and parallelize test cases


tjprescott:
Currently, `EvalsRunner` runs a single folder by joining its test files into a single `jsonl` file and running those test cases in parallel.

If you want to run multiple batches of test cases, or the same batch multiple times, those are handled manually in `run.py` or `avc eval run`. Unfortunately, they are run in series...

It would be nice if EvalRunner simply handled the parallelization automatically. If `num_runs=3` then it should run all three instances of the test at the same time instead of sequentially, which makes the test take 3 times as long.

For test cases, it should accept all test paths and create the minimal number of test batches to run them all... then it should parallelize those.

For example, if you pass two test folders, each with `test-config.yaml`, it would have to create a minimum of two test batches. However, it should run those batches in parallel.

There's plenty of prior art within the `_apiview_reviewer.py` file to guide handling the parallelization.

PR #12439, opened 2025-10-09
[AVC] Granular run options + parallelize

marcarro:
Rebased/Adapted changes from https://github.com/Azure/azure-sdk-tools/pull/12392 onto the new structure.

 ## Changes
- Added new discovery system in `_discovery.py` that collects workflow based on input test path(s) as `EvaluationTarget` objects that are then passed into the `EvaluationRunner`
- Replaced sequential execution with parallel execution
- `--test-paths/-p` is now optional. When skipped, it looks inside `evals/tests` and runs all workflows.

## Usage
```bash
# Run all workflows
/evals> python run.py

# Run a workflow
/evals> python run.py -p tests/mention_action

# Run a testcase
/evals> python run.py -p tests/mention_action/no_action.json

# Mix it up (e.g. workflow + testcase + workflow) 
/evals> python run.py -p tests/mention_action tests/thread_resolution_action/csharp_update_kb_generic.json tests/apiview_legacy/small 
```

Closes https://github.com/Azure/azure-sdk-tools/issues/12402, Closes https://github.com/Azure/azure-sdk-tools/issues/12424

### DIFF
--- a/packages/python-packages/apiview-copilot/ci.yml
+++ b/packages/python-packages/apiview-copilot/ci.yml
@@ -42,7 +42,7 @@ stages:
         steps:
           - script: |
               python -m pip install --upgrade pip
-              pip install -r requirements.txt
+              pip install -r dev_requirements.txt
               pip install pytest
               pytest tests --junitxml=pytest-results.xml
             workingDirectory: packages/python-packages/apiview-copilot


packages/python-packages/apiview-copilot/evals/_config_loader.py
```diff
@@ -93,13 +93,14 @@ def load_workflow_config(path: str | os.PathLike) -> WorkflowConfig:
     input_path = Path(path).resolve()
 
     # require a directory (we expect a folder containing test-config.yaml)
-    if not input_path.exists() or not input_path.is_dir():
-        _fail(f"Workflow path must be a directory containing 'test-config.yaml': {input_path}")
+    if not input_path.exists():
+        _fail(f"Workflow path must be either a directory containing 'test-config.yaml' or a sibling of 'test-config.yaml': {input_path}")
```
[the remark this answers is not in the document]
marcarro: Input paths can be directories (i.e. single workflows)

marcarro:
Closes https://github.com/Azure/azure-sdk-tools/issues/12424, testcases should now be able to be written in .yaml files. There is an example in `evals/tests/mention_action/no_action.yaml`

packages/python-packages/apiview-copilot/evals/_runner.py
```diff
@@ -1,130 +1,313 @@
...
-        )
-
-    def create_temporary_jsonl_file(self, directory: pathlib.Path) -> pathlib.Path:
+        return {}
+
+    def in_ci(self) -> bool:
+        return bool(os.getenv("TF_BUILD"))
+    
+    @property
+    def azure_ai_project(self) -> dict[str, str]:
```
tjprescott: These look like Java getters. Why? You can just make these properties public.

packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -140,22 +140,6 @@ def show_results(self, processed_results: dict) -> None:
         """
         pass
 
-    def post_process(
```
tjprescott, line 143: Why was this removed?

packages/python-packages/apiview-copilot/evals/_discovery.py
```diff
@@ -0,0 +1,204 @@
...
+
+from evals._config_loader import WorkflowConfig, load_workflow_config
+
+
+@dataclasses.dataclass(frozen=True)
+class EvaluationTarget:
+    """Represents a single evaluation unit - either a file or workflow.
+    """
+    path: Path
+    type: Literal["single_test", "workflow", "workflow_directory"]
```
tjprescott, line 18: So, "single_test" makes me worry... if I have tests A, B, C, and D in a folder with test-config, and I run `avc eval run -p A C` it should not consider A and C to be single tests... it should recognize that they both have the same configuration file and package them together. Does this do that?

Also, what's the difference between a "workflow" and a "workflow_directory"?

packages/python-packages/apiview-copilot/evals/_discovery.py
```diff
@@ -0,0 +1,204 @@
...
+    def summary(self) -> str:
+        """Human-readable summary of discovered targets."""
+        return (
+            f"Discovered {len(self.targets)} targets "
+            f"({self.total_test_files} test files, "
+            f"types: {', '.join(sorted(self.workflow_types))})"
+        )
+
+
+class EvaluationDiscovery:
```
tjprescott: Why is this a class with nothing but static methods? Why not just make `discover_targets` be a function that you call?

packages/python-packages/apiview-copilot/evals/_discovery.py
```diff
@@ -0,0 +1,204 @@
...
+    """Discovers and normalizes evaluation targets from various input formats."""
+    
+    @staticmethod
+    def discover_targets(test_paths: list[str] | None, tests_root: Path | None = None) -> DiscoveryResult:
```
tjprescott: This seems like a lot of code. The strategy I'd use is:

1. Have a dictionary of str, list[str]. The key is the config path, and the values are the test paths
2. iterate through all of the test paths and expand them, grouping them by the config path. If a target is a folder, expand to include all of the test files in that folder.
3. Once done, each key-value pair is the minimal bundles of tests that should be packaged together

packages/python-packages/apiview-copilot/evals/_runner.py
```diff
@@ -1,130 +1,313 @@
...
+                    **self._context.credential_kwargs,
+                )
+                all_run_results.append({jsonl_file.name: result})
+            
+            # Process results with evaluator
+            guideline_ids = set()
+            processed_results = evaluator.process_results(all_run_results, guideline_ids)
+
+            # NOTE: Skipping showing individual results to reduce noise in parallel runs
+            #    evaluator.show_results(processed_results)
```
tjprescott: Change this to a TODO. You probably should still show the results, once everything is finished.

packages/python-packages/apiview-copilot/evals/_runner.py
marcarro: Wasnt sure how accessible they should be. Will remove

packages/python-packages/apiview-copilot/evals/_custom.py
marcarro: All sub-procedures it was calling (establish baseline, calculate coverage) were removed, and the only implementation of `post_process` was in `CustomAPIViewReview`, which after removing the procedures basically did nothing, so it seemed to make sense to completely remove it altogether.

packages/python-packages/apiview-copilot/evals/_discovery.py
marcarro: A workflow is a single workflow (e.g. mention_action). A workflow directory is a directory that contains many different workflows (i.e. apiview_legacy, or even the entire `tests` folder)

tjprescott: If this type isn't used for anything but debugging, I'd recommend getting rid of it to just keep the code more concise (and avoid the risk that in the future someone does take a dependency on it).

packages/python-packages/apiview-copilot/evals/_runner.py
tjprescott: If they should be accessible, just store them as `self.azure_ai_project` instead of `self._azure-ai_project`. In python you _can_ get it either way, but you communicate intent. There are cases for a property-based get, but if you are just re-exposing an internal property, that's not one of them (though it's very common IIRC in Java)

packages/python-packages/apiview-copilot/tests/evals/discovery_test.py
```diff
@@ -0,0 +1,452 @@
...
+        self.name = name
+        self.kind = kind
+
+
+def test_discover_targets_empty_paths():
+    with tempfile.TemporaryDirectory() as temp_dir:
+        tests_root = Path(temp_dir)
+        
+        with pytest.raises(ValueError, match="No workflows found"):
+            discover_targets(None, tests_root)
```
tjprescott: I thought "empty paths" was supposed to return all things?

packages/python-packages/apiview-copilot/tests/evals/discovery_test.py
```diff
@@ -0,0 +1,452 @@
...
+
+def test_discover_targets_nonexistent_tests_root():
+    nonexistent = Path("/nonexistent/path")
+    
+    with pytest.raises(ValueError, match="Tests directory not found"):
+        discover_targets(None, nonexistent)
+
+
+@patch('evals._discovery.load_workflow_config')
+def test_discover_single_test_file(mock_load_config):
```
tjprescott: Rather than create all of these structured via code, it would likely be cleaner and easier to create fake test hierarchy in the `tests/evals` folder (`tests/evals/paths` or something) and target those. It should simplify the code her significantly.

packages/python-packages/apiview-copilot/cli.py
```diff
@@ -1016,12 +1021,20 @@ def load_arguments(self, command):
             ac.argument(
                 "num_runs", type=int, options_list=["--num-runs", "-n"], help="Number of times to run the test case."
             )
+            ac.argument(
+                "max_parallel", 
+                type=int, 
+                options_list=["--max-parallel", "-j"], 
+                default=3,
+                help="Maximum number of evaluations to run in parallel (default: 3, set to 1 for sequential)"
+            )
```
tjprescott: I would think the default should be dynamic based on the number of CPU cores you have. Might make sense to default to None and set this in code.

packages/python-packages/apiview-copilot/tests/evals/discovery_test.py
marcarro: When a directory with no workflows or sub-workflows inside is passed in, the behavior is that nothing executes. Everything is run when *no paths* are passed in

tjprescott: Got it.

tjprescott (approved):

PR #12444, opened 2025-10-09
Add script/pipeline for manual cpex attestation

smw-ms:


eng/pipelines/cpex-attestation-manual.yml
```diff
@@ -0,0 +1,54 @@
...
+    - ProductSku
+    - Feature
+  displayName: Target Type
+- name: Status
+  type: int
+  displayName: Status (0 - Incomplete, 1 - Complete, 3 - N/A)
+  values:
+    - 0
+    - 1
+    - 3
```
praveenkuttappan, line 33: why skipped 2?  :)

smw-ms: maybe they wanted to make sure that if a user entered 2 thinking it meant complete that it wouldn't mess with anything. not too sure.

PR #12450, opened 2025-10-10
[azsdkcli] lint/format python

l0lawrence:


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
```diff
@@ -27,6 +27,43 @@ public PythonLanguageSpecificChecks(
     }
     public string SupportedLanguage => "Python";
 
+    /// <summary>
+    /// Ensures azure-sdk-tools is installed and available for Python operations.
+    /// </summary>
+    /// <param name="packagePath">The package path to determine repository root</param>
+    /// <param name="cancellationToken">Cancellation token</param>
+    /// <returns>CLICheckResponse indicating success or failure</returns>
+    private async Task<CLICheckResponse> EnsureAzureSdkToolsInstalledAsync(string packagePath, CancellationToken cancellationToken = default)
```
benbp: I don't think we need to have the `Async` suffix for our methods. It's more useful when there are both async and non-async implementations.


```suggestion
    private async Task<CLICheckResponse> EnsureAzureSdkToolsInstalled(string packagePath, CancellationToken cancellationToken = default)
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
```diff
@@ -135,15 +172,120 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, Canc
...
+
+            // Ensure azure-sdk-tools is installed before proceeding
+            var installCheckResult = await EnsureAzureSdkToolsInstalledAsync(packagePath, cancellationToken);
+            if (installCheckResult.ExitCode != 0)
+            {
+                return installCheckResult;
+            }
+
+            // Check if Python is available
+            var pythonCheckResult = await _processHelper.Run(new("python", ["--version"], timeout: TimeSpan.FromSeconds(10)), cancellationToken);
```
benbp: Do we need to check that it's python 3 and not python 2? Or is that not really as much of a concern these days.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
```diff
@@ -135,15 +172,120 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, Canc
...
+                ("azpysdk pylint", new[] { "azpysdk", "pylint", packagePath }),
+                ("azpysdk mypy", new[] { "azpysdk", "mypy", packagePath }),
+                ("azpysdk veriftypes", new[] { "azpysdk", "veriftypes", packagePath }),
+                ("pyright", new[] { "pyright", packagePath })
+            };
+
+            foreach (var (toolName, command) in lintingTools)
+            {
+                _logger.LogInformation("Executing command: {ToolName} - {Command}", toolName, string.Join(" ", command));
+                var result = await _processHelper.Run(new(command[0], command.Skip(1).ToArray(), workingDirectory: packagePath, timeout: timeout), cancellationToken);
```
benbp, line 155: We could drop the `await` and use `Task.WhenAll` here so the linting runs concurrently.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
l0lawrence: its not a super big concern min supported python version is 3.9 rn

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
```diff
@@ -27,6 +27,43 @@ public PythonLanguageSpecificChecks(
     }
     public string SupportedLanguage => "Python";
 
+    /// <summary>
+    /// Ensures azure-sdk-tools is installed and available for Python operations.
+    /// </summary>
+    /// <param name="packagePath">The package path to determine repository root</param>
+    /// <param name="cancellationToken">Cancellation token</param>
+    /// <returns>CLICheckResponse indicating success or failure</returns>
+    private async Task<CLICheckResponse> EnsureAzureSdkToolsInstalled(string packagePath, CancellationToken cancellationToken = default)
```
l0lawrence: For right now I have this here pre-verifySetup being merged to ensure its installed

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
```diff
@@ -135,15 +172,104 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, Canc
...
+            var lintingTasks = lintingTools.Select(async tool =>
+            {
+                var (toolName, command) = tool;
+                var result = await _processHelper.Run(new(command[0], command.Skip(1).ToArray(), workingDirectory: packagePath, timeout: timeout), cancellationToken);
+                _logger.LogInformation("{ToolName} completed with exit code {ExitCode}", toolName, result.ExitCode);
+                return (toolName, result);
+            });
+
+            // Wait for all linting tools to complete
+            var allResults = await Task.WhenAll(lintingTasks);
```
benbp: For this to be concurrent you'll also want to remove the `await` keyword above, otherwise it will still block on each call.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
l0lawrence: jenny is going to add a python process helper that will assist with this

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
```diff
@@ -134,15 +171,104 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, bool
...
+
+            var timeout = TimeSpan.FromMinutes(10); 
+
+            // Run multiple linting tools
+            var lintingTools = new[]
+            {
+                ("pylint", new[] { "azpysdk", "pylint", packagePath }),
+                ("mypy", new[] { "azpysdk", "mypy", packagePath }),
+                ("veriftypes", new[] { "azpysdk", "veriftypes", packagePath }),
+                ("pyright", new[] { "azpysdk", "pyright", packagePath })
```
l0lawrence: pyright is optional we can remove this

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
```diff
@@ -134,15 +171,104 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, bool
...
+            }
+
+            var timeout = TimeSpan.FromMinutes(10); 
+
+            // Run multiple linting tools
+            var lintingTools = new[]
+            {
+                ("pylint", new[] { "azpysdk", "pylint", packagePath }),
+                ("mypy", new[] { "azpysdk", "mypy", packagePath }),
+                ("veriftypes", new[] { "azpysdk", "veriftypes", packagePath }),
```
l0lawrence: lets remove this too for now

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
```diff
@@ -132,17 +132,88 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, bool
         }
     }
 
-    public async Task<CLICheckResponse> LintCodeAsync(string packagePath, bool fixCheckErrors = false, CancellationToken cancellationToken = default)
```
l0lawrence, line 135: build was failing

samvaity: Im confused you should have async since you are `await` ing [here](https://github.com/Azure/azure-sdk-tools/pull/12450/files#diff-766f3ddc1406b30800268baf841353f3eb980cf6783514db5c5dff659a576da9R155) ?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
```diff
@@ -132,17 +132,88 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, bool
...
+            var lintingTasks = lintingTools.Select(async tool =>
+            {
+                var (toolName, command) = tool;
+                var result = await _processHelper.Run(new(command[0], command.Skip(1).ToArray(), workingDirectory: packagePath, timeout: timeout), cancellationToken);
+                _logger.LogInformation("{ToolName} completed with exit code {ExitCode}", toolName, result.ExitCode);
+                return (toolName, result);
+            });
+
+            // Wait for all linting tools to complete
+            var allResults = Task.WhenAll(lintingTasks);
```
timovv: Need to await this?
```suggestion
            var allResults = await Task.WhenAll(lintingTasks);
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
```diff
@@ -132,17 +132,88 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, bool
...
+
+            // Analyze results
+            var failedTools = allResults.Result.Where(r => r.result.ExitCode != 0).ToList();
```
timovv: This is blocking right now but once you add the await above will become
```suggestion
            var failedTools = allResults.Where(r => r.result.ExitCode != 0).ToList();
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
timovv: I'm pretty sure it does need an await -- the processes should run in parallel but we still need to wait for all of them to complete before we can proceed? https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.whenall?view=net-9.0

The alternative is calling Task.Result which will block.

timovv: Oh I think we might not be talking about the same await :)

l0lawrence: guys I read this so wrong

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
timovv: The async will be needed again once `await` is added to the Task.WhenAll. But it's not needed right now since the use of the await is inside an async lambda function, not the main function body.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
timovv: Now that we have `Task.WhenAll` I think stuff should run in parallel even with this await

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageSpecificChecks.cs
l0lawrence: I think* because it was wrapped in the async tool -> it was fine with it, fixed it overall either way

l0lawrence:
will add changelog edit after refactor pr

PR #12454, opened 2025-10-10
Revise tsp-client usage instructions in README

raych1:
Updated usage instructions to use `npm exec --prefix` for running tsp-client commands. The reason is `tsp-client` requires the working directory to be the root folder of SDK repository.

### DIFF
--- a/eng/common/tsp-client/README.md
+++ b/eng/common/tsp-client/README.md
@@ -25,7 +25,8 @@ npm ci
 
 ## Usage
 
-After installation, you can run tsp-client using `npm exec --prefix` from the root directory of your SDK repository.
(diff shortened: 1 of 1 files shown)


eng/common/tsp-client/README.md
```diff
@@ -25,30 +25,31 @@ npm ci
 
 ## Usage
 
-After installation, you can run tsp-client by navigating to the directory and using npm exec:
+After installation, you can run tsp-client using `npm exec --prefix` from the root directory of your SDK repository.
```
JoshLove-msft: For several commands, you need to have the cwd be the client library dir, e.g. sync/generate. Not sure about init. But I think the key point is that when invoking the local package version of tsp-client, users shouldn't be expected to navigate to the package directory, but instead use --prefix and pass in the path to the eng/common.

Issue #12287, opened 2025-09-30
[Tool]: VerifySetup tool for V1


l0lawrence:
### 🔧 Tool Name

verifySetup

### 📄 What Should This Tool Do?

Given a list of requirements for the mcp release tools, the requirements for all the languages  (may be hardcoded as of right now) and a list of the languages the user wants to generate for, check if each requirement is installed for the specified languages. If any requirement is seen as not installed, then provide guidance on what to do.

### 📥 Inputs and 📤 Outputs

**Inputs**:
- List of languages we want to check requirements for

**Outputs**:
- Guidance on what to do for requirements that are not installed.

### 🏗️ Execution Context

Both MCP and CLI

### 💻 CLI Command and Flags (if applicable)

azsdk-verify-setup

ronniegeraghty:
@l0lawrence Do we need a sub issue for [.NET] List Requirements?

JennyPng:
[...]

PR #12457, opened 2025-10-10
Add Spec for VerifySetup

JennyPng:
Adding VerifySetup spec 

- https://github.com/Azure/azure-sdk-tools/issues/12287
- [original spec](https://loop.cloud.microsoft/p/eyJ3Ijp7InUiOiJodHRwczovL21pY3Jvc29mdC5zaGFyZXBvaW50LmNvbS8%2FbmF2PWN6MGxNa1ltWkQxaUlUSk9NR2xKUVZCS1N6QjVkVnBRUmpaTmFtcE1VVzF6WWpCSVIxVXpjakZLYVZwalZsSmFibll4U2pJeVlYbERaa3hCZW5sVGNGVkhNMDFIVHpSa1lWQW1aajB3TVVvM1EwUTFOVUl6VEZCRlVFOVhSMEZVVmtoWlEwTklSVW8zV1ZwRU0wbEtKbU05Sm1ac2RXbGtQVEUlM0QiLCJyIjpmYWxzZX0sInAiOnsidSI6Imh0dHBzOi8vbWljcm9zb2Z0LnNoYXJlcG9pbnQuY29tLzpmbDovci9jb250ZW50c3RvcmFnZS94OEZOTy14dHNrdUNSWDJfZk1USExYbmh1RGtqdG1sRmdtU1hPcGRJZTFNL0RvY3VtZW50JTIwTGlicmFyeS9Db3BpbG90L1VudGl0bGVkLmxvb3A%2FZD13NDc2MWMwNmFkZTk1NGU5NGE2MzNhMTU1YTRlYWExZDUmY3NmPTEmd2ViPTEmbmF2PWN6MGxNa1pqYjI1MFpXNTBjM1J2Y21GblpTVXlSbmc0Ums1UExYaDBjMnQxUTFKWU1sOW1UVlJJVEZodWFIVkVhMnAwYld4R1oyMVRXRTl3WkVsbE1VMG1aRDFpSVRKT01HbEpRVkJLU3pCNWRWcFFSalpOYW1wTVVXMXpZakJJUjFVemNqRkthVnBqVmxKYWJuWXhTakl5WVhsRFpreEJlbmxUY0ZWSE0wMUhUelJrWVZBbVpqMHdNVW8zUTBRMU5VUkxXVUpSVlZCR1R6WlRVa2hMVFUwMVFrdFhVMDlXU1U5V0ptTTlKVEpHSm1ac2RXbGtQVEVtWVQxTWIyOXdRWEJ3Sm5BOUpUUXdabXgxYVdSNEpUSkdiRzl2Y0Mxd1lXZGxMV052Ym5SaGFXNWxjaVo0UFNVM1FpVXlNbmNsTWpJbE0wRWxNakpVTUZKVVZVaDRkR0ZYVG5saU0wNTJXbTVSZFdNeWFHaGpiVlozWWpKc2RXUkROV3BpTWpFNFdXbEZlVlJxUW5CVFZVWlJVMnR6ZDJWWVZtRlZSVmt5VkZkd2NWUkdSblJqTWtsM1UwVmtWazB6U1hoVGJXeGhXVEZhVTFkdE5USk5WVzk1VFcxR05WRXlXazFSV0hBMVZUTkNWbEo2VGs1U01EZ3dXa2RHVVdaRVFYaFRhbVJFVWtSVk1WRnFUazFWUlZaUlZERmtTRkZXVWxkVFJteEVVVEJvUmxOcVpGcFhhMUY2VTFWdkpUTkVKVEl5SlRKREpUSXlhU1V5TWlVelFTVXlNbVppWWpFellUY3hMV0l6TURJdE5EQTBOUzFpTVdFM0xUZzNOMkZrT0RRMFlUazRPQ1V5TWlVM1JBJTNEJTNEIiwiciI6ZmFsc2V9LCJpIjp7ImkiOiJmYmIxM2E3MS1iMzAyLTQwNDUtYjFhNy04NzdhZDg0NGE5ODgifX0%3D)

### DIFF
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -76,7 +76,7 @@ What are we trying to achieve with this design?
 
 _Known cases where this approach doesn't work or has limitations._
 
-#### Exception 1: Hard-Coded Requirements (V1 Limitation)
+#### Exception 1: Hard-Coded Requirements (Scenario 1 Limitation)
 
 **Description:**
 
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -278,7 +278,7 @@ To fix missing requirements:
   3. Add MAVEN_HOME/bin to your PATH environment variable
   4. Restart your IDE
 
-After installing, run 'azsdk verify-setup --langs python' again to verify.
+After installing, run 'azsdk verify-setup --langs python,java' again to verify.
 ```
 
 #### MCP Agent Interaction
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -227,10 +227,10 @@ azsdk verify-setup --all
 }
 ```
 
-**When Automatically Invoked**:
+**When Invoked**:
 
-1. **MCP Server Start**: Automatically check requirements for detected language when server starts
-2. **Before Generation**: When user requests SDK generation, verify requirements first
+- **Before Tool Calls**: When user requests a tool, verify requirements first
+- **On user request**: When user asks the agent to check their setup, environment, installations, etc.
 
 ### User Experience
 
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -149,8 +149,8 @@ Developer experience depends on this decision - blocking is safer but less infor
 
 The `verify-setup` tool checks whether required dependencies for SDK development are installed in the environment. It operates in two modes:
 
-1. **Automatic**: Runs when MCP server starts, checking requirements for the detected language
-2. **Explicit**: Developer or agent explicitly invokes the tool with specific language(s)
+1. **Implicit**: Agent will suggest to run this tool before other tools, checking requirements for the detected language of the repo
+2. **Explicit**: Developer or agent explicitly invokes the tool
 
 The tool uses a hard-coded registry of requirements (V1) that includes:
 
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -15,15 +15,15 @@
 
 ## Definitions
 
-**Environment Setup**: The process of ensuring all required tools, SDKs, and dependencies are installed and properly configured for SDK development work.
+**Environment Setup**: The process of ensuring all required tools and dependencies are installed and properly configured for SDK development work.
 
-**Requirements**: Language-specific and core tools needed to successfully run Dev Inner Loop MCP tools (e.g., TypeSpec compiler, language SDKs, build tools).
+**Requirements**: Language-specific and core tools needed to successfully run Azure SDK MCP tools (e.g., TypeSpec compiler, language SDKs, build tools).
 
 **Verification Check**: A command or script that determines whether a specific requirement is installed and accessible in the developer's environment.
 
-**Lazy Loading**: Checking only the requirements for the language(s) currently being worked on, rather than checking all languages upfront.
+**Lazy Loading**: Checking only the requirements for the repository currently being worked on, rather than checking all requirements upfront.
 
-**Source of Truth**: A single, authoritative configuration that defines what requirements are needed for each language and how to install them.
+**Source of Truth**: A single, authoritative configuration that defines what requirements are needed for each tool and how to install them.
 
 ---
 
@@ -36,12 +36,8 @@
 - **.NET**: Developers manually discover missing dependencies when build or generation commands fail. No proactive verification.
 - **Java**: Same as .NET - reactive discovery through failures.
 - **JavaScript**: Same as .NET - reactive discovery through failures.
-- **Python**: Some repos have `azpysdk` which can be checked, but no unified verification across all required tools.
+- **Python**: `azure-sdk-for-python` has `azpysdk` which can be used for some validation checks, but no unified verification across all required tools.
 - **Go**: Same as .NET - reactive discovery through failures.
-
-**Data plane vs. Management plane:**
-
-Both data plane and management plane SDK development currently lack proactive environment verification. TypeSpec-based generation (common in data plane) and Swagger-based generation (common in management plane) each have their own tooling requirements, but neither has automated verification.
 
 ### Why This Matters
 
@@ -151,7 +147,7 @@ Developer experience depends on this decision - blocking is safer but less infor
 
 ### Overview
 
-The `verify-setup` tool checks whether required dependencies for SDK development are installed in the developer's environment. It operates in two modes:
+The `verify-setup` tool checks whether required dependencies for SDK development are installed in the environment. It operates in two modes:
 
 1. **Automatic**: Runs when MCP server starts, checking requirements for the detected language
 2. **Explicit**: Developer or agent explicitly invokes the tool with specific language(s)
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -1,0 +1,101 @@
+# V1 Verify Setup Spec
+
+## Purpose
+
+- A tool that checks if requirements for MCP release tools are installed in the developer's environment
+- Informs the user what is missing and how to resolve it.
+
+
+## Assumptions / Scope for V1
+
+- Requirements are hard-coded
+- The developer is working with all five languages, and will generate all languages locally
+- Execution context: MCP and CLI
+    - CLI command: azsdkcli-verify-setup
+
+## Design
+### Prereq
+
+- There is a hard-coded list of all requirements for MCP release tools
+- For each requirement, there are instructions on how to check it and how to install it
+
+For ex.
+
+```
+"core":
+  [
+    ... // non language-specific requirements   
+  ]  
+"python":
+  [
+    {
+      "requirement": "azpysdk"
+      "check": ["azpysdk", "--help"]
+      "install_instruction": ["1. Ensure your venv is activated", "2. Run `python -m pip install eng/tools/azure-sdk-tools[build]", "3. Check that it was installed with `azpysdk --help`"]               
+    },
+    ...     
+  ],
+
+"java": [...]
+```
+
+
+### Inputs
+
+- `--langs`: List of languages to check requirements forBy default, checks requirements for the current repo's lang (can use LanguageResolver logic to detect)
+
+
+Example Usages: 
+
+- I'm in `azure-sdk-for-python` and I want to check Python requirements
+    - Ask agent: "check my environment setup"
+    - `azsdkcli-package-verify-setup --langs python`
+- I want to check that I have Java and Python requirements
+    - Ask agent: "check my environment setup for java and python"
+    - `azsdkcli-package-verify-setup --langs java,python`
+- I want to check that I have requirements for all languages
+    - Ask agent: "check my environment setup for all languages"
+    - `azsdkcli-package-verify-setup --all`
+
+### Outputs
+
+- What requirements are missing, and how to install them
+
+For ex.
+
+```
+"
+
+Python Requirements to Install:
+1. You're missing azpysdk. Ensure you've activated your virtual environment, and run:
+`python -m pip install eng/tools/azure-sdk-tools[build]`
+
+...
+
+Java Requirements to Install:
+1. ...
+
+"
+```
+
+## Flow
+### Invoking the tool:
+
+- Execute every time MCP server starts (using language detected from current repo path)
+- Execute when performing local generation for user requested languages.
+    - I want to locally generate my SDK for python and java 
+        - `azsdkcli-package-verify-setup --langs java,python`
+- User explicitly asks agent to verify setup / runs the CLI command
+- (V2 idea) Running another tool fails -> suggest to verify setup
+
+### Tool flow:
+
+- for each language, iterate through requirements and run check command; track missing requirements
+- for each missing requirement, get instructions for resolving, and return
+
+## Open Questions
+
+- In the future, when we want to dynamically get requirements rather than relying on a hard-coded source, how would we get the correct installation instructions? 
+- If multiple/all languages are passed in, and one language fails verifySetup, do we continue with remaining languages or block the whole flow?
+    - There seems to be group agreement over preference for lazy loading and not blocking the developer for languages that they're not actively working on
+
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -19,7 +19,7 @@
 
 **Requirements**: Language-specific and core tools needed to successfully run Azure SDK MCP tools (e.g., TypeSpec compiler, language SDKs, build tools).
 
-**Verification Check**: A command or script that determines whether a specific requirement is installed and accessible in the developer's environment.
+**Verification Check**: A command or script that determines whether a specific requirement is installed and accessible in the environment.
 
 **Lazy Loading**: Checking only the requirements for the repository currently being worked on, rather than checking all requirements upfront.
 
@@ -231,7 +231,6 @@ azsdk verify-setup --all
 
 1. **MCP Server Start**: Automatically check requirements for detected language when server starts
 2. **Before Generation**: When user requests SDK generation, verify requirements first
-3. **(V2 Idea) On Failure**: When another tool fails due to missing dependencies, suggest running verify-setup. 
 
 ### User Experience
 
@@ -267,6 +266,11 @@ Java Requirements:
 ---
 
 To fix missing requirements:
+📦 Maven:
+  1. Download the latest version of Maven.
+  2. Set MAVEN_HOME environment variable to the Maven installation path
+  3. Add MAVEN_HOME/bin to your PATH environment variable
+  4. Restart your IDE
 
 📦 azpysdk:
    1. Navigate to your azure-sdk-for-python repo
@@ -505,7 +509,7 @@ Too heavy for a V1 environment verification tool. This could be a future option 
    - Expected: Correctly identify which are installed vs. missing
 
 3. **Multi-Language**: Run with `--langs python,java`
-   - Expected: Check requirements for both languages, report combined results
+   - Expected: Check requirements for both languages, report combined results, and ensure no other languages were evaluated
 
 4. **Current Repo Detection**: Run without parameters in different language repos
    - Expected: Automatically detect and check correct language
@@ -542,6 +546,11 @@ Verify:
 - [Python MCP Implementation Reference](https://github.com/Azure/azure-sdk-for-python/blob/ad3e09230bfbf31030c375102ecef366f4bcb35a/eng/tools/mcp/azure-sdk-python-mcp/main.py)
 
 ---
+
+## Future Work
+
+- Determine how to dynamically get requirements, instead of relying on a hard-coded config
+- Explore invoking verifySetup when another tool fails due to missing dependencies
 
 ## Meeting Notes
 
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -36,7 +36,7 @@
 - **.NET**: Developers manually discover missing dependencies when build or generation commands fail. No proactive verification.
 - **Java**: Same as .NET - reactive discovery through failures.
 - **JavaScript**: Same as .NET - reactive discovery through failures.
-- **Python**: `azure-sdk-for-python` has `azpysdk` which can be used for some validation checks, but no unified verification across all required tools.
+- **Python**: `azure-sdk-for-python` has `azpysdk` which can be used for some validation checks, but no proactive verification.
 - **Go**: Same as .NET - reactive discovery through failures.
 
 ### Why This Matters
@@ -266,18 +266,18 @@ Java Requirements:
 ---
 
 To fix missing requirements:
-📦 Maven:
-  1. Download the latest version of Maven.
-  2. Set MAVEN_HOME environment variable to the Maven installation path
-  3. Add MAVEN_HOME/bin to your PATH environment variable
-  4. Restart your IDE
-
 📦 azpysdk:
    1. Navigate to your azure-sdk-for-python repo
    2. Ensure your virtual environment is activated
    3. Run: python -m pip install -e eng/tools/azure-sdk-tools[build]
    4. Verify installation with: azpysdk --help
 
+📦 Maven:
+  1. Download the latest version of Maven.
+  2. Set MAVEN_HOME environment variable to the Maven installation path
+  3. Add MAVEN_HOME/bin to your PATH environment variable
+  4. Restart your IDE
+  
 After installing, run 'azsdk verify-setup --langs python' again to verify.
 ```
 
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -31,7 +31,7 @@ For ex.
     {
       "requirement": "azpysdk"
       "check": ["azpysdk", "--help"]
-      "install_instruction": ["1. Ensure your venv is activated", "2. Run `python -m pip install eng/tools/azure-sdk-tools[build]", "3. Check that it was installed with `azpysdk --help`"]               
+      "install_instruction": ["1. Ensure your venv is activated", "2. Run `python -m pip install eng/tools/azure-sdk-tools[build]`", "3. Check that it was installed with `azpysdk --help`"]               
     },
     ...     
   ],
@@ -42,7 +42,7 @@ For ex.
 
 ### Inputs
 
-- `--langs`: List of languages to check requirements forBy default, checks requirements for the current repo's lang (can use LanguageResolver logic to detect)
+- `--langs`: List of languages to check requirements for. By default, checks requirements for the current repo's lang (can use LanguageResolver logic to detect)
 
 
 Example Usages: 
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -277,7 +277,7 @@ To fix missing requirements:
   2. Set MAVEN_HOME environment variable to the Maven installation path
   3. Add MAVEN_HOME/bin to your PATH environment variable
   4. Restart your IDE
-  
+
 After installing, run 'azsdk verify-setup --langs python' again to verify.
 ```
 
@@ -565,9 +565,9 @@ Verify:
    - Richard suggested lang-specific copilot-instructions.md (similar to PowerShell installation approach)
 
 3. **Lazy Loading**: Agreement on "lazy loading" approach
-   - Only check requirements for current language being developed
+   - Only check requirements for current repository being developed
    - Don't block the developer with requirements for languages they're not using
-   - Example: Dev in Python repo starts MCP → check only Python requirements
+   - Example: Dev in Python repo starts MCP → check only core and Python requirements
 
 4. **Tool vs. Feature Question**: Discussion whether this needs to be standalone tool
    - Alternative: When other MCP tools fail, they return specific installation instructions
--- a/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
+++ b/tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
@@ -1,101 +1,576 @@
-# V1 Verify Setup Spec
+# Spec: 1-env-setup - verify-setup
 
-## Purpose
+## Table of Contents
 
-- A tool that checks if requirements for MCP release tools are installed in the developer's environment
-- Informs the user what is missing and how to resolve it.
+- [Definitions](#definitions)
+- [Background / Problem Statement](#background--problem-statement)
+- [Goals and Exceptions/Limitations](#goals-and-exceptionslimitations)
+- [Design Proposal](#design-proposal)
+- [Alternatives Considered](#alternatives-considered)
+- [Open Questions](#open-questions)
+- [Implementation Plan](#implementation-plan)
+- [Testing Strategy](#testing-strategy)
 
+---
 
-## Assumptions / Scope for V1
+## Definitions
 
-- Requirements are hard-coded
-- The developer is working with all five languages, and will generate all languages locally
-- Execution context: MCP and CLI
-    - CLI command: azsdkcli-verify-setup
+**Environment Setup**: The process of ensuring all required tools, SDKs, and dependencies are installed and properly configured for SDK development work.
 
-## Design
-### Prereq
+**Requirements**: Language-specific and core tools needed to successfully run Dev Inner Loop MCP tools (e.g., TypeSpec compiler, language SDKs, build tools).
 
-- There is a hard-coded list of all requirements for MCP release tools
-- For each requirement, there are instructions on how to check it and how to install it
+**Verification Check**: A command or script that determines whether a specific requirement is installed and accessible in the developer's environment.
 
-For ex.
+**Lazy Loading**: Checking only the requirements for the language(s) currently being worked on, rather than checking all languages upfront.
+
+**Source of Truth**: A single, authoritative configuration that defines what requirements are needed for each language and how to install them.
+
+---
+
+## Background / Problem Statement
+
+### Current State
+
+**Current state per language:**
+
+- **.NET**: Developers manually discover missing dependencies when build or generation commands fail. No proactive verification.
+- **Java**: Same as .NET - reactive discovery through failures.
+- **JavaScript**: Same as .NET - reactive discovery through failures.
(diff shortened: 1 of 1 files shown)


tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+- [Alternatives Considered](#alternatives-considered)
+- [Open Questions](#open-questions)
+- [Implementation Plan](#implementation-plan)
+- [Testing Strategy](#testing-strategy)
+
+---
+
+## Definitions
+
+**Environment Setup**: The process of ensuring all required tools, SDKs, and dependencies are installed and properly configured for SDK development work.
```
maririos: what do you mean for SDKs?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+
+**Requirements**: Language-specific and core tools needed to successfully run Dev Inner Loop MCP tools (e.g., TypeSpec compiler, language SDKs, build tools).
```
maririos: because it covers "making sure the mcp works" then the scope of `environment setup` should be beyond Dev Inner Loop

benbp:
We also have tools that are not language specific with their own dependency requirements, so I think an approach that links a requirements definition at a tool or class level will be useful. For example, `azsdk_<tool category>_verify_env` and `azsdk <cmd-group> <cmd> --verify-env`.

I'm also wondering a bit whether a config based approach will make expressing hierarchies and/or references too complicated. We may have some dependencies that are common to multiple languages/tool areas, and if we express them in code it will be easier to compose dependency groups that can be applied to an area. I think it's a good idea to think about how we will keep this in sync with the moving targets that is the repos, but if we drive the source of truth to be from the tool itself (tool -> generated install docs -> repos/pipelines) then maybe the problem is easier to solve?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+2. **Friction for New Contributors**: New team members or open-source contributors struggle to get started
+3. **MCP Agent Limitations**: When the agent attempts to run tools and they fail due to missing dependencies, the agent can't distinguish between actual errors and setup issues
+4. **Context Switching Pain**: When developers work across multiple language repos, they must manually remember and install different requirements for each language
+
+**Agent vs. Manual Workflow:**
+
+- **With Agent (Current)**: Agent attempts a command → fails → developer must interpret error → manually install missing tool → retry
+- **With Agent (Proposed)**: Agent checks setup first → proactively tells developer what's missing and how to fix it → developer installs → proceed with confidence
+- **Manual CLI (Current)**: Developer runs command → fails → searches documentation or asks teammates → installs tool → retry
+- **Manual CLI (Proposed)**: Developer explicitly runs `azsdk verify-setup` → sees clear list of missing items → follows install instructions → proceeds
```
benbp, line 56: Echoing my main comment, I think `--verify-setup` as a flag for a command will be easier to reason about for users. If I'm running some command then I can check the setup verification for that command or for the parent command group (so `azsdk package --verify-setup` could validate for the language a package is being generated for, etc.).

benbp: Possibly what I'm suggesting is out of scope, and we would have two modes - verify setup for a whole language scope, and verify setup just for a tool/command area.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+
+---
+
+## Definitions
+
+**Environment Setup**: The process of ensuring all required tools, SDKs, and dependencies are installed and properly configured for SDK development work.
+
+**Requirements**: Language-specific and core tools needed to successfully run Dev Inner Loop MCP tools (e.g., TypeSpec compiler, language SDKs, build tools).
+
+**Verification Check**: A command or script that determines whether a specific requirement is installed and accessible in the developer's environment.
```
maririos:
```suggestion
**Verification Check**: A command or script that determines whether a specific requirement is installed and accessible in the environment.
```

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+
+**Lazy Loading**: Checking only the requirements for the language(s) currently being worked on, rather than checking all languages upfront.
```
maririos: This will work beyond what each language needs, right?

```suggestion
**Lazy Loading**: Checking only the requirements for the tools(s) currently being worked on, rather than checking all tools requirements upfront.
```

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+
+**Source of Truth**: A single, authoritative configuration that defines what requirements are needed for each language and how to install them.
```
maririos:
```suggestion
**Source of Truth**: A single, authoritative configuration that defines what requirements are needed for each tool and how to install them.
```

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+## Background / Problem Statement
+
+### Current State
+
+**Current state per language:**
+
+- **.NET**: Developers manually discover missing dependencies when build or generation commands fail. No proactive verification.
+- **Java**: Same as .NET - reactive discovery through failures.
+- **JavaScript**: Same as .NET - reactive discovery through failures.
+- **Python**: Some repos have `azpysdk` which can be checked, but no unified verification across all required tools.
```
maririos: what do you mean with some repos? could this be more specific?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+- **Go**: Same as .NET - reactive discovery through failures.
+
+**Data plane vs. Management plane:**
+
+Both data plane and management plane SDK development currently lack proactive environment verification. TypeSpec-based generation (common in data plane) and Swagger-based generation (common in management plane) each have their own tooling requirements, but neither has automated verification.
```
maririos: Management plane is also TypeSpec and most of the SDKs are from TypeSpec :)
One big difference is that for management plane, there is a script that installs all of the tooling in the machine that does the SDK generation. @raych1 has also mentioned this in meetings.

Not sure if it is worth calling this out, or completely remove it

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+- [ ] Support language-specific requirement checking (lazy loading based on current context)
+- [ ] Work seamlessly in both MCP (agent) and CLI execution modes
+- [ ] Minimize developer friction when working across multiple language repositories
+- [ ] Establish a foundation for dynamic requirement discovery in future versions
+
+### Exceptions and Limitations
+
+_Known cases where this approach doesn't work or has limitations._
+
+#### Exception 1: Hard-Coded Requirements (V1 Limitation)
```
maririos: @ronniegeraghty did we agree on `milestone`?

```suggestion
#### Exception 1: Hard-Coded Requirements (Milestone 1 Limitation)
```

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+- Potential drift: Hard-coded requirements may become outdated if not regularly maintained
+- No single source of truth across repos
+
+**Workaround:**
+
+For V1, this is acceptable to prove the concept. V2 will address dynamic requirement discovery (see Open Questions).
+
+---
+
+#### Exception 2: Installation Instruction Specificity
```
maririos, line 97: this was so helpful to read :) thank you for adding it

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+
+---
+
+## Design Proposal
+
+### Overview
+
+The `verify-setup` tool checks whether required dependencies for SDK development are installed in the developer's environment. It operates in two modes:
+
+1. **Automatic**: Runs when MCP server starts, checking requirements for the detected language
```
maririos: how are the languages detected?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+{
+    "categories": {
+        "core": [...],
+        "java": [...],
+        "javascript": [...],
+        "dotnet": [...],
+        "go": [...],
+        "python": [
+            {
+                "requirement": "python",
```
maririos, line 177: why is the category repeated here? to address "core" cases?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+### Detailed Design
+
+#### Requirements Registry Structure
+
+The requirements are organized into categories based on language, in addition to a "core" section for non-language specific requirements:
+
+```json
+{
+    "categories": {
+        "core": [...],
```
maririos, line 170: talking about core, is the spec repo part of this?

It would be great if `verify-setup` step will work if invoked from the specs repository.
what would be the implications for that too?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+{
+  "languages": ["python", "java"],
+}
+```
+
+**When Automatically Invoked**:
+
+1. **MCP Server Start**: Automatically check requirements for detected language when server starts
+2. **Before Generation**: When user requests SDK generation, verify requirements first
+3. **(V2 Idea) On Failure**: When another tool fails due to missing dependencies, suggest running verify-setup.
```
maririos: it might be good to move this to another place that would be more easily discovered when working on V2

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+  ✅ Python 3.8+
+  ❌ azpysdk - MISSING
+
+Java Requirements:
+  ✅ Java JDK 11+
+  ❌ Maven - MISSING
+
+---
+
+To fix missing requirements:
```
maririos, line 268: should this be per language?
what about the error in Java?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+**Why not chosen:**
+
+Too heavy for a V1 environment verification tool. This could be a future option for developers who want container-based development.
+
+---
+
+## Open Questions
+
+- [ ] **Question 1**: How should we handle dynamic requirement discovery in V2?
+  - Context: Hard-coding requirements in V1 works but isn't maintainable long-term
```
maririos, line 429: have you engaged with EngSys to see if they have some ideas on how this should happen? If not, that would be a good step for V2

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+**Test Scenarios**:
+
+1. **Clean Environment**: Run in fresh environment with no tools installed
+   - Expected: All requirements reported as missing with instructions
+
+2. **Partial Installation**: Install some but not all requirements
+   - Expected: Correctly identify which are installed vs. missing
+
+3. **Multi-Language**: Run with `--langs python,java`
+   - Expected: Check requirements for both languages, report combined results
```
maririos: and make sure no other languages were evaluated

maririos (commented):
Overall looks great! my comments are for specific clarifications that would help with clarity

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+    - Parse `.devcontainer/devcontainer.json` files from each repo
+    - Use CI pipeline definitions (e.g., `ci.yml`) as source of truth
+    - Create a new eng/common config file specifically for dev requirements
+    - Leverage devcontainers CLI to interrogate container definitions
+  - **Proposal**: Explore eng/common config file that each repo can customize, with a standard schema
+
+- [ ] **Question 2**: Should installation instructions be in per-language copilot-instructions.md instead?
+  - Context: copilot-instructions already has instructions for installing Powershell if it's missing
+
+- [ ] **Question 3**: Should each tool be responsible for checking their own requirements instead of having a central tool?
```
m-redding, line 440: I was wondering about this since I noticed the language checks are already checking their own requirements. But I don't think it would have to be an either/or thing. It can be its own tool and other tools could also just call parts of the implementation of this tool or something. Imo the tool is helpful on its own even if the other tools are also checking requirements.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,576 @@
...
+
+```json
+{
+  "languages": ["python", "java"],
+}
+```
+
+**When Automatically Invoked**:
+
+1. **MCP Server Start**: Automatically check requirements for detected language when server starts
```
m-redding: When you say starts do you mean when the server itself is started up? Where does that output end up?

m-redding (approved):
overall tool design looks good for .NET

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: I'm using the existing abstraction for implementing language specific logic that Timo added https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/docs/per-language.md

the language of the repo is detected by looking for the Language-Settings.ps1 file in the repo - from testing though, it seems like copilot very frequently will just provide the correct language arg

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: The categories are just intended to organize the requirements, so the list under "python" category contains every installation that needs to be checked when developing for python

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: The intention is when the user types something into copilot for the first time, that can automatically start the MCP server, and we want copilot to just suggest/run the verifySetup tool in the first interaction

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,581 @@
...
+
+---
+
+## Goals and Exceptions/Limitations
+
+### Goals
+
+What are we trying to achieve with this design?
+
+- [ ] Proactively verify that required dependencies are installed before running MCP operations
```
samvaity, line 68: The goal would be to lazy load env requirements based on the package path not proactively or upfront?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
samvaity: I agree, loved this little callout for specificity.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,581 @@
...
+**Description:**
+
+If a developer requests verification for multiple languages (e.g., `--langs python,java`) and one language's requirements fail, it's unclear whether to:
+
+- Block the entire operation
+- Continue checking remaining languages and report all failures
+
+**Impact:**
+
+Developer experience depends on this decision - blocking is safer but less informative, while continuing provides more information but may be confusing.
```
samvaity, line 126: ✅

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,581 @@
...
+   3. Run: python -m pip install -e eng/tools/azure-sdk-tools[build]
+   4. Verify installation with: azpysdk --help
+
+📦 Maven:
+  1. Download the latest version of Maven.
+  2. Set MAVEN_HOME environment variable to the Maven installation path
+  3. Add MAVEN_HOME/bin to your PATH environment variable
+  4. Restart your IDE
+
+After installing, run 'azsdk verify-setup --langs python' again to verify.
```
samvaity: nit: 
```suggestion
After installing, run 'azsdk verify-setup --langs python,java' again to verify.
```

samvaity (approved):

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
ronniegeraghty:
```suggestion
#### Exception 1: Hard-Coded Requirements (Scenario 1 Limitation)
```

I went with the phrasing `Scenario 1` to keep it related to the fact this is all based around a scenario, but removed the `V` so there wasn't any more confusion on if this was tied to a specific version release of the tool

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
ronniegeraghty: I believe output from an MCP can be seen in VSCode's Output panel: 
image
You just need to choose to view output from the specific MCP server in the dropdown on the top right.
So, errors that happen during start up would be visible to the user here. And when something errors, VSCode usually pops this panel open automatically.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
ronniegeraghty: Agree with @m-redding. Users will usually want to do all necessary installation first, rather than keep installing new things as they go through the process, and possible lose their place in the process. 

It would also be interesting to see if having to do installation tasks in the middle of the process would cause Copilot issue with tracking where it is in the process. 🤔

ronniegeraghty (approved):

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
benbp: IMO the MCP logging should not be used by tooling to surface messages we expect the user to read, it's more for debugging if something goes catastrophically wrong. The tool responses to the LLM and it's corresponding summary should be the only interaction point.

Similarly we should not run any expensive operations as part of server start as I don't think it fits within the assumptions of what an MCP server will do (expose tools to the LLM and run them). Having instructions for the LLM to call verify or having tools call it under the hood seems more in line with conventions here.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: yes, running it from rest-api-specs can only check for the core requirements, not any language-specific requirements

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,581 @@
...
+Java Requirements:
+  ✅ Java JDK 11+
+  ❌ Maven - MISSING
+
+---
+
+To fix missing requirements:
+📦 azpysdk:
+   1. Navigate to your azure-sdk-for-python repo
+   2. Ensure your virtual environment is activated
```
l0lawrence, line 271: will we have instructions to tell copilot to ask the user to choose the venv they want to run in?

JennyPng: yes this can be added to the instructions in the python SDK repo

l0lawrence (approved):

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,581 @@
...
+Too heavy for a V1 environment verification tool. This could be a future option for developers who want container-based development.
+
+---
+
+## Open Questions
+
+- [ ] **Question 1**: How should we handle dynamic requirement discovery in V2?
+  - Context: Hard-coding requirements in V1 works but isn't maintainable long-term
+  - Options:
+    - Parse `.devcontainer/devcontainer.json` files from each repo
```
raych1, line 431: IMO, prefer referencing the `devcontainer.json` if it’s actively maintained by the language team. Otherwise, keeping the configuration in a file under `eng/common` is acceptable since it shouldn’t require frequent updates.

Note that we don’t have a single `ci.yml` that includes all dependencies for every language.

raych1 (approved):

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
benbp: I added some ideas on this in https://github.com/Azure/azure-sdk-tools/pull/12457#issuecomment-3407751610 and based on the latest update to the notes it seems like the concept of a single source of truth has been discussed/brainstormed now?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
benbp: This is why I think the v1 hardcoded requirements might be better expressed in the c# source vs. json. Right now the prototype requires referencing a config file by language name, but I could imagine tool classes or methods having attributes that mark what requirements groups they require and then being able to use inheritance to group them all together vs. having to deal with a config format that will likely grow in complexity.

```
[SystemRequirements([RequirementCategories.Python])]
public class MyPythonTool()
{
    [SystemRequirements([RequirementCategories.AnUncommonPythonTool])]
    public DefaultCommandResponse SomeRareMethodWithALazyRequirement()
    {
    }
}
```

This could be premature optimization/over-engineering though. If all we want right now is a flat set of requirements for 5 languages, then it's probably not necessary.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: Ok, then I won't actually call the tool on server start, and just add to each copilot-instructions to use the tool before trying to run other tools?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
l0lawrence: I think milestone one for us with the assumptions we are given is a flat set of requirements, but we can most definitely look into this for milestone two

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
benbp: Yeah I think instructions works as a starting point. If we have some sort of tool-level requirements then I could also see the possibility of adding data to `.NextSteps` in error responses about calling the verify tool.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: The goal is for Copilot to proactively suggest running this tool before running other tools - but by default the tool will only check core requirements and the language requirements of the working repo. however the user could specify multiple languages to verify for as well

JennyPng:
Since there's majority approval on the design for Scenario 1, I've gathered the remaining open questions in an issue to discuss for Scenario 2 https://github.com/Azure/azure-sdk-tools/issues/12632 @maririos @benbp and I'll go ahead and merge this PR today/tomorrow?

praveenkuttappan (approved):

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,581 @@
...
+
+- Determine how to dynamically get requirements, instead of relying on a hard-coded config
+- Explore invoking verifySetup when another tool fails due to missing dependencies
+
+## Meeting Notes
+
+### 10/9 Discussion - Key Takeaways
+
+1. **Source of Truth**: Agreement that there should be one source-of-truth config (from eng/common or each repo) for all requirements
+   - Consider devcontainers as potential source
```
richardpark-msft, line 560: Caveat - you can't totally do this because devcontainers are Linux, and we'll see a lot of Windows developers. However, it's a start.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,581 @@
...
+---
+
+## Design Proposal
+
+### Overview
+
+The `verify-setup` tool checks whether required dependencies for SDK development are installed in the environment. It operates in two modes:
+
+1. **Automatic**: Runs when MCP server starts, checking requirements for the detected language
+2. **Explicit**: Developer or agent explicitly invokes the tool with specific language(s)
```
richardpark-msft: I don't think you should do this **Automatic** "don't start the MCP" bit - basically it says you can't use any actions in here unless you have _all_ tools installed, even for MCP tools you're not using. 

I would cut this this check, in favor of just having the "verify installed tools" MCP tool instead, and on-demand/lazy checks. At least then the failures are relevant to the thing you're specifically trying to do, rather than an odd error message that comes out of nowhere.

Also, I'd wager that interactively dealing with chat is nicer than having to attempt to restart the MCP over and over again.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,581 @@
...
+  ✅ Java JDK 11+
+  ❌ Maven - MISSING
+
+---
+
+To fix missing requirements:
+📦 azpysdk:
+   1. Navigate to your azure-sdk-for-python repo
+   2. Ensure your virtual environment is activated
+   3. Run: python -m pip install -e eng/tools/azure-sdk-tools[build]
```
richardpark-msft, line 272: In your experiments has Copilot been able to follow all these instructions?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,581 @@
...
+│                  │                               │
+│  ┌───────────────▼───────────────────────────┐  │
+│  │ 2. Load Requirements Registry            │  │
+│  │    - Core requirements                    │  │
+│  │    - Language-specific requirements       │  │
+│  └───────────────┬───────────────────────────┘  │
+│                  │                               │
+│  ┌───────────────▼───────────────────────────┐  │
+│  │ 3. For Each Requirement:                 │  │
+│  │    - Run check command                    │  │
```
richardpark-msft, line 327: Parallelized? Might as well take advantage of it to do this as quickly as possible.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
```diff
@@ -0,0 +1,581 @@
...
+
+### Detailed Design
+
+#### Requirements Registry Structure
+
+The requirements are organized into categories based on language, in addition to a "core" section for non-language specific requirements:
+
+```json
+{
+    "categories": {
```
richardpark-msft, line 169: Why are we doing this as a JSON file, and not just code? Code is _way_ more flexible and could be just as readable as JSON.

richardpark-msft (changes requested):

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: V1 is just Copilot informing the user of the instructions, not doing any installation for the user, but yes it's been able to return the correct instructions

there are some screenshots from testing results on the prototype PR https://github.com/Azure/azure-sdk-tools/pull/12476

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: yes it will be parallelized

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: Yep I'll change the wording to make it clearer - we decided in a resolved discussion with Ben on just updating copilot-instructions to suggest running this tool before others, not to actually call it on server startup

timovv (approved):
I think this is fine as a first cut, we can iterate and refine further in the next milestone.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
benbp: Agree on this

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
benbp: Also forgive me if this was discussed, but how long does tool validation take? Like if I run a new MCP server session the day after, do I still have to wait X seconds for verify to run? Am I always going to get prompted by the MCP client to ensure that I have the tools installed? I'm not sure what the best balance is here UX wise.

I wonder if it makes sense for the MCP server to run some non-blocking checks in the background after starting and then caching the results. So on startup we check to see if the user is in a package dir, run tool checks, then cache the results. Then anytime the user makes a tool call we just check the cached result and it feels instant to them. If the user happens to call `verify` while this is happening, we just block on the cache read until the background work is complete.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
benbp: Maybe for v2, I think we could still start running the tool in the background to cache results, see https://github.com/Azure/azure-sdk-tools/pull/12457#discussion_r2456697386

benbp:
LGTM. I have some comments in here that are mostly for follow-on improvements. My main issue with the current spec is the file based config, but I won't block on it as it's easy to refactor if needed.

benbp (approved):

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
richardpark-msft: @JennyPng, you've answered my question fully, so we're good. @benbp has extended this comment chain, and now owns its resolution. ;)

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
richardpark-msft: Oh, I fully think you can return the instructions :) 

I meant, how well does Copilot follow them? Ie, is it able to figure out that it needs to try to install Python, etc... and then attempts to do it? Or is it just a response, and Copilot has no clue?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
richardpark-msft: Offline talk with @JennyPng - this is fine (from my perspective) for V1, but let's not keep diving into config files when we are all programming in the same language and can just write code to do it.

Can we add this to the next revision of the spec, or file an issue, so we don't forget?

richardpark-msft (approved):
We're okay, but I'd like to see some issues filed for any "future refactor" work mentioned in here, so we don't forget it.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: If I'm understanding correctly, yes Copilot understands the returned instructions and formats it in a human-readable way - for V1 it won't attempt to actually do any action, it just displays the instructions
image

maririos (approved):

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
haolingdong-msft: `1. **MCP Server Start**: Automatically check requirements for detected language when server starts`
Seems current agreement is not to invoke the tool automatically on MCP server start? Do we want to remove above sentence?

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
haolingdong-msft: To confirm, in V1, will the agent install the enviornment for you? Or just output the instructions and let user to install manually?

haolingdong-msft (approved):
Thanks @JennyPng for the detailed spec! Mostly LGTM. Approve to unlock merge. Just leave some comments for clarification and understanding purpose.

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: It only outputs instructions, it does not install for you

lmazuel (approved):

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: Resolving open convos so I can merge, but I've moved all these open questions to this issue https://github.com/Azure/azure-sdk-tools/issues/12632

tools/azsdk-cli/specs/1-env-setup-env-verification.spec.md
JennyPng: Resolving open convos so I can merge, but I've moved all these open questions to this issue https://github.com/Azure/azure-sdk-tools/issues/12632

PR #12458, opened 2025-10-10
Add V1 Scenario Spec, Environment Setup Spec, and Spec Template

ronniegeraghty:
- Introduced the V1 Scenario specification detailing the end-to-end workflow for generating and releasing preview SDKs across .NET, Java, JavaScript, Python, and Go using the Health Deidentification service.
- Created the Environment Setup verification spec outlining the proactive verification of required tools and dependencies for SDK development.
- Added a spec template to standardize the creation of future design specifications, including sections for definitions, background, goals, design proposals, alternatives considered, and open questions.

### DIFF
--- a/tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
+++ b/tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
@@ -126,7 +126,8 @@ Without a concrete end-to-end scenario, we risk building tools in isolation that
 - **Release management** - Outer loop activity beyond Scenario 1 scope
 - **Publishing packages to package managers** - Outer loop activity beyond Scenario 1 scope
 
-_Note: Scenario 1 focuses on local development workflow up to the point where SDKs are ready for PR creation. The outer loop process is out of scope. This is a test scenario - do not commit changes or create release PRs._
+_Note: Scenario 1 focuses on local development workflow up to the point where SDKs are ready for PR creation. The outer loop process is out of scope. This is a test scenario - do not commit changes or create releases._
+_If you create a PR to compare the generated library with the existing library and validate PR CI, it must be created as a draft PR, DO NOT merge it, and add `[DO NOT REVIEW OR MERGE]` to the title._
 
 ---
 


tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
maririos, line 1: curious about the file name. Do we need the `0-`? what is the purpose of the first number?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
+# Spec: V1 Scenario
+
+## Overview
+
+The V1 scenario defines the end-to-end workflow for generating and releasing preview SDKs across all five languages (.NET, Java, JavaScript, Python, Go) using the Health Deidentification service as the test case.
```
maririos: did we get confirmation from the management plane team that this spec will work for them too?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+
+**Service**: [Health Deidentification](https://github.com/Azure/azure-rest-api-specs/tree/ded7abde9c48ba84df36b53dfcaef48a2c134097/specification/healthdataaiservices/HealthDataAIServices.DeidServices)  
+**Mode**: Works in both agent and CLI modes  
+**Goal**: Prove complete workflow from setup → generate → validate → release
```
maririos:
```suggestion
**Goal**: Prove complete SDK local workflow from setup → generate → validate until release
```

tools/azsdk-cli/docs/specs/README.md
```diff
@@ -0,0 +1,151 @@
...
+- **Stage 3 (Customizing)**: Covers customizing TypeSpec or SDK libraries
+- **Stage 6 (Package)**: Covers package metadata and documentation updates
+- **Stage 99 (Operations)**: For architecture-level or cross-cutting concerns between inner and outer loop
+
+### 2. Open a Pull Request
+
+1. Create a new branch for your spec
+2. Add your completed spec file to this `docs/specs/` directory
+3. Open a PR with a clear title like: `[Spec] Environment Setup Verification Design`
+4. Tag relevant stakeholders and language representatives as reviewers:
```
maririos, line 53: [codeowners](https://github.com/Azure/azure-sdk-tools/blob/main/.github/CODEOWNERS#L62) is automatically adding everyone on the `azsdk-cli` group. Do we still need an explicit list?

tools/azsdk-cli/docs/specs/README.md
```diff
@@ -0,0 +1,151 @@
...
+   - **.NET**: @dotnet-driver
+   - **Java**: @java-driver
+   - **JavaScript**: @js-driver
+   - **Python**: @python-driver
+   - **Go**: @go-driver
+
+### 3. Gather Feedback and Obtain Sign-Off
+
+- Team members will use the PR review process to comment on and discuss the spec
+- Each SDK language team should provide sign-off via **PR approval**
```
maririos:
```suggestion
- Each SDK language team should provide sign-off via **PR approval** (management plane and data plane)
```

tools/azsdk-cli/docs/specs/README.md
```diff
@@ -0,0 +1,151 @@
...
+   - How will this work across different SDK languages?
+
+6. **Alternatives Considered**
+   - What other approaches were evaluated?
+   - Why was this design chosen over alternatives?
+
+7. **Open Questions**
+   - Unresolved items that need discussion
+   - Areas where input is specifically needed
+
```
maririos, line 103: if the spec proposal doesn't work or apply for a language, there should be a specific section in this document with that information (who and why)

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
jsquire: I just left a comment asking for the clarification, but Mariana beat me to it!

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
+# Spec: V1 Scenario
```
jsquire:
```suggestion
# Spec: V1 scenario
```

nit:  If we're looking to conform to the Microsoft documentation standards, the headers should be sentence case.    I'm good either way, but figured it was worth mentioning if we're going to start building up an official specs area,

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+
+**Service**: [Health Deidentification](https://github.com/Azure/azure-rest-api-specs/tree/ded7abde9c48ba84df36b53dfcaef48a2c134097/specification/healthdataaiservices/HealthDataAIServices.DeidServices)  
+**Mode**: Works in both agent and CLI modes  
+**Goal**: Prove complete workflow from setup → generate → validate → release
+
+---
+
+## Why V1 Scenario Matters
+
+Without a concrete scenario, we risk building tools in isolation that don't integrate well. V1 provides:
```
jsquire:
```suggestion
Without a concrete end-to-end scenario, we risk building tools in isolation that don't integrate well. V1 provides:
```

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+
+- Clear definition of "done" for all teams
```
jsquire:
```suggestion
- Clear definition of successful completion for all teams
```

nit:  I can be "done" and fail miserably.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+- Scope boundaries to prevent feature creep
+- Foundation for cross-language consistency
+
+---
+
+## Context & Assumptions
+
+### Environment
+
+- Windows machine with freshly cloned repos (specs + all 5 language repos)
```
jsquire, line 77: This will include a copy of Health Deidentification already, since its an existing library.    Generation will update or fail rather than satisfy the scenario.

We'll need to explicitly remove it as part of the setup.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+- Existing SDKs already present in language repos
+- Test resources already provisioned
+- Test recordings already exist (playback mode only)
+- TypeSpec modifications are local only (single repo)
+
+### In Scope for V1
+
+- **All five languages** (.NET, Java, JavaScript, Python, Go) - no exceptions, all must pass
+- **Preview release** (not first preview, no architect review required)
```
jsquire: You know this library has already been released, right?    We don't actually want to release based on this exercise.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+- **TypeSpec-based generation** from Health Deidentification service - creating non-compatible version that ignores existing code customizations
+- **With or without client.tsp** - handles both scenarios
```
jsquire: This doesn't test the agent.   It's a core generator feature test.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+- **Playback testing** using existing test recordings
```
jsquire: Again, this tests the language infrastructure.   We only need to validate that the generator ran successfully.   Ensuring the generator works correctly is out of scope.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+- **Both data plane and management plane** APIs
```
jsquire: I assume this means there's a Health Deidentification RM TSP out there?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+
+```text
+1. Environment Setup → verify-setup
+   └─ Check all requirements for all 5 languages
+
+2. Generating → generate-sdk
+   └─ Generate SDK code, tests, samples from TypeSpec
+   └─ Validate: build, test (playback mode), samples
+
+3. Update Package/Docs/Metadata → update-package
```
jsquire, line 144: I assume this is existing functionality, not something new?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+   └─ Run all PR CI checks locally to ensure green PR
+
+5. Releasing (Outer Loop) → create-release-plan, create-prs, validate-release, release, close-plan
+   └─ Create release plan → create all PRs → validate readiness → release all → confirm & close
+   └─ All 5 languages released successfully
+```
+
+---
+
+## Stage Details
```
jsquire, line 156: This is particularly helpful!

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,148 @@
...
+**Tool**: `verify-setup` ([#12287](https://github.com/Azure/azure-sdk-tools/issues/12287))  
+**Action**: Check all requirements upfront for all languages  
+**Success**: All tools/SDKs installed, user knows what's missing
+
+### 2. Generating
+
+**Tool**: `generate-sdk` ([#11403](https://github.com/Azure/azure-sdk-tools/issues/11403))  
+**Action**: Generate SDK code, tests, samples from TypeSpec for Health Deidentification  
+**Success**: Clean generation for all 5 languages  
+**Validation**: `build-sdk`, `run-tests` (playback mode), `validate-samples` - ensure generated code compiles, tests pass, and samples are valid
```
jsquire: How do the tests get recorded?    How do the test resources get created?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: To get a specific ordering of files in the dir. 0 = scenario files, other number are to their according stage in the dev inner loop. I believe this was a request from @lmazuel

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: I believe all we need for MGMT plane for this scenario is to confirm the generate and build tools work for mgmt plane libraries. But no, we didn't get confirmation from them

tools/azsdk-cli/docs/specs/README.md
ronniegeraghty: codeowners will only require one or two people from that list approve for the merge to go through and I think we wanted it stated that we are using approvals to get sign off from each language, and their approvals are required.

tools/azsdk-cli/docs/specs/README.md
ronniegeraghty: This would go in the `Goals and Exceptions/Limitations` section as one of the exceptions.

ronniegeraghty:
> it would be better to remove the `1-env-setup....` spec so this review is merged as the template and guidance, and then a new PR can focus on the design itself for `1-env-setup...`

Removed

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: I believe our scenario is looking at updating the libraries (with a breaking version).
I was under the assumption we didn't want any initial library setup steps (work to create a new library in the SDK repos) to be part of this scenario. So we'd be updating an existing. 
@maririos, was that your assumption as well?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: I think we want to be able to run these in the scenario but for our testing we wouldn't run the release steps. 
When we finish the work on this scenario, if all the tools are released on the mcp server, and service team should be able to do all the activities. 

added a note on line 55 related to this.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: I believe what we are testing (or validating) with these is that the agent is able to properly invoke local generation for all languages for both data and mgmt planes.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Yes just added a link to the MGMT Plane spec.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Yes, this stage covers existing steps in our release process. The work to do for this stage is in creating MCP tools to invoke the different automations we have for doing these activities per language.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Since we are starting with a set of existing libraries, it's assumed that recorded test have already been created and we are only running tests against recordings, no live tests. 
This is called out in the `Out of scope for V1` section: `Live test execution` & `Test resource management`

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
jsquire: I think there's a disconnect there.    If we generate into an existing library that has already been released, then the flow won't match "this is an initial beta."      We're using a variation of that phrase in a couple of different spots.    I also do not want us committing any changes to this area in .NET - we don't own this library.

When I suggested it, my understanding is that what we did would be off to the side, NOT integrated into the actual live repo.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
jsquire: Exactly - `handles both scenarios` is out of scope, then.    That's a generator responsibility, not whatever invokes the generator.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
lmazuel: I think I like that the spec will always be the first file in the folder, but I can be convinceed otherwise if you all hate it

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
lmazuel: +1 with @ronniegeraghty 's assumption, V1 shouldn't be about new setup, it's a hard problem

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
lmazuel: I don't think V1 is about "this is an initial beta", but an updatet. @ronniegeraghty, we are not explicit about that I believe in the doc, that the library already exists

@jsquire , we don't want to merge anything, this is an exercise.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
lmazuel: I agree that which files are run is a tsp-client problem, and the agent shouldn't care. I think the real point is that editing client.tsp is not part of the flow of V1

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
lmazuel: We want to check that the agent can run the test with a prompt like "Can you run my  tests" or something simillar

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,152 @@
...
+2. Create all PRs (one per language)
+3. Validate release readiness
+4. Execute release for all languages
+5. Confirm all languages released and close release plan
+
+**Success**: All 5 languages released successfully, release plan closed
+
+---
+
+## Success Criteria
```
lmazuel, line 186: I wonder if we should have example of prompts that we want to ssee working consistencly after V1 is done. Like:
- Is my machine ready
- Generate the code
- Please run my tests

As a success criteria, I should be able to start myself from scratch, and run those commands on my machine, and it should work just fine.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,152 @@
...
+
+V1 is complete when:
+- ✅ Complete workflow executes successfully for all 5 languages
+- ✅ Works in both agent and CLI modes
```
lmazuel: To my point before, I'd like more than "works" with the explicit set of CLI commands and English prompt that is supposed to cover the entire V1 flow.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
maririos: @haolingdong-msft will this service work for management plane? (does it even have a management plane spec?) and if not, could you provide one that will work for management plane?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
maririos: Correct: editing is not part of V1.
The clarification here is that we want to work on an sdk that has minimal to non-client code customization. Whether there are customizations on client.tsp shouldn't affect our V1. => with or without client.tsp.
Maybe this could be worded differently? `SDK should not have previous client code customizations nor it should create new client customizations`

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
maririos: ohh I love this. didn't we have a doc that included prompts per scenario? we should add them here (the ones that apply for V1)

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: [Link to MGMT Plane Spec](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/healthdataaiservices/HealthDataAIServices.Management)

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Yeah maybe this can be better stated in the doc
Currently I have it as `**Preview release** (not first preview, no architect review required)`

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: How does having 
```
- Working with or creating new SDK code customizations
- Creating new client.tsp customizations
```
in the out of scope section look?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,461 @@
...
+**Service**: Health Deidentification
+- [Health Deidentification Data Plane Spec](https://github.com/Azure/azure-rest-api-specs/tree/ded7abde9c48ba84df36b53dfcaef48a2c134097/specification/healthdataaiservices/HealthDataAIServices.DeidServices)  
+- [Health Deidentification MGMT Plane Spec](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/healthdataaiservices/HealthDataAIServices.Management)
+
+**Modes**: Works in both [agent mode](#agent-mode) and [CLI mode](#cli-mode)  
+**Goal**: Prove complete SDK local workflow from setup → generate → validate until release
+
+---
+
+## Definitions
```
jsquire, line 32: Beautiful.   Love this!

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,461 @@
...
+
+V1 is complete when:
+
+- ✅ Complete workflow executes successfully for all 5 languages
+- ✅ Works in both [agent mode](#agent-mode) and [CLI mode](#cli-mode)
+- ✅ Documentation exists for both modes
+- ✅ Repeatable with consistent results
+- ✅ Agent prompts produce expected behavior
+- ✅ CLI commands execute with expected options and outputs
+
```
jsquire, line 197: Can we add an item or note here about  "Workflow was completed locally but NOT committed to the repository?"

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,461 @@
...
+**Options:**
+- `--languages <list>`: Comma-separated list of languages to check (default: all)
+- `--verbose`: Show detailed output for each check
+- `--fix`: Attempt to automatically install missing dependencies
+
+**Expected Output:**
+```
+Checking environment setup for 5 languages...
+
+✓ .NET SDK 8.0.100 installed
```
jsquire:
```suggestion
✓ .NET SDK 9.0.205 installed
```

Just to be pedantic.   The 9.x SDK for .NET should be installed.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,461 @@
...
+   └─ Validate: versions, READMEs, changelogs
+
+4. Validating → run-pr-checks
+   └─ Run all PR CI checks locally to ensure green PR
+
+5. Releasing (Outer Loop) → create-release-plan, create-prs, validate-release, release, close-plan
+   └─ Create release plan → create all PRs → validate readiness → release all → confirm & close
+   └─ All 5 languages released successfully
+```
+
```
jsquire, line 153: 6.  STOP.   Don't actually commit this or release it.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,461 @@
...
+- **[Playback testing](#playback-mode)** using existing test recordings
+- **Both [data plane](#data-plane) and [management plane](#management-plane)** APIs
+- **[Agent](#agent-mode) and [CLI modes](#cli-mode)** - both must work
+- **Existing changelog generation** - no changes to current process
+
+### Out of scope for V1
+
+- Working with or creating new [SDK code customizations](#code-customizations)
+- Creating new [TypeSpec customizations](#typespec-customizations) ([client.tsp](#client-tsp))
+- Breaking changes
```
jsquire: Committing this to the repository or releasing it.   Seriously.   Don't do this or Mariana will come to your house and give you really mean looks.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Added

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Added

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: added

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+- [Success Criteria](#success-criteria)
+- [Agent Prompts](#agent-prompts)
+- [CLI Commands](#cli-commands)
+- [Open Questions](#open-questions)
+
+---
+
+## Overview
+
+The V1 scenario defines the end-to-end workflow for generating and validating [preview SDKs](#preview-release) across all five languages (.NET, Java, JavaScript, Python, Go) using the Health Deidentification service as the test case.
```
richardpark-msft:
```suggestion
The V1 scenario defines the end-to-end workflow for generating and validating [preview SDKs](#preview-release) across all 5 languages (.NET, Java, JavaScript, Python, Go) using the Health Deidentification service as the test case.
```

You're using the "all 5 languages" consistently, but if it's not spelled the same then it's not easily searchable.

So either "all 5 languages" or "all five languages", but not both.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+- **Existing changelog generation** - no changes to current process
+
+### Out of scope for V1
+
+- Working with or creating new [SDK code customizations](#code-customizations)
+- Creating new [TypeSpec customizations](#typespec-customizations) ([client.tsp](#client-tsp))
+- Breaking changes
+- Live test execution
+- Test resource management
+- Linux/macOS support
```
richardpark-msft: I know this is just V1, but it's useful to know _why_ we decided some things were out of scope.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+- First preview version
+- GA version
+- Committing changes to repositories
+- Creating or merging pull requests
+- Release management
+- Publishing packages to package managers
+
+_Note: V1 focuses on local development workflow up to the point where SDKs are ready for PR creation. The outer loop release process is out of scope. This is a test scenario - do not commit changes or create release PRs._
```
richardpark-msft: Is 'release process' the outer loop or is there more? (if so, do we even need to say 'outer loop'? It's confusing enough with us saying 'inner loop').

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+
+⚠️  STOP: This is a test scenario only. Do NOT commit these changes or create release PRs.
+```
+
+---
+
+## Stage Details
+
+### 1. Environment Setup
+**Tool**: `verify-setup` ([#12287](https://github.com/Azure/azure-sdk-tools/issues/12287))
```
richardpark-msft, line 160:
```suggestion

**Tool**: `verify-setup` ([#12287](https://github.com/Azure/azure-sdk-tools/issues/12287))  
```

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+
+- Windows machine with freshly cloned repos (specs + all 5 language repos)
+- Existing SDKs already present in language repos
+- Test resources already provisioned
+- Test recordings already exist ([playback mode](#playback-mode) only)
+- [TypeSpec](#typespec) modifications are local only (single repo)
+
+### In Scope for V1
+
+- **All five languages** (.NET, Java, JavaScript, Python, Go) - no exceptions, all must pass
```
richardpark-msft: 'pass' in this case is just "must generate without error", correct? 

Or are we going to have some small tests to use them as well?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+
+### Environment
+
+- Windows machine with freshly cloned repos (specs + all 5 language repos)
+- Existing SDKs already present in language repos
+- Test resources already provisioned
+- Test recordings already exist ([playback mode](#playback-mode) only)
+- [TypeSpec](#typespec) modifications are local only (single repo)
+
+### In Scope for V1
```
richardpark-msft: Some questions to clarify the scope:
 
- Is making this work in Visual Studio (not VSCode) in scope?
- Is using arbitrary models in scope or are we saying "we're testing with Claude Sonnet 4"

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+- **All five languages** (.NET, Java, JavaScript, Python, Go) - no exceptions, all must pass
+- **[Preview release](#preview-release)** (not first preview, no architect review required)
+- **[TypeSpec](#typespec)-based generation** from Health Deidentification service - creating non-compatible version that ignores existing [code customizations](#code-customizations)
+- **With or without [client.tsp](#client-tsp)** - handles both scenarios
+- **[Playback testing](#playback-mode)** using existing test recordings
+- **Both [data plane](#data-plane) and [management plane](#management-plane)** APIs
+- **[Agent](#agent-mode) and [CLI modes](#cli-mode)** - both must work
+- **Existing changelog generation** - no changes to current process
+
+### Out of scope for V1
```
richardpark-msft: Some suggestions for "currently out of scope but soon-ish in-scope"

- Visual Studio support (really, this is just us ensuring we test it)
- GitHub's `copilot` CLI, and using the Copilot Coding Agent
- Multiple models (I'm _guessing_ we're going to say Claude Sonnet 4 only, based on ES recommendations).

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+
+## Definitions
+
+_Terms used throughout this spec with precise meanings:_
+
+- **<a id="agent-mode"></a>Agent Mode**: Interactive mode where the user provides natural language prompts to the AI agent (GitHub Copilot), and the agent executes the appropriate tools and commands automatically.
+
+- **<a id="cli-mode"></a>CLI Mode**: Direct command-line interface mode where the user manually executes specific `azsdk` commands with explicit options and parameters.
+
+- **<a id="preview-release"></a>Preview Release**: A pre-GA (beta) release of an SDK package (not the first preview). Does not require architect review. Indicates the API is not yet stable and may have breaking changes in future releases.
```
richardpark-msft: I think this goes a bit too far with redefining a term. I don't think I could find anyone that assumes "preview release" only refers to subsequent previews, and not the first one. 

I'm curious if you could just inline part of this explanation in the few spots that use it. It's really more of a constraint for individual pieces of this doc, than part of the term.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+**Modes**: Works in both [agent mode](#agent-mode) and [CLI mode](#cli-mode)  
+**Goal**: Prove complete SDK local workflow from setup → generate → validate
+
+---
+
+## Definitions
+
+_Terms used throughout this spec with precise meanings:_
+
+- **<a id="agent-mode"></a>Agent Mode**: Interactive mode where the user provides natural language prompts to the AI agent (GitHub Copilot), and the agent executes the appropriate tools and commands automatically.
```
richardpark-msft, line 36: Does this count me using Copilot coding agent, or the `copilot` CLI? 

(If so, we should just straight up say "in VSCode only")

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
haolingdong-msft: Hi @maririos and @ronniegeraghty, I've already replied the same question you asked in the loop. Paste [here](https://microsoft.sharepoint.com/:fl:/g/contentstorage/CSP_ae9817b7-57b5-49d4-9f3a-e08717b934f8/Ed40N6qt1iFDu24Aw3_CwecBtBK8_dbyLSrBDDvr0h2iGA?e=880sz8&nav=cz0lMkZjb250ZW50c3RvcmFnZSUyRkNTUF9hZTk4MTdiNy01N2I1LTQ5ZDQtOWYzYS1lMDg3MTdiOTM0ZjgmZD1iJTIxdHhlWXJyVlgxRW1mT3VDSEY3azAtSTdjVzdINkZSRlBuS0VaS096ZWQwdXZURWhxU1hVS1JacTVoYmhfZ29xNCZmPTAxMkM2QkNVRzZHUTMyVkxPV0VGQjNXM1FBWU43NEZRUEgmYz0lMkYlM0ZtaW5pZmllZCUzRDZlNzc4ZGQzLTk4ZTktNDAzNy04NGEwLTljOWMwMjE2NzY5YyUyNnNlcSUzRDg2OTMmYT1Mb29wQXBwJnA9JTQwZmx1aWR4JTJGbG9vcC1wYWdlLWNvbnRhaW5lciZ4PSU3QiUyMnclMjIlM0ElMjJUMFJUVUh4dGFXTnliM052Wm5RdWMyaGhjbVZ3YjJsdWRDNWpiMjE4WWlGMGVHVlpjbkpXV0RGRmJXWlBkVU5JUmpkck1DMUpOMk5YTjBnMlJsSkdVRzVMUlZwTFQzcGxaREIxZGxSRmFIRlRXRlZMVWxweE5XaGlhRjluYjNFMGZEQXhNa00yUWtOVlIwcFJTVFpOUjBsR1JVUTFRVmxOVnpSQlNVSldVRWt5TWpjJTNEJTIyJTJDJTIyaSUyMiUzQSUyMmE3NGMyM2U4LThhYTAtNDYyMy05MjAzLTg2MzE4MDM3NjIzOSUyMiU3RA%3D%3D) in case it didn't come through your end. Thanks.

From me:
> I have checked with mgmt language owners, dotnet may have some concern with the selected service. Other languages are good with this service. Please refer to @haiyuazhang's comment above for more details. Thanks.

From @haiyuazhang:
> The chosen service still relies on AutoRest.Csharp for generating the .NET SDK. I’m considering whether we should pick a service from [here](https://github.com/Azure/azure-sdk-for-net/blob/main/doc/GeneratorMigration/Library_Inventory.md#management-plane-libraries-using-typespec-azure-typespechttp-client-csharp-mgmt) to test if the underlying tooling works well with the new management plane generator

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+- ✅ Works in both [agent mode](#agent-mode) and [CLI mode](#cli-mode)
+- ✅ Documentation exists for both modes
+- ✅ Repeatable with consistent results
+- ✅ Agent prompts produce expected behavior
+- ✅ CLI commands execute with expected options and outputs
+- ✅ Workflow runs completely locally *Nothing is committed to the repo*
+
+---
+
+## Agent Prompts
```
haolingdong-msft, line 200: Thanks @ronniegeraghty ! I love this detailed prompts.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
lirenhe: As we are using a service with both data plane and mgmt. plane, is data plane for GO also in the scope?  @richardpark-msft

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
haolingdong-msft: Please also kindly pay attention to  @lirenhe’s comment in above loop on how we select the service.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+
+- **<a id="preview-release"></a>Preview Release**: A pre-GA (beta) release of an SDK package (not the first preview). Does not require architect review. Indicates the API is not yet stable and may have breaking changes in future releases.
+
+- **<a id="playback-mode"></a>Playback Mode**: Test execution mode that uses pre-recorded HTTP interactions instead of making live calls to Azure services. Enables fast, reliable testing without requiring live Azure resources.
+
+- **<a id="typespec"></a>TypeSpec**: The specification language used to define Azure service APIs. SDK code is generated from TypeSpec specifications located in the azure-rest-api-specs repository.
+
+- **<a id="typespec-customizations"></a>TypeSpec Customizations**: Modifications made to TypeSpec files (such as [client.tsp](#client-tsp)) to customize the SDK generation process. These customizations affect what code is generated but do not modify the generated code itself. Examples include renaming operations, changing parameter types, or customizing client interfaces.
```
haolingdong-msft: May I know does V1 include update tsp files? If yes, what is the scope? Is it changes to client.tsp only? Or include changes to other source tsp files as well? 
(Add @ArthurMa1978, @lirenhe and @chunyu3 as it seems related with typespec authoring.)

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+2. Execute `generate-sdk` for Health Deidentification service
+3. Execute `update-package` to update versions, changelogs, READMEs
+4. Execute `run-pr-checks` locally to validate all checks pass
+5. Report status and next steps
+
+### Environment Setup
+
+**Prompt:**
+```
+Check if my environment is ready to generate SDKs for all languages.
```
haolingdong-msft, line 224: Personally I feel usually user will not prompt to check environment. More nature way to me is: user prompts ‘generate Java Sdk for xxx service’. Then if environment issue appears, the agent will call the tool to set up environment. What do you think?

tools/azsdk-cli/docs/specs/README.md
```diff
@@ -0,0 +1,169 @@
+# Dev Inner Loop Design Specs
+
+This directory contains design specifications for the Azure SDK Dev Inner Loop tools and processes. All major design proposals, feature additions, and process changes should be documented and reviewed here.
```
haolingdong-msft, line 3: I really like the idea of putting design docs together here. Because currently seems we have .doc, .loop,  .md design files and each doc has different structure/format. So in the future our design should use .md and  follow the  structure you described below?

tools/azsdk-cli/docs/specs/README.md
```diff
@@ -0,0 +1,169 @@
...
+   - Example: `1-env-setup-env-verification.spec.md`
+3. Fill out all sections of the template (see [What to Include](#what-to-include) below)
+
+#### Dev Inner Loop Stages
+
+Use these stage prefixes when naming your spec files:
+
+| Stage # | Stage Name | Prefix | Example Spec Name |
+|---------|------------|--------|-------------------|
+| 1 | Environment Setup | `1-env-setup` | `1-env-setup-env-verification.spec.md` |
```
chrisradek, line 32: Should we add `0` for scenarios too?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+
+## Workflow
+
+```text
+1. Environment Setup → verify-setup
+   └─ Check all requirements for all 5 languages
+
+2. Generating → generate-sdk
+   └─ Generate SDK code, tests, samples from TypeSpec
+   └─ Validate: build, test (playback mode), samples
```
raych1, line 142: How do we trigger the validation after the generation? I recommend that we control the workflow by the copilot instructions.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+
+⚠️  STOP: This is a test scenario only. Do NOT commit these changes or create release PRs.
+```
+
+---
+
+## Stage Details
+
+### 1. Environment Setup
+**Tool**: `verify-setup` ([#12287](https://github.com/Azure/azure-sdk-tools/issues/12287))
```
raych1, line 160: Does it include cloning SDK language repos?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+
+Environment check complete: 4/5 languages ready
+Run with --fix to install missing dependencies
+```
+
+### 2. Generate SDK
+
+**Command:**
+```bash
+azsdk generate-sdk --service healthdataaiservices --spec-path <path-to-specs> --languages .NET,Java,JavaScript,Python,Go
```
raych1, line 315: I'm curious about the origin of this design. It differs significantly from my design spec, which has already been implemented and delivered, including the naming and interfaces. I'm concerned that making this the V1 goal given the timeline. 

@ronniegeraghty I'd recommend trying out the current experience for generating first, then we can revisit the goal based on it.  CC: @maririos @lmazuel

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,406 @@
...
+
+**Prompt:**
+```
+Update the package metadata for Health Deidentification SDKs to prepare for a preview release.
+```
+
+**Expected Agent Activity:**
+1. Execute `update-package` for all 5 languages
+2. Update version numbers for [preview release](#preview-release)
+3. Update changelogs with recent changes
```
raych1: Changelog update is out of v1 scope for data plane SDK since they are manually updated as current practices.

tools/azsdk-cli/docs/specs/README.md
m-redding: I don't think anyone from .NET is included as a codeowner so not sure if someone should be added or if the explicit list is good enough

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: The outer loop is more than just the release process. But we also want to say that pushing to the repos is out of scope too which is also outer loop. So maybe just saying outer loop stuff is out of scope is simpler.

ronniegeraghty: changed the wording. let me know what you think?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Updated the section to include reasoning for why things are out of scope. A few of them are just to limit initial complexity.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Really what I'm trying to say here is that support for completing these activities is available across the languages. If the inputs given would normally cause the generation to fail, then it should also fail in our case, and we should report any error message. I think in a later milestone we'll look at handling failures to help the user fix them, but that's out of scope for now.

ronniegeraghty: Updated the language here and also added a specific call out in the out-of-scope section talking about tooling to resolve errors/failures. Let me know what you think?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: These are all good questions that I don't believe we considered when scoping the scenario. 
I believe model shouldn't affect our work because the exact model being used shouldn't change the way we implement a tool. We may tune verbiage for better interaction with different models, but I think we should wait on this till we have our Evals set up, which will give us an easier way to test if changes are causing better or worse responses from certain models. @lmazuel, @maririos, @jsquire, Do all this sound right to you? 

We didn't scope the editor or AI chat tool originally. I believe our tooling only works with VSCode currently so other editors would be out of scope. @maririos & @praveenkuttappan can you confirm this? 

I'll add to the scenario spec that only VSCode is in scope and other editors are out of scope.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Good call outs. Like I said above, I believe testing multiple modes will come when we have our AI evals set up. 

But yest trying different editors and agents that support MCP will be coming in a future scenario.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Agreed. Simplified it and called out Frist or subsequent preview release inline.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Updated in scope and out of scope sections to say the only env in scope is VSCode with GitHub Copilot.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Yes, it should be.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Any type of TypeSpec authoring is out of scope. 
I added a part to specifically call this out in the out of scope section. 
Here's the text:
```
- **Creating new [TypeSpec customizations](#typespec-customizations) ([client.tsp](#client-tsp))** - Limits scope; we're not focusing on helping authors create code customizations or client.tsp customizations at this stage
```

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: > Please also kindly pay attention to @lirenhe’s comment in above loop on how we select the service.

I responded in the loop, but it would be great if we could bring that conversation here to keep all conversations on the scenario in one place.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: I believe your probably right, I think we just wanted a way to test the individual tools using both agent prompts and CLI commands.

tools/azsdk-cli/docs/specs/README.md
ronniegeraghty: Yes, that's the plan. We'll use the PR to have async discussions on the spec and if any larger topics arise, we'll discuss them in our Thursday evening (your Friday morning) recurring meeting.

tools/azsdk-cli/docs/specs/README.md
ronniegeraghty: Good catch. You can tell I worked on this before creating my scenario spec. 
Just added the updates.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: I believe this should be discussed in the `### Detailed Design: [Detailed explanation with diagrams, code samples, or workflows as appropriate]` section for the spec on the validation items. I would consider it part of their workflow details on what cause an AI to believe it should use the tooling. 

But yes, copilot instructions are probably the simplest answer right now. But eventually we won't be able to rely on copilot instructions once we start looking at supporting scenarios outside our own repos.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: No. In the context assumptions we have it stated that we are assuming the repos (spec and all 5 language repos) have been freshly cloned in the env. Here is the text: 
```
## Context & Assumptions

### Environment

- Windows machine with freshly cloned repos (specs + all 5 language repos)
```

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Yes, I do need to revisit these commands. These were created as I was working with Copilot to create the spec and it added these as place holders. I'll work with the leads for each stage to get the right command syntax.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Good call out. I updated this to say mgmt plane only and added ChangeLog updates for data plane to the out of scope section.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
+# Spec: Scenario 1
```
ronniegeraghty, line 1: I've updated the naming of the scenario from `V1 Scenario` to `Scenario 1` to address confusion on whether or not this scenario is tied to a specific version release of azsdk. I'll updated the name of the file before merging but I don't want to lose any comments on it until they are all resolved.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+---
+
+## Overview
+
+Scenario 1 defines the end-to-end workflow for generating and validating [preview SDKs](#preview-release) across all 5 languages (.NET, Java, JavaScript, Python, Go) using the Health Deidentification service as the test case.
+
+**Service**: Health Deidentification
+
+- [Health Deidentification Data Plane Spec](https://github.com/Azure/azure-rest-api-specs/tree/ded7abde9c48ba84df36b53dfcaef48a2c134097/specification/healthdataaiservices/HealthDataAIServices.DeidServices)  
+- [Health Deidentification MGMT Plane Spec](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/healthdataaiservices/HealthDataAIServices.Management)
```
ronniegeraghty, line 25: We are currently discussing if this is the right service to use for MGMT Plane since the generation of the .NET MGMT plane library for this service still requires `AutoRest.Csharp`. 

CC: @lirenhe & @haiyuazhang: Carrying over the conversation from the loop. 

@lirenhe & @haiyuazhang, will the current MGMT Plane generation tooling fail when trying to generate the .NET library for Health Deidentificaiton? Or will it just generate a library that is considered breaking from the existing one?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Carried the conversation on the loop to this PR in a comment below.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
richardpark-msft: >  but I think we should wait on this till we have our Evals set up, which will give us an easier way to test if changes are causing better or worse responses from certain models

Actually, this is why I'm suggesting picking a _single_ model, and saying that's what we're testing with (ie, reduce the scope of Scenario 1).

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
richardpark-msft: Really good!

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
richardpark-msft: LGTM.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
haiyuazhang: @ronniegeraghty  It's fine to select the health data service for .NET management plane validation. We're beginning to migrate from AutoRest.Csharp to the new management generator, and the migration will likely be completed before the validation starts. If not, we can still proceed using AutoRest.Csharp. as it still works.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
richardpark-msft: @lirenhe, yeah, I'm in for the full ride :) I haven't even looked at the service, but I'm okay if it fails - we'd have to fix that stuff anyways, and this is the right time for it.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
maririos: >We didn't scope the editor or AI chat tool originally. I believe our tooling only works with VSCode currently so other editors would be out of scope. @maririos & @praveenkuttappan can you confirm this?

It works for Java too as @samvaity has been working with Ben on making sure the MCP works on their IDE.
It would be good to do a test for .NET and see what fails there. I wouldn't consider this part of V1 though as part of that work is dependent on VS and not on us

maririos: I am indifferent at this time to pick a specific model. According to the task/step the model might change and I prefer having that flexibility to pick the one that performs the best. But this would be a better question from the experience you all get by executing each step.
I am not 100% convinced we have enough information at this point for that to be decided

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
maririos: Correct. For the specific flow you are mentioning @haolingdong-msft please follow up on the design spec that Jenny has in https://github.com/Azure/azure-sdk-tools/pull/12457

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
maririos: we picked this service because it looked simple. If we need to select a different one for management plane so we are not blocked on .NET, we can do that

maririos: @jsquire do you have an opinion on the generation the .NET management plane team should use for this effort?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+
+- **<a id="playback-mode"></a>Playback Mode**: Test execution mode that uses pre-recorded HTTP interactions instead of making live calls to Azure services. Enables fast, reliable testing without requiring live Azure resources.
+
+- **<a id="typespec"></a>TypeSpec**: The specification language used to define Azure service APIs. SDK code is generated from TypeSpec specifications located in the azure-rest-api-specs repository.
+
+- **<a id="typespec-customizations"></a>TypeSpec Customizations**: Modifications made to TypeSpec files (such as [client.tsp](#client-tsp)) to customize the SDK generation process. These customizations affect what code is generated but do not modify the generated code itself. Examples include renaming operations, changing parameter types, or customizing client interfaces.
+
+- **<a id="client-tsp"></a>client.tsp**: An optional TypeSpec customization file that allows SDK authors to customize the generated client interface without modifying the service specification. This is a specific type of [TypeSpec customization](#typespec-customizations).
```
samvaity: I feel like this can just be a part of typespec customizations, do we need to segregate this?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+- **<a id="release-plan"></a>Release Plan**: A coordinated release workflow that tracks the release of SDK packages across multiple languages, ensures all validations pass, and manages the PR and release process end-to-end.
+
+---
+
+## Why Scenario 1 Matters
+
+Without a concrete end-to-end scenario, we risk building tools in isolation that don't integrate well. Scenario 1 provides:
+
+- Clear definition of successful completion for all teams
+- Repeatable test case for validation
```
samvaity:
```suggestion
- Repeatable validation framework to ensure reliability
```

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+- Scope boundaries to prevent feature creep
```
samvaity:
```suggestion
- Defined scope boundaries to prevent feature creep
```

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+
+- **<a id="release-plan"></a>Release Plan**: A coordinated release workflow that tracks the release of SDK packages across multiple languages, ensures all validations pass, and manages the PR and release process end-to-end.
+
+---
+
+## Why Scenario 1 Matters
+
+Without a concrete end-to-end scenario, we risk building tools in isolation that don't integrate well. Scenario 1 provides:
+
+- Clear definition of successful completion for all teams
```
samvaity: Clear success benchmark for all teams

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+
+## Context & Assumptions
+
+### Environment
+
+- Windows machine with freshly cloned repos (specs + all 5 language repos)
+- Existing SDKs already present in language repos
+- Test resources already provisioned
+- Test recordings already exist ([playback mode](#playback-mode) only)
+- [TypeSpec](#typespec) modifications are local only (single repo)
```
samvaity, line 81: Is this common enough workflow to be in consideration for V1 scenario?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+
+### In Scope for Scenario 1
+
+- **All 5 languages** (.NET, Java, JavaScript, Python, Go) - All activities defined in this scenario must be supported across all languages listed. Each activity should behave consistently with existing tooling—successfully completing or erroring out as expected given the same inputs.
+- **Subsequent [preview releases](#preview-release)** - Releases after the first preview that do not require architect review and where initial library setup has already been completed
+- **[TypeSpec](#typespec)-based generation** from Health Deidentification service - creating non-compatible version that ignores existing [code customizations](#code-customizations)
```
samvaity: what is the meaning of `non-compatible version` exactly?

samvaity: If the scope is to ignore code customizations, I think the expectation remains to have code that builds and runs, even if it differs from the initial beta version.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+- **With or without [client.tsp](#client-tsp)** - handles both scenarios
```
samvaity: We should clarify on this and chose one path to keep behavior/results consistent?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+- **Other editors (IntelliJ, Visual Studio, etc.)** - All editors besides VS Code are out of scope; VS Code is the only supported development environment for scenario 1.
+
+**Outer Loop Activities:**
+
+- **Committing changes to repositories** - Scenario 1 focuses on local development experience (inner loop), not outer loop integration
+- **Creating or merging pull requests** - Scenario 1 focuses on inner loop; PR creation engages with outer loop
+- **Release management** - Outer loop activity beyond Scenario 1 scope
+- **Publishing packages to package managers** - Outer loop activity beyond Scenario 1 scope
+
+_Note: Scenario 1 focuses on local development workflow up to the point where SDKs are ready for PR creation. The outer loop process is out of scope. This is a test scenario - do not commit changes or create release PRs._
```
samvaity: I agree that for inner-loop efforts, we shouldn't create release PRs. However, is there a better way to validate changes? Maybe we could consider creating PRs that aren't merged, but still provide visibility into the changes and allow us to benefit from pipeline runs.
One for each language, that way we can spot the operation gaps better?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+
+---
+
+## Stage Details
+
+### 1. Environment Setup
+
+**Tool**: `verify-setup` ([#12287](https://github.com/Azure/azure-sdk-tools/issues/12287))  
+**Action**: Check all requirements upfront for all languages  
+**Success**: All tools/SDKs installed, user knows what's missing
```
samvaity, line 162: Just confirming, this tooling intent is to guide the user for missing and not install correct?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
lmazuel: I think the point is that there isn't any client.tsp update, so everything go through tsp-client, which abstract the concept of input file (maybe it's client.tsp. maybe it's main.tsp, it's not important)

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
lmazuel: Creating PR we don't merge is fine to me

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
lmazuel: That's my understanding, @LibbaLawrence and @JennyPng to confirm

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
lmazuel: I think the point being made is that the changes in TypeSpec are not big enough to require manual code modification in the target language

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
JennyPng: yes

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+... (similar for other languages)
+
+Generation complete: 5/5 languages successful
+```
+
+### 3. Update Package Metadata
+
+**Command:**
+```bash
+azsdk update-package --service healthdataaiservices --version 1.1.0-beta.1 --languages .NET,Java,JavaScript,Python,Go
```
raych1, line 352: Same here for the gaps of commands and interfaces. I will include them in the spec of its own tool.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
```diff
@@ -0,0 +1,425 @@
...
+**Expected Agent Activity:**
+1. Execute `verify-setup` for .NET, Java, JavaScript, Python, Go
+2. Report which requirements are met and which are missing
+3. Provide guidance on how to install missing dependencies
+
+### Generate SDK
+
+**Prompt:**
+```
+Generate the Health Deidentification SDK for all languages from the TypeSpec in azure-rest-api-specs.
```
raych1, line 236: What is the user context? spec repo or sdk repo? Can we call it out explicitly?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: In our docs we have a section that says we currently test with Claude Sonnet 4 and 4.1, and GPT-4 and 5. 
image

So.... should we limit testing to those to practice what our docs preach?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Noting that these will change as the specs for these tools come in.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Sounds like this is okay for now and we can pivot if it becomes an issue for mgmt plane later.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
raych1: Should we just put pseudocode as example for now and later update it to reference each individual tool’s spec when it's available?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Great point, combined it into one and updated the spec to use the `typespec customization` term.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Took this suggestion.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Are you asking when compared to generating locally but using the typespec in the remote API Specs Repo? 

I'm not sure if this is the most common workflow but I believe the idea was this would reduce complexity as we try to manage what can fit into a single sprint. 

We also eventually have to look into so scenarios that happen outside our own repos, and pushing anything not local to a later scenario helps us not have to worry about the other repo scenarios. (for now)

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: @samvaity is right that this refers to the fact we are ignoring code customizations. This point doesn't mention that we'd expect the code not to build or run.

I added a definition for non-compatible version: 
```
Non-Compatible Version**: A new version of an SDK library that contains breaking changes, rendering it incompatible with previous versions. In this scenario, non-compatibility occurs because existing code customizations are ignored during regeneration, resulting in a modified API surface that breaks backward compatibility.
```

Do you think we need to explicitly call it out somewhere that the produced libraries should be buildable? I believed that to be inferred since building the libraries is part of validating the generation stage. I would also say the goal isn't to always produce buildable & run-able libraries in this scenario. The goal it is to have the tooling work as expected, so if the generator when used normally to created libraries would produce a non-buildable library due to issues in the typespec or the generator, we should be behaving in the same way.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Wouldn't we need to create separate spaces in the SDK repos for use to create PRs to merge our codes into? 
Because if we create PRs to merge the code into the dir where the actual library code lives then it'll be the service team who is assigned as the reviewers for that PR. So we'd have to tell them they should be ignoring these. 
Or we could create them as draft PRs? If they are draft PRs does only the PR author have the ability to take the PR from draft to open? Or could the service team accidentally do this then approve and merge it?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Yes

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Yes, I added an update that calls these out as samples and they will be replaced when the spec for that specific tool is created.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Added the following to the Env context assumption to call this out: 
`All [agent mode](#agent-mode) interactions are performed in VS Code with GitHub Copilot with the Azure/azure-rest-api-specs repository is open in VSCode.`

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: Added to the in scope section

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
samvaity: The definition addition helps resolve this.
Validating/testing should infer buildable code so no need to explicitly mention.

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
samvaity: I think we can do draft prs with title to include [DO NOT REVIEW] ?
create separate spaces in the SDK repos -> is an interesting question as this is an existing SDK it can take over the existing SDK space ?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
samvaity: Sorry but @JennyPng in the setup spec the `Action` is to do lazy loading and not check all requirements upfront, can you confirm if the action here is indeed reflecting it right wrt ongoing spec PR?

tools/azsdk-cli/docs/specs/0-v1-scenario.spec.md
ronniegeraghty: I extended the note to say the following 
> _Note: Scenario 1 focuses on local development workflow up to the point where SDKs are ready for PR creation. The outer loop process is out of scope. This is a test scenario - do not commit changes or create releases._
_If you create a PR to compare the generated library with the existing library and validate PR CI, it must be created as a draft PR, DO NOT merge it, and add `[DO NOT REVIEW OR MERGE]` to the title._

This adds instructions for those who want to create PRs for testing.

samvaity (approved):

Issue #10598, opened 2025-05-12
Java duplicate artifactIds: create-apireview.yml and Create-APIReview.ps1


JimSuplizio:
Similar to [validate-all-packages.yml](https://github.com/Azure/azure-sdk-tools/issues/10229) issue, create-apireview.yml used a list Artifacts' Names passed in to get the packageInfo files which will no longer work for Java. The solution is to allow either a list of Artifacts or PackageInfo files.

PR #12467, opened 2025-10-13
Updated create-apireview template to use packageInfo

raych1:
The change is to support Java v2 library which needs to get `groupId` and `artifactId` from `packageInfo` to differenciate same `artifactId`. 
Refer to https://github.com/Azure/azure-sdk-tools/issues/10598 for details.

* Added a new `PackageInfoFiles` parameter to both the pipeline template and the PowerShell script

* Refactored parameter handling in `Create-APIReview.ps1` to prioritize `PackageName`, then `Artifacts`, then `PackageInfoFiles`, ensuring backward compatibility with previous usages while supporting the new method.

* Refactored the `ProcessPackage` function to work with package info objects

### DIFF
--- a/eng/common/scripts/Create-APIReview.ps1
+++ b/eng/common/scripts/Create-APIReview.ps1
@@ -193,7 +193,7 @@ function ProcessPackage($packageInfo)
         # Call function with appropriate parameters
         if ($supportsPackageInfoParam) {
             LogInfo "Calling $FindArtifactForApiReviewFn with packageInfo parameter"
-            $packages = &$FindArtifactForApiReviewFn $ArtifactPath $pkgArtifactName $packageInfo
+            $packages = &$FindArtifactForApiReviewFn $ArtifactPath $packageInfo
         }
         else {
             LogInfo "Calling $FindArtifactForApiReviewFn with legacy parameters"
--- a/eng/common/scripts/Detect-Api-Changes.ps1
+++ b/eng/common/scripts/Detect-Api-Changes.ps1
@@ -132,13 +132,13 @@ foreach ($packageInfoFile in $packageInfoFiles)
     # Check if the function supports the packageInfo parameter
     $functionInfo = Get-Command $FindArtifactForApiReviewFn -ErrorAction SilentlyContinue
     $supportsPackageInfoParam = $false
-    
+
     if ($functionInfo -and $functionInfo.Parameters) {
         # Check if function specifically supports packageInfo parameter
         $parameterNames = $functionInfo.Parameters.Keys
         $supportsPackageInfoParam = $parameterNames -contains 'packageInfo'
     }
-    
+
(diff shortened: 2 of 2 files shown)


eng/common/scripts/Create-APIReview.ps1
```diff
@@ -296,42 +309,84 @@ function ProcessPackage($packageName)
...
-# if package name param is not empty then process only that package
-if ($PackageName)
-{
-    Write-Host "Processing $($PackageName)"
-    $result = ProcessPackage -packageName $PackageName
-    $responses[$PackageName] = $result 
+# Initialize working variable
+$ProcessedPackageInfoFiles = @()
+
+if ($PackageName -and $PackageName -ne "") {
```
weshaggard, line 330: I assume these branches can be removed once we convert all the usages of to use PackageInfo's?

eng/common/scripts/Create-APIReview.ps1
```diff
@@ -296,42 +309,84 @@ function ProcessPackage($packageName)
     return 0
 }
 
-$responses = @{}
-# Check if package config file is present. This file has package version, SDK type etc info.
-if (-not $ConfigFileDir)
-{
+Write-Host "Artifact path: $($ArtifactPath)"
+Write-Host "Source branch: $($SourceBranch)"
+Write-Host "Package Info Files: $($PackageInfoFiles)"
```
weshaggard, line 314: Will this output as you expect given it is an object array?

eng/common/scripts/Create-APIReview.ps1
```diff
@@ -172,12 +173,32 @@ function IsApiviewStatusCheckRequired($packageInfo)
...
+        $supportsPackageInfoParam = $false
+        
+        if ($functionInfo -and $functionInfo.Parameters) {
+            # Check if function specifically supports packageInfo parameter
+            $parameterNames = $functionInfo.Parameters.Keys
+            $supportsPackageInfoParam = $parameterNames -contains 'packageInfo'
+        }
+        
+        # Call function with appropriate parameters
+        if ($supportsPackageInfoParam) {
```
weshaggard, line 194: Another case where I would like to clean this up later once we get everyone updated to use packageInfo.

eng/common/scripts/Create-APIReview.ps1
raych1: Yes, it works well.
image

eng/common/scripts/Create-APIReview.ps1
raych1: That's correct.

eng/common/scripts/Create-APIReview.ps1
raych1: That's correct.

Issue #12331, opened 2025-10-03
[JavaScript] Lisst Requirements V1


l0lawrence:
Provide Jenny with the requirements needed for your language generation steps for V1, this includes what is needed to run any of your generation/build code, and CI pr checks.

samvaity:
### Java SDK Environment Setup Requirements
Requirements Checklist

#### **1. Node.js & npm**

*   [ ] Node.js **≥ 20.0.0** installed
    ```bash
    node --version && npm --version
    ```
*   [ ] npm included with Node.js installation

***

#### **2. Java Development Kit (JDK)**

*   [ ] JDK **≥ 17** installed (Microsoft OpenJDK recommended)
    ```bash
    java -version
    ```
*   [ ] `JAVA_HOME` set to JDK installation path
*   [ ] `%JAVA_HOME%\bin` added to `PATH`

***

#### **3. Apache Maven**

*   [ ] Latest Apache Maven installed
    ```bash
    mvn --version
    ```
*   [ ] `MAVEN_HOME` set to Maven installation path
*   [ ] `%MAVEN_HOME%\bin` added to `PATH`

***

#### **4. TypeSpec Tools**

*   [ ] Install/update TypeSpec tools:
    ```bash
    npm install -g @azure-tools/typespec-client-generator-cli@latest @typespec/compiler@latest
    ```
*   [ ] Verify versions:
    ```bash
    tsp-client --version && tsp --version
    ```
    *   `tsp-client` ≥ 0.24.0
    *   `tsp` ≥ 1.0.0

***

#### **5. Combined Environment Check**

Run this command to confirm all tools:

```bash
node --version && npm --version && tsp-client --version && tsp --version && java -version && mvn --version
```

### **Additional Notes**

*   Always install **latest stable versions**.
*   Configure environment variables for persistence (`JAVA_HOME`, `MAVEN_HOME`, `PATH`).
*   Restart VS Code after setup for environment refresh.
*   Administrator privileges may be required for installations.

PR #12476, opened 2025-10-13
Add Verify Setup Tool

JennyPng:
Design spec: https://github.com/Azure/azure-sdk-tools/pull/12457

Adding an MCP tool that checks the environment for necessary requirements to run MCP tools.

To run in CLI for all languages: `dotnet run --project Azure.Sdk.Tools.Cli -- verify setup --all`

## manual testing results
These results require updating copilot-instructions in respective repos with the instructions linking to `verify-setup.instructions.md`

### azure-sdk-for-go
image

### azure-sdk-for-python
example of agent asking for venv path
image

example of follow-up checking after installing missing reqs
image

### azure-sdk-for-java
image

### azure-sdk-for-javascript
image

example of asking for multiple langs
image

### azure-sdk-for-dotnet
image

### azure-rest-api-specs

image

### example of checking all languages within azure-sdk-for-java

image

### example of catching outdated version
checking python when there's a venv with python 3.8
image

## focus areas for review
- location of and accuracy of RequirementsV1.json config 

## next steps after merging this
- respective language owners should check that the tool works expected (will make issues)
- update copilot-instructions across language repos to link to this tool's specific instructions, as well as add instructions to suggest running this tool before any other tools run
- Address remaining open design questions for the next iteration: https://github.com/Azure/azure-sdk-tools/issues/12632


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,121 @@
...
+            },
+            {
+                "requirement": "Node.js >= 22.0.0",
+                "check": ["npm", "--version"],
+                "instructions": ["Download Node.js"]
+            },
+            {
+                "requirement": "tsp-client >= 0.24.0",
+                "check": ["tsp-client", "--version"],
+                "instructions": ["npm install -g @azure-tools/typespec-client-generator-cli@latest @typespec/compiler@latest"]
```
raych1: `tsp-client` should be installed from the shared location. Refer to https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/tsp-client/README.md

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,121 @@
+{
+    "categories": {
```
raych1, line 2: @weidongxu-microsoft @haolingdong-msft @msyyc @tadelesh @MaryGao 
can you please review the list and call out if there're any tools missing which are needed for mgmt-plane SDK?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,121 @@
...
+            },
+            {
+                "requirement": "github CLI >= 2.30.0",
+                "check": ["gh", "--version"],
+                "instructions": ["Download and install GitHub CLI"]
+            }
+        ],
+        "java": [
+            {
+                "requirement": "java >= 17",
```
haolingdong-msft: I believe java 11 would be enough. Add @weidongxu-microsoft for confirmation

weidongxu-microsoft: Yeah, 11 is good for emitter.

BTW, should we do `javac --version` (upon Java compiler)? To make sure user has JDK installed (not only a JRE that does not have compiler).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,121 @@
...
+            {
+                "requirement": "azcore",
+                "check": ["go", "list", "-m", "github.com/Azure/azure-sdk-for-go/sdk/azcore"],
+                "instructions": ["go get -u github.com/Azure/azure-sdk-for-go/sdk/azcore toolchain@none go@1.23.0"]
+            },
+            {
+                "requirement": "azidentity",
+                "check": ["go", "list", "-m", "github.com/Azure/azure-sdk-for-go/sdk/azidentity"],
+                "instructions": ["go get -u github.com/Azure/azure-sdk-for-go/sdk/azidentity toolchain@none go@1.23.0"]
+            }
```
tadelesh: These two libraries are not tools.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
JennyPng: @samvaity specified version 17 here https://github.com/Azure/azure-sdk-tools/issues/12331 so I'll leave it at the higher version?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    private static readonly List<string> LANGUAGES = new() { "python", "java", "dotnet", "javascript", "go" };
+    private const int COMMAND_TIMEOUT_IN_SECONDS = 30;
+
+    public override CommandGroup[] CommandHierarchy { get; set; } = [
+        SharedCommandGroups.Verify,
+    ];
+
+    private readonly Option<string> languagesParam = new("--langs", "-l")
```
benbp: I think it's a bit clearer to use `--languages`. Since this already has the `-l` alias we don't need to abbreviate the long form.

Also you can use the enum for the option to get built-in validation and not have to hardcode the help text languages (which would need to be updated if we added a language).

```suggestion
    private readonly Option<SdkLanguage> languagesParam = new("--languages", "-l")
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/PythonRequirementsCheck.cs
```diff
@@ -0,0 +1,92 @@
...
+    private string? FindOrCreateVenv(string packagePath)
+    {
+        try
+        {
+            if (string.IsNullOrWhiteSpace(packagePath))
+            {
+                packagePath = Environment.CurrentDirectory;
+            }
+
+            var candidates = new[] { ".venv", "venv", ".env", "env" };
```
l0lawrence: should we allow the user to give us a venv they are using as an input to our check requirements for python, or tweak this check a bit? they could have a venv named something else `python -m venv potato` for example and that wouldn't get caught here b/c `potato/bin/activate`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,333 @@
...
+/// </summary>
+[McpServerToolType, Description("This tool verifies that the environment is set up with the required installations to run MCP release tools.")]
+public class VerifySetupTool : MCPTool
+{
+    private readonly IProcessHelper processHelper;
+    private readonly ILogger<VerifySetupTool> logger;
+
+    private readonly ILanguageSpecificResolver<IEnvRequirementsCheck> envRequirementsCheck;
+
+    private readonly string PATH_TO_REQS = Path.Combine(AppContext.BaseDirectory, "Configuration", "RequirementsV1.json");
```
benbp: If we go with the config file approach you will have to embed it into the assembly so that this works when we run as a single file executable.

I have an example of this that you can steal from:

Embed resource in csproj: https://github.com/Azure/azure-sdk-tools/blob/9339f65c02eac516aeabefe9a65a8d14069b7418/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Azure.Sdk.Tools.Cli.csproj#L58-L60

Reference assembly resource in tool: https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Example/Quokka.cs#L29-L33

```
            var assembly = Assembly.GetExecutingAssembly();
            var names = assembly.GetManifestResourceNames();
            using var stream = assembly.GetManifestResourceStream("Azure.Sdk.Tools.Cli.Images.quokka.ansi");
            using var reader = new StreamReader(stream);
            return reader.ReadToEnd();
```

eng/common/instructions/azsdk-tools/verify-setup.instructions.md
```diff
@@ -0,0 +1,6 @@
+---
+description: 'Verify Setup'
+---
+Your goal is to identify the project repo root, and pass in the packagePath to the Verify Setup tool. For a language repo, pass in the language via the `langs` parameter (supported parameters are: python,java,dotnet,javascript,go). For a non-language repo, do not pass in any languages to just check the core requirements. Summarize the output of the tool to inform if the user's development environment is correctly set up for SDK development and release tasks.
```
benbp: Is this for MCP mode or CLI or both? The lang parameter naming differs between them right now. Also I think the supported parameters should only be sourced from the MCP tool help text (via the `Description` attribute argument) or the CLI help text, so we don't have to hardcode it here.

Providing an example CLI command and/or MCP tool name to call is probably helpful for the LLM as well.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,333 @@
...
+                AllRequirementsSatisfied = true,
+                Results = new List<RequirementCheckResult>()
+            };
+
+            foreach (var req in reqsToCheck)
+            {
+                logger.LogInformation("Checking requirement: {Requirement}, Check: {Check}, Instructions: {Instructions}",
+                    req.requirement, req.check, req.instructions);
+
+                var result = await RunCheck(req, packagePath, ct);
```
benbp: I think it's probably better to run these checks concurrently via something like:

```
var result = RunCheck(req, packagePath, ct);
results.Append(result);

await Task.WhenAll(results);
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
haolingdong-msft: Add @haiyuazhang to comment on dotnet.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,111 @@
...
+        "python": [
+            {
+                "requirement": "python >= 3.9",
+                "check": ["python", "--version"],
+                "instructions": ["Install Python from https://www.python.org/downloads/"]
+            },
+            {
+                "requirement": "azpysdk",
+                "check": ["azpysdk", "--help"],
+                "instructions": ["Ensure your virtual environment is activated", "python -m pip install eng/tools/azure-sdk-tools[build]", "Check that it was installed with `azpysdk --help`"]
```
msyyc, line 107:
```suggestion
                "instructions": ["Ensure your virtual environment is activated", "python -m pip install eng/tools/azure-sdk-tools[build,ghtools,sdkgenerator]", "Check that it was installed with `azpysdk --help`"]
```

Python need additional dependencies for sdk generation.

eng/common/instructions/azsdk-tools/verify-setup.instructions.md
JennyPng: It's intended for MCP to enforce copilot passing in the correct params - I'll remove the hard-coded params and add an example

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/PythonRequirementsCheck.cs
JennyPng: (discussed offline but adding here for visibility) I'll look into overloading the MCP tool function to take in a venv path as a param

eng/common/instructions/azsdk-tools/verify-setup.instructions.md
benbp: If the input param for the method uses the enum type we have for languages instead of string, then I believe the schema returned to the MCP client for `tools/list` will show what the values are, so it shouldn't need to be documented.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,342 @@
...
+    {
+        var langs = parseResult.GetValue(languagesParam);
+        var allLangs = parseResult.GetValue(allLangOption);
+        var parsed = allLangs ? LANGUAGES : ParseLanguages(langs);
+        var packagePath = parseResult.GetValue(SharedOptions.PackagePath);
+        return await VerifySetup(parsed, packagePath, ct);
+    }
+
+    [McpServerTool(Name = "azsdk_verify_setup"), Description("Verifies the developer environment for MCP release tool requirements. Accepts a list of supported languages to check requirements for (java, python, dotnet, javascript, go), and the packagePath of the repo to check.")]
+    public async Task<VerifySetupResponse> VerifySetup(List<string> langs = null, string packagePath = null, CancellationToken ct = default)
```
benbp: Related to my above comment. Realized we actually shouldn't need to note the languages in the description if we use the enum:

```suggestion
    public async Task<VerifySetupResponse> VerifySetup(List<SdkLanguage> languages = null, string packagePath = null, CancellationToken ct = default)
```

JennyPng: I updated this to use SdkLanguage instead of a string - I am leaving this arg name as `langs` for now because so far copilot has only generated input with `langs` and never `languages` even if I update the instructions....

benbp: :D that's interesting! Is it possibly due to the instructions still referencing langs?

> For example, in `azure-sdk-for-js`, run `VerifySetup` with `(langs=javascript, packagePath=<path>/azure-sdk-for-js)`

eng/common/instructions/azsdk-tools/verify-setup.instructions.md
```diff
@@ -0,0 +1,10 @@
+---
+description: 'Verify Setup'
+---
+This tool verifies the developer's environment for SDK development and release tasks. It returns what requirements are missing for the specified languages and repo.
+
+Your goal is to identify the project repo root, and pass in the packagePath to the Verify Setup tool. For a language repo, pass in the language. For a non-language repo, do not pass in any languages to just check the core requirements. Summarize the output of the tool.
+
+For example, in `azure-sdk-for-js`, run `VerifySetup` with `(langs=javascript, packagePath=<path>/azure-sdk-for-js)`.
```
praveenkuttappan: Tool name is `azsdk_verify_setup`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,126 @@
+{
+    "categories": {
+        "core": [
+            {   
+                "requirement": "Node.js >= 22.16.0",
+                "check": ["node", "--version"],
+                "instructions": ["Download Node.js"]
+            },
+            {
+                "requirement": "npm",
```
praveenkuttappan: is this check required when checking node? npm is by default installed with latest node. right?

l0lawrence: @timovv

timovv: Yeah Node should include npm so this probably isn't necessary

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,126 @@
...
+                "check": ["npm", "--version"],
+                "instructions": ["Download Node.js"]
+            },
+            {
+                "requirement": "tsp-client >= 0.24.0",
+                "check": ["tsp-client", "--version"],
+                "instructions": ["cd eng/common/tsp-client", "`npm ci`"]
```
timovv, line 12: Some of these instructions are commands to run and some of them are plain English. Should we add something like 'run this command' to the instructions for the ones that are commands?

```suggestion
                "instructions": ["Run these commands", "cd eng/common/tsp-client", "`npm ci`"]
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,319 @@
...
+            }
+
+            var results = await Task.WhenAll(checkTasks);
+
+            foreach (var (req, result) in results)
+            {
+                if (result.ExitCode != 0)
+                {
+                    logger.LogWarning("Requirement check failed for {Requirement}. Suggested install command: {Instruction}", req.requirement, req.instructions);
+                    response.AllRequirementsSatisfied = false;
```
praveenkuttappan: Can you also please set ``ResponseError` here so this tool execution will be marked as failure in telemetry which will be helpful to understand setup issues in customer env?
ResponseError = "Setup requirement check failed"

or you can add each `ResponseError` from checks to `RepsonseErrors` in parent response object.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,319 @@
...
+        // Per-language requirements
+        var reqGetter = null as IEnvRequirementsCheck;
+        if (languages == null || languages.Count == 0)
+        {
+            // Detect language if none given
+            reqGetter = await envRequirementsCheck.Resolve(packagePath);
+
+            if (reqGetter == null)
+            {
+                logger.LogWarning("Could not resolve requirements checker for the specified languages. Checking only core requirements. Please provide languages using --languages option to check language requirements.");
```
praveenkuttappan: nit: language list is empty in this code block so warning should indicate packagePath

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,319 @@
...
+        {
+            // Unable to parse the version from the output
+            return requiredVersion;
+        }
+
+        string installedVersion = outputVersionMatch.Value;
+
+        logger.LogInformation($"Installed version: {installedVersion}");
+
+        int comparison = CompareVersions(installedVersion, requiredVersion);
```
praveenkuttappan: I think we should make it easier by comparing just major and minor version and avoid comparing patch version?
@l0lawrence @benbp @richardpark-msft

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/EnvRequirementsCheck.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Sdk.Tools.Cli.Models;
+
+public abstract class EnvRequirementsCheck
+{
+    private readonly string PATH_TO_REQS = Path.Combine(AppContext.BaseDirectory, "Configuration", "RequirementsV1.json");
+
+    public async Task<List<SetupRequirements.Requirement>> ParseRequirements(string checkCategory, CancellationToken ct = default)
```
l0lawrence: nit - could this just be a helper function? I'm not sure if that is .NET-ish or not

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
JennyPng: even when changing that part to `languages` it still uses `langs` for some reason 😓

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
JennyPng: Copilot seems to be able to interpret and format it reasonably
image

but I can add it for more clarity?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/EnvRequirementsCheck.cs
JennyPng: It probably could, I wasn't sure where the best place to put it would be

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
benbp: I would assume if patch doesn't matter then `requiredVersion` in the config can just be `>= M.m` ?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,126 @@
+{
+    "categories": {
+        "core": [
+            {   
+                "requirement": "Node.js >= 22.16.0",
+                "check": ["node", "--version"],
+                "instructions": ["Download Node.js"]
```
benbp, line 7: npm gets installed via the node install. Is the LLM is smart enough to figure this out or will it try to install both via the same process?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,126 @@
...
+        "javascript": [
+            {
+                "requirement": "pnpm",
+                "check": ["pnpm", "--version"],
+                "instructions": ["npm install -g pnpm", "Run pnpm install in the repo root."]
+            }
+        ],
+        "dotnet": [
+            {
+                "requirement": "dotnet",
```
benbp: I think we are going to hit some interesting edge cases. Like some other tools in this config, dotnet has multiple ways it can be installed (`apt`, install script into `~/.dotnet`, etc.) and it doesn't work nicely if multiple methods get used together (e.g. the dotnet binary resolves to `/usr/bin/dotnet` which references the apt installations, but the LLM decided to use the install script and the new SDK is located in `~/.dotnet`.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/VerifySetupResponse.cs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using System.Text.Json.Serialization;
+using System.Text;
+namespace Azure.Sdk.Tools.Cli.Models;
+
+public class VerifySetupResponse : CommandResponse
+{
+    [JsonPropertyName("allRequirementsSatisfied")]
```
benbp:
```suggestion
    [JsonPropertyName("all_requirements_satisfied")]
```

(note to myself - I should add an analyzer to enforce snake_case for json property names)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ILanguageSpecificResolver.cs
```diff
@@ -13,4 +14,13 @@ public interface ILanguageSpecificResolver<T> where T : class
...
\ No newline at end of file
+
+    /// <summary>
+    /// Resolves language-specific service implementations for a list of language identifiers.
+    /// This is used by tools that need to check or operate across multiple languages in one request.
+    /// </summary>
+    /// <param name="languages">A list of language identifiers (e.g. "python", "java").</param>
+    /// <param name="ct">Cancellation token.</param>
+    /// <returns>A list of resolved service instances (null for languages that could not be resolved).</returns>
+    public List<T?> Resolve(List<SdkLanguage> languages, CancellationToken ct = default);
```
benbp: Defensive programming nit, but maybe use `HashSet` instead of `List` so we can't pass duplicate languages?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,326 @@
...
+            ">=" => comparison >= 0 ? string.Empty : requiredVersion,
+            "<=" => comparison <= 0 ? string.Empty : requiredVersion,
+            ">" => comparison > 0 ? string.Empty : requiredVersion,
+            "<" => comparison < 0 ? string.Empty : requiredVersion,
+            "=" => comparison == 0 ? string.Empty : requiredVersion,
+            _ => requiredVersion,
+        };
+    }
+
+    private int CompareVersions(string version1, string version2)
```
m-redding: You could see if this might work for your case - https://learn.microsoft.com/en-us/dotnet/api/system.version.tryparse?view=net-9.0

I used it in the DotNetLanguageSpecificChecks https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs#L155-L159

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,121 @@
+{
+    "categories": {
+        "core": [
```
m-redding, line 3: the verify readme common script uses python and pip - https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Verify-Readme.ps1#L85, it may be worth considering adding it to the core requirements since I think that script will be run for all languages in the validate step

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,121 @@
...
+            }
+        ],
+        "javascript": [
+            {
+                "requirement": "pnpm",
+                "check": ["pnpm", "--version"],
+                "instructions": ["npm install -g pnpm", "Run pnpm install in the repo root."]
+            }
+        ],
+        "dotnet": [
```
m-redding: I think .NET at some point will need an additional more complicated check for https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=windows%2Cnet8#prerequisites

Is that the kind of thing that would go in DotNetRequirementsCheck? I still need to do some research about how to validate that prerequisite but I think it might be more work to validate than the other checks represented here

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
JennyPng: for V1 the LLM only provides instructions and doesn't try to install anything

from my testing so far, the LLM doesn't deviate from the instructions this tool returns - so I have not seen it try to install npm

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
weidongxu-microsoft: Higher version would definitely work. We can have this as 17, till we have feedback from users.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,126 @@
...
+            {
+                "requirement": "github CLI >= 2.30.0",
+                "check": ["gh", "--version"],
+                "instructions": ["Download and install GitHub CLI"]
+            },
+            {
+                "requirement": "python",
+                "check": ["python", "--version"],
+                "instructions": ["Install Python from https://www.python.org/downloads/"]
+            }
```
JennyPng, line 33: note: python appears twice in the config because the [Verify-Readme](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Verify-Readme.ps1#L85) script needs Python to run pip, but we don't expect users to need a venv to check core requirements, and all the Python specific checks attempt to invoke from a specific venv

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
timovv: If it's fine with Copilot I'm fine with it too

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,126 @@
...
+        "javascript": [
+            {
+                "requirement": "pnpm",
+                "check": ["pnpm", "--version"],
+                "instructions": ["npm install -g pnpm", "Run pnpm install in the repo root."]
+            }
+        ],
+        "dotnet": [
+            {
+                "requirement": "dotnet",
```
m-redding:
```suggestion
                "requirement": "dotnet >= 9.0.306",
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
m-redding: In .NET the patch version of the SDK does matter, instead of simplifying by removing patch from the comparison I think it would be better to consider my comment below about using System.Version https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-version that already does all of this

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/EnvRequirementsCheck.cs
```diff
@@ -0,0 +1,43 @@
...
+public abstract class EnvRequirementsCheck
+{
+    public async Task<List<SetupRequirements.Requirement>> ParseRequirements(string checkCategory, CancellationToken ct = default)
+    {
+        var assembly = Assembly.GetExecutingAssembly();
+        var names = assembly.GetManifestResourceNames();
+        using var stream = assembly.GetManifestResourceStream("Azure.Sdk.Tools.Cli.Configuration.RequirementsV1.json");
+        using var reader = new StreamReader(stream);
+        var requirementsJson = await reader.ReadToEndAsync(ct);
+        var setupRequirements = JsonSerializer.Deserialize<SetupRequirements>(requirementsJson);
```
m-redding: You could use DeserializeAsync here

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ILanguageSpecificResolver.cs
```diff
@@ -13,4 +14,13 @@ public interface ILanguageSpecificResolver<T> where T : class
...
\ No newline at end of file
+
+    /// <summary>
+    /// Resolves language-specific service implementations for a list of language identifiers.
+    /// This is used by tools that need to check or operate across multiple languages in one request.
+    /// </summary>
+    /// <param name="languages">A list of language identifiers (e.g. "python", "java").</param>
+    /// <param name="ct">Cancellation token.</param>
+    /// <returns>A list of resolved service instances (null for languages that could not be resolved).</returns>
+    public List<T?> Resolve(HashSet<SdkLanguage> languages, CancellationToken ct = default);
```
m-redding: This feels different enough that it doesn't feel like it should be an overload to me but it's subjective enough I'm ok if you want to  leave it

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/DotNetRequirementsCheck.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Sdk.Tools.Cli.Models;
+
+public class DotNetRequirementsCheck : EnvRequirementsCheck, IEnvRequirementsCheck
+{
+    public async Task<List<SetupRequirements.Requirement>> GetRequirements(string packagePath, CancellationToken ct = default)
+    {
+        return await base.ParseRequirements("dotnet", ct);
```
m-redding: I see that you have it this way because of the language string, but I think it would make more sense to have GetRequirements called with the parsed requirements as a param instead of having it always called first thing. if there's a way to do it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/EnvRequirementsCheck.cs
m-redding:
```suggestion
        var setupRequirements = await JsonSerializer.DeserializeAsync<SetupRequirements>(stream, cancellationToken: ct);
```
You shouldn't need to read it into a string unless I'm missing something

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,338 @@
...
+
+    private async Task<List<SetupRequirements.Requirement>> GetCoreRequirements(CancellationToken ct)
+    {
+        var assembly = Assembly.GetExecutingAssembly();
+        var names = assembly.GetManifestResourceNames();
+        using var stream = assembly.GetManifestResourceStream("Azure.Sdk.Tools.Cli.Configuration.RequirementsV1.json");
+        using var reader = new StreamReader(stream);
+        var requirementsJson = await reader.ReadToEndAsync(ct);
+
+        var setupRequirements = JsonSerializer.Deserialize<SetupRequirements>(requirementsJson);
```
m-redding: I feel like the config file shouldn't need to be deserialized again for every category of requirements. I think the requirements should be parsed once and then passed to the language implementations

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
JennyPng: I did actually run into an edge case related to this where installing node only through `nvm` and not `sudo apt` causes a failure due to how `ProcessHelper` spawns the shell (no `nvm activate` logic or using `.bashrc` I think) 😓 

for milestone 1, do you think these edge cases need to be addressed (with added instructions from the LLM to the user, or adding more robust env var configuring in `ProcessHelper`, etc.)? @benbp

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/PythonRequirementsCheck.cs
```diff
@@ -0,0 +1,103 @@
...
+                var path = Path.Combine(packagePath, c);
+                if (Directory.Exists(path))
+                {
+                    return Path.GetFullPath(path);
+                }
+            }
+
+            // no venv found, create one
+            var venvPath = Path.Combine(packagePath, ".venv");
+
```
JennyPng: @l0lawrence do you think it makes sense to auto create a venv and/or even try to find a venv? it may be strange now that we request the user to pass in a specific path

l0lawrence: Right now since we are just checking what they have already installed and aren't doing any installation Im good with removing the create a venv/find a venv part. One question I do have is do you know how good the llm is at choosing to pass in the venv path? If it doesn't choose to pass through a path and we end up setting it to null, are we then going to check the python in path (if its set) for the dependencies and say something like "we are checking the python set in path", or are you thinking to error and say "no pass in a venv path" and expect the llm to correct itself? 

Image

l0lawrence: oh I did not read above code we are erroring :) cool

JennyPng: From testing, if you're in the Python repo and checking the Python reqs, the LLM pretty reliably will ask you for a venvpath or at least autofill in some venv path if you have 1+ venvs in your repo. 

In other scenarios with other repos and/or checking multiple languages, it's a bit more finicky, but I tried emphasizing in the instructions prompt to always ask for a venv path (although, if the user is in a non python repo, is it even expected that they would want to provide a venv path?)

l0lawrence: they might want to check multiple langs to see if they can do something and that could be in a diff repo than python, we can work on seeing what prompt gives the best result could be something to sync with evaluation people on :)

JennyPng: sounds good, I'll add prompt refining as an open question/area to this issue https://github.com/issues/assigned?issue=Azure%7Cazure-sdk-tools%7C12632 - and I'll remove the auto venv creation

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ILanguageSpecificResolver.cs
samvaity: +1 on this feeling different enough to not being an overload. Thoughts on changing it to `ResolveLanguages`?

samvaity: For the return type should we consider if a dictionary (e.g., Dictionary<SdkLanguage, T?>) to be more expressive?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/JavaRequirementsCheck.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Sdk.Tools.Cli.Models;
+
+public class JavaRequirementsCheck : IEnvRequirementsCheck
+{
+    public List<SetupRequirements.Requirement> GetRequirements(string packagePath, Dictionary<string, List<SetupRequirements.Requirement>> categories, CancellationToken ct = default)
+    {
+        return categories.TryGetValue("java", out var requirements) ? requirements : new List<SetupRequirements.Requirement>();
```
samvaity, line 14: I see for V1 the implementations of `IEnvRequirementsCheck` for each language look same. Do you think they would differ/vary as we proceed in future?
My suggestion would be to start small and use a Map/Dictionary to get the specific language requirements from the List<SetupRequirements>  as the implementation suggests instead ?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/PythonRequirementsCheck.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            // Update checks to use venv path
+            var binDir = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "Scripts" : "bin";
+            req.check[0] = Path.Combine(venvPath, binDir, req.check[0]);
+        }
+
+        return reqs;
+    }
+
+    private string FindOrCreateVenv(string packagePath)
```
samvaity: should these go into helpers?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,314 @@
...
+    {
+        var langs = parseResult.GetValue(languagesParam);
+        var allLangs = parseResult.GetValue(allLangOption);
+        var parsed = allLangs ? Enum.GetValues<SdkLanguage>().ToHashSet() : langs.ToHashSet();
+        var packagePath = parseResult.GetValue(SharedOptions.PackagePath);
+        var venvPath = parseResult.GetValue(venvOption);
+        return await VerifySetup(parsed, packagePath, venvPath, ct);
+    }
+
+    [McpServerTool(Name = "azsdk_verify_setup"), Description("Verifies the developer environment for MCP release tool requirements. Accepts a list of supported languages to check requirements for, and the packagePath of the repo to check. Accepts a specific Python virtual environment path to use for Python requirements checks.")]
```
samvaity, line 86: I understand that the Python environment setup currently requires users to provide an environment path.
- Is there a more standard or recommended approach for configuring environment paths in Python? For example, is there a commonly used location or package path for these setups?
- For V1 milestone, since we are not installing anything, requesting this information may be extending our scope. In that case, would we also need to specify download locations for other components, such as Java? Thinking out loud. 
What do you think @JennyPng @LibbaLawrence

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,314 @@
...
+
+    private const int COMMAND_TIMEOUT_IN_SECONDS = 30;
+
+    public override CommandGroup[] CommandHierarchy { get; set; } = [
+        SharedCommandGroups.Verify,
+    ];
+
+    private readonly Option<List<SdkLanguage>> languagesParam = new("--languages", "-l")
+    {
+        Description = $"List of space-separated languages to check requirements for ({string.Join(", ", Enum.GetNames(typeof(SdkLanguage)))}). Defaults to current repo's language.",
```
samvaity: The list formatting (string.Join(", ", ...)) does not match the description text to provide languages as comma-separated?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,314 @@
...
+    };
+
+
+    private readonly Option<bool> allLangOption = new("--all")
+    {
+        Description = "Check requirements for all supported languages.",
+        DefaultValueFactory = _ => false
+    };
+
+    private readonly Option<string> venvOption = new("--venv-path", "-v")
```
samvaity: -v in most standard form is used for checking version. We should probably consider updating it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,314 @@
...
+
+            var results = await Task.WhenAll(checkTasks);
+
+            foreach (var (req, result) in results)
+            {
+                if (result.ExitCode != 0)
+                {
+                    logger.LogWarning("Requirement check failed for {Requirement}. Suggested install command: {Instruction}", req.requirement, req.instructions);
+
+                    response.AllRequirementsSatisfied = false;
```
samvaity: Do we need the bool AllRequirementsSatisfied, currently don't see it any different from result.exitcode?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,314 @@
...
+        var venvPath = parseResult.GetValue(venvOption);
+        return await VerifySetup(parsed, packagePath, venvPath, ct);
+    }
+
+    [McpServerTool(Name = "azsdk_verify_setup"), Description("Verifies the developer environment for MCP release tool requirements. Accepts a list of supported languages to check requirements for, and the packagePath of the repo to check. Accepts a specific Python virtual environment path to use for Python requirements checks.")]
+    public async Task<VerifySetupResponse> VerifySetup(HashSet<SdkLanguage> langs = null, string packagePath = null, string venvPath = null, CancellationToken ct = default)
+    {
+        try
+        {
+            List<SetupRequirements.Requirement> reqsToCheck = await GetRequirements(langs, packagePath ?? Environment.CurrentDirectory, venvPath, ct);
```
samvaity, line 91: How are we limiting the verification is applied to only the specified languages?

samvaity, line 91: We should also add handling for unsupported language

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ILanguageSpecificResolver.cs
timovv: Maybe also personal style but I think I'd prefer an overload of `Resolve` that just takes a single `SdkLanguage` instead of a many-to-many:

```cs
public Task<T?> Resolve(SdkLanguage language);
```

Then it could be also used for resolving a single language, and doing the thing for a list/set of languages would be as simple as something like `await Task.WhenAll(languages.Select(resolver.Resolve);`.

I would also prefer it return `Task` for consistency with the other method. Or, alternatively, we could rename the existing method to `ResolveAsync` to disambiguate.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/DotNetRequirementsCheck.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Sdk.Tools.Cli.Models;
+
+public class DotNetRequirementsCheck : IEnvRequirementsCheck
+{
+    public List<SetupRequirements.Requirement> GetRequirements(string packagePath, Dictionary<string, List<SetupRequirements.Requirement>> categories, CancellationToken ct = default)
+    {
+        return categories.TryGetValue("dotnet", out var requirements) ? requirements : new List<SetupRequirements.Requirement>();
```
samvaity, line 14: What are your thoughts on parsing the JSON once globally and each language-specific checker could be given the relevant slice of the config?
We could consider injecting the config data so that these classes don’t need to know about JSON at all. 
Citing for performance reasons if this is going to be revisited dont mind the current design as it works.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,314 @@
...
+            VerifySetupResponse response = new VerifySetupResponse
+            {
+                AllRequirementsSatisfied = true,
+                Results = new List<RequirementCheckResult>()
+            };
+
+            // Start all checks concurrently
+            var checkTasks = new List<Task<(SetupRequirements.Requirement req, DefaultCommandResponse result)>>();
+            
+            foreach (var req in reqsToCheck)
```
samvaity, line 101: Possibly future scope, currently, it appears the tool checks each language’s requirements sequentially. Given the small number of languages and that each check involves spawning a process (which can be IO-bound waiting for the process output), there is an opportunity for parallelism.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
JennyPng: The expected type for the langs arg is `List<SdkLangauge>` so passing in a value not defined in the `SdkLanguage` enum will automatically fail when calling this function, both from MCP and CLI

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
JennyPng: It's hard to standardize because devs can name virtual environments arbitrarily, as well as have multiple venvs anywhere. For example, the Python repo could have different venvs for each separate service directory. 

Without knowing the specific venv, it would be difficult to verify the Python setup since all the Python specific requirements should be installed into a venv, not the outer system.

Currently, I have a fallback where if no venv path is provided, it does attempt to find a venv based on the most generic possible names in the user's working directory, but that's only intended as a fallback

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/JavaRequirementsCheck.cs
JennyPng: The main use case currently is that Python has unique logic for updating commands with a venv path. I can imagine that in the future, if we want to fetch requirements in a more flexible/dynamic way than from reading a central config, each language may have more differences?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/DotNetRequirementsCheck.cs
JennyPng: I think that is similar to what I'm trying to do in my current approach, I parse the JSON only once in the tool class itself here https://github.com/Azure/azure-sdk-tools/pull/12476/files/c9c13b47b49bf13460d5bacb2e3ebee7812f5b13#diff-79c84bf73e629cc27a03b11185c9f62d2486af16919cceb81edf3757c3fcedc2R253 and pass the parsed dictionary to each language getter to return its own requirements from that 🤔 I tried to keep the config-getting logic simple since that implementation may change very soon for the future iterations

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
benbp: Yeah I think all the hardening is for a later milestone

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
JennyPng: The tool is currently concurrently starting each requirement check and awaiting all to finish - is there a way to parallelize it further?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ILanguageSpecificResolver.cs
JennyPng: sounds good, refactored according to timo's suggestion

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
benbp: @samvaity this should already be running concurrently now (I had the same comment originally). If we want to fully parallelize (vs. just non-blocking I/O) we can thread these operations but I'm not sure that will buy us much more perf, but probably worth investigating at some point.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
MaryGao: These environments look good for JS.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/DotNetRequirementsCheck.cs
m-redding: I agree that it would be nicer to have each langauge check get its relevant slice, but since this is the initial version and may change that seems reasonable to me

m-redding (approved):
This lgtm for milestone 1!

timovv (approved):
Looks good to merge for scenario 1

l0lawrence (approved):
lgtm excited to try it in v1

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/VerifySetupResponse.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+        return sb.ToString();
+    }
+}
+
+public class RequirementCheckResult
+{
+    public string Requirement { get; set; }
+    public List<String> Instructions { get; set; }
+    public string? Output { get; set; }
```
samvaity, line 53: Could we add docs for these properties and model?

samvaity: Here and other model classes.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ILanguageSpecificResolver.cs
```diff
@@ -13,4 +14,12 @@ public interface ILanguageSpecificResolver<T> where T : class
...
     /// <returns>The resolved service instance, or null if no matching service is found.</returns>
     public Task<T?> Resolve(string packagePath, CancellationToken ct = default);
+
+    /// <summary>
+    /// Resolves the language-specific service implementation for a given Sdk language.
+    /// </summary>
+    /// <param name="language">A language identifier.</param>
+    /// <param name="ct">Cancellation token.</param>
+    /// <returns>A list of resolved service instances (null for languages that could not be resolved).</returns>
+    public Task<T?> Resolve(SdkLanguage language, CancellationToken ct = default);
```
samvaity, line 24: This appears to be a different method rather than just an overload. Should we consider naming it ResolveAll?

samvaity, line 24: Based on the documentation comment (docstring) the return type should be a list of service instances. To match the docstring, the return type should be: `Task<List<T?>>`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/DotNetRequirementsCheck.cs
samvaity: @JennyPng If you're planning to revisit this, could you file a follow-up issue for better tracking? Extracting config loading and using dependency injection for language-specific requirement slices would be a useful improvement.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/JavaRequirementsCheck.cs
samvaity: I’m good with moving forward with V1 using this approach.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,312 @@
...
+            throw new Exception("Failed to parse requirements JSON.");
+        }
+
+        return setupRequirements.categories;
+    }
+
+    private string CheckVersion(string output, SetupRequirements.Requirement req)
+    {
+        // Return empty string if version requirement is satisfied, else return version to upgrade to
+        string versionPattern = @"(>=|<=|>|<|=)\s*([\d\.]+)";
```
samvaity: Declare as a constant for reuse and clarity.
Same for below.

samvaity (changes requested):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ILanguageSpecificResolver.cs
timovv: I would consider this an overload

samvaity: @timovv I initially did think the same but the return type here makes a difference. The new method returns multiple services (a list of results), which makes it fundamentally different from resolving a single service by path or language. For instance, if both methods "resolve a service" but one uses a package path and the other uses a language, then overloading would make sense.
Wdyt?

JennyPng: I updated this function to just return a single service not a list - I'll update the outdated docstring

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/DotNetRequirementsCheck.cs
JennyPng: sure, I made the issue here: https://github.com/Azure/azure-sdk-tools/issues/12742

samvaity (approved):
@JennyPng thanks for making the changes, excited to see this in action! :shipit:

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/PythonRequirementsCheck.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    private string FindVenv(string packagePath)
+    {
+        // try to find existing venv in package path if none was provided
+        if (string.IsNullOrWhiteSpace(packagePath))
+        {
+            packagePath = Environment.CurrentDirectory;
+        }
+
+        var candidates = new[] { ".venv", "venv", ".env", "env" };
```
richardpark-msft, line 68: `.env` shouldn't be used, as that's typically the name of the .env file that we use for loading up the environment variables.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -0,0 +1,126 @@
...
+                "check": ["golangci-lint", "--version"],
+                "instructions": ["https://golangci-lint.run/docs/welcome/install/"]
+            },
+            {
+                "requirement": "generator",
+                "check": ["generator", "--help"],
+                "instructions": ["go install github.com/Azure/azure-sdk-for-go/eng/tools/generator@latest"]
+            }
+        ],
+        "python": [
```
richardpark-msft, line 98: Can these dependencies just be emitted into the dev_requirements.txt when the project is generated?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/GoRequirementsCheck.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Sdk.Tools.Cli.Models;
+
+public class GoRequirementsCheck : IEnvRequirementsCheck
+{
+    public List<SetupRequirements.Requirement> GetRequirements(string packagePath, Dictionary<string, List<SetupRequirements.Requirement>> categories, CancellationToken ct = default)
+    {
+        return categories.TryGetValue("go", out var requirements) ? requirements : new List<SetupRequirements.Requirement>();
```
richardpark-msft, line 14: It seems odd to me that the only function of this code is to go into a dictionary it didn't load to grab something with a hardcoded name.

It'd make more sense to just have these defined, in code, and then do away with `categories`.

CC: @benbp

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -0,0 +1,313 @@
...
+        var langs = parseResult.GetValue(languagesParam);
+        var allLangs = parseResult.GetValue(allLangOption);
+        var parsed = allLangs ? Enum.GetValues<SdkLanguage>().ToHashSet() : langs.ToHashSet();
+        var packagePath = parseResult.GetValue(SharedOptions.PackagePath);
+        var venvPath = parseResult.GetValue(venvOption);
+        return await VerifySetup(parsed, packagePath, venvPath, ct);
+    }
+
+    [McpServerTool(Name = "azsdk_verify_setup"), Description("Verifies the developer environment for MCP release tool requirements. Accepts a list of supported languages to check requirements for, and the packagePath of the repo to check. Accepts a specific Python virtual environment path to use for Python requirements checks.")]
+    public async Task<VerifySetupResponse> VerifySetup(HashSet<SdkLanguage> langs = null, string packagePath = null, string venvPath = null, CancellationToken ct = default)
```
richardpark-msft, line 87: I think we can consider just saying that the .venv has to be in a particular spot, with a particular name, and do away with this parameter.

Otherwise, we might end up having to add and propagate parameters for all languages as they come up with requirements.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/GoRequirementsCheck.cs
richardpark-msft: But otherwise, this single line looks amazing.

richardpark-msft (approved):
I left some comments on there - it'd be good to think about the design, and whether it's possibly a leaky abstraction, but this is perfectly fine for v1.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/GoRequirementsCheck.cs
praveenkuttappan: We need to combine language services into one class for each language instead of spinning off a language class for each feature.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/VerifySetupResponse.cs
```diff
@@ -0,0 +1,54 @@
...
+// Licensed under the MIT License.
+using System.Text.Json.Serialization;
+using System.Text;
+namespace Azure.Sdk.Tools.Cli.Models;
+
+/// <summary>
+/// Response model for setup verification operations
+/// </summary>
+public class VerifySetupResponse : CommandResponse
+{
```
praveenkuttappan: @JennyPng Can you please add language as field int this response and update it. This will help us tag this tool call language level telemetry.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
l0lawrence: we can talk with Izzy and see, I didn't think we added things to the dev_reqs file when we generate but can double check

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/GoRequirementsCheck.cs
timovv: > We need to combine language services into one class for each language instead of spinning off a language class for each feature.

Can you elaborate on the rationale here? If we do that then the language specific class would grow significantly in size as we add more features and have poor cohesion. I would prefer having the language-specific bits be specific to each feature so that we can keep separation of concerns.

praveenkuttappan (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
JennyPng: I'm a little skeptical of how this would affect the experience of Python devs, but I'll add this to an issue for future consideration for the next scenario https://github.com/Azure/azure-sdk-tools/issues/12632

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/VerifySetup/GoRequirementsCheck.cs
JennyPng: resolving this so I can merge but I made note of this open topic on an issue for discussing Scenario 2 https://github.com/Azure/azure-sdk-tools/issues/12632

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
JennyPng: added to https://github.com/Azure/azure-sdk-tools/issues/12632

Issue #10229, opened 2025-04-07
Java duplicate artifactIds: validate-all-packages.yml, Validate-All-Packages.ps1 and Validate-Package.ps1


JimSuplizio:
validate-all-packages.yml calls Validate-All-Packages.ps1 with the list of Parameters.Artifacts and only uses the $artifact.name. This is, in turn, passed on to Validate-Package.ps1 which, the first thing it does, is load the PackageInfo file. This should probably looping through the PackageInfo files and passing them along instead of looking at Artifacts. There's a lot of other functions calls, especially around the API Review and Update-DevOps-Release-WorkItem.ps1 calls that only use the name. It's unclear what changes are going to need to be made to update the DevOps release item to make but I believe this particular file is going to have cascading changes that hit a number of other common files.

PR #12478, opened 2025-10-13
Updated validate pkg template to use packageInfo

raych1:
The change is to support Java v2 library which needs to get `groupId` and `artifactId` from `packageInfo` to differenciate same `artifactId`. 
Refer to https://github.com/Azure/azure-sdk-tools/issues/10229 for details.

* Added a new `PackageInfoFiles` parameter to both the pipeline template and the PowerShell script

* Refactored parameter handling in `Validate-All-Packages.ps1` to prioritize `Artifacts`, then `PackageInfoFiles`, ensuring backward compatibility with previous usages while supporting the new method.

* Refactored the `ProcessPackage` function to work with package info objects

* Moved functions in the `Validate-Package.ps1` to `Validate-All-Packages.ps1` and the former script will be deleted in separate PR

* Deprecated `Update-DevOps-Release-WorkItem.ps1` and consolidated the code into the `DevOps-WorkItem-Helpers.ps1`

### DIFF
--- a/eng/common/scripts/Validate-All-Packages.ps1
+++ b/eng/common/scripts/Validate-All-Packages.ps1
@@ -178,7 +178,7 @@ function ProcessPackage($packageInfo)
     $isShipped = IsVersionShipped $pkgName $versionString
     if ($isShipped) {
         Write-Host "Package work item already exists for version [$versionString] that is marked as shipped. Skipping the update of package work item."
-        exit 0
+        return
     }
 
     Write-Host "Validating package $pkgName with version $versionString."


eng/common/scripts/Prepare-Release.ps1
```diff
@@ -141,18 +142,18 @@ if ($null -eq $newVersionParsed)
   exit 1
 }
 
-&$EngCommonScriptsDir/Update-DevOps-Release-WorkItem.ps1 `
```
weshaggard, line 144: I assume we should also update this script to use the newly created function.

eng/common/scripts/Validate-All-Packages.ps1
```diff
@@ -12,39 +12,287 @@ Param (
...
+        Message = ""
+    }
+    ValidateChangeLog $changeLogPath $versionString $changeLogStatus
+
+    # API review and package name validation
+    $fullPackageName = $pkgName
+
+    # If there's a groupId that means this is Java and pkgName = GroupId+ArtifactName
+    # but the VerifyAPIReview requires GroupId:ArtifactName
+    if ($packageInfo.PSObject.Members.Name -contains "Group") {
```
weshaggard, line 200: Is this the core of the changes? The rest seem mostly like refactoring, which is fine but I want to understand what we are actually changing.

eng/common/scripts/Prepare-Release.ps1
raych1: There're only two places calling this script. One is the `validate-package.ps1` and the other is the `prepare-release.ps1`. Both of them have been updated in this PR. 

This script will be deleted as part of the clean-up task.

eng/common/scripts/Validate-All-Packages.ps1
raych1: Line#32 to line#165 were copied from the original `validate-package.ps1`, which will be deleted as part of the clean-up task.

Main changes include:
`ProcessPackage` function was updated to take `packageInfo` parameter.  When `parameters.ArtifactList` is provided for current usage, code was added(line#265) to parse them and create `packageInfo`, and then it will be passed into `ProcessPackage` function.

The following snippet is the original code, which means `VerifyAPIReview` already consumes the full name including `groupId` for Java package name.

```
# API review and package name validation
$fulPackageName = $pkgName
$groupId = $null
if ($pkgInfo.PSObject.Members.Name -contains "Group") {
    $groupId = $pkgInfo.Group
}
if ($groupId){
    $fulPackageName = "${groupId}:${pkgName}"
}
```
@praveenkuttappan do you know if Java uses `groupId:artifactId` as identifier in APIReview service? It seems to be the case from the `apiview`, e.g. https://spa.apiview.dev/review/c12fd0a7ffe74c2c8349a181345bffff?activeApiRevisionId=fe96d87ce3f84e48b7b42bc99b8082cd

eng/common/scripts/Prepare-Release.ps1
weshaggard: I think I'm OK with deleting that script but please do update the docs at https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/444/Marking-a-Package-as-In-Release what references it. I don't expect anyone is using it.

weshaggard (approved):

PR #12492, opened 2025-10-14
[AVC] Make evals take into consideration rationale similarity

tjprescott:
Updates evals to simplify structure and take rationale similarity into account.

Modified filter_existing_comments so that it better passes the evals. 

Updates the output so it looks more like unit test output
image

Coverts all eval test cases to yaml.

### DIFF
--- a/packages/python-packages/apiview-copilot/.gitignore
+++ b/packages/python-packages/apiview-copilot/.gitignore
@@ -5,3 +5,4 @@ raw_requests
 build
 .runs
 *.log
+flow.flex.yaml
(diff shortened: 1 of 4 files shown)


packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -662,42 +679,40 @@ def process_results(self, raw_results: list, guideline_ids: set = None) -> dict:
 
                 run_summary = {
                     "total_tests": 0,
-                    f"{self.correct_field_name}s": 0,
+                    "correct_actions": 0,
                     "test_results": [],
-                    f"{self.comparison_field}_breakdown": self.breakdown_categories.copy(),
+                    "action_breakdown": self.breakdown_categories.copy(),
```
marcarro: I would just say to add `"max_score": 0` here and we're good 👍

marcarro (approved):

PR #12494, opened 2025-10-14
CPEX Attestation: Add Warning for unknown states

smw-ms:
Fixes bug where Kpi Id was set equal to comment message when lifecycle was unknown


maririos (approved):
Talked offline about the implementation in this PR

PR #12498, opened 2025-10-14
Update issue closure and reminder timelines

samvaity:
Currently, after the `needs-author-feedback` label is assigned, the GitHub event processor follows this workflow:

- The issue is tracked as awaiting a response from the author.
- If the author does not respond within 7 days, the issue is marked as stale using a scheduled rule.
- If there is still no response after 14 days, the issue is closed by another scheduled rule.

Similarly, for label, `issue-addressed` updated 
- to close issues after 3 business days instead of the previous 7 calendar days

At present, the dashboard tracks these issues until they are closed due to inactivity and the wait times are a little too long.

The proposed change is to reduce these waiting periods to:
- 3 business days before marking as stale
- 5 business days before closing


samvaity:
The current PR will be closed in favor of broader SLA/automation proposal.
- Aggressive timelines not ideal for OSS contributors.
- Better approach: update SLA tracking logic rather than shortening response windows.
- @ronniegeraghty to explore SLA adjustments and automation improvements.

PR #12500, opened 2025-10-15
Refactor: Migrate to OpenAI and streamline Result pattern usage

radhgupta:
# Migrate from Persistent Agent to OpenAI and Refactor Result Pattern Usage

## Overview
This PR modernizes the agent architecture by migrating from a persistent agent model to OpenAI while also cleaning up unnecessary Result pattern usage for better code clarity and performance.

### 2. Removed Result Pattern from InputValidator
- **Rationale**: Input validation should fail fast - there's no point in continuing execution if the input is invalid
- Simplified validation logic by throwing exceptions directly rather than wrapping in Result pattern
- Reduced complexity and improved performance by eliminating unnecessary Result object creation
- Made error handling more straightforward and explicit

### 3. Moved Iterative Flow to Orchestrator Class
- Centralized the iterative processing logic in the orchestrator
- Improved separation of concerns and code organization

### 4. Added BuildWorkflow.cs
- Introduced dedicated workflow class for build operations (compile typespec and build generated library)
- Provides structured approach to build pipeline management
- Improves code organization and reusability

### DIFF
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/OpenAIService.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/OpenAIService.cs
@@ -185,20 +185,52 @@ internal class OpenAIService
             }
         }
         
-        var finalResponse = await chatClient.CompleteChatAsync(messages, cancellationToken: cancellationToken);
+        // Keep calling until we get a non-tool response
+        ChatCompletion finalResponse;
+        int maxIterations = 10; // Prevent infinite loops
+        int iteration = 0;
         
-        string finalResult;
-        if (finalResponse?.Value?.Content != null && finalResponse.Value.Content.Count > 0 && finalResponse.Value.Content[0] != null)
+        do
         {
-            finalResult = finalResponse.Value.Content[0].Text;
+            var response = await chatClient.CompleteChatAsync(messages, cancellationToken: cancellationToken);
+            finalResponse = response.Value;
+            
+            // If we get more tool calls, handle them
+            if (finalResponse.FinishReason == ChatFinishReason.ToolCalls)
+            {
+                messages.Add(new AssistantChatMessage(finalResponse));
+                
+                foreach (var toolCall in finalResponse.ToolCalls)
+                {
+                    if (toolCall is ChatToolCall functionCall)
+                    {
+                        Logger.LogDebug("Executing additional tool: {ToolName} (iteration {Iteration})", 
+                            functionCall.FunctionName, iteration + 1);
+                        var argumentsJson = functionCall.FunctionArguments.ToString();
+                        var result = await ToolExecutor.ExecuteToolCallAsync(functionCall.FunctionName, argumentsJson, validationContext, cancellationToken);
+                        messages.Add(new ToolChatMessage(functionCall.Id, result));
+                    }
+                }
+            }
+            
+            iteration++;
+        } 
+        while (finalResponse.FinishReason == ChatFinishReason.ToolCalls && iteration < maxIterations);
+        
+        if (iteration >= maxIterations)
+        {
+            Logger.LogWarning("Maximum tool call iterations reached ({MaxIterations}), stopping", maxIterations);
+        }
+        
+        if (finalResponse?.Content != null && finalResponse.Content.Count > 0 && finalResponse.Content[0] != null)
+        {
+            return finalResponse.Content[0].Text;
         }
         else
         {
-            Logger.LogWarning("OpenAI response did not contain any content.");
-            finalResult = string.Empty;
+            Logger.LogWarning("OpenAI response did not contain any content after tool calls.");
+            return string.Empty;
         }
-        
-        return finalResult;
     }
 
     /// <summary>
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
@@ -59,7 +59,7 @@ namespace Azure.Tools.GeneratorAgent.Configuration
         
 
         // OpenAI Settings
-        public string? OpenAIApiKey => Configuration.GetSection("OpenAI:ApiKey").Value;
+        public string? OpenAIApiKey => Configuration.GetSection("OpenAI:ApiKey").Value ?? EnvironmentVariables.OpenAIApiKey;
         public string? OpenAIEndpoint => Configuration.GetSection("OpenAI:Endpoint").Value;
         public string OpenAIModel => Configuration.GetSection("OpenAI:Model").Value ?? "gpt-4o";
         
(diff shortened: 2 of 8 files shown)


src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/Azure.Tools.GeneratorAgent.Tests.csproj
```diff
@@ -12,13 +12,13 @@
...
     <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
-    <PackageReference Include="Azure.AI.Agents.Persistent" Version="1.1.0-beta.3" />
     <PackageReference Include="Azure.Core" Version="1.47.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="9.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="9.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="9.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="9.0.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="9.0.0" />
+    <PackageReference Include="OpenAI" Version="2.1.0" />
```
m-redding:
```suggestion
```
nit you should be able to remove these because the tests would get these references transitively through the project reference on Azure.Tools.GeneratorAgent

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
```diff
@@ -57,6 +57,12 @@ public AppSettings(IConfiguration configuration, ILogger<AppSettings> logger)
         public int DelayBetweenFixesMs => 
             int.Parse(Configuration.GetSection("AzureSettings:DelayBetweenFixesMs").Value ?? "500");
         
+
+        // OpenAI Settings
+        public string? OpenAIApiKey => Configuration.GetSection("OpenAI:ApiKey").Value;
```
m-redding: Imo I think it would be nice to have an environment variable option for this too, personally I like the environment variable option because I know it can't get accidentally checked in with a commit or something

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/OpenAIService.cs
```diff
@@ -0,0 +1,263 @@
...
+            if (toolCall is ChatToolCall functionCall)
+            {
+                Logger.LogDebug("Executing tool: {ToolName}", functionCall.FunctionName);
+                var argumentsJson = functionCall.FunctionArguments.ToString();
+                var result = await ToolExecutor.ExecuteToolCallAsync(functionCall.FunctionName, argumentsJson, validationContext, cancellationToken);
+                messages.Add(new ToolChatMessage(functionCall.Id, result));
+            }
+        }
+        
+        var finalResponse = await chatClient.CompleteChatAsync(messages, cancellationToken: cancellationToken);
```
m-redding: I'm curious are there ever cases where the agent might send additional tool calls or any cases where this isn't the final response? Or is this pretty much always the case where this will be the client.tsp output?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/BuildWorkflow.cs
```diff
@@ -0,0 +1,45 @@
...
+        LocalLibraryGenerationService compileService,
+        LibraryBuildService buildService,
+        ILogger<BuildWorkflow> logger)
+    {
+        CompileService = compileService ?? throw new ArgumentNullException(nameof(compileService));
+        BuildService = buildService ?? throw new ArgumentNullException(nameof(buildService));
+        Logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    public async Task<Result<object>> ExecuteAsync(ValidationContext context, CancellationToken cancellationToken)
```
m-redding, line 25: This is all looking really good! Sometimes splitting things up can be hard to follow, but the way you have each thing organized is really nice, the method and parameter names make it really obvious what each method is responsible for, so I feel like I can read each method on its own and understand what it's doing without needing to look at anything that it calls

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GenerationOrchestrator.cs
```diff
@@ -0,0 +1,140 @@
...
+            if (fixes.Count == 0)
+            {
+                Logger.LogWarning("No fixes generated for compilation/build errors - errors may not be addressable");
+                throw new InvalidOperationException("No addressable errors found in compilation/build output");
+            }
+
+            Logger.LogInformation("Generated {FixCount} fixes for iteration {Iteration}", fixes.Count, currentIteration + 1);
+
+            // Apply fixes and apply the changes
+            var patchSuccess = await GenerateAndApplyFixesAsync(fixes, validationContext, cancellationToken);
```
m-redding, line 111: This name slightly confused me because I assumed "Generate" meant generate code but it means have an LLM generate, but I can't think of a better name  and the name does say exactly what it does so I think it's fine. But just commenting in case there's something you think of later

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Security/InputValidator.cs
```diff
@@ -26,225 +26,193 @@ internal static class InputValidator
...
+            return path;
         }
 
         /// <summary>
         /// Validates a TypeSpec path for security and format.
         /// </summary>
         /// <param name="path">The TypeSpec path to validate</param>
         /// <param name="isLocalPath">True if this is a local filesystem path, false if it's a relative repository path</param>
-        public static Result<string> ValidateTypeSpecDir(string? path, bool isLocalPath = true)
+        public static string ValidateTypeSpecDir(string? path, bool isLocalPath = true)
```
m-redding: Nice I think it makes sense for this to throw instead

m-redding: nit but since the method actually makes some changes to the path maybe it's worth naming it ValidateAndNormalizeTypeSpecDir or something like that

There are also quite a few of the same thing in this file, this is definitely a nit but just in my opinion only including validate in the name makes it feel like it's just doing checks and not changing anything

m-redding (approved):
Nice I think this looks good!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/OpenAIService.cs
radhgupta: The current implementation has a limitation that it only handles one round of tool calls, but the agent could potentially need multiple rounds of investigation before providing the final client.tsp output. FIxing it

radhgupta: Fixed!

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
radhgupta: Fixed!

m-redding (approved):

PR #12415, opened 2025-10-08
Update CODEOWNERS to include ai-evals directory

maririos:
Added CODEOWNERS for ai-evals directory.

PR #12505, opened 2025-10-15
Restore azsdk-cli auto-PR label

benbp:
Seems like this was inadvertently removed in https://github.com/Azure/azure-sdk-tools/pull/12415


maririos (approved):
Thank you!!

jsquire (approved):

Issue #12489, opened 2025-10-14
AzSDK MCP server: Create release plan to check for existing release plan


praveenkuttappan:
Create release plan takes API spec as a param and check for an existing release plan for the product. It creates a release plan if there is no release plan found with same exact API spec PR.

We should also check if there is already a release plan for the same product and for same API version that is either in `In progress` or 'New` or `Not started`. If a release plan is found, then the tools should send a response to show existing release plan for the product and ask user to verify the release plan and confirm if they still want to create a new release plan.

Tool should also add an optional param to force create a release plan for the product. If user confirms to create a release plan anyway then this flag should be set to true and create a release plan without checking for existing release plans.

smw-ms:
After PR (https://github.com/Azure/azure-sdk-tools/pull/125160)
**Chat 1**
<img width="1543" height="1280" alt="Image" src="https://github.com/user-attachments/assets/5a32df10-a8f0-43bb-a283-68764ac2d3f3" />

<img width="1547" height="1252" alt="Image" src="https://github.com/user-attachments/assets/d64dacf2-796a-4025-bfe5-3eeab1303532" />

**Chat 2**

<img width="1618" height="1276" alt="Image" src="https://github.com/user-attachments/assets/fd3c900f-1e9c-4d91-80b1-4f25e6927dbe" />

<img width="1555" height="1283" alt="Image" src="https://github.com/user-attachments/assets/9da0fbd2-2d4a-464b-889e-9b1e081ebcb5" />

PR #12516, opened 2025-10-15
Allow forced creation when release plans already exist

smw-ms:
Issue: https://github.com/Azure/azure-sdk-tools/issues/12489

### DIFF
--- a/eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
+++ b/eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
@@ -13,7 +13,16 @@ Follow these steps in order to create or manage a release plan for an API specif
 - If no pull request is available, prompt the user to provide the API spec pull request link
 - Validate that the provided pull request link is accessible and valid
 
-## Step 2: Gather Release Plan Information
+## Step 2: Check for Existing Release Plan
+- Ask the user if they already have an existing release plan
+- If they confirm:
+    - Query the existing release plan using either:
+        - The release plan number, or
+        - The API spec pull request link
+    - Display the existing release plan details (Release Plan ID, status, associated languages, SDK PRs).
+- If no existing release plan is found, continue to Step 3 to gather required details for creating a new one. 
+
+## Step 3: Gather Release Plan Information
 Collect the following required information from the user. Do not create a release plan with temporary values. Confirm the values with the user before proceeding to create the release plan.
 If any details are missing, prompt the user accordingly:
 
@@ -25,7 +34,7 @@ If any details are missing, prompt the user accordingly:
     - "beta" for preview API versions
     - "stable" for GA API versions
 
-## Step 3: Create Release Plan
+## Step 4: Create Release Plan
 - If the user doesn't know the required details, direct them to create a release plan using the release planner
 - Provide this resource: [Release Plan Creation Guide](https://eng.ms/docs/products/azure-developer-experience/plan/release-plan-create)
 - Once all information is gathered, use `azsdk_create_release_plan` to create the release plan
@@ -33,18 +42,18 @@ If any details are missing, prompt the user accordingly:
 - Display the newly created release plan details to the user for confirmation
 - Refer to #file:sdk-details-in-release-plan.instructions.md to identify languages configured in the TypeSpec project and add them to the release plan
 
-## Step 4: Update SDK Details in Release Plan
+## Step 5: Update SDK Details in Release Plan
 - Refer to #file:sdk-details-in-release-plan.instructions.md to add languages and package names to the release plan
 - If the TypeSpec project is for a management plane, refer to #file:verify-namespace-approval.instructions.md if this is first release of SDK.
 
-## Step 5: Link SDK Pull Requests (if applicable)
+## Step 6: Link SDK Pull Requests (if applicable)
 - Ask the user if they have already created SDK pull requests locally for any programming language
 - If SDK pull requests exist:
     - Collect the pull request links from the user
     - Use `azsdk_link_sdk_pull_request_to_release_plan` to link each SDK pull request to the release plan
     - Confirm successful linking for each SDK pull request
 
-## Step 6: Summary
+## Step 7: Summary
 - Display a summary of the completed actions:
     - Release plan status (created or existing)
     - Linked SDK pull requests (if any)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
```diff
@@ -136,6 +137,38 @@ public async Task<ReleasePlanDetails> GetReleasePlanAsync(int releasePlanId)
             return await MapWorkItemToReleasePlanAsync(releasePlanWorkItems[0]);
         }
 
+        public async Task<List<ReleasePlanDetails>> GetReleasePlansForProductAsync(string productTreeId, string specApiVersion)
+        {
+            try
+            {
+                var query = $"SELECT [System.Id] FROM WorkItems WHERE [System.TeamProject] = '{Constants.AZURE_SDK_DEVOPS_RELEASE_PROJECT}' AND [Custom.ProductServiceTreeID] = '{productTreeId}' AND [System.WorkItemType] = 'Release Plan' AND [System.State] IN ('New','Not Started','In Progress')";
```
praveenkuttappan: This should skip the test release plans created with Tag `Release Planner App Test`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -235,21 +237,37 @@ public async Task<ObjectCommandResponse> CreateReleasePlan(string typeSpecProjec
...
+                    var existingReleasePlan = await devOpsService.GetReleasePlanAsync(specPullRequestUrl);
+                    if (existingReleasePlan != null && existingReleasePlan.WorkItemId > 0)
                     {
-                        Message = $"Release plan already exists for the pull request: {specPullRequestUrl}. Release plan link: {existingReleasePlan.ReleasePlanLink}",
-                        Result = existingReleasePlan
-                    };                    
-                }
+                        return new ObjectCommandResponse
+                        {
+                            Message = $"Release plan not created. Release plan already exists for the pull request: {specPullRequestUrl}. Release plan link: {existingReleasePlan.ReleasePlanLink}",
```
praveenkuttappan: Following sentence will make agent assume that request failed. So, let's remove that part.
`Release plan not created`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -235,21 +237,37 @@ public async Task<ObjectCommandResponse> CreateReleasePlan(string typeSpecProjec
...
+                            Result = existingReleasePlan
+                        };
+                    }
 
+                    logger.LogInformation("Checking for existing release plans for product: {productTreeId}", productTreeId);
+                    var existingReleasePlans = await devOpsService.GetReleasePlansForProductAsync(productTreeId, specApiVersion);
+                    existingReleasePlans = existingReleasePlans.Where(releasePlan => releasePlan.WorkItemId > 0).ToList();
```
praveenkuttappan: Why is `where` required? Is DevOps service returning release plans without WorkItemId?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -235,21 +237,37 @@ public async Task<ObjectCommandResponse> CreateReleasePlan(string typeSpecProjec
...
+                    if (existingReleasePlans.Any())
+                    {
+                        return new ObjectCommandResponse
+                        {
+                            Message = $" Release plan not created. Release plan(s) already exist for the product: {productTreeId}."
```
praveenkuttappan:
```suggestion
                            Message = $"An active release plan already exists for the product: {productTreeId}. " 
```

eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
```diff
@@ -31,25 +24,34 @@ If any details are missing, prompt the user accordingly:
...
 - Refer to #file:sdk-details-in-release-plan.instructions.md to identify languages configured in the TypeSpec project and add them to the release plan
 
-## Step 5: Update SDK Details in Release Plan
+### Step 3a: Handle Existing Release Plans
+- When `azsdk_create_release_plan` returns existing release plans.
+  - Extract and display key information: Release Plan ID, status, associated languages, SDK PRs
+  - Present the three options:
+    1. **Work with the existing release plan** - Use the current release plan and make any needed updates
+    2. **Force create a new release plan** - Create a completely new release plan even though one already exists
+    3. **Cancel** - Don't proceed with release plan creation
```
praveenkuttappan: **Cancel** should not be a choice. right? What's the use of it?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
praveenkuttappan: This should also check whether existing release plan is for same SDK release type.

maririos: ohh interesting. This is a great find!

maririos: > This should skip the test release plans created with Tag `Release Planner App Test`

with the condition that if the mcp is running under test mode, then we do want to make sure we only list the work items that have that tag :)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
maririos: but also, at this point we don't know if the release plan was created or not. we just know that one already exists

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -235,21 +237,37 @@ public async Task<ObjectCommandResponse> CreateReleasePlan(string typeSpecProjec
...
-                var existingReleasePlan = await devOpsService.GetReleasePlanAsync(specPullRequestUrl);
-                if (existingReleasePlan != null && existingReleasePlan.WorkItemId > 0)
+                ValidateCreateReleasePlanInputAsync(typeSpecProjectPath, serviceTreeId, productTreeId, specPullRequestUrl, sdkReleaseType, specApiVersion);
+                
+                if (!forceCreateReleasePlan)
                 {
-                    return new ObjectCommandResponse
+                    // Check for existing release plan for the given pull request URL.
+                    logger.LogInformation("Checking for existing release plan for pull request URL: {specPullRequestUrl}", specPullRequestUrl);
+                    var existingReleasePlan = await devOpsService.GetReleasePlanAsync(specPullRequestUrl);
```
maririos, line 343: if there are multiples Release plans with the same API spec PR, wouldn't this return a list of release plans?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
maririos: should the error also say that the user could force the creation? (that way we avoid putting that logic on the instructions file?)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
praveenkuttappan: Correct. So we should avoid the release plans with test app tag when test mode env variable is present.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
praveenkuttappan: I think service method fetches only latest release plan. @smw-ms can you please check and confirm this?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
praveenkuttappan: > should the error also say that the user could force the creation? (that way we avoid putting that logic on the instructions file?)

@smw-ms You can put a statement in `nextSteps` property in response to say "Prompt user to confirm whether to use existing release plan or force create a new release plan"

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
smw-ms: only one release plan is returned. looking at the code there is no sorting happening, it just returns the first linked release plan it comes across, so there is no guarantee that it is the latest release plan

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
```diff
@@ -136,6 +137,38 @@ public async Task<ReleasePlanDetails> GetReleasePlanAsync(int releasePlanId)
...
+                    {
+                        releasePlans.Add(releasePlan);
+                    }
+                }
+
+                return releasePlans;
+            }
+            catch (Exception ex)
+            {
+                logger.LogError($"Failed to get release plans for product id {productTreeId}. Error: {ex.Message}");
```
benbp:
```suggestion
                logger.LogError(ex, $"Failed to get release plans for product id {productTreeId}");
```

We should always log like this so we don't swallow the stack trace.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -108,6 +108,12 @@ IEnvironmentHelper environmentHelper
             Required = true,
         };
 
+        private readonly Option<bool> forceCreateReleasePlanOpt = new("--force-create-release")
```
benbp: Would the more conventional `--force` be sufficient?

```suggestion
        private readonly Option<bool> forceCreateReleasePlanOpt = new("--force")
```

eng/common/instructions/azsdk-tools/create-release-plan.instructions.md
```diff
@@ -31,25 +25,26 @@ If any details are missing, prompt the user accordingly:
     - "beta" for preview API versions
     - "stable" for GA API versions
 
-## Step 4: Create Release Plan
+## Step 3: Create Release Plan
```
praveenkuttappan: Agent should prompt the user to check if they have an existing release plan before we ask for the details to create. This will help with the use case if a user is already aware of previously created release plan so asking for details and to attempt to create a release plan will give bad user experience.
If they confirm about an existing release plan, then the agent should query existing release plan using release plan number or using pr.

smw-ms: done

praveenkuttappan (approved):

PR #12524, opened 2025-10-16
[azsdkcli] Add boolean success/fail to RunAllTests

timovv:
While we take time to consider a more detailed contract for running tests, having a bool for success/failure is probably a good first step.


l0lawrence (approved):
lgtm as a starting point

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Tests/JavaScriptTestRunner.cs
```diff
@@ -1,20 +1,19 @@
 using Azure.Sdk.Tools.Cli.Helpers;
-using Azure.Sdk.Tools.Cli.Models;
 
 namespace Azure.Sdk.Tools.Cli.Services.Tests;
 
 public class JavaScriptTestRunner(IProcessHelper processHelper) : ITestRunner
 {
-    public async Task RunAllTests(string packagePath, CancellationToken ct = default)
+    public async Task<bool> RunAllTests(string packagePath, CancellationToken ct = default)
```
samvaity, line 7: Instead of the boolean as return type. We should consider setting DefaultCommandResponse as the return type that way we don't have to do the translation twice? From result code -> bool -> DefaultCommandResponse

timovv: It's a conscious decision not to return a Response here since I don't think we should couple the test runner interface to the concept of a command response.

samvaity: I think providing a response model would help to pass and parse test results in failure cases to provide better guidance to the users?

timovv: The test output is already available since it gets output by the process helper. Whether or not we want to return additional information from the test runner is an open question, but whatever information we do decide we want to return I do not believe that the command Response type is the appropriate vehicle for it. The test runner is not a command -- it's an interface that tools or other parts of the CLI app could use -- and so it doesn't make sense to me that it should return a command response.

As I said in the PR description, this is just a first step. If we decide in future that it would be useful to return more information, next steps, etc., then we can adjust at that time, but it's not a discussion I think we should have right now.

samvaity: I agree that we should improve the return type in the future so it's more specific and helpful than just a command response. However, at the moment, I don't see much value in having two translations either. But happy to punt this for future iterations.

PR #12531, opened 2025-10-16
validation running prompt

l0lawrence:
Will add to each copilot-instructions:

```
# Check PR and Validation Checks of SDK
Run [check package validation](..\eng\common\instructions\azsdk-tools\check-package-validation.instructions.md) to check if the package validation has passed for the SDK package.
```

prompt: Run all PR checks locally for the Health Deidentification SDKs before I create pull requests.
Outputs:

image

### DIFF
--- a/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
+++ b/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
@@ -26,11 +26,9 @@ Check the validation checks for the SDK package by collecting the required infor
       - generatedcodechecks
       - samples
 
-
 2. **Execute Check**:
     - Use the `azsdk_package_run_check` tool with the package path and check type.
     - If the package path is not provided, use the package name and language to determine the package path for each language and prompt the user to confirm the path before proceeding.
-
 
 3. **Present Results**:
     - If the package has passed all validation checks, highlight and finish
--- a/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
+++ b/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
@@ -18,5 +18,6 @@ Run the validation checks for the SDK package by collecting the required informa
 
 ## Expected User Interaction Flow
 1. Ask: "What is the exact path of the package you want to check for validation?"
+2. Ask: "Please specify the check type you want to run (e.g., 'all', 'lint', 'format', etc.)"
 3. Execute the check using the provided information
 4. Display results and next steps
--- a/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
+++ b/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
@@ -1,8 +1,8 @@
 ---
-description: 'Check if the package pr checks and/or validation has passed for the SDK package.'
+description: 'Check if the package validation checks have passed for the SDK package.'
 ---
 ## Goal
-Check the PR and Validation checks for the SDK package by collecting the required information from the user and executing the run_check command.
+Check the validation checks for the SDK package by collecting the required information from the user and executing the run_check command.
 
 ## Instructions
 1. **Collect Required Information**:
@@ -15,14 +15,19 @@ Check the PR and Validation checks for the SDK package by collecting the require
       - Go
     - Prompt the user to select the check type to run (default to "all" if not specified):
       - all
+      - changelog
+      - dependency
+      - readme
+      - cspell
+      - snippets
       - linting
       - format
-      - changelog
-      - cspell
-      - readme
+      - checkaotcompat
+      - generatedcodechecks
+      - samples
 
 
-2. **Execute Readiness Check**:
+2. **Execute Check**:
     - Use the `azsdk_package_run_check` tool with the package path and check type.
     - If the package path is not provided, use the package name and language to determine the package path for each language and prompt the user to confirm the path before proceeding.
 
--- a/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
+++ b/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
@@ -11,7 +11,6 @@ Check the validation checks for the SDK package by collecting the required infor
 
 2. **Execute Check**:
     - Use the `azsdk_package_run_check` tool with the package path and check type.
-    - If the package path is not provided, use the package name and language to determine the package path for each language and prompt the user to confirm the path before proceeding.
 
 3. **Present Results**:
     - If the package has passed all validation checks, highlight and finish
--- a/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
+++ b/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
@@ -17,7 +17,6 @@ Check the validation checks for the SDK package by collecting the required infor
     - If the package is not ready, display the specific check types that need to be fixed for each language.
 
 ## Expected User Interaction Flow
-1. Ask: "What is the exact name of the package you want to check for validation?"
-2. Ask: "Please select the programming language for this package: Python, Java, JavaScript, .NET, Go, or All"
+1. Ask: "What is the exact path of the package you want to check for validation?"
 3. Execute the check using the provided information
 4. Display results and next steps
--- a/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
+++ b/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
@@ -6,13 +6,7 @@ Check the validation checks for the SDK package by collecting the required infor
 
 ## Instructions
 1. **Collect Required Information**:
-    - Prompt the user for the exact package name
-    - Prompt the user to select the programming language from the following options (case sensitive):
-      - Python
-      - Java
-      - JavaScript
-      - .NET
-      - Go
+    - Prompt the user for the absolute package path to the package directory for the SDK package they want to check.
     - Prompt the user specify values for each of the parameters required by the azsdk_package_run_check tool. Present the user with options to pick from for the allowed values specified by the parameter schema.
 
 2. **Execute Check**:
--- a/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
+++ b/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
@@ -2,7 +2,7 @@
 description: 'Check if the package validation checks have passed for the SDK package.'
 ---
 ## Goal
-Check the validation checks for the SDK package by collecting the required information from the user and executing the run_check command.
+Check the validation checks for the SDK package by collecting the required information from the user and executing the azsdk_package_run_check command.
 
 ## Instructions
 1. **Collect Required Information**:
@@ -13,18 +13,7 @@ Check the validation checks for the SDK package by collecting the required infor
       - JavaScript
       - .NET
       - Go
-    - Prompt the user to select the check type to run (default to "all" if not specified):
-      - all
-      - changelog
-      - dependency
-      - readme
-      - cspell
-      - snippets
-      - linting
-      - format
-      - checkaotcompat
-      - generatedcodechecks
-      - samples
+    - Prompt the user specify values for each of the parameters required by the azsdk_package_run_check tool. Present the user with options to pick from for the allowed values specified by the parameter schema.
 
 2. **Execute Check**:
     - Use the `azsdk_package_run_check` tool with the package path and check type.
--- a/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
+++ b/eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
@@ -1,0 +1,38 @@
+---
+description: 'Check if the package pr checks and/or validation has passed for the SDK package.'
+---
+## Goal
+Check the PR and Validation checks for the SDK package by collecting the required information from the user and executing the run_check command.
+
+## Instructions
+1. **Collect Required Information**:
+    - Prompt the user for the exact package name
+    - Prompt the user to select the programming language from the following options (case sensitive):
+      - Python
+      - Java
+      - JavaScript
+      - .NET
+      - Go
+    - Prompt the user to select the check type to run (default to "all" if not specified):
+      - all
+      - linting
+      - format
+      - changelog
+      - cspell
+      - readme
+
+
+2. **Execute Readiness Check**:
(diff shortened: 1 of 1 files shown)


eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,38 @@
...
+      - JavaScript
+      - .NET
+      - Go
+    - Prompt the user to select the check type to run (default to "all" if not specified):
+      - all
+      - linting
+      - format
+      - changelog
+      - cspell
+      - readme
```
l0lawrence: will add the other vals as added to repo

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,38 @@
...
+
+
+2. **Execute Readiness Check**:
```
l0lawrence:
```suggestion
2. **Execute Validation Check**:
```

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,38 @@
+---
```
l0lawrence, line 1: Do we want to add `package test` under here with `package run-checks`?

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,38 @@
...
+description: 'Check if the package pr checks and/or validation has passed for the SDK package.'
```
samvaity: Why are we referring to them as `pr checks` instead of just `checks`? It seems like users might want to run lint checks or other checks without necessarily creating a PR. Including `pr` feels unnecessary.

This comment applies to the text overall in the spec.

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,38 @@
...
+      - changelog
+      - cspell
+      - readme
+
+
+2. **Execute Readiness Check**:
+    - Use the `azsdk_package_run_check` tool with the package path and check type.
+    - If the package path is not provided, use the package name and language to determine the package path for each language and prompt the user to confirm the path before proceeding.
+
+
```
samvaity, line 14: nit: extra line

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,38 @@
+---
+description: 'Check if the package pr checks and/or validation has passed for the SDK package.'
+---
+## Goal
+Check the PR and Validation checks for the SDK package by collecting the required information from the user and executing the run_check command.
+
+## Instructions
+1. **Collect Required Information**:
+    - Prompt the user for the exact package name
```
samvaity: Do we want to prompt the user for the package path instead of the package name?

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,38 @@
...
+    - If the package path is not provided, use the package name and language to determine the package path for each language and prompt the user to confirm the path before proceeding.
+
+
+3. **Present Results**:
+    - If the package has passed all validation checks, highlight and finish
+    - If the package is not ready, display the specific check types that need to be fixed for each language.
+
+## Expected User Interaction Flow
+1. Ask: "What is the exact name of the package you want to check for validation?"
+2. Ask: "Please select the programming language for this package: Python, Java, JavaScript, .NET, Go, or All"
```
samvaity: I haven't encountered a situation where this wasn't inferred from the package directory the user was in. Are these instructions meant to be more of a precaution?

l0lawrence: this is so the user can say "run X,Y,Z lang" and prompt the user or try to infer where the pkg will live vs just running 1 lang and being done

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
l0lawrence: I had it as package name because if we want multiple languages to be run the package path would not be the same

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
l0lawrence: the spec refers to them as pr checks so for consistency I applied that language naming here -- let me remove everywhere since i have the other pr

timovv (approved):
Looks fine to me, we can adjust based on how it performs in practice

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,41 @@
+---
+description: 'Check if the package validation checks have passed for the SDK package.'
+---
+## Goal
+Check the validation checks for the SDK package by collecting the required information from the user and executing the run_check command.
```
benbp: What is `run_check` supposed to refer to? We should reference command/tool names exactly I think.

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Instructions
+1. **Collect Required Information**:
+    - Prompt the user for the exact package name
+    - Prompt the user to select the programming language from the following options (case sensitive):
```
benbp: Would it be more flexible to say something like below? That way we aren't hard-coding the enum values for language and check type which could drift in the future. OR I think we should consider auto-generating these instructions (or leveraging `[Description]` instead).

> Prompt the user specify values for each of the parameters required by the `azsdk_package_run_check` tool. Present the user with options to pick from for the allowed values specified by the parameter schema.

l0lawrence: ah good idea so it wont get stale

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
m-redding: But aren't the package names also a little different for each language too?

l0lawrence: yes good point, in my running of this I'm using a human worded version of the package name for example "keyvault keys" kind of like how the v0 spec had the wording set up "healthdeidentification sdk". Are there sdks that wildly vary naming wise from that? tbh I have stuck to the common sdks I know for testing

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
l0lawrence: @timovv you want to add on?

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,30 @@
+---
+description: 'Check if the package validation checks have passed for the SDK package.'
```
samvaity: @LibbaLawrence Should these be appended in here https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/instructions/azsdk-tools/local-sdk-workflow.instructions.md ?

maririos: maybe mentioned, but keeping this separate so people can call it independent of the E2E flow

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
maririos: it depends. how is the package name used?

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
l0lawrence: oh good catch let me add it as part c

l0lawrence: ah actually I believe that file is just for build/generate and then it gets tied into the copilot-instructions so its the equiv of this file for validate

maririos: Ohhh yeah I think you are right

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
l0lawrence: prompts the user to give it, then if in a specific lang repo it defaults to that package naming structure and tries to find the path for that language, if a different language other than the repo you are in is requested it tries to map to it which ends up in it trying to discover the pkg path  
image
I'll swap it over sameeksha it ends up thinking more to end up at the same point if you want an alternate lang

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,23 @@
...
+---
+## Goal
+Check the validation checks for the SDK package by collecting the required information from the user and executing the azsdk_package_run_check command.
```
benbp:
```suggestion
Check the validation checks for the SDK package by collecting the required information from the user and executing the `azsdk_package_run_check` tool.
```

benbp (approved):

timovv (approved):

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,22 @@
+---
+description: 'Check if the package validation checks have passed for the SDK package.'
+---
+## Goal
+Check the validation checks for the SDK package by collecting the required information from the user and executing the `azsdk_package_run_check` tool.
```
praveenkuttappan:
```suggestion
Run validation checks for the SDK package by collecting the required information from the user and executing the `azsdk_package_run_check` tool.
```

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,22 @@
+---
+description: 'Check if the package validation checks have passed for the SDK package.'
```
praveenkuttappan: Is it better to say ` Run package validation for the SDK package`

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,22 @@
...
+---
+## Goal
+Check the validation checks for the SDK package by collecting the required information from the user and executing the `azsdk_package_run_check` tool.
+
+## Instructions
+1. **Collect Required Information**:
+    - Prompt the user for the absolute package path to the package directory for the SDK package they want to check.
```
praveenkuttappan, line 9: Isn't this tool triggered for the package in current repo in context? Why does this need to be absolute path? We should check this can be simplified to take just package name or relative path to package in one of the future enhancements.

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,22 @@
...
+    - Prompt the user specify values for each of the parameters required by the azsdk_package_run_check tool. Present the user with options to pick from for the allowed values specified by the parameter schema.
```
praveenkuttappan:
```suggestion
    - Prompt the user to specify values for each of the parameters required by the azsdk_package_run_check tool. Present the user with options to pick from for the allowed values specified by the parameter schema.
```

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
l0lawrence: Yes it can pick that up, absolute path in my mind was helpful to specify if the user wants to validate different languages from different repos (so in python repo but they validate java, or in rest-spec but validate python). I can look into changing the input to relative path if we decide those scenarios aren't super likely or if we want to use --language in association

timovv (approved):

praveenkuttappan (approved):

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
```diff
@@ -0,0 +1,22 @@
+---
+description: 'Run the package validation checks for the SDK package.'
+---
+## Goal
+Run the validation checks for the SDK package by collecting the required information from the user and executing the `azsdk_package_run_check` tool.
+
+## Instructions
+1. **Collect Required Information**:
+    - Prompt the user for the absolute package path to the package directory for the SDK package they want to check.
+    - Prompt the user to specify values for each of the parameters required by the azsdk_package_run_check tool. Present the user with options to pick from for the allowed values specified by the parameter schema.
```
weshaggard, line 10: What are the parameters? Are we expecting the agent to query the tool for those parameters?

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
weshaggard: Ideally the user would just need to give the package name and the agent would be able to find the path needed for the tool to run.

eng/common/instructions/azsdk-tools/check-package-validation.instructions.md
l0lawrence: yes - re https://github.com/Azure/azure-sdk-tools/pull/12531#discussion_r2456538651

it seems to work well from what I've seen->

image

weshaggard: Thanks. If we have the context of the repo I assume could the user not just give the package name instead of the full path to the package? Also are we intentionally scoping to a single package vs say a service directory?

l0lawrence: yes originally I had packageName but I swapped to path after this discussion and playing around with the idea of wanting to validate for multiple languages from one copilot session (one repo), as getting the path if you are say in the rest-api-specs repo from a packageName ended up asking the user for the path in the end anyway -- https://github.com/Azure/azure-sdk-tools/pull/12531#discussion_r2474168854. Yes this is just for a single package in the milestone 1 scenario

weshaggard: Thanks. We might want to keep experimenting to see if there are ways to make this easier beyond making the user copy and paste the full package path.

praveenkuttappan (approved):

PR #12534, opened 2025-10-16
Provide Mocked MCP tool answers

jeo02:
- Manually sending the tool responses instead of letting it happen automatically. 
- This lets us provide "mocked" responses. Planning on also creating a set of mocked responses to fall back on if the MCP Tool response was not provided.
- Turned off auto function invocation.

### DIFF
--- a/tools/ai-evals/azsdk-mcp/Scenarios/AzsdkTypeSpecGeneration_Step02_TypespecValidation.cs
+++ b/tools/ai-evals/azsdk-mcp/Scenarios/AzsdkTypeSpecGeneration_Step02_TypespecValidation.cs
@@ -38,7 +38,7 @@ namespace Azure.Sdk.Tools.McpEvals.Scenarios
 
             // Pass the expected outcome through the additional context. 
             var additionalContext = new ExpectedToolInputEvaluatorContext(json.ExpectedOutcome, s_toolNames);
-            var result = await expectedToolInputEvaluator.EvaluateAsync(fullChat, response, additionalContext: [additionalContext]);
+            var result = await scenarioRun.EvaluateAsync(fullChat, response, additionalContext: [additionalContext]);
 
             // 4. Assert the results
             EvaluationRating[] expectedRatings = [EvaluationRating.Good, EvaluationRating.Exceptional];


tools/ai-evals/azsdk-mcp/Helpers/ChatCompletion.cs
```diff
@@ -54,6 +54,75 @@ public async Task<ChatResponse> GetChatResponseAsync(IEnumerable<ChatMessage> ch
...
+                // several AIContent types.
+                var message = response.Messages.First();
+
+                conversationMessages.Add(message);
+
+                var functionCalls = message.Contents.OfType<FunctionCallContent>();
+                foreach (var functionCall in functionCalls)
+                {
+                    // Use the expected tool result if we have it.
+                    if (expectedToolResults.TryGetValue(functionCall.Name, out var expectedToolResult))
```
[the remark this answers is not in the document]
jeo02: Yea cause the function call has the normalized name.

tools/ai-evals/azsdk-mcp/Helpers/ChatCompletion.cs
```diff
@@ -54,6 +54,75 @@ public async Task<ChatResponse> GetChatResponseAsync(IEnumerable<ChatMessage> ch
...
+
+            var conversationMessages = chat.ToList();
+
+            var chatOptions = new ChatOptions
+            {
+                Tools = [.. tools]
+            };
+
+            var maxToolCalls = expectedToolResults.Count;
+
```
praveenkuttappan: nit: Remove these empty lines between vars

tools/ai-evals/azsdk-mcp/Helpers/ChatCompletion.cs
```diff
@@ -54,6 +54,75 @@ public async Task<ChatResponse> GetChatResponseAsync(IEnumerable<ChatMessage> ch
...
+            var toolsCalled = 0;
+
+            var chatInitialIndex = conversationMessages.Count;
+
+            while (response.FinishReason == ChatFinishReason.ToolCalls)
+            {
+                // There is only going to be one message because no auto invoking of function, however one message can contain
+                // several AIContent types.
+                var message = response.Messages.First();
```
praveenkuttappan: Is it always guaranteed to have a message or is it possible to be empty response? If empty is possible then we should use FirstOrDefault() and handle the null when message is referred.

tools/ai-evals/azsdk-mcp/Helpers/ChatCompletion.cs
```diff
@@ -54,6 +54,75 @@ public async Task<ChatResponse> GetChatResponseAsync(IEnumerable<ChatMessage> ch
...
+
+                        conversationMessages.Add(errorResponseMessage);
+                    }
+                    
+                    toolsCalled++;
+                }
+
+                if(toolsCalled >= maxToolCalls)
+                {
+                    break;
```
praveenkuttappan: Shouldn't we report if tools count is more than expected tool count?

tools/ai-evals/azsdk-mcp/Helpers/TestSetup.cs
```diff
@@ -43,7 +43,7 @@ public static IChatClient GetChatClient()
         {
             var azureClient = GetAzureOpenAIClient();
             return new ChatClientBuilder(azureClient.GetChatClient(AzureOpenAIModelDeploymentName).AsIChatClient())
-                .UseFunctionInvocation()
+                //.UseFunctionInvocation()
```
praveenkuttappan: nit: Remove if not required

tools/ai-evals/azsdk-mcp/Helpers/ChatCompletion.cs
jeo02: I'm going to be testing the new instructions, but the problem was the lack of natural stopping points. The LLM would want to continue calling the next tool until the instruction set is done. 

I'll test with the new instruction set and see how it works.

tools/ai-evals/azsdk-mcp/Helpers/ChatCompletion.cs
jeo02: yea I can do FirstOrDefault I'm not sure if its guaranteed, but the condition is if there's a tool call continue so there should be a message. But better safe than sorry.

PR #12552, opened 2025-10-17
[AVC] Create GithubManager and workflow to create parser-related Github issues

tjprescott:
Creates a `GithubManager` class that uses a Github App to interface with Github.

Create an `@mention` workflow to create a Github issue for parser-related APIView bugs. 

Refactors `MentionWorkflow` into an abstract base class to make it easier and more maintainable to add future mention workflows.

**TODO**
- [x] Update the target repo from `tjprescott/azure-sdk-tools` to `Azure/azure-sdk-tools`
- [x] Simplify the summarize input payload for the create github action workflow

### DIFF
--- a/packages/python-packages/apiview-copilot/src/mention/_open_parser_issue_workflow.py
+++ b/packages/python-packages/apiview-copilot/src/mention/_open_parser_issue_workflow.py
@@ -1,7 +1,9 @@
 import json
+
 import prompty
 from src._github_manager import GithubManager
 from src._utils import get_prompt_path
+
 from ._base import MentionWorkflow
 
 
@@ -9,8 +11,8 @@ class OpenParserIssueWorkflow(MentionWorkflow):
     prompty_filename = "parse_conversation_to_github_issue.prompty"
     summarize_prompt_file = "summarize_github_actions.prompty"
     deduplication_prompt_file = "deduplicate_parser_issue.prompty"
-    
-    REPO_OWNER = "tjprescott"
+
+    REPO_OWNER = "Azure"
     REPO_NAME = "azure-sdk-tools"
 
     LANGUAGE_LABELS = {
@@ -37,13 +39,13 @@ class OpenParserIssueWorkflow(MentionWorkflow):
         """Execute the parser issue workflow"""
         recent_issues = self._fetch_recent_parser_issues()
         dedup_result = self._check_for_duplicate_issue(plan, recent_issues)
-        
+
         if dedup_result["action"] == "no-op":
             # in recent_issues, find the issue with the matching number
             issue_number = dedup_result["issue"]
             for issue in recent_issues:
                 if issue["number"] == issue_number:
-                     return {
+                    return {
                         "action": "no-op",
                         "url": issue["html_url"],
                         "title": issue["title"],
@@ -51,7 +53,7 @@ class OpenParserIssueWorkflow(MentionWorkflow):
                         "created_at": issue["created_at"],
                     }
             raise ValueError(f"Issue number {issue_number} from deduplication prompt not found in recent issues.")
-        
+
         issue = self._create_parser_issue(plan)
         return {
             "action": "create",
@@ -66,9 +68,7 @@ class OpenParserIssueWorkflow(MentionWorkflow):
         client = GithubManager.get_instance()
         metadata_query = ["workflow: parser-issue", "source: APIView Copilot"]
         issues = client.search_issues(
-            owner=self.REPO_OWNER,
-            repo=self.REPO_NAME,
-            query=f'{" ".join(metadata_query)} in:body is:issue state:open'
+            owner=self.REPO_OWNER, repo=self.REPO_NAME, query=f'{" ".join(metadata_query)} in:body is:issue state:open'
         )
         return issues
 
@@ -81,10 +81,10 @@ class OpenParserIssueWorkflow(MentionWorkflow):
             "package_name": self.package_name,
             "code": self.code,
             "error_context": error_context,
-            "existing_issues": self._format_issues_for_dedup(recent_issues)
+            "existing_issues": self._format_issues_for_dedup(recent_issues),
         }
         raw_dedup = prompty.execute(dedup_prompt_path, inputs=dedup_inputs)
-        
+
         try:
             return json.loads(raw_dedup)
         except json.JSONDecodeError:
@@ -98,7 +98,7 @@ class OpenParserIssueWorkflow(MentionWorkflow):
                 "number": issue["number"],
                 "title": issue["title"],
                 "body": issue.get("body", "")[:500],  # truncate
-                "created_at": issue["created_at"]
+                "created_at": issue["created_at"],
             }
             for issue in issues
         ]
@@ -110,13 +110,9 @@ class OpenParserIssueWorkflow(MentionWorkflow):
         body = self._add_metadata_to_body(plan.get("body"))
         labels = self._build_issue_labels()
         return client.create_issue(
-            owner=self.REPO_OWNER,
-            repo=self.REPO_NAME,
-            title=plan.get("title"),
-            body=body,
-            labels=labels
+            owner=self.REPO_OWNER, repo=self.REPO_NAME, title=plan.get("title"), body=body, labels=labels
         )
-    
+
     def _add_metadata_to_body(self, body: str) -> str:
         """Inject parser issue metadata into issue body."""
         return f"""<!-- workflow: parser-issue -->
@@ -132,4 +128,3 @@ class OpenParserIssueWorkflow(MentionWorkflow):
         if language_label:
             labels.append(language_label)
         return labels
-


packages/python-packages/apiview-copilot/src/_github_manager.py
```diff
@@ -0,0 +1,273 @@
...
+        Returns a singleton instance of GithubManager.
+        If force_new is True, creates a new instance.
+        """
+        if cls._instance is None or force_new:
+            cls._instance = cls(use_app_auth=use_app_auth)
+        return cls._instance
+
+    def __init__(self, use_app_auth: bool = True):
+        settings = SettingsManager()
+        self._pat = settings.get("azuresdk-github-pat")
```
weshaggard: I assume this shouldn't be needed any more.

packages/python-packages/apiview-copilot/src/_github_manager.py
```diff
@@ -0,0 +1,273 @@
...
+        self._client = httpx.Client(
+            timeout=self._timeout,
+            headers={
+                "Accept": "application/vnd.github+json",
+                "User-Agent": self._user_agent,
+            },
+        )
+
+        self._app_id = settings.get("github_app_id")
+        self._client_secret = settings.get("github_app_client_secret")
```
weshaggard: What is the client secret here?

packages/python-packages/apiview-copilot/src/_github_manager.py
```diff
@@ -0,0 +1,273 @@
...
+    def _create_app_jwt(self) -> str:
+        """
+        Mint a GitHub App JWT using Azure Key Vault 'sign' (non-exportable key).
+        """
+        settings = SettingsManager()
+        keyvault_url = settings.get("github_app_keyvault_url")
+        key_name = settings.get("github_app_key_name")
+        app_id = settings.get("github_app_id")
+
+        credential = DefaultAzureCredential()
```
weshaggard: Do you want to use DAC here? I would suggest a more specific set of credentials.

packages/python-packages/apiview-copilot/scripts/decode_qr_code.py
```diff
@@ -0,0 +1,23 @@
+import base64
+import sys
+
+
+def main():
```
tjprescott: This is no longer necessary

packages/python-packages/apiview-copilot/src/_github_manager.py
tjprescott: Vestiges. I removed it.

packages/python-packages/apiview-copilot/src/_github_manager.py
```diff
@@ -0,0 +1,279 @@
...
+        """
+        headers = {"Authorization": f"token {token}"}
+        last_exc = None
+        for attempt in range(1, self._max_retries + 1):
+            try:
+                resp = self._client.request(method, url, headers=headers, json=json)
+                if resp.status_code in (429, 502, 503, 504) or (
+                    resp.status_code == 403 and "secondary rate limit" in resp.text.lower()
+                ):
+                    raise httpx.HTTPStatusError("Retryable", request=resp.request, response=resp)
```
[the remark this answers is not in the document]
tjprescott: Will go with this for now and update if it becomes an issue.

packages/python-packages/apiview-copilot/src/_github_manager.py
tjprescott: I updated it to use our `get_credential` method, but under the hood this is just DefaultAzureCredential. What more specific set of credentials do you suggest?

packages/python-packages/apiview-copilot/src/_mention.py
```diff
@@ -51,10 +52,54 @@ def handle_mention_request(*, comments: list[str], language: str, package_name:
...
+        results = json.loads(raw_results)
+        return results
+    except json.JSONDecodeError:
+        return raw_results
+
+
+def _post_github_issue(*, plan: dict):
+    client = GithubManager.get_instance()
+    issue = client.create_issue(
+        owner="tjprescott", repo="azure-sdk-tools", title=plan.get("title"), body=plan.get("body")
```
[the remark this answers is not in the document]
tjprescott: No, but this will need to be updated to be "Azure"

packages/python-packages/apiview-copilot/scripts/decode_qr_code.py
tjprescott: Removed.

packages/python-packages/apiview-copilot/src/_github_manager.py
```diff
@@ -0,0 +1,256 @@
...
+            "Authorization": f"Bearer {app_jwt}",
+            "Accept": "application/vnd.github+json",
+            "User-Agent": self._user_agent,
+            "X-GitHub-Api-Version": "2022-11-28",
+        }
+        resp = self._client.post(url, headers=headers, json={})
+        resp.raise_for_status()
+        return resp.json()
+
+    def _get_installation_token(self, owner: str) -> str:
```
marcarro, line 292: If `self._installation_tokens` is cache, you could use it here. Check if the owner is in the dict, if it is, return the token if not expired. If it **is** expired, remove that entry and create a new one

packages/python-packages/apiview-copilot/src/_github_manager.py
weshaggard: In general you should prefer a specific chained credential and only include the credential types that you expect to use. The biggest issue with DAC is that it probes too many options and you often end up using a credential you don't want to use.

marcarro:
More of a general question: what would happen if the same parser bug gets reported multiple times?

packages/python-packages/apiview-copilot/src/mention/_open_parser_issue_workflow.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class OpenParserIssueWorkflow(MentionWorkflow):
+    prompty_filename = "parse_conversation_to_github_issue.prompty"
+    summarize_prompt_file = "summarize_github_actions.prompty"
+
+    def execute_plan(self, plan: dict):
+        client = GithubManager.get_instance()
+        issue = client.create_issue(
+            owner="tjprescott", repo="azure-sdk-tools", title=plan.get("title"), body=plan.get("body")
```
AlitzelMendez: does this mean that we are opening all the issues in your fork?

tjprescott: While I'm testing, yes. I will switch it before merging.

packages/python-packages/apiview-copilot/src/mention/_open_parser_issue_workflow.py
```diff
@@ -1,15 +1,84 @@
...
 
     def execute_plan(self, plan: dict):
+        """Execute the parser issue workflow"""
+        recent_issues = self._fetch_recent_parser_issues()
+        dedup_result = self._check_for_duplicate_issue(plan, recent_issues)
+        
+        if dedup_result["action"] == "no-op":
+            return {
+                "action": "no-op",
+                "existing_issue": dedup_result["existing"],
```
marcarro: `existing_issue` should be just `issue` for consistency with below return

packages/python-packages/apiview-copilot/src/_github_manager.py
tjprescott: I replaced it with the same ChainedTokenCredential Ali used in APIView.

packages/python-packages/apiview-copilot/evals/tests/mention_summarize/open_parser_issue.yaml
```diff
@@ -0,0 +1,11 @@
+testcase: mention_summarize_open_parser_issue
```
tjprescott, line 1: @marcarro please add a summarize scenario where an issue already exists. In this case, the summarize step should respond something like "Thank you for your feedback! It looks like there's already an existing issue for this: <link>".

packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -86,6 +101,20 @@ def _filter_existing_comment(testcase: str, response: str, language: str, existi
...
 
+def _deduplicate_parser_issue(testcase: str, response: str, language: str, package_name: str, code: str, error_context: str, existing_issues: str):
+    prompty_path = Path(__file__).parent.parent / "prompts" / "mention" / "deduplicate_parser_issue.prompty"
+    prompty_kwargs = {
+        "testcase": testcase,
+        "response": response,
+        "language": language,
+        "package_name": package_name,
+        "code": code,
+        "error_context": error_context,
```
tjprescott: This is more "issue_context" than error, per se.

packages/python-packages/apiview-copilot/prompts/mention/deduplicate_parser_issue.prompty
```diff
@@ -0,0 +1,120 @@
...
+# Guidelines
+
+**Duplicates** have the same ROOT CAUSE even if the specific code differs:
+- "Literal[foo]" vs "Literal[bar]" = DUPLICATE (both unquoted literals)
+- "missing quotes in Literal" vs "bare string in type hint" = DUPLICATE (same underlying issue)
+- "parser crashes on Union types" vs "parser crashes on Optional types" = MIGHT BE DUPLICATE (check if same parse failure mode)
+
+**Not duplicates** have different root causes:
+- "unquoted Literal" vs "deeply nested generics" = NOT DUPLICATE (different parser failures)
+- "missing import statement" vs "incorrect type annotation" = NOT DUPLICATE (different error types)
```
tjprescott: You should use a completely different example in the prompt relative to the evals (or the sample) because otherwise you are just special casing your tests.

packages/python-packages/apiview-copilot/prompts/mention/deduplicate_parser_issue.prompty
```diff
@@ -0,0 +1,120 @@
...
+
+- You will be given a NEW parser issue with its language, package, code snippet, and error context.
+- You will also be given a list of EXISTING recent parser issues.
+- Your job is to determine if the NEW issue is a duplicate of ANY existing issue.
+- Focus on the UNDERLYING BUG, not the specific package or code snippet. Multiple different code examples can have the same root cause.
+- Your response MUST be valid json (DO NOT include the ```json at the start or ``` at the end of your response!):
+```json
+{
+  "action": "<no-op|create>",
+  "existing": issue_number_or_null,
```
tjprescott: I don't see any guidance provided on this. It should basically reflect, if not null, what is the *best* existing issue, as this will be important to the summary.

packages/python-packages/apiview-copilot/prompts/mention/summarize_github_actions.prompty
```diff
@@ -0,0 +1,51 @@
...
+      "created_at": "2025-10-15T19:36:45Z",
+      "body": "**Bad code snippet:**\n```python\nfoo: Literal[Foo]\n```\n\n**Description:**\nThe APIView parser is incorrectly displaying literal values in type annotations without quotes. According to Python typing conventions, literal values should be quoted, e.g., `Literal[\"Foo\"]`.\n\n**Expected (good) code snippet:**\n```python\nfoo: Literal[\"Foo\"]\n```\n\nPlease update the parser to ensure literal values are properly quoted in the APIView output.",
+    }
+---
+system:
+You are an AI that summarizes the results of an @mention request from APIView. This request required taking some action on Github.
+
+# INSTRUCTIONS
+
+## NO FAILURES
```
tjprescott, line 38: Summary also need to understand whether a _new_ issue was created or an existing issue was found for this. Need an eval case for this as well.

packages/python-packages/apiview-copilot/src/mention/_open_parser_issue_workflow.py
```diff
@@ -0,0 +1,84 @@
...
+            "issue": issue
+        }
+
+    def _fetch_recent_parser_issues(self):
+        """Fetch recent open parser issues from GitHub."""
+        client = GithubManager.get_instance()
+        issues = client.search_issues(
+            owner=self.REPO_OWNER,
+            repo=self.REPO_NAME,
+            query="label:parser is:issue state:open"
```
tjprescott: We can't query on label because we cannot create new labels (even if it lets you--don't!). You could query for "APIView parser" or something. Possibly also azure-sdk-automation as the author, though that would ignore issues posted by humans (which I'd advise against).

packages/python-packages/apiview-copilot/src/mention/_open_parser_issue_workflow.py
```diff
@@ -0,0 +1,84 @@
...
+            "package_name": self.package_name,
+            "code": self.code,
+            "error_context": error_context,
+            "existing_issues": self._format_issues_for_dedup(recent_issues)
+        }
+        raw_dedup = prompty.execute(dedup_prompt_path, inputs=dedup_inputs)
+        
+        try:
+            return json.loads(raw_dedup)
+        except json.JSONDecodeError:
```
tjprescott, line 90: If the output is in a bad format, you should error out. You can also force prompty to use a json schema so it can't return bad JSON.

packages/python-packages/apiview-copilot/src/mention/_open_parser_issue_workflow.py
```diff
@@ -0,0 +1,84 @@
...
+            # Default to creating new issue if we can't parse the response
+            return {"action": "create", "existing": None}
+
+    def _format_issues_for_dedup(self, issues: list) -> str:
+        """Format issues for deduplication prompt input."""
+        formatted_issues = [
+            {
+                "number": issue["number"],
+                "title": issue["title"],
+                "body": issue.get("body", "")[:500],  # truncate
```
tjprescott, line 100: Arguably if there's a long issue we should summarize it rather than truncate, but this is probably good enough for now.

packages/python-packages/apiview-copilot/src/mention/_open_parser_issue_workflow.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def _create_parser_issue(self, plan: dict):
+        """Create a new parser issue on GitHub."""
+        client = GithubManager.get_instance()
+        return client.create_issue(
+            owner=self.REPO_OWNER,
+            repo=self.REPO_NAME,
+            title=plan.get("title"),
+            body=plan.get("body"),
+            labels=["parser"]
```
tjprescott: We cannot create new labels. EngSys owns those and will not be happy with copilot creating new labels. Any key words need to be embedded in the title (most reliable) and/or the body.

tjprescott (commented):
A handful of comments. One red flag--we cannot create new tags.

packages/python-packages/apiview-copilot/prompts/mention/deduplicate_parser_issue.prompty
```diff
@@ -52,14 +52,28 @@ You are an agent that detects duplicate parser issues for apiview.
 - You will also be given a list of EXISTING recent parser issues.
 - Your job is to determine if the NEW issue is a duplicate of ANY existing issue.
 - Focus on the UNDERLYING BUG, not the specific package or code snippet. Multiple different code examples can have the same root cause.
+- If the NEW issue is a duplicate of any existing issue, respond with action "no-op" and include the existing issue's payload INCLUDING its URL under a 'url' field verbatim.
```
tjprescott: Since you already retrieve the similar issues from Github, my recommendation here would be to simple get the "most similar issue number" and then get the actual payload from what you already pulled. I've found that asking the LLM to pass through data verbatim is often unreliable, and it wastes tokens since you already have the data in memory.

marcarro:
Before merging, remember to switch away from the fork on the issue posting/search procedures

PR #12556, opened 2025-10-17
[AVC] Use cached results for evaluations

### LINKED ISSUE #12403: [AVC] Support running evals with cached results
When writing test cases, it would be very helpful to be able to run the test case with cached API calls. For example, if you need to debug into the output code for your test, having to re-run the whole test and make API calls is slow, costly and unnecessary. It would be helpful if you could used a cached result so that instead of making a live call to an LLM, it simply returns the cached value. This would allow you to quickly reach the output code.

marcarro:
Use cached results for evaluations instead of making LLM calls every time. Useful for debugging.

Closes #12403


packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -19,8 +20,72 @@
 from src._settings import SettingsManager
 from src._utils import get_prompt_path
 
+def get_cached_result(workflow_name: str, testcase: str) -> dict | None:
+    """Get cached result for a specific testcase in a workflow."""
+    cache_file = Path(__file__).parent / "tests/cache" / f"{workflow_name}.jsonl"
+    if not cache_file.exists():
+        return None
```
tjprescott: Yeah, this isn't going to work if you target individual test cases, and it should.

packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -19,8 +20,72 @@
...
+def save_cached_result(workflow_name: str, testcase: str, result: dict):
+    """Save result to cache for a specific testcase."""
+    cache_dir = Path(__file__).parent / "tests/cache"
+    cache_dir.mkdir(exist_ok=True)
+    
+    cache_file = cache_dir / f"{workflow_name}.jsonl"
+    try:
+        cache_entry = {"testcase": testcase, "actual": result["actual"]}
+        with open(cache_file, 'a', encoding='utf-8') as f:
+            f.write(json.dumps(cache_entry, ensure_ascii=False) + '\n')
```
tjprescott: Here I think we need to take the results that are in JSONL format and split them into individual JSON files that can be cached individually. The structure of the `tests/cache` folder will ultimately mirror that of the `tests` folder.

packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -19,8 +20,72 @@
...
+    except (IOError, KeyError):
+        # Failed to cache - just continue
+        pass
 
-def _review_apiview(query: str, language: str):
+def cached_target_function(workflow_name: str):
```
tjprescott: How does this work if you are targeting (and caching) a single test case?

packages/python-packages/apiview-copilot/evals/_custom.py
```diff
@@ -19,8 +20,72 @@
...
+    """Decorator to add caching to target functions."""
+    _use_cache = os.environ.get('EVALS_USE_CACHE', 'false').lower() == 'true'
```
tjprescott: Why an ENV var? Could this be a parameter on `cached_target_function` (which seems to be acting like a decorator factory)?

packages/python-packages/apiview-copilot/evals/_custom.py
marcarro: On the first run of an un-cached individual testcase with `--use-cache`, a cache file is created with the name of the workflow parent of the testcase. On the second run of the testcase, it looks for the workflow parent name and hits the cache. It adds testcases to each workflow jsonl file as needed.

packages/python-packages/apiview-copilot/evals/_custom.py
marcarro: What I found throughout testing was that approaches like having module level variables or even function attributes (something like `target_function.use_cache = self.use_cache`) consistently didn't work. I inspected this and the target function that is instantiated vs the target function run by `evaluate()` have different ID's, so maybe the evaluation library is making a copy of the target function, and somehow the attributes are lost between the threads. Not sure. This is the workaround I came up with but open to a better approach.

packages/python-packages/apiview-copilot/evals/_custom.py
tjprescott: How well does this work if you have nested folders? It seems more natural for the cache folder structure to simply mirror the test folder structure rather than mapping a folder name to a file name.

```
foo
-> bar
```

Becomes `foo.jsonl` and `foo_bar.jsonl`? But this becomes fraught with difficulty if you have

```
foo_bar
-> bar
```

Because now `foo_bar_bar.jsonl` becomes difficult (or impossible) to parse.

packages/python-packages/apiview-copilot/evals/_custom.py
tjprescott: `def cached_target_function(workflow_name: str, use_cache: bool = false):`

Assuming you've tried this and it didn't work?

marcarro: Nope, I'm not exactly sure how it would work.
```python
USE_CACHE = False

@cached_target_function("mention_action", use_cache=USE_CACHE)  # Gets decorated with 'False'?
def _mention_action_workflow(...):
```

Don't really know the specifics of how decorator arguments are set or if they can be modified at runtime. Anyhow, I'm working on updating the caching logic and I dropped the use of an env variable + decorators altogether 😄 should be up soon.

PR #12557, opened 2025-10-17
added knowledge base and removed unnecessary Result Pattern Useage

radhgupta:
### 1.  Added Knowledge Base for Stronger Context
- **Added `KnowledgeBaseService`** to load and manage TypeSpec-specific knowledge from `knowledge.md`
- **Enhanced `OpenAIService`** to combine agent instructions with knowledge base content in system messages
- **Created comprehensive `knowledge.md`** with detailed TypeSpec customization guidance, common patterns

### 2. Removed Result Pattern Where Not Required
- **Simplified return types** in methods where early exit was more appropriate than error propagation
- **Removed Result wrapper** from `TypeSpecPatchApplicator.ApplyPatchAsync()` - now returns `Task` and throws exceptions for immediate error handling
- **Streamlined error flow** to exit early on critical failures instead of continuing with wrapped errors

### 3. Consolidated Agent Response Parser
- **Created unified `AgentResponseParser`** static class with centralized JSON parsing logic
- **Removed duplicate response parsing code** from multiple services
- **Added robust JSON cleaning** with support for markdown code blocks, mixed content, and various JSON formats

### 4. Removed Complicated Custom Exception
- **Simplified exception handling** by removing overly complex custom exception types that weren't providing value

### DIFF
--- a/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ToolExecutor.cs
+++ b/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ToolExecutor.cs
@@ -56,26 +56,29 @@ internal class ToolExecutor
 
     private async Task<string> ExecuteGetTypeSpecFileAsync(ValidationContext validationContext, string argumentsJson, CancellationToken cancellationToken)
     {
-        if (string.IsNullOrEmpty(argumentsJson))
+        if (!string.IsNullOrEmpty(argumentsJson))
         {
-            argumentsJson = "{}"; // Default to empty object
-        }
-        
-        // Parse arguments to get the filename
-        using var args = JsonSerializer.Deserialize<JsonDocument>(argumentsJson);
-        if (args?.RootElement.TryGetProperty("path", out var pathElement) == true)
-        {
-            var filename = pathElement.GetString();
-            if (string.IsNullOrEmpty(filename))
+
+            // Parse arguments to get the filename
+            using var args = JsonSerializer.Deserialize<JsonDocument>(argumentsJson);
+            if (args?.RootElement.TryGetProperty("path", out var pathElement) == true)
             {
-                throw new ArgumentException("Missing or empty 'path' in arguments");
+                var filename = pathElement.GetString();
+                if (string.IsNullOrEmpty(filename))
+                {
+                    throw new ArgumentException("Missing or empty 'path' in arguments");
+                }
+
+                var result = await ToolHandler.GetTypeSpecFileAsync(filename, validationContext, cancellationToken);
+                return JsonSerializer.Serialize(result);
             }
-            
-            var result = await ToolHandler.GetTypeSpecFileAsync(filename, validationContext, cancellationToken);
-            return JsonSerializer.Serialize(result);
+
+            throw new ArgumentException("Missing 'path' property in arguments");
         }
-        
-        throw new ArgumentException("Missing 'path' property in arguments");
+        else
+        {
+            return "{}";
+        }
     }
 
     private static string CreateErrorResponse(string errorMessage)


src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentResponseParser.cs
```diff
@@ -22,10 +22,12 @@ public static IEnumerable<RuleError> ParseErrors(string rawResponse)
         {
             try
             {
-                // Extract JSON from markdown code blocks if present
-                var jsonContent = ExtractJsonFromResponse(rawResponse);
+                if (string.IsNullOrWhiteSpace(rawResponse))
+                {
+                    throw new InvalidOperationException("Agent response is Null or empty");
+                }
```
m-redding: nit - you check if response is null/empty in CleanJsonResponse as well but just return instead of throwing. Is that because you want to throw in this case and not others? Otherwise I think you could probably only check in one place

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentResponseParser.cs
```diff
@@ -39,71 +41,123 @@ public static IEnumerable<RuleError> ParseErrors(string rawResponse)
                     return Enumerable.Empty<RuleError>();
                 }
 
-                return response.Errors.Select(e => new RuleError(e.Type, e.Message));
+                return response.Errors.Select(e => new RuleError(
+                    string.IsNullOrWhiteSpace(e.Type) ? "UnspecifiedError" : e.Type, 
+                    string.IsNullOrWhiteSpace(e.Message) ? "No message provided" : e.Message));
             }
-            catch (JsonException ex)
+            catch (Exception ex)
```
m-redding: Curious why change this to Exception? If it's not a JSON exception I wonder if it might make more sense to just let it bubble up on its own

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentResponseParser.cs
```diff
@@ -39,71 +41,123 @@ public static IEnumerable<RuleError> ParseErrors(string rawResponse)
...
+                }
+                return patchRequest;
             }
-            
-            // Check if response starts with ``` (without json specifier)
-            if (trimmed.StartsWith("```"))
+            catch (Exception ex)
             {
-                return ExtractFromMarkdownBlock(trimmed, "```");
+                throw new InvalidOperationException($"Failed to parse patch JSON: {ex.Message}", ex);
```
m-redding: Same sort of thing with this one, I'm curious do you feel like the exception is too cryptic as it is? So it's better to re-throw as the same one? If patchRequest is null then it seems like that failure wouldn't need to be rethrown

src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/LibraryBuildServiceTests.cs
```diff
@@ -0,0 +1,658 @@
...
+
+namespace Azure.Tools.GeneratorAgent.Tests
+{
+    [TestFixture]
+    internal class LibraryBuildServiceTests
+    {
+        private static Mock<ILogger<LibraryBuildService>> CreateMockLogger()
+        {
+            return new Mock<ILogger<LibraryBuildService>>();
+        }
```
m-redding, line 16: nit/non blocking since almost all the tests use this you could always pre-seed this in a [SetUp] method

src/dotnet/GeneratorAgent/tests/Azure.Tools.GeneratorAgent.Tests/LocalLibraryGenerationServiceTests.cs
```diff
@@ -17,282 +10,209 @@ namespace Azure.Tools.GeneratorAgent.Tests
...
-            using var fixture = new TestEnvironmentFixture();
-            
-            Assert.DoesNotThrow(() => fixture.CreateService());
+            // Arrange
+            var appSettings = CreateTestAppSettings();
+            var logger = NullLogger<LocalLibraryGenerationService>.Instance;
+            var processExecutionService = CreateMockProcessExecutionService().Object;
+
+            // Act & Assert
+            Assert.DoesNotThrow(() => new LocalLibraryGenerationService(appSettings, logger, processExecutionService));
```
m-redding, line 22: nit/nonblocking but I typically try to be selective about including tests that only test that something doesn't throw, sometimes it makes sense but other times it's just kind of like a smoke test. For constructors you'd probably see all the tests failing if the constructor starting throwing for just standard inputs

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/OpenAIService.cs
```diff
@@ -119,13 +123,33 @@ public async Task<string> GenerateFixesAsync(IList<Fix> fixes, ValidationContext
...
+        var systemMessage = AppSettings.AgentInstructions;
         
-        // Clean the response - remove markdown code blocks if present
-        var cleanedResult = CleanJsonResponse(result);
-        
-        Logger.LogDebug("OpenAI Response: {Response}", cleanedResult);
+        if (KnowledgeBaseService.IsKnowledgeBaseAvailable)
+        {
+            var knowledgeBase = KnowledgeBaseService.GetKnowledgeBase();
+            systemMessage = $"{systemMessage}\n\n## TypeSpec Knowledge Base\n\n{knowledgeBase}";
```
m-redding, line 142: Curious - what's the value of having a knowledge base abstracted as a separate concept vs. including it in the system prompt itself? Is it easier to extend/more explicit with the fact it's adding additional context?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ToolExecutor.cs
```diff
@@ -30,32 +27,40 @@ public ToolExecutor(ITypeSpecToolHandler toolHandler, ILogger<ToolExecutor> logg
...
-        // This tool doesn't need arguments
         var result = await ToolHandler.ListTypeSpecFilesAsync(validationContext, cancellationToken);
         return JsonSerializer.Serialize(result);
     }
 
     private async Task<string> ExecuteGetTypeSpecFileAsync(ValidationContext validationContext, string argumentsJson, CancellationToken cancellationToken)
     {
+        if (string.IsNullOrEmpty(argumentsJson))
+        {
+            argumentsJson = "{}"; // Default to empty object
```
m-redding: nit - could you set args to null or something instead of deserializing an empty object?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Tools/TypeSpecToolHandler.cs
```diff
@@ -53,12 +52,12 @@ public async Task<ListTypeSpecFilesResponse> ListTypeSpecFilesAsync(ValidationCo
                 };
 
                 Logger.LogInformation("Returning {Count} TypeSpec files with complete content for analysis", response.Files.Count);
+                
                 return response;
             }
             catch (Exception ex)
             {
-                Logger.LogError(ex, "Error handling list_typespec_files tool call");
```
m-redding, line 60: Why delete the logging from here? Is it logged/caught upstream?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
```diff
@@ -63,8 +63,8 @@ public AppSettings(IConfiguration configuration, ILogger<AppSettings> logger)
         public string? OpenAIEndpoint => Configuration.GetSection("OpenAI:Endpoint").Value;
         public string OpenAIModel => Configuration.GetSection("OpenAI:Model").Value ?? "gpt-4o";
         
-        public string TypespecEmitterPackage => "@typespec/http-client-csharp";
-        public string TypespecCompiler => "@typespec/compiler";
+        public string TypespecEmitterPackage => Configuration.GetSection("AzureSettings:TypespecEmitterPackage").Value ?? "@typespec/http-client-csharp";
+        public string TypespecCompiler => Configuration.GetSection("AzureSettings:TypespecCompiler").Value ?? "@typespec/compiler";
```
m-redding, line 67: (non-blocking / conversational) This made me think of an idea - maybe some future evolution could experiment with trying to get these from the tspconfig.yml and then falling back to a default or app settings value, maybe you could just steal the implementation from tsp client
Because it may even be a consideration to stick to tsp compile long-term so it's 3p friendly

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
```diff
@@ -72,14 +71,12 @@ public async Task<Dictionary<string, string>> GetTypeSpecFilesAsync(string commi
...
             {
                 string errorMessage = $"Failed to deserialize GitHub API response for '{AppSettings.AzureSpecRepository}/{typeSpecDir}' at commit '{commitId}'. Please verify the repository path and commit ID are correct.";
-                Logger.LogError(ex, errorMessage);
                 throw new InvalidOperationException(errorMessage, ex);
             }
 
             if (contents == null)
             {
                 string errorMessage = $"GitHub API returned null content for '{AppSettings.AzureSpecRepository}/{typeSpecDir}' at commit '{commitId}'";
-                Logger.LogError(errorMessage);
```
m-redding, line 82: Same with these, are the errors logged / flagged to the user upstream? Or did you find them too verbose or unhelpful for users?

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/knowledge.md
```diff
@@ -0,0 +1,171 @@
+# Agent Knowledge: TypeSpec Customization for Azure SDK
```
m-redding, line 1: (non-blocking / conversational) at some point when there's a more unified evals framework it'd be nice to have some evaluations to help make the decisions about what to / not to include here more easily data-driven

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/TypeSpecFileService.cs
```diff
@@ -260,7 +260,7 @@ public async Task<Result<bool>> UpdateTypeSpecFileAsync(string fileName, string
             }
             catch (Exception ex)
             {
-                return Result<bool>.Failure(new SecurityException($"TypeSpec filename validation failed: {fileName}", ex));
+               throw new SecurityException($"TypeSpec filename validation failed: {fileName}", ex);
```
m-redding: I know this is already there but I'm curious why SecurityException? It sounds like that's a permissions thing, which I'm not sure if it applies in this case

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorAnalysisService.cs
```diff
@@ -28,31 +28,39 @@ public ErrorAnalysisService(OpenAIService aiService, ILogger<ErrorAnalysisServic
...
         /// Analyzes errors from a compilation/build result using regex parsing with AI fallback for unparsed content.
         /// Note: This method assumes the result contains a ProcessException (recoverable build/compilation errors).
-        /// General exceptions (unrecoverable errors) should be handled by the caller before reaching this method.
         /// </summary>
-        public async Task<Result<IEnumerable<Fix>>> GenerateFixesFromResultAsync(string errorOutput, ValidationContext validationContext, CancellationToken cancellationToken)
+        public async Task<List<Fix>> GenerateFixesFromFailureLogsAsync(string errorOutput, CancellationToken cancellationToken)
         {
+            if (string.IsNullOrWhiteSpace(errorOutput))
+            {
+                throw new InvalidOperationException("No output available for analysis");
```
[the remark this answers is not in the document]
m-redding: I feel like I remember that this is checked earlier in the process? You may not even need to check it

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ProcessExecutionService.cs
```diff
@@ -71,25 +68,9 @@ public virtual async Task<Result<object>> ExecuteAsync(
...
-            catch (InvalidOperationException ex)
-            {
-                Logger.LogError(ex, "Invalid operation starting process: {Command}", command);
-                return Result<object>.Failure(ex);
-            }
             catch (Exception ex) when (ex is not OperationCanceledException)
             {
-                Logger.LogError(ex, "Unexpected error executing command: {Command}", command);
-                return Result<object>.Failure(ex);
+                throw new InvalidOperationException("Failed to execute process", ex);
```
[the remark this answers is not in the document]
radhgupta: This is unrecoverable exception therefore we directly throw rather than using Result Pattern

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/ErrorAnalysisService.cs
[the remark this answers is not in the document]
radhgupta: I think you are right; however, I was planning of writing the service in the sense that it could be used independently and handles all the edge cases! that is the reason why I wanted to add a null check here

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/OpenAIService.cs
radhgupta: Good question! The reason I wanted to have a seperate class was because of the following benefits:

- Content updates without code changes - just replace the .md file
- Graceful fallback - service works even if knowledge file is missing
- Clear separation - core AI instructions vs domain knowledge
- Future flexibility - easy to extend to multiple knowledge sources or dynamic loading

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ToolExecutor.cs
radhgupta: I set it to empty object because I thought if I set it to null the json deserializer would throw exception  which might be too wordy rather if we deserialize empty object, it will not have json proterpty "path" and we will get a argument exception "missing 'path' property in arguments from line 78.
I may be wrong please feel free to correct me :)

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Tools/TypeSpecToolHandler.cs
radhgupta: Yes, it logged upstream.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Configuration/AppSettings.cs
radhgupta: This sounds quite interesting; we can add it to the milestones.

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/GitHubFileService.cs
radhgupta: These error msgs are logged upstream in program.cs  where  we log errorMessage

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentResponseParser.cs
radhgupta: fixed

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/AgentResponseParser.cs
radhgupta: fixed

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/OpenAIService.cs
m-redding: Thanks for explaining, that makes sense to me

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/Agent/ToolExecutor.cs
m-redding: Ahh sorry my comment was unclear, I meant something like this if this might work because you're right that setting it to null would throw

```csharp
try{
    JsonDocument? args = null;
    if (!string.IsNullOrEmpty(argumentsJson))
    {
        args = // deserialize
    }
    // if args is null throw with no property path
}
finally
{
    args?.Dispose();
}
```

m-redding:
```suggestion
            throw new ArgumentException("Missing or empty 'path' in arguments");
```
Ahh ok my original comment was unclear, I mean doing `JsonDocument? args = null` and then only calling deserialize if `argumentsJson` was not null, but actually I thought about this more and what about something like this

m-redding: Or just wrapping the whole deserialize/parsing stuff inside of an `if (!string.IsNullOrEmpty(argumentsJson))`

radhgupta: Makes sense! thanks :)

src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/knowledge.md
maririos, line 1: @chrisradek for context

maririos: and @swathipil

Issue #12510, opened 2025-10-15
AzSDK MCP server: Update SDK details should validate package name


praveenkuttappan:
We have copilot instructions to identify the package name for each language. But copilot still sends invalid package names in request. For e.g. Go package name should always begin with `sdk/`

We should also tweak https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/instructions/azsdk-tools/sdk-details-in-release-plan.instructions.md to make sure copilot follows this guideline.

smw-ms:
Before changes: 
<img width="716" height="1248" alt="Image" src="https://github.com/user-attachments/assets/73f64222-7853-4a4c-9d92-b23438455196" />

smw-ms:
After PR: https://github.com/Azure/azure-sdk-tools/pull/12559 - 
<img width="731" height="1292" alt="Image" src="https://github.com/user-attachments/assets/49a85a71-60fe-45fb-b774-ed8d09f8db8a" />

smw-ms:
After commit: https://github.com/Azure/azure-sdk-tools/pull/12559/commits/085e3b20bf421712c08e8ebdd974584412206b9a
<img width="1498" height="1232" alt="Image" src="https://github.com/user-attachments/assets/2e9ae6da-8dd0-4a37-b4ce-0cd8d5ad3d7f" />

<img width="1527" height="1121" alt="Image" src="https://github.com/user-attachments/assets/5837a2ef-0058-48a3-893e-92004959fc0f" />

maririos:
@praveenkuttappan could you remind me why do we need this?

maririos:
The package name is evaluated in the spec repo and in tspconfig

PR #12559, opened 2025-10-18
Validate package names when adding to release plan

smw-ms:
Issue :  https://github.com/Azure/azure-sdk-tools/issues/12510


eng/common/instructions/azsdk-tools/sdk-details-in-release-plan.instructions.md
```diff
@@ -9,13 +9,19 @@ description: 'Identify languages configured in the TypeSpec project and add it t
    - For .NET, use namespace property to get package name.
    - For JavaScript, use `packagedetails:name` property to get package name.
    - For Go, use module name and remove `github.com/Azure/azure-sdk-for-go/` to get package name
-3. Map the language name in emitter to one of the following in Pascal case(except .NET):
+3. Validate package names against these prefix rules before proceeding:
```
benbp: This seems like it might be better as a tool call than an instructions doc

> Call `azsdk_package_validate_name` before proceeding

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -633,5 +643,36 @@ public async Task<DefaultCommandResponse> UpdateLanguageExclusionJustification(i
                 return new DefaultCommandResponse { ResponseError = $"Failed to update release plan with language exclusion justification: {ex.Message}" };
             }
         }
+
+        [McpServerTool(Name = "azsdk_package_validate_name"), Description("Validate package name based on language")]
+        public ValidationResponse ValidatePackageName(string language, string packageName)
```
maririos: We have logic on the spec repo that validates package names. I wonder if this would be an opportunity to copy/reutilize/update and be consistent in both places:
- Validation rules: https://github.com/Azure/azure-rest-api-specs/wiki/TypeSpec-Validation#rules
- Validation code (need to look for the specific properties rules): https://github.com/Azure/azure-rest-api-specs/blob/main/eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts

maririos (commented):
We should complement the validation rules. 
General information can be found here: https://eng.ms/docs/products/azure-developer-experience/develop/namespace-review?tabs=management

I have added a comment with more information about places where we are already doing package name validation

eng/common/instructions/azsdk-tools/sdk-details-in-release-plan.instructions.md
```diff
@@ -4,18 +4,19 @@ description: 'Identify languages configured in the TypeSpec project and add it t
...
 **Goal**: Identify languages configured in the TypeSpec project and generate the json object with language and package name.
 1. Identify the language emitter configuration in the `tspconfig.yaml` file in the TypeSpec project root.
-2. Identify the package name or namespace for each language emitter.
+2. Identify and display the package name or namespace for each language emitter.
    - For Java and Python, use `emitter-output-dir` for package name if it exists. Otherwise use `package-dir` to get the package name as fallback approach.
    - For .NET, use namespace property to get package name.
    - For JavaScript, use `packagedetails:name` property to get package name.
    - For Go, use module name and remove `github.com/Azure/azure-sdk-for-go/` to get package name
-3. Map the language name in emitter to one of the following in Pascal case(except .NET):
+3. Use `azsdk_package_validate_name` to validate each of the package names before proceeding.If any of the package names are invalid, correct the package name.
```
praveenkuttappan: We don't need this reference here in instruction. validate package name should not be a tool.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
praveenkuttappan: This shouldn't be a tool. We just need to validate the package name in request to update the SDK details in the release plan.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -633,5 +643,36 @@ public async Task<DefaultCommandResponse> UpdateLanguageExclusionJustification(i
...
+        {
+            try
+            {
+                var languagePrefixMap = new Dictionary<string, List<string>>
+                (StringComparer.OrdinalIgnoreCase)
+                {
+                    { ".NET", new List<string> { "Azure." } },
```
praveenkuttappan: Let's this one for  .NET. It's not always Azure.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -633,5 +643,36 @@ public async Task<DefaultCommandResponse> UpdateLanguageExclusionJustification(i
...
+                    { "JavaScript", new List<string> { "@azure/", "@azure-rest/" } },
+                    { "Java", new List<string> { "azure-", "com.azure." } },
```
praveenkuttappan: Same here. There are scopes other than com.azure.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -633,5 +643,36 @@ public async Task<DefaultCommandResponse> UpdateLanguageExclusionJustification(i
...
             }
         }
+
+        [McpServerTool(Name = "azsdk_package_validate_name"), Description("Validate package name based on language")]
+        public ValidationResponse ValidatePackageName(string language, string packageName)
+        {
+            try
+            {
+                var languagePrefixMap = new Dictionary<string, List<string>>
+                (StringComparer.OrdinalIgnoreCase)
```
praveenkuttappan: @maririos Is there a rule defined in TypeSpec validation? If V2 tspconfig.yaml has a stricter convention for package name, then we can remove these checks for all language. Main issue I am noticing is for go which does not have package name begins with `sdk/` when agent sends the request.

maririos: the validations for package names are part of typespec validation. I was actually wondering why we need this here

eng/common/instructions/azsdk-tools/sdk-details-in-release-plan.instructions.md
samvaity: +1 It can be an internal check before release, but don't think we need a tool for that?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -451,6 +451,16 @@ public async Task<DefaultCommandResponse> UpdateSDKDetailsInReleasePlan(int rele
                     return $"Unsupported SDK language found. Supported languages are: {string.Join(", ", supportedLanguages)}";
                 }
 
+                // Validate SDK Package names
+                foreach (var sdk in SdkInfos)
+                {
+                    var response = ValidatePackageName(sdk.Language, sdk.PackageName);
+                    if (response.IsValid != true)
+                    {
+                        return new DefaultCommandResponse { ResponseError = response.ResponseError, Message = response.Message};
```
benbp: Should this collect all possible errors before returning in case there are multiple? You can assign the list of errors to the `ResponseErrors` array in `DefaultCommandResponse`.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
benbp: We should also avoid hardcoding language strings and instead use the `SdkLanguage` enum in all places where we filter by language. This makes future changes easier, and also allows MCP tool schemas to enumerate valid input values.

praveenkuttappan (approved):

PR #12561, opened 2025-10-20
Update approve script to set correct login

weshaggard:
@benbp I wonder if you were hitting this as well but while debugging I found I was logged into TME which is why it failed to download the PR list.


eng/scripts/Approve-Eng-Common-Sync-PRs.ps1
```diff
@@ -17,13 +17,22 @@ gh auth status
...
   Write-Error "Please login via gh auth login"
-  exit 1
+  exit $LASTEXITCODE
 }
 
 $ToolsRepo = "Azure/azure-sdk-tools"
 
 $ghloggedInUser = (gh api user -q .login)
 # Get a temp access token from the logged in az cli user for azure devops resource
+$account = (az account show | ConvertFrom-Json)
```
benbp:
```suggestion
$account = (az account show -o json | ConvertFrom-Json)
```

JSON isn't the default on all machines.

benbp:
A bit extra but we could also consider setting the subscription back to what it was after.

PR #12562, opened 2025-10-20
Expose package SdkType in Apiview URI

helen229:
Closes https://github.com/Azure/azure-sdk-tools/issues/12242

corresponding PR for API view side changes:
https://github.com/Azure/azure-sdk-tools/pull/12398

Same as https://github.com/Azure/azure-sdk-tools/pull/12323, this pr could not merge due to build issue, creating a new branch to resolve the issue.

This PR updates the API review submission scripts to include package type information in API review requests. The changes enable the scripts to pass the SdkType property from package metadata through to the APIView service.

Key Changes:

Added $packageType parameter to API review submission functions
Extracted package type from packageInfo.SdkType metadata
Updated function calls to pass package type through the request chain

### DIFF
--- a/eng/common/scripts/Detect-Api-Changes.ps1
+++ b/eng/common/scripts/Detect-Api-Changes.ps1
@@ -22,7 +22,7 @@ Param (
 $configFileDir = Join-Path -Path $ArtifactPath "PackageInfo"
 
 # Submit API review request and return status whether current revision is approved or pending or failed to create review
-function Submit-Request($filePath, $packageName)
+function Submit-Request($filePath, $packageName, $packageType)
 {
     $repoName = $RepoFullName
     if (!$repoName) {
@@ -39,6 +39,7 @@ function Submit-Request($filePath, $packageName)
     $query.Add('packageName', $packageName)
     $query.Add('language', $LanguageShort)
     $query.Add('project', $DevopsProject)
+    $query.Add('packageType', $packageType)
     $reviewFileFullName = Join-Path -Path $ArtifactPath $packageName $reviewFileName
     # If CI generates token file then it passes both token file name and original file (filePath) to APIView
     # If both files are passed then APIView downloads the parent directory as a zip
@@ -126,6 +127,7 @@ foreach ($packageInfoFile in $packageInfoFiles)
 {
     $packageInfo = Get-Content $packageInfoFile | ConvertFrom-Json
     $pkgArtifactName = $packageInfo.ArtifactName ?? $packageInfo.Name
+    $packageType = $packageInfo.SdkType
 
     LogInfo "Processing $($pkgArtifactName)"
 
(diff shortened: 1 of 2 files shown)


eng/common/scripts/Detect-Api-Changes.ps1
```diff
@@ -126,6 +127,7 @@ foreach ($packageInfoFile in $packageInfoFiles)
 {
     $packageInfo = Get-Content $packageInfoFile | ConvertFrom-Json
     $pkgArtifactName = $packageInfo.ArtifactName ?? $packageInfo.Name
+    $packageType = $packageInfo.SdkType
```
maririos, line 130: This code is new to me, where is the information from `packageInfo` being created at? 
I was expecting to see `packageType` being set here according to the type of files, but I could be very wrong

helen229: $packageInfo = Get-Content $packageInfoFile | ConvertFrom-Json

from here. 

Since Ray changed packageMetaData to packageInfo  last week so I also updated to $packageInfo.SdkType

raych1:
@helen229 I'm curious that if [this API](https://github.com/Azure/azure-sdk-tools/blob/06abfe4918dafe5ba5c6b4f7251b33589992ae65/eng/common/scripts/Helpers/ApiView-Helpers.ps1#L51) also need to update?

helen229:
> @helen229 I'm curious that if [this API](https://github.com/Azure/azure-sdk-tools/blob/06abfe4918dafe5ba5c6b4f7251b33589992ae65/eng/common/scripts/Helpers/ApiView-Helpers.ps1#L51) also need to update?

Thanks for calling out this! I didn't update the corresponding controller for this API, this API is Endpoint: GET /AutoReview/GetReviewStatus, I think it's not related to our issues but I'll wait for Praveen for confirmation.

eng/common/scripts/Detect-Api-Changes.ps1
maririos: My question is more, are we sure packageInfo contains the correct SDKType? is it calculated as we expect?

weshaggard:
merging because all the sync PRs have merged and I don't want eng/common to be out of sync

praveenkuttappan:
> > @helen229 I'm curious that if [this API](https://github.com/Azure/azure-sdk-tools/blob/06abfe4918dafe5ba5c6b4f7251b33589992ae65/eng/common/scripts/Helpers/ApiView-Helpers.ps1#L51) also need to update?
> 
> Thanks for calling out this! I didn't update the corresponding controller for this API, this API is Endpoint: GET /AutoReview/GetReviewStatus, I think it's not related to our issues but I'll wait for Praveen for confirmation.

@raych1 @helen229  I don't think this needs an update. It just checks the status of API review so no need to pass SDK type.

PR #12564, opened 2025-10-20
[azsdk-cli] Organize command line hierarchy

benbp:
This PR takes a first pass at organizing our command line hierarchy. This ensures alignment with https://github.com/Azure/azure-sdk-tools/pull/12071 as well as organizing the rest of the CLI commands not specified in that doc.

One area that I'm still not sure of is that we can probably merge the top-level `spec-workflow` and `typespec` commands, since they are both typespec related? @praveenkuttappan @maririos @chrisradek thoughts?

The new structure (levels 1 and 2):

```
⇉ ⇉ ⇉ dotnet run --no-restore -- -h
Description:
  azsdk cli - A Model Context Protocol (MCP) server that facilitates tasks for anyone working with the Azure SDK team.

Usage:
  azsdk [command] [options]

Options:
  -h, --help    Show help and usage information
  --version     Show version information
  --debug       Enable debug logging [default: False]
  -o, --output  The format of the output. Supported formats are: plain, json [default: plain]

Commands:
  mcp            Starts the MCP server (stdio mode)
  package, pkg   Package operations
  config         SDK service configuration commands
  azp, pipeline  Azure Pipelines commands
  release-plan   Manage release plans in AzureDevops
  tsp, typespec  Commands for setting up or working with TypeSpec projects
  spec-workflow  TypeSpec SDK generation commands
  test           Test operations for SDK packages
  eng            Internal azsdk engineering system commands
  example        Example tool demonstrating framework features
```

```
⇉ ⇉ ⇉ dotnet run --no-restore -- pkg -h
Description:
  Package operations

Usage:
  azsdk pkg [command] [options]

Options:
  -h, --help    Show help and usage information
  --debug       Enable debug logging [default: False]
  -o, --output  The format of the output. Supported formats are: plain, json [default: plain]

Commands:
  validate <All|Changelog|Cspell|Dependency|Format|Linting|Readme|Snippets>  Run validation checks for SDK packages [default: All]
  readme                                                                     README operations for SDK packages
  release-readiness                                                          Checks release readiness of a SDK package
  build                                                                      Builds SDK source code for a specified language and project
  generate                                                                   Generates SDK code for a specified language based on the provided 'tspconfig.yaml' or 'tsp-location.yaml'
  release                                                                    Run the release pipeline for the package
  test                                                                       Test operations for SDK packages
```

```
⇉ ⇉ ⇉ dotnet run --no-restore -- config -h
Description:
  SDK service configuration commands

Usage:
  azsdk config [command] [options]

[...]

PR #12566, opened 2025-10-20
Fix tool filtering to work off mcp tool attribute name, not method name

benbp:
Fix tool filtering to work off the mcp tool attribute name, not the C# method name.

The `GetCommand` method is going to change slightly after #12564 goes in, but the core logic change to tool filtering is the same so I figured I'd get this up for review.

```
⇉ ⇉ ⇉ dotnet run --no-restore -- mcp  --tools '*hello*'
...
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://127.0.0.1:36615
...

{"jsonrpc":"2.0","id":3,"method":"tools/list","params":{}}

info: ModelContextProtocol.Server.McpServer[570385771]
      Server (azsdk 0.5.3.0) method 'tools/list' request handler called.
info: ModelContextProtocol.Server.McpServer[1867955179]
      Server (azsdk 0.5.3.0) method 'tools/list' request handler completed.

{"result":{"tools":[{"name":"azsdk_hello_world","description":"Echoes the message back to the client","inputSchema":{"type":"object","properties":{"message":{"type":"string"}},"required":[
"message"]}},{"name":"azsdk_hello_world_fail","description":"Echoes the message back to the client with a failure","inputSchema":{"type":"object","properties":{"message":{"type":"string"}}
,"required":["message"]}}]},"id":3,"jsonrpc":"2.0"}
```

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
@@ -130,8 +130,8 @@ namespace Azure.Sdk.Tools.Cli.Services
                     toolMethods = toolMethods.Where(m =>
                     {
                         var attr = m.GetCustomAttribute<McpServerToolAttribute>();
-                        return toolMatchList.Any(glob => FileSystemName.MatchesSimpleExpression(glob, attr?.Name));
-                    }).ToList();
+                        return attr?.Name is not null && toolMatchList.Any(glob => FileSystemName.MatchesSimpleExpression(glob, attr.Name));
+                    });
                 }
 
                 foreach (var toolMethod in toolMethods)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
```diff
@@ -121,23 +122,32 @@ public static void RegisterInstrumentedMcpTools(IServiceCollection services, str
                     continue;
                 }
 
-                foreach (var toolMethod in toolType.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance))
+                var toolMethods = toolType.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance)
```
scbedd, line 125: How much does adding this reflection slow down startup? Before we eliminated the reflection for Getting the active tools using the static toolslist, we saw something like a 95% slowdown from that reflection alone. Is this reflection not as impactful?

benbp: Good question. It's not noticeable from what I can tell but I haven't measured (server startup is <1s with tool filtering). In the near/mid future I'm probably going to have to remove this entirely in favor of source generation so that we can support AOT.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
```diff
@@ -121,23 +122,32 @@ public static void RegisterInstrumentedMcpTools(IServiceCollection services, str
...
+                                    .Where(m => m.GetCustomAttribute<McpServerToolAttribute>() is not null);
+
+                if (toolMatchList.Length > 0)
+                {
+                    toolMethods = toolMethods.Where(m =>
+                    {
+                        var attr = m.GetCustomAttribute<McpServerToolAttribute>();
+                        return toolMatchList.Any(glob => FileSystemName.MatchesSimpleExpression(glob, attr?.Name));
```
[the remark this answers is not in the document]
scbedd: Copilot is probably right about this but if we pass null it'll probably explode which is fine for an error.

PR #12567, opened 2025-10-20
add token usage data to telemetry

benbp:
This PR adds data for token usage to our telemetry for commands that make backend calls to openai via microagent or azure agent service.

It does this by updating the telemetry processor to look for a list of properties that may have been set via `Activity.SetCustomProperty()` and promoting them to tags. Classes within an activity scope can set these properties via `Activity.Current?.SetCustomProperty()`. In the future, we will ideally add some middleware to the MCP server and the CLI parser to auto-set tags like package, language, etc. based on well known parameters.

It also changes the telemetry opt-out in debug mode to just remove the exporter (skip upload). This way we can still get the telemetry properties dumped to console when `--debug` is enabled (very useful for testing).

**OLD DISCUSSION FROM PRIOR CHANGES, IGNORE NOT RELEVANT**

[EDIT] I am going to try out the approach suggested in https://github.com/Azure/azure-sdk-tools/pull/12567#issuecomment-3424008171 relevant to my questions/concerns below:

I'm not confident in the approach I took here, so would love to get feedback. The core issue I faced is that we register the token usage helper in DI with `AddScoped` but the telemetry service is a singleton via `AddSingleton`. The benefit of scoped is that we get tracking of token usage by any class within the request scope. That means if we have multiple backend agent calls from one MCP tool call, those will all get added to the token usage count that we submit for the telemetry event without having to propagate a token usage helper instance ourselves (DI does it for us). However because the telemetry service is a singleton it can't take a dependency on the scoped token usage helper. I think we still want telemetry service as a singleton so that we can track a root/parent activity id to record all tool calls that exist within an MCP server session, as detailed in our telemetry spec.

What I have done to work around this is introduce a static `TelemetryContext` class that we add to our `TelemetryProcessor`. The processor is wired up when we initialize the telemetry service and handles `OnEnd` events for any `Activity`, so that we can add the final token usage to the event before it gets uploaded. Now at the start of any CLI command or MCP tool call we invoke `TelemetryContext.Reset(tokenUsageHelper)` using the injected token usage helper, and any future scoped tracker classes could also be added onto the context.

Overall this approach of managing state by patching a global object feels super brittle, as it's easy to miss that `Reset()` needs to be called if we ever refactor or add new modes. I'd love if anyone has ideas about a simpler way that I could do this.

This will change slightly after #12566 goes in but the main logic for review will be as presented.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Extensions/OpenTelemetryExtensions.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Extensions/OpenTelemetryExtensions.cs
@@ -8,8 +8,6 @@ using Azure.Sdk.Tools.Cli.Telemetry;
 using Azure.Sdk.Tools.Cli.Telemetry.InformationProvider;
 using Azure.Monitor.OpenTelemetry.Exporter;
 using Microsoft.Extensions.Azure;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using OpenTelemetry.Resources;
 using OpenTelemetry.Trace;
@@ -31,13 +29,9 @@ public static class OpenTelemetryExtensions
                 {
                     options.Version = assemblyName.Version.ToString();
                 }
-#if !DEBUG
                 var collectTelemetry = Environment.GetEnvironmentVariable("AZSDKTOOLS_COLLECT_TELEMETRY");
                 options.IsTelemetryEnabled = string.IsNullOrEmpty(collectTelemetry)
                     || (bool.TryParse(collectTelemetry, out var shouldCollect) && shouldCollect);
-#else
-                options.IsTelemetryEnabled = false;
-#endif
             });
 
         services.AddSingleton<ITelemetryService, TelemetryService>();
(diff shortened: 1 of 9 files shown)


timovv:
> because the telemetry service is a singleton it can't take a dependency on the scoped token usage helper

I wonder if it would be possible to invert this dependency. Have the token usage helper depend on the telemetry service or some intermediary and use a callback/event pattern to propogate the necessary information back up to the service. Just spitballing here, no idea whether that is actually feasible or whether it would in fact help with the problem.

Overall I agree that we should try and avoid the static but if the alternatives are worse I would be fine with having it in.

benbp:
@timovv That's a great idea I'll try it out. Anytime we make a call to `tokenUsageHelper.Add()` it could update some context bag in the telemetry service, and we could possibly ensure it's scoped by the activity/request id to avoid any global state collisions/races. This way it's the helper itself responsible for telemetry not the caller, so there's no coupling with the tool handlers.

benbp:
@timovv I have a new and much simpler approach using some of the built-in features of `Activity` that we discussed in DM. Let me know what you think. If references to `Activity.Current` get out of hand across the codebase then we could reasses this approach I think.

timovv (approved):

PR #12574, opened 2025-10-20
[azsdk-cli] Log if process timeout is especially long

benbp:
[azsdk-cli] Log if process timeout is especially long

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/Options/ProcessOptions.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/Options/ProcessOptions.cs
@@ -16,7 +16,7 @@ public interface IProcessOptions
 
 public class ProcessOptions : IProcessOptions
 {
-    public static readonly int DEFAULT_PROCESS_TIMEOUT_SECONDS = 120;  // Default timeout of 2 minutes
+    public static readonly TimeSpan DEFAULT_PROCESS_TIMEOUT = TimeSpan.FromMinutes(2);
     private static readonly TimeSpan maxProcessTimeout = TimeSpan.FromHours(2);
 
     private const string CMD = "cmd.exe";
@@ -81,7 +81,7 @@ public class ProcessOptions : IProcessOptions
             throw new ArgumentOutOfRangeException(nameof(timeout), "timeout cannot exceed 2 hours.");
         }
 
-        this.Timeout = timeout ?? TimeSpan.FromSeconds(DEFAULT_PROCESS_TIMEOUT_SECONDS);
+        this.Timeout = timeout ?? DEFAULT_PROCESS_TIMEOUT;
         this.LogOutputStream = logOutputStream;
     }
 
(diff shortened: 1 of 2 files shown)


samvaity (approved):
Looks good to me, except for the copilot space comment.
I'm also wondering if there's a way to set an upper limit on the process runner task besides just logging.

benbp:
> Looks good to me, except for the copilot space comment.

@samvaity I have the trailing space in the optional timeout log value, that way we don't have two spaces before the word `in` if we don't inject the log.

> I'm also wondering if there's a way to set an upper limit on the process runner task besides just logging.

Do you mean like a max allowed timeout value? We can definitely do a runtime check but a compile time check would be even better. I'll see what I can do

timovv:
Might need to be a big number if there is a hard cap, I know some live test runs in the JS repo at least can go on for 30 minutes plus.

benbp:
Yeah I was thinking like 1-2 hours or something. That way MCP or CLI agents don't get hard stuck if left to do work in the background (idk how long claude/codex will wait by default).

benbp:
Updated to add an argument exception to the options constructor. I think a compile time check would require a custom type/analyzer/source gen which seems like more trouble than it's worth.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/Options/ProcessOptions.cs
```diff
@@ -16,8 +16,10 @@ public interface IProcessOptions
 
 public class ProcessOptions : IProcessOptions
 {
+    public static readonly int DEFAULT_PROCESS_TIMEOUT_SECONDS = 120;  // Default timeout of 2 minutes
```
timovv: nit: maybe this could be a TimeSpan too?
```suggestion
    public static int DefaultProcessTimeout { get; } = TimeSpan.FromMinutes(2);
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/Options/ProcessOptions.cs
```diff
@@ -17,6 +17,8 @@ public interface IProcessOptions
 public class ProcessOptions : IProcessOptions
 {
     public static readonly int DEFAULT_PROCESS_TIMEOUT_SECONDS = 120;  // Default timeout of 2 minutes
+    private static readonly TimeSpan maxProcessTimeout = TimeSpan.FromHours(2);
```
samvaity, line 20: This is a fair timeout, I know in Java running tests in record mode can sometimes take over an hour. Just sharing a general thought as how this affects the user experience for such lengthy operations, but I assume they were aware of this when they made their decision.

timovv (approved):
If long-running stuff ends up being common it might be interesting to investigate some way of backgrounding things (like how VSCode Copilot can run terminal commands in the background if it thinks it'll take a while)

benbp:
> If long-running stuff ends up being common it might be interesting to investigate some way of backgrounding things (like how VSCode Copilot can run terminal commands in the background if it thinks it'll take a while)

Yeah it doesn't seem like MCP has a first class concept of a background task, but I could see us creating a simple implementation of it. I would argue that people doing long running azsdk commands in an AI-assisted way are probably better off just running them via claude code with sub-agents or codex in a separate terminal window/tmux pane.

timovv:
To that point I guess there's nothing stopping VS Code Copilot from running the CLI command itself in a background terminal.

PR #12576, opened 2025-10-21
Add Dependency Management for Java

samvaity:
image

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Services/Languages/JavaLanguageSpecificChecksTests.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Services/Languages/JavaLanguageSpecificChecksTests.cs
@@ -97,7 +97,7 @@ namespace Azure.Sdk.Tools.Cli.Tests.Services.Languages
         }
 
         /// <summary>
-        /// Sets up successful Maven install command (Azure SDK approach).
+        /// Sets up successful Maven install command.
         /// </summary>
         private void SetupSuccessfulMavenInstall()
         {
@@ -150,38 +150,40 @@ namespace Azure.Sdk.Tools.Cli.Tests.Services.Languages
         }
 
         /// <summary>
-        /// Sets up successful Maven javadoc command.
+        /// Sets up successful Maven linting command with all tools including javadoc.
+        /// This matches the fail-safe accumulation approach.
         /// </summary>
-        private void SetupSuccessfulJavadoc()
+        private void SetupSuccessfulMavenLinting()
         {
+            var pomPath = Path.Combine(JavaPackageDir, "pom.xml");
             MockProcessHelper.Setup(x => x.Run(It.Is<ProcessOptions>(p => 
                 (p.Command == "mvn" || p.Command == "cmd.exe") && 
-                p.Args.Contains("javadoc:jar")), 
+                p.Args.Contains("install") &&
+                p.Args.Contains("--no-transfer-progress") &&
+                p.Args.Contains("-DskipTests") &&
+                p.Args.Contains("-Dgpg.skip") &&
+                p.Args.Contains("-DtrimStackTrace=false") &&
+                p.Args.Contains("-Dmaven.javadoc.skip=false") &&
+                p.Args.Contains("-Dcodesnippet.skip=true") &&
+                p.Args.Contains("-Dspotless.skip=false") &&
+                p.Args.Contains("-Djacoco.skip=true") &&
+                p.Args.Contains("-Dshade.skip=true") &&
+                p.Args.Contains("-Dmaven.antrun.skip=true") &&
+                p.Args.Contains("-Dcheckstyle.failOnViolation=false") &&
+                p.Args.Contains("-Dcheckstyle.failsOnError=false") &&
+                p.Args.Contains("-Dspotbugs.failOnError=false") &&
+                p.Args.Contains("-Drevapi.failBuildOnProblemsFound=false") &&
+                p.Args.Contains("-am") &&
+                p.Args.Contains("-f") &&
+                p.Args.Contains(pomPath) &&
+                p.WorkingDirectory == JavaPackageDir &&
+                p.Timeout == TimeSpan.FromMinutes(15)), 
                 It.IsAny<CancellationToken>()))
                 .ReturnsAsync(new ProcessResult { 
                     ExitCode = 0, 
                     OutputDetails = [(StdioLevel.StandardOutput, 
                         "[INFO] Building jar: /path/to/target/test-1.0-javadoc.jar\n" +
                         "[INFO] BUILD SUCCESS"
-                    )] 
-                });
-        }
-
-        /// <summary>
-        /// Sets up failed Maven javadoc command.
-        /// </summary>
-        private void SetupFailedJavadoc()
-        {
-            MockProcessHelper.Setup(x => x.Run(It.Is<ProcessOptions>(p => 
-                (p.Command == "mvn" || p.Command == "cmd.exe") && 
-                p.Args.Contains("javadoc:jar")), 
-                It.IsAny<CancellationToken>()))
-                .ReturnsAsync(new ProcessResult { 
-                    ExitCode = 1, 
-                    OutputDetails = [(StdioLevel.StandardError, 
-                        "[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.4.1:jar (default) on project test: MavenReportException: Error while generating Javadoc:\n" +
-                        "[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.\n" +
-                        "[ERROR] Command line was: /usr/lib/jvm/java-11-openjdk/bin/javadoc @options @packages"
                     )] 
                 });
         }
@@ -446,8 +448,7 @@ namespace Azure.Sdk.Tools.Cli.Tests.Services.Languages
         {
             // Arrange
             SetupSuccessfulMavenVersionCheck();
-            SetupSuccessfulMavenInstall();
-            SetupSuccessfulJavadoc();
+            SetupSuccessfulMavenLinting();
 
             // Act
             var result = await LangService.LintCodeAsync(JavaPackageDir, false, CancellationToken.None);
@@ -464,7 +465,6 @@ namespace Azure.Sdk.Tools.Cli.Tests.Services.Languages
             // Verify Maven commands were called correctly
             MockProcessHelper.Verify(x => x.Run(It.Is<ProcessOptions>(p => IsMavenVersionCheck(p)), It.IsAny<CancellationToken>()), Times.Once);
             MockProcessHelper.Verify(x => x.Run(It.Is<ProcessOptions>(p => IsMavenInstallCommand(p)), It.IsAny<CancellationToken>()), Times.Once);
-            MockProcessHelper.Verify(x => x.Run(It.Is<ProcessOptions>(p => IsJavadocCommand(p)), It.IsAny<CancellationToken>()), Times.Once);
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
```diff
@@ -309,6 +316,76 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, Canc
...
+    /// <param name="packagePath">The package directory path</param>
+    /// <param name="pomPath">The path to the pom.xml file</param>
+    /// <param name="cancellationToken">Cancellation token</param>
+    /// <returns>Analysis results</returns>
+    private async Task<CLICheckResponse> AnalyzeDependencyTreeAsync(string packagePath, string pomPath, CancellationToken cancellationToken)
+    {
+        var command = "mvn";
+        var args = new[] { "dependency:tree", "-Dverbose", "-f", pomPath };
+
+        var result = await _processHelper.Run(new(command, args, workingDirectory: packagePath, timeout: TimeSpan.FromMinutes(5)), cancellationToken);
```
benbp, line 292: Do we want an explicit path for if exit code is non-zero? Warning "Dependecy analysis found issues" could be misleading if the output is like "mvn not installed" or something. Also if this will eventually get exposed to MCP then none of the command error info will be returned to the client.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Services/Languages/JavaLanguageSpecificChecksTests.cs
```diff
@@ -150,13 +150,34 @@ private void SetupFailedSnippetUpdate()
...
         /// <summary>
-        /// Sets up successful Maven javadoc command.
+        /// Sets up successful Maven linting command with all tools including javadoc.
+        /// This matches the fail-safe accumulation approach.
         /// </summary>
-        private void SetupSuccessfulJavadoc()
+        private void SetupSuccessfulMavenLinting()
         {
+            var pomPath = Path.Combine(JavaPackageDir, "pom.xml");
             MockProcessHelper.Setup(x => x.Run(It.Is<ProcessOptions>(p =>
```
benbp, line 159: I wonder if at some point it would be worth creating a `MavenHelper` with `MavenOptions` that way the mocks can just look for `It.Is<MavenOptions>` and maybe filter on the args a bit more easily vs. having to enumerate all the args. Then we wouldn't have to check `mvn || cmd.exe` either.

samvaity: Makes sense, I will look into it. Follow-up issue #12594

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
samvaity: We only reach here from a non-error non-exit path for the installation errors. But I see your point there could be other unrelated errors are escaped before dependency analysis. Let me update.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
```diff
@@ -111,31 +111,34 @@ public async Task<CLICheckResponse> LintCodeAsync(string packagePath, bool fixCh
...
             var command = "mvn";
             var args = new List<string>
             {
                 "install",
                 "--no-transfer-progress",
                 "-DskipTests",
                 "-Dgpg.skip",
-                "-Dmaven.javadoc.skip=true",
+                "-DtrimStackTrace=false",
+                "-Dmaven.javadoc.skip=false",
```
l0lawrence, line 125: so now javadoc is being run with everything else?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
```diff
@@ -257,8 +248,24 @@ public async Task<CLICheckResponse> LintCodeAsync(string packagePath, bool fixCh
         }
         results.Add(("RevAPI", revapiHasIssues));
 
-        _logger.LogInformation("Linting results parsed: Checkstyle ran={CheckstyleRan} issues={CheckstyleIssues}, SpotBugs ran={SpotBugsRan} issues={SpotBugsIssues}, RevAPI ran={RevapiRan} issues={RevapiIssues}",
-            checkstyleRan, checkstyleHasIssues, spotbugsRan, spotbugsHasIssues, revapiRan, revapiHasIssues);
+        // Check for Javadoc execution and results (included in linting)
+        var javadocRan = output.Contains("maven-javadoc-plugin") && !output.Contains("-Dmaven.javadoc.skip=true");
```
l0lawrence: wouldnt skip not be true because we set it to false above, does this check anything?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
```diff
@@ -257,8 +248,24 @@ public async Task<CLICheckResponse> LintCodeAsync(string packagePath, bool fixCh
...
+        var javadocHasIssues = false;
+        if (javadocRan)
+        {
+            // Look for javadoc errors or warnings
+            var javadocSuccess = !output.Contains("Javadoc Warnings") &&
```
l0lawrence: should this get added into the ParseLintingResults() fn or is it quite different from the other formats?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
```diff
@@ -309,6 +316,76 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, Canc
...
+            // Validate Maven and POM prerequisites
+            var pomPath = Path.Combine(packagePath, "pom.xml");
+            var prerequisiteCheck = await ValidateMavenPrerequisitesAsync(packagePath, pomPath, cancellationToken);
+            if (prerequisiteCheck != null)
+            {
+                return prerequisiteCheck;
+            }
+
+            // Azure SDK for Java uses BOM-based dependency management
+            // Focus on analysis and validation rather than automatic fixes
```
l0lawrence: nit: I think this comment could be removed?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
samvaity: yes, I realized I could combine it with the other linting runs no need for separate command.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
samvaity: Simplified the linting result parsing.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
```diff
@@ -111,31 +111,34 @@ public async Task<CLICheckResponse> LintCodeAsync(string packagePath, bool fixCh
...
+            // The -am flag ensures parent POMs are built/resolved automatically
             var command = "mvn";
             var args = new List<string>
             {
                 "install",
                 "--no-transfer-progress",
                 "-DskipTests",
                 "-Dgpg.skip",
-                "-Dmaven.javadoc.skip=true",
+                "-DtrimStackTrace=false",
```
raych1, line 124: I'm wondering if this has any overlap with current `build` command.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
```diff
@@ -309,6 +249,75 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, Canc
...
+        else
+        {
+            var errorMessage = "Dependency analysis found issues - check Maven output for conflicts or build errors";
+            _logger.LogWarning(errorMessage);
+
+            return new CLICheckResponse(1, output, errorMessage)
+            {
+                NextSteps = [
+                    "Add Azure SDK BOM to dependencyManagement section in pom.xml: https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/boms/azure-sdk-bom",
+                    "Use Azure SDK BOM to manage versions: <groupId>com.azure</groupId> <artifactId>azure-sdk-bom</artifactId> <version>1.3.0</version>",
```
raych1: Should we replace the detailed value with a link to the file in `main` so that any future updates are automatically reflected?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
```diff
@@ -344,4 +353,37 @@ public async Task<CLICheckResponse> UpdateSnippetsAsync(string packagePath, Canc
...
+        return [
+            ("Checkstyle", HasCheckstyleIssues(output)),
+            ("SpotBugs", HasSpotBugsIssues(output)), 
+            ("RevAPI", HasRevapiIssues(output)),
+            ("Javadoc", HasJavadocIssues(output))
+        ];
+    }
+
+    private static bool HasCheckstyleIssues(string output) =>
+        output.Contains("Checkstyle violations") &&
```
raych1: Nit: should we ignore case to make this more tolerant?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
samvaity: >                    "Add Azure SDK BOM to dependencyManagement section in pom.xml: https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/boms/azure-sdk-bom",

This line should work for that ?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
samvaity: Is the build command located centrally? I was assuming in the `build` command we would be skipping linting, but if not we can definitely reuse.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
raych1: It does. I feel like the first line is enough and we can remove the second line, as the version is included in the linked page.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs
raych1: Build command is configured in the [Java repo](https://github.com/Azure/azure-sdk-for-java/blob/478feeb7f0193d1a155d23420505cb62ccc8e4be/eng/swagger_to_sdk_config.json#L30). I think it skips linting, as linting is considered too much for code compile. But it's flexible to update the command in the language repo without touching this code base.

samvaity: This command is supposed to include linting and hence different from the build command. https://github.com/Azure/azure-sdk-tools/issues/12594 should address standardizing calling for maven operations further.

raych1 (approved):

l0lawrence (approved):
lgtm thank you!

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Services/Languages/JavaLanguageSpecificChecksTests.cs
```diff
@@ -150,13 +150,34 @@ private void SetupFailedSnippetUpdate()
...
+        private void SetupSuccessfulMavenLinting()
         {
+            var pomPath = Path.Combine(JavaPackageDir, "pom.xml");
             MockProcessHelper.Setup(x => x.Run(It.Is<ProcessOptions>(p => 
                 (p.Command == "mvn" || p.Command == "cmd.exe") && 
-                p.Args.Contains("javadoc:jar")), 
+                p.Args.Contains("install") &&
+                p.Args.Contains("--no-transfer-progress") &&
+                p.Args.Contains("-DskipTests") &&
+                p.Args.Contains("-Dgpg.skip") &&
```
haolingdong-msft, line 164: Hi @samvaity, May I understand more on where is the command used? Because I saw a lot of skips here, but I believe currently CI will not skip these? If it is for validation, should we make it the same as CI? Srikanta once express the concern on skipping these checks. Add @srnagar as well for awareness.

Issue #12183, opened 2025-09-22
Strategies for updating client.tsp


chrisradek:
Using this issue to identify different efforts to feed updates to client.tsp

weidongxu-microsoft:
As information

Current impl in Java mitigation for SDK migration to TypeSpec is done as
1. agent read a generated CHANGELOG.md
2. it also read the [guide to mitigate breaks](https://github.com/weidongxu-microsoft/java-sdk-tools/blob/main/mcp/assets/migrate-instructions.md?plain=1#L47-L78), with instructions about how to parse the information in CHANGELOG
3. it writes the client.tsp, according to the guide above

This "write client.tsp" is done as a single tool (with supporting tools like generateSdk, buildSdk, getChangelog).

For this impl, it would be a throw-away one. We won't use this anymore, after TypeSpec migration is completed for all services.

---

The normal breaking changes mitigation likely going to follow a different guide.

The main difference is that, for "SDK migration to TypeSpec", we are sure there is absolutely no API change. Hence all change in SDK is supposedly caused by either TypeSpec or emitter.
But when in normal breaking changes, it is pretty valid that the cause is a break in API (e.g. service removes an API due to security reason) -- in such case the instructions would be very different.

haolingdong-msft:
Hi @chrisradek , thanks for coordinating the efforts on updating `client.tsp`!

For your information as well, Java previously implements a tool to update client name in the java mcp server. 

It will basically do the following:

- update client.tsp to add @@clientName 
- regenerate after adding @@clientName
- update existing source code, tests, samples and documentation that refer to the old class. Because there are customized code, which will not be updated by regeneration.
- compile and build the sdk, if it fails, fix according to error messages until build succeeds.
- delete the un-referenced classes.

Code can be found here: https://github.com/Azure/azure-sdk-for-java/blob/fa6652ef2e2ca7429e64119c9eb1a6c4b49e4bdd/eng/tools/mcp/azure-sdk-java-mcp/src/client-name-update.ts 

Whole picture of Java MCP can be found in my initial PR: https://github.com/Azure/azure-sdk-for-java/pull/45927

Please note that the java mcp server configuration is already deleted in the [PR](https://github.com/Azure/azure-sdk-for-java/pull/46435) (thanks @samvaity) as we now centralize mcp tools to azure-sdk-tools.

chrisradek:
Thank you for that info @haolingdong-msft!

[...]

Issue #12420, opened 2025-10-08
[TypeSpec] Improve client customizations instructions


chrisradek:
Currently in the azure-rest-api-specs repo, there are copilot instructions to reference the llms.txt files generated for typespec and typespec-azure docs.

In testing applying client customizations to specs, I'm finding that while copilot will reference the correct docs, it doesn't always use the typespec azure libraries correctly.

For example, I prompted copilot to split a service into 2 clients for only the python SDK, and specified which operations belonged to which client. It correctly used the `@client` decorator, but it never applied scoping correctly.

Now, if I asked copilot a follow-up along the lines of "How are you sure this only applies to python", it again looks up the documentation and correctly identifies it needs to use the scopes parameter, but I can't get it to do that on its own proactively.

We should look into either updating the typespec-azure docs to put more emphasis on when each parameter should be used/give more examples, and/or update instructions to provide more guidance on patterns to look for (e.g. scope) when applying customizations.

PR #12577, opened 2025-10-21
Spec: customizing clients

chrisradek:
Using this PR to help collect scenarios that drive `client.tsp` customizations.  Related to #12183

Current status:
Lacking enough scenarios that are triggered by some automatable process for driving client.tsp updates. As such, the value proposition for creating a language-agnostic workflow is relatively low given we already have tooling be written for .NET, and a prompt for Go's scenario. Need input from other language teams to tip the scales!


tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,93 @@
...
+
+---
+
+## Definitions
+
+_Terms used throughout this spec with precise meanings:_
+
+- **<a id="typespec"></a>TypeSpec**: The specification language used to define Azure service APIs. SDK code is generated from TypeSpec specifications located in the azure-rest-api-specs repository.
+
+- **<a id="typespec-customizations"></a>TypeSpec Customizations**: SDK-specific customizations made in the `client.tsp` file to control SDK generation. This term does not refer to modifications of the service API TypeSpec files.
```
maririos: we should add `SDK customizations` too so the distinction between the two is clear

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,93 @@
...
+TypeSpec specs are written to describe service APIs. However, our team supports generating SDKs in no less than 5 languages, each differing in their idioms. The `client.tsp` file acts as a TypeSpec entry point for SDK emitters, applying language-specific customizations to the service spec that are then used by each respective SDK emitter.
+
+Service teams cannot be expected to understand the nuances of each SDK language and how to best customize the generated code for each. Applying changes also requires understanding how to use [Azure.ClientGenerator.Core](https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/).
+
+### Current State
+
+Currently, feedback for adding client customizations to `client.tsp` can come in 2 ways:
+
+1. From manual reviews, e.g. API View feedback
+2. From automated processes, e.g. code analyzers
```
maririos: we are missing the scenario where we need the SDKs to be [dependable](https://azure.github.io/azure-sdk/general_introduction.html). This will come in the form of breaking changes mitigations.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,93 @@
...
+
+TypeSpec specs are written to describe service APIs. However, our team supports generating SDKs in no less than 5 languages, each differing in their idioms. The `client.tsp` file acts as a TypeSpec entry point for SDK emitters, applying language-specific customizations to the service spec that are then used by each respective SDK emitter.
+
+Service teams cannot be expected to understand the nuances of each SDK language and how to best customize the generated code for each. Applying changes also requires understanding how to use [Azure.ClientGenerator.Core](https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/).
+
+### Current State
+
+Currently, feedback for adding client customizations to `client.tsp` can come in 2 ways:
+
+1. From manual reviews, e.g. API View feedback
```
maririos: Consider modifying it to `From reviews` because they can come from APIView Copilot (not manual :) )

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,93 @@
...
+
+**.NET**:
+The .NET SDK has code analyzers that run during their SDK build (`dotnet build`) that identifies problematic code (e.g. class names that don't follow .NET naming conventions) and suggests fixes.
+Many of these fixes are handled by adding customizations to `client.tsp` for .NET, such as renaming models with `@clientName`. Currently these fixes are manually applied after SDK builds fail.
+_.NET has a clear flow for identifying and applying `client.tsp` customizations via: Emit SDK -> Build SDK -> Apply fixes -> Repeat._
+
+**Go**:
+The Go SDK has a [breaking changes](https://github.com/Azure/azure-sdk-for-go/blob/instruction/.github/prompts/go-sdk-breaking-changes-review.prompts.md) prompt that analyzes the generated SDK's changelog for specific patterns and maps those to `client.tsp` customizations where possible.
+_We don't currently have a clear step in our process for identifying breaking changes, but the flow for Go is: Emit SDK -> Analyze changelog & apply fixes -> Repeat._
+
```
maririos, line 69: We know about the Java scenario which is when the build fails (bc of analyzers or because of breaking changes)

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,93 @@
...
+**All Languages**:
+All language SDKs have their APIs reviewed in API View. Common feedback includes renaming models per language, or changing operations that appear in a client. We have documentation for customizing SDKs such as [renaming types](https://azure.github.io/typespec-azure/docs/howtos/generate-client-libraries/09renaming/), but recent (10/14/2025) testing shows that even with the current documentation, AI agents have difficulty applying these customizations correctly in a consistent manner: particularly when `scope` is required.
+
+### Why This Matters
+
+We should minimize the amount of time service teams spend on customizing their TypeSpec for each SDK language. Let them focus on what they are experts on - the service API - rather than figuring out how to apply SDK-specific feedback to their TypeSpec.
+
+For automated processes like .NET, we can reduce the time spent in the outer loop of SDK build failures by proactively applying customizations to `client.tsp` before creating a PR in the azure-rest-api-specs repo.
```
maririos: Modifying the wording here will make this a principle of this effort, and it is to help user apply customizations to their SDKs.
Also, avoid using implementation details like `before creating a PR in the specs repo`. That is more an implementation detail and now the why it matters

maririos (commented):
Thanks for putting this together. General feedback:
-  I see the document doesn't talk about customizations to the SDK per se. Could you please update it and make sure it covers that case too?
- From my understanding, customization happens when there is a trigger saying a change is needed. The trigger then follows a process. This document should identify the triggers that we know of: build failure (caused by analyzers or breaking changes), feedback received by either people or tooling, and for user experience (breaking changes). With each trigger we can look: how to identify a change needs to happen? do we need to classify it to see how to address the customization? what is the success criteria?
- Once the triggers are identified, then what is the process to address them. According to the trigger the process can be different(?), but hopefully they all use the same steps/base. This is the part that is more confusing and that needs more time. Building the bones for that will be key in this doc

chrisradek:
> I see the document doesn't talk about customizations to the SDK per se. Could you please update it and make sure it covers that case too?

I was thinking SDK customizations was out of scope for `client.tsp` customizations, though it might make sense to merge the efforts @samvaity is doing on that front.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: Ah, the scenario I heard about was when TypeSpec changes cause the client customizations (the Java code) to fail. My understanding was that's the work @samvaity was focusing on - how to automatically fix the client customizations.

I also saw Java's earlier MCP server, but that had a cookbook for how to apply renames in `client.tsp`.

Would love to hear differently though! I don't know how they (or anyone outside of Go) detects breaking changes today.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: Ah, I was thinking this would fall under reviews (caught in API view) or automated processes, but I can call this out separately since I don't know how these are actually handled today except for Go.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
lmazuel: +1

SDK Customization are customization that couldn't be expressed in client.tsp, and have been done in the target language. Those are always located in the language repo, along the package. They may require pre-generation of some constructs through client.tsp.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,93 @@
...
+
+- **<a id="typespec-customizations"></a>TypeSpec Customizations**: SDK-specific customizations made in the `client.tsp` file to control SDK generation. This term does not refer to modifications of the service API TypeSpec files.
+
+- **<a id="api-view"></a>API View**: A web-based tool for reviewing SDK APIs. It allows language architects and SDK team members to provide feedback on just the SDK APIs without needing to understand the underlying implementation.
+
+---
+
+## Background / Problem Statement
+
+TypeSpec specs are written to describe service APIs. However, our team supports generating SDKs in no less than 5 languages, each differing in their idioms. The `client.tsp` file acts as a TypeSpec entry point for SDK emitters, applying language-specific customizations to the service spec that are then used by each respective SDK emitter.
```
lmazuel: Technically 7. I assume you don't talk about C++ and Rust, but they exist, and we ship SDKs :). C++ is mostly scoped to a set of core packages, and Rust is growing to catch up everything. They are not different in their process, it's just "yet another emitter".

This is a good example of client.tsp with Rust:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/storage/Microsoft.BlobStorage/client.tsp

chrisradek: I tried hedging with "no less than 5" :smile: - those 5 are just the ones currently represented in azsdk-cli where language detection is used. But yes you're absolutely right.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,93 @@
...
+
+Service teams cannot be expected to understand the nuances of each SDK language and how to best customize the generated code for each. Applying changes also requires understanding how to use [Azure.ClientGenerator.Core](https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/).
```
lmazuel: Not sure it's the main argument for client.tsp
- First, customization can be cross-language, so you do it once and for all. That applies for client structuration, some naming, etc.
- Then, you can put language specific as needed

IOW, it's incorrect to say that client.tsp is only for language specific changes. It's for SDK specific changes, and some instructions can apply to several of them. We do support for instance modifiers like `!python` to say "applies to everything but Python", because cross-language modification is important.

This means that when using AI to write client.tsp, we may have an interesting challenge at some point if each AI wants to write in the language they own, and nothing else, like ending up with:

```
@clientName(Foo, "bar", "python")
@clientName(Foo, "bar", "js")
@clientName(Foo, "bar", "csharp")
@clientName(Foo, "bar", "java")
```

which ideally we avoid.

The opposite problem is true, if you are working on Java, the AI needs to understand that this line applies to Java right now:
```
@clientName(Foo, "bar", "!python")
```

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Current State
+
+Currently, feedback for adding client customizations to `client.tsp` can come in 2 ways:
```
lmazuel: Technically three. We should support just someone opening Copilot, and asking to write in the client.tsp directly.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,93 @@
...
+
+For automated processes like .NET, we can reduce the time spent in the outer loop of SDK build failures by proactively applying customizations to `client.tsp` before creating a PR in the azure-rest-api-specs repo.
+
+For manual processes like API View reviews, we can improve the authoring experience to reduce the tedium and cognitive load of applying customizations manually.
+
+---
+
+## Goals and Exceptions/Limitations
+
+### Goals
```
lmazuel, line 76: My two cents here:

We have two different distinct components, that are important to separate from a SRP perspective:

1- Having a system that given a prompt, knows how to write a client.tsp directive
2- Having a system or guidance, on how to plug source of feedback to that system

For instance, for 1:
- Given a name change directive, where to do it? Adding a new clientName? Modifying the languages of an existing ones?
- Given an unclear directive, identify in TCGC what are the valid options, and make a suggestions. TCGC decorators in a finite list, we should be able to explictly describe all of them, to give the AI enough to understand what to use

That core behavior doesn't care where the feedback is coming from, it just knows that given a prompt, its job is to write the correct client.tsp

For 2:

We need a guidance system on how to build prompts for system 1. For instance, if it's coming from APIView, wrapping this under a template:
"Those comments are coming from architects for the language XXX: <insert APIView feedback here>"

If coming from C#, something like:
"The C# analyzer needs to modify the client.tsp for C# specifically for those symbols: <insert things here>"

If we know C# system is only about names, we can do better:
"The C# analyzer needs to modify the client.tsp for C# names using clientName, specifically for those symbols: <insert things here>"


What we want to avoid, is that C#  and Go and APIView, etc. rewrite their own "how to write a client.tsp" logic. This should be centralizer, and APIView/C#/Go all gives instructions to a unique client.tsp dedicated AI. This also means that a human can use the system straight from Copilot, with the same value and intelligence used for the automated system. IOW, the system 1 does not care if the feedback is manual or automated, it doesn't even know.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: Fair point. Really what I want to stress is that service teams think about the service API, they shouldn't be expected to know the ways that our SDKs need to be customized (that's what our efforts are for!)

This doesn't detract from your argument, but I was curious, so I ran a script (thanks copilot!) to see how `@clientName` is used in the specs repo. It found 5724 occurrences, with 4867 of those being scoped to a single SDK (C# being the biggest offender). Only 715 had no scope.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
samvaity: @chrisradek this spec addresses a gap: it focuses on automating spec-level customizations (renames, client splits, etc.), whereas the existing TspClientUpdateTool deals with code-level customizations. To achieve a seamless inner-loop, these must integrate into one unified workflow where the CLI handles both updating the client.tsp and updating any hand-written SDK code in tandem. 

Example: If a .NET analyzer error says “Class X should end with ‘Client’”, the tool should prefer to add a decorator in client.tsp. After updating the spec and regenerating, but there in an error in the handwritten code (not solvable by spec changes), where the current code patch tool (TspClientUpdateTool) would handle it.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
samvaity: The current spec focuses on client.tsp changes, but as a tooling there’s an implicit coordination needed between the Azure REST API spec repository (where client.tsp lives) and the language SDK repos (where custom code lives). 
I suggest we have a single integrated tool, effectively touching both: updating the spec file and the custom code. Ideally, the tool could output a pair of PRs – one to the azure-rest-api-specs with the TypeSpec changes, and one to the azure-sdk-for- with the updated code – or if using a monorepo approach, a single PR if all changes are in one place.

Currently, without the integrated tool concept the **order of operations** becomes crucial and the user will have to perform sequence execution of commands to get the desired code updates.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,96 @@
...
+
+What are we trying to achieve with this design?
+
+- [ ] Identify workflows that trigger the feedback required for identifying `client.tsp` customizations
+- [ ] Improve the service team experience for authoring `client.tsp` customizations
+
+## Design Proposal
```
samvaity, line 98: SDK developer’s inner loop, the `TspClientUpdateTool` (existing) handle post-generation code fixes, while Spec #12577 aims to handle pre-generation spec fixes. If we leave them disjointed, we’d have overlapping functionality and possibly confusion about what to run when. Integration means a developer can run one unified workflow to apply TypeSpec customizations and update custom code, obtaining an up-to-date SDK that incorporates cross-language feedback and passes all checks, without juggling multiple tools.

This is my proposal for revised architecture where we adapt/enhance the existing TspCLientUpdateTool to encompass the customization story:

    ┌────────────────────────────────────────────────────────────┐
    │ 1. Gather Inputs and Context                               │
    │ ────────────────────────────────────────────────────────── │
    │ • Detect SDK language using LanguageServiceResolver        │
    │ • Load tsp-location.yaml or --spec-path                    │
    │ • Collect APIView comments, analyzer warnings              │
    └────────────────────────────────────────────────────────────┘
                              ↓
    ┌────────────────────────────────────────────────────────────┐
    │ 2. Apply Spec-Level Customizations (Spec #12577)           │
    │ ────────────────────────────────────────────────────────── │
    │ • AI generates TypeSpec decorators (e.g., @friendlyName)   │
    │ • Update client.tsp                                        │
    │ • Compile TypeSpec to validate                             │
    │ • If error → refine AI prompt and retry (loop max 2x)      │
    └────────────────────────────────────────────────────────────┘
                              ↓
    ┌────────────────────────────────────────────────────────────┐
    │ 🔷 TspClientUpdateTool Workflow (Currently)                │
    │ ────────────────────────────────────────────────────────── │
    │ ┌────────────────────────────────────────────────────────┐ │
    │ │ 3. SDK Code Generation                                 │ │
    │ │ • Generate SDK code using updated TypeSpec             │ │
    │ │ • Output to --new-gen directory                        │ │
    │ └────────────────────────────────────────────────────────┘ │
    │                          ↓                                 │
    │ ┌────────────────────────────────────────────────────────┐ │
    │ │ 4. Code Customization Updates (AI Patch)v│ │
    │ │ • AI generates patches from diff + current code        │ │
    │ │ • Apply patches using ClientCustomizationCodePatchTool │ │
    │ │ • Fallback rules for trivial changes                   │ │
    │ └────────────────────────────────────────────────────────┘ │
    │                          ↓                                 │
    │ ┌────────────────────────────────────────────────────────┐ │
    │ │ 5. Validation and Iteration                            │ │
    │ │ • Compile SDK code                                     │ │
    │ │ • If build fails → retry AI patching (loop max 2x)     │ │
    │ │ • Optional: run tests, surface logical issues          │ │
    │ └────────────────────────────────────────────────────────┘ │
    └────────────────────────────────────────────────────────────┘
                              ↓
    ┌────────────────────────────────────────────────────────────┐
    │ 6. Finalize Outputs (TBD - can just do git diff)           │
    │ ────────────────────────────────────────────────────────── │
    │ • Output updated client.tsp (spec diff)                   │
    │ • Output updated SDK source files                         │
    │ • Summary log of changes and citations                    │
    └────────────────────────────────────────────────────────────┘

Would love to know your thoughts too @lmazuel and @maririos.

samvaity (commented):
The spec introduces updating the TypeSpec itself. Looking at it from customizations pov, there’s isn't a direct functional overlap (spec vs code), but there are potential redundancies and conflicts that we would need to manage:
1. Dual Fixing of Issues: If an API review requests “Rename class FooClient to FooServiceClient in all languages,” spec automation (e.g., @clientName in Spec #12577) should handle this before code generation. Otherwise, code patches may duplicate the effort or cause build failures. Coordinating spec fixes before code generation prevents redundant work and keeps design changes at the source.

2. Redundant Documentation or Guidelines: Both spec and code fixes use language guidelines (e.g., .NET class naming). If spec automation applies conventions first, code patchers won’t need to rename classes, streamlining the process and letting code updates focus on areas spec can't cover, like test code.

3. Workflow Confusion: Separate spec and code fixes can confuse developers about which tool to run and in what order, leading to unnecessary patches or incomplete updates. Merging workflows removes this ambiguity.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: I think you're right about combining the 2 into a single workflow. How does your existing tool get triggered today? Thinking of the "Gather Inputs and Context" section, seems like this tool could be referenced in the "next steps" of a build/validation step if they fail (being careful we don't get stuck in a loop)

Also recognizing we'll need to be able to invoke it on-demand e.g. when wanting to address API View comments.

chrisradek:
Current status:
I've expanded this design to include the gaps we identified and also speak to the code customization aspect.

Note that the actual design is still TBD - I'll create issues for each of the gaps - these will feed back into the design. For example, one gap is around copilot's current limitations for proposing client.tsp updates. We have https://github.com/Azure/azure-sdk-tools/issues/12420 that can be a potential solution.

Taking out of draft so the current background/requirements/gaps can be agreed on.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,116 @@
...
+_.NET has a clear flow for identifying and applying `client.tsp` customizations via: Emit SDK -> Build SDK -> Apply fixes -> Repeat._
+
+**Go**:
+The Go SDK has a [breaking changes](https://github.com/Azure/azure-sdk-for-go/blob/instruction/.github/prompts/go-sdk-breaking-changes-review.prompts.md) prompt that analyzes the generated SDK's changelog for specific patterns and maps those to `client.tsp` customizations where possible.
+_We don't currently have a clear step in our process for identifying breaking changes, but the flow for Go is: Emit SDK -> Analyze changelog & apply fixes -> Repeat._
+
+**All Languages**:
+All language SDKs have their APIs reviewed in API View. Common feedback includes renaming models per language, or changing operations that appear in a client. We have documentation for customizing SDKs such as [renaming types](https://azure.github.io/typespec-azure/docs/howtos/generate-client-libraries/09renaming/), but recent (10/14/2025) testing shows that even with the current documentation, AI agents have difficulty applying these customizations correctly in a consistent manner: particularly when `scope` is required.
```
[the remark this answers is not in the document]
chrisradek: The date is indeed correct - 2025.

jsquire:
//fyi:  @m-redding

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Current State
+
+Currently, feedback for adding client customizations to `client.tsp` can come in a few forms:
+
+1. From reviews, e.g. API View feedback
+2. From automated processes, e.g. code analyzers
+3. From manually triggered processes, e.g. breaking changes analysis
+
+**.NET**:
```
m-redding: We also have an API compat tool for .NET that runs with `dotnet build` - https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/overview, it will fail the build if an API is missing with an error like `"error CP0008: Type 'Foo.Bar.MyClass' does not implement interface 'Foo.Bar.IMyInterface' on lib/net8.0/MyAssembly.dll but it does on [Baseline] lib/net8.0/MyAssembly.dll"`

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,116 @@
...
+There has already been some work done to support identifying and fixing code customizations - customizations made directly to generated SDKs rather than to the TypeSpec files - in the `TspClientUpdateTool`. This tool exists as an MCP tool/CLI command and currently works _after_ an SDK has been generated from TypeSpec.
+
+#### Gaps
+
+The current state exposes some gaps we have today:
+
+- **AI agents are not experts on client customizations.** They make mistakes when applying customizations to `client.tsp`, even with access to the existing documentation.
+- **AI agents don't know when to suggest client customizations.** There is no mechanism in azsdk-cli today to steer AI agents to focus on applying `client.tsp` customizations.
+- **Nothing shared across languages.** Each language team has their own process for identifying and applying `client.tsp` customizations.
+- **Difficult to apply changes from outside the inner loop.** We lack a process for updating `client.tsp` from outside the azure-rest-api-specs repo.
```
m-redding: +1 to this, it's been awkward in .NET trying to make adjustments without a local copy of the specs repo.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,116 @@
...
+
+What are we trying to achieve with this design?
+
+- [ ] Improve the service team experience for authoring `client.tsp` customizations
+- [ ] Simplify the process of _how_ to apply `client.tsp` customizations for language teams
+- [ ] Integrate with the existing workflow for client code customizations
+
+Put another way, we have the following areas to focus on in the design proposal:
+
+1. Make AI an expert on `client.tsp` customizations. The base expertise shouldn't have to be redefined by each language team.
```
m-redding: @radhgupta did some work on giving the .NET generator agent additional context, I think that would be good to centralize as part of this - https://github.com/Azure/azure-sdk-tools/blob/main/src/dotnet/GeneratorAgent/src/Azure.Tools.GeneratorAgent/knowledge.md

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Definitions
+
+_Terms used throughout this spec with precise meanings:_
+
+- **<a id="typespec"></a>TypeSpec**: The specification language used to define Azure service APIs. SDK code is generated from TypeSpec specifications located in the azure-rest-api-specs repository.
+
+- **<a id="typespec-customizations"></a>TypeSpec Customizations**: SDK-specific customizations made in the `client.tsp` file to control SDK generation. This term does not refer to modifications of the service API TypeSpec files.
+
+- **<a id="code-customizations"></a>Code Customizations**: Hand-written modifications made directly to generated SDK code after generation. These customizations exist within the SDK language repositories and must be preserved across regeneration. Examples include adding convenience methods, custom error handling, or language-specific optimizations. Also known as "handwritten code" or "customization layer."
```
m-redding: I'm not sure if other languages have stuff like this but in .NET there are additional customizations that can be done by defining partial classes and regenerating - https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/.tspd/docs/customization.md

m-redding (approved):
Adding more context for .NET but agree with the explanation of the problem/gaps/state/etc

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,116 @@
...
+2. From automated processes, e.g. code analyzers
+3. From manually triggered processes, e.g. breaking changes analysis
+
+**.NET**:
+The .NET SDK has code analyzers that run during their SDK build (`dotnet build`) that identifies problematic code (e.g. class names that don't follow .NET naming conventions) and suggests fixes.
+Many of these fixes are handled by adding customizations to `client.tsp` for .NET, such as renaming models with `@clientName`. Currently these fixes are manually applied after SDK builds fail.
+_.NET has a clear flow for identifying and applying `client.tsp` customizations via: Emit SDK -> Build SDK -> Apply fixes -> Repeat._
+
+**Go**:
+The Go SDK has a [breaking changes](https://github.com/Azure/azure-sdk-for-go/blob/instruction/.github/prompts/go-sdk-breaking-changes-review.prompts.md) prompt that analyzes the generated SDK's changelog for specific patterns and maps those to `client.tsp` customizations where possible.
```
l0lawrence: In python we have a breaking changes check as well that we could try to leverage in a similar way - https://github.com/Azure/azure-sdk-for-python/tree/main/scripts/breaking_changes_checker

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
l0lawrence: If I remember correctly Swathi had done some work with applying apiview feedback using a chatmode might be worth syncing there to see where that ended up she might have some flow ideas if you havent already

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,116 @@
...
+- [Design Proposal](#design-proposal)
+- [Alternatives Considered](#alternatives-considered)
+
+---
+
+## Definitions
+
+_Terms used throughout this spec with precise meanings:_
+
+- **<a id="typespec"></a>TypeSpec**: The specification language used to define Azure service APIs. SDK code is generated from TypeSpec specifications located in the azure-rest-api-specs repository.
```
benbp: nit: a bunch of empty anchor links that don't appear to be used?

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,116 @@
+# Spec: Customizing Client TypeSpec - [WIP]
```
haolingdong-msft:
```suggestion
# Spec: Customizing Client TypeSpec
```

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,116 @@
...
+_We don't currently have a clear step in our process for identifying breaking changes, but the flow for Go is: Emit SDK -> Analyze changelog & apply fixes -> Repeat._
+
+**All Languages**:
+All language SDKs have their APIs reviewed in API View. Common feedback includes renaming models per language, or changing operations that appear in a client. We have documentation for customizing SDKs such as [renaming types](https://azure.github.io/typespec-azure/docs/howtos/generate-client-libraries/09renaming/), but recent (10/14/2025) testing shows that even with the current documentation, AI agents have difficulty applying these customizations correctly in a consistent manner: particularly when `scope` is required.
+
+There has already been some work done to support identifying and fixing code customizations - customizations made directly to generated SDKs rather than to the TypeSpec files - in the `TspClientUpdateTool`. This tool exists as an MCP tool/CLI command and currently works _after_ an SDK has been generated from TypeSpec.
```
haolingdong-msft: There are also work being done on fixing client.tsp to resolve breaking changes. @chunyu3 @tadelesh @weidongxu-microsoft demoed once in the all hands meeting.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
haolingdong-msft: > AI agents have difficulty applying these customizations correctly in a consistent manner: particularly when `scope` is required.

Is it possible to add an example? Because I thought GenAI could help on updating `client.tsp` given an API-vew comments. So would be very appreciated to see the gaps in real world example.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
weidongxu-microsoft: Agree. This could be a solution to separate concern of "what to change" and "how to change".

Former be language specific and may require information input (output from code analyzer, changelog summary, apiview comment, etc.) and knowledge (what is preferrable in this language, what need to be fixed and what does not, etc.).

Latter be simpler, that more focused on TCGC abilities, and how to apply these features, e.g.
- rename a symbol
- rename a client / operation group
- modify doc on client-side
- change client initialization
- override client method
- etc.

This may help on the "**Nothing shared across languages.**" gap above. Latter part is shareable across languages.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### Gaps
```
weidongxu-microsoft: Wondering whether the availability of "information producer" belong to the gaps.

To be specific, I'd wondering whether it be helpful to have some tools to get the information, for the application of client.tsp customization, e.g.
- "code analyzer result" tool, to get suggestions from .NET code analyzer (may need the ability to run it first)
- "apiview comments" tool, to get conversations in apiview, and summarize what need to change
- "changelog summary" tool, to maybe just read CHANGELOG.md (or run a new calculation on SDK changes, compared to latest stable) and summarize what need to change

PS: if these tools already exist or in design, ignore me

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: Oops - I copied these from an existing spec but you're right, anchors aren't being used.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: I think some of those tools exist, some don't. For example, we have tools to build the generated SDKs, and run validation - so those could potentially trigger the client.tsp customization flow on failure. Once the flow has started, it should at least run build/validation steps to make sure the customizations made things better.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: I'll add some examples - though it's not consistent so sometimes a prompt will work and then the next time it won't 😅 

We do have multiple efforts though to improve this - Swati has a chatmode specifically for making agents better at responding to API view feedback, and Radhika has a markdown file with instructions to smooth the experience for c# because the results without any additional instructions has been inconsistent.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
ronniegeraghty: It's been my assumption that most of the local work requires both the Specs repo and 1-All the SDK Language repos to be cloned locally. 

@praveenkuttappan do we have tools that are currently creating PRs for repos? Or is that done by out tool invoking a pipeline that has the PR creation logic?

chrisradek: For this spec, I'm going to focus the scenario where repos are cloned locally, or git patches can be applied locally. Will defer automatically creating PRs/commits (say from a CI pipeline) for the next phase.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,186 @@
...
+
+**Python**:
+The Python SDK has [breaking changes](https://github.com/Azure/azure-sdk-for-python/tree/main/scripts/breaking_changes_checker) checks that may be resolvable via `client.tsp` customizations, but there is no automated process for identifying or applying these client.tsp customizations today.
+
+**All Languages**:
+All language SDKs have their APIs reviewed in API View. Common feedback includes renaming models per language, or changing operations that appear in a client. We have documentation for customizing SDKs such as [renaming types](https://azure.github.io/typespec-azure/docs/howtos/generate-client-libraries/09renaming/), but recent (10/14/2025) testing shows that even with the current documentation, AI agents have difficulty applying these customizations correctly in a consistent manner: particularly when `scope` is required or customizations are spread over multiple languages/sessions. For example, when creating custom clients using `@client` with different naming for different languages, many times the agent would properly create the custom client for the first language prompted, but then incorrectly use a `@clientName` for the next language prompted.
```
tadelesh: Could you share more details about how the agent gets TCGC docs for customization? I'm a little curious about the problem here is the complicity or the doc quality.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Detailed Design
+
+#### 1. TypeSpec client customizations reference document
+
+A TypeSpec client customizations reference document will serve as the foundation for teaching AI agents about the customizations available via the `Azure.ClientGenerator.Core` library. It includes concise documentation and examples for the decorators available in the library, as well as cover some common scenarios.
+
+This is a living document - it can be updated over time as new decorators are added to the library or new common scenarios are identified. This is _not_ intended to contain guidance for every scenario each language SDK may encounter. For example, Go breaking changes detection looks for specific patterns in the changelog to identify client.tsp customizations. These patterns would not be covered in this document.
+
+This will live in `eng/common/knowledge/customizing-client-tsp.md` so that all language SDK repos can reference it when needed.
```
tadelesh: Would it be possible to have such doc just in the existing TCGC's decorator [doc](https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/decorators/) or client related [doc](https://azure.github.io/typespec-azure/docs/howtos/generate-client-libraries/00howtogen/)? If we could not add the doc to the original source of these decorators, I'm worried about the sync or update things.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: Currently, they are retrieved by the agent when it thinks more info about TypeSpec - there are copilot instructions guiding the agent to read the `llms.txt` files from our official docs sites and this includes the TCGC docs.

However, there is an issue with the quality of the docs - even if the agent is directed to read the TCGC specific docs before doing any work, it still struggles to produce correct results (for example - using decorator `@` notation instead of augment decorator `@@` notation, incorrect scopes, or mixing decorators that shouldn't be mixed)

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: I originally tried using the original sources, and modifying what's generated so that we get better results from the AI agent.

I won't pretend to be an expert on the best way to provide context to the LLM :smile: but the way I went about coming up with a reference was to create several prompts that would require client.tsp customizations, and directly referenced knowledge bases (This included the official generated docs, chatmode, the reference I propose here, and a reference .NET created). I'd run these all (separately) through copilot on the same specs and compare the changes to what was expected. What seemed to get the most consistent results was to cut down on the explanatory text for the decorators, and provide concise examples. Also including scenarios for the common customizations we see - just a handful of decorators make up the majority of customizations.

We can definitely look at how we can change the official doc generation to better match what works for agents, but I think we're still learning what that looks like and can benefit from being able to quickly make tweaks to improve things vs updating the `tspd` and official docs anytime we want to test.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: Resolving now that we've combined the workflows

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: We've updated the design so that the new tool takes in any context. This could be build errors, apiview comments, user request. Then the tool will attempt to address whatever was provided with TypeSpec updates, then SDK updates if things went unaddressed.

Then we can have tools like an apiview comments tool that either has a `nextStep` to invoke the customization tool, or it can call our tool directly in code.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,510 @@
...
+Many of these fixes are handled by adding customizations to `client.tsp` for .NET, such as renaming models with `@clientName`. Currently these fixes are manually applied after SDK builds fail.
+_.NET has a clear flow for identifying and applying `client.tsp` customizations via: Emit SDK -> Build SDK -> Apply fixes -> Repeat._
+
+**Go**:
+The Go SDK has a [breaking changes](https://github.com/Azure/azure-sdk-for-go/blob/instruction/.github/prompts/go-sdk-breaking-changes-review.prompts.md) prompt that analyzes the generated SDK's changelog for specific patterns and maps those to `client.tsp` customizations where possible.
+_We don't currently have a clear step in our process for identifying breaking changes, but the flow for Go is: Emit SDK -> Analyze changelog & apply fixes -> Repeat._
+
+**Python**:
+The Python SDK has [breaking changes](https://github.com/Azure/azure-sdk-for-python/tree/main/scripts/breaking_changes_checker) checks that may be resolvable via `client.tsp` customizations, but there is no automated process for identifying or applying these client.tsp customizations today.
```
swathipil: We have a GH issue today to incorporate the breaking changes tool directly into the Python APIView parser, where breaking public API changes would show up in the diagnostics banner. I believe Travis mentioned that they plan on having APIView Copilot add comments to help resolve diagnostics (such as linting errors), so this could be part of the existing flow from reviews automatically.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,581 @@
...
+
+1. **Context Classifier**
+   Analyzes requests and routes to: Phase A (TypeSpec can help), Success (done), or Failure (too complex/stalled). Phase B is not a classifier decision—it activates automatically on Phase A build failures when customization files exist.
+
+1. **Phase A – [TypeSpec Customizations](#typespec-customizations):**
+   Apply `client.tsp` decorators, regenerate SDK, validate build, return to classifier
+
+1. **Phase B – [Code Customizations](#code-customizations):**
+
+   - **Activation**: Phase A build fails AND customization files exist (Java: `/customization/` or `*Customization.java`, Python: `*_patch.py`, .NET: partial classes)
```
lmazuel, line 146: This raises the question of if at some point, the AI could suggests to create code customization, as it would feel confident there is no other solution.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,581 @@
...
+
+   - **Design Rationale**: Phase B operates under three principles:
+     1. **Spec-First Always**: Only activates after Phase A build failures when customization files exist
+     2. **Narrow Scope**: Mechanical transformations only; uncertain/complex cases get manual guidance. v2 may add pattern-based features while keeping error-driven activation
+     3. **Safety Net**: Handles ~10% (build errors TypeSpec cannot solve); Phase A solves ~80%, remaining ~10% get manual guidance
+
+   - **Scope** (<20 lines, <5 files, deterministic only):
+     - ✅ **In**: Remove duplicates, update references, add imports, rename keywords, update type annotations
+     - ❌ **Out**: Convenience methods, architecture changes, visibility (use `@access`), error handling, complex logic
```
lmazuel, line 155: Curious how this list was built?

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,581 @@
...
+```
+
+**Example 3: Build failure handling**
+
+```json
+{
+  "tool": "azsdk_package_customize_code",
+  "arguments": {
+    "packagePath": "/path/to/sdk",
+    "customizationRequest": "Build failed with: error CS0246: The type or namespace name 'FooModel' could not be found"
```
lmazuel, line 328: Should have a language attached right? Feels like a build fails. Just making sure that we agree that language here is an important part of the context.t

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,581 @@
...
+
+**Entry Point:** API review feedback
+
+**Problem:** Model name doesn't follow .NET casing conventions ("Id" vs "ID").
+
+**Workflow Execution:**
+
+| Phase                 | Action                                                                                                                        | Result                                                              |
+| --------------------- | ----------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------- |
+| **Phase A: TypeSpec** | Analyze feedback requirements<br/>Apply `@clientName` with proper scoping for .NET<br/>Regenerate .NET SDK<br/>Validate build | SDK regenerates successfully<br/>Build passes<br/>No Phase B needed |
```
lmazuel, line 408: Could we have a sample that needs update? Is this in scope?

lmazuel (approved):

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
samvaity: Just going through the existing customization files present in the repo today and some via Typespec capabilities. I think this list can be altered based on the future experiences but mentioned it as an initial starting point.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
samvaity: The package path includes the language indicator, which is how we currently extract language information for our existing tools in MCP.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
samvaity: Yes, via Manual Guidance
The AI should suggest creating customization files, but through the manual guidance route rather than creating files automatically. Because creating a customization file is quite in the zone of architectural decision plus the initial bootstrapping complexity for the customization dir setup etc. Providing it via manual guidance I think gives user the control to decide to introduce customization layer or not.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
samvaity: That's an interesting question. I believe this would be outside the scope of the current specification. Builds can fail if they reference an old model name and need updates, especially if the samples were manually written instead of generated. Generated samples should be covered, but in general, excluding samples from the build might require manual updates or using `azsdk_package_samples_generate` to regenerate them. So we can assume, sample updates are not part of the customization workflow.  I will add it to the scope.

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,621 @@
...
+
+1. **Context Classifier**
+   Analyzes requests and routes to: Phase A (TypeSpec can help), Success (done), or Failure (too complex/stalled). Phase B is not a classifier decision—it activates automatically on Phase A build failures when customization files exist.
+
+1. **Phase A – [TypeSpec Customizations](#typespec-customizations):**
+   Apply `client.tsp` decorators, regenerate SDK, validate build, return to classifier
+
+1. **Phase B – [Code Customizations](#code-customizations):**
+
+   - **Activation**: Phase A build fails AND customization files exist (Java: `/customization/` or `*Customization.java`, Python: `*_patch.py`, .NET: partial classes)
```
timovv, line 146: JavaScript has a somewhat different approach to customization where authors make edits to the generated code directly, and we reconcile those edits with changes to the generated code upon regeneration (so later on) using a 3-way merge. I think the described flow should be able to handle this, but wanted to call it out just in case.

See here https://github.com/Azure/azure-sdk-for-js/wiki/Modular-(DPG)-Customization-Guide

samvaity: yes, I have made a note of that for Javascript being different for the customization file presence detection https://github.com/orgs/Azure/projects/865/views/4?pane=issue&itemId=148287910

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
```diff
@@ -0,0 +1,621 @@
...
+- [ ] **Stall detection**: Tool detects when same error appears twice consecutively
+- [ ] **Context limit**: Tool monitors context size and fails gracefully when exceeding character limits
+- [ ] **Build validation**: All scenarios complete with no build errors (warnings acceptable)
+
+---
+
+## Success Criteria
+
+- [ ] **Automated Detection**: Correctly identify build failure types across all supported languages
+- [ ] **Phase A Success (80% Coverage)**: Successfully apply [TypeSpec Customizations](#typespec-customizations) for 80%+ of feature requests and specification issues
```
l0lawrence, line 576: are we tracking this with telemetry? how are we going to track the files/lines changed?

l0lawrence (approved):

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
chrisradek: With telemetry we can track if client.tsp changes were made, and if the tool succeeds due to SDK regeneration/rebuild passing. We haven't discussed tracking actual diffs yet - that could be interesting to let us see if _we_ think the changes look good overtime too.

timovv (approved):

timovv (approved):

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
m-redding, line 146: In .NET, would it be if any partial classes exist or would a partial class specifically for the "class to customize" need to exist?

m-redding (approved):

tools/azsdk-cli/docs/specs/3-customizing-client-tsp.spec.md
samvaity: For .NET, it would be  any partial classes exist outside of Generated folder.

Issue #12463, opened 2025-10-13
[APIView] UX Feedback - Comment Severity and Bulk Resolution


tjprescott:
Some UX feedback: 
- [x]  the comment header and bulk-resolution UI do not honor the theme (light/dark mode) (https://github.com/Azure/azure-sdk-tools/pull/12479)
- [ ] <del>the "Related" badge should be relocated to the comment header, rather than at the end of the first comment</del>
- [x] the "Related" badge color should be higher-contrast to be easier to read (in light and dark mode)

Some UX design questions:
- [ ] If you are closing a batch of comments, should the UX allow you to up- or downvote the batch and add a comment to the batch? Currently, this UX doesn't concern itself with that, and only deals with resolving issues. Realistically, that means that the first comment will likely get the attention (if any) per out feedback mechanisms, and the others will offer no signal and that itself can skew our metrics.

AlitzelMendez:
Travis, do you know if the feedback that “the ‘Related’ badge should be relocated to the comment header rather than at the end of the first comment” was consistent? I spoke with Crystal about this, and she has some doubts from a UX standpoint. One of her questions was whether it matters to architects to see at a glance how many related comments a comment has when it is collapsed or not expanded,

tjprescott:
@AlitzelMendez I know that Anna and Kashif both favored that, as do I. There was no one that said they preferred it where it was. It was not an exhaustive survey and most architects did not respond at all (which means they likely didn't try it).

AlitzelMendez:
@ccbarragan could you help us with the design question please
`If you are closing a batch of comments, should the UX allow you to up- or downvote the batch and add a comment to the batch? Currently, this UX doesn't concern itself with that, and only deals with resolving issues. Realistically, that means that the first comment will likely get the attention (if any) per out feedback mechanisms, and the others will offer no signal and that itself can skew our metrics.`


you can see the related issues feature in here: https://spa.apiviewstagingtest.com/review/96672963bb5747db9d65d32df5f82d5a?activeApiRevisionId=2668702652604dad92d75feb1321c7a4

I think is valid to upvote/downvote the batch? can this overcomplicate the ux?

ccbarragan:
[...]

PR #12587, opened 2025-10-21
[API View] Bulk Comment Resolution feedback

AlitzelMendez:
closes: https://github.com/Azure/azure-sdk-tools/issues/12463

Enables users to resolve multiple related comments simultaneously with optional voting and reply messages.

## Features
- **Batch Resolution**: Resolve multiple comments in one operation
- **Batch Voting**: Apply votes during resolution
- **Resolution Comments**: Add optional reply when resolving which is applied to every comment

## Demo
https://github.com/user-attachments/assets/69e50b2d-2d93-433c-8ec4-1dfa7fedd642

### DIFF
--- a/src/dotnet/APIView/ClientSPA/src/app/_components/shared/related-comments-dialog/related-comments-dialog.component.html
+++ b/src/dotnet/APIView/ClientSPA/src/app/_components/shared/related-comments-dialog/related-comments-dialog.component.html
@@ -93,29 +93,27 @@
     </div>
 
     <div class="resolution-comment-section mt-4">
-      <!-- Feedback Section -->
-      <div class="feedback-buttons-section mb-3">
-        <div class="d-flex align-items-center vote-buttons">
+      <div class="d-flex align-items-center justify-content-between mb-2">
+        <label class="form-label fw-semibold mb-0">
+          Message (Optional):
+        </label>
+        <div class="d-flex align-items-center vote-section">
           <button 
             type="button"
-            class="btn btn-link p-1 vote-btn"            
+            class="btn btn-link p-2 vote-btn"
             title="Upvote selected issues"
             (click)="toggleBatchVote('up')">
-            <span [ngClass]="hasBatchUpvote() ? 'bi bi-hand-thumbs-up-fill emoji-active' : 'bi bi-hand-thumbs-up text-muted'"></span>
+            <span [ngClass]="hasBatchUpvote() ? 'bi bi-hand-thumbs-up-fill emoji-active fs-5' : 'bi bi-hand-thumbs-up text-muted fs-5'"></span>
           </button>
           <button 
             type="button"
-            class="btn btn-link p-1 vote-btn"
+            class="btn btn-link p-2 vote-btn"
             title="Downvote selected issues"
             (click)="toggleBatchVote('down')">
-            <span [ngClass]="hasBatchDownvote() ? 'bi bi-hand-thumbs-down-fill emoji-active' : 'bi bi-hand-thumbs-down text-muted'"></span>
+            <span [ngClass]="hasBatchDownvote() ? 'bi bi-hand-thumbs-down-fill emoji-active fs-5' : 'bi bi-hand-thumbs-down text-muted fs-5'"></span>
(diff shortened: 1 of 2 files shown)


src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -453,5 +475,32 @@ private async Task AssertOwnerAsync(ClaimsPrincipal user, CommentItemModel comme
...
+                        comment.Downvotes.RemoveAll(u => u == userName);
+                    }
+                    break;
+                case FeedbackVote.Down:
+                    if (comment.Downvotes.RemoveAll(u => u == userName) == 0)
+                    {
+                        comment.Downvotes.Add(userName);
+                        comment.Upvotes.RemoveAll(u => u == userName);
+                    }
+                    break;
```
tjprescott, line 497: If I read this correctly, it seems like if there's an up or down vote, we clear any existing votes and replace it. I don't think we should do that. Since APIVIew actually just tracks the names, not the actual count, we should just add the name of whoever is doing the resolution to the relevant collection (up or down) if it's not already there.

So like if you upvoted something and then I bulk downvote and it includes the one you upvoted, in the end it should show one upvote (yours) and one downvote (mine).

src/dotnet/APIView/ClientSPA/src/app/_components/shared/related-comments-dialog/related-comments-dialog.component.html
```diff
@@ -17,17 +17,38 @@
...
+            (click)="toggleBatchVote('up')">
+            <span [ngClass]="hasBatchUpvote() ? 'bi bi-hand-thumbs-up-fill emoji-active fs-5' : 'bi bi-hand-thumbs-up text-muted fs-5'"></span>
+          </button>
+          <button 
+            type="button"
+            class="btn btn-link p-2 vote-btn"
+            title="Downvote selected issues"
+            (click)="toggleBatchVote('down')">
+            <span [ngClass]="hasBatchDownvote() ? 'bi bi-hand-thumbs-down-fill emoji-active fs-5' : 'bi bi-hand-thumbs-down text-muted fs-5'"></span>
+          </button>
```
tjprescott: I think these should be moved to be below related commenst and just above the reply box.

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
AlitzelMendez: this is per user, so basically it checks your user like

Ali upvote this already (specifically Ali) yes? if she did let's remove that upvote and downvote that

but we can't put Ali in both places

src/dotnet/APIView/ClientSPA/src/app/_components/shared/related-comments-dialog/related-comments-dialog.component.html
AlitzelMendez: I did this change:

image

what do you think?

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
tjprescott: Got it! Thanks for the explanation!

src/dotnet/APIView/ClientSPA/src/app/_components/shared/related-comments-dialog/related-comments-dialog.component.html
tjprescott: I might suggest making it flush right so it's located generally where you see it normally, but it looks better to me!

AlitzelMendez: sure!! :D

AlitzelMendez: image

changed :)

tjprescott (approved):

PR #12563, opened 2025-10-20
Added spec for 'package updater' tools

raych1:
- Ported the [original docs specs](https://microsoft-my.sharepoint.com/:w:/p/raychen/IQAhDpImqdGbSIqU7iXKHUOqAaQ8j7Rw_sGUOtxjq55pxwc?e=ukoWtP)  into the new spec layout.

- Added new content and clarifications to match the updated format.

- See issue https://github.com/Azure/azure-sdk-tools/issues/11827 for scope and follow-ups.

PR #12608, opened 2025-10-22
Added update-changelog tool

raych1:
Key changes:
- Changelog Update Tool: Automates changelog content generation for SDK packages

- Configuration-Based Tool: Reusable base class for tools requiring configuration-driven operations

- Enhanced Configuration Helper: Extended `SpecGenSdkConfigHelper` with generic configuration support for multiple tool types

- Test Coverage


tools/azsdk-cli/Azure.Sdk.Tools.Cli/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 0.5.4 (2025-10-22)
```
praveenkuttappan: 0.5.4 was already released. You will need to rebase and add this under 0.5.5

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/SpecGenSdkConfigHelper.cs
```diff
@@ -15,7 +15,8 @@ public interface ISpecGenSdkConfigHelper
     {
         // Config value retrieval methods
         Task<T> GetConfigValueFromRepoAsync<T>(string repositoryRoot, string jsonPath);
-        Task<(BuildConfigType type, string value)> GetBuildConfigurationAsync(string repositoryRoot);
+        Task<(ConfigContentType type, string value)> GetBuildConfigurationAsync(string repositoryRoot);
```
praveenkuttappan, line 18: Why can't we use just GetConfigurationAsync from caller instead of adding a wrapper for get build config which is just calling GetConfigurationAsync?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,197 @@
...
+
+        /// <summary>
+        /// Updates the changelog for the specified package.
+        /// For management-plane packages: invokes changelog generation script.
+        /// For data-plane packages: returns guidance for manual editing.
+        /// </summary>
+        /// <param name="packagePath">Absolute path to the SDK package directory.</param>
+        /// <param name="ct">Cancellation token for the operation.</param>
+        /// <returns>A response indicating the result of the changelog update operation.</returns>
+        [McpServerTool(Name = "azsdk_package_update_changelog_content"), Description("Update changelog content for an SDK package.")]
```
praveenkuttappan, line 59: Can we shorten this tool name to just ` azsdk_package_update_changelog`. Updating a change log implies updating it's contents.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,197 @@
...
+                NextSteps = ["Update the version if it's a release."]
+            };
+        }
+
+        /// <summary>
+        /// Determines the SDK repository root from the package path.
+        /// </summary>
+        /// <param name="packagePath">The package path to start searching from.</param>
+        /// <returns>The SDK repository root path, or null if not found.</returns>
+        private string? DetermineSdkRepositoryRoot(string packagePath)
```
praveenkuttappan: Can you move this to PackageHelper. I think this will be helpful for other tools.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,197 @@
...
+            return null;
+        }
+
+        /// <summary>
+        /// Determines the package type (mgmt vs data-plane) for the specified package.
+        /// </summary>
+        /// <param name="packagePath">The package path.</param>
+        /// <param name="sdkRepoRoot">The SDK repository root.</param>
+        /// <returns>The package type string.</returns>
+        private string DeterminePackageType(string packagePath, string sdkRepoRoot)
```
praveenkuttappan, line 151: Same here.  It will be helpful to move this also to helper class if other tools need to know the SDK type

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
timovv: We already have a helper for this in GitHelper

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,197 @@
...
+        {
+            // Prepare script parameters
+            var scriptParameters = new Dictionary<string, string>
+            {
+                { "SdkRepoPath", sdkRepoRoot },
+                { "PackagePath", packagePath }
+            };
+
+            // Create and execute process options for the update-changelog script
+            var processOptions = await CreateProcessOptions(ConfigType.UpdateChangelog, sdkRepoRoot, packagePath, scriptParameters, 15);
```
praveenkuttappan, line 125: Why do you need to pass a dictionary if repo path and package path and same values as separate params too?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// Handles changelog guidance for data-plane packages.
+        /// </summary>
+        /// <returns>A noop response with guidance for manual editing.</returns>
+        private DefaultCommandResponse HandleDataPlaneChangelog()
+        {
+            logger.LogInformation("Data-plane package detected, providing manual editing guidance");
+            
+            return new DefaultCommandResponse
+            {
+                Result = "noop",
```
praveenkuttappan, line 139: nit: should this be more readable Result if it's passed to user or agent? I am not sure how is `noop` interpreted on agent side.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ConfigBasedTool.cs
```diff
@@ -0,0 +1,354 @@
...
+using Azure.Sdk.Tools.Cli.Models;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Sdk.Tools.Cli.Tools.Package;
+
+/// <summary>
+/// Base class for tools that need configuration-based operations, process execution, and common response patterns.
+/// Provides shared functionality for configuration management, validation, response creation, and process execution.
+/// </summary>
+public abstract class ConfigBasedTool : MCPTool
```
timovv, line 12: Seems like this class consists mainly of helper functions. Maybe it would be beneficial to convert the helpers into a helper class so that any helpers can be used without needing to inherit from `ConfigBasedTool` if we think they'll be used elsewhere? Otherwise, if we think these are one-offs, we could also just keep the helpers as private in the ChangelogUpdateTool.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ConfigBasedTool.cs
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    /// <summary>
+    /// Creates a failure response with the specified error message.
+    /// Works with any response type that inherits from CommandResponse.
+    /// </summary>
+    /// <typeparam name="T">The response type that inherits from CommandResponse.</typeparam>
+    /// <param name="message">The error message to include in the response.</param>
+    /// <returns>A response indicating failure.</returns>
+    protected T CreateFailureResponse<T>(string message) where T : CommandResponse, new()
```
timovv: I don't know if this helper is necessary and if it is maybe it could be constrained to the ChangelogUpdateTool specifically.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ConfigBasedTool.cs
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    /// <summary>
+    /// Validates that the specified file path exists.
+    /// </summary>
+    /// <typeparam name="T">The response type that inherits from CommandResponse.</typeparam>
+    /// <param name="filePath">The file path to validate.</param>
+    /// <param name="fileDescription">Description of the file for error messages (e.g., "configuration file").</param>
+    /// <returns>A failure response if validation fails, null if validation passes.</returns>
+    protected T? ValidateFilePath<T>(string filePath, string fileDescription = "file") where T : CommandResponse, new()
```
timovv: Again I don't think this helper is providing much value.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
praveenkuttappan: Any reason not to use DiscoverRepoRoot in GitHelper?
https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs#L177C21-L177C38?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,197 @@
...
+                    logger.LogInformation($"Detected SDK repository root at: {currentPath}");
+                    return currentPath;
+                }
+                
+                currentPath = Path.GetDirectoryName(currentPath);
+            }
+
+            logger.LogWarning($"Could not determine SDK repository root from package path: {packagePath}");
+            return null;
+        }
```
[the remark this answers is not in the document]
praveenkuttappan: It's unused so you should remove this code.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// <param name="packagePath">The package path.</param>
+        /// <param name="sdkRepoRoot">The SDK repository root.</param>
+        /// <returns>The package type string.</returns>
+        private string DeterminePackageType(string packagePath, string sdkRepoRoot)
+        {
+            // Analyze path structure for common patterns
+            var relativePath = Path.GetRelativePath(sdkRepoRoot, packagePath);
+
+            // Common patterns for management-plane packages
+            if (relativePath.Contains("arm-", StringComparison.OrdinalIgnoreCase) ||
```
praveenkuttappan, line 157: This can be a problem. What is a service or data plane package is added with prefix `farm-` and this code will treat it as management. Individual language repo already has a logic to identify whether a package is mgmt or client. Can we reuse it or use the same logic?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,197 @@
...
+                relativePath.Contains("mgmt-", StringComparison.OrdinalIgnoreCase) ||
+                relativePath.Contains("resourcemanager", StringComparison.OrdinalIgnoreCase))
+            {
+                logger.LogInformation("Package type determined from path: mgmt");
+                return "mgmt";
+            }
+
+            // Default to data-plane if no clear indicators
+            logger.LogInformation("Package type defaulted to: data-plane");
+            return "data-plane";
```
praveenkuttappan, line 167: I suggest changing this to `client` to match sdk type we use everywhere else in engsys.

timovv: Maybe SDK type should be an enum?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
timovv: +1, I think we need to use language-specific logic for determining whether a package is ARM or data plane. Having this be heuristic-based where we check the path for a few different words seems brittle.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/SpecGenSdkConfigHelper.cs
raych1: `GetBuildConfigurationAsync` was added before and I just keep it for compatibility. It's planned to refactor `SdkBuildTool` to use the new method `GetConfigurationAsync` in next PR.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
raych1: We intended to use `azsdk_package_update_changelog_content` since it won't change `release date` and `version`. Refer to the spec PR for the discussion.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
raych1: Thanks. I understand it's a bit redundant and we can extract them from the dictionary. 
The dictionary is intentionally meant to be the parameters passed to the external language script. The separate `sdkRepoRoot/packagePath` parameters are there for internal use and keep the method contract explicit.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
raych1: Agreed it's not friendly to human while it's actually friendly to the agent. We will have actual operation instead of 'noop' in next version.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ConfigBasedTool.cs
raych1: We do have `specGenSdkConfigHelper` for config parsing/substitution, and the choice to add `ConfigBasedTool` was intentional to remove a lot of repeated boilerplate across five tools (`update-changelog, update-metadata, update-ci, update-version, build`). The base class centralizes the Process orchestration, shared validation, standardized responses.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
raych1: I couldn't find any existing code that determines the SDK type. The only place interacting with the SDK Type is the release stage code. Can you share me with the location if we have it in other places? @praveenkuttappan @timovv

maririos:
@raych1 to be clear, this PR implementation depends on https://github.com/Azure/azure-sdk-tools/pull/12563 approval, right?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,170 @@
...
+            IProcessHelper processHelper,
+            IGitHelper gitHelper)
+            : base(specGenSdkConfigHelper, logger, processHelper)
+        {
+            this.gitHelper = gitHelper;
+        }
+
+        public override CommandGroup[] CommandHierarchy { get; set; } = [SharedCommandGroups.Package];
+
+        private const string UpdateCommandName = "update-changelog";
```
samvaity, line 37: Should we make this as a subcommand group? `azsdk package update --changelog` So that we can potentially extend it for other metadata updates, `azsdk package update --readme` etc?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ConfigBasedTool.cs
```diff
@@ -0,0 +1,354 @@
...
+        };
+    }
+
+    /// <summary>
+    /// Validates that the specified package path exists and is accessible.
+    /// </summary>
+    /// <typeparam name="T">The response type that inherits from CommandResponse.</typeparam>
+    /// <param name="packagePath">The package path to validate.</param>
+    /// <returns>A failure response if validation fails, null if validation passes.</returns>
+    protected T? ValidatePackagePath<T>(string packagePath) where T : CommandResponse, new()
```
samvaity, line 73: Should we move this to a common helper, I know we use `packagePath` in many of our existing tools but don't necessarily validate the path.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,170 @@
...
+                if (validationError != null) 
+                {
+                    return validationError;
+                }
+
+                // Get SDK repository root
+                string sdkRepoRoot = gitHelper.DiscoverRepoRoot(packagePath);
+                if (string.IsNullOrEmpty(sdkRepoRoot))
+                {
+                    return CreateFailureResponse<DefaultCommandResponse>($"Failed to discover local sdk repo with packagePath: {packagePath}.");
```
samvaity, line 80: nit: This can be simplified to.
`return new DefaultCommandResponse { ResponseError = $"Failed to discover local sdk repo with packagePath: {packagePath}.}" };`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
samvaity: +1 on @praveenkuttappan's logic here "`farm`- and this code will treat it as management."
This would be incorrect. Also, it could lead to misclassification if naming conventions aren’t followed consistently across SDKs.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
samvaity: +1 on enum for SDK type.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// <returns>A noop response with guidance for manual editing.</returns>
+        private DefaultCommandResponse HandleDataPlaneChangelog()
+        {
+            logger.LogInformation("Data-plane package detected, providing manual editing guidance");
+            
+            return new DefaultCommandResponse
+            {
+                Result = "noop",
+                Message = "Data-plane changelog untouched; manual edits required.",
+                NextSteps = ["Update the version if it's a release."]
```
samvaity, line 141: The next steps here should be more elaborate since manual intervention is needed. Since we dont provide auto-fix we should provide helpful guidance here imo.

raych1:
> @raych1 to be clear, this PR implementation depends on #12563 approval, right?

@maririos , that's correct. This PR depends on the spec approval.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ChangelogUpdateTool.cs
benbp: I had been thinking this could follow a similar approach to what we do with checks:

```
azsdk package update  # defaults to all
azsdk package update changelog
azsdk package update ...
```

However typing this out I think `update` is maybe a misleading/overloaded verb to use here. If I read `package update` I think "get a new package version."

raych1:
Closed this PR and replaced it by a [new PR](https://github.com/Azure/azure-sdk-tools/pull/12821)

PR #12609, opened 2025-10-22
Add tool status in response

praveenkuttappan:
Add a tool status property in tool response

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Helpers/OutputHelperTests.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Helpers/OutputHelperTests.cs
@@ -41,7 +41,7 @@ internal class OutputHelperTests
     }
   ],
   ""suggested_fix"": ""a test suggested fix"",
-  ""status"": ""Succeeded""
+  ""operation_status"": ""Succeeded""
 }";
 
         var output = new OutputHelper(OutputHelper.OutputModes.Json);
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
@@ -53,8 +53,8 @@ public abstract class CommandResponse
     /// <summary>
     /// Status shows whether the command operation was successful.
     /// </summary>
-    [JsonPropertyName("status")]
-    public Status CommandStatus
+    [JsonPropertyName("operation_status")]
+    public Status OperationStatus
     {
         get
         {


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs
```diff
@@ -144,7 +144,7 @@ public async Task<SdkReleaseResponse> ReleasePackageAsync(string packageName, st
                     PackageName = packageName,
                     Language = language,
                     ReleasePipelineStatus = "Failed",
-                    ReleaseStatusDetails = $"Error: {ex.Message}"
+                    ResponseError = $"Error: {ex.Message}"
```
benbp, line 147: FYI in case it's relevant, if response error is set then we only print that during `ToString()`. For json serialized to MCP it will still send back all properties.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
```diff
@@ -44,6 +44,18 @@ public int ExitCode
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public List<string>? NextSteps { get; set; }
 
+    /// <summary>
+    /// ToolSucceeded shows whether the tool operation was successful.
+    /// </summary>
+    [JsonPropertyName("tool_succeeded")]
```
benbp: Do we want this to be `status`/`result` maybe? Then we can start with `succeeded` and `failed` but I could see things like `timeout` / `cancelled` etc. being useful in the future.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Helpers/OutputHelperTests.cs
```diff
@@ -40,7 +40,8 @@ public void TestTypedJsonOutput()
       ""message"": ""message2""
     }
   ],
-  ""suggested_fix"": ""a test suggested fix""
+  ""suggested_fix"": ""a test suggested fix"",
+  ""status"": 0
```
benbp: The json property should still resolve to the `enum.ToString()` value so it shows up in the JSON as `"succeeded"` etc.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs
praveenkuttappan: I am a bit confused. As long as ResponseError is passed in MCP json response, we are fine. We also have pipelinestatus field in this case.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
```diff
@@ -44,6 +50,18 @@ public int ExitCode
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public List<string>? NextSteps { get; set; }
 
+    /// <summary>
+    /// Status shows whether the command operation was successful.
+    /// </summary>
+    [JsonPropertyName("command_status")]
```
benbp: Can this not just be `status` or `result`? Then it makes sense for both a command and tool invocation context.

praveenkuttappan: Nope. There is a status field in one of the sub response so `Status` field will hide that status.

praveenkuttappan: I can change it to result

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Helpers/OutputHelperTests.cs
```diff
@@ -40,7 +40,8 @@ public void TestTypedJsonOutput()
       ""message"": ""message2""
     }
   ],
-  ""suggested_fix"": ""a test suggested fix""
+  ""suggested_fix"": ""a test suggested fix"",
+  ""command_status"": ""Succeeded""
```
praveenkuttappan:
```suggestion
  ""result"": ""Succeeded""
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
praveenkuttappan:
```suggestion
    [JsonPropertyName("result")]
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
```diff
@@ -44,6 +50,18 @@ public int ExitCode
...
+    public Status CommandStatus
```
praveenkuttappan:
```suggestion
    public Status Result
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
praveenkuttappan: I have changed the json response to status. Unfortunately, both status and result fiels are already used by sub responses so I cannot use it without changing them.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
```diff
@@ -44,6 +50,18 @@ public int ExitCode
...
 
+    /// <summary>
+    /// Status shows whether the command operation was successful.
+    /// </summary>
+    [JsonPropertyName("status")]
+    public Status CommandStatus
+    {
+        get
+        {
+            return string.IsNullOrEmpty(ResponseError) && (ResponseErrors == null || ResponseErrors.Count == 0)? Status.Succeeded : Status.Failed;
```
samvaity, line 61: Just to be clear, this property cannot be set but only supposed to be inferred form ResponseError property?

benbp: I could see scenarios where we want to set failure but don't have an error message, or in the future if we want to set some 3rd status field, however for now I think it's simpler if there's only a single way to indicate we've failed and that's to set the error property.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
benbp: Ok if it's already overloaded then I guess whatever works is fine. Maybe `operation_status` or just `tool/command`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
praveenkuttappan: Correct. We can change it to setter later if we want.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
praveenkuttappan: Done.

PR #12612, opened 2025-10-22
Enable test proxy; build in setup phase not run phase

LarryOsterman:


tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
```diff
@@ -76,10 +87,19 @@ public override async Task<IterationResult> RunAsync(
...
+            // If --test-proxy is in the command line arguments, set AZURE_TEST_MODE environment variable to "playback" to indicate that we are going to use the test proxy.
+            // Otherwise set test mode to `live`.
+            if (arguments.Contains("test-proxy"))
+            {
+                Environment.SetEnvironmentVariable("AZURE_TEST_MODE", "playback");
+            }
+            else
+            {
+                Environment.SetEnvironmentVariable("AZURE_TEST_MODE", "live");
+            }
```
heaths, line 139: These cause a rebuild - at least for recorded tests. That's part of how the `#[recorded::test]` attribute works. Not sure if/how you'd change that. So this really needs to be during the build phase. But if you *only* need these at runtime, sure, this works.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
```diff
@@ -76,10 +87,19 @@ public override async Task<IterationResult> RunAsync(
...
+            // set up the params for cargo. Some of the arguments were established in the setup stage.
+            string finalParams = $"test {_testCommand} -- --test-results {_targetResultsDirectory}/{testName}-results.json \"{testName}\" {arguments}";
```
heaths: I don't think we should run it through `cargo test` if you already built it. That said, perhaps figuring out the path to the test isn't worth it. The initial run will take a little longer until `cargo` figures out it doesn't need to build (even in WSL this is noticeable) so that might skew results. Unfortunately, it doesn't seem there's a `cargo test --no-build` option (there's a `cargo test --no-run` that just builds tests?!) that computes the path for you.

Also take into account my previous comment about those env vars. As is, I'm worried this PR didn't actually move the needle.

heaths (changes requested):

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
heaths: We could build with JSON and then process the output to get the final executable, something like this (in bash using `jq`):

```bash
cargo build -p azure_core --test perf --message-format json | jq -s '.[] | select(.executable) | .executable'
```

Note that JSON messages go to `stdout` and there will be multiple JSON objects, which is why I use `jq -s` (slurp) here.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
LarryOsterman: Yeah, I know that. However, the only thing that gets rebuilt is the actual tests, not the entire package. That saves a *significant* amount of time running the tests in a loop (which is the default).

And it only triggers a rebuild if the `--debug` switch is set, which is never the case in the perf tests.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
LarryOsterman: I chose to use `cargo test` for exactly that reason: Figuring out the name of the binary which was built is a non-trivial operation. I wish that wasn't the case :(.

Otherwise, I'd be using the binary directly.

And to be clear: This PR isn't about moving the needle on the perf tests - it won't. It's about cleaning up the actual tests, and enabling support for the existing `--test-proxy` command line switch.

After we GA, I'm hoping to add support for generating a local cargo.toml file in the `setup` phase which will allow us to do build-to-build comparisons of the packages (similar to what is done for all the other languages). But in order to do that, these changes need to be made first.

RickWinter (approved):

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
heaths: It rebuilds `azure_core_test` on up, so that's at least 3 crates. Per our offline discussion, I think we could probably avoid this, though.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
LarryOsterman: Current PR uses --message-format to calculate the binary name.

mikeharder (approved):

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
```diff
@@ -60,10 +65,55 @@ public UtilEventArgs(string methodName, string[] methodParams)
...
+            _testCommand = $"--{flavor} --package {primaryPackage} --test perf ";
+            var result = await Util.RunAsync(_cargoName, $"build {_testCommand} --message-format=json", WorkingDirectory, log: false);
+            // Look for errors in the build.
+            if (result.ExitCode != 0)
+            {
+                return (result.StandardOutput, result.StandardError, null);
+            }
+
+            Console.WriteLine("Parsing build output to find test executables...");
+            string[] json_elements = result.StandardOutput.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
```
heaths, line 84: I knew new lines are platform-dependent for text files, but - perhaps a silly question - is the same true of stdout? I guess I never really thought much of it. IMO, I usually just split on "\n" in the few times I'd had to process stdout as text and expect new lines.

heaths: I mean, I assume you tested this on your Windows machine, so if it worked there, it'll work anywhere (Windows is the weirdest in this case).

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
```diff
@@ -60,10 +65,55 @@ public UtilEventArgs(string methodName, string[] methodParams)
...
+                        }
+                    }
+                }
+                catch (JsonException e)
+                {
+                    Console.WriteLine($"Exception {e} parsing line as JSON, skipping line.");
+                    // Ignore lines that are not valid JSON
+                }
+            }
+            if (executables.Length != 1)
```
heaths, line 105: Not a bad check. Alternatively, though, I'd expect the last executable to be "the one". Whatever works.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
```diff
@@ -76,24 +126,56 @@ public override async Task<IterationResult> RunAsync(
...
+            {
+                Environment.SetEnvironmentVariable("AZURE_TEST_MODE", "playback");
+            }
+            else
+            {
+                Environment.SetEnvironmentVariable("AZURE_TEST_MODE", "live");
+            }
             ProcessResult result = new ProcessResult(0, String.Empty, String.Empty);
-            if (IsTest)
+            if (_executablePath == String.Empty)
```
heaths, line 141: Couldn't we just fail? Or is this a common enough case it's nice to fall back. Is time measured from the start of `RunAsync` being called on us?

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
LarryOsterman: Works fine on Windows, I'll be honest and say I didn't test on Linux, but I should.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
LarryOsterman: I thought about that - there should never be more than one executable generated, so I *think* this is safe.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
LarryOsterman: I don't know - I was trying to be extra cautious here in case the logic above didn't work. But failing is an option as well.

And time is measured from the start of the run loop in the test application. Time in the perf automation app is irrelevant.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
heaths: I'm not worried about linux, which uses "\n" anyway. My understanding is that `WriteFile` in text mode in Windows (and, therefore, anything that uses it) translates "\n" in output streams to "\r\n" anyway. Then there's Mac which - though I'm not actually sure if it's true today with OSX/macOS - uses "\r" only.

PR #12615, opened 2025-10-22
[azsdkcli] update spec for validation

l0lawrence:


timovv (approved):

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
```diff
@@ -267,10 +267,9 @@ Run all PR checks locally for the Health Deidentification SDKs before I create p
 ```
 
 **Expected Agent Activity:**
-1. Execute `run-pr-checks` for all 5 languages
-2. Run build, test, lint, breaking change detection
-3. Report which checks passed and which failed
-4. Provide guidance on fixing any failures
+1. Execute `run-checks` for all 5 languages
+1. Report which checks passed and which failed
+1. Provide general guidance on fixing failures
```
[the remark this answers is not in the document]
timovv: Technically Markdown doesn't care about this and honestly I prefer it since it's easier to insert something in the middle of the list this way.

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
```diff
@@ -145,8 +145,8 @@ _If you create a PR to compare the generated library with the existing library a
    └─ Update versions, changelogs, READMEs, metadata files
    └─ Validate: versions, READMEs, changelogs
 
-4. Validating → run-pr-checks
-   └─ Run all PR CI checks locally to ensure green PR
+4. Validating → run-checks
+   └─ Run commonly failing PR CI checks locally to ensure green PR
```
samvaity: We have the ability to run all checks using the validation tool, so why is this change necessary?

l0lawrence: we have a majority but for milestone 1 we dont have all the ci checks -- path length too long for example is left out

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
```diff
@@ -379,41 +378,43 @@ Updating package metadata for Health Deidentification SDK...
...
-### 4. Run PR Checks
+### 4. Run Validation Checks
 
-> Sample: will be changed when the spec for this tool is added
 
 **Command:**
 ```bash
-azsdk run-pr-checks --service healthdataaiservices --languages .NET,Java,JavaScript,Python,Go
+# Run all checks (default)
+azsdk package run-checks --package-path <path_to_sdk_package>/
```
benbp: I have this as `azsdk package validate` in https://github.com/Azure/azure-sdk-tools/pull/12564 but alternatively `azsdk package checks` ?

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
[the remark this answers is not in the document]
benbp: Agree, I always use `1. 1. 1.`

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
samvaity: We should just list the checks covered then in that case.

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
samvaity: I prefer this too `azsdk package validate`

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
```diff
@@ -379,41 +378,43 @@ Updating package metadata for Health Deidentification SDK...
...
+
+# Run specific check type
+azsdk package run-checks <check-type> --package-path <path_to_sdk_package>/
```
samvaity: we should also include the command for the option fixErrors?

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
```diff
@@ -379,41 +378,43 @@ Updating package metadata for Health Deidentification SDK...
...
+- `all`: Run all available validation checks (default)
+- `changelog`: Run changelog validation check
+- `dependency`: Run dependency analysis check  
+- `readme`: Run README validation check
+- `cspell`: Run spelling validation check
+- `snippets`: Run snippet update check
+- `linting`: Run code linting check
+- `format`: Run code formatting check
+- `checkaotcompat`: .NET AOT compatibility validation
+- `generatedcodechecks`: .NET generated code validation
```
samvaity, line 407: Just curious, how is this different from other checks ?

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
```diff
@@ -379,41 +378,43 @@ Updating package metadata for Health Deidentification SDK...
...
+- `samples`: Validate samples
```
samvaity: Is this V1 milestone, don't seem to have language specific tasks yet on this?

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
l0lawrence: we do below

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
l0lawrence: validate sounds good to me

l0lawrence: sorry I forgot about the rename pr ben :)

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
l0lawrence: Oh yes I wasn't sure on that one b/c fixing errors isn't in milestone 1. I could add with a (future milestone) comment or such

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
m-redding: We have a script that is run in our CIs that does a bunch of stuff - https://github.com/Azure/azure-sdk-for-net/blob/main/eng/scripts/CodeChecks.ps1 and it didn't make sense to rewrite them and split them up at this point

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
l0lawrence: yes it is it should be on each language-specific issue in the descr., as discussed in the sync I believe* it only applies to JS here: https://github.com/Azure/azure-sdk-tools/blob/300e6499e42b4f53fb4fca688a4b04d14ab5d8ae/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageSpecificChecks.cs#L30. Python for ex just runs mypy under our linting step and that covers our samples

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
maririos: is it only for generated code? or for custom code too.
(are there any document or settings checks)?

l0lawrence: @m-redding would be able to explain best, but from my understanding .NET lumps all their checks together under these scripts in CI so this was the reflection of that

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
m-redding: Just to make sure I'm understanding each call to the tool does checks for one language, right?

m-redding (approved):

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
m-redding: It regenerates code, exports the API, checks the solution file is correct, re-generates the snippets and does additional .NET specific README validations beyond the common script

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
l0lawrence: yup! in the future if there is a common path of doing all langs at once we can make it take in a list of paths etc.

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
samvaity: From above what you noted @m-redding I do feel this is more like a release-readiness check than a validation check then?
Also the current naming makes it seem focused only for generated code?

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
samvaity: Same case as python for Java. Wanted to make sure we note it, if it is only applicable to JS, should we add it as
>- `samples`: Validate samples for JS or something similar as we did for only.NET checks?

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
m-redding: Hmm I'm not sure, it runs in the CIs for every PR

m-redding: It does more than check generated code but code checks felt too vague to me. I'm not stuck on that name if someone has a better idea

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
```diff
@@ -263,14 +263,13 @@ Update the package metadata for Health Deidentification SDKs to prepare for a pr
 
 **Prompt:**
 ```
-Run all PR checks locally for the Health Deidentification SDKs before I create pull requests.
+Run all validation checks locally for the Health Deidentification SDKs before I create pull requests.
```
samvaity: nit: Given we wanted to update `all` checks I would be consistent with that terminology.

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
l0lawrence: can do!

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
```diff
@@ -145,8 +145,8 @@ _If you create a PR to compare the generated library with the existing library a
    └─ Update versions, changelogs, READMEs, metadata files
    └─ Validate: versions, READMEs, changelogs
 
-4. Validating → run-pr-checks
-   └─ Run all PR CI checks locally to ensure green PR
+4. Validating → run-checks
+   └─ Run commonly failing validation checks locally to ensure green PR
```
samvaity:
```suggestion
   └─ Run validation checks locally to ensure green PR
```

samvaity (approved):

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
timovv:
```suggestion
- `samples`: Validate samples for JS
```

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
maririos: I wonder if we expect language specific checks, if that should be it's own group so that people don't have to parse names, they can either call `azsdk package run-checks -languageSpecific .net`  or something like that
(could also happen later if we see it is a valid scenario)

maririos: `dotnetchecks` ?

haolingdong-msft:
Thanks @l0lawrence for the spec. From my understanding, seems the logics are cross dp and dp? Is there any work needed from mgmt side?

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
weidongxu-microsoft: Java had a similar validation step, for SDK generated from TypeSpec source.
https://github.com/Azure/azure-sdk-for-java/blob/main/eng/scripts/Compare-CurrentToCodegeneration.ps1#L169-L207

It maybe not as complex as .NET. Basically it does 2 things
1. call `tsp-client update` to re-generate code from tsp-location.yaml, this verifies that TypeSpec is up-do-date.
2. call `git diff` to check if there is difference when code is freshly generated. It verifies that there is no manual code change (we'd expect correct customization that be recognized by emitter, but not dev ad hoc edit of code that would be overridden by next invocation of emitter).

maririos: @l0lawrence feel free to add them individually for now, or separate them into a standard design for the languages

m-redding: I think for .NET this actually relates to the longer discussion here - https://github.com/Azure/azure-sdk-tools/pull/12563#discussion_r2449474956

My suggestion is leaving this as-is (or doing the language thing - I don't have a preference) for the first scenario. I would like to do some work to adjust/split up a lot of these scripts that we have in the .NET repo to be more friendly for running in multiple contexts (local, cli, pipelines, etc). Because right now our scripts are really written only for the pipelines. But I would prefer to do this at the same time as any work from the outcome of the discussion above if possible.

l0lawrence:
> Thanks @l0lawrence for the spec. From my understanding, seems the logics are cross dp and dp? Is there any work needed from mgmt side?

For validation checks we are just running what we have running in CI for each language ([verify-readme, verify-changelog, cspell, update-snippets](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/ILanguageSpecificChecks.cs)). From my local testing+the pipelines I saw it seemed aligned between the two, but could you confirm would there be differences in how these checks are invoked for for mgmt-plane?

maririos (approved):

haolingdong-msft:
@l0lawrence, I checked the [code](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs), and don't see mgmt specific logics. I also checked with @XiaofeiCao and we don't think Java mgmt wrote specific logics from mgmt on those checkings. (/cc @weidongxu-microsoft as well) We can definitely improve later if specific logics for mgmt needed.

For other languages, @tadelesh,  @msyyc, @MaryGao, @haiyuazhang, could you please help to double confirm on @l0lawrence's question? https://github.com/Azure/azure-sdk-tools/pull/12615#issuecomment-3444415796

haolingdong-msft (approved):

tadelesh:
> @l0lawrence, I checked the [code](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs), and don't see mgmt specific logics. I also checked with @XiaofeiCao and we don't think Java mgmt wrote specific logics from mgmt on those checkings. (/cc @weidongxu-microsoft as well) We can definitely improve later if specific logics for mgmt needed.
> 
> For other languages, @tadelesh, @msyyc, @MaryGao, @haiyuazhang, could you please help to double confirm on @l0lawrence's question? [#12615 (comment)](https://github.com/Azure/azure-sdk-tools/pull/12615#issuecomment-3444415796)

Checked the current [code](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageSpecificChecks.cs) and should be same for mgmt.

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
raych1: @m-redding 
This script appears to do multiple things: re-generate the code, validate readme, update snippets, and export api, etc. However, we already have the `validate` tool and the `generate` tool covering these parts: re-generate the code, validate readme, update snippets. For `export api`, it is expected to be covered by the `package update-metadata` tool. 

That said, this [specific check](https://github.com/Azure/azure-sdk-tools/blob/197218080c45416fe128c92986349bc997689924/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs#L33) for .net can likely be deprecated once the respective tools are in place.

raych1 (approved):

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
m-redding: Hmm I think at some point the group may need to decide how much overlap we want to avoid between tools. If the goal is:
> Run validation checks locally to **ensure green PR**

to me that means users should be able to **run only this check** and know they'll have a green PR. It shouldn't only work if you run it in the entire flow. 

But to truly achieve that goal there would be a ton of overlap with a lot of different tools (I mean, should we also run the tests?). May be something worth re-visiting as the tooling evolves.

samvaity: I’d like to share my thoughts: for each language, it would be ideal to break down validations so users can easily understand what might have failed, rather than relying on a complex script that runs multiple processes at once. While I see the benefit of `generatedCodeChecks`, considering the overlap, making these checks modular and having an --all option seems practical(what we are trying to do here). That way, for both Java(Compare-CurrentCodeGeneration) and .NET, if any checks are missed above, they’ll be covered, and we won’t necessarily need to use the script, which would be preferable. The overlaps between tools make scopes blurry which is not ideal. We can start with smaller pieces, rather than the accumulated `generatedCodeChecks` is my thought @m-redding what do you think?

cc: @maririos, @l0lawrence @raych1

haiyuazhang:
> > @l0lawrence, I checked the [code](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs), and don't see mgmt specific logics. I also checked with @XiaofeiCao and we don't think Java mgmt wrote specific logics from mgmt on those checkings. (/cc @weidongxu-microsoft as well) We can definitely improve later if specific logics for mgmt needed.
> > For other languages, @tadelesh, @msyyc, @MaryGao, @haiyuazhang, could you please help to double confirm on @l0lawrence's question? [#12615 (comment)](https://github.com/Azure/azure-sdk-tools/pull/12615#issuecomment-3444415796)
> 
> Checked the current [code](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageSpecificChecks.cs) and should be same for mgmt.

looks good for .NET mgmt.

msyyc:
> @l0lawrence, I checked the [code](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageSpecificChecks.cs), and don't see mgmt specific logics. I also checked with @XiaofeiCao and we don't think Java mgmt wrote specific logics from mgmt on those checkings. (/cc @weidongxu-microsoft as well) We can definitely improve later if specific logics for mgmt needed.
> 
> For other languages, @tadelesh, @msyyc, @MaryGao, @haiyuazhang, could you please help to double confirm on @l0lawrence's question? [#12615 (comment)](https://github.com/Azure/azure-sdk-tools/pull/12615#issuecomment-3444415796)

Looks good to python mgmt.

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
m-redding: I addressed this already above. I did not write this script and I'm not going to rewrite any scripts until we have alignment on https://github.com/Azure/azure-sdk-tools/pull/12563#discussion_r2449474956

lmazuel (approved):

benbp (approved):

praveenkuttappan (approved):
LGTM

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
```diff
@@ -379,41 +378,44 @@ Updating package metadata for Health Deidentification SDK...
...
 ```
 
 **Options:**
-- `--service <name>`: Service name to check (required)
-- `--languages <list>`: Languages to check (default: all)
-- `--checks <list>`: Specific checks to run (build,test,lint,breaking-changes)
-- `--parallel`: Run checks in parallel
-- `--fail-fast`: Stop on first failure
+- `--package-path <path>`: Path to the specific SDK package directory (required)
+- `--fix`: Option to fix the errors reported (Milestone 2)
```
ronniegeraghty, line 395: Do we need a languages option if the user is only working on a single language or subset of languages? Or is the tool doing something smart to figure this out without the need of an extra option?

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
```diff
@@ -379,41 +378,44 @@ Updating package metadata for Health Deidentification SDK...
...
-... (similar for other languages)
-
-PR checks complete: 5/5 languages passed all checks
+[ERROR] Some checks failed
+[NEXT STEPS]
+Successful checks: Spelling, Snippets, Format
+Failed checks: Changelog, README, Linting
+Address the issues identified above before proceeding with package release.
+Re-run the package checks after making corrections to verify all issues are resolved.
+Note: The following checks are not implemented for this language: Dependency, AOT Compatibility, Generated Code, Sample Validation
```
ronniegeraghty, line 418: Do we still output next steps for CLI command output? 
Also, non-blocking, but do we want this to be more specific about which checks failed in which languages?

ronniegeraghty (approved):
Not blocking on these. More just questions/thoughts for updates that can be made in the future.

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
l0lawrence: it takes in a single path so it works on A language at a time, in the future we can expand this to take in a list

tools/azsdk-cli/docs/specs/0-scenario-1.spec.md
l0lawrence: this runs on one language at a time

PR #12626, opened 2025-10-23
Add eval into azsdk cli pipeline

praveenkuttappan:


eng/pipelines/templates/stages/archetype-sdk-tool-dotnet.yml
```diff
@@ -214,6 +223,21 @@ extends:
 
                 - template: /eng/pipelines/templates/steps/upload-llm-artifacts.yml
 
+          - ${{ if ne(parameters.EvalProject, '') }}:
```
weshaggard, line 226: @praveenkuttappan as we chatted lets remove this from this shared template and make it a job template we can include inside of the cli and instruction pipelines.

Issue #12614, opened 2025-10-22
[API View] Logging fails without providing feedback


AlitzelMendez:
Rigth now when a log in for api view fail, we are not returning any feedback and it just enters into a infinite loop of please loging

PR #12647, opened 2025-10-24
[API View] Fix: Infinite Redirect Loop for Users with Wrong GitHub Organization

AlitzelMendez:
closes: https://github.com/Azure/azure-sdk-tools/issues/12614

## Problem
Users authenticated with a GitHub account that's not in the required organization (Azure/Microsoft) experience an infinite redirect loop when accessing the SPA. Instead of seeing an error message, they get stuck redirecting between the SPA and login page.

**Root Cause:** Cookie authentication was redirecting API requests to HTML pages, returning HTTP 200 with HTML content. The SPA couldn't handle this and kept retrying.

## Solution
- **Backend:** Modified cookie authentication events to return proper HTTP status codes (`403`/`401`) for API requests instead of HTML redirects
- **Frontend:** Added interceptor to catch `403` errors and redirect to the `/Unauthorized` page with helpful instructions

✅ Users with wrong organization now see the `/Unauthorized` page with clear instructions  
✅ No more infinite redirect loop

### DIFF
--- a/src/dotnet/APIView/ClientSPA/src/app/_services/http-error-interceptor/http-error-interceptor.service.ts
+++ b/src/dotnet/APIView/ClientSPA/src/app/_services/http-error-interceptor/http-error-interceptor.service.ts
@@ -16,7 +16,6 @@ export class HttpErrorInterceptorService implements HttpInterceptor{
             const now = Date.now();            
             if (now - HttpErrorInterceptorService.lastRedirectTime > HttpErrorInterceptorService.REDIRECT_COOLDOWN_MS) {
               HttpErrorInterceptorService.lastRedirectTime = now;
-              
               // Remove 'spa.' only if it appears at the start of the hostname
               const url = new URL(window.location.origin);
               if (url.hostname.startsWith('spa.')) {


src/dotnet/APIView/ClientSPA/src/app/app.component.ts
```diff
@@ -55,7 +55,12 @@ export class AppComponent  implements OnInit{
           body.classList.add(theme);
         }
         this.loadHighlightTheme(this.getHighlightTheme(theme));
-      });
+      },
+      error: () => {
+        // If user profile fails (e.g., 403 Forbidden), use default theme
```
maririos, line 60: and once they log in they will go back to the spa side of APIView?

AlitzelMendez: no, when this happens it sends you to the unauthorized page where explains what to do to get to the organization public, there is a button at the button to retry loging, if that works it will send you to the main page

maririos: is the main page common for both cases? or is there a `main spa` (idk how to say it)

AlitzelMendez: oh no, there is not main spa, spa is only when you interact with a review/revision but for the "homepage" you always access to spa! I can show you tomorrow

Issue #11857, opened 2025-08-27
[API View] Version overwriting issue on release branches


AlitzelMendez:
## Summary
API Views generated from release branches are being overwritten by scheduled pipeline runs from `main`, causing version mismatches and breaking existing reviews. This impacts teams performing GA releases from release branches, where a stable API View is needed during the review cycle.

## Example reported: 
- Python release pipeline run - [Pipelines - Run 20250810.1 logs](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fdev.azure.com%2Fazure-sdk%2Finternal%2F_build%2Fresults%3FbuildId%3D5202325%26view%3Dlogs%26j%3Db70e5e73-bbb6-5567-0939-8415943fadb9%26t%3D56a7d543-abfc-52f1-3602-2fe2c068a1d3&data=05%7C02%7CAlitzel.Mendez%40microsoft.com%7Cbfad0d273f3e43764bac08dde4ec6250%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638918426091191846%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=pPXLLRifbWbSnzWvhz7Q16sxm9oZQEQIwMpoJY0eCn0%3D&reserved=0)

[...]

PR #12658, opened 2025-10-25
[API View] Compare package version also for pre-release

AlitzelMendez:
closes: https://github.com/Azure/azure-sdk-tools/issues/11857

### Problem
Pre-release versions were skipping version comparison, causing version mismatches:
- UI displayed: `12.28.0b1`
- Actual code file: `12.27.0` ❌

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
+++ b/src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
@@ -276,7 +276,7 @@ namespace APIViewWeb.Controllers
                         // But any manual pipeline run at release time should compare against all approved revisions to ensure hotfix release doesn't have API change
                         // If review surface doesn't match with any approved revisions then we will create new revision if it doesn't match pending latest revision
 
-                        bool considerPackageVersion = !String.IsNullOrWhiteSpace(codeFile.PackageVersion) && !(new AzureEngSemanticVersion(codeFile.PackageVersion, codeFile.Language).IsPrerelease);
+                        bool considerPackageVersion = !String.IsNullOrWhiteSpace(codeFile.PackageVersion);
 
                         if (compareAllRevisions)
                         {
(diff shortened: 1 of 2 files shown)


src/dotnet/APIView/APIViewWeb/Controllers/AutoReviewController.cs
```diff
@@ -276,7 +276,7 @@ public async Task<ActionResult> CreateApiReview(
                         // But any manual pipeline run at release time should compare against all approved revisions to ensure hotfix release doesn't have API change
                         // If review surface doesn't match with any approved revisions then we will create new revision if it doesn't match pending latest revision
 
-                        bool considerPackageVersion = !String.IsNullOrWhiteSpace(codeFile.PackageVersion) && !(new AzureEngSemanticVersion(codeFile.PackageVersion, codeFile.Language).IsPrerelease);
+                        bool considerPackageVersion = !String.IsNullOrWhiteSpace(codeFile.PackageVersion);
```
maririos, line 279: I am not familiar with this code :( but looking at the history, it seems it was intentional that Dozie added the part you are removing.
See: https://github.com/Azure/azure-sdk-tools/pull/10788

Is `packageVersion=null` no longer a thing?

AlitzelMendez: we are preserving the `packageVersion=null` but removing the `prerelease` that has been in there even longer!! but I think @praveenkuttappan  has some context if this is necessary or if we can do this change

praveenkuttappan: Dozie made this change to create a revision for each package version even if API surface is same but it was done so only for GA version. It will start creating a new revision even for preview release with this new change to avoid confusion of `where is the revision for my package version`

maririos: Thanks for the clarifications

PR #12685, opened 2025-10-28
Use bench command rather than perf command to build perf tests

LarryOsterman:


tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
```diff
@@ -69,11 +66,10 @@ public UtilEventArgs(string methodName, string[] methodParams)
             // just make sure we have the target directory cleaned up of previous results in case of a test issue in a previous test / run
             _targetResultsDirectory = await CleanupAsync(project);
             Directory.CreateDirectory(_targetResultsDirectory);
-            _debug = debug;
 
             string flavor = debug ? "debug" : "release";
-            _testCommand = $"--{flavor} --package {primaryPackage} --test perf ";
-            var result = await Util.RunAsync(_cargoName, $"build {_testCommand} --message-format=json", WorkingDirectory, log: false);
+            string testCommand = $"--{flavor} --package {primaryPackage} --bench perf ";
+            var result = await Util.RunAsync("cargo", $"build {testCommand} --message-format=json", WorkingDirectory, log: false);
```
heaths, line 72: Nit: I'd prefer the command line in a single `format` just so there's not this extra level of indirection we have to refer to. More typical that way. If you were going to reuse it like you did before, that'd make more sense but this is a local.

heaths (approved):

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
```diff
@@ -69,11 +66,10 @@ public UtilEventArgs(string methodName, string[] methodParams)
             // just make sure we have the target directory cleaned up of previous results in case of a test issue in a previous test / run
             _targetResultsDirectory = await CleanupAsync(project);
             Directory.CreateDirectory(_targetResultsDirectory);
-            _debug = debug;
 
             string flavor = debug ? "debug" : "release";
-            _testCommand = $"--{flavor} --package {primaryPackage} --test perf ";
-            var result = await Util.RunAsync(_cargoName, $"build {_testCommand} --message-format=json", WorkingDirectory, log: false);
+            string testCommand = $"--{flavor} --package {primaryPackage} --bench perf ";
```
RickWinter, line 71: Nit: perfCommand

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
RickWinter, line 120: @mikeharder These should be perfName not testName.

tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/Rust.cs
LarryOsterman: That was intentional, actually - the name of a perf test is always "perf" because it's a perf test.

PR #12696, opened 2025-10-29
Add spec for repo script calling contract

benbp:
This PR breaks out the script-calling discussion raised in the package updater spec (https://github.com/Azure/azure-sdk-tools/pull/12563) into its own spec so we can focus debate.

Proposed implementation: https://github.com/Azure/azure-sdk-tools/pull/12714

### DIFF
--- a/tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
+++ b/tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
@@ -294,7 +294,6 @@ Tests to call sample scripts using different parameter types. Invoked from mcp, 
 - Test multi-alias references
 - Test sanitization of parameters
 - Test powershell execution
-- Test
 
 ### Integration Tests
 
@@ -306,7 +305,7 @@ Developers go through generation and release process using `azsdk` tooling with 
 
 ### Cross-Language Validation
 
-- Existing CI in langauge repos should catch issues as it is migrated to use `azsdk` as a pipeline entrypoint for tasks.
+- Existing CI in language repos should catch issues as it is migrated to use `azsdk` as a pipeline entrypoint for tasks.
 - Potential CI to validate that scripts referenced in the contract file support `Alias`, if applicable
 
 ---
--- a/tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
+++ b/tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
@@ -104,7 +104,7 @@ The design proposal below removes some aspects of the contract initially propose
 
 ### Overview
 
-When a command/tool is invoked in `azsdk` that is known to be repository-specific (i.e. the command requires a package path or a call to get the repository root), it should first check a centralized config in that repository. If the config contains an override entry for the command, call the override specified. If not, fall back to the implementation in the CLI/MCP server (or fail if not implemented).
+When a command/tool is invoked in `azsdk` that is known to be repository-specific (i.e. the command requires a package path or a call to get the repository root), it should first check a centralized config in that repository. If the config contains an override entry for the command, call the override specified. If not, fall back to the implementation in the CLI/MCP server (or fail/no-op if not implemented).
 
 ### Detailed Design
 
--- a/tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
+++ b/tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
@@ -19,15 +19,7 @@
 
 _Define any terms that might be ambiguous or interpreted differently across teams. Establish shared understanding before diving into the design._
 
-**Example:**
-
-- **Generate SDK**: In this context, "generate SDK" means...
-- **Environment Setup**: Refers to...
-- **Tool**: A discrete unit of functionality that...
-
-**Your definitions:**
-
-- **[Splatting]**: [A powershell feature that allows a key/value object to be transformed into script parameters]
+- **[Splatting]**: A powershell feature that allows a key/value object to be transformed into script parameters
 
 ---
 
--- a/tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
+++ b/tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
@@ -1,0 +1,342 @@
+# Spec: 7 Repository Tooling Contract - (no tool)
+
+## Table of Contents
+
+- [Definitions](#definitions)
+- [Background / Problem Statement](#background--problem-statement)
+- [Goals and Exceptions/Limitations](#goals-and-exceptionslimitations)
+- [Design Proposal](#design-proposal)
+- [Alternatives Considered](#alternatives-considered) _(optional)_
+- [Open Questions](#open-questions)
+- [Success Criteria](#success-criteria)
+- [Implementation Plan](#implementation-plan)
+- [Testing Strategy](#testing-strategy)
+- [Documentation Updates](#documentation-updates)
+
+---
+
+## Definitions
+
+_Define any terms that might be ambiguous or interpreted differently across teams. Establish shared understanding before diving into the design._
+
+**Example:**
+
+- **Generate SDK**: In this context, "generate SDK" means...
+- **Environment Setup**: Refers to...
+- **Tool**: A discrete unit of functionality that...
+
+**Your definitions:**
+
+- **[Splatting]**: [A powershell feature that allows a key/value object to be transformed into script parameters]
+
+---
+
+## Background / Problem Statement
+
+There are no firm guidelines for calling scripts and/or binaries that are specific to a language repository from the `azsdk` tool.
+Without a standardized approach, it is difficult to reason about potential breaking changes when `azsdk`
+is tightly coupled to the call pattern of scripts in other repositories.
+
+This spec builds upon the ideas outlined in [6-package-updater-spec.md](https://github.com/Azure/azure-sdk-tools/blob/cc5bae3dd44b02c01ebb8d7546fdbeb0838b665d/tools/azsdk-cli/docs/specs/6-package-updater.spec.md#proposed-approaches-for-the-intergration-between-language-specific-logic-and-tools) and the [preceding discussion](https://github.com/Azure/azure-sdk-tools/pull/12563#discussion_r2449474956). It seemed easier to focus that discussion into this separate spec.
+
+### Current State
+
+- Scripts can be referenced directly in `azsdk` by a hardcoded file path in the process helper [example](https://github.com/Azure/azure-sdk-tools/blob/826bd97dc1580e65b82ac7e4c2c4cd35a1ff8c3a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs#L54)
+- There are no CI tests in the target repositories to prevent breaking changes being made to referenced scripts.
+- There is no formal versioning/dependency between `azsdk` and the repository state.
+
+### Why This Matters
+
+There is years worth of hardened automation that has been driving the engineering system in the language repositories.
+As the `azsdk` tool is being developed it offers opportunities to streamline some of this automation and also make it accessible
+to AI agents.
+It seems likely that existing repository scripts will always exist in some capacity (see below), so a formalized approach to referencing
+these scripts will have the following benefits:
+
+- Improved velocity - tactical script changes can be made with confidence while not requiring an `azsdk` release.
+  - As `azsdk` matures and gains usage/impact, the validations/requirements to release a new version will increase.
+  - Small, language-specific changes can be made without needing to understand `azsdk` development.
+  - Breakglass changes can be made to resolve urgent issues (unblocking releases, security issues, etc.)
+- Easier migration of scripts to `azsdk`.
+  - It will take time to centralize automation into `azsdk`. There needs to be a bridge between `azsdk` and existing scripts that can support a gradual migration while maintaining reliability.
+- New automation experiments can be made in language repositories, even if the desired end state is a generic interface in `azsdk`.
+- Language specific edge cases can exist without cluttering the cli/mcp tool list.
+- Depending on calling a script (i.e. Export-Apis.ps1) or command line (msbuild /p:ExportApis) that have a set of parameters unique to our repo will be fragile and make versioning `azsdk` with the source changes in the repo very difficult. It will also make this tool very difficult to use in other repo contexts (i.e. the openai or mcp repos).
+
+---
+
+## Goals and Exceptions/Limitations
+
+### Goals
+
+- [ ] Simplify SDK automation
+  - Make automation entrypoints easier to understand and invoke
+  - Invoke operations in the same way locally and in pipelines. Remove multiple entrypoints for the same logic.
+  - Reduce layers of indirection and complexity
+  - Improve testability of changes
+- [ ] Reduce tight coupling between `azsdk` and language repository automation
+  - `azsdk` cli commands and mcp tools should not be bound to a repository structure that isn't designed to version with it
+- [ ] Enable development in both areas
+  - Keep toil low for small, tactical changes to scripts in a repository
+  - Support an efficient on-ramp for script logic to be migrated to `azsdk` without requiring intensive release coordination
+
+### Limitations
+
+Managing parameter mappings can still cause tight coupling or potential breakages.
+The design proposal below outlines the use of `[Alias]` for powershell cmdlets to address this.
+However, this approach means that direct calls to repository tools (e.g. `msbuild`) should be wrapped in repository scripts.
+
+Adding schemas for parameters could improve reliability but at the cost of requiring changes to `azsdk` and/or scripts to be coordinated.
+It would also require additional up front work to capture parameter schemas for existing scripts, along with CI checks to ensure the
+scripts implement the schema.
+
+## Design Proposal
+
+The original idea for this approach is copied from the [package updater spec](https://github.com/Azure/azure-sdk-tools/blob/cc5bae3dd44b02c01ebb8d7546fdbeb0838b665d/tools/azsdk-cli/docs/specs/6-package-updater.spec.md#proposed-approaches-for-the-intergration-between-language-specific-logic-and-tools) by [@raych1](http://github.com/raych1):
+
+> Define a small, well-documented contract/API (parameters, expected inputs/outputs, exit codes, JSON schema). Language repo implements the contract (script or program). CLI does normalization, validation, and invokes the repo-side contract. Optionally implement language-specific logic in CLI for cases where shared helpers or BCL types are beneficial; otherwise call the repo script.
+
+> Pros: Best of both worlds: a small, stable contract (parameters, exit codes, JSON output) that CLI depends on while letting repo implement details. Repo can evolve implementation without breaking consumers so long as contract remains stable. CLI can provide shared normalization, validation, telemetry, and orchestration, while repo scripts handle repo-local tasks (file paths, tool variants). Easier to test and mock (CLI tests against a known contract; repo tests ensure script adheres to contract).
+
+> Cons: Requires upfront design of the contract and discipline to keep it stable. Still requires versioning decisions (how to roll out contract changes across repos). Some duplication of wrapper code may persist (but minimized if contract is well-designed).
+
+The design proposal below removes some aspects of the contract initially proposed, only keeping the script path and lookup tags.
+
+### Overview
+
+When a command/tool is invoked in `azsdk` that is known to be repository-specific (i.e. the command requires a package path or a call to get the repository root), it should first check a centralized config in that repository. If the config contains an override entry for the command, call the override specified. If not, fall back to the implementation in the CLI/MCP server (or fail if not implemented).
+
+### Detailed Design
+
+In `eng/azsdk-automation-contract.json` in a language repository, a set of mappings can be defined that
+correspond to CLI commands and/or MCP tools.
+
+```
+[
+  {
+    "tags": ["FooBarCheck", "azsdk_foo_bar", "foobar"],
+    "command": "eng/scripts/FooBar.ps1"
+  }
+]
+```
+
+Any script parameters will be [splatted](https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_splatting) in the script call, and the mapping enabled via the `[Alias]` attribute in the powershell cmdlet. No contract is defined for parameters.
+
+By supporting `[Alias(...)]` script authors can easily coordinate changes to parameter names in `azsdk` without having to
+synchronize releases of the binary with script changes in the local repository.
+Aliases used by older `azsdk` versions can be kept around for backwards compatibility.
+
+To onboard an existing script to be used by the `azsdk` tool, all that needs to be done is add aliases for parameters.
+Alternatively, normalize the keys used by an `azsdk` invocation of the script to match its existing parameter names.
+
+```pwsh
+# eng/scripts/FooBar.ps1
+param(
+  [Alias('fooBarOption', 'foo-bar')]
+  [string] $FooBar
+)
+
+echo $FooBar
+```
+
+On the `azsdk` side, a service will be written to enable the following call patterns:
+
+- Within a helper method: a key/value parameters object can be created manually and passed to `Invoke`
+- In CLI mode: a key/value parameters object can be created from arguments and options, based on the parse result from `System.CommandLine`
+  - The key for arguments will be the key given in the CLI help text
+  - The key for options will be the non-aliased option flag value with leading hyphens trimmed, e.g. `--repo-path` becomes `repo-path`
+- In MCP mode: a key/value parameters object will be created from the variable names and values that are used to generate the tool's JSON schema
+
+```csharp
+public async Task<CLICheckResponse> FooBarCheck(string fooBarOption, CancellationToken ct = default)
+{
+    if (RepositoryService.HasImplementation("FooBarCheck"))
+    {
+        return RepositoryService.Invoke("FooBarCheck", new()
+        {
+            [nameof(fooBarOption)] = fooBarOption
+        });
+    }
+
+    // do fallback work here or throw not implemented
+}
+```
+
+```csharp
+// RepositoryService simple implementation
+public async Task<CLICheckResponse> Invoke(string commandName, OrderedDictionary<string, SomeParameterUnionType> params)
+{
+    var scriptPath = GetCommand(commandName);
+    var paramJson = JsonSerializer.Serialize(params, new JsonSerializerOptions { WriteIndented = false });
+    var command = $"$params = ('{paramJson}' | ConvertFrom-Json -AsHashtable); & {scriptPath} @params";
+    var options = new PowershellOptions(args: command);
+    var result = await _powershellHelper.Run(options, ct);
+    return new CLICheckResponse(result.ExitCode, result.Output);
+}
+```
+
+### Cross-Language Considerations
+
+_How does this design work across different SDK languages?_
+
+| Language   | Approach | Notes |
+|------------|----------|-------|
+| .NET       | [How it works in .NET] | [Any specific considerations] |
+| Java       | [How it works in Java] | [Any specific considerations] |
+| JavaScript | [How it works in JS] | [Any specific considerations] |
+| Python     | [How it works in Python] | [Any specific considerations] |
+| Go         | [How it works in Go] | [Any specific considerations] |
+
+### User Experience
+
+[How will developers interact with this? Show examples of commands, outputs, or workflows]
+
+```bash
+# foobar command maps to eng/scripts/FooBar.ps1 in config
+# --foo-bar option maps to `foo-bar` key via helper method
+# The following powershell is executed by the repository service:
+#   $params = '{"foo-bar": "foobar"}' | ConvertFrom-Json -AsHashtable
+#   eng/scripts/FooBar.ps1 @params
+$ azsdk foobar --foo-bar foobar
+foobar
+```
+
+### Architecture Diagram
+
+TBD
+
+---
+
+## Alternatives Considered _(optional)_
+
+_What other approaches did you evaluate? Why was this design chosen?_
+
+_Micro-Alternative 0_:
+
+As an alternative to the embedded param splatting command, the logic could instead be checked into `eng/common/scripts/invoke-script.ps1`
+and then called directly via `_powershellHelper.Run(engCommonInvokeScript, ['-Json', paramJson]);`. This would provide optionality
+to make script invocation changes en-masse across the repos without requiring an `azsdk` release. This idea was shelved because
+it wouldn't support repositories without `eng/common` syncing. If a need arises, both modes could be supported (invoke eng/common if
+it exists, then fall back to embedded).
+
+**Copied from the [package updater spec](https://github.com/Azure/azure-sdk-tools/blob/cc5bae3dd44b02c01ebb8d7546fdbeb0838b665d/tools/azsdk-cli/docs/specs/6-package-updater.spec.md#proposed-approaches-for-the-intergration-between-language-specific-logic-and-tools) by [@raych1](http://github.com/raych1):**
+
+### Alternative 1: CLI class-based integration
+
+**Description**
+
+Implement language-specific behavior inside the CLI (C# classes / services), e.g. DotNetLanguageSpecificChecks.cs style. CLI performs discovery, normalization, and calls language-specific logic directly (possibly invoking lower-level commands like msbuild/tools).
+
+**Pros:**
+
+Centralized logic: all language integrations live inside the CLI — easier to reason about and test in one place. Leverages shared helpers, BCL types. Avoids an extra layer of script indirection (fewer wrappers). Easier to integrate complex cross-language logic (normalization, LLM hooks, common result schema).
+
+**Cons:**
+
+Tight coupling to language repo layout and script parameters → brittle when files/params move. CLI releases must be coordinated with language repo changes (versioning friction). Harder to change language-specific behavior quickly — needs a CLI release or pinned CLI in each repo. Some repo-specific tasks (exact file locations, repo branching differences) are easier in scripts that live next to the source.
+
+### Alternative 2: Repo-hosted script integration (script-first)
+
+Keep the authoritative implementation in each language repo as scripts (PowerShell or other). CLI calls those scripts (or a small wrapper script living in the language repo) as the integration point. Example: Export-API.ps1 or update-metadata.ps1 inside the language repo.
+
+**Pros:**
+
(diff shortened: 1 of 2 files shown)


tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+
+> Cons: Requires upfront design of the contract and discipline to keep it stable. Still requires versioning decisions (how to roll out contract changes across repos). Some duplication of wrapper code may persist (but minimized if contract is well-designed).
+
+The design proposal below removes some aspects of the contract initially proposed, only keeping the script path and lookup tags.
+
+### Overview
+
+When a command/tool is invoked in `azsdk` that is known to be repository-specific (i.e. the command requires a package path or a call to get the repository root), it should first check a centralized config in that repository. If the config contains an override entry for the command, call the override specified. If not, fall back to the implementation in the CLI/MCP server (or fail if not implemented).
```
weshaggard: Some tools might not need to fail if not implemented they could potentially support no-op for some languages.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+- Within a helper method: a key/value parameters object can be created manually and passed to `Invoke`
+- In CLI mode: a key/value parameters object can be created from arguments and options, based on the parse result from `System.CommandLine`
+  - The key for arguments will be the key given in the CLI help text
+  - The key for options will be the non-aliased option flag value with leading hyphens trimmed, e.g. `--repo-path` becomes `repo-path`
+- In MCP mode: a key/value parameters object will be created from the variable names and values that are used to generate the tool's JSON schema
+
+```csharp
+public async Task<CLICheckResponse> FooBarCheck(string fooBarOption, CancellationToken ct = default)
+{
+    if (RepositoryService.HasImplementation("FooBarCheck"))
```
weshaggard, line 146: Would RepositoryService handle both the config and the cli specific implementation registration?

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+On the `azsdk` side, a service will be written to enable the following call patterns:
+
+- Within a helper method: a key/value parameters object can be created manually and passed to `Invoke`
+- In CLI mode: a key/value parameters object can be created from arguments and options, based on the parse result from `System.CommandLine`
+  - The key for arguments will be the key given in the CLI help text
+  - The key for options will be the non-aliased option flag value with leading hyphens trimmed, e.g. `--repo-path` becomes `repo-path`
+- In MCP mode: a key/value parameters object will be created from the variable names and values that are used to generate the tool's JSON schema
+
+```csharp
+public async Task<CLICheckResponse> FooBarCheck(string fooBarOption, CancellationToken ct = default)
```
weshaggard, line 144: I assume you define all parameters here in this function signature?

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
benbp: The no-op behavior would probably belong in the C# tool's language provider class. I'll update the wording to indicate fail/no fail based on implementation.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
benbp: Yeah, I was thinking the `RepositoryService` could handle config lookup and caching. The current examples show using it directly, but for the language provider classes I think it makes sense to do this check/invoke at a higher level rather than in each language implementation (so we only dispatch to a language provider if there's no implementation override).

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
benbp: For the simple case where we're using some helper method, the caller can construct the dictionary directly and use whatever key names they want (ideally just `nameof`).

We can do some logic to parse the key names generically if the invocation is wired up at the CLI command/MCP tool dispatch level.

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+  - As `azsdk` matures and gains usage/impact, the validations/requirements to release a new version will increase.
+  - Small, language-specific changes can be made without needing to understand `azsdk` development.
+  - Breakglass changes can be made to resolve urgent issues (unblocking releases, security issues, etc.)
+- Easier migration of scripts to `azsdk`.
+  - It will take time to centralize automation into `azsdk`. There needs to be a bridge between `azsdk` and existing scripts that can support a gradual migration while maintaining reliability.
+- New automation experiments can be made in language repositories, even if the desired end state is a generic interface in `azsdk`.
+- Language specific edge cases can exist without cluttering the cli/mcp tool list.
+- Depending on calling a script (i.e. Export-Apis.ps1) or command line (msbuild /p:ExportApis) that have a set of parameters unique to our repo will be fragile and make versioning `azsdk` with the source changes in the repo very difficult. It will also make this tool very difficult to use in other repo contexts (i.e. the openai or mcp repos).
```
m-redding: Also export-api is a terrible example because it hasn't had a change in parameters i.e. contract since May 2021

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+## Goals and Exceptions/Limitations
+
+### Goals
+
+- [ ] Simplify SDK automation
+  - Make automation entrypoints easier to understand and invoke
+  - Invoke operations in the same way locally and in pipelines. Remove multiple entrypoints for the same logic.
+  - Reduce layers of indirection and complexity
+  - Improve testability of changes
+- [ ] Reduce tight coupling between `azsdk` and language repository automation
```
m-redding: I don't think this proposal solves this problem. It's just adding another layer of coupling that's actually harder to understand.

I also don't think this proposal considers any of the nuance with the tools. Some tools cannot run outside of our repos, such as the validate tool. We have no idea what CI checks other repos do, how can we write prompts for that even if we did allow someone to swap in a script that runs their CIs?

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+  - `azsdk` cli commands and mcp tools should not be bound to a repository structure that isn't designed to version with it
```
m-redding: This seems like overengineering. In my opinion the process, tools, scripts, and repository structure for developing SDKs doesn't change often enough for this to be significant.

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
m-redding: I don't think this proposal really addresses anything in regard to making the tooling usable in other contexts. I can share a bit about the openai repo and using the tooling there since that's what I'm familiar with.

Some tools would be very useful in that context, particularly client.tsp customizations and generating code. That's why dotnet has decided to build 3p versions of those tools separately from this project.

But others are not, such as updating package metadata or validation checks. I don't think we would provide much value by decoupling that from our repos. For example, the validation check tool would be so useful for partners. They can make sure their pipelines will pass without needing to babysit a PR. How can we provide that value without being tightly coupled to the specific pipelines that run in each language repo?

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+
+---
+
+## Goals and Exceptions/Limitations
+
+### Goals
+
+- [ ] Simplify SDK automation
+  - Make automation entrypoints easier to understand and invoke
+  - Invoke operations in the same way locally and in pipelines. Remove multiple entrypoints for the same logic.
```
m-redding: I think THIS is where we should really heavily focus these efforts.
This is where I see the value:

- Moving in-line scripts like this into standalone scripts: https://github.com/Azure/azure-sdk-for-net/blob/main/eng/common/pipelines/templates/steps/validate-filename.yml#L4-L18
- Moving pipeline and ci.yml logic into their associated scripts (to avoid https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs#L222-L252)
- Untangling scripts so it's easier to run them locally
- Creating friendlier output from PowerShell scripts that is actionable for LLMs
- Not duplicating logic in the CLI that we already have in the language repos - package type detection, reading ci.yml files, looking at build properties, etc

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+  - `azsdk` cli commands and mcp tools should not be bound to a repository structure that isn't designed to version with it
+- [ ] Enable development in both areas
+  - Keep toil low for small, tactical changes to scripts in a repository
+  - Support an efficient on-ramp for script logic to be migrated to `azsdk` without requiring intensive release coordination
+
+### Limitations
+
+Managing parameter mappings can still cause tight coupling or potential breakages.
+The design proposal below outlines the use of `[Alias]` for powershell cmdlets to address this.
+However, this approach means that direct calls to repository tools (e.g. `msbuild`) should be wrapped in repository scripts.
```
maririos: do you have an example for this?

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+  - Improve testability of changes
+- [ ] Reduce tight coupling between `azsdk` and language repository automation
+  - `azsdk` cli commands and mcp tools should not be bound to a repository structure that isn't designed to version with it
+- [ ] Enable development in both areas
+  - Keep toil low for small, tactical changes to scripts in a repository
+  - Support an efficient on-ramp for script logic to be migrated to `azsdk` without requiring intensive release coordination
+
+### Limitations
+
+Managing parameter mappings can still cause tight coupling or potential breakages.
```
m-redding: I also think this proposal is limited because of what I mentioned above. Some of these tools are tightly coupled **conceptually** and no matter how many scripts we write we can't change that.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+
+> Cons: Requires upfront design of the contract and discipline to keep it stable. Still requires versioning decisions (how to roll out contract changes across repos). Some duplication of wrapper code may persist (but minimized if contract is well-designed).
+
+The design proposal below removes some aspects of the contract initially proposed, only keeping the script path and lookup tags.
+
+### Overview
+
+When a command/tool is invoked in `azsdk` that is known to be repository-specific (i.e. the command requires a package path or a call to get the repository root), it should first check a centralized config in that repository. If the config contains an override entry for the command, call the override specified. If not, fall back to the implementation in the CLI/MCP server (or fail/no-op if not implemented).
```
maririos, line 100: similar to what we do with `Language-Settings.ps1` files?

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+synchronize releases of the binary with script changes in the local repository.
+Aliases used by older `azsdk` versions can be kept around for backwards compatibility.
+
+To onboard an existing script to be used by the `azsdk` tool, all that needs to be done is add aliases for parameters.
+Alternatively, normalize the keys used by an `azsdk` invocation of the script to match its existing parameter names.
+
+```pwsh
+# eng/scripts/FooBar.ps1
+param(
+  [Alias('fooBarOption', 'foo-bar')]
```
maririos, line 128: why would it have different aliases?

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
```diff
@@ -0,0 +1,342 @@
...
+    var scriptPath = GetCommand(commandName);
+    var paramJson = JsonSerializer.Serialize(params, new JsonSerializerOptions { WriteIndented = false });
+    var command = $"$params = ('{paramJson}' | ConvertFrom-Json -AsHashtable); & {scriptPath} @params";
+    var options = new PowershellOptions(args: command);
+    var result = await _powershellHelper.Run(options, ct);
+    return new CLICheckResponse(result.ExitCode, result.Output);
+}
+```
+
+### Cross-Language Considerations
```
maririos: @benbp for clarity can you remove the sections that we shouldn't review and that are from the template?

maririos:
@benbp is the idea of the PR just to drive a conversation? or is it to also get it signed off and merged into the repo

weshaggard:
> @benbp is the idea of the PR just to drive a conversation? or is it to also get it signed off and merged into the repo

Both. I would like to see the decisions captured and committed to the repo.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
weshaggard: Only in potential cases where we are making a change that we need to be compatible with different names.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
weshaggard: similar but a little more structured and defined.

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
scbedd: I think he means the difference between a pshell script that is invoked, with `Language-Settings` providing any language-repo-specific shims, versus a direct tool call `shell out to msbuild` where the call _is_ the contract?

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
benbp: Also in cases where we want to pass mcp or cli commands directly to the script override. For example if we implement a middleware that allows us to detect script overrides before touching tool/command code, we can pass the cli flags and/or mcp parameters (which differ in naming) directly to the script with matching aliases.

tools/azsdk-cli/docs/specs/7-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
+# Spec: 7 Repository Tooling Contract - (no tool)
```
raych1:
```suggestion
# Spec: 99 Repository Tooling Contract - (no tool)
```
It's architecture level spec.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
+# Spec: 99 Repository Tooling Contract - (no tool)
```
weshaggard, line 1: https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/docs/per-language.md 

Should we update that doc or incorporate them?

benbp: That doc seems like a good place, I'll add some pieces of the spec over there and also have it cross reference the `process-calling.md` doc I'm updating in https://github.com/Azure/azure-sdk-tools/pull/12714

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Detailed Design
+
+In `eng/azsdk-cli-command-overrides.json` in a language repository, a set of mappings can be defined that
+correspond to CLI commands and/or MCP tools.
+
+```
+[
+  {
+    "tags": ["FooBarCheck", "azsdk_foo_bar", "foobar"],
```
msyyc, line 110: I understand "FooBarCheck" is implementation within `azsdk`, "foobar" is command part of `azsdk foobar ...` then what does "azsdk_foo_bar" mean? Should language owner add all these 3 name into `tag` when trying to override to use script of language repo?

haolingdong-msft, line 110: Could I learn more on: why do we need so many tags for one script? I thought 'tag' and 'command' is like key value pairs, we just need one key to get the command.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+    "command": "eng/scripts/FooBar.ps1"
```
haolingdong-msft, line 111: should the script be `powershell` or it is ok to use other script language?

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Why This Matters
+
+There is years worth of hardened automation that has been driving the engineering system in the language repositories.
+As the `azsdk` tool is being developed it offers opportunities to streamline some of this automation and also make it accessible
+to AI agents.
+It seems likely that existing repository scripts will always exist in some capacity (see below), so a formalized approach to referencing
+these scripts will have the following benefits:
+
+- Improved velocity - tactical script changes can be made with confidence while not requiring an `azsdk` release.
```
haolingdong-msft, line 48: I love this, this is important especially when there is a bug found, we can fix in the script and merge the code in language repo, we don't need to wait for `azsdk` to be released to make the change effective.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+
+> Define a small, well-documented contract/API (parameters, expected inputs/outputs, exit codes, JSON schema). Language repo implements the contract (script or program). CLI does normalization, validation, and invokes the repo-side contract. Optionally implement language-specific logic in CLI for cases where shared helpers or BCL types are beneficial; otherwise call the repo script.
+
+> Pros: Best of both worlds: a small, stable contract (parameters, exit codes, JSON output) that CLI depends on while letting repo implement details. Repo can evolve implementation without breaking consumers so long as contract remains stable. CLI can provide shared normalization, validation, telemetry, and orchestration, while repo scripts handle repo-local tasks (file paths, tool variants). Easier to test and mock (CLI tests against a known contract; repo tests ensure script adheres to contract).
```
haolingdong-msft, line 92: From my understanding, current design is more flexible, there are two approaches to let each language to integrate with azsdk cli. 
One is to write C# code directly into azsdk cli. The other is to write script in each language repo and  invoke the script from C# code. 
Do we have recommendation on when to choose which approach?

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+# eng/scripts/FooBar.ps1
+param(
+  [Alias('fooBarOption', 'foo-bar')]
+  [string] $FooBar
+)
+
+echo $FooBar
+```
+
+On the `azsdk` side, a service will be written to enable the following call patterns:
```
haolingdong-msft, line 135: > a service will be written to enable the following call patterns:

Sorry I might be lack of context on engsys. By the service, do you mean something like 'tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs'? 

e.g. for update change log tool, Java will write a script in sdk repo to implement the logics and add a service called `JavaUpdateChangeLog.cs` to invoke the script with the parameters.

haolingdong-msft (commented):
Thanks @benbp for adding the design doc! It makes the language side work clearer. Just adding some follow up questions on integration and implementation details for each language .

haolingdong-msft:
Hi @benbp @raych1,  Shall we add some guidelines on how each language implements a script?
- How to write script when encouter error? Always log error to console, or to return a json object with status code and error message
- Is there any preference on the script language to use? Or it is completely flexible.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
haolingdong-msft: A follow up question: if we need to change the script's input parameters, we still need to wait for `azsdk` to be release? Is my understanding correct?

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
weidongxu-microsoft: If possible, can we use a somewhat more meaningful name in the examples?

E.g. in `azsdk foobar --foo-bar foobar` it is really confusing as the command name is `foobar`, the option name is `foo-bar`, and the value of the option is again `foobar`

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+
+Adding schemas for parameters could improve reliability but at the cost of requiring changes to `azsdk` and/or scripts to be coordinated.
+It would also require additional up front work to capture parameter schemas for existing scripts, along with CI checks to ensure the
+scripts implement the schema.
+
+## Design Proposal
+
+The original idea for this approach is copied from the [package updater spec](https://github.com/Azure/azure-sdk-tools/blob/cc5bae3dd44b02c01ebb8d7546fdbeb0838b665d/tools/azsdk-cli/docs/specs/6-package-updater.spec.md#proposed-approaches-for-the-intergration-between-language-specific-logic-and-tools) by [@raych1](http://github.com/raych1):
+
+> Define a small, well-documented contract/API (parameters, expected inputs/outputs, exit codes, JSON schema). Language repo implements the contract (script or program). CLI does normalization, validation, and invokes the repo-side contract. Optionally implement language-specific logic in CLI for cases where shared helpers or BCL types are beneficial; otherwise call the repo script.
```
weshaggard, line 90: While this is good for the discussion I think we can remove this from the committed design doc.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+
+The design proposal below removes some aspects of the contract initially proposed, only keeping the script path and lookup tags.
+
+### Overview
+
+When a command/tool is invoked in `azsdk` that is known to be repository-specific (i.e. the command requires a package path or a call to get the repository root), it should first check a centralized config in that repository. If the config contains an override entry for the command, call the override specified. If not, fall back to the implementation in the CLI/MCP server (or fail/no-op if not implemented).
+
+### Detailed Design
+
+In `eng/azsdk-cli-command-overrides.json` in a language repository, a set of mappings can be defined that
```
weshaggard, line 104: Do we think overrides is needed in the file name? I would just go with `eng/azsdk-cli-commands.json`

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+```
+[
+  {
+    "tags": ["FooBarCheck", "azsdk_foo_bar", "foobar"],
+    "command": "eng/scripts/FooBar.ps1"
+  }
+]
+```
+
+Any script parameters will be [splatted](https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_splatting) in the script call, and the mapping enabled via the `[Alias]` attribute in the powershell cmdlet. No contract is defined for parameters.
```
weshaggard, line 116: I want to challenge `No contract is defined for parameters`. We have to have some notion of what parameters are going to come through how else will a script know what parameters to add? Alias can definitely help with flexibility and back-compat, but we still need to know what parameters and what names to use. We cannot just change the names at will as it will break existing implementation.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+
+By supporting `[Alias(...)]` script authors can easily coordinate changes to parameter names in `azsdk` without having to
```
weshaggard, line 118: I don't think this is correct. If you change the parameters, you must first support them in the scripts before you can update to the newer version of the azsdk with those changes. 

I think we need to actually put some of these things to the test because I fear we are not building a system that will adapt to the existing scripts. We need to define some set of required input parameters otherwise the calls to the scripts will not be able to work correctly if it doesn't receive any of the parameters that it needs to do its work.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+synchronize releases of the binary with script changes in the local repository.
+Aliases used by older `azsdk` versions can be kept around for backwards compatibility.
+
+To onboard an existing script to be used by the `azsdk` tool, all that needs to be done is add aliases for parameters.
```
weshaggard, line 122: If there is no contract for the parameters how does a script know what parameter/alias to define?

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+
+echo $FooBar
+```
+
+On the `azsdk` side, a service will be written to enable the following call patterns:
+
+- Within a helper method: a key/value parameters object can be created manually and passed to `Invoke`
+- In CLI mode: a key/value parameters object can be created from arguments and options, based on the parse result from `System.CommandLine`
+  - The key for arguments will be the key given in the CLI help text
+  - The key for options will be the non-aliased option flag value with leading hyphens trimmed, e.g. `--repo-path` becomes `repo-path`
```
weshaggard, line 140: This sounds a lot like defining a contract for parameters. Why wouldn't it be the same for the scripts? They should have the same contract.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+
+[How will developers interact with this? Show examples of commands, outputs, or workflows]
+
+```bash
+# foobar command maps to eng/scripts/FooBar.ps1 in config
+# --foo-bar option maps to `foo-bar` key via helper method
+# The following powershell is executed by the repository service:
+#   $params = '{"foo-bar": "foobar"}' | ConvertFrom-Json -AsHashtable
+#   eng/scripts/FooBar.ps1 @params
+$ azsdk foobar --foo-bar foobar
```
weshaggard, line 181: While PS supports splatting we could also implement our own version of splatting that would essentially output `--foo-bar "foobar"` which would enable other scripting language calls like npm or python directly. Those scripts already have to arg parse and can implement aliasing if they wish.

richardpark-msft (commented):
Mostly just docs and type naming.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
m-redding: I think having a contract for parameters makes sense my only problem would be with including them in the specs and enforcing them across all languages.

I don't know exactly what kind of contract you have in mind, but I would hope it would be something that individual languages can define and implement as they implement their own pieces for each tool. At least for me, I've found that we all have way less in common across languages than I would have expected. I think it would really slow down design and therefore the development of each tool to force upfront design and normalization of script inputs.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
raych1: It's added in [this PR](https://github.com/Azure/azure-sdk-tools/pull/12714/files#diff-18ce93b1228809d2436033bc7e9b2dc3cd8df6a1d0e3aad8ff4d044e579149ddR144-R165)

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
raych1: I would say it depends on the change. If the change of the parameter still keeps the same alias, it won't need change in the `azsdk-cli` code. Otherwise, the parameter mapping in the `azsdk-cli` code will need to be updated. e.g. when adding a new parameter with a brand-new alias.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
raych1: Good question. It's answered in a separate PR: https://github.com/Azure/azure-sdk-tools/pull/12714/files#diff-18ce93b1228809d2436033bc7e9b2dc3cd8df6a1d0e3aad8ff4d044e579149ddR13

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
raych1: Same here, this is answered in the PR: https://github.com/Azure/azure-sdk-tools/pull/12714/files#diff-18ce93b1228809d2436033bc7e9b2dc3cd8df6a1d0e3aad8ff4d044e579149ddR168-R172

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
raych1: I discussed with Ben before, and we'd like to support `powershell` only for now to keep things simple at the beginning. Since most existing scripts are written in `powershell` and it's relatively easy to wrap other languages if needed, we can revisit this design later based on the actual usage.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
ronniegeraghty: Building on this, are we certain that the desire will always be to override the default behavior of the command? 
Is it possible in some cases we'd want to prepend or append extra steps to the default behavior?

benbp:
After talking a bit more with @weshaggard and reading through these comments, we are exploring ideas for a slightly more defined schema.

- The primary integration/override point will be top level mcp tools or CLI calls
- A contract/schema can be generated that those calls to script overrides (pwsh, python, etc.)
- The MCP and CLI middleware can detect overrides at the tool/cmd call level and redirect to the script override
- Where there is not a direct mapping between the tool/cli UX and repo scripts (which may run at a lower level along with other scripts), a higher level script in the repo can be created that matches the tool/cli schema and orchestrates lower level script calls
- The schema will match the tool/cli invocation schema and can be used to enforce that repo scripts follow the same format

@weshaggard did I summarize this correctly?

I wonder if there's an idea here around just piggybacking on the existing mcp tool schema for tools to be overridden and copying that over to a repo contract config.

I will be out of office for the next few weeks, so @raych1 and @weshaggard will take ownership of this spec.

l0lawrence:
> * Where there is not a direct mapping between the tool/cli UX and repo scripts (which may run at a lower level along with other scripts), a higher level script in the repo can be created that matches the tool/cli schema and orchestrates lower level script calls

This feels like it is entering into a wrapper scripts, which I thought that we had agreed upon was not a good approach as it can become more brittle and harder to test?

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
weshaggard: I'm going to do some exploring here to see what options make sense but right now I would like the contract to match the command line options we define for the cli. If that cli tool is specific to a language, then maybe the contract can be more specific but if that cli tool is defined to be generic and support all languages then we would ideally have the same contract.

m-redding:
We're going to need to set up another meeting to discuss this then, because that's not what we decided on in the meeting

weshaggard:
@benbp I think that is a fair summary and just to add more context I plan to do some prototyping of how we might consume the cli in our pipelines to replace existing script calls to see if there are issues.

I do not consider this design complete yet and I do think we need more follow-ups, and I think I will be able to help drive that once I do some prototyping to see how things work.

weshaggard (changes requested):
There is more work to be done here.

tools/azsdk-cli/docs/specs/99-repo-tooling-contract.md
```diff
@@ -0,0 +1,322 @@
...
+Managing parameter mappings can still cause tight coupling or potential breakages.
+The design proposal below outlines the use of `[Alias]` for powershell cmdlets to address this.
+However, this approach means that direct calls to repository tools (e.g. `msbuild`) should be made from the CLI, or in
+some special cases wrapped in a powershell script.
+
+Adding schemas for parameters could improve reliability but at the cost of requiring changes to `azsdk` and/or scripts to be coordinated.
+It would also require additional up front work to capture parameter schemas for existing scripts, along with CI checks to ensure the
+scripts implement the schema.
+
+## Design Proposal
```
jsquire, line 86: Just making sure that I'm following the end-to-end scenario being proposed.   The way that I'm reading this:

- There is one file in the language repositories related to directing DevEx platform commands

- All DevDex CLI and MCP tools will use this file to determine how to invoke repository-specific scripts

- The goal is to provide a neutral format for redirection from specific DevEx command to existing scripts.  For example, the `platform-update` command for .NET could be redirected to the existing `Export-API.ps1` script.

- This obviates the need for most language-specific implementations to have an embedded implementation in the DevEx platform or to write a set of wrapper scripts to transform commands to existing script calls.

Do I have that right?

haolingdong-msft:
Hi @weshaggard @benbp  @raych1. Just would like to share my two cents on the integration approach:

At the begining of inner loop, I raised [concern](https://teams.microsoft.com/l/message/19:meeting_YmI2NTNjZjctMjNhOC00MDViLWFlY2QtYzVhNjNkYzhjNGI3@thread.v2/1752806634239?context=%7B%22contextType%22%3A%22chat%22%7D) on the original design shared by @raych1 about defining contract/API and `swagger_to_sdk.json` configuration to let each language use script to integrate. Later I was told this is the agreed design, so I just volunteered to work with Ray on the build sdk [task](https://github.com/Azure/azure-sdk-tools/issues/11402) to prove the concept. 

Summarize my concern at that time:

The approach feels a bit traditional/waterfall rather than agentic/agile, which might slow down development speed a bit. Because it requires upfront confirmation across all languages, defining APIs and contracts, and identifying sharable components at the very begining.

It might be similar concerns as @m-redding's comment below. @m-redding Please corrrect me if I understand wronly. 😄 

> I'm not sure if there are many common logics across language  At least for me, I've found that we all have way less in common across languages than I would have expected. I think it would really slow down design and therefore the development of each tool to force upfront design and normalization of script inputs.

When working on the build sdk task, I also left a comment for clarifications on how to write script: https://github.com/Azure/azure-sdk-tools/issues/11402#issuecomment-3134950291

-----

With that said. From this PR, **I might want to get more information and clarification on:**

- How can we make the development experience more agile and agentic?
- For script responses: should success be logged to stdout and errors to stderr, or should we consider a structured response like JSON?
- Where should the contract (script input parameters) be defined? How does AzureSdk.CLI pass input parameters to the script?
- Where should each language place its script, and how does AzCli know where to find it?

Thanks in advance for your guidance!

PR #12714, opened 2025-10-30
[azsdk-cli] Add service for calling repository script overrides 

benbp:
Implements the spec detailed at https://github.com/Azure/azure-sdk-tools/pull/12696

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
@@ -7,28 +7,23 @@ namespace Azure.Sdk.Tools.Cli.Services;
 /// .NET-specific implementation of language repository service.
 /// Uses tools like dotnet CLI, MSBuild, NuGet, etc. for .NET development workflows.
 /// </summary>
-public class DotNetLanguageSpecificChecks : ILanguageSpecificChecks
+public class DotNetLanguageSpecificChecks(
+    ILogger<DotNetLanguageSpecificChecks> logger,
+    IProcessHelper processHelper,
+    IPowershellHelper powershellHelper,
+    IGitHelper gitHelper,
+    IRepositoryScriptService repositoryScriptService
+) : ILanguageSpecificChecks
 {
-    private readonly IProcessHelper _processHelper;
-    private readonly IGitHelper _gitHelper;
-    private readonly IPowershellHelper _powershellHelper;
-    private readonly ILogger<DotNetLanguageSpecificChecks> _logger;
+    private readonly IProcessHelper _processHelper = processHelper;
+    private readonly IGitHelper _gitHelper = gitHelper;
+    private readonly IPowershellHelper _powershellHelper = powershellHelper;
+    private readonly IRepositoryScriptService _repositoryScriptService = repositoryScriptService;
+    private readonly ILogger<DotNetLanguageSpecificChecks> _logger = logger;
     private const string DotNetCommand = "dotnet";
     private const string RequiredDotNetVersion = "9.0.102"; // TODO - centralize this as part of env setup tool
     private static readonly TimeSpan CodeChecksTimeout = TimeSpan.FromMinutes(6);
     private static readonly TimeSpan AotCompatTimeout = TimeSpan.FromMinutes(5);
-
-    public DotNetLanguageSpecificChecks(
-        IProcessHelper processHelper,
-        IPowershellHelper powershellHelper,
-        IGitHelper gitHelper,
-        ILogger<DotNetLanguageSpecificChecks> logger)
-    {
-        _processHelper = processHelper;
-        _powershellHelper = powershellHelper;
-        _gitHelper = gitHelper;
-        _logger = logger;
-    }
 
     public async Task<CLICheckResponse> CheckGeneratedCode(string packagePath, bool fixCheckErrors = false, CancellationToken ct = default)
     {
@@ -50,28 +45,15 @@ public class DotNetLanguageSpecificChecks : ILanguageSpecificChecks
                 return new CLICheckResponse(1, "", "Failed to determine service directory from the provided package path.");
             }
 
-            var repoRoot = _gitHelper.DiscoverRepoRoot(packagePath);
-            var scriptPath = Path.Combine(repoRoot, "eng", "scripts", "CodeChecks.ps1");
-            if (!File.Exists(scriptPath))
-            {
-                _logger.LogError("Code checks script not found at: {ScriptPath}", scriptPath);
-                return new CLICheckResponse(1, "", $"Code checks script not found at: {scriptPath}");
-            }
+            var (invoked, result) = await _repositoryScriptService.TryInvoke("CodeChecks", packagePath, new()
+                {
+                    { "ServiceDirectory", serviceDirectory },
+                    { "SpellCheckPublicApiSurface", true }
+                }, ct);
 
-            var args = new[] { scriptPath, "-ServiceDirectory", serviceDirectory, "-SpellCheckPublicApiSurface" };
-            var options = new PowershellOptions(scriptPath, args, workingDirectory: repoRoot, timeout: CodeChecksTimeout);
-            var result = await _powershellHelper.Run(options, ct);
-
-            if (result.ExitCode == 0)
-            {
-                _logger.LogInformation("Generated code checks completed successfully");
-                return new CLICheckResponse(result.ExitCode, result.Output);
-            }
-            else
-            {
-                _logger.LogWarning("Generated code checks for package at {PackagePath} failed with exit code {ExitCode}", packagePath, result.ExitCode);
-                return new CLICheckResponse(result.ExitCode, result.Output, "Generated code checks failed");
-            }
+            return invoked ?
+                new CLICheckResponse(result.ExitCode, result.Output) :
+                new CLICheckResponse(1, "", "No implementation found for CodeChecks in repository.");
         }
         catch (Exception ex)
         {
@@ -224,10 +206,10 @@ public class DotNetLanguageSpecificChecks : ILanguageSpecificChecks
         try
         {
             var csprojFiles = Directory.GetFiles(packagePath, "*.csproj", SearchOption.AllDirectories);
-            var mainCsprojFile = csprojFiles.FirstOrDefault(f => 
+            var mainCsprojFile = csprojFiles.FirstOrDefault(f =>
                 Path.GetFileNameWithoutExtension(f).Equals(packageName, StringComparison.OrdinalIgnoreCase));
             var csprojFile = mainCsprojFile ?? csprojFiles.FirstOrDefault();
-            
+
             if (csprojFile == null)
             {
                 _logger.LogDebug("No .csproj file found in package path: {PackagePath}", packagePath);
@@ -235,15 +217,15 @@ public class DotNetLanguageSpecificChecks : ILanguageSpecificChecks
             }
 
             var projectContent = await File.ReadAllTextAsync(csprojFile, ct);
-            
+
             var hasAotOptOut = projectContent.Contains("<AotCompatOptOut>true</AotCompatOptOut>", StringComparison.OrdinalIgnoreCase);
-            
+
             if (hasAotOptOut)
             {
                 _logger.LogInformation("Found AotCompatOptOut=true in project file: {CsprojFile}", csprojFile);
                 return true;
             }
-            
+
             return false;
         }
         catch (Exception ex)
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
@@ -141,7 +141,8 @@ public class RepositoryScriptService(
         }
 
         var paramJson = JsonSerializer.Serialize(args, jsonOptions);
-        var command = $"$params = ('{paramJson}' | ConvertFrom-Json -AsHashtable); & {scriptPath} @params";
+        var escapedScriptPath = scriptPath.Replace("'", "''");
+        var command = $"$params = @'\n{paramJson}\n'@ | ConvertFrom-Json -AsHashtable; & '{escapedScriptPath}' @params";
 
         var options = new PowershellOptions(args: [command], workingDirectory: workingDirectory);
         var result = await powershellHelper.Run(options, ct);
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
@@ -1,0 +1,163 @@
+using System.Collections.Specialized;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.Sdk.Tools.Cli.Helpers;
+
+namespace Azure.Sdk.Tools.Cli.Services;
+
+public interface IRepositoryScriptService
+{
+    Task<string> GetCommand(string commandName, string packagePath, CancellationToken ct);
+    Task<bool> HasImplementation(string commandName, string packagePath, CancellationToken ct);
+    Task<(bool invoked, ProcessResult result)> TryInvoke(
+        string commandName,
+        string packagePath,
+        OrderedDictionary args,
+        bool invokeFromRepoRoot = true,
+        CancellationToken ct = default
+    );
+    Task<(bool invoked, ProcessResult result)> TryInvoke(string commandName, string packagePath, OrderedDictionary args, CancellationToken ct = default);
+}
+
+public class RepoCommandContract
+{
+    [JsonPropertyName("tags")]
+    public List<string> Tags { get; set; } = [];
+
+    [JsonPropertyName("command")]
+    public string Command { get; set; } = string.Empty;
+}
+
+/// <summary>
+/// Provides repository-specific script discovery and execution by loading the automation config,
+/// caching command mappings per package path, and invoking the associated PowerShell scripts.
+/// </summary>
+/// <param name="logger">The logger used to record informational and warning messages.</param>
+/// <param name="gitHelper">Helper that locates the repository root for a given package path.</param>
+/// <param name="powershellHelper">Helper responsible for running PowerShell commands.</param>
+public class RepositoryScriptService(
+    ILogger<RepositoryScriptService> logger,
+    IGitHelper gitHelper,
+    IPowershellHelper powershellHelper
+) : IRepositoryScriptService
+{
+    public string ScriptConfig = Path.Join("eng", "azsdk-cli-command-overrides.json");
+
+    private readonly Dictionary<string, Dictionary<string, string>> commandCache = [];
+
+    private readonly JsonSerializerOptions jsonOptions = new()
+    {
+        WriteIndented = false,
+        ReadCommentHandling = JsonCommentHandling.Skip,
+        AllowTrailingCommas = true
+    };
+
+    private async Task Load(string repoRoot, CancellationToken ct)
+    {
+        if (commandCache.TryGetValue(repoRoot, out var _))
+        {
+            return;
+        }
+
+        var contractPath = Path.Join(repoRoot, ScriptConfig);
+        if (File.Exists(contractPath))
+        {
+            var json = await File.ReadAllTextAsync(contractPath, ct);
+            var contract = JsonSerializer.Deserialize<List<RepoCommandContract>>(json, jsonOptions);
+
+            if (contract == null)
+            {
+                logger.LogWarning("Failed to deserialize contract at {ContractPath}", contractPath);
+                return;
+            }
+
+            foreach (var command in contract)
+            {
+                foreach (var tag in command.Tags)
+                {
+                    if (!commandCache.ContainsKey(repoRoot))
+                    {
+                        commandCache[repoRoot] = new Dictionary<string, string>();
+                    }
+                    commandCache[repoRoot][tag] = Path.Join(command.Command);
+                }
+            }
+        }
+    }
+
+    public async Task<string> GetCommand(string commandName, string repoRoot, CancellationToken ct)
+    {
+        await Load(repoRoot, ct);
+
+        if (string.IsNullOrEmpty(repoRoot) ||
+            !commandCache.TryGetValue(repoRoot, out var commandMap) ||
+            !commandMap.TryGetValue(commandName, out string? value))
+        {
+            return null;
+        }
+
+        if (!File.Exists(Path.Join(repoRoot, value)))
+        {
+            throw new Exception($"Script override '{value}' for command '{commandName}' in '{ScriptConfig}' does not exist in the repository");
+        }
+
+        return value;
+    }
+
+    public async Task<bool> HasImplementation(string commandName, string repoRoot, CancellationToken ct)
+    {
+        await Load(repoRoot, ct);
+
+        if (string.IsNullOrEmpty(repoRoot) ||
+            !commandCache.TryGetValue(repoRoot, out var commandMap) ||
+            !commandMap.TryGetValue(commandName, out string? _))
+        {
+            return false;
+        }
+
+        return true;
+    }
+
+    public async Task<(bool invoked, ProcessResult result)> TryInvoke(
+        string commandName,
+        string packagePath,
+        OrderedDictionary args,
+        bool invokeFromRepoRoot = true,
+        CancellationToken ct = default
+    )
+    {
+        var repoRoot = gitHelper.DiscoverRepoRoot(packagePath);
+        var scriptPath = await GetCommand(commandName, repoRoot, ct);
+        if (scriptPath == null)
+        {
+            return (false, new());
+        }
+
+        string workingDirectory = "";
+        if (invokeFromRepoRoot)
+        {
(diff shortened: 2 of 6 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
```diff
@@ -50,28 +45,15 @@ public async Task<CLICheckResponse> CheckGeneratedCode(string packagePath, bool
...
-            if (!File.Exists(scriptPath))
-            {
-                _logger.LogError("Code checks script not found at: {ScriptPath}", scriptPath);
-                return new CLICheckResponse(1, "", $"Code checks script not found at: {scriptPath}");
-            }
-
-            var args = new[] { scriptPath, "-ServiceDirectory", serviceDirectory, "-SpellCheckPublicApiSurface" };
-            var options = new PowershellOptions(scriptPath, args, workingDirectory: repoRoot, timeout: CodeChecksTimeout);
-            var result = await _powershellHelper.Run(options, ct);
+            var (invoked, result) = await _repositoryScriptService.TryInvoke("CodeChecks", packagePath, new()
```
samvaity, line 48: Should `HasImplementation` be checked here in case there is an existing implementation that needs to be used instead of calling the script? I am not sure how we would use `HasImplementation`.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
```diff
@@ -50,28 +45,15 @@ public async Task<CLICheckResponse> CheckGeneratedCode(string packagePath, bool
...
-            {
-                _logger.LogInformation("Generated code checks completed successfully");
-                return new CLICheckResponse(result.ExitCode, result.Output);
-            }
-            else
-            {
-                _logger.LogWarning("Generated code checks for package at {PackagePath} failed with exit code {ExitCode}", packagePath, result.ExitCode);
-                return new CLICheckResponse(result.ExitCode, result.Output, "Generated code checks failed");
-            }
+            return invoked ?
```
samvaity, line 54: Also, for other cases, if not script config if found is that equivalent to not invoked?

samvaity: Should the fallback implementation then go here?

benbp: Yes the fallback implementation would go here. Though from my read of the original code in this specific check, this is the single required implementation currently.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+        OrderedDictionary args,
+        bool invokeFromRepoRoot = true,
+        CancellationToken ct = default
+    )
+    {
+        var repoRoot = gitHelper.DiscoverRepoRoot(packagePath);
+        var scriptPath = await GetCommand(commandName, repoRoot, ct);
+        if (scriptPath == null)
+        {
+            return (false, new());
```
praveenkuttappan, line 133: shouldn't we add message here to inform the user that script config is missing for the command?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
praveenkuttappan, line 48: This is great instead of a bunch hardcoded dependency.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: The config is optional, so I think it's up to each tool author whether they want to warn or fail in these cases. Otherwise a default warning could be misleading.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
benbp: `HasImplementation` just means whether there's an override script, not a fallback implementation in code. Perhaps I should rename or remove. Originally I envisioned the flow like:

```
if (repoScriptService.HasImplementation)
{
    repoScriptService.Invoke(...);
}
else
{
    fallback();
}
```

But I decided `TryInvoke` was just simpler (it checks the override + script existence under the hood). I kept `HasImplementation` in case we would want to use it anywhere, since `GetCommand` throws if the script doesn't exist on disk, but maybe it's better to just remove it?

benbp:
Also, one issue in this PR is the repo service will not reload repo config changes in a long running mcp server. I'm going to add another commit to this PR or the next to reload the config if the timestamp has changed.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        return (true, result);
+    }
+
+
+    public async Task<(bool invoked, ProcessResult result)> TryInvoke(string commandName, string packagePath, OrderedDictionary args, CancellationToken ct)
+    {
+        return await TryInvoke(commandName, packagePath, args, true, ct);
+    }
```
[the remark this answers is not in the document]
benbp: I have the overrides so that the cancellation token is required, otherwise we have to do `(..., invokeFromRepoRoot = true, CancellationToken ct = default)`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
samvaity: Yes, I think it's safe to remove it! 
I had the same thought when I read `TryInvoke` but just wanted to get it clarified.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        var contractPath = Path.Join(repoRoot, ScriptConfig);
+        if (File.Exists(contractPath))
+        {
+            var json = await File.ReadAllTextAsync(contractPath, ct);
+            var contract = JsonSerializer.Deserialize<List<RepoCommandContract>>(json, jsonOptions);
+
+            if (contract == null)
+            {
+                logger.LogWarning("Failed to deserialize contract at {ContractPath}", contractPath);
```
richardpark-msft, line 70: We should throw on this - let's not let malformed data cause us to silently fail.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+    Task<(bool invoked, ProcessResult result)> TryInvoke(string commandName, string packagePath, OrderedDictionary args, CancellationToken ct = default);
+}
+
+public class RepoCommandContract
+{
+    [JsonPropertyName("tags")]
+    public List<string> Tags { get; set; } = [];
+
+    [JsonPropertyName("command")]
+    public string Command { get; set; } = string.Empty;
```
richardpark-msft, line 28: Some documentation on this - what is 'command'? Like, a full command invocation with args? Just the name of a powershell script? Or is it just a label -something user friendly to use?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+        bool invokeFromRepoRoot = true,
+        CancellationToken ct = default
+    );
+    Task<(bool invoked, ProcessResult result)> TryInvoke(string commandName, string packagePath, OrderedDictionary args, CancellationToken ct = default);
+}
+
+public class RepoCommandContract
+{
+    [JsonPropertyName("tags")]
+    public List<string> Tags { get; set; } = [];
```
richardpark-msft, line 25: What are tags used for? To search? To group in some query?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+namespace Azure.Sdk.Tools.Cli.Services;
+
+public interface IRepositoryScriptService
+{
+    Task<string> GetCommand(string commandName, string packagePath, CancellationToken ct);
+    Task<bool> HasImplementation(string commandName, string packagePath, CancellationToken ct);
+    Task<(bool invoked, ProcessResult result)> TryInvoke(
+        string commandName,
+        string packagePath,
+        OrderedDictionary args,
```
richardpark-msft, line 15: How do I invoke something that's just a positional argument, not a flag+value?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+
+namespace Azure.Sdk.Tools.Cli.Services;
+
+public interface IRepositoryScriptService
+{
+    Task<string> GetCommand(string commandName, string packagePath, CancellationToken ct);
+    Task<bool> HasImplementation(string commandName, string packagePath, CancellationToken ct);
+    Task<(bool invoked, ProcessResult result)> TryInvoke(
+        string commandName,
+        string packagePath,
```
richardpark-msft, line 14: Is 'packagePath' the full package path? Just the 'sdk/<package path>'? (needs docs)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+using Azure.Sdk.Tools.Cli.Helpers;
+
+namespace Azure.Sdk.Tools.Cli.Services;
+
+public interface IRepositoryScriptService
+{
+    Task<string> GetCommand(string commandName, string packagePath, CancellationToken ct);
+    Task<bool> HasImplementation(string commandName, string packagePath, CancellationToken ct);
+    Task<(bool invoked, ProcessResult result)> TryInvoke(
+        string commandName,
```
richardpark-msft, line 13: CommandName is...path to a powershell script? An exe?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+        string packagePath,
+        OrderedDictionary args,
+        bool invokeFromRepoRoot = true,
+        CancellationToken ct = default
+    );
+    Task<(bool invoked, ProcessResult result)> TryInvoke(string commandName, string packagePath, OrderedDictionary args, CancellationToken ct = default);
+}
+
+public class RepoCommandContract
```
richardpark-msft, line 22: This is a _very_ generic name. What is it actually? Command aliases to commands?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+    Task<string> GetCommand(string commandName, string packagePath, CancellationToken ct);
+    Task<bool> HasImplementation(string commandName, string packagePath, CancellationToken ct);
+    Task<(bool invoked, ProcessResult result)> TryInvoke(
+        string commandName,
+        string packagePath,
+        OrderedDictionary args,
+        bool invokeFromRepoRoot = true,
+        CancellationToken ct = default
+    );
+    Task<(bool invoked, ProcessResult result)> TryInvoke(string commandName, string packagePath, OrderedDictionary args, CancellationToken ct = default);
```
richardpark-msft, line 19: What is the 'invoked' value for here? Wouldn't ProcessResult have enough information to know? 

(and if it doesn't, what were you thinking you'd do if it's false/true?)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
richardpark-msft: Can't we just throw instead? Why are we trying to create two return values?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
```diff
@@ -0,0 +1,164 @@
...
+    private readonly JsonSerializerOptions jsonOptions = new()
+    {
+        WriteIndented = false,
+        ReadCommentHandling = JsonCommentHandling.Skip,
+        AllowTrailingCommas = true
+    };
+
+    private async Task Load(string repoRoot, CancellationToken ct)
+    {
+        if (commandCache.TryGetValue(repoRoot, out var _))
```
raych1, line 57: The cache might be invalidated when the local repo switches to a different branch.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: Having no command override isn't an error condition. Fallback code existing or not is the potential error condition, but that's up to the caller. I have two return values so we can more concisely call `TryInvoke` instead of having to have a condition like `if (HasOverride)` everywhere, then if we didn't invoke anything the fallback code path can be executed. We could also consider returning a response or null, but that feels less clear to me than having named return variables.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: I have been going back and forth on this. The primary intent is that changes to the repo can't break `azsdk` commands as easily. In cases where we DO have fallback implementations and someone messes up the contract config, it's not like it will break all users of `azsdk`.

But on the other hand, we may want to fail fast when the config is broken. If we think a script override is getting called but we're actually executing the fallback code, that could cause for some confusing debugging sessions (especially in a pipeline context).

At the very least I was thinking about adding some common CI checks for the config files (it's valid, the scripts it references exist, etc.).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: It's just a script path, I will rename it and add a doc comment.

benbp: I wonder if we want to do any validation to disallow users from entering commands into the config (e.g. disallow spaces).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: Just to have a many to one mapping from `azsdk` code to the script. For a couple reasons:

- Backwards/forwards compatibility, e.g. we change the command reference in the CLI but have older versions still out in the wild. Or we plan on changing the command reference but timing the `azsdk` release with the repo change in another config is tedious.
- Flexibility with automatic invocations from different call sites (mcp, cli, helper code). I am envisioning middleware where we can annotate a command or mcp tool to be repo script overridable. In those cases, we would inspect the command name or mcp tool name, which will differ and try to resolve it against the tags list. This might not be necessary, but it adds almost no complexity to have an array type for tags instead of a string.
  - For example `azsdk_code_checks` (mcp), `checks` (cli) and `CodeChecks` (helper code) could all resolve.
  - Same for the parameters, a flag like `--service-directory` could be automatically parsed to `service-directory` and the override script can add `[Alias('service-directory')]` to its parameter declarations. That way the original `-ServiceDirectory` powershell parameter stays unchanged but accept multiple possible names without us having to create an internal mapping in `azsdk` from CLI options/MCP parameters to powershell parameters.

benbp: See https://github.com/Azure/azure-sdk-for-net/pull/53592 for what the config looks like.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: All positional arguments still have named variables on the powershell side and can be passed via flag+value. I don't think we should try to support positional arguments as it will add a lot of complexity (now we need to know the parameter schema).

benbp:
```pwsh
$ function foo([string]$foo) { echo $foo }
$ foo abc
abc
$ foo -foo abc
abc
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: Good question. I'm not sure. Most cases it will just run from repo root, so it doesn't matter. But it's probably a good idea for us have a `PackagePath` type used across the codebase instead of using string to represent it. FYI @praveenkuttappan

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: It's a script path. I'll update as this is unclear.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: Agree. Perhaps something like `RepositoryScriptOverrideConfig` ?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: The value represents whether we actually found a script override. I haven't been able to think of a better name but agree `invoked` is not clear. Any ideas?

We could also return a null process result instead to indicate that we never tried to run anything and therefore have no process result? But I like the idea of having separate named returns vs. just having to know that we intend a null response to mean "we didn't find an override". So maybe `Task<(bool noOverrideFound, ProcessResult result)>`?

If it's true, then the caller is done and can return the process result. If it's false, then the caller can either execute the c#-based fallback implementation, throw a not implemented error, or return a no-op.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
benbp: My plan was to have the cache be updated when the config timestamp changes but I haven't gotten to it yet.

benbp:
Tests failing on windows due to path separator assertions. Will fix tomorrow. @richardpark-msft yep you called this one.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
haolingdong-msft: Seems the design spec is still using `if (repoScriptService.HasImplementation)`, we may need to update the spec to be consistent.

haolingdong-msft (approved):

tools/azsdk-cli/docs/process-calling.md
```diff
@@ -107,3 +110,72 @@ var result = await powershellHelper.Run(cmdOptions, cancellationToken);
...
+        { "SpellCheckPublicApiSurface", true }
+    }, ct);
+
+// Optionally provide a fallback implementation or no-op instead of failing
+return invoked ?
+    new CLICheckResponse(result.ExitCode, result.Output) :
+    new CLICheckResponse(1, "", "No implementation found for CodeChecks in repository.");
+```
+
+
```
haolingdong-msft, line 156: Just curious, Seems the design spec has the part on decribing how language integrate using script. Do we want  to put them together?

tools/azsdk-cli/docs/process-calling.md
```diff
@@ -9,6 +9,9 @@ This document contains details for calling external processes.
   * [Result type: ProcessResult](#result-type-processresult)
   * [PowershellHelper: scripts and one-liners](#powershellhelper-scripts-and-one-liners)
   * [Best practices for external processes](#best-practices-for-external-processes)
+* [Running Language Repository Scripts](#running-language-repository-scripts)
+  * [When to use a script overrid](#when-to-use-a-script-override)
```
weshaggard, line 13:
```suggestion
  * [When to use a script override](#when-to-use-a-script-override)
```

tools/azsdk-cli/docs/process-calling.md
```diff
@@ -107,3 +110,72 @@ var result = await powershellHelper.Run(cmdOptions, cancellationToken);
...
+
+## Running Language Repository Scripts
+
+Powershell scripts in language repositories should not be called directly, but instead invoked via the `RepositoryScriptService`. This service references a config in the target language repository and looks for any script overrides that match a tag. If there is a match then a tool should invoke that script. Tools can also provide a fallback implementation.
+
+See the [repo-tooling-contract](./specs/99-repo-tooling-contract.md) spec for more details on this requirement.
+
+### When to use a script override
+
+It should be a goal that script automation living in a language repository eventually be implemented in the `azsdk` tool directly.
```
weshaggard, line 122: I don't necessarily think every tool will be implemented directly in the cli, I would like more of them to be as it makes sense, but I don't see a world where all will be implemented directly in the tool. This is also why I'm not sure the term `override` is correct.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotNetLanguageSpecificChecks.cs
weshaggard, line 48: I was not expecting this to be a script invoke wrapper that is called by language specific tools. I was expecting this to be handled in the language service itself. For example with something like UpdateMetadata tool we would look for a script that handles the in the config and would expect a specific set of parameters (i.e. repo root and package path) and the language would implement a script or update an existing script to take those parameters to do the work.

weshaggard (changes requested):
I think we still have more work to figure out here.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
raych1: +1 on having a `PackagePath` type. We might use the absolute path in cases the command is run from a non-root folder.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
raych1: I think we should add validation and report the errors to help with the troubleshooting.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
raych1: I assume this mechanism will be in place before we roll out 'azsdk-cli' to the real users. It just needs to remind developer to restart the tool after updating the config.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/RepositoryScriptService.cs
raych1: +1 on the silent fail approach. A malformed configuration or script shouldn't impact fallback scenarios. For debugging overridden behaviors, we can improve it by well-structured logging.

tools/azsdk-cli/docs/process-calling.md
```diff
@@ -107,3 +110,72 @@ var result = await powershellHelper.Run(cmdOptions, cancellationToken);
...
+    new CLICheckResponse(1, "", "No implementation found for CodeChecks in repository.");
+```
+
+
+The corresponding repository config will look something like:
+
+```json
+[
+  {
+    "tags": [ "CodeChecks", "azsdk_package_check_code" ],
```
msyyc, line 162: It would be nice to add a section to tell language owner how/where to find the right tag `CodeCheck` and `azsdk_package_check_code` when they need to add override script in SDK repo.

PR #12715, opened 2025-10-30
Add static markdown to show list of tools available in current MCP se…

praveenkuttappan:
This PR adds a markdown to show list of tools supported by Azure SDK MCP server. In the future, it will be automatically updated when a new MCP server version is released.


praveenkuttappan:
@lmazuel FYI...

maririos (approved):

PR #12725, opened 2025-10-31
[sample generation] Support extra context and content inlining from relative links

deyaaeldeen:
Adds support for
- `--extra-context <path to file or folder>` that adds more context to the sample generation context
- parsing of relative links in input prompts and inline the content in there. This parsing happens at the top level only and doesn't happen in nested files.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
@@ -43,7 +43,10 @@ namespace Azure.Sdk.Tools.Cli.Helpers
             foreach (Match refMatch in refMatches)
             {
                 var refName = refMatch.Groups[1].Value.Trim();
-                var refPath = refMatch.Groups[2].Value.Trim();
+                // Handle both angle bracket URLs and bare URLs
+                var refPath = !string.IsNullOrEmpty(refMatch.Groups[2].Value) 
+                    ? refMatch.Groups[2].Value.Trim()  // <url> format
+                    : refMatch.Groups[3].Value.Trim(); // bare url format
                 referenceDefinitions[refName] = refPath;
             }
             
@@ -63,7 +66,7 @@ namespace Azure.Sdk.Tools.Cli.Helpers
                 var match = matches[i];
                 string linkText, linkPath;
 
-                // Determine if this is an inline link [text](path) or reference link [text][ref]
+                // Determine link type based on which groups are captured
                 if (!string.IsNullOrEmpty(match.Groups[2].Value))
                 {
                     // Inline link: [text](path)
@@ -88,15 +91,61 @@ namespace Azure.Sdk.Tools.Cli.Helpers
                         continue;
                     }
                 }
+                else if (!string.IsNullOrEmpty(match.Groups[5].Value))
+                {
+                    // Shortcut reference link: [text][]
+                    linkText = match.Groups[5].Value;
+                    var refName = linkText.Trim();
+                    
+                    // Look up the reference definition using the link text as reference name
+                    if (referenceDefinitions.TryGetValue(refName, out var refPath))
+                    {
+                        linkPath = refPath;
+                    }
+                    else
+                    {
+                        // Reference not found, skip this link
+                        logger.LogWarning("Reference definition not found for: {refName}", refName);
+                        continue;
+                    }
+                }
+                else if (!string.IsNullOrEmpty(match.Groups[6].Value))
+                {
+                    // Implicit reference link: [text]
+                    linkText = match.Groups[6].Value;
+                    var refName = linkText.Trim();
+                    
+                    // Look up the reference definition using the link text as reference name
+                    if (referenceDefinitions.TryGetValue(refName, out var refPath))
+                    {
+                        linkPath = refPath;
+                    }
+                    else
+                    {
+                        // Reference not found, skip this link
+                        logger.LogWarning("Reference definition not found for: {refName}", refName);
+                        continue;
+                    }
+                }
+                else if (!string.IsNullOrEmpty(match.Groups[7].Value))
+                {
+                    // Autolink: <url> - replace with just the URL
+                    var autoUrl = match.Groups[7].Value;
+                    result = result.Substring(0, match.Index) + autoUrl + result.Substring(match.Index + match.Length);
+                    continue;
+                }
                 else
                 {
                     // This shouldn't happen with our regex, but just in case
                     continue;
                 }
 
-                // Skip if it looks like a URL (starts with http/https)
+                // Skip if it looks like a URL (starts with http/https/mailto/etc.)
                 if (linkPath.StartsWith("http://", StringComparison.OrdinalIgnoreCase) || 
-                    linkPath.StartsWith("https://", StringComparison.OrdinalIgnoreCase))
+                    linkPath.StartsWith("https://", StringComparison.OrdinalIgnoreCase) ||
+                    linkPath.StartsWith("mailto:", StringComparison.OrdinalIgnoreCase) ||
+                    linkPath.StartsWith("ftp://", StringComparison.OrdinalIgnoreCase) ||
+                    linkPath.StartsWith("file://", StringComparison.OrdinalIgnoreCase))
                 {
                     continue;
                 }
@@ -153,10 +202,10 @@ namespace Azure.Sdk.Tools.Cli.Helpers
             return result;
         }
 
-        [GeneratedRegex(@"\[([^\]]+)\]\(([^)]+)\)|\[([^\]]+)\]\[([^\]]+)\]", RegexOptions.Compiled)]
+        [GeneratedRegex(@"\[([^\[\]]*(?:\[[^\[\]]*\][^\[\]]*)*)\]\(([^\s)]+)(?:\s+[""'][^""']*[""'])?\)|\[([^\[\]]*(?:\[[^\[\]]*\][^\[\]]*)*)\]\[([^\[\]]*)\]|\[([^\[\]]*(?:\[[^\[\]]*\][^\[\]]*)*)\]\[\]|\[([^\[\]]*(?:\[[^\[\]]*\][^\[\]]*)*)\](?!\[)(?!\()|\<([^<>\s]+)\>", RegexOptions.Compiled | RegexOptions.Singleline)]
         private static partial Regex MarkdownLinkRegex();
 
-        [GeneratedRegex(@"^\s*\[([^\]]+)\]:\s*(.+)$", RegexOptions.Compiled | RegexOptions.Multiline)]
+        [GeneratedRegex(@"^\s*\[([^\[\]]+)\]:\s*(?:<([^<>]+)>|([^\s]+))(?:\s+(""[^""]*""|'[^']*'|\([^)]*\)))?\s*$", RegexOptions.Compiled | RegexOptions.Multiline)]
         private static partial Regex ReferenceDefinitionRegex();
     }
 }
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
@@ -1,15 +1,26 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using Microsoft.Extensions.Logging;
+using System.Text.RegularExpressions;
 
 namespace Azure.Sdk.Tools.Cli.Helpers
 {
     /// <summary>
     /// Helper class for processing and manipulating prompt content.
     /// </summary>
-    public static class PromptHelper
+    public static partial class PromptHelper
     {
+        // Maximum file size to read (1 MB)
+        private const long MaxFileSizeBytes = 1024 * 1024;
+        
+        // Compiled regex for better performance when matching markdown links: [text](path)
+        private static readonly Regex markdownLinkRegex =
+            MarkdownLinkRegex();
+
+        // Regex for matching reference definitions: [ref]: path
+        private static readonly Regex referenceDefinitionRegex =
+            ReferenceDefinitionRegex();
+
         /// <summary>
         /// Expands relative file links in markdown-style text by inlining their content.
         /// Supports links in the format [text](relative/path/to/file.ext)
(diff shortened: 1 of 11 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// <returns>Text with file links expanded to include their content</returns>
+        public static async Task<string> ExpandRelativeFileLinksAsync(string text, string basePath, ILogger logger, CancellationToken ct = default)
+        {
+            if (string.IsNullOrWhiteSpace(text))
+            {
+                return text;
+            }
+
+            // Regex to match markdown links: [text](path)
+            var linkPattern = @"\[([^\]]*)\]\(([^)]+)\)";
```
richardpark-msft: Two comments:

### Regex

Your regex also matches: `[](http://)`

I think you want:

```regexp
\[([^\]]+)\]\(([^)]+)\)
```



### Other valid markdown link styles

Markdown also has these styles of links:

```text
Hello, blah link here [hello][world]
[world]: https://microsoft.com
```

These are more common in READMEs, but I don't think you want to skip them.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
```diff
@@ -0,0 +1,113 @@
...
+            var regex = new System.Text.RegularExpressions.Regex(linkPattern);
```
christothes: Should we make this a static field and make the regex compiled also?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
```diff
@@ -0,0 +1,113 @@
...
+            var matches = regex.Matches(text);
+            if (matches.Count == 0)
+            {
+                return text;
+            }
+
+            var result = text;
+            var expandedFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase); // Track to prevent infinite loops
+
+            logger.LogDebug("Found {linkCount} potential file links in prompt", matches.Count);
```
christothes, line 62:
```suggestion
            logger.LogDebug($"Found {matches.Count} potential file links in prompt");
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
```diff
@@ -0,0 +1,113 @@
...
+                        result = result.Substring(0, match.Index) + expandedContent + result.Substring(match.Index + match.Length);
+                    }
+                    else if (!File.Exists(fullPath))
+                    {
+                        logger.LogWarning("Referenced file not found: {linkPath} (resolved to {fullPath})", linkPath, fullPath);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    logger.LogWarning(ex, "Failed to expand file link: {linkPath}", linkPath);
```
richardpark-msft: Let the exception flow out in these helper classes. I'd only catch and eat an exception when you're at the "UI" layer (like the actual command, or MCP tool), since you have to translate that to a result.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
```diff
@@ -0,0 +1,113 @@
...
+
+";
+
+                        // Replace the link with the expanded content
+                        result = result.Substring(0, match.Index) + expandedContent + result.Substring(match.Index + match.Length);
+                    }
+                    else if (!File.Exists(fullPath))
+                    {
+                        logger.LogWarning("Referenced file not found: {linkPath} (resolved to {fullPath})", linkPath, fullPath);
+                    }
```
christothes:
```suggestion
                    if (File.Exists(fullPath))
                    {
                        if (!expandedFiles.Contains(fullPath))
                        {
                            expandedFiles.Add(fullPath);
                            logger.LogDebug("Expanding file link: {linkPath} -> {fullPath}", linkPath, fullPath);

                            var fileContent = await File.ReadAllTextAsync(fullPath, ct);
                        
                            // Create expanded content with clear boundaries
                            var expandedContent = $@"

## Referenced File: {linkText} ({linkPath})

```
{fileContent}
```

";

                            // Replace the link with the expanded content
                            result = result.Substring(0, match.Index) + expandedContent + result.Substring(match.Index + match.Length);
                        }
                    }
                    else
                    {
                        logger.LogWarning("Referenced file not found: {linkPath} (resolved to {fullPath})", linkPath, fullPath);
                    }

```

christothes: The formatting looks broken for this suggestion.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
richardpark-msft: If you're curious about syntax: https://github.com/Azure/azure-sdk-tools/blob/eb2c44c1f68dc3dc5e446ce61eb2e2c2db27972e/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/ReadMeGeneratorTool.cs#L275

richardpark-msft (commented):
Some things worth addressing.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleGeneration/SampleLanguageContext.cs
```diff
@@ -40,14 +40,49 @@ public virtual string GetSampleGenerationInstructions() =>
...
+        }
+
+        var packagePath = pathList[0]; // First path is the package path
+        var additionalPaths = pathList.Skip(1); // Remaining paths are additional context
+
+        var sourceInputProvider = GetSourceInputProvider();
+        var sourceInputs = sourceInputProvider.Create(packagePath).ToList();
+        
+        // Add additional paths as SourceInput entries
+        foreach (var path in additionalPaths)
```
christothes:
```suggestion
        if (!paths.Any())
        {
            throw new ArgumentException("At least one path must be provided", nameof(paths));
        }

        var packagePath = paths.First(); // First path is the package path
        var additionalPaths = paths.Skip(1); // Remaining paths are additional context

        var sourceInputProvider = GetSourceInputProvider();
        var sourceInputs = sourceInputProvider.Create(packagePath).ToList();
        
        // Add additional paths as SourceInput entries
        foreach (var path in additionalPaths)
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleGeneration/SampleLanguageContext.cs
```diff
@@ -40,14 +40,49 @@ public virtual string GetSampleGenerationInstructions() =>
...
+        {
+            if (!string.IsNullOrWhiteSpace(path))
+            {
+                var fullPath = Path.GetFullPath(path.Trim());
+                sourceInputs.Add(new SourceInput(fullPath, 
+                    IncludeExtensions: Array.Empty<string>(), // Include all files for additional context
```
christothes: This comment doesn't seem to make sense in this context.

christothes (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
```diff
@@ -0,0 +1,162 @@
...
+                else
+                {
+                    logger.LogWarning("Referenced file not found: {linkPath} (resolved to {fullPath})", linkPath, fullPath);
+                }
+            }
+
+            return result;
+        }
+
+        [GeneratedRegex(@"\[([^\]]+)\]\(([^)]+)\)|\[([^\]]+)\]\[([^\]]+)\]", RegexOptions.Compiled)]
```
weshaggard: Are all mark down links discoverable via regex?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
```diff
@@ -0,0 +1,162 @@
...
+        /// <summary>
+        /// Expands relative file links in markdown-style text by inlining their content.
+        /// Supports links in the format [text](relative/path/to/file.ext)
+        /// </summary>
+        /// <param name="text">The text containing potential file links</param>
+        /// <param name="basePath">The base directory to resolve relative paths from</param>
+        /// <param name="logger">Logger for debug and warning messages</param>
+        /// <param name="ct">Cancellation token</param>
+        /// <returns>Text with file links expanded to include their content</returns>
+        public static async Task<string> ExpandRelativeFileLinksAsync(string text, string basePath, ILogger logger, CancellationToken ct = default)
```
weshaggard, line 33: What is the purpose of inlining the referenced doc? Is there a particular issue we are trying to address?

deyaaeldeen: The main scenario here is that the user wants generate samples using a prompt in a markdown file that links to a local file containing a code sample. This prompt could slightly vary across target languages but the code sample stays the same. So instead of inlining the same code sample in all prompts, we want to just link to it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
deyaaeldeen: Good point, I expanded this logic to catch all cases in https://github.com/Azure/azure-sdk-tools/pull/12725/commits/23282be457287c38d6b46de36c7de2d9edce2686

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
```diff
@@ -96,18 +192,20 @@ public static async Task<string> ExpandRelativeFileLinksAsync(string text, strin
...
                 {
-                    logger.LogWarning(ex, "Failed to expand file link: {linkPath}", linkPath);
+                    logger.LogWarning("Referenced file not found: {linkPath} (resolved to {fullPath})", linkPath, fullPath);
                 }
             }
 
             return result;
         }
+
+        [GeneratedRegex(@"\[([^\[\]]*(?:\[[^\[\]]*\][^\[\]]*)*)\]\(([^\s)]+)(?:\s+[""'][^""']*[""'])?\)|\[([^\[\]]*(?:\[[^\[\]]*\][^\[\]]*)*)\]\[([^\[\]]*)\]|\[([^\[\]]*(?:\[[^\[\]]*\][^\[\]]*)*)\]\[\]|\[([^\[\]]*(?:\[[^\[\]]*\][^\[\]]*)*)\](?!\[)(?!\()|\<([^<>\s]+)\>", RegexOptions.Compiled | RegexOptions.Singleline)]
```
richardpark-msft, line 205: I'm not sure if the reason the regex is so complex is that it's trying to handle a lot of alternate cases, but I would absolutely be okay with writing _multiple_ simpler regexes to avoid this.

Given the low stakes here, I'm fine with this PR, as-is, but we should file an issue to come back and address this.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
```diff
@@ -40,17 +60,92 @@ public static async Task<string> ExpandRelativeFileLinksAsync(string text, strin
...
+                    // This shouldn't happen with our regex, but just in case
+                    continue;
+                }
+
+                // Skip if it looks like a URL (starts with http/https/mailto/etc.)
                 if (linkPath.StartsWith("http://", StringComparison.OrdinalIgnoreCase) || 
-                    linkPath.StartsWith("https://", StringComparison.OrdinalIgnoreCase))
+                    linkPath.StartsWith("https://", StringComparison.OrdinalIgnoreCase) ||
+                    linkPath.StartsWith("mailto:", StringComparison.OrdinalIgnoreCase) ||
+                    linkPath.StartsWith("ftp://", StringComparison.OrdinalIgnoreCase) ||
```
richardpark-msft, line 147: What decade was this markdown written in? ;)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
```diff
@@ -40,17 +60,92 @@ public static async Task<string> ExpandRelativeFileLinksAsync(string text, strin
...
+                else if (!string.IsNullOrEmpty(match.Groups[7].Value))
+                {
+                    // Autolink: <url> - replace with just the URL
+                    var autoUrl = match.Groups[7].Value;
+                    result = result.Substring(0, match.Index) + autoUrl + result.Substring(match.Index + match.Length);
+                    continue;
+                }
+                else
+                {
+                    // This shouldn't happen with our regex, but just in case
```
richardpark-msft, line 139: If it shouldn't happen, just throw otherwise we'll never detect regex problems.

richardpark-msft (approved):
Approved, but please file follow-up issue for that regex. I can still see it, in my dreams.

weshaggard (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
weshaggard: I also have my worries about using Regex here but agree with moving forward for now.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
deyaaeldeen: Wouldn't that be not so great UX? I don't think the user would care too much about the validity of their markdown.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/PromptHelper.cs
deyaaeldeen: Agreed, it is too messy. I opened https://github.com/Azure/azure-sdk-tools/issues/12748 to track this.

PR #12730, opened 2025-10-31
Add Response properties as custom properties so we can add them to t…

praveenkuttappan:
This change will add package name, language and SDK type to telemetry for SDK release tool, Package check tool and release readiness tool.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
@@ -46,23 +46,33 @@ public class InstrumentedTool(
             activity?.SetTag(TagName.ToolResponse, content);
             activity?.SetStatus(ActivityStatusCode.Ok);
 
-            if (result.Content?.FirstOrDefault() is TextContentBlock contentBlock)
+            try
             {
-                try
+                foreach (var c in result.Content)
                 {
-                    var responseDict = JsonSerializer.Deserialize<Dictionary<string, object>>(contentBlock.Text);
-                    if (responseDict != null)
+                    // Process only TextContentBlock for custom properties
+                    // Other content types are binary, audiio etc.
+                    if (c is TextContentBlock contentBlock)
                     {
-                        foreach (var kvp in responseDict)
+                        var responseDict = JsonSerializer.Deserialize<Dictionary<string, object>>(contentBlock.Text);
+                        if (responseDict != null)
                         {
-                            activity?.SetCustomProperty(kvp.Key, kvp.Value?.ToString() ?? string.Empty);
+                            foreach (var kvp in responseDict)
+                            {
+                                if (kvp.Value != null)
+                                {
+                                    activity?.SetCustomProperty(kvp.Key, JsonSerializer.Serialize(kvp.Value));
+                                }
+                            }                  
                         }
                     }
                 }
-                catch (JsonException jsonEx)
-                {
-                    logger.LogWarning($"Failed to deserialize contentBlock.Text for telemetry properties: {jsonEx.Message}");
-                }
+            }
+            catch (JsonException ex)
+            {
+                activity?.AddException(ex);
+                activity?.SetStatus(ActivityStatusCode.Error, ex.Message);
+                logger.LogError(ex, "Failed to deserialize contentBlock.Text for telemetry properties");
             }
             return result;
         }
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/MCPToolBase.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/MCPToolBase.cs
@@ -2,6 +2,10 @@
 // Licensed under the MIT License.
 using System.CommandLine;
 using System.Diagnostics;
+using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.Sdk.Tools.Cli.Attributes;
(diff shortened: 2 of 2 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
```diff
@@ -46,6 +46,24 @@ public override async ValueTask<CallToolResult> InvokeAsync(RequestContext<CallT
...
+            if (result.Content?.FirstOrDefault() is TextContentBlock contentBlock)
+            {
+                try
+                {
+                    var responseDict = JsonSerializer.Deserialize<Dictionary<string, object>>(contentBlock.Text);
+                    if (responseDict != null)
+                    {
+                        foreach (var kvp in responseDict)
+                        {
+                            activity?.SetCustomProperty(kvp.Key, kvp.Value?.ToString() ?? string.Empty);
```
benbp: I wonder if it's better to process the raw json? If we serialize list types back to object then they'll stringify as `System.Collections.List` I think.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
```diff
@@ -46,6 +46,24 @@ public override async ValueTask<CallToolResult> InvokeAsync(RequestContext<CallT
             activity?.SetTag(TagName.ToolResponse, content);
             activity?.SetStatus(ActivityStatusCode.Ok);
 
+            if (result.Content?.FirstOrDefault() is TextContentBlock contentBlock)
```
benbp: We should also have a corresponding update to the `MCPToolBase` for CLI invocation (though since CLI telemetry isn't currently working, it's a bit moot right now)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
```diff
@@ -46,6 +46,24 @@ public override async ValueTask<CallToolResult> InvokeAsync(RequestContext<CallT
...
+                    {
+                        foreach (var kvp in responseDict)
+                        {
+                            activity?.SetCustomProperty(kvp.Key, kvp.Value?.ToString() ?? string.Empty);
+                        }
+                    }
+                }
+                catch (JsonException jsonEx)
+                {
+                    logger.LogWarning($"Failed to deserialize contentBlock.Text for telemetry properties: {jsonEx.Message}");
```
benbp: Is it worth adding a telemetry custom property to indicate a parse failure? (probably not)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
```diff
@@ -46,6 +46,24 @@ public override async ValueTask<CallToolResult> InvokeAsync(RequestContext<CallT
             activity?.SetTag(TagName.ToolResponse, content);
             activity?.SetStatus(ActivityStatusCode.Ok);
 
+            if (result.Content?.FirstOrDefault() is TextContentBlock contentBlock)
+            {
```
richardpark-msft: The `if` condition here is kind of odd to me. Why doesn't InvokeAsync only return (result.Content[0] as TextContentBlock) value, if we're going to throw away all the other content?

What is the other content in this case? Is it not valuable for telemetry?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
richardpark-msft: @benbp's point is good - also, I just noticed the signature of it takes an `object` as the second parameter. Do we even need to stringify it?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
```diff
@@ -46,6 +46,24 @@ public override async ValueTask<CallToolResult> InvokeAsync(RequestContext<CallT
...
+                try
```
richardpark-msft: The 'if' condition makes a lot of assumptions about the output. I'm assuming this is because we're enforcing it in our tools, somehow?

richardpark-msft (commented):
Small questions.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
praveenkuttappan: MCP tool result contains either TexContent or AudioContent or Binary content or ResourceContent. Our tool response is all Text content. I am filtering out any other types here, so we don't break this if MCP protocol adds a new type.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
praveenkuttappan: Done.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
praveenkuttappan: done

weshaggard (approved):

PR #12733, opened 2025-10-31
Rename CLICheckResponse as PackageCheckResponse

praveenkuttappan:
Rename response class


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Package/PackageCheckResponse.cs
```diff
@@ -45,14 +45,14 @@ protected override string Format()
 }
 
 /// <summary>
-/// CLI check response for cookbook/documentation reference responses.
+/// Package check response for cookbook/documentation reference responses.
 /// </summary>
-public class CookbookCLICheckResponse : CLICheckResponse
+public class CookbookPackageCheckResponse : PackageCheckResponse
```
samvaity, line 50: Unrelated to current change, @LibbaLawrence Is this class still in use?

PR #12746, opened 2025-11-03
Update response for SDK generation and SDK build tools to include package name and language in response

praveenkuttappan:
Changed response type from DefaultCommandResponce to PackageOperationResponse. This will help us add package and language to telemetry for these tools.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageService.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/LanguageService.cs
@@ -10,15 +10,19 @@ namespace Azure.Sdk.Tools.Cli.Services.Languages
         public static readonly ImmutableDictionary<string, SdkLanguage> RepoToLanguageMap = new Dictionary<string, SdkLanguage>()
         {
             { "azure-sdk-for-net", SdkLanguage.DotNet },
-            {  "azure-sdk-for-go", SdkLanguage.Go  },
+            { "azure-sdk-for-go", SdkLanguage.Go },
             { "azure-sdk-for-java", SdkLanguage.Java },
             { "azure-sdk-for-js", SdkLanguage.JavaScript },
             { "azure-sdk-for-python", SdkLanguage.Python },
-            { "azure-sdk-for-rust" , SdkLanguage.Rust}
+            { "azure-sdk-for-rust", SdkLanguage.Rust}
         }.ToImmutableDictionary();
 
         public static SdkLanguage GetLanguageForRepo(string repoName)
         {
+            if (string.IsNullOrEmpty(repoName))
+            {
+                return SdkLanguage.Unknown;
+            }
             if (RepoToLanguageMap.TryGetValue(repoName.ToLower(), out SdkLanguage language))
             {
                 return language;


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Package/PackageResponseBase.cs
```diff
@@ -38,5 +57,23 @@ public void SetPackageType(string packageType)
                 PackageType = type;
             }
         }
+
+        public static SdkLanguage GetLanguageFromRepo(string? repoName)
+        {
+            if (string.IsNullOrEmpty(repoName))
```
samvaity: When can this happen? We don't expect users to pass in this info so cannot be any string ideally?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Package/PackageResponseBase.cs
```diff
@@ -38,5 +57,23 @@ public void SetPackageType(string packageType)
...
+            {
+                return SdkLanguage.Unknown;
+            }
+            repoName = repoName.ToLower();
+            return repoName switch
+            {
+                "azure-sdk-for-net" => SdkLanguage.DotNet,
```
samvaity: Would it make sense to use a constants map for this, since the repo names are fixed?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Package/PackageResponseBase.cs
praveenkuttappan: If path is invalid or caller of this public method passes null.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Package/PackageResponseBase.cs
raych1: +1. I see this map can be used in other places and can we move it to common constant class?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/SdkLanguage.cs
```diff
@@ -4,6 +4,8 @@ namespace Azure.Sdk.Tools.Cli.Models;
 
 public enum SdkLanguage
 {
```
raych1, line 6: Can you also add `rust` since rust team is catching up onboarding to this project? CC: @ronniegeraghty

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Package/PackageOperationResponse.cs
```diff
@@ -0,0 +1,76 @@
...
+            }
+            if (Duration > 0)
+            {
+                output.AppendLine($"Duration: {Duration}ms");
+            }
+            if (!string.IsNullOrEmpty(PackageName))
+            {
+                output.AppendLine($"Package: {PackageName}");
+            }
+            if (Language != SdkLanguage.Unknown)
```
raych1: I think we should always include the value in the output even if it's invalid to help user/developer understand the issue. Same for the following `SdkType`.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkBuildTool.cs
```diff
@@ -89,35 +93,59 @@ public async Task<DefaultCommandResponse> BuildSdkAsync(
...
+                return CreateFailureResponse($"An error occurred: {ex.Message}", null);
+            }
+        }
+
+        private async Task<PackageInfo> GetPackageInfo(string packagePath, CancellationToken ct)
+        {
+            PackageInfo? packageInfo = null;
+            try
+            {
+                var packageInfoHelper = await packageInfoResolver.Resolve(packagePath, ct);
```
raych1, line 114: Can we log an error when language specific `packageinfohelper` fail to be resolved?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkBuildTool.cs
```diff
@@ -89,35 +93,59 @@ public async Task<DefaultCommandResponse> BuildSdkAsync(
...
+            catch (Exception ex)
+            {
+                logger.LogError(ex, "Error occurred while parsing package path");
             }
+            return packageInfo;
         }
 
         // Helper method to create failure responses along with setting the failure state
-        private DefaultCommandResponse CreateFailureResponse(string message)
+        private PackageOperationResponse CreateFailureResponse(string message, PackageInfo? packageInfo)
```
raych1:
```suggestion
        private PackageOperationResponse CreateFailureResponse(string message, PackageInfo? packageInfo = null)
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkBuildTool.cs
```diff
@@ -89,35 +93,59 @@ public async Task<DefaultCommandResponse> BuildSdkAsync(
...
+            {
+                var packageInfoHelper = await packageInfoResolver.Resolve(packagePath, ct);
+                if (packageInfoHelper != null)
+                {
+                    packageInfo = await packageInfoHelper.ResolvePackageInfo(packagePath, ct);
+                }
+            }
+            catch (Exception ex)
+            {
+                logger.LogError(ex, "Error occurred while parsing package path");
```
raych1:
```suggestion
                logger.LogError(ex, $"Error occurred while parsing package path: {packagePath}");
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkBuildTool.cs
raych1: So that the caller doesn't need to pass the 'null' value.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkGenerationTool.cs
```diff
@@ -250,21 +254,23 @@ private bool IsValidSha(string sha)
...
+                SdkRepoName = sdkRepoName
             };
         }
 
         // Helper method to create success responses (no SetFailure needed)
-        private DefaultCommandResponse CreateSuccessResponse(string message)
+        private PackageOperationResponse CreateSuccessResponse(string message, string sdkRepoName)
         {
-            return new DefaultCommandResponse
+            return new PackageOperationResponse
```
raych1, line 269: Why don't we correlate the generation to package level?

praveenkuttappan: What do you mean by package level? I will rename PackageResponse to PackageWorkItem response since that is about devops work item info.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/Package/PackageResponseBase.cs
praveenkuttappan: Removing this method to move it to a common class.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkGenerationTool.cs
praveenkuttappan: I don't see any package name in tsp result. Does SDK generation returns the generated package path or package name? It is currently setting only the repo name so this tool execution will have language tag at least.

raych1: The result doesn't include the package name currently, but we should correlate the telemetry to `package` scope because SDK generation works on the `package` scope.

praveenkuttappan: We can update the package name in response whenever it is available in response and that's all required to correlate this to package. There is nothing else required to add package telemetry for tools. Can you make a change in SDK generation tool to extract the package name and add it in the response. I would prefer to get this merged and get whatever data is available to telemetry.

scbedd (approved):

PR #12753, opened 2025-11-04
Update response type for release plan and TypeSpec operations

praveenkuttappan:
Update response type for release plan, SDK release workflow, and TypeSpec operations so that these tools include properties like TypeSpec project path, Package type, language in response and this will also get added to telemetry.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/TypeSpec/TypeSpecBaseResponse.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/TypeSpec/TypeSpecBaseResponse.cs
@@ -11,7 +11,7 @@ namespace Azure.Sdk.Tools.Cli.Models.Responses.TypeSpec
         [Telemetry]
         [JsonPropertyName("typespec_project")]
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
-        public string TypeSpecProject { get; set; } = string.Empty;
+        public required string TypeSpecProject { get; set; } = string.Empty;
         [JsonPropertyName("package_type")]
         [Telemetry]
         public SdkType PackageType { get; set; }
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TspClientHelper.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TspClientHelper.cs
@@ -44,7 +44,8 @@ public class TspClientHelper : ITspClientHelper
             {
                 ResponseError = isCli
                     ? "Failed to convert swagger to TypeSpec project, see details in the above logs."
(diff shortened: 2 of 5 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/TypeSpec/TypeSpecBaseResponse.cs
```diff
@@ -3,13 +3,16 @@
...
 {
-    public class TypeSpecBaseResponse
+    /// <summary>
+    /// Serves as a base class for TypeSpec-related responses, providing common telemetry properties.
+    /// </summary>
+    public abstract class TypeSpecBaseResponse : CommandResponse
     {
         [Telemetry]
         [JsonPropertyName("typespec_project")]
         public string? TypeSpecProject { get; set; }
```
samvaity: Is this supposed to be TypeSpecProject path?

samvaity: Why is this optional?

scbedd: Good catch my eyes just glazed over this detail.

I _think_ that it's nullable to handle when we are unable to parse a TypeSpecProject path successfully? This same `string?` TypeSpecProject is present on [ReleasePlanBaseResponse.cs](https://github.com/Azure/azure-sdk-tools/pull/12753/files#diff-2815dcf4d474f1e0d7a30d5e2af6207ed633f0ee919abde6d480fa44e5274645). 

Let @praveenkuttappan speak to it but I'm thinking this is a simple

> it's always nullable everywhere

because of situations on the ReleasePlanResponse MUST sometimes fail because of badly configured typespec or the like.

samvaity: Badly configured typespec sounds like an error response in itself but making this optional just opens those fallback gates which I think can be avoided. Not too attached, will let @praveenkuttappan decide :)

praveenkuttappan: As Scott mentioned, some tools does not know the typespec project path so it's optional. If TypeSpec project path is known to the tool, then it will be included in the response. @samvaity It is TypeSpec project path relative to the spec repo root so we know which TSP project is using which tool.

praveenkuttappan: I have made TypeSpecProject as a required field so it will throw compiler error if not set when creating response.
@samvaity @scbedd

PR #12777, opened 2025-11-05
Clean-up language registration and detection

weshaggard:
- Move to using extension method without need for LanguageImplementations class
- Move to using the repo name instead of Language-settings.ps1 to detect language


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/SdkLanguage.cs
```diff
@@ -19,3 +21,36 @@ public enum SdkLanguage
     [JsonPropertyName("Rust")]
     Rust
 }
+
+public static class SdkLanguageExtensions
```
timovv: This made me think this class had extension methods, would `SdkLanguageHelpers` be a more appropriate name?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Extensions/LanguageSpecificExtensions.cs
timovv, line 1: We have extensions for telemetry in this namespace as well. Are we saying that those should move to ServiceRegistrations.cs too?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
```diff
@@ -219,7 +219,7 @@ public string GetRepoName(string pathInRepo)
                 throw new InvalidOperationException($"Unable to parse repository name from remote URL: {uri}");
             }
 
-            string repoName = segments[^1].TrimEnd(".git".ToCharArray());
+            string repoName = segments[segments.Length - 1].Replace(".git", "");
```
timovv: Any reason to change from `^1`?

```suggestion
            string repoName = segments[^1].Replace(".git", "");
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/LanguageSpecificResolver.cs
```diff
@@ -22,8 +20,8 @@ public class LanguageSpecificResolver<T>(
 {
     public async Task<T?> Resolve(string packagePath, CancellationToken ct = default)
     {
-        var language = await DetectLanguageAsync(packagePath, ct);
-        return await Resolve(language ?? default, ct);
+        var language = DetectLanguage(packagePath, ct);
+        return await Resolve(language, ct);
```
timovv, line 24: I guess these don't need to be async anymore.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
timovv: Also just curious about this approach, I know all the different options have trade offs but would we ever expect people to be working from a branch on the -pr repos and if so would that work using this logic?

timovv (approved):
Looks good just a couple of qs but nothing blocking

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
weshaggard: This `^1` syntax is newer and didn't realize that was supported now. I will switch that back. 

You make a good point about the private repos. We should add support for them as well.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/LanguageSpecificResolver.cs
weshaggard: You are probably right I can clean more of the async methods up.

praveenkuttappan (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Extensions/LanguageSpecificExtensions.cs
weshaggard: I'm not necessarily trying to remove the namespace just these types.

praveenkuttappan:
Thanks for this change. I am working on a similar change to replace all feature specific language classes to LanguageService class and we can get rid of language resolver logic. I will have PR available soon.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/LanguageSpecificResolver.cs
weshaggard: Upon further investigation, I don't think I want to change those yet. That is more detailed and we many end up needing this to be aysnc again so keeping them async for now.

timovv (approved):

weshaggard:
Closing this as @praveenkuttappan is picking up these changes in his work.

Issue #12464, opened 2025-10-13
[APIView] UX: Adopt a similar feedback mechanism as Github Copilot


tjprescott:
**Current**

Currently, our feedback UX is, for a bad comment, to give it a thumbs down and then leave a comment mentioning `@azure-sdk` and describing the problem. This kicks of a feedback cycle in Copilot that updates the KB (if warranted) and issues a reply.

The downsides to this is that architects often get it wrong, either failing to mention azure-sdk or just giving a thumbs down and deleting or otherwise not completing the loop. Because of how this is stored in Cosmos, it's also not easy for us to connect the downvoted comment to the follow-up, making it harder to ascertain whether the follow-up was performed.

We plan to use up- and downvotes as part of our confidence scores, so having them engage with this mechanism is important.

**Alternative**

When GH Copilot leaves a PR comment, it uses the same thumbs up/down mechanism for feedback. However, when a thumbs-down is given, it automatically brings up a pop-up like this:

<img width="659" height="587" alt="Image" src="https://github.com/user-attachments/assets/f9a38d49-8476-4ebe-a2dd-2d34b9dc7556" />

This is great because it allows the user to pick the "why it's wrong" reasons instead of having to type them, allows them to provide more detail, and then just hit submit. We could immediately store that with the downvoted comment (making it easier for us to audit), and package that as if it were an `@azure-sdk` mention. It should help us drive greater follow-up on comments.

It would also be logical to require this pop-up if an architect _deletes_ a comment, as this like a "super downvote".

Finally, the pop-up would help clarify that a downvote shouldn't be used for an arguably valid comment that the architect just doesn't agree with. I've done this plenty with GH, initially giving a suggestion a thumbs down but then realizing it's not necessarily an invalid or wrong comment, just that I don't side with the AI on the suggestion. In that case, the AI quality is truly neutral and no up or down vote is appropriate.

ccbarragan:
@tjprescott I like this proposal a lot, makes sense to reaffirm what the thumbs up / down actions actually do. 

Curious if users might also want the ability to **react** to a comment as opposed to leaving feedback around its quality (sounds like they might). Common pattern is to separate these two actions. Example:

<img width="1472" height="864" alt="Image" src="https://github.com/user-attachments/assets/5e6e30a5-fdde-4106-8767-f0a0ff37a0c7" />

tjprescott:
For APIView, I think adding reactions could muddy our feedback story.

[...]

PR #12786, opened 2025-11-05
[API View] Feedback mechanism for AI Generated comments

AlitzelMendez:
closes: https://github.com/Azure/azure-sdk-tools/issues/12464
Implements optional feedback collection when users downvote or delete AI-generated comments to help improve AI comment quality.

When users interact negatively with AI-generated comments (downvote 👎 or delete 🗑️), a feedback dialog appears asking why. Users must select at least one reason from 4 predefined options and can optionally provide additional details.

**Feedback Options:**
- Information is factually incorrect
- APIView tool limitation or quirk  
- Azure SDK design decision/guideline
- Other

### Key Behavior
**Single Comment Operations:**
• ✅ Downvote requires selecting at least one feedback reason
• ✅ Feedback dialog appears when thumbs down is clicked
• ✅ Users must select at least one reason from 4 options
• ✅ Users can optionally provide additional context
• ✅ Downvote only persists if feedback is submitted (not canceled)
• ✅ Feedback stored with timestamp and user attribution

**Deletion (Both Single & Bulk):**
• ✅ Deletion requires provide a reason
• ✅ Delete button remains disabled until reason is selected
• ✅ Optional additional comments can be provided
• ✅ No separate confirmation dialog - feedback collection replaces it
• ✅ Feedback stored with `isDelete: true` flag

**Bulk Operations:**
• ✅ Downvote feedback integrated inline within bulk resolution dialog
• ✅ Feedback section appears when thumbs down is selected
• ✅ Users must provide at least one reason for downvote to persist
• ✅ Feedback section expands/collapses based on vote selection
• ✅ Multiple reasons can be selected via multiselect dropdown
• ✅ Same feedback flow for both single and bulk operations

### Changes
**Frontend:**
- New reusable feedback dialog component
- Optimized comment thread state management
- Full dark mode support

**Backend:**
- New API endpoint: `POST /comments/{reviewId}/{commentId}/feedback`
- Feedback stored on `CommentItemModel.Feedback` property
- Includes comprehensive unit tests

## Demo 
https://github.com/user-attachments/assets/52441d53-f214-4e21-a07e-2e1a968a12a5

## Bulk Operation feedback demo

https://github.com/user-attachments/assets/ba9ead62-6ca6-4271-bfa5-3e0048c8ff2f

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/LeanModels/CommentItemModel.cs
+++ b/src/dotnet/APIView/APIViewWeb/LeanModels/CommentItemModel.cs
@@ -1,4 +1,3 @@
-#nullable enable
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
 using System;
@@ -73,7 +72,7 @@ namespace APIViewWeb.LeanModels
         public List<string> MemoryIds { get; set; } = [];
         public float ConfidenceScore { get; set; }
 
-        public List<CommentFeedback>? Feedback { get; set; } = [];
+        public List<CommentFeedback> Feedback { get; set; } = [];
         public static CommentSeverity ParseSeverity(string value)
         {
             return value?.ToUpperInvariant() switch


tjprescott:
image
@AlitzelMendez  this is no longer true, right?

src/dotnet/APIView/APIViewWeb/LeanModels/CommentItemModel.cs
```diff
@@ -32,6 +32,15 @@ public enum CommentSource
         Diagnostic
     }
 
+    public class CommentFeedback
+    {
+        public List<string> Reasons { get; set; } = new List<string>();
+        public string Comment { get; set; } = string.Empty;
+        public bool IsDelete { get; set; }
+        public string SubmittedBy { get; set; }
+        public DateTime? SubmittedOn { get; set; }
```
tjprescott: Why is this optional?

src/dotnet/APIView/APIViewWeb/LeanModels/CommentItemModel.cs
```diff
@@ -63,6 +72,7 @@ public class CommentItemModel
         public List<string> MemoryIds { get; set; } = [];
         public float ConfidenceScore { get; set; }
 
+        public List<CommentFeedback> Feedback { get; set; } = [];
```
tjprescott, line 75: This should probably be nullable. Since it only even applies to AI-generated comments, there's no point storing even empty list for non-AI comments.

src/dotnet/APIView/APIViewWeb/LeanModels/CommentFeedbackRequest.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+
+namespace APIViewWeb.LeanModels
+{
+    public class CommentFeedbackRequest
+    {
+        public List<string> Reasons { get; set; } = new();
```
tjprescott: My gut tells me we should store these as enums, as that will allow copilot to take deterministic action instead of having to parser intent.

src/dotnet/APIView/ClientSPA/src/app/_components/shared/ai-comment-feedback-dialog/ai-comment-feedback-dialog.component.ts
```diff
@@ -0,0 +1,73 @@
...
+  selectedReasons: string[] = [];
+  additionalComments: string = '';
+
+  readonly feedbackReasons = [
+    'This comment is factually incorrect',
+    'This is an APIView rendering bug',
+    'This is an accepted APIView rendering choice',
+    'This is an accepted SDK design pattern',
+    'The guideline cited here is out-of-date'
+  ];
```
tjprescott: I think we should map these to enum values and store them as such.

AlitzelMendez:
> I think we should store the "reasons" as enums. The dialog should also appear when using bulk resolution (currently on UX Test it doesn't).

like when clicking thumbs down inside the bulk resolution dialog, trigger this feedback pop up?

@ccbarragan could you provide your feedback in here? what would be a good way to integrate this on the bulk resolution? 
I have the feeling that triggering another pop up is to many layers of interaction, but is this, ok? or do you have another suggestion on how to add this?

tjprescott:
> like when clicking thumbs down inside the bulk resolution dialog, trigger this feedback pop up?

Yes. Bulk resolution is supposed to be the same as repeating the process for each item, so if downvoting an individual item triggers the dialogue, then so should downvoting a batch.

AlitzelMendez:
> Yes. Bulk resolution is supposed to be the same as repeating the process for each item, so if downvoting an individual item triggers the dialogue, then so should downvoting a batch.

Oh yes! I just want to think about ways to implement this… Should we create another pop-up? Expand a window inside the same pop-up? Add a small window to the right? Remove the option to provide more comments (since they already have the comment window) and instead show a small, clean window with options stating that it’s required to select at least one option—but they don’t need to click Submit? well but they can select more than one!

I will try to mock this ideas tomorrow morning

AlitzelMendez:
@ccbarragan here are some suggestions/questions

## Inline Feedback Section
In this one we have the control of, if you selected downvote you have to provide a reason, the button gets disable after clicking downvote.

It makes the experience harder, but we guarantee feedback if thumb down is provided

https://github.com/user-attachments/assets/2f81e940-e317-4d85-bada-34062f1c2e51

## Feedback After Resolve
This one is a more common pattern, but I think everyone is going to close it, not providing feedback and we will be on a weird place of, should we not downvote then? which is not going to be clear/easily visible

https://github.com/user-attachments/assets/ed61ccd4-cb83-4912-9ebf-297e7e34ba19

## Layered Dialog
Now that I see implemented it looks good! :) but still I have the question if this is a good/acceptable user experience.

The workflow would be similar at the current one, if not reason provided.... the thumbs downs don't persist

https://github.com/user-attachments/assets/403cbaa6-bfce-40f9-9595-339459681ffc

tjprescott:
@AlitzelMendez thanks for these! I think we should consider whether we want bulk operations to be like one big transaction or discrete. The answer I think helps inform which approach is the "right" one.

ccbarragan:
@AlitzelMendez awesome explorations! shared some thoughts at our UX sync today. My preference on the three would be the first option inline feedback section, but displaying the feedback checkbox list after message / thumbs down / input field (instead of above).

AlitzelMendez:
Updating with offline conversations to have the whole context here

- For delete operation we want to always request a reason, if a reason is not provided, we should not delete 
image

- We should have a similar experience on bulk resolution

src/dotnet/APIView/APIViewWeb/LeanModels/CommentItemModel.cs
AlitzelMendez: changed

src/dotnet/APIView/APIViewWeb/LeanModels/CommentFeedbackRequest.cs
AlitzelMendez: make sense, changed

src/dotnet/APIView/APIViewWeb/LeanModels/CommentItemModel.cs
AlitzelMendez: oh no wait! this is be a lot more problematic if it is null, first we will need to check if null before adding and will habe some implications in the whole class having collection as a possible null

PR #12791, opened 2025-11-06
Add documentation for environment variables

deyaaeldeen:


maririos (changes requested):
Blocking this PR until the team is in agreement with `Using non-Azure OpenAI providers`

PR #12811, opened 2025-11-07
Add CODEOWNERS for tsp-client version control

raych1:


.github/CODEOWNERS
```diff
@@ -24,6 +24,11 @@
 ###########
 /eng/common/pipelines/codeowners-linter.yml @weshaggard
 
+###########
+# Files used to manage tsp-client version and installation
+###########
+/eng/common/tsp-client @catalinaperalta @raych1 @azure/azure-sdk-eng
```
jsquire: Aren't you already in the eng group?  Why do you need a named entry as well?

weshaggard: I think we should probably remove azure-sdk-eng from this one.

raych1: Removed eng group entry.

PR #12812, opened 2025-11-07
Updated the schema of swagger_to_sdk_config

raych1:
Extended the schema for some integrated scripts.

### DIFF
--- a/tools/spec-gen-sdk/src/types/SwaggerToSdkConfigSchema.json
+++ b/tools/spec-gen-sdk/src/types/SwaggerToSdkConfigSchema.json
@@ -149,6 +149,22 @@
             }
           }
         },
+        "updateChangelogContentScript": {
+          // Update changelog content.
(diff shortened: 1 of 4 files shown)


tools/spec-gen-sdk/src/types/SwaggerToSdkConfigSchema.json
```diff
@@ -149,6 +149,22 @@
             }
           }
         },
+        "updateChangelogContentScript": {
+          // Update changelog content.
+          "$ref": "#/definitions/RunOptions"
+        },
+        "updateCiScript": {
```
weshaggard, line 156: Is this one still needed?

raych1: Yes, it needs to be supported in the package update stage but not in the scope of milestone 1. I prefer to add it together with other options in this PR.

PR #12820, opened 2025-11-10
Language check is case sensitive in release tool

### LINKED ISSUE #12768: Language check is case sensitive in release tool
https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs#L66

Language validation is case sensitive in release tool. So request is failing if agent passes python instead of Python.

Fix: Check should not be case sensitive and release pipeline should run irrespective of the language casing.

jeo02:
- changed to case insensitive
- Resolves #12768

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs
@@ -63,7 +63,7 @@ namespace Azure.Sdk.Tools.Cli.Tools.Package
                     response.ReleaseStatusDetails = "Package name cannot be null or empty. ";
                     isValidParams = false;
                 }
-                if (string.IsNullOrWhiteSpace(language) || !ValidLanguages.Contains(language))
+                if (string.IsNullOrWhiteSpace(language) || !ValidLanguages.Contains(language, StringComparer.OrdinalIgnoreCase))
                 {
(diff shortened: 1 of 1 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs
```diff
@@ -63,7 +63,7 @@ public async Task<SdkReleaseResponse> ReleasePackageAsync(string packageName, st
                     response.ReleaseStatusDetails = "Package name cannot be null or empty. ";
                     isValidParams = false;
                 }
-                if (string.IsNullOrWhiteSpace(language) || !ValidLanguages.Contains(language))
+                if (string.IsNullOrWhiteSpace(language) || !ValidLanguages.Contains(language, StringComparer.OrdinalIgnoreCase))
```
maririos, line 66: I haven't checked, but are there any other places where we need to be explicit about it?

praveenkuttappan (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs
jeo02: I can poke around I have no doubt that it's the only place

PR #12826, opened 2025-11-11
Add sample translation

deyaaeldeen:
This pull request introduces a new feature to the Azure SDK CLI tool that enables sample code translation between programming languages. The main addition is the `SampleTranslatorTool`, which automates the discovery, translation, and output of sample files from one language to another using AI-powered translation. This tool is now integrated into the CLI and supports batch processing, context-aware translation, and output management.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
@@ -38,8 +38,6 @@ namespace Azure.Sdk.Tools.Cli.Tools.Samples
     {
         private readonly ILanguageSpecificResolver<SampleLanguageContext> sampleContextResolver;
         private readonly IMicroagentHostService microagentHostService;
-        private readonly IFileHelper fileHelper;
-
         public SampleTranslatorTool(
             IMicroagentHostService microagentHostService,
             ILogger<SampleTranslatorTool> logger,
@@ -51,7 +49,6 @@ namespace Azure.Sdk.Tools.Cli.Tools.Samples
         {
             this.sampleContextResolver = sampleContextResolver;
             this.microagentHostService = microagentHostService;
-            this.fileHelper = fileHelper;
         }
 
         public override CommandGroup[] CommandHierarchy { get; set; } = [SharedCommandGroups.Samples];
@@ -81,12 +78,20 @@ namespace Azure.Sdk.Tools.Cli.Tools.Samples
             Required = false
         };
 
+        private readonly Option<int> batchSizeOption = new("--batch-size")
+        {
+            Description = "Number of samples to process in each batch to avoid overwhelming the AI model",
+            Required = false,
+            DefaultValueFactory = _ => SampleConstants.DefaultBatchSize,
+        };
+
         protected override Command GetCommand() => new("translate", "Translates sample files from source language to target package language")
         {
             fromOption,
             toOption,
             overwriteOption,
-            modelOption
+            modelOption,
+            batchSizeOption
         };
 
         public override async Task<CommandResponse> HandleCommand(ParseResult parseResult, CancellationToken ct)
@@ -95,10 +100,11 @@ namespace Azure.Sdk.Tools.Cli.Tools.Samples
             string toPackagePath = parseResult.GetValue(toOption) ?? throw new ArgumentException("to is required");
             bool overwrite = parseResult.GetValue(overwriteOption);
             var model = parseResult.GetValue(modelOption);
+            int batchSize = parseResult.GetValue(batchSizeOption);
 
             try
             {
-                await TranslateSamplesAsync(fromPackagePath, toPackagePath, overwrite, model, ct);
+                await TranslateSamplesAsync(fromPackagePath, toPackagePath, overwrite, model, batchSize, ct);
                 return new DefaultCommandResponse { Message = "Sample translation completed successfully" };
             }
             catch (ArgumentException ex)
@@ -113,7 +119,7 @@ namespace Azure.Sdk.Tools.Cli.Tools.Samples
             }
         }
 
-        private async Task TranslateSamplesAsync(string fromPackagePath, string toPackagePath, bool overwrite, string? model, CancellationToken ct)
+        private async Task TranslateSamplesAsync(string fromPackagePath, string toPackagePath, bool overwrite, string? model, int batchSize, CancellationToken ct)
         {
             // Determine source language and get samples directory from source package path
             var sourceLanguageService = GetLanguageService(fromPackagePath) ?? throw new ArgumentException("Unable to determine source language for package (resolver returned null). Ensure repository structure and Language-Settings.ps1 are correct.");
@@ -144,7 +150,7 @@ namespace Azure.Sdk.Tools.Cli.Tools.Samples
             logger.LogInformation("Found {count} sample files to translate", sampleFiles.Count);
 
             // Load target package context for translation
-            var targetPackageContext = await targetContext.LoadContextAsync([packageInfo.PackagePath], 2000000, 50000, ct);
+            var targetPackageContext = await targetContext.LoadContextAsync([packageInfo.PackagePath], SampleConstants.MaxContextCharacters, SampleConstants.MaxCharactersPerFile, ct);
             
             if (string.IsNullOrWhiteSpace(targetPackageContext))
             {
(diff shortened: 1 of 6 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
```diff
@@ -0,0 +1,355 @@
...
+using Azure.Sdk.Tools.Cli.Helpers;
+using Azure.Sdk.Tools.Cli.Microagents;
+using Azure.Sdk.Tools.Cli.Models;
+using ModelContextProtocol.Server;
+using Azure.Sdk.Tools.Cli.Services;
+using Azure.Sdk.Tools.Cli.Samples;
+using Azure.Sdk.Tools.Cli.Tools.Core;
+using Azure.Sdk.Tools.Cli.Services.Languages;
+
+namespace Azure.Sdk.Tools.Cli.Tools.Samples
```
praveenkuttappan, line 17: shouldn't this be under Tools.Package namespace? I see that the tool is taking package path as a param. I think in general all sample tool classes should be moved under package namespace.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
```diff
@@ -0,0 +1,355 @@
...
+        public override async Task<CommandResponse> HandleCommand(ParseResult parseResult, CancellationToken ct)
+        {
+            string fromPackagePath = parseResult.GetValue(fromOption) ?? throw new ArgumentException("from is required");
+            string toPackagePath = parseResult.GetValue(toOption) ?? throw new ArgumentException("to is required");
+            bool overwrite = parseResult.GetValue(overwriteOption);
+            var model = parseResult.GetValue(modelOption);
+
+            try
+            {
+                await TranslateSamplesAsync(fromPackagePath, toPackagePath, overwrite, model, ct);
```
praveenkuttappan: Is it intentional not to expose this as a MCP tool and keep it as a CLI command only?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
```diff
@@ -0,0 +1,355 @@
...
+
+        protected override Command GetCommand() => new("translate", "Translates sample files from source language to target package language")
+        {
+            fromOption,
+            toOption,
+            overwriteOption,
+            modelOption
+        };
+
+        public override async Task<CommandResponse> HandleCommand(ParseResult parseResult, CancellationToken ct)
```
praveenkuttappan, line 97: Can you create a custom response model for samples that inherits PackageResponseBase class? Response should contain the proeprties like `language`, `package name` etc whenever applicable so these informations are also added in telemetry. I am in the middle of updating new tool documentation to update response type requirement.

christothes (approved):

timovv (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
deyaaeldeen: Samples is not under package at the moment. There is no support for nested namespaces yet. Ideally, I want this to be azsdk package samples translate.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
deyaaeldeen: Yes

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
deyaaeldeen: That is a good point. However, let me address this in a follow-up PR because it is more involved and the sample generator doesn't have this either yet.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
```diff
@@ -0,0 +1,361 @@
...
+        };
+
+        private readonly Option<int> batchSizeOption = new("--batch-size")
+        {
+            Description = "Number of samples to process in each batch to avoid overwhelming the AI model",
+            Required = false,
+            DefaultValueFactory = _ => SampleConstants.DefaultBatchSize,
+        };
+
+        protected override Command GetCommand() => new("translate", "Translates sample files from source language to target package language")
```
deyaaeldeen, line 88: @maririos @ronniegeraghty You brought up today in the sync whether this can be renamed. I still think translate is a good name and copilot seems to [agree](https://m365.cloud.microsoft/chat/entity1-d870f6cd-4aa5-4d42-9626-ab690c041429/eyJpZCI6IlZYTmxjbFl4ZkdoMGRIQnpPaTh2YzNWaWMzUnlZWFJsTFdsdWRDNXZabVpwWTJVdVkyOXRMM3hQU1VRNlpETXlZVFZsTkRVdE9XUmxOQzAwWkdVMExXSTBNMlF0WVRFNU0yVm1PV1ppT1dVeGZHTmtZVEl5WkdJMExUQmxZMk10TkRnM05TMDVOak5pTFRaa05qY3hNV0k1WlRCaU9Yd3lNREkxTFRFeExURXhWREU1T2pFd09qVXdMalEzT1RrNE1ETmEiLCJzY2VuYXJpbyI6InNoYXJlTGlua1ZpYVJpY2hDb3B5IiwicHJvcGVydGllcyI6eyJwcm9tcHRTb3VyY2UiOiJ1c2VyIiwiY2xpY2tUaW1lc3RhbXAiOiIyMDI1LTExLTExVDE5OjEwOjQ5LjU5OVoifSwiY2hhdFR5cGUiOiJ3ZWIiLCJ2ZXJzaW9uIjoxLjF9). Let me know what do you think.

timovv (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
```diff
@@ -0,0 +1,361 @@
...
+    /// <summary>
+    /// Represents a source sample file discovered for translation.
+    /// </summary>
+    /// <param name="FilePath">The full path to the sample file</param>
+    /// <param name="Content">The content of the sample file</param>
+    /// <param name="SourceLanguage">The detected source language</param>
+    /// <param name="Extension">The file extension</param>
+    internal record SourceSampleFile(string FilePath, string Content, string SourceLanguage, string Extension);
+
+    [McpServerToolType, Description("Translates sample files from one language to another")]
```
praveenkuttappan, line 36: Should we say `convert` instead of translate? Similarly call SampleConverterTool instead of SampleTranslatorTool? 
//cc: @maririos

deyaaeldeen: I already have a discussion on this in https://github.com/Azure/azure-sdk-tools/pull/12826#pullrequestreview-3449686383. Let's keep the discussion in one place.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Samples/SampleTranslatorTool.cs
maririos: Copilot will agree with what you ask :)
> You’re absolutely right—“translate” usually implies natural language (e.g., English → Spanish), which could confuse developers. For programming languages, we want something that clearly signals code adaptation or generation, not text translation.
Best Fit: If the feature takes existing samples and outputs equivalent ones in another language, samples convert or samples port are the most intuitive.
If the feature uses AI to create new samples from scratch, samples generate feels right.

Using the word `Convert` makes sense to me.
Input -> convert -> output

praveenkuttappan: yup. Convert and ConverterTool seem better.

deyaaeldeen: Copilot isn’t reasoning here, it simply suggests translate by default; you can verify that by running the same prompt yourself. I understand and agree that “translate” can carry different meanings depending on context but that’s true for many technical terms. In the context of samples, however, it’s hard to imagine a realistic scenario where “translate” would be misunderstood as anything other than what we intend. Let’s keep the current naming for now and revisit only if we see clear customer feedback suggesting otherwise.

PR #12843, opened 2025-11-11
Added update-version tool

raych1:
This is [the design spec](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/docs/specs/6-package-updater.spec.md)

Changes Overview:
- for mgmt-plane package, run `update-version` script
- for data-plane package, return `noop` with guidance on next steps
- include `running verify setup tool` in the next steps for failure responses
- updated changelog
- updated docs


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
```diff
@@ -0,0 +1,242 @@
...
+                            $"Version updated" + 
+                                (!string.IsNullOrWhiteSpace(releaseDate) ? $" and release date set to {releaseDate}." : "."),
+                            ["Review the updated version and release date", "Run validation checks"]);
+                    }
+                }
+            }
+
+            // Run default logic to update version
+            logger.LogInformation("Running default logic to update version for the package...");            
+            return await languageService.UpdateVersionAsync(packagePath, version, releaseDate, ct);
```
weshaggard: Shouldn't you also pass down releaseType?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
```diff
@@ -0,0 +1,242 @@
...
+                    var scriptParameters = new Dictionary<string, string>
+                    {
+                        { "SdkRepoPath", sdkRepoRoot },
+                        { "PackagePath", packagePath }
+                    };
+
+                    // Add optional parameters if provided
+                    if (!string.IsNullOrWhiteSpace(releaseType))
+                    {
+                        scriptParameters["ReleaseType"] = releaseType;
```
weshaggard, line 169: Shouldn't we always pass these values even if they are empty?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
```diff
@@ -0,0 +1,242 @@
...
+
+            if (!Directory.Exists(packagePath))
+            {
+                return PackageOperationResponse.CreateFailure($"Package path does not exist: {packagePath}");
+            }
+
+            // Set default release type to 'beta' if both Version and ReleaseType are empty or null
+            if (string.IsNullOrWhiteSpace(version) && string.IsNullOrWhiteSpace(releaseType))
+            {
+                releaseType = "beta";
```
weshaggard, line 131: Do we plan to default the releaseDate as well? Or do we want to let the languages handle that?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/SpecGenSdkConfigHelper.cs
```diff
@@ -276,7 +276,15 @@ public async Task<PackageOperationResponse> ExecuteProcessAsync(
 
                 if (result.ExitCode != 0)
                 {
-                    return PackageOperationResponse.CreateFailure($"Process failed with exit code {result.ExitCode}. Output:\n{trimmedOutput}", packageInfo);
+                    return PackageOperationResponse.CreateFailure(
+                        $"Process failed with exit code {result.ExitCode}. Output:\n{trimmedOutput}",
+                        packageInfo,
+                        nextSteps: [
+                            "Check the running logs for details about the error",
```
weshaggard: For these next steps wouldn't it be better to use the tool names instead of duplicated text we hope will pick the right tool?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
raych1: Currently it's only used by the language-specific script and isn't needed here.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
raych1: They're optional parameters and don't have to be passed in when they're empty. Script can distinguish between `not provided` vs `explicitly empty`.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
raych1: It makes sense to have a default value; we should set the current date as the default in the tool and pass it to the language script.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
weshaggard: I would like to keep the args consistent for both the language contract and the script contract.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/SpecGenSdkConfigHelper.cs
raych1: MCP tool name is slightly different from the CLI name. e.g. `azsdk_verify_setup` vs `azsdk verify setup`. I used the descriptive text to make it more compatible with both agent and CLI scenarios.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
raych1: Updated.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
raych1: Updated.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
```diff
@@ -0,0 +1,262 @@
...
+        ISpecGenSdkConfigHelper specGenSdkConfigHelper): base(languageServices, gitHelper, logger)
+    {
+        _specGenSdkConfigHelper = specGenSdkConfigHelper;
+    }
+
+    public override CommandGroup[] CommandHierarchy { get; set; } = [SharedCommandGroups.Package];
+
+    private const string UpdateVersionCommandName = "update-version";
+
+    private static readonly Option<string?> ReleaseTypeOption = new("--release-type", "-t")
```
weshaggard, line 37: Any reason this isn't `--releaseType` to match the parameters other places like in the script call?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/SpecGenSdkConfigHelper.cs
weshaggard: OK we can try the text but there are 2 things that don't sit well with me. First we are duplicating these strings in various places and second, we are assuming this text will be enough for the LLM to find the appropriate tool we want it to run next. This is something we should keep an eye on in testing.

weshaggard (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
raych1: We use the [kebab-format](https://github.com/Azure/azure-sdk-tools/blob/8f39e5daac2697d23815da8110f89a1744fc190f/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedOptions.cs#L77) for the CLI parameter name.

weshaggard: Ok something we might want to think a little more about to make things consistent between cli args and scripts we call with those args.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/SpecGenSdkConfigHelper.cs
raych1: I will definitely keep an eye on how the agent responds on the `next_steps`.

I made these next steps as default value in the `CreateFailure` method to remove the duplicated text.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
```diff
@@ -0,0 +1,255 @@
...
+            logger.LogInformation("Repository root discovered: {SdkRepoRoot}", sdkRepoRoot);
+            var languageService = GetLanguageService(packagePath);
+            if (languageService == null)
+            {
+                return PackageOperationResponse.CreateFailure("Tooling error: unable to determine language service for the specified package path.", nextSteps: ["Create an issue at the https://github.com/Azure/azure-sdk-tools/issues/new", "contact the Azure SDK team for assistance."]);
+            }
+
+            // Check for package type
+            var packageInfo = await languageService.GetPackageInfo(packagePath, ct);
+            if (packageInfo?.SdkType == SdkType.Management)
```
raych1, line 151: @haolingdong-msft @haiyuazhang @msyyc @MaryGao @tadelesh @skywing918 
this is the place to determine if a package is mgmt-plane type or data-plane type.

Both the `update-changelog-content` tool and the `update-version` tool have this logic in tool's code. Language script doesn't need to check on the SDK type unless script has different logic per SDK type.

The `update-metadata` tool doesn't have this logic since language script is expected to handle both mgmt-plane SDK and data-plane SDK. Let me know if it's not the case.

Currently the `languageService.GetPackageInfo` method doesn't set the `SdkType`. We already have issues created for redmond language team to work on this. 

Before the `languageService.GetPackageInfo` supports the `SdkType` setting, you will need to take two actions for your local code base in order to do integration testing. 
- add the following one-line code before line#151
```c#
packageInfo.SdkType = SdkType.Management;
```
- run `dotnet build` for this project

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/VersionUpdateTool.cs
```diff
@@ -0,0 +1,255 @@
...
+            {
+                // For management-plane packages, execute configured version update script
+                var (configContentType, configValue) = await _specGenSdkConfigHelper.GetConfigurationAsync(sdkRepoRoot, SpecGenSdkConfigType.UpdateVersion);
+                if (configContentType != SpecGenSdkConfigContentType.Unknown && !string.IsNullOrEmpty(configValue))
+                {
+                    logger.LogInformation("Found valid configuration for updating version. Executing configured script...");
+
+                    // Prepare script parameters
+                    var scriptParameters = new Dictionary<string, string>
```
raych1, line 160: @haolingdong-msft @haiyuazhang @msyyc @MaryGao @tadelesh @skywing918 
this is the place to prepare the parameters for the language script call.

haolingdong-msft: Thanks for tagging us to highlight this part. It is quite helpful for us to do integration!

PR #12855, opened 2025-11-12
Grab package info using the script, which is the authority

### LINKED ISSUE #12838: Package Info is missing SDK type for Go

https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.cs#L40

Currently SDK type is set in CI and Engsys tools using the following code, and we should use the same logic to set the properties in MCP package info so values match.
https://github.com/Azure/azure-sdk-for-go/blob/main/eng/scripts/Language-Settings.ps1#L63

richardpark-msft:
- Grab package info using the script, which is the authority. Otherwise we'll just fall out of sync.
- Addressing some lint type issues that showed up
- Renaming files to match the class.

Fixes #12838


richardpark-msft:
@praveenkuttappan, this just gets the information _from_ the script. Until we're ready to do a full port I don't think it's a good idea to have two copies of the code that have no checks on each other.

praveenkuttappan (approved):
Looks good

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.cs
```diff
@@ -39,131 +35,76 @@ public GoLanguageService(
...
+        //   "VersionFile": "/home/ripark/src/az/sdk/messaging/azservicebus/internal/constants.go",
+        //   "ModuleName": "azservicebus"
+        // }
         logger.LogDebug("Resolving Go package info for path: {packagePath}", packagePath);
-        var (repoRoot, relativePath, fullPath) = Parse(gitHelper, packagePath);
-        var (packageName, packageVersion) = await TryGetPackageInfoAsync(fullPath, ct);
-        
-        if (packageName == null)
+        string[] args = ["-c", $". {commonPS1}; Get-GoModuleProperties('{packagePath}') | ConvertTo-Json"];
+        var processResult = await processHelper.Run(new ProcessOptions(
```
weshaggard: Please use PowershellOptions https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/Options/PowershellOptions.cs

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.cs
```diff
@@ -39,131 +35,76 @@ public GoLanguageService(
...
+            workingDirectory: repoRoot), ct);
+
+        if (processResult.ExitCode != 0)
         {
-            logger.LogWarning("Could not determine package name for Go package at {fullPath}", fullPath);
+            throw new Exception($"Failed to extract package properties for {packagePath}: {processResult.Output}");
         }
-        if (packageVersion == null)
+
+        var goModuleProperties = JsonSerializer.Deserialize<GoModulePropertiesPowershell>(processResult.Output)
```
weshaggard: While parsing the stdout might work in can be cause trouble if someone decides to output other random crap. We probably want to move writing this to a file instead. Similar to https://github.com/Azure/azure-sdk-for-go/blob/main/eng/common/scripts/Save-Package-Properties.ps1 but I don't think that will work in its current form, so we can stick with this for now.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.cs
```diff
@@ -39,131 +35,76 @@ public GoLanguageService(
...
+        //   },
+        //   "VersionFile": "/home/ripark/src/az/sdk/messaging/azservicebus/internal/constants.go",
+        //   "ModuleName": "azservicebus"
+        // }
         logger.LogDebug("Resolving Go package info for path: {packagePath}", packagePath);
-        var (repoRoot, relativePath, fullPath) = Parse(gitHelper, packagePath);
-        var (packageName, packageVersion) = await TryGetPackageInfoAsync(fullPath, ct);
-        
-        if (packageName == null)
+        string[] args = ["-c", $". {commonPS1}; Get-GoModuleProperties('{packagePath}') | ConvertTo-Json"];
```
weshaggard: I'm good with this for now but I do think we should eventually move the implementation from the script into the c# code and invert this dependency and have the script call the cli.

raych1: Why don't we move this implementation up to the base class given it just calls the common script, and it is same across languages?

weshaggard: We could but we would need to go through the common helper instead of calling `Get-GoModuleProperties` directly.

richardpark-msft: > I'm good with this for now but I do think we should eventually move the implementation from the script into the c# code and invert this dependency and have the script call the cli.

Agreed, this is just a simple stop-gap that makes sure we don't have to maintain two implementations. I'd like to just cut-over at some point.

> Why don't we move this implementation up to the base class given it just calls the common script, and it is same across languages?

@raych1, I think we're hoping people will just spend time porting over the logic they need, into C#, which I'll eventually do. This stop-gap is just in place so we can work on that as time allows, without blocking anyone.

raych1: The `Get-GoModuleProperties` function just [check on the module name](https://github.com/Azure/azure-sdk-for-go/blob/7f2ba924c3449c39e38df89e4a481a0ea9824524/eng/scripts/Language-Settings.ps1#L63) to determine the `SdkType`. I think we should do that in C# directly instead of calling the script.

Other languages except JS also have similar logic checking on the package name to determine the `SdkType`.

raych1: @richardpark-msft I meant we could just set the `SdkType` for now based on the module name because it's the current request. For the other missing properties, we could port them later.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Services/Languages/GoLanguageServicesTests.cs
```diff
@@ -0,0 +1,210 @@
...
+            var resp = await LangService.AnalyzeDependencies(tempDir.DirectoryPath, false, CancellationToken.None);
+            Assert.That(resp.ExitCode, Is.EqualTo(0));
+
+            var identityLine = File.ReadAllLines(Path.Join(tempDir.DirectoryPath, "go.mod"))
+                .Where(line => line.Contains("azidentity"))
+                .Select(line => line.Trim())
+                .First();
+            Assert.That(identityLine, Is.Not.EqualTo("github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.10.0"));
+
+            resp = await LangService.FormatCode(tempDir.DirectoryPath, false, CancellationToken.None);
```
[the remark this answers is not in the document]
richardpark-msft: Fixed, but really we should have just been checking the exit code.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.cs
richardpark-msft: Done, removed some boilerplate, which was nice.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.cs
richardpark-msft: Yeah, this is always a concern - I don't know if there are good guidelines, or even enforcement, for making sure that errors/info go to stderr or not.

praveenkuttappan (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.cs
weshaggard: No there are no good guidelines or enforcements. It is safer to not rely on std-out if possible.

PR #12875, opened 2025-11-12
Move Env Setup Spec

JennyPng:
move to right directory


lmazuel (approved):
Specs were separated in two folders, this PR just moves all back together

PR #12883, opened 2025-11-13
[AVC] Add webapp authentication

tjprescott:
This PR adds authentication for the AVC webapp, which was previously lacking. This adds server side authentication in addition to AppService-level Easy Auth.

### DIFF
--- a/packages/python-packages/apiview-copilot/cli.py
+++ b/packages/python-packages/apiview-copilot/cli.py
@@ -1,7 +1,3 @@
-import logging
-
-from azure.core.exceptions import ClientAuthenticationError
-
 # -------------------------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for
@@ -16,6 +12,7 @@ Command line interface for APIView Copilot.
 
 import asyncio
 import json
+import logging
 import os
 import pathlib
 import sys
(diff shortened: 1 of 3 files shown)


packages/python-packages/apiview-copilot/cli.py
```diff
@@ -953,6 +959,24 @@ def analyze_comments(language: str, start_date: str, end_date: str, environment:
...
+def _build_auth_header(base_url):
+    """
+    Helper to build Authorization header with Bearer token for WEBAPP_ENDPOINT requests.
+    """
+    from src._credential import get_credential
+
+    credential = get_credential()
+    # Determine correct scope based on endpoint
+    if "apiview.dev" in base_url:
+        scope = "api://apiview/.default"
```
tjprescott: I should include the APP_ID for production here before merging.

packages/python-packages/apiview-copilot/cli.py
```diff
@@ -953,6 +959,24 @@ def analyze_comments(language: str, start_date: str, end_date: str, environment:
     print(f"Unique CreatedBy values ({len(created_by_set)}): {sorted(created_by_set)}")
 
 
+def _build_auth_header(base_url):
```
tjprescott, line 964: @AlitzelMendez you will need something similar to this in APIView to reconnect to it.

PR #12885, opened 2025-11-13
Update generation tools to use tsp-client helper

raych1:
Key Changes
- Moved `tsp-client init` call to the `TspClientHelper` class
- Migrated the generation tool to use the `TspClientHelper`
- Consolidated the package response creation methods


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkGenerationTool.cs
```diff
@@ -42,7 +42,7 @@ INpxHelper npxHelper
             Required = false,
         };
 
-        private readonly Option<string> emitterOpt = new("--emitter-options", "-o")
+        private readonly Option<string> emitterOpt = new("--emitter-options", "-e")
         {
             Description = "Emitter options in key-value format. Example: 'package-version=1.0.0-beta.1'",
```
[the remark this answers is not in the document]
raych1: Changed it because of the conflicts with --output flag.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TspClientHelper.cs
```diff
@@ -96,4 +101,40 @@ public async Task<TspToolResponse> UpdateGenerationAsync(string tspLocationPath,
...
+        {
+            arguments.AddRange(additionalArgs);
+        }
+
+        var npxOptions = new NpxOptions(
+            "@azure-tools/typespec-client-generator-cli",
+            arguments.ToArray(),
+            logOutputStream: true,
+            workingDirectory: workingDirectory
+        );
```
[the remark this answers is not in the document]
raych1: Added it back.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TspClientHelper.cs
```diff
@@ -96,4 +103,41 @@ public async Task<TspToolResponse> UpdateGenerationAsync(string tspLocationPath,
             TypeSpecProject = outputDirectory
         };
     }
+
+    public async Task<TspToolResponse> InitializeGenerationAsync(string workingDirectory, string tspConfigPath, string[]? additionalArgs = null, CancellationToken ct = default)
+    {
+        logger.LogInformation("tsp-client init: {tspConfig} in {workingDir}", tspConfigPath, workingDirectory);
+
+        // Build arguments list dynamically
+        var arguments = new List<string> { "tsp-client", "init", "--update-if-exists", "--tsp-config", tspConfigPath };
```
weshaggard, line 112: Do we even need the update now that we are using "--update-if-exits"?

weshaggard (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TspClientHelper.cs
raych1: The `customized-code` tool uses the update command.

samvaity (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TspClientHelper.cs
weshaggard: It is find to keep but I think they could use this now as well. @catalinaperalta I know we talked about potentially deprecating "update" but I don't know if that is something we will ever do in practice.

raych1: The `customized-code` scenario happens in the SDK repo. Using the `init` command would need additional work, such as constructing the URL for `tspconfig.yaml` or identifying the local typespec project folder, because `tspconfig` is the required parameter for the `init` command.  CC: @samvaity

catalinaperalta: tsp-client init (with or without the update-if-exists flag) is run from the root of the repo, while tsp-client update is run from the client library directory. Most service partners use tsp-client update during active development of their client libraries since they're typically already in that directory and can run other library specific commands for example running their tests and samples. I dont think we have enough data to say that service partners would find it easier to run the init command from the root of the repository every time they're making updates to their library.

PR #12892, opened 2025-11-13
Get correct Package Version for CodeFiles

chidozieononiwu:
The `assemblySymbol.Identity.Version` is not valid in the case of non-GA versions.
This gets the more familiar package version that corresponds to the nugget file version.
e.g. `2.0.0-alpha.20251113.57` rather than `2.0.0.0`


tools/apiview/parsers/csharp-api-parser/CSharpAPIParser/TreeToken/CodeFileBuilder.cs
```diff
@@ -81,7 +81,7 @@ public CodeFile Build(IAssemblySymbol assemblySymbol, bool runAnalysis, List<Dep
                 Language = "C#",
                 ParserVersion = CurrentVersion,
                 PackageName = assemblySymbol.Name,
-                PackageVersion = assemblySymbol.Identity.Version.ToString()
+                PackageVersion = GetAssemblyInformationalVersion(assemblySymbol) ?? assemblySymbol.Identity.Version.ToString()
```
christothes: Could we add a test for this?

christothes (approved):

Issue #12897, opened 2025-11-14
[js tool] clean up samples folder


skywing918:
if run `build sample`, should clean up the `samples` folder

PR #12898, opened 2025-11-14
Clean up Samples folder if run command for build samples

skywing918:
https://github.com/Azure/azure-sdk-tools/issues/12897

1. Clean up Samples folder if run command for build samples
2. git retry fail should be warnning


tools/js-sdk-release-tools/src/mlc/clientGenerator/modularClientPackageGenerator.ts
```diff
@@ -33,6 +33,7 @@ export async function generateAzureSDKPackage(options: ModularClientPackageOptio
         if (await exists(packageJsonPath)) originalNpmPackageInfo = await getNpmPackageInfo(packageDirectory);
 
         await cleanUpPackageDirectory(packageDirectory, options.runMode);
+        await cleanupSamplesFolder(packageDirectory);
```
MaryGao: this should be alongside with sample build, where is this step?

Issue #12265, opened 2025-09-29
Introduce separate package.json and package-lock.json JS specific dependencies installation


skywing918:
npm install -g @azure-tools/js-sdk-release-tools

skywing918:
close as completed

PR #12899, opened 2025-11-14
Pin js-sdk-release-tools under eng/common/tools

skywing918:
https://github.com/Azure/azure-sdk-tools/issues/12265
Updated usage instructions to use npm exec --prefix for running js-sdk-release-tools commands. The reason is js-sdk-release-tools requires the working directory to be the root folder of SDK repository.

### DIFF
--- a/eng/common/js-sdk-release-tools/package.json
+++ b/eng/common/js-sdk-release-tools/package.json
@@ -1,0 +1,5 @@
+{
+  "dependencies": {
+    "@azure-tools/js-sdk-release-tools": "2.14.9"
+  }
+}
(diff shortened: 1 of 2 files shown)


eng/common/js-sdk-release-tools/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@azure-tools/js-sdk-release-tools": "2.14.9"
```
weshaggard, line 3: If this is only used in the JS repo then we should just put it there and not under eng/common.

PR #12904, opened 2025-11-14
[azsdkcli] error message update to fix breaking change bug

l0lawrence:
When the services were refactored the error messages were changed which broke the package tool check to ensure that not implemneted checks dont fail the run

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/PackageCheckTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/PackageCheckTool.cs
@@ -582,7 +582,7 @@ namespace Azure.Sdk.Tools.Cli.Tools.Package
 
         private static bool IsNotImplemented(PackageCheckResponse response)
         {
-            return response.ResponseError != null && response.ResponseError.Contains("not implemented", StringComparison.OrdinalIgnoreCase);
+            return response.ResponseError != null && response.ResponseError.Contains("This is not an applicable operation for this language", StringComparison.OrdinalIgnoreCase);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/PackageCheckTool.cs
```diff
@@ -582,7 +582,7 @@ private static PackageCheckResponse CreateUnsupportedLanguageResponse(string pac
 
         private static bool IsNotImplemented(PackageCheckResponse response)
         {
-            return response.ResponseError != null && response.ResponseError.Contains("not implemented", StringComparison.OrdinalIgnoreCase);
+            return response.ResponseError != null && response.ResponseError.Contains("This is not an applicable operation for this language", StringComparison.OrdinalIgnoreCase);
```
weshaggard: Do we really want this to be the generic message? Shouldn't we have a Not Applicable and a Not Implemented? Perhaps given this is a user message a Not Implemented isn't very helpful.

l0lawrence: The responseError message is returned from the shared languageService class so I believe it will be the same for each language that doesn't have an overload. I don't think there is much value in overloading those base functions for each language to change the message to be specific to NotApplicable or NotImplemented. I have no strong feelings however if we want to change its wording to remove not applicable overall

For the package check tool itself returning "these are not implemented : List of not implemented" -- I agree there isn't a ton of value in telling them what didn't run when they select to run all validation other than they now know what that language doesn't cover validation wise. I need to catch the notImplemented checks to ensure that validation doesn't fail, but could remove the output for not implemented

weshaggard: I agree with not returning this message to the user but should we just make them no-op in the default case so there isn't any real error.

weshaggard (approved):

praveenkuttappan (approved):

Issue #12909, opened 2025-11-14
[azsdk verify setup] version flag isn't supported for gocov-xml and gocov-html command


raych1:
`gocov` tool has been installed but the command to check its installation is not right, which causes the following errors:
```text
[ERROR] Requirement check failed for gocov-xml. Error: Command gocov-xml --version failed with exit code 2. Output: flag provided but not defined: -version
Usage of gocov-xml:
  -source string
        Absolute path to source. Defaults to current working directory..
[ERROR] Requirement check failed for gocov-html. Error: Command gocov-html --version failed with exit code 2. Output: flag provided but not defined: -version
```

JennyPng:
addressing this in this PR https://github.com/Azure/azure-sdk-tools/pull/12943

PR #12943, opened 2025-11-17
Update Go Checks in Verify Setup

JennyPng:
- bug fix on some Go check commands that used nonexistent flags
- also removing mention of GOROOT because it seems like manually modifying that is not recommended

resolving bug https://github.com/Azure/azure-sdk-tools/issues/12909


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -62,7 +62,7 @@
             {
                 "requirement": "go",
                 "check": ["go", "version"],
-                "instructions": ["Download and install the Go version specified in globals.yml", "Ensure your GOPATH and GOROOT environment variables are set correctly", "Add GOROOT/bin to your PATH environment variable"]
+                "instructions": ["Download and install the Go version specified in globals.yml", "Ensure your GOPATH environment variable is set correctly", "Ensure GOPATH/bin is set in your PATH environment variable"]
```
richardpark-msft: This error message is a bit weird since GOPATH is the edge-case. I don't even have that variable set, and you wouldn't need it, so let's not send anyone down that rabbithole:

```suggestion
                "instructions": ["Download and install the latest Go versiop of the compiler on https://go.dev/dl/"]
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -62,7 +62,7 @@
...
             },
             {
                 "requirement": "go-junit-report",
```
richardpark-msft: You were asking me if Go developers need all of these tools. This is one of those examples where no person would ever install these. Either vscode provides the proper GUI for it, or they'll use some other Go tool.

AFAIK, this adapter is _only_ used in CI, since CI understands junit reports.

If this is _only_ for interactive normal Go package developers then I'd remove this step. If this is _also_ for CI, then I guess you have to leave it, but please file an issue to revisit this when you do your next design. The last thing we want is local developers having to install tools that are only required in CI.

richardpark-msft: You can also delete the 'gocov' related lines as well - they're only for CI. Ref: https://github.com/Azure/azure-sdk-for-go/issues/25217#issuecomment-3353469678.

JennyPng: since azsdk CLI will be plugged into CI in the future, I'll keep these reqs in here but add a bool so they get filtered out in non-CI mode for local development

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -64,6 +64,11 @@
                 "check": ["go", "version"],
                 "instructions": ["Download and install the latest Go version of the compiler on https://go.dev/dl/"]
             },
+            {
+                "requirement": "goimports",
+                "check": ["goimports", "."],
```
JennyPng: @richardpark-msft  `goimports` is a requirement used by another tool that I wanted to add. does this check make sense to you?

it doesn't seem to have a flag I can use as a check since running it without args causes it to hang, there's no version flag, and `goimports -help` returns an exit code of 2 for some reason which reads as a failure...

richardpark-msft: If you look in the Go language service you can see how I do it. The issue with these formatters is they expect a file or stdin, and that's it. (I also swapped to gofmt, which still has the same issue, btw)

It does make it seem like simply doing a path search for the binary, rather than executing it, might be better. Is that an option here? Otherwise you can write custom code to do this, but that's not compatible with your current JSON config file approach.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
richardpark-msft: Sounds good. See another comment I have above - you're going to run into some more issues with the other formatting tool we use (gofmt) if we assume they have some no-op switch.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/SetupRequirements.cs
```diff
@@ -18,5 +18,8 @@ public class Requirement
         
         [JsonPropertyName("instructions")]
         public List<string> instructions { get; set; }
+
+        [JsonPropertyName("is_ci_only")]
+        public bool? isCiOnly { get; set; }
```
weshaggard: What makes a check CI only?

JennyPng: It's a requirement that the developer doesn't need to install locally, but it's used in CI. I recall we wanted to use azsdk CLI in the CI in the future, so this keeps CI requirements in the tool while not forcing a developer to install it

weshaggard: Shouldn't we at least validate that these tools exist locally? If a tool is needed it would need to exist locally or in CI.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
weshaggard: @richardpark-msft do we have a tool in the cli that currently needs go-junit-report? If not we probably should remove it from the list. If we do we probably still need to ensure this tool exists if it is run locally.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/SetupRequirements.cs
praveenkuttappan: +1 this. There is no point adding a requirement only for CI which is something to be done in CI setup and not by verify setup tool or CLI command.

JennyPng: Got it, I'll remove the CI requirements

weshaggard (approved):

PR #12945, opened 2025-11-17
Spec Proposal: Add Pagination Support to Tool Request/Response

chidozieononiwu:
- Use the latest CSharp Parser Changes


src/dotnet/APIView/APIViewWeb/APIViewWeb.csproj
```diff
@@ -77,6 +77,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\..\tools\apiview\parsers\csharp-api-parser\CSharpAPIParser\CSharpAPIParser.csproj" />
```
AlitzelMendez: Why do we want APIView to depend on CSharp parser? The dependency should flow the other way.

I think we should take a different approach. If you want to reference the value in a single place, add the value to https://github.com/Azure/azure-sdk-tools/blob/main/src/dotnet/APIView/APIViewWeb/Helpers/ApiViewConstants.cs with a name like `CSharpParserVersion`, and use that constant in both places.

This avoids adding the unnecessary project reference while still maintaining a single source of truth. Note that CSharpAPIParser already references APIView, so it can access ApiViewConstants without creating any new dependencies.

PR #12948, opened 2025-11-17
[test proxy] path normalization for multipart/form-data Content-Disposition headers

m-redding:
This PR adds cross-platform path normalization for multipart/form-data Content-Disposition headers. Some libraries set filepaths as filenames, which is okay according to the HTTP spec. This causes issues in the test proxy which does a byte comparison. It will only succeed if the test was recorded and played back on the same OS. This will normalize paths when the test proxy goes to sanitize bodies, converting windows paths to linux paths.


tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordedTestSanitizer.cs
```diff
@@ -171,6 +171,10 @@ public byte[] SanitizeMultipartBody(string boundary, byte[] raw)
                                 newValue = new StringValues(newBody.Length.ToString());
                             }
                         }
+                        if (h.Key == "Content-Disposition")
```
m-redding: I ended up adding the normalization here because:
- it applies to values from both recordings and incoming requests
- code is already looping through the header values as strings and looking at the values
- I looked at implementing at a few different places, but having to take bytes, pass around the encoding, decode, normalize, re-encode, etc added a lot of complexity when there was already a spot I could do it easily

The downside is that this runs with each sanitizer, which there are over 130 sanitizers that need to be applied, which means that this runs over 130 times for each recording and incoming request, but I think this is comparatively trivial since we're already looping through the headers in the body 130 times, this is just adding some normalization to one kind of header in that loop

scbedd: Only place it could be.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/MultipartUtilities.cs
```diff
@@ -431,6 +433,133 @@ private static void RemoveCurrentBoundary(
...
+            }
+
+            var spanValue = value.AsSpan();
+            var semicolonIndex = spanValue.IndexOf(';');
+            if (semicolonIndex == -1) // no header parameters
+            {
+                return value;
+            }
+
+            var outputValue = new StringBuilder(value.Length + 5);
```
scbedd: Why value.Length + 5?

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/MultipartUtilities.cs
```diff
@@ -431,6 +433,133 @@ private static void RemoveCurrentBoundary(
...
+                remaining = remaining[nonWhitespace..];
+
+                // get param slice: "name=value" and param name and value
+                var nextSemicolon = remaining.IndexOf(';');
+                var param = nextSemicolon == -1 ? remaining : remaining.Slice(0, nextSemicolon + 1);
+                var equalIndex = param.IndexOf('=');
+                var paramName = equalIndex == -1 ? param : param.Slice(0, equalIndex);
+                var paramValue = equalIndex == -1 ? ReadOnlySpan<char>.Empty : param.Slice(equalIndex + 1);
+
+                if (paramName.SequenceEqual("filename".AsSpan()))
```
scbedd, line 481: I do not like having specific code for handling `filename` vs `filename*`.

I can understand normalizing paths generally within a `Content-Disposition` header, but having the specific code handling for the specific param seems like a bad pattern. It's very specific to this specific param, instead of being a bit more general.

Is there anything stopping us from normalizing backslashes in the Content-Disposition header generally?

scbedd: I can see exactly why you implemented this way, and I'm not convinced that this is the right solution. It feels extremely tightly bound to an azure-specific problem.

m-redding: > azure-specific problem

This is for the unbranded OpenAI library not specific to Azure

> Is there anything stopping us from normalizing backslashes in the Content-Disposition header generally?

The filename* definition is from [RFC 5987](https://www.rfc-editor.org/rfc/rfc5987), the handling of it needs to be different from the filename header parameter to some degree because I need to decode the value using the charset defined in the first part of the header value, then normalize, then re-encode using that same value. OpenAI only uses UTF-8 but I don't think we should assume UTF-8 encoding.

I can add a helper to at least combine the part after decoding, and before re-encoding, but there's going to have to be some specific logic to handle the encoding in the filename* extended parameter vs no encoding in the filename parameter.

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/MultipartUtilities.cs
m-redding: Ah I had a comment but must have deleted it on accident, it's just to give a small buffer although this method shouldn't change the length of the header value so I can just make this value.Length

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/MultipartUtilities.cs
m-redding: > Is there anything stopping us from normalizing backslashes in the Content-Disposition header generally?

I guess to directly answer this it's the filename* parameter value being encoded is what's stopping us

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Common/RecordedTestSanitizer.cs
```diff
@@ -163,13 +163,9 @@ public byte[] SanitizeMultipartBody(string boundary, byte[] raw)
                     foreach (var h in section.Headers)
                     {
                         var newValue = h.Value;
-                        if (h.Key == "Content-Length")
+                        if (h.Key == "Content-Disposition")
```
m-redding: Oops just realized I deleted the wrong if - need to fix this

m-redding: This file should be unchanged

scbedd: I will sign off after the reversion of this change 👍

tools/test-proxy/Azure.Sdk.Tools.TestProxy/Sanitizers/ContentDispositionFilePathSanitizer.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Text;
+using Azure.Sdk.Tools.TestProxy.Common;
+using Azure.Sdk.Tools.TestProxy.Common.Exceptions;
+using Microsoft.AspNetCore.WebUtilities;
+
+namespace Azure.Sdk.Tools.TestProxy.Sanitizers
+{
+    public class ContentDispositionFilePathSanitizer : RecordedTestSanitizer
+    {
+        public ContentDispositionFilePathSanitizer()
```
scbedd, line 13: I _believe_ my mapper should handle this. You're going to actually need to run a registration to be certain though.

m-redding: I tested this with the OpenAI repo and it works

PR #12949, opened 2025-11-17
Fixing a bug where "go get" can auto-update the go version directive, which can break tooling

richardpark-msft:
Don't auto-upgrade the Go version directive for older modules. We're in a transition state, and that would break areas that aren't ready yet.

Also, some small fixes:
- Use gofmt, instead of goimports. gofmt is included with the Go compiler. In the future it'd be nice to allow them to choose, but this is the standard setup on CI.
- Rename all the constants to make it more obvious which one is Unix, and which is Windows.
- Fix some checks so their cross-platform-ness is explicit, even though they would have worked fine since the .exe suffix is implicit on Windows.


samvaity (approved):
lgtm!

PR #12984, opened 2025-11-18
Add PackageInfo SdkType for Java

### LINKED ISSUE #12836: Package Info is missing SDK type for Java

https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaLanguageService.cs#L37

Currently SDK type is set in CI and Engsys tools using following code, and we should use the same logic to set the properties in MCP package info so values match.

https://github.com/Azure/azure-sdk-for-java/blob/main/eng/scripts/Language-Settings.ps1#L163

samvaity:
Closes #12836

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/SdkType.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/SdkType.cs
@@ -9,5 +9,7 @@ public enum SdkType
     [JsonPropertyName("mgmt")]
     Management,
     [JsonPropertyName("client")]
-    Dataplane
+    Dataplane,
+    [JsonPropertyName("spring")]
+    Spring
 }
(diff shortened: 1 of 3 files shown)


praveenkuttappan (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/SdkType.cs
```diff
@@ -9,5 +9,7 @@ public enum SdkType
     [JsonPropertyName("mgmt")]
     Management,
     [JsonPropertyName("client")]
-    Dataplane
+    Dataplane,
+    [JsonPropertyName("spring")]
+    Spring
```
timovv, line 14: What's this SDK type? Is it Java-specific? Could we add a quick docstring here to say that if so?

timovv (approved):

raych1 (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/SdkType.cs
samvaity: yeah, its java specific

Issue #12981, opened 2025-11-18
[js-api-parser] Refactor JS API view Parser - Phase 0


KarishmaGhiya:
Refactoring the JS API view Parser - Phase 0:
## Phase 0: Foundation ✅
- [x] Create `TokenGenerator` interface
- [x] Create generator registry (`src/tokenGenerators/index.ts`)
- [x] Integrate into `buildMemberLineTokens` with fallback
- [x] Implement Function generator
- [x] Implement Enum generator

KarishmaGhiya:
Completed with #12986 and #13053

Issue #12982, opened 2025-11-18
[js-api-parser] Refactor how we track deprecated APIView items


KarishmaGhiya:
Today we track item deprecation in each generator with a separate parameter that we pass to the generator. In an ideal world the token generators take a single parameter: the `ApiItem` (or subclass of) and that item has all the information we need to decide whether it's deprecated or not.

For example, for ApiEnum, we can do: `const deprecated = item.tsdocComment?.deprecatedBlock !== undefined;`

This issue tracks investigating whether that can work for us and if so, removing the `deprecated` argument to the `generate` function, replacing it with the logic above.

Also putting the discussion thread for context for more information we'll need in future https://github.com/Azure/azure-sdk-tools/pull/12986#discussion_r2547062791

PR #12986, opened 2025-11-19
[js-api-parser] Refactor Js parser: phase0

KarishmaGhiya:
This PR does the following tasks from issue #12981

Create TokenGenerator interface
Create generator registry (src/tokenGenerators/index.ts)
Integrate into buildMemberLineTokens with fallback
Implement Enum generator


tools/apiview/parsers/js-api-parser/src/generate.ts
```diff
@@ -310,6 +311,14 @@ function mayHaveChildren(item: ApiItem): boolean {
  * @param deprecated Whether the Api is deprecated or not
  */
 function buildMemberLineTokens(line: ReviewLine, item: ApiItem, deprecated: boolean) {
+   for (const generator of generators) {
+    if (generator.isValid(item)) {
+      line.Tokens.push(...generator.generate(item, deprecated));
+      console.log(`Generated tokens for ${item.canonicalReference.toString()} using custom generator: ${generator.constructor.name}`);
+      console.log(generator.generate(item, deprecated));
```
maorleger: If logging the generated tokens helps us maybe we can just call it once like:

```ts
const tokens = generator.generate(item, deprecated)
console.log(`Generated tokens for ${item.canonicalReference.toString()} using custom generator: ${generator.constructor.name}`);
console.log(tokens)
line.tokens.push(...tokens)
return
```

The function call is probably super cheap, so its not a big deal, but unnecessary to call it twice

tools/apiview/parsers/js-api-parser/src/tokenGenerators/enum.ts
```diff
@@ -0,0 +1,39 @@
+import { ApiItem, ApiItemKind } from "@microsoft/api-extractor-model";
+import { ReviewToken, TokenKind } from "../models";
+import { ITokenGenerator } from "./interfaces";
+
+function isValid(item: ApiItem): boolean {
```
maorleger: Maybe we can use a typeguard instead of a boolean, then the `generate` function can take the subclass

What I mean is:

```ts
function isValid(item: ApiItem): item is ApiEnum {
  return item.kind === ApiItemKind.Enum;
}

function generate(item: ApiEnum): ReviewToken[] { // notice that this can take ApiEnum now
```

maorleger: I also wonder about the name `isValid`, maybe `applicableTo` or `canGenerate` or something like that? 

No strong feelings here, it's easy to rename later if we decide on a better name

tools/apiview/parsers/js-api-parser/src/tokenGenerators/enum.ts
```diff
@@ -0,0 +1,39 @@
...
+    const enumMembers = item.members;
+    for (const member of enumMembers) {
+        tokens.push({Kind: TokenKind.MemberName, Value: member.displayName, IsDeprecated: deprecated});
+        tokens.push({Kind: TokenKind.Punctuation, Value: ",", HasPrefixSpace: true, HasSuffixSpace: true, IsDeprecated: deprecated});
+    }
+    // Remove last comma
+    if (enumMembers.length > 0) {
+        tokens.pop();
+    }
+    tokens.push({Kind: TokenKind.Punctuation, Value: "}", HasPrefixSpace: true, HasSuffixSpace: true, IsDeprecated: deprecated});
```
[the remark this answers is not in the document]
maorleger: oooh nice catch if true 😄

tools/apiview/parsers/js-api-parser/src/tokenGenerators/index.ts
```diff
@@ -0,0 +1,6 @@
+import { enumTokenGenerator } from "./enum";
+import { ITokenGenerator } from "./interfaces";
+
+export const generators: ITokenGenerator[] = [
```
maorleger: nit: in typescript we dont usually include the `I` prefix, so the interface would just be `TokenGenerator`

(looks like you may have been writing c# code lately? 😸 )

tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
```diff
@@ -0,0 +1,22 @@
...
+     */
+    isValid(item: ApiItem): boolean;
+
+    /**
+     * Generates ReviewTokens from the given ApiItem.
+     * @param item - The ApiItem to process.
+     * @param deprecated - Indicates if the Api is deprecated.
+     * @returns An array of ReviewTokens generated from the ApiItem.
+     */
+    generate(item: ApiItem, deprecated: boolean): ReviewToken[];
```
maorleger: I bet we can use the ApiItem model to check if its deprecated instead of passing this boolean around, I would suggest just checking out the interface or chatting with copilot about what's available to us. Maybe it's a jsdoc annotation or something that is passed through to the model json?

maorleger: Try this:

```ts
const isDeprecated = item.tsdocComment?.deprecatedBlock !== undefined;
```

tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
```diff
@@ -0,0 +1,22 @@
+import { ApiItem } from '@microsoft/api-extractor-model';
+import { ReviewToken } from '../models';
+
+/**
+ * Interface for token generators that create ReviewTokens from ApiItems.
+ */
+export interface ITokenGenerator {
```
jeremymeng: Is it possible to get some compile time help by using generics?
```suggestion
export interface ITokenGenerator<T = ApiItem> {
```
Each generator then can be more specific,

tools/apiview/parsers/js-api-parser/src/tokenGenerators/index.ts
KarishmaGhiya: Oooooopsss

tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
jeremymeng: It's possible that only a parent is marked as `@deprecated` but we still want to access that info in descendants.

maorleger: huh interesting - do you have any examples I can look at? I don't doubt it's a real case I just want to make sure I can recognize it when I see it 👍 

I guess another way we can tackle this is maybe a helper function like:

```
// iterative pseudocode, recursive version is fine too
function isDeprecated(node):
    current = node
    while current != undefined:
        if current.docComment.deprecatedBlock == true:
            return true
        current = current.parent  // or .parentApiItem
    return false
```

So like, walk up the tree and return true if any of its parent nodes are deprecated

But wouldn't we want to only show `@deprecated` on the line item itself? Like if its a method and the class itself is deprecated I probably dont need to see it marked as deprecated, only the class itself.

But totally open to discussion - let me know what you had in mind!

My goal really is to have each of those generators take a single argument (the ApiItem or subclasses of) and infer everything from the ApiItem model itself. We can include `options` if we must but I'd like to avoid it as much as possible

jeremymeng: I forgot exactly what leads to the current implementation. It likely has to do with the limitation of the output from api-extractor: sometimes it doesn't have a well-formed AST node for an API and we have to fall back to use excerpt tokens text. If that's the only reason we don't really need recursion.

```json
          "kind": "Class",
          "canonicalReference": "@azure/identity!UsernamePasswordCredential:class",
          "docComment": "/**\n * Enables authentication to Microsoft Entra ID with a user's username and password. This credential requires a high degree of trust so you should only use it when other, more secure credential types can't be used.\n *\n * @deprecated\n *\n * UsernamePasswordCredential is deprecated. Use a more secure credential. See https://aka.ms/azsdk/identity/mfa for details.\n */\n",
          "excerptTokens": [
            {
              "kind": "Content",
              "text": "export declare class UsernamePasswordCredential implements "
            },
            {
              "kind": "Reference",
              "text": "TokenCredential",
              "canonicalReference": "@azure/core-auth!TokenCredential:interface"
            },
            {
              "kind": "Content",
              "text": " "
            }
          ],
```

>But wouldn't we want to only show @deprecated on the line item itself? Like if its a method and the class itself is deprecated I probably dont need to see it marked as deprecated, only the class itself.

Yes, that should be enough too, unless we want to apply a CSS class to the whole range of deprecated item.

tools/apiview/parsers/js-api-parser/src/tokenGenerators/enum.ts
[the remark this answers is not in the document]
KarishmaGhiya: Just checked, it's true. The above won't work for recursive.

tools/apiview/parsers/js-api-parser/src/tokenGenerators/enum.ts
KarishmaGhiya: yeah we can decide on a better name. just that the `valid` term is very intuitive. Let's visit this as I make progress on more tokens.

tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
KarishmaGhiya: How does this look?
```
export interface TokenGenerator<T extends ApiItem = ApiItem> {
  
    isValid(item: ApiItem): item is T;

   
    generate(item: ApiItem, deprecated: boolean): ReviewToken[];
}
```

KarishmaGhiya:
@jeremymeng  @maorleger  I addressed some of the comments. Can you review again? I added tests too.

tools/apiview/parsers/js-api-parser/src/tokenGenerators/enum.ts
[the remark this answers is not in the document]
KarishmaGhiya: For now, I'll take Copilot's suggestion, and then we can discuss how we want to refine `buildMember` further.

tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
jeremymeng: could the parameter in `generate` be T as well?

KarishmaGhiya: `generate(item: T, deprecated: boolean): ReviewToken[]` This?

tools/apiview/parsers/js-api-parser/src/tokenGenerators/enum.ts
maorleger: sounds good to me 👍

tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
KarishmaGhiya: Will revisit this discussion when I work on this issue https://github.com/Azure/azure-sdk-tools/issues/12982

tools/apiview/parsers/js-api-parser/src/generate.ts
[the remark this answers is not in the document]
KarishmaGhiya: Gonna let the console.log statements be there for comparison with old v/s refactored generation. Till we refactor everything and verify it's correctness.

Issue #12889, opened 2025-11-13
Conditional Testing Evals


jeo02:
If we run the evals in another language repository right now it will run all the evals that would be specific to the azure-rest-api-spec repo.
- Pass in the repository name.
- Group Scenarios into different groups per repo + general that should be run everywhere. ex. call creates release plan tool should be able to run in any repository. But evals that expect a specific tool set from instructions are specific.

PR #12995, opened 2025-11-19
Azsdk cli eval tests pipeline

### LINKED ISSUE #12767: Copilot instruction pipeline trigger for evals
Creating a new pipeline to trigger evals when changes to copilot instructions or eng/common/instruction/azsdk-cli are merged.
Will need to exist for:
- Each language
- azure-rest-api-specs repository

Would require a slightly different pipeline template to clone checkout language repo and checkout tools repo before running eval.

jeo02:
- triggers evaluation on copilot instruction changes.
- Placing it in commons as it should be triggered from any language. For now only the rest-api-specs will have the pipeline until conditional testing is implemented: https://github.com/Azure/azure-sdk-tools/issues/12889
- Resolves #12767

### DIFF
--- a/eng/common/pipelines/ai-evals-tests.yml
+++ b/eng/common/pipelines/ai-evals-tests.yml
@@ -77,14 +77,14 @@ jobs:
         workingDirectory: '${{ parameters.ToolsRepoName }}/${{ parameters.EvalProject }}'
         inlineScript: |
           echo "Logged in to Azure"
-          echo "Running eval in project ${{ parameters.EvalProject }}"
+          echo "Running eval in project ${{ parameters.ToolsRepoName }}/${{ parameters.EvalProject }}"
           dotnet test /p:ArtifactsPackagesDir=$(Build.ArtifactStagingDirectory) --logger trx
       env:
         DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
         DOTNET_CLI_TELEMETRY_OPTOUT: 1
         DOTNET_MULTILEVEL_LOOKUP: 0
         AZURE_OPENAI_MODEL_DEPLOYMENT_NAME: ${{ parameters.Model }}
-        AZURE_OPENAI_ENDPOINT: $(Build.Repository.Name)/${{ parameters.OpenAIEndPoint }}
+        AZURE_OPENAI_ENDPOINT: ${{ parameters.OpenAIEndPoint }}
         REPOSITORY_NAME: $(Build.Repository.Name)
         COPILOT_INSTRUCTIONS_PATH_MCP_EVALS: $(System.DefaultWorkingDirectory)/$(Build.Repository.Name)/.github/copilot-instructions.md
 


jeo02:
[Pipeline](https://dev.azure.com/azure-sdk/internal/internal%20Team/_build?definitionId=7985&_a=summary)

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+    - checkout: self
+    - task: UseDotNet@2
+      displayName: "Use .NET SDK 9.0.x"
+      retryCountOnTaskFailure: 3
+      inputs:
+        packageType: sdk
+        version: 9.0.x
+        performMultiLevelLookup: true
+
+    - task: UseDotNet@2
```
praveenkuttappan, line 49: Do we need this?

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+      - template: /eng/pipelines/templates/variables/globals.yml
+      - template: /eng/pipelines/templates/variables/image.yml
+    displayName: 'Run AI Eval'
+    pool:
+      name: $(LINUXPOOL)
+      image: $(LINUXVMIMAGE)
+      os: linux
+
+    steps:
+    - checkout: self
```
praveenkuttappan: Do we want to checkout everything or just eng/ and .github? All we need is copilot instructions and instruction files in eng/common. right?

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+        SkipCheckoutNone: true
+        Repositories:
+          - Name: ${{ parameters.ToolsRepoOwner }}/${{ parameters.ToolsRepoName }}
+            WorkingDirectory: $(System.DefaultWorkingDirectory)/${{ parameters.ToolsRepoName }}
+        Paths:
+          - 'tools/**'
+    
+    - task: AzureCLI@2
+      displayName: 'Run eval'
+      condition: and(succeeded(), ne(variables['Skip.Eval'], 'true'))
```
praveenkuttappan: No need to add skip variable in condition for this pipeline since this is just for eval.

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+      inputs:
+        azureSubscription: opensource-api-connection
```
weshaggard, line 74: Are these credentials needed to connect to the Eval service? This will limit who can run this pipeline because it will need to remain internal and only folks with permission will be able to run it.

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,94 @@
+trigger: none
+
+pr:
+  branches:
+    include:
+      - main
+  paths:
+    include:
+      - .github/copilot-instructions.md
```
weshaggard, line 9: What is the plan for running this in other repos? Should we start by having this only in the tools repo and run whenever the common instructions change? Or maybe only have it in the specs repo if the repo level instructions are changing more heavily there.

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+      - eng/common/instructions/azsdk-tools/**
+
+parameters:
+  - name: EvalProject
+    type: string
+    default: 'tools/azsdk-cli/Azure.Sdk.Tools.Cli.Evaluations'
+  - name: OpenAIEndPoint
+    type: string
+    default: 'https://ai-prmarottai3149546654251245.openai.azure.com/'
```
weshaggard, line 18: Is this the shared openai end-point? Do we need to have a specific instance?

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+      displayName: "Use .NET SDK 8.0.x"
+      retryCountOnTaskFailure: 3
+      inputs:
+        packageType: sdk
+        version: 8.0.x
+        performMultiLevelLookup: true
+
+    - template: /eng/common/pipelines/templates/steps/sparse-checkout.yml
+      parameters:
+        SkipCheckoutNone: true
```
weshaggard, line 59: Why not sparse check out the repo we are running in as well? If all we need is the .github/copilot-instructions.md maybe we just clone that.

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+        Repositories:
+          - Name: ${{ parameters.ToolsRepoOwner }}/${{ parameters.ToolsRepoName }}
+            WorkingDirectory: $(System.DefaultWorkingDirectory)/${{ parameters.ToolsRepoName }}
+        Paths:
+          - 'tools/**'
```
weshaggard, line 67: Don't we need the eng/common directory? Or will that come from the repo we are testing in?

weshaggard: Also is the tools directory enough to run the eval tests? I know there might be some things in the root and eng folder, although those might be getting added to the paths automatically.

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+          echo "Running eval in project ${{ parameters.EvalProject }}"
+          dotnet test /p:ArtifactsPackagesDir=$(Build.ArtifactStagingDirectory) --logger trx
+      env:
+        DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
+        DOTNET_CLI_TELEMETRY_OPTOUT: 1
+        DOTNET_MULTILEVEL_LOOKUP: 0
+        AZURE_OPENAI_MODEL_DEPLOYMENT_NAME: ${{ parameters.Model }}
+        AZURE_OPENAI_ENDPOINT: ${{ parameters.OpenAIEndPoint }}
+        REPOSITORY_NAME: $(Build.Repository.Name)
+        COPILOT_INSTRUCTIONS_PATH_MCP_EVALS: $(System.DefaultWorkingDirectory)/.github/copilot-instructions.md
```
weshaggard: Why is this an env variable? Is the test code using this to find the instructions?

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+      displayName: 'Run eval'
+      condition: and(succeeded(), ne(variables['Skip.Eval'], 'true'))
+      inputs:
+        azureSubscription: opensource-api-connection
+        scriptType: 'bash'
+        scriptLocation: 'inlineScript'
+        workingDirectory: '${{ parameters.ToolsRepoName }}/${{ parameters.EvalProject }}'
+        inlineScript: |
+          echo "Logged in to Azure"
+          echo "Running eval in project ${{ parameters.EvalProject }}"
```
weshaggard: Why is the ArtifactsPackagesDir needed?

eng/common/pipelines/ai-evals-tests.yml
jeo02: yes, the test code is using this to get the Copilot instructions

eng/common/pipelines/ai-evals-tests.yml
jeo02: Not sure @praveenkuttappan ?

eng/common/pipelines/ai-evals-tests.yml
jeo02: Well i guess the thought being it will be triggered by changes rather than manual runs. If someone is to run it manually should be internal.

eng/common/pipelines/ai-evals-tests.yml
jeo02: The eng/common should be in the repository we are in. 
We are checking out the tools repo only to get the evals project as everything else we need is in the repository we are in.

eng/common/pipelines/ai-evals-tests.yml
jeo02: yea we can also do that and then just remove the checkout: self. I think we would also need the eng/common/instructions part as well

eng/common/pipelines/ai-evals-tests.yml
jeo02: I assume that its because I have a dependency on the Azure.Sdk.Tools.Cli which has some dependency on another project that uses dotnet 8

eng/common/pipelines/ai-evals-tests.yml
jeo02: yes I'll sparse checkout it

eng/common/pipelines/ai-evals-tests.yml
jeo02: I needed the openai endpoint so I think it will always point to the specific instance not sure how you can make it more general?

eng/common/pipelines/ai-evals-tests.yml
jeo02: So, the pipeline I have in the comments is for the specs repository, so I plan on starting there. Overall, every language would be able to run their specific test cases against their own copilot instructions.

Also have general tests to trigger x tool that should work across all repositories that should be checked.

eng/common/pipelines/ai-evals-tests.yml
weshaggard: @praveenkuttappan this looks like your instance, should we use the shared one?

praveenkuttappan: yes. It seems evryone is using the instance I created because that has gpt5 model. But we should switch to a shared instance if we have one.

weshaggard: I don't know what that shared instance is @benbp might know but I know we have limited number of instances we can have in our subscription, so it is best to share if we don't need anything specific.

eng/common/pipelines/ai-evals-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+      parameters:
+        SkipCheckoutNone: true
+        Repositories:
+          - Name: $(Build.Repository.Name)
+            Commitish: $(Build.SourceVersion)
+            WorkingDirectory: $(System.DefaultWorkingDirectory)/$(Build.Repository.Name)
+          - Name: ${{ parameters.ToolsRepoOwner }}/${{ parameters.ToolsRepoName }}
+            WorkingDirectory: $(System.DefaultWorkingDirectory)/${{ parameters.ToolsRepoName }}
+        Paths:
+          - 'tools/**'
```
weshaggard, line 67: Any reason this path is quoted and the others aren't. I would make them consistent one way or the other I don't think the quotes are needed but also don't hurt.

weshaggard (approved):
I'm OK with the general idea but I do worry about the fact this pipeline requires secrets and as such will not be able to auto-trigger on all PRs. For example, it will not be able to trigger on copilot generated PRs someone will need to trigger it manually via a comment.

eng/common/pipelines/ai-evals-tests.yml
weshaggard: Ok let's be sure to name the pipelines consistently for each language using the standard convention `<lang> - <name>`.

eng/common/pipelines/ai-evals-tests.yml
jeo02: Honestly didn't catch that I can remove it

jeo02:
Made an issue to address all of the fixes here. 
https://github.com/Azure/azure-sdk-tools/issues/13049

weshaggard:
@jeo02 please do not override and merge this until all the sync PRs are merged.

PR #12998, opened 2025-11-19
Change default value of SkipReleaseStage to false

weshaggard:
Change the default back to what it was before https://github.com/Azure/azure-sdk-tools/pull/11287 as it currently blocks publishing all the existing tools. The cli should still be protected and only deployable via main builds.


weshaggard:
FYI @benbp @scbedd

PR #13009, opened 2025-11-20
Used another way to create package info array

raych1:
Fixed a [JSON parse error](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5592577&view=logs&j=fbf07650-9b13-5e87-a054-01a63335f5b0&t=1c726fd2-74dc-51d2-977c-c4e89f5457ee):
```
Formatted command: . 'D:\a\_work\1\azure-sdk-for-java\eng\common\scripts\Create-APIReview.ps1' -PackageInfoFiles ('[
  "D:\a\_work\1/packages-signed/PackageInfo/azure-sdk-template.json",
  "D:\a\_work\1/packages-signed/PackageInfo/azure-sdk-template-two.json"
]' | ConvertFrom-Json -NoEnumerate) -ArtifactList ('[]' | ConvertFrom-Json | Select-Object Name) -ArtifactPath 'D:\a\_work\1/packages-signed' -ArtifactName packages-signed -APIKey '***' -PackageName '' -SourceBranch 'update-pipeline-script-for-v2' -DefaultBranch 'main' -ConfigFileDir 'D:\a\_work\1/packages-signed/PackageInfo' -BuildId '5592577' -RepoName 'Azure/azure-sdk-for-java' -MarkPackageAsShipped $True
========================== Starting Command Output ===========================
"C:\Program Files\PowerShell\7\pwsh.exe" -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". 'D:\a\_work\_temp\a4d2f187-5ffe-47e8-8251-3646f08da395.ps1'"
ConvertFrom-Json : Conversion from JSON failed with error: Bad JSON escape sequence: \a. Path '', line 2, position 7
```

**Test run:**
[template-batch-release](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5592983&view=results)

### DIFF
--- a/eng/common/scripts/Create-APIReview.ps1
+++ b/eng/common/scripts/Create-APIReview.ps1
@@ -373,7 +373,8 @@ elseif ($ArtifactList -and $ArtifactList.Count -gt 0) {
 elseif ($PackageInfoFiles -and $PackageInfoFiles.Count -gt 0) {
     # Lowest Priority: Direct PackageInfoFiles (new method)
     Write-Host "Using PackageInfoFiles parameter with $($PackageInfoFiles.Count) files"
-    $ProcessedPackageInfoFiles = $PackageInfoFiles  # Use as-is
+    # Filter out empty strings or whitespace-only entries
+    $ProcessedPackageInfoFiles = $PackageInfoFiles | Where-Object { -not [string]::IsNullOrWhiteSpace($_) }
 }
 else {
     Write-Error "No package information provided. Please provide either 'PackageName', 'ArtifactList', or 'PackageInfoFiles' parameters."
@@ -382,7 +383,7 @@ else {
 
 # Validate that we have package info files to process
 if (-not $ProcessedPackageInfoFiles -or $ProcessedPackageInfoFiles.Count -eq 0) {
-    Write-Error "No package info files found after processing parameters."
+    Write-Error "No package info files found after processing parameters. Or PackageInfoFiles parameter contains only empty or whitespace entries, please check the artifact settings."
     exit 1
 }
 
--- a/eng/common/scripts/Validate-All-Packages.ps1
+++ b/eng/common/scripts/Validate-All-Packages.ps1
@@ -282,12 +282,13 @@ elseif ($PackageInfoFiles -and $PackageInfoFiles.Count -gt 0)
 {
     # Direct PackageInfoFiles (new method)
     Write-Host "Using PackageInfoFiles parameter with $($PackageInfoFiles.Count) files"
-    $ProcessedPackageInfoFiles = $PackageInfoFiles
+    # Filter out empty strings or whitespace-only entries
+    $ProcessedPackageInfoFiles = $PackageInfoFiles | Where-Object { -not [string]::IsNullOrWhiteSpace($_) }
 }
 
 # Validate that we have package info files to process
 if (-not $ProcessedPackageInfoFiles -or $ProcessedPackageInfoFiles.Count -eq 0) {
-    Write-Error "No package info files found after processing parameters."
+    Write-Error "No package info files found after processing parameters. Or PackageInfoFiles parameter contains only empty or whitespace entries, please check the artifact settings."
     exit 1
 }
 


eng/common/pipelines/templates/steps/validate-all-packages.yml
```diff
@@ -34,7 +34,7 @@ steps:
             -BuildDefinition $(System.CollectionUri)$(System.TeamProject)/_build?definitionId=$(System.DefinitionId) `
             -PipelineUrl $(System.CollectionUri)$(System.TeamProject)/_build/results?buildId=$(Build.BuildId) `
             -IsReleaseBuild $$(SetAsReleaseBuild) `
-            -PackageInfoFiles ('${{ convertToJson(parameters.PackageInfoFiles) }}' | ConvertFrom-Json -NoEnumerate)
+            -PackageInfoFiles @('${{ join(''',''', parameters.PackageInfoFiles) }}')
```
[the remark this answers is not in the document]
raych1: Good reminder. However, the suggestion cannot handle the case @(''). I've added the check on the script.

eng/common/pipelines/templates/steps/create-apireview.yml
```diff
@@ -41,7 +41,7 @@ steps:
       inputs:
         filePath: ${{ parameters.SourceRootPath }}/eng/common/scripts/Create-APIReview.ps1
         arguments: >
-          -PackageInfoFiles ('${{ convertToJson(parameters.PackageInfoFiles) }}' | ConvertFrom-Json -NoEnumerate)
+          -PackageInfoFiles @('${{ join(''',''', parameters.PackageInfoFiles) }}')
```
weshaggard, line 44: This assumes that this is a list is that the expectation?

eng/common/scripts/Create-APIReview.ps1
```diff
@@ -373,7 +373,8 @@ elseif ($ArtifactList -and $ArtifactList.Count -gt 0) {
 elseif ($PackageInfoFiles -and $PackageInfoFiles.Count -gt 0) {
     # Lowest Priority: Direct PackageInfoFiles (new method)
     Write-Host "Using PackageInfoFiles parameter with $($PackageInfoFiles.Count) files"
-    $ProcessedPackageInfoFiles = $PackageInfoFiles  # Use as-is
+    # Filter out empty strings or whitespace-only entries
```
weshaggard, line 376: Do we end up with empty entries?

eng/common/pipelines/templates/steps/create-apireview.yml
weshaggard, line 44: Based on https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#join wouldn't this just be `join(',', parameters.PackageInfoFiles)`

weshaggard: I assume this is some escaping but do these all need to be single quotes? Do we need to have the outer quotes?

raych1: Yes, `PackageInfoFiles` is a list.

eng/common/scripts/Create-APIReview.ps1
raych1: Theoretically, the `artifacts` or `release artifact` won't be empty. However, it's defensible to have this check.

This is the test run without setting `artifacts` input:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5595306&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=526cb169-4442-5378-373a-5a950554349c

eng/common/pipelines/templates/steps/create-apireview.yml
raych1: Yes, expect the output to the powershell to be something like :  @('file1.json','file2.json','file3.json')

eng/common/scripts/Create-APIReview.ps1
weshaggard: I would assume if there are no artifacts the list would be empty but not an individual entry.

eng/common/pipelines/templates/steps/create-apireview.yml
weshaggard: Why is are the quotes around the outside needed? `@(' ... join ')`

raych1: The outside single quotation marks are used as the beginning quotation mark for the first array element and the ending quotation mark for the last array element.
Let me give an example, say `PackageInfoFiles` equals [`a/file1.json`, `a/file2.json`].
The join operation will make the value (referred as `inner value`) of ${{...}} to be `a/file1.json','a/file2.json`.
Then @('`inner value`') becomes `@('a/file1.json','a/file2.json')`

eng/common/scripts/Create-APIReview.ps1
raych1: From the yml, the empty `PackageInfoFiles` will be like `@('')`, which has an empty string element instead of an empty array. See [the comment](https://github.com/Azure/azure-sdk-tools/pull/13009#discussion_r2548127008) for why the yml passes in an array with an empty string element.

eng/common/pipelines/templates/steps/create-apireview.yml
weshaggard: Ok I think understand what is happening now. While it is not optimal, I cannot come up with a better way to do it with the current expression functions we have available.

weshaggard (approved):

PR #13029, opened 2025-11-21
Fix release_sdk tool failure due to package name mismatch

### LINKED ISSUE #12600: AzSDK MCP server: release_sdk tool failure due to package name mismatch
I see few failures in telemetry for `azsdk_release_sdk` tool when user provides a package name that does not have an exact match. Currently we fail these requests with message "package does not exists" and this might be confusing. For e.g. user was trying to release package `armmongocluster` but exact package name is `sdk/resourcemanager/mongocluster/armmongocluster`

solution:
If an exact package is not found with the given input, then we should attempt to find all packages with name that contains the given input and respond with a message. 

`The package <package name> could not be found. Did you mean one of the following available packages?
[<List of available package names with given word in the name>]
Next step: Select the package and run SDK release tool`

smw-ms:
Issue: https://github.com/Azure/azure-sdk-tools/issues/12600


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs
```diff
@@ -73,7 +73,41 @@ public async Task<SdkReleaseResponse> ReleasePackageAsync(string packageName, st
...
                 // Get the package work item from DevOps
-                var package = await devopsService.GetPackageWorkItemAsync(packageName, language);
+                PackageWorkitemResponse? package = null;
+                try
+                {
+                    package = await devopsService.GetPackageWorkItemAsync(packageName, language);
+                } 
+                catch (Exception ex)
+                {
+                    if (ex.Message.Contains("Failed to find package work item with package name", StringComparison.OrdinalIgnoreCase))
```
weshaggard: Is there any other exception type or error code we can use beyond the string text? Depending on hard-coded string text like this tends to break overtime if someone tweaks the string.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
weshaggard, line 1049: Why not do the partial package query directly here instead of throwing this exception? If we don't want to do that in all cases just add that as a parameter. That will prevent the need of throwing and catching exceptions and handling that.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs
```diff
@@ -73,7 +73,41 @@ public async Task<SdkReleaseResponse> ReleasePackageAsync(string packageName, st
...
+                    {
+                        logger.LogInformation(ex, "Exact package not found; falling back to partial matches");
+                        var packages = await devopsService.ListPartialPackageWorkItemAsync(packageName, language);
```
praveenkuttappan: We should get the list of matching packages if exact check did not return a work item irrespective of exception reasons. So get the list if package  == null

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs
```diff
@@ -73,7 +73,41 @@ public async Task<SdkReleaseResponse> ReleasePackageAsync(string packageName, st
...
+                        if (packages == null || packages.Count == 0)
+                        {
+                            return new SdkReleaseResponse
+                            {
+                                ReleasePipelineStatus = "Failed",
```
praveenkuttappan: Set language property in response.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Package/SdkReleaseTool.cs
```diff
@@ -73,7 +73,41 @@ public async Task<SdkReleaseResponse> ReleasePackageAsync(string packageName, st
...
+                                ResponseError = $"No package work item found for package '{packageName}' in language '{language}'. Please check the package name and language and also make sure that SDK is merged to main branch in the specific language repo.",
+                            };
+
+                        }
+                        
+                        var packageNames = packages.Where(p => p != null && !string.IsNullOrEmpty(p.PackageName))
```
praveenkuttappan: `!string.IsNullOrEmpty(p.PackageName)` Is it going to be true in any case? Query fetches the work items that contains the input in package name.

praveenkuttappan (approved):

Issue #12506, opened 2025-10-15
Update APIReview scripts to use full name for Java libraries


raych1:
The full name of Java libraries will be `groupId:artifactId`.

Here're the relevant scripts:
https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Create-APIReview.ps1
https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Detect-Api-Changes.ps1

PR #13040, opened 2025-11-22
Include groupId in the pkg name when create api view request

raych1:
Fixed https://github.com/Azure/azure-sdk-tools/issues/12506

[Test Run](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5601715&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=658c70a5-b0a1-5272-a37a-3405847182a4)

[New Test Run](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5606853&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3)

### DIFF
--- a/eng/common/scripts/Detect-Api-Changes.ps1
+++ b/eng/common/scripts/Detect-Api-Changes.ps1
@@ -22,12 +22,15 @@ Param (
 $configFileDir = Join-Path -Path $ArtifactPath "PackageInfo"
 
 # Submit API review request and return status whether current revision is approved or pending or failed to create review
-function Submit-Request($filePath, $packageName, $packageType, $group)
+function Submit-Request($filePath, $packageInfo)
 {
+    $packageName = $packageInfo.ArtifactName ?? $packageInfo.Name
+    $packageType = $packageInfo.SdkType
+    
     # Construct full package name with groupId if available
     $fullPackageName = $packageName
-    if ($group) {
-        $fullPackageName = "${group}:$packageName"
+    if ($packageInfo.PSObject.Members.Name -contains "Group" -and $packageInfo.Group) {
+        $fullPackageName = "$($packageInfo.Group):$packageName"
     }
     $repoName = $RepoFullName
     if (!$repoName) {
@@ -170,7 +173,7 @@ foreach ($packageInfoFile in $packageInfoFiles)
         if ($isRequired -eq $True)
         {
             $filePath = $pkgPath.Replace($ArtifactPath , "").Replace("\", "/")
-            $respCode = Submit-Request -filePath $filePath -packageName $pkgArtifactName -packageType $packageType -group $packageInfo.Group
+            $respCode = Submit-Request -filePath $filePath -packageInfo $packageInfo
             if ($respCode -ne '200')
             {
                 $responses[$pkgArtifactName] = $respCode


eng/common/scripts/Detect-Api-Changes.ps1
```diff
@@ -36,7 +41,7 @@ function Submit-Request($filePath, $packageName, $packageType)
     $query.Add('commitSha', $CommitSha)
     $query.Add('repoName', $repoName)
     $query.Add('pullRequestNumber', $PullRequestNumber)
-    $query.Add('packageName', $packageName)
+    $query.Add('packageName', $fullPackageName)
```
raych1, line 47: @praveenkuttappan can we merge this change directly or any change needs to make in API view server side?

CC: @helen229 @AlitzelMendez

eng/common/scripts/Detect-Api-Changes.ps1
```diff
@@ -22,8 +22,13 @@ Param (
 $configFileDir = Join-Path -Path $ArtifactPath "PackageInfo"
 
 # Submit API review request and return status whether current revision is approved or pending or failed to create review
-function Submit-Request($filePath, $packageName, $packageType)
+function Submit-Request($filePath, $packageName, $packageType, $group)
 {
+    # Construct full package name with groupId if available
+    $fullPackageName = $packageName
+    if ($group) {
+        $fullPackageName = "${group}:$packageName"
```
weshaggard: Would it be better to do this on the java side before calling this method instead of adding the group parameter here?

raych1: What do you mean on the java side? `PackageInfo` has separated property: `Group` and `ArtifactName` and we should directly compute based on their value instead of checking on the language.

In addition, not all the places are expected to use the full package name. for example, line #48, only needs the artifact name without `Group`.

weshaggard: Do we need to update that to also include the groupid?

weshaggard: Are we going to make the name property in the packageinfo be the full name now in java?

raych1: It's not planned. At the very least, the "+" delimiter cannot be used as in the URL query part. API view currently uses `:` as delimiter to construct the full package name, here's [the sample code](https://github.com/Azure/azure-sdk-tools/blob/2fb44c5f6443b81ebdbac2fb9910efd958d2ef1c/eng/common/scripts/Validate-All-Packages.ps1#L198).

weshaggard (approved):

eng/common/scripts/Detect-Api-Changes.ps1
```diff
@@ -165,7 +170,7 @@ foreach ($packageInfoFile in $packageInfoFiles)
         if ($isRequired -eq $True)
         {
             $filePath = $pkgPath.Replace($ArtifactPath , "").Replace("\", "/")
-            $respCode = Submit-Request -filePath $filePath -packageName $pkgArtifactName -packageType $packageType
+            $respCode = Submit-Request -filePath $filePath -packageName $pkgArtifactName -packageType $packageType -group $packageInfo.Group
```
weshaggard: not all packageInfo have a Group property so if this script is ran in strict mode this call will fail. We might need to handle that.

raych1: Updated the `Submit-Request` to take the `PackageInfo` obj as input then check the property safely.

weshaggard (approved):

eng/common/scripts/Detect-Api-Changes.ps1
praveenkuttappan: This looks fine to me.

Issue #12595, opened 2025-10-21
[APIView] Cross-Language Feature Design


tjprescott:
A requirement for the archboard revamp process to enable automated reviews and approvals is a way to have a cross-language conversation in APIView.

**This issue is not to track implementation, but to track the need to develop and vet a solution, as this is still very ill-defined.**

We have discussed a couple ways this might manifest, based on the work that has already been done in APIView to enable cross-language IDs.

**Cross-Langauge Conversations Tab**

This would present a unified conversations tab that would include comments (and context snippets) for comments across all languages for a given review, allowing any architect to see the conversations for every other review. In this view, comments could be marked as applicable to one or more languages to identify and resolve cross-language impacts.

This view would allow replies to existing comments, but would not allow the creation of new comments (just like the current conversations tab today)

**Cross-language Review Page**

This would present some common form--likely TypeSpec--where architects could leave feedback and mark that feedback as relevant to one or more languages. This view would show only syntax that has a cross-language ID, and architects would be able to easily see how that concept is rendered in their language.

**Dashboard**

There is some sketch of a cross-language dashboard that would show the "10,000 ft view" for all languages for a given package.

tjprescott:
Okay, as of 11/5 we have found a way to move this forward as a pilot. We already have logic that ties TypeSpec reviews to their associated language-specific SDK reviews. We will leverage that existing logic to demonstrate a proof-of-concept for converting the package-name label to a related-packages switcher.

tjprescott:
After an internal meeting on Nov 10, we agreed to use the TypeSpec spec itself as the organizing principle around cross-language efforts rather than the service tree ID.

- Only services with a TypeSpec spec will be able to participate in the cross-language features (this was already true!)
- We identified the need to spend more time examining and testing the UX for specs which are still in flux (the spec is not yet approved, or the namespaces are not approved).
- APIView will still need to work for "orphan" reviews that have no associated TypeSpec (for example: supported by Swagger only)

PR #13048, opened 2025-11-24
[APIView] Cross Language and Namespace Review Proposal

tjprescott:
Related to: https://github.com/Azure/azure-sdk-tools/issues/12595

For reviewing, it's recommended to use the rich diff view:
image

### DIFF
--- a/src/dotnet/APIView/proposals/parse_tspconfig.py
+++ b/src/dotnet/APIView/proposals/parse_tspconfig.py
@@ -5,6 +5,7 @@ import urllib.request
 import os
 import yaml
 import re
+import requests
 import urllib.parse
 
 
@@ -27,13 +28,50 @@ def _parse_typespec(yaml_url):
     parsed_url = urllib.parse.urlparse(yaml_url)
     path_parts = parsed_url.path.strip('/').split('/')
     typespec_ns = None
-    # TODO: Actually compile the typespec if available to get the namespace, service name and description...
 
-    # Look for pattern: .../<provider>/<namespace>/tspconfig.yaml
-    if len(path_parts) >= 3 and path_parts[-1].lower() == 'tspconfig.yaml':
-        typespec_ns = path_parts[-2]
-    elif path_parts:
-        typespec_ns = path_parts[-1]
+    # Try to extract namespace from main.tsp if available
+    # Download main.tsp from the same directory as tspconfig.yaml
+    if len(path_parts) >= 1:
+        tsp_dir = '/'.join(path_parts[:-1])
+        main_tsp_url = None
+        if 'githubusercontent.com' in yaml_url:
+            # Extract user, repo, and branch from path
+            path_split = parsed_url.path.strip('/').split('/')
+            # Format: /<user>/<repo>/<branch>/.../<namespace>/tspconfig.yaml
+            if len(path_split) >= 4:
+                user = path_split[0]
+                repo = path_split[1]
+                branch = path_split[2]
+                rest = '/'.join(path_split[3:-1])  # skip tspconfig.yaml
+                main_tsp_url = f'https://raw.githubusercontent.com/{user}/{repo}/{branch}/{rest}/main.tsp'
+            else:
+                main_tsp_url = None
+        elif 'github.com' in yaml_url and '/blob/' in yaml_url:
+            # If user passed a blob URL, convert to raw
+            repo_parts = yaml_url.split('github.com/', 1)[-1].split('/blob/', 1)
+            user_repo = repo_parts[0]
+            main_tsp_url = f'https://raw.githubusercontent.com/{user_repo}/{tsp_dir}/main.tsp'
+        elif yaml_url.startswith('http'):
+            # Generic case
+            main_tsp_url = urllib.parse.urljoin(yaml_url, '../main.tsp')
+        # Try to download and parse main.tsp
+        if main_tsp_url:
+            try:
+                resp = requests.get(main_tsp_url, timeout=5)
+                if resp.status_code == 200:
+                    # Look for: namespace <NamespaceName>;
+                    match = re.search(r'namespace\s+([\w\.]+)\s*;', resp.text)
+                    if match:
+                        typespec_ns = match.group(1)
+            except Exception:
+                pass
+
+    # Fallback to URL-based extraction if not found
+    if not typespec_ns:
+        if len(path_parts) >= 3 and path_parts[-1].lower() == 'tspconfig.yaml':
+            typespec_ns = path_parts[-2]
+        elif path_parts:
+            typespec_ns = path_parts[-1]
     if typespec_ns:
         ns_type = 'management' if 'Management' in typespec_ns else 'data-plane'
         return {'typespec': {'namespace': typespec_ns, 'type': ns_type}}


src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+
+This approach resolves several open questions about identity, namespace changes, and review attachment, and provides a robust foundation for future enhancements.
+
+### 2. Integration Points
+
+#### **When just a TypeSpec `tspconfig.yaml` is available**
+
+User can manually create a Product in APIView by providing the `tspconfig.yaml` file or its URI. APIView will:
+
+* Parse the file to extract Product metadata (name, description, namespaces)
```
maririos: I am trying to make it to an example. 
You mean the name comes from here? https://github.com/Azure/azure-rest-api-specs/blob/main/specification/appconfiguration/data-plane/AppConfiguration/main.tsp#L10

src/dotnet/APIView/proposals/ProjectCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+model Product {
+  /** GUID identifier for the Product. */
+  Id: string;
+  /** Human-friendly name of the Product. Otherwise, defaults to the namespace. */
+  DisplayName?: string;
+  Description?: string;
+  /** Service-level contacts. */
+  Owners?: string[];
+  /** Convenience field for the latest approved (or proposed) namespace. */
+  Namespace: string;
```
maririos: which namespace? it depends on each language so it is a list

src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+  /** GUID identifier for the Product. */
+  Id: string;
+  /** Human-friendly name of the Product. Otherwise, defaults to the namespace. */
+  DisplayName?: string;
+  Description?: string;
+  /** Service-level contacts. */
+  Owners?: string[];
+  /** Convenience field for the latest approved (or proposed) namespace. */
+  Namespace: string;
+  NamespaceInfo: ProductNamespaceInfo;
```
maririos: what would go here?

src/dotnet/APIView/proposals/ProjectCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+  NamespaceInfo: ProductNamespaceInfo;
+  /** Associated Review IDs. */
+  ReviewIds: string[];
```
maririos: because we are keeping the Review - Revision structure, right?

src/dotnet/APIView/proposals/ProjectCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+3. If there is no approved namespace for that language, create a new namespace proposal in proposed status, updating the previous proposal:
+  a. Proposed → Withdrawn
+  b. Rejected → Rejected (keep the same)
+
+---
+
+
+## Open Questions
+
+1. Do we care about just namespace or also package name? The current proposal has both fields, so approval, rejection, etc. is implicitly for both. Is that acceptable?
```
maririos: my understanding is that the only language that uses namespace and package name differently is Java

src/dotnet/APIView/proposals/ProjectCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+2. The namespace behavior proposed would mean that a Product could have multiple rejected namespaces, and other Products would not be able to use those namespaces. Is that acceptable?
+
+---
+
+## Next Steps
+
+* Validate with EngSys that tspconfig-derived metadata can be reliably emitted
+* Create migration script for existing Reviews
+* Create new endpoint in APIView for registering new Products manually with `tspconfig.yaml`
+* Implement CrossLanguagePackageId handling in language parsers
```
maririos, line 305: It is no needed in TCGC because that is the work that Izzy did? or there is still work needed there

maririos (commented):
Could you provide couple of examples that would help land the proposal?

src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+
+Including rejected, withdrawn, and superseded proposals.
+
+### ✓ Improve stakeholder experience
+
+Navigation, dashboards, and discussions become cross-language by default.
+
+### ✓ Provide a stable home for TypeSpec reviews
+
+Every Product has exactly one TypeSpec Review.
```
praveenkuttappan: This can be confusing. One Azure service/product can have more than one typespec project.

src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+
+### ✓ Provide a means to group related Products under a higher-level umbrella (future work).
+
+Products can manually be aggregated into service portfolios in the future. For example, KeyVault Keys, Secrets, and Certificates
```
praveenkuttappan: We should look at how Azure SDK docs are grouped and take the similar grouping ideas here. It has a language independent grouping.

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: Do we have an example of this?

src/dotnet/APIView/proposals/ProjectCentricProposal.md
tjprescott: This is helper for the latest approved or proposed TypeSpec namespace, since that's the canonical one. But we could also omit this and you just get whichever language-specific namespace you want.

src/dotnet/APIView/proposals/ProjectCentricProposal.md
tjprescott: Yes. Now the way things are structured in APIView today, we wouldn't necessarily have this field--we would have a "ProductId" field on Reviews, but I included it here to illustrate the connection. We might also want to expose this IMO.

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: This is where you look for the language-specific namespace approvals and history.

src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+### Namespace Proposal Updates
+
+```tsp
+/** Aggregated namespace information for a Product. */
+model ProductNamespaceInfo {
+  /** Currently approved namespaces. */
+  ApprovedNamespaces: NamespaceDecisionEntry[];
+  /** Full history of namespace proposals and decisions. */
+  NamespaceHistory: NamespaceDecisionEntry[];
+}
```
tjprescott: @maririos here's the answer to your above question. The idea is that any new namespace update will go in NamespaceHistory first, and then once a Namespace has been approved, it would go into the ApprovedNamespaces collection, which could simply be a convenience property.

src/dotnet/APIView/proposals/ProjectCentricProposal.md
tjprescott: Then this approach of the `NamespaceDecisionEntry` containing the namespace and package name is probably fine. If Java disagreed with one or the other they would just iterate until both were right.

src/dotnet/APIView/proposals/ProjectCentricProposal.md
tjprescott: AFAIK the only parser that sends this is Python. While there are helper methods in TCGC to standardize the logic, every language parser will need to send this info, which means parser updates are necessary.

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: The idea is to parser the tspconfig.yaml only:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/appconfiguration/data-plane/AppConfiguration/tspconfig.yaml

So you'd get namespaces and a placeholder name (in this case, probably `Appconfiguration` from the service-dir). Once you actually have a TypeSpec review created for this product, it would update the name to "Azure App Configuration". 

So I probably need to update this language.

From the app config example, it would parse:
- service-dir -> rough name ("appconfiguration")
- languages: (python, C#, javascript, java)
- proposed namespaces and package name

src/dotnet/APIView/proposals/ProjectCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+---
+
+## High-Level Design
+
+
+### 1. Introduce a `Product` Entity
+
+A Product:
+
+* Has a stable, opaque ID (a GUID generated by APIView), ensuring identity does not depend on the TypeSpec namespace.
```
praveenkuttappan, line 81: What's the use of this GUID? Why don't we use the relative path to spec project as the ID?

src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+
+Products can manually be aggregated into service portfolios in the future. For example, KeyVault Keys, Secrets, and Certificates
+could be grouped under a single KeyVault service portfolio.
+
+---
+
+## High-Level Design
+
+
+### 1. Introduce a `Product` Entity
```
praveenkuttappan: Product is already a well defined entity for Azure services in service tree. So this will be confusing when creating an entity of `Product` type that is just an entity for each typespec project.

src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+
+### Product Identity
+
+Each Product is assigned a stable, opaque identifier (ProductId), which is a GUID generated by APIView at creation time. This ensures that the Product’s identity is not tied to the TypeSpec namespace or any other mutable property. The ProductId is used as the primary key for all cross-language grouping and review attachment logic.
+
+### Namespace Matching
+
+The Product’s human-facing namespace (as defined in TypeSpec) is tracked separately from the ProductId. APIView maintains a namespace history for each Product, recording all proposed, approved, rejected, and withdrawn namespaces over time. This allows Products to evolve their namespaces without losing historical context or breaking identity links.
+
+When a new Review (for any language) is uploaded, APIView uses the Review’s CrossLanguagePackageId to find the correct Product. The attachment is successful if the CrossLanguagePackageId matches any namespace in the Product’s namespace history. This decouples identity from namespace and supports scenarios where namespaces change or are superseded.
```
praveenkuttappan: We should also mention how this impacts any manual review gen of uploading source or token file to APIView.

src/dotnet/APIView/proposals/ProjectCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+
+The Product’s human-facing namespace (as defined in TypeSpec) is tracked separately from the ProductId. APIView maintains a namespace history for each Product, recording all proposed, approved, rejected, and withdrawn namespaces over time. This allows Products to evolve their namespaces without losing historical context or breaking identity links.
+
+When a new Review (for any language) is uploaded, APIView uses the Review’s CrossLanguagePackageId to find the correct Product. The attachment is successful if the CrossLanguagePackageId matches any namespace in the Product’s namespace history. This decouples identity from namespace and supports scenarios where namespaces change or are superseded.
+
+This approach resolves several open questions about identity, namespace changes, and review attachment, and provides a robust foundation for future enhancements.
+
+### 2. Integration Points
+
+#### **When just a TypeSpec `tspconfig.yaml` is available**
```
praveenkuttappan, line 114: This does not create the review itself. right? What's the use case and benefit for a service team to register in APIView just to create their metadata in APIView?

src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+* Create a Product from `tspconfig.yaml`
+* Attach the TypeSpec review as the canonical spec for that Product
+
+#### **When a language Review is uploaded**
+
+APIView will automatically:
+
+* Look at `CrossLanguagePackageId`
+* Find the matching Product
+* Attach the Review to the Product
```
praveenkuttappan: Assuming that review is going to be tagged under an unknown product if product is not already registered. In that case, we will need to have a feature or support an option to manually link to correct product.

src/dotnet/APIView/proposals/ProjectCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+```tsp
+/** New top-level Product entity */
+model Product {
+  /** GUID identifier for the Product. */
+  Id: string;
+  /** Human-friendly name of the Product. Otherwise, defaults to the namespace. */
+  DisplayName?: string;
+  Description?: string;
+  /** Service-level contacts. */
+  Owners?: string[];
```
praveenkuttappan, line 158: There is another project driven by engsys for code owner and service owner and its planned to be stored in DevOps work item. What's the use case of this owner information. We should avoid maintaining the owner info here and take benefit of centrally managed owner info.

src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+from Reviews.
+
+```diff
+ model Review {
+  ... existing fields ...
+-    NamespaceReviewStatus: NamespaceReviewStatus;
+-    NamespaceApprovalRequestedBy: string;
+-    NamespaceApprovalRequestedOn?: utcDateTime;
++    ProductId?: string;
++    IsTypeSpecReview?: boolean;
```
praveenkuttappan: Why do we need this? Review already has a `Language` property. We can always have a wrapper property in .NET side if it is helpful but I don 't think we should store this extra bool in DB.

src/dotnet/APIView/proposals/ProjectCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+
+We will also remove the `NamespaceReviewStatus` enum.
+
+---
+
+## Key Workflows
+
+### Workflow 1 — Product Creation via TypeSpec Review
+
+1. EngSys parses `tspconfig.yaml` from a TypeSpec Review
```
praveenkuttappan, line 240: Is this part of TypeSpec engsys pipeline or is it something done by TypeSpec APIview parser?

src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+
+### Alternative Workflow 1 — Manual Product Creation
+
+1. User uploads `tspconfig.yaml` or provides its URI
+2. APIView parses it and creates a Product
+3. User uploads TypeSpec Review and language Reviews later, which will attach automatically
+
+### Alternative Workflow 2 — Existing Reviews Migration
+
+1. Script creates Products for existing TypeSpec Reviews
```
praveenkuttappan: Existing TypeSpec reviews does not have any source information so we need to think about this migration and how to get the details from tspconfig.yml

src/dotnet/APIView/proposals/ProjectCentricProposal.md
praveenkuttappan: Java package name is combination of group name and package name. So effectively approving the package name implicitly approves namespace. But a good question for language architects.

src/dotnet/APIView/proposals/ProductCentricProposal.md
```diff
@@ -0,0 +1,306 @@
...
+  a. Proposed → Withdrawn
+  b. Rejected → Rejected (keep the same)
+
+---
+
+
+## Open Questions
+
+1. Do we care about just namespace or also package name? The current proposal has both fields, so approval, rejection, etc. is implicitly for both. Is that acceptable?
+2. The namespace behavior proposed would mean that a Product could have multiple rejected namespaces, and other Products would not be able to use those namespaces. Is that acceptable?
```
praveenkuttappan: This is going to be a rare scenario. But I don't think we should block this. What happens if a service team decides to split their spec to two projects and one of them contain previous package name and second project contains a new package name? So, we should support moving package hierarchy under a new product in that case.

src/dotnet/APIView/proposals/ProjectCentricProposal.md
tjprescott: The purpose of the GUID is to make sure that the first proposed namespace doesn't become the ID. Similarly, if initial relative path is wrong we don't want that codified as the ID. Also, if you just upload a tspconfig.yaml you wouldn't have that path.

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: Maybe `Project` instead?

src/dotnet/APIView/proposals/ProjectCentricProposal.md
tjprescott: Correct, this would only create the Product in APIView, with no Reviews. The use case is where a service wants to get namespaces approved without having their TypeSpec complete or any SDKs generated.

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: Well, when a token file is received, it will try to create an APIRevision and link it to a Review, and, if that review is not already linked, to a Product. Just like if there is no review, it will create one, if there is no product, it will create one. This would probably be the most common case. 

Hopefully it will not be necessary to manually link things frequently, in which case we could handle the occasional edge case in the DB. Worse case, yeah, we have an option (probably an admin-only option) to do this via UI.

src/dotnet/APIView/proposals/ProjectCentricProposal.md
tjprescott: Right now this would be storing Github usernames, which is how everything else is done in APIView. The ability to cross-reference an existing store of this info would be great, assuming they are compatible.

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: Yeah I'm fine with that. Since this is just the TypeSpec for the service, I'm including any kind of "calculated properties" on the model.

src/dotnet/APIView/proposals/ProjectCentricProposal.md
tjprescott: This would be that magical whatever script either we write or Central EngSys does that gives a consistent, usable output from a tsp-config.yaml. It would not be the TypeSpec APIView parser (which works with a tsp code model, not YAML files).

src/dotnet/APIView/proposals/ProductCentricProposal.md
maririos: - is `service-dir` unique and present in all tspconfig?
- what happens if an API spec doesn't have a tspconfig? the APIViews for those typespecs won't be displayed as a product?

src/dotnet/APIView/proposals/ProjectCentricProposal.md
maririos: Agree on the parser updates. I remember when we talked to JS they mentioned they would also have to change how their codegen works to expose that. So there is also work involved for the codegen of each language

src/dotnet/APIView/proposals/ProductCentricProposal.md
maririos: For SDKs docs they use the 'package display name' that is in the 'package' work item in devops.
> In each item for your release, you should have a value for Service Name and Package Display Name. The docs table of contents is generated from this data.

https://eng.ms/docs/products/azure-developer-experience/develop/sdk-docs

Not sure about API spec docs though

maririos: (I am asking around :) https://github.com/Azure/azure-sdk-tools/issues/13109 )

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: This is just describing a manual flow where you create the Product by uploading a tsp-config.yaml (practically speaking, EVERY TypeSpec project must have one). The product would also get created automatically when a TypeSpec review was created (most likely by CI)

src/dotnet/APIView/proposals/ProjectCentricProposal.md
tjprescott: Yes, and since we were changing the process, that's why we deliberately only collaborated with Python (or I guess, at the time it was "us" [Python] collaborating with "you") 😆

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: I believe this is mentioned elsewhere in the scenario descriptions. Is there something missing?

src/dotnet/APIView/proposals/ProductCentricProposal.md
maririos: Is it true that every typespec project has a tspconfig.yaml? Will the CI block a PR if they don't have it?

tjprescott: Yes. For all intents and purposes, every TypeSpec project must have a tspconfig.yaml. I don't think it's enforced directly by the CI, but more by the compiler. If you don't use tspconfig you have to use command line arguments, which none of our specs are going to do.

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: I think this is just a restatement of the idea that "Product" is not the right term, yes? Or are you saying we release SDKs that combinations of multiple TypeSpec projects?

tjprescott: Renamed to Project.

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: Renamed to Project.

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: This is out of scope for this proposal--just a mention that this proposal would support such an additional level of hierarchy.

src/dotnet/APIView/proposals/ProductCentricProposal.md
tjprescott: Yeah, this is on review, so you're right... just look at the language! Removed.

src/dotnet/APIView/proposals/parse_tspconfig.py
```diff
@@ -0,0 +1,147 @@
+# Parse TypeSpec config YAML files to extract language-specific metadata
+import sys
```
tjprescott, line 2: Added this draft script. Here's what it outputs for the yaml files included:
```
C:\repos\azure-sdk-tools\src\dotnet\APIView>python proposals\parse_tspconfig.py proposals\ai-agents.yaml
--- ai-agents.yaml ---
{'language': 'python', 'package_name': 'azure-ai-agents', 'namespace': 'azure.ai.agents'}
{'language': 'csharp', 'package_name': 'Azure.AI.Agents.Persistent', 'namespace': 'Azure.AI.Agents.Persistent'}
{'language': 'typescript', 'package_name': '@azure/ai-agents', 'namespace': '@azure/ai-agents'}

C:\repos\azure-sdk-tools\src\dotnet\APIView>python proposals\parse_tspconfig.py proposals\storage.yaml  
--- storage.yaml ---
{'language': 'python', 'package_name': 'azure-storage-blob', 'namespace': 'azure.storage.blob'}
{'language': 'csharp', 'package_name': 'Azure.Storage.Blob', 'namespace': 'Azure.Storage.Blob'}
{'language': 'java', 'package_name': 'azure-storage-blob', 'namespace': 'com.azure.storage.blob'}
{'language': 'typescript', 'package_name': '@azure-rest/azure-storage-blob', 'namespace': '@azure-rest/azure-storage-blob'}
{'language': 'rust', 'package_name': 'azure_storage_blob', 'namespace': 'azure_storage_blob'}

C:\repos\azure-sdk-tools\src\dotnet\APIView>python proposals\parse_tspconfig.py proposals\keyvault-mgmt.yaml
--- keyvault-mgmt.yaml ---
{'language': 'python', 'package_name': 'azure-mgmt-keyvault', 'namespace': 'azure.mgmt.keyvault'}
{'language': 'csharp', 'package_name': 'Azure.ResourceManager.KeyVault', 'namespace': 'Azure.ResourceManager.KeyVault'}
{'language': 'java', 'package_name': 'azure-resourcemanager-keyvault', 'namespace': 'com.azure.resourcemanager.keyvault'}
{'language': 'typescript', 'package_name': '@azure/arm-keyvault', 'namespace': '@azure/arm-keyvault'}
{'language': 'go', 'package_name': 'sdk/resourcemanager/keyvault/armkeyvault/v2', 'namespace': 'sdk/resourcemanager/keyvault/armkeyvault/v2'}
```

src/dotnet/APIView/proposals/ProjectCentricProposal.md
```diff
@@ -0,0 +1,307 @@
...
+
+1. The namespace behavior proposed would mean that a Project could have multiple rejected namespaces, and other Projects would not be able to use those namespaces. Is that acceptable?
+
+---
+
+## Next Steps
+
+* Validate with EngSys that tspconfig-derived metadata can be reliably emitted
+* Create migration script for existing Reviews
+* Create new endpoint in APIView for registering new Projects manually with `tspconfig.yaml`
```
praveenkuttappan, line 304: This should also be exposed as a MCP tool to register new namespace.

tjprescott:
I opened a new implementation issue with the approved proposal: https://github.com/Azure/azure-sdk-tools/issues/13301

I also opened an issue for the script creation, including the POC for guidance: https://github.com/Azure/azure-sdk-tools/issues/13302

Closing this PR.

PR #13053, opened 2025-11-24
[js-api-parser] add function token generator

KarishmaGhiya:
Adding function token generator

### DIFF
--- a/tools/apiview/parsers/js-api-parser/src/tokenGenerators/function.ts
+++ b/tools/apiview/parsers/js-api-parser/src/tokenGenerators/function.ts
@@ -1,49 +1,153 @@
-import { ApiFunction, ApiItem, ApiItemKind, ExcerptTokenKind } from "@microsoft/api-extractor-model";
+import {
+  ApiFunction,
+  ApiItem,
+  ApiItemKind,
+  ExcerptTokenKind,
+  Parameter,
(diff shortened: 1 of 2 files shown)


tools/apiview/parsers/js-api-parser/src/tokenGenerators/function.ts
```diff
@@ -0,0 +1,52 @@
...
+    tokens.push({Kind: TokenKind.Keyword, Value: "export", HasSuffixSpace: true, IsDeprecated: deprecated});
+    
+    // Add function keyword
+    tokens.push({Kind: TokenKind.Keyword, Value: "function", HasSuffixSpace: true, IsDeprecated: deprecated});
+    
+    // Add function name
+    tokens.push({Kind: TokenKind.MemberName, Value: item.displayName, IsDeprecated: deprecated});
+    
+    // Process parameters and return type from excerptTokens
+    for (const excerpt of item.excerptTokens) {
```
jeremymeng: Should we try using properties available in `ApiFunction` first (`parameters`/`typeParameters`) before falling back to use the `excerptTokens`?

KarishmaGhiya: Will do

Issue #12507, opened 2025-10-15
Update scripts of processing release plan work items to use full name for Java libraries


raych1:
The full name of Java libraries will be `groupid:artifactId`.

Here're the related scripts:
https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Validate-All-Packages.ps1

PR #13076, opened 2025-11-26
Use full package name in DevOps work item

raych1:
Fixed https://github.com/Azure/azure-sdk-tools/issues/12507

Key Changes:
- added a new function to return full package name
- pass through the `groupId` when dealing with the dev ops work item
- updated `Get-PkgProperties` to match both of the `name` and `groupId` if `Group` exists

### DIFF
--- a/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
+++ b/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
@@ -139,14 +139,14 @@ function Invoke-Query($fields, $wiql, $output = $true)
   return $workItems
 }
 
-function BuildHashKeyFromNonNullArgs()
+function BuildHashKeyNoNull()
 {
   $filterNulls = $args | Where-Object { $_ }
-  # if we had any non-nulls then return it
-  if (!$filterNulls) {
+  # if we had any nulls then return null
+  if (!$filterNulls -or $args.Count -ne $filterNulls.Count) {
     return $null
   }
-  return BuildHashKey @filterNulls
+  return BuildHashKey $args
 }
 
 function BuildHashKey()
@@ -241,7 +241,7 @@ function FindLatestPackageWorkItem($lang, $packageName, $groupId = $null, $outpu
 
 function FindPackageWorkItem($lang, $packageName, $version, $groupId = $null, $outputCommand = $true, $includeClosed = $false, $ignoreReleasePlannerTests = $true, $tag = $null)
 {
-  $key = BuildHashKeyFromNonNullArgs $lang $packageName $version $groupId
+  $key = BuildHashKey $lang $packageName $version $groupId
   if ($key -and $packageWorkItems.ContainsKey($key)) {
     return $packageWorkItems[$key]
   }
@@ -300,7 +300,7 @@ function FindPackageWorkItem($lang, $packageName, $version, $groupId = $null, $o
 
   foreach ($wi in $workItems)
   {
-    $localKey = BuildHashKeyFromNonNullArgs $wi.fields["Custom.Language"] $wi.fields["Custom.Package"] $wi.fields["Custom.PackageVersionMajorMinor"] $wi.fields["Custom.GroupId"]
+    $localKey = BuildHashKey $wi.fields["Custom.Language"] $wi.fields["Custom.Package"] $wi.fields["Custom.PackageVersionMajorMinor"] $wi.fields["Custom.GroupId"]
     if (!$localKey) {
       $packageWorkItemWithoutKeyFields[$wi.id] = $wi
       Write-Host "Skipping package [$($wi.id)]$($wi.fields['System.Title']) which is missing required fields language, package, or version."


eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -266,3 +266,25 @@ function Split-ArrayIntoBatches {
...
+        $groupId = $PackageInfo.Group
+        if ($groupId) {
+            if ($UseColonSeparator) {
+                return "${groupId}:$($PackageInfo.ArtifactName)"
+            }
+            return "${groupId}+$($PackageInfo.ArtifactName)"
+        }
+    }
+    return $PackageInfo.Name
+}
```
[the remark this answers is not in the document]
raych1: Added tests

eng/common/pipelines/templates/steps/verify-changelog.yml
```diff
@@ -14,6 +14,9 @@ parameters:
 - name: Condition
   type: string
   default: succeeded()
+- name: GroupId
```
weshaggard: Why is this needed? We should be using the [verify-changelogs.yml](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/steps/verify-changelogs.yml) which uses the PackageInfo

eng/common/scripts/Helpers/Package-Helpers.ps1
```diff
@@ -266,3 +266,25 @@ function Split-ArrayIntoBatches {
 
   return , $batches
 }
+
+# Get the full package name based on packageInfo properties
+# Returns Group+ArtifactName if Group exists and has a value, otherwise returns Name
+# If UseColonSeparator switch is enabled, returns Group:ArtifactName format (colon separator)
+function Get-FullPackageName {
```
weshaggard: Why not put this function directly on the PackageProps class in Pacakge-Properties.ps1

eng/common/pipelines/templates/steps/verify-changelog.yml
raych1: Sounds good. I updated it by the package info file path.

eng/common/scripts/Helpers/Package-Helpers.ps1
raych1: I thought about that but ended up here. Moved it.

eng/common/scripts/Package-Properties.ps1
```diff
@@ -576,3 +576,25 @@ function Get-PkgPropsForEntireService ($serviceDirectoryPath) {
 
     return $projectProps
 }
+
+# Get the full package name based on packageInfo properties
+# Returns Group+ArtifactName if Group exists and has a value, otherwise returns Name
+# If UseColonSeparator switch is enabled, returns Group:ArtifactName format (colon separator)
+function Get-FullPackageName {
```
weshaggard, line 594: Any not make this a function on the PackageProps type itself?

eng/common/pipelines/templates/steps/verify-changelog.yml
weshaggard: We shouldn't need any changes to this file. My goal is to eventually delete it once we convert everyone over to verify-changelogs.yml. Can you convert usages to that instead of needing to edit this?

eng/common/scripts/Package-Properties.ps1
raych1: We could do that. With that, for the cases where have the `packageInfo` JSON, we will need to call a new function to create `PackageProps` object from the JSON before invoking this member function. It is more flexible to have it as individual function to allow matching both of the `PackageProps` object and `PackageInfo` JSON object.

weshaggard: I see when we parse from the JSON object we aren't getting a PackageProps object. That is a good reason to keep it separate.

eng/common/pipelines/templates/steps/verify-changelog.yml
raych1: I reverted all the related changes to this template. I will convert Java pipeline template to use the `verify-changelogs.yml`. I created [this PR](https://github.com/Azure/azure-sdk-tools/pull/13102) to allow `forRelease` setting.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -513,6 +526,7 @@ function CreateOrUpdatePackageWorkItem($lang, $pkg, $verMajorMinor, $existingIte
     return
   }
   $pkgName = $pkg.Package
+  $pkgGroupId = if ($pkg.PSObject.Properties.Name -contains "GroupId") { $pkg.GroupId } else { $pkg.Group }
```
raych1: package csv uses `GroupId` as the name while `PackageInfo` and `work item` use `Group`. We might want to make them consistent in the future.

weshaggard: I would prefer to call this GroupId for the new field if possible to make this consistent. Ideally we would also update the PackageProps to rename that to GroupId as well but that might be a little more work.

weshaggard: Are we making a bad assumption that if Group is a property if GroupId isn't? Should we check to see if Group exists as well?

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -1277,6 +1295,7 @@ function Update-DevOpsReleaseWorkItem {
 
   $packageInfo = [PSCustomObject][ordered]@{
     Package = $packageName
+    Group = $groupId
```
weshaggard: For new fields and places lets standardize on calling them GroupId so it is clear that is what it is for.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
raych1: If that's the direction, let's make the new field called `GroupId` in work item. I will create an issue for tracking the future update on the `PackageProps`.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
raych1: Updated it to use '`GroupId` for new fields.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
raych1: Checking both now.

raych1: Created https://github.com/Azure/azure-sdk-tools/issues/13226

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -512,6 +525,15 @@ function CreateOrUpdatePackageWorkItem($lang, $pkg, $verMajorMinor, $existingIte
     Write-Host "Cannot create or update because one of lang, pkg or verMajorMinor aren't set. [$lang|$($pkg.Package)|$verMajorMinor]"
     return
   }
+
+  # PackageProp object uses Group, while other places use GroupId, such as in work item fields and package csv files.
+  $pkgGroupId = if ($pkg.PSObject.Properties.Name -contains "GroupId") {
+    $pkg.GroupId
+  } elseif ($pkg.PSObject.Properties.Name -contains "Group") {
```
weshaggard: Is this still needed for the PackageProps?

raych1: You're right, it processes the data from dev-ops work item and package csv. Both of them use `GroupId`.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -238,9 +239,13 @@ function FindLatestPackageWorkItem($lang, $packageName, $outputCommand = $true,
   return $latestWI
 }
 
-function FindPackageWorkItem($lang, $packageName, $version, $outputCommand = $true, $includeClosed = $false, $ignoreReleasePlannerTests = $true, $tag = $null)
+function FindPackageWorkItem($lang, $packageName, $version, $groupId = $null, $outputCommand = $true, $includeClosed = $false, $ignoreReleasePlannerTests = $true, $tag = $null)
 {
-  $key = BuildHashKeyNoNull $lang $packageName $version
```
weshaggard, line 243: Maybe we should update the BuildHashKeyNoNull to have a version that allows null and only includes it if the value isn't null.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -523,6 +543,7 @@ function CreateOrUpdatePackageWorkItem($lang, $pkg, $verMajorMinor, $existingIte
   $fields = @()
   $fields += "`"Language=${lang}`""
   $fields += "`"Package=${pkgName}`""
+  $fields += "`"GroupId=${pkgGroupId}`""
```
weshaggard, line 538: Should we only include this clause if groupId isn't null?

eng/common/scripts/Package-Properties.ps1
```diff
@@ -230,16 +230,30 @@ class PackageProps {
...
     $allPkgProps = Get-AllPkgProperties -ServiceDirectory $ServiceDirectory
-    $pkgProps = $allPkgProps.Where({ $_.Name -eq $PackageName -or $_.ArtifactName -eq $PackageName });
+    
+    if ([string]::IsNullOrEmpty($GroupId)) {
+        $pkgProps = $allPkgProps.Where({ $_.Name -eq $PackageName -or $_.ArtifactName -eq $PackageName });
+    }
+    else {
+        $pkgProps = $allPkgProps.Where({ 
+            ($_.Name -eq $PackageName -or $_.ArtifactName -eq $PackageName) -and 
+            ($_.PSObject.Properties.Name -contains "Group" -and $_.Group -eq $GroupId)
```
weshaggard, line 254:
```suggestion
            ($_.PSObject.Properties.Name -contains "GroupId" -and $_.GroupId -eq $GroupId)
```

weshaggard: Nevermind this looks to be a packageprops/info so I guess it would still be Group.

weshaggard (approved):
Couple potential clean-ups but otherwise looks good.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
raych1: I'd rather keeping it updated to the input value even if it's null.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
raych1: Updated the function to build key based on non-null arguments.

eng/common/scripts/Package-Properties.ps1
raych1: The input object comes from the `PackageProps` type and we have to access it by `Group`.

eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -139,14 +139,14 @@ function Invoke-Query($fields, $wiql, $output = $true)
   return $workItems
 }
 
-function BuildHashKeyNoNull()
+function BuildHashKeyFromNonNullArgs()
```
weshaggard: BuildHashKey already builds from all non-null values.

I don't remember why we added the BuildHashKeyNoNull function but I suggest keeping it unchanged and just update your callers to use BuildHashKey function directly.

raych1: All right, updated.

PR #13096, opened 2025-12-01
fix 2459: miss message in history issue

chunyu3:
Fix https://github.com/Azure/azure-sdk-pr/issues/2459

- We will include all the messages in history include those message with 'has_result: false', so will remove the filter rule: ignore 'has_result: false' message.
- refine the code, remove empty file 'promptGenerator.ts', and rename 'promptGeneratorV2.ts' to 'promptGenerator.ts'
- set the thinking message relay time as the Bot reply time to resolve ordering issue.

### DIFF
--- a/eng/common-tests/ChangeLog-Operations.Tests.ps1
+++ b/eng/common-tests/ChangeLog-Operations.Tests.ps1
@@ -1,0 +1,227 @@
+Import-Module Pester
+
+BeforeAll {
+    . $PSScriptRoot/../common/scripts/ChangeLog-Operations.ps1
+}
+
+Describe "Parse-ChangelogContent" {
+    It "Should parse changelog text with multiple sections" {
+        $changelogText = "### Breaking Changes`n`n- Removed deprecated API ``oldMethod()```n- Changed return type of ``getData()```n`n### Features Added`n`n- Added new ``newMethod()`` API`n- Added support for async operations`n`n### Bugs Fixed`n`n- Fixed null pointer exception in ``processData()``"
+
+        $result = Parse-ChangelogContent -ChangelogText $changelogText -InitialAtxHeader "#"
+
+        $result | Should -Not -BeNullOrEmpty
+        $result.ReleaseContent | Should -Not -BeNullOrEmpty
+        $result.Sections | Should -Not -BeNullOrEmpty
+        $result.Sections.Keys | Should -HaveCount 3
+        $result.Sections.ContainsKey("Breaking Changes") | Should -BeTrue
+        $result.Sections.ContainsKey("Features Added") | Should -BeTrue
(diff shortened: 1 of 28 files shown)


tools/sdk-ai-bots/azure-sdk-qa-bot/src/turn/ThinkingHandler.ts
```diff
@@ -50,7 +52,7 @@ export class ThinkingHandler {
   }
 
   // separate this method from cancelTimer to make sure complete message is always shown
-  public async stop(reply: CompletionResponsePayload | RagApiError, currentPrompt: Prompt) {
+  public async stop(replyStartTime: Date, reply: CompletionResponsePayload | RagApiError, currentPrompt: Prompt) {
```
JiaqiZhang-Dev, line 55: could `replyStartTime` just be `replyTime`?

tools/sdk-ai-bots/azure-sdk-qa-bot/src/turn/ThinkingHandler.ts
```diff
@@ -195,7 +199,7 @@ export class ThinkingHandler {
       conversationId,
       activityId: replyActivityId,
       reply,
-      timestamp: prompt.timestamp,
```
JiaqiZhang-Dev, line 198: what does the original prompt.timestamp mean? could we remove the timestamp from prompt model?

tools/sdk-ai-bots/azure-sdk-qa-bot/src/turn/ThinkingHandler.ts
chunyu3: @JiaqiZhang-Dev When bot reply, it will first reply a placeholder 'Thinking...', we should use this timestamp. it is the start reply time, it is not the actual time we get the reply from backend service.
To clarify those two for future maintenance, I prefer to 'replyStartTime'

tools/sdk-ai-bots/azure-sdk-qa-bot/src/turn/ThinkingHandler.ts
chunyu3: prompt is the query message( compare to the reply message), not the prompt, promt.timestamp is the timestamp when user send the qa query. 

the variable name is a little confusing. we will update late.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/turn/ThinkingHandler.ts
JiaqiZhang-Dev: Got it, thanks for clarify

PR #13102, opened 2025-12-01
Added 'forRelease' input parameter

raych1:
Make it ready for release pipeline integration

### DIFF
--- a/eng/common/scripts/Verify-ChangeLogs.ps1
+++ b/eng/common/scripts/Verify-ChangeLogs.ps1
@@ -34,6 +34,8 @@ foreach($propertiesFile in $packageProperties) {
         continue
   }
 
+  Write-Host "Verifying changelog for $($PackageProp.Name)"
+
   $validChangeLog =  Confirm-ChangeLogEntry -ChangeLogLocation $PackageProp.ChangeLogPath -VersionString $PackageProp.Version -ForRelease $ForRelease
 
   if (-not $validChangeLog) {


eng/common/pipelines/templates/steps/verify-changelogs.yml
raych1, line 2: @weshaggard looks like scanning thru the `packageInfo` folder wouldn't work for the multiple package scenarios which want to skip the chagnelog validation.  We might need to add the support to take a single `Package Info File` input. 

e.g. [Java spring CI](https://github.com/Azure/azure-sdk-for-java/blob/6daf347877504ac4c1be536f14f5e8b15647a73d/sdk/spring/ci.yml#L272)

[place to call 'verifychangelog.yml'](https://github.com/Azure/azure-sdk-for-java/blob/6daf347877504ac4c1be536f14f5e8b15647a73d/eng/pipelines/templates/jobs/ci.yml#L400)

maririos:
@raych1 Could you provide more context on what the purpose of adding this property is and how it would change the existing behavior?

eng/common/pipelines/templates/steps/verify-changelogs.yml
weshaggard: Or we figure out another way to filter out the verification. For cases like spring it seems like we should disable it completely for that pipeline.

raych1: The annoying thing is only one package within spring group - [`azure-spring-data-cosmos`](https://github.com/Azure/azure-sdk-for-java/blob/6daf347877504ac4c1be536f14f5e8b15647a73d/sdk/spring/ci.yml#L258) requires the validation.  Otherwise, 

Python also has same situation for group packages. i.e. https://github.com/search?q=repo%3AAzure%2Fazure-sdk-for-python%20skipVerifyChangelog&type=code

weshaggard: https://github.com/Azure/azure-sdk-for-python/blob/8884ad72c87a9a618b4ab50c299e73548e780e36/eng/common/scripts/Verify-ChangeLogs.ps1#L11

Doesn't that handle this property?

raych1: As synced offline, the linked code handles the flag set in the yml files.

raych1:
> @raych1 Could you provide more context on what the purpose of adding this property is and how it would change the existing behavior?

@maririos the change is to align with current behavior implemented by [another template](https://github.com/Azure/azure-sdk-tools/blob/eb97e12be7f013d5f2d6333c8cf1603725601ce4/eng/common/pipelines/templates/steps/verify-changelog.yml#L11). The goal is to deprecate the former template and use the one updated in this PR.

PR #13104, opened 2025-12-01
[AVC] Changed evals run flag from --use-cache to --use-recording

marcarro:


packages/python-packages/apiview-copilot/evals/run.py
```diff
@@ -22,13 +22,13 @@
         help="Paths to directories containing test files.",
     )
     parser.add_argument(
-        "--use-cache",
+        "--use-recording",
         action="store_true",
-        help="Enable caching of evaluation results to speed up runs.",
+        help="Enable recording of evaluation results to speed up runs.",
```
tjprescott: Should say something like "Use recordings instead of making actual LLM calls to speed runs. If recordings are not available, LLM calls will be made and saved as recordings."

This is more accurate to what it actually does.

packages/python-packages/apiview-copilot/cli.py
```diff
@@ -1177,10 +1177,10 @@ def load_arguments(self, command):
                 help="The full paths to the folder(s) containing the test files. Must have a `test-config.yaml` file. If omitted, runs all workflows.",
             )
             ac.argument(
-                "use_cache",
-                options_list=["--use-cache"],
+                "use_recording",
+                options_list=["--use-recording"],
                 action="store_true",
-                help="Use cached results for testcases when available.",
+                help="Use recorded results for testcases when available.",
```
tjprescott: Same as above.

PR #13030, opened 2025-11-21
List tools command to list MCP tools, description and CLI command

praveenkuttappan:
This PR adds CLI command to list all MCP tools, description and CLI command in Az SDK MCP server.

usage:
`azsdk list -o json` (to get json output)
`azsdk list`

PR #13108, opened 2025-12-01
Automate MCP list documentation

### LINKED ISSUE #12039: AzSDK MCP server: auto generate CLI documentation
We need to auto generate CLI documentation whenever new changes are merged to Azure SDK tools MCP server.

Solution:

1. Have script or command within the MCP server that generates a markdown file for entire CLI supported by MCP server.
2. Include this markdown in azure-sdk-tools repo.
3. Run the script as part of PR pipeline and add modified markdown file to pull request if cli documentation file is updated.

### LINKED ISSUE #12655: List of MCP tool
In order to help telemetry, we need the list of MCP function that exists. This list is useful because:
- It should not contain test function like azsdk_hello_world
- If a function is never called, the telemetry couldn't know it exists

We'd like to keep this list in the repo, as documentation and importable spot for the PowerBI dashboard.

Ideally, we make sure with CI that each time someone adds a MCP function, the file is up to date.

jeo02:
- Leverage the PR created on release for incrementing package number to also update documentation.
- Added custom job to publish pipeline that is only activated when we are verifying changelog.
- Updated MCP list docs.
- Resolves #12039
- Resolves #12655


tools/azsdk-cli/update-mcp-tools-docs.ps1
```diff
@@ -0,0 +1,123 @@
...
+}
+
+function Get-McpToolsList {
+    param([string]$azsdkPath)
+    
+    if (-not (Test-Path $azsdkPath)) {
+        throw "Azure SDK CLI executable not found at: $azsdkPath"
+    }
+    
+    $jsonOutput = & $azsdkPath list --output json 2>&1
```
benbp: @praveenkuttappan I'm just going through https://github.com/Azure/azure-sdk-tools/pull/13030 that supports this and it seems like a lot of extra complexity and ongoing boilerplate to couple MCP tools list with the command classes. It doesn't necessarily map 1:1 with commands to support tools list either.

I think we could instead start the MCP server here and call `tools/list` via json-rpc (see example https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/mcp/azure-sdk-mcp.ps1#L58-L66).

If we wanted to support something like `azsdk mcp list` locally for users then we should be able to use the mcp sdk functions to figure this out and return the correct data, instead of changing our class structure and requiring duplication/hardcoding of tool names.

praveenkuttappan: @benbp Summer needs the tools list, description for the eval and documentation needs both tool and CLI commands. @jeo02 was trying json-rpc earlier and Wes suggested to follow the approach taken my microsoft mcp. So changes in 13030 was done to map tool name to command name.

benbp: @weshaggard @praveenkuttappan is the idea that we want the documentation to be able to indicate that X command can also be called as Y tool? If not, then the approach introduces a brittle coupling that developers have to know to manually keep up to date when adding/changing commands. If so, then I would think we could use the type system instead of manual annotations to indicate shared methods.

praveenkuttappan: Doc will show tool name and corresponding CLI command so user can find the CLI command for a tool.

eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
```diff
@@ -119,6 +120,9 @@ stages:
             # run this locally and figure out what the inputs should be
             arguments: '-Version $(NewVersion) -ChangelogPath "${{ parameters.ToolDirectory }}/CHANGELOG.md"'
             pwsh: true
+        
+        - ${{ each job in parameters.CustomReleaseJobs }}:
```
praveenkuttappan: @jeo02 We cannot add a custom job here. We will have to add this as custom post release step instead of a job.

tools/azsdk-cli/auto-documentation.yml
```diff
@@ -0,0 +1,40 @@
...
+      displayName: 'Checkout current branch'
+    
+    - template: /eng/common/pipelines/templates/steps/sparse-checkout.yml
+      parameters:
+        Repositories:
+          - Name: Azure/azure-sdk-tools
+            Commitish: main
+            WorkingDirectory: $(System.DefaultWorkingDirectory)/azure-sdk-tools-main
+        Paths:
+          - 'tools/azsdk-cli/**'
```
praveenkuttappan: why do we need this one when repo is already checked out at line 10? All of the below steps are using default checked out path and not this sparse checked path.

tools/azsdk-cli/auto-documentation.yml
```diff
@@ -0,0 +1,40 @@
...
+        projects: '$(System.DefaultWorkingDirectory)/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Azure.Sdk.Tools.Cli.csproj'
+        arguments: '--configuration Release'
+    
+    - task: PowerShell@2
+      displayName: 'Generate MCP tools documentation'
+      inputs:
+        targetType: 'filePath'
+        filePath: '$(System.DefaultWorkingDirectory)/tools/azsdk-cli/update-mcp-tools-docs.ps1'
+        arguments: >-
+          -AzsdkCliPath $(System.DefaultWorkingDirectory)/artifacts/bin/Azure.Sdk.Tools.Cli/Release/net8.0/azsdk
```
praveenkuttappan: This will break if .NET version is different which ends up creating executable in different path.

tools/azsdk-cli/auto-documentation.yml
```diff
@@ -0,0 +1,40 @@
...
+    
+    - task: DotNetCoreCLI@2
+      displayName: 'Build Azure SDK CLI (Release)'
+      inputs:
+        command: 'build'
+        projects: '$(System.DefaultWorkingDirectory)/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Azure.Sdk.Tools.Cli.csproj'
+        arguments: '--configuration Release'
+    
+    - task: PowerShell@2
+      displayName: 'Generate MCP tools documentation'
```
praveenkuttappan: We can avoid building the exe using existing script that will download the exe.
1. Run ` pwsh ./eng/common/mcp/azure-sdk-mcp.ps1 -InstallDirectory  $(System.DefaultWorkingDirectory)`
2. Pass  $(System.DefaultWorkingDirectory)/azsdk as the executable path

tools/azsdk-cli/update-mcp-tools-docs.ps1
```diff
@@ -0,0 +1,123 @@
...
+function Get-VersionFromCsproj {
+    param([string]$csprojPath)
+    
+    if (-not (Test-Path $csprojPath)) {
+        throw "Csproj file not found at: $csprojPath"
+    }
+    
+    [xml]$csprojXml = Get-Content $csprojPath
+    return $csprojXml.Project.PropertyGroup.VersionPrefix
+}
```
praveenkuttappan: You can run` .\azsdk.exe --version` to get the current MCP version.

eng/pipelines/templates/stages/archetype-sdk-publish-net.yml
praveenkuttappan: As discussed, let's add the custom job outside the current job and include create pr step in it.

eng/common/pipelines/templates/jobs/ai-eval-job.yml
```diff
@@ -58,40 +53,42 @@ jobs:
...
+            Commitish: ${{ parameters.EvalRepoCommit }}
+            WorkingDirectory: $(System.DefaultWorkingDirectory)/${{ split(parameters.EvalRepo, '/')[1] }}
+          - Name: ${{ parameters.TargetRepo }}
+            Commitish: ${{ parameters.TargetRepoCommit }}
+            WorkingDirectory: $(System.DefaultWorkingDirectory)/${{ split(parameters.TargetRepo, '/')[1] }}
         Paths:
           - 'tools/**'
-          - eng/common/**
-          - .github/copilot-instructions.md
+          - 'eng/common/**'
```
benbp:
```suggestion
```

All of `eng` is included by default in this template.

tools/azsdk-cli/auto-documentation.yml
```diff
@@ -0,0 +1,39 @@
...
+      name: $(LINUXPOOL)
+      image: $(LINUXVMIMAGE)
+      os: linux
+    
+    steps:
+    - checkout: self
+      displayName: 'Checkout current branch'
+    
+    - template: /eng/pipelines/templates/steps/install-dotnet.yml
+
```
benbp:
```suggestion
```

I don't think this is necessary since we're installing the self contained version.

tools/azsdk-cli/update-mcp-tools-docs.ps1
```diff
@@ -0,0 +1,125 @@
...
+    }
+    
+    $versionOutput = & $exePath --version 2>&1
+    
+    if ($LASTEXITCODE -ne 0) {
+        throw "Failed to run azsdk --version command. Exit code: $LASTEXITCODE"
+    }
+    
+    # Remove the +sha part from version (e.g., "0.5.9+adef6a4419984a5852e4118d05f695b90d007044" -> "0.5.9")
+    $version = $versionOutput.Trim() -replace '\+.*$', ''
```
benbp: I just put up https://github.com/Azure/azure-sdk-tools/pull/13182 which will remove the sha from the version (I don't think it's helpful to have anyway).

eng/common/pipelines/templates/jobs/ai-eval-job.yml
jeo02: oh woops that part of another PR that somehow snuck onto here I'll remove it but will be using that advice in the other PR :)

tools/azsdk-cli/update-mcp-tools-docs.ps1
```diff
@@ -0,0 +1,125 @@
...
+    
+    foreach ($tool in $mcpTools) {
+        $name = $tool.mcpToolName
+        $command = $tool.commandLine
+        $description = $tool.description
+
+        $description = $description -replace '\|', '\|'
+        
+        # Skip commands (not tools)
+        if ([string]::IsNullOrEmpty($name)) {
```
praveenkuttappan: Let's not skip the commands without a tool name. There are few CLI commands without a MCP tool that we still want to be listed in doc.

praveenkuttappan (approved):
One suggestion to include CLI command without MCP tool in the doc. Otherwise looks fine to me.

tools/azsdk-cli/docs/mcp-commands.md
```diff
@@ -0,0 +1,37 @@
+# Tools available in Azure SDK MCP server
+
+This document provides a comprehensive list of all MCP (Model Context Protocol) tool commands supported by the Azure SDK MCP server version 0.5.9.
+
+## Tools list
+
+| Command | Description |
+|---------|-------------|
+| azsdk azp analyze | Analyze azure pipeline for failures. Set analyzeWithAgent to false unless requested otherwise by the user |
```
weshaggard: Should we format as code? Don't we also want to include the mcp-tool name?

tools/azsdk-cli/auto-documentation.yml
```diff
@@ -0,0 +1,37 @@
...
+    - task: PowerShell@2
+      displayName: 'Generate MCP tools documentation'
+      inputs:
+        targetType: 'filePath'
+        filePath: '$(System.DefaultWorkingDirectory)/tools/azsdk-cli/update-mcp-tools-docs.ps1'
+        arguments: '-AzsdkExePath $(System.DefaultWorkingDirectory)/azsdk'
+        pwsh: true
+        workingDirectory: '$(System.DefaultWorkingDirectory)'
+    
+    - template: /eng/common/pipelines/templates/steps/create-pull-request.yml
```
weshaggard, line 35: Don't we plan to use the same PR as the version update?

tools/azsdk-cli/docs/mcp-commands.md
jeo02: No that was a misunderstanding and it's back to including the tools

tools/azsdk-cli/auto-documentation.yml
jeo02: Originally yes but realized you can't pass a custom job down and run it inside a task. Therefore, we wouldn't be able to run the job before that PR is made.

weshaggard: If we compute the branch name the same for both and use that branch name each job will update that same branch and thus the same PR.

tools/azsdk-cli/update-mcp-tools-docs.ps1
```diff
@@ -0,0 +1,120 @@
...
+"@
+    
+    foreach ($tool in $mcpTools) {
+        $name = $tool.mcpToolName
+        $command = $tool.commandLine
+        $description = $tool.description
+
+        $description = $description -replace '\|', '\|'
+        
+        $markdown += "| $name | $command | $description |`n"
```
weshaggard, line 90:
```suggestion
        $markdown += "| $name | ``$command`` | $description |`n"
```
If we use quotes around it might make the command line stick out a little better.

tools/azsdk-cli/docs/mcp-tools.md
```diff
@@ -1,50 +1,50 @@
...
+| azsdk_run_typespec_validation | azsdk tsp validate | Run TypeSpec validation. Provide absolute path to TypeSpec project root as param. This tool runs TypeSpec validation and TypeSpec configuration validation. |
+| azsdk_typespec_check_project_in_public_repo | azsdk tsp check-public-repo | Check if TypeSpec project is in public spec repo. Provide absolute path to TypeSpec project root as param. |
+| azsdk_typespec_init_project | azsdk tsp init | Use this tool to initialize a new TypeSpec project. Returns the path to the created project. |
+| azsdk_update_language_exclusion_justification |  | Update language exclusion justification in release plan work item. This tool is called to update justification for excluded languages in the release plan. Optionally pass a language name to explicitly request exclusion for a specific language. |
+| azsdk_update_sdk_details_in_release_plan |  | Update the SDK details in the release plan work item. This tool is called to update SDK language and package name in the release plan work item. sdkDetails parameter is a JSON of list of SDKInfo and each SDKInfo contains Language and PackageName as properties. |
```
weshaggard, line 48: Do we not have any cli commands that don't have mcptools linked?

jeo02: yea you're right, was using the wrong .exe to run the script should have them now.

tools/azsdk-cli/docs/mcp-tools.md
```diff
@@ -1,50 +1,57 @@
...
-| azsdk_customized_code_update | Update customized TypeSpec-generated client code |
-| azsdk_typespec_check_project_in_public_repo | Check if TypeSpec project is in public spec repo. Provide absolute path to TypeSpec project root as param. |
-| azsdk_update_language_exclusion_justification | Update language exclusion justification in release plan work item. This tool is called to update justification for excluded languages in the release plan. Optionally pass a language name to explicitly request exclusion for a specific language. |
-| azsdk_update_sdk_details_in_release_plan | Update the SDK details in the release plan work item. This tool is called to update SDK language and package name in the release plan work item. sdkDetails parameter is a JSON of list of SDKInfo and each SDKInfo contains Language and PackageName as properties. |
-| azsdk_verify_setup | Verifies that the user's environment is setup with the necessary installations to use SDK MCP tools. |
\ No newline at end of file
+| Name | Command | Description |
+|------|---------|-------------|
+| azsdk_analyze_log_file | `azsdk azp log analyze` | Analyzes a log file for errors and issues |
```
weshaggard, line 9: Image

we should see if we can make the command not wrap if possible.

weshaggard (approved):

tools/azsdk-cli/docs/mcp-tools.md
jeo02: https://github.com/Azure/azure-sdk-tools/issues/13212
Made an issue for it

Issue #13080, opened 2025-11-26
Remove package name approval on APIView for .NET management plane.


maririos:
Remove package name approval on APIView for .NET management plane.

This requirement was introduced last year [introduced as a bug](https://github.com/Azure/azure-sdk-for-net/pull/43599/files) and we have confirmed with Arthur Ma that his check is no needed.

praveenkuttappan:
Fixed in https://github.com/Azure/azure-sdk-for-net/pull/54283

PR #13131, opened 2025-12-02
Skip namespace approval check for mgmt plane

praveenkuttappan:
Package name approval check is not required for management plane. But a bug was introduced in https://github.com/Azure/azure-sdk-for-net/pull/43599/files and it mandated package name approval for .NET.

Fixes issue #https://github.com/Azure/azure-sdk-tools/issues/13080

### DIFF
--- a/eng/common/scripts/Create-APIReview.ps1
+++ b/eng/common/scripts/Create-APIReview.ps1
@@ -283,7 +283,7 @@ function ProcessPackage($packageInfo)
                     # Check if package name is approved. Preview version cannot be released without package name approval
                     if (!$pkgNameStatus.IsApproved)
                     {
-                        if (IsApiviewStatusCheckRequired $packageInfo)
+                        if ($packageInfo.SdkType -eq "client" -and $packageInfo.IsNewSdk)
                         {
                             Write-Error $($pkgNameStatus.Details)
                             return 1


eng/common/scripts/Create-APIReview.ps1
```diff
@@ -283,7 +283,7 @@ function ProcessPackage($packageInfo)
                     # Check if package name is approved. Preview version cannot be released without package name approval
                     if (!$pkgNameStatus.IsApproved)
                     {
-                        if (IsApiviewStatusCheckRequired $packageInfo)
+                        if ($packageInfo.SdkType -eq "client" -and $packageInfo.IsNewSdk)
```
weshaggard, line 286: Why this change? The default condition already matches this.

weshaggard: Shouldn't we update this in the .NET repo? https://github.com/Azure/azure-sdk-for-net/pull/43599/files

praveenkuttappan: .NET management plane requires apiview approval for GA but it does not require package name approval using APIView for mgmt plane. .NET is the only language that requires apiview approval for GA release.

 .NET repo has overridden the fn `IsApiviewStatusCheckRequired` and return true if package type is client or mgmt.

weshaggard: Should we instead pass the release type to the function so the language can still decide instead of us hard-coding it in the common script?

praveenkuttappan: Submitted a PR ion .NET repo to handle this in language side. https://github.com/Azure/azure-sdk-for-net/pull/54283

praveenkuttappan:
Submitted https://github.com/Azure/azure-sdk-for-net/pull/54283 to fix the issue

PR #13136, opened 2025-12-02
[AVC] Changed run_evals signature to make -p flag truly optional

marcarro:
# Running all tests

### Before
`python cli.py eval run -p`

### After
`python cli.py eval run`

## Cause
`-p` was required even with no value passed because `run_evals()` did not provide a default argument for what's supposed to be an optional one.

### DIFF
--- a/packages/python-packages/apiview-copilot/cli.py
+++ b/packages/python-packages/apiview-copilot/cli.py
@@ -180,10 +180,12 @@ def _local_review(
     reviewer.close()
 
 
-def run_evals(test_paths: list[str] = [], num_runs: int = 1, save: bool = False, use_recording: bool = False):
+def run_evals(test_paths: list[str] = None, num_runs: int = 1, save: bool = False, use_recording: bool = False):
     """
     Runs the specified test case(s).
     """
+    if test_paths is None:
+        test_paths = []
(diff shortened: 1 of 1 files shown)


packages/python-packages/apiview-copilot/cli.py
```diff
@@ -180,10 +180,12 @@ def _local_review(
     reviewer.close()
 
 
-def run_evals(test_paths: list[str], num_runs: int = 1, save: bool = False, use_recording: bool = False):
+def run_evals(test_paths: list[str] = None, num_runs: int = 1, save: bool = False, use_recording: bool = False):
     """
     Runs the specified test case(s).
     """
+    if test_paths is None:
+        test_paths = []
```
tjprescott, line 188: How does this work if you are in some weird directory? I feel like it would just start crawling everything from CWD. Might make more sense to start at `evals\tests` if you don't provide anything.

marcarro: `_discovery.py` defaults to `evals/tests` when no test paths are provided

[Here specifically](https://github.com/Azure/azure-sdk-tools/blob/main/packages/python-packages/apiview-copilot/evals/_discovery.py#L55)

PR #13169, opened 2025-12-03
[http-fault-injector] Update deps of Java samples

mikeharder:
- floats dep versions, so should not require further updates


tools/http-fault-injector/sample-clients/java/http-fault-injector-client/pom.xml
mikeharder, line 1: copied simpler pom.xml from storage-blobs example

tools/http-fault-injector/sample-clients/java/storage-blobs/src/main/java/FaultInjectorHttpClient.java
```diff
@@ -28,7 +28,7 @@ public final class FaultInjectorHttpClient implements HttpClient {
      */
     public FaultInjectorHttpClient(HttpClient httpClient) {
         this(httpClient, Utils.DEFAULT_HTTP_FAULT_INJECTOR_SCHEME,
-             Util.DEFAULT_HTTP_FAULT_INJECTOR_HOST, Utils.DEFAULT_HTTP_FAULT_INJECTOR_HTTP_PORT);
+             Utils.DEFAULT_HTTP_FAULT_INJECTOR_HOST, Utils.DEFAULT_HTTP_FAULT_INJECTOR_HTTP_PORT);
```
mikeharder, line 31: unrelated bugfix

PR #13208, opened 2025-12-04
Update Group in the work items

raych1:
Only update the script partially in order to make the sync script work from release csv to dev-ops work items.


eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1
```diff
@@ -253,6 +253,7 @@ function FindPackageWorkItem($lang, $packageName, $version, $outputCommand = $tr
   $fields += "System.Tags"
   $fields += "Custom.Language"
   $fields += "Custom.Package"
+  $fields += "Custom.Group"
```
weshaggard, line 256: I assume this is groupId now correct?

PR #13218, opened 2025-12-05
Add script for local tool description eval workflow

smw-ms:
Resolves issue: https://github.com/Azure/azure-sdk-tools/issues/13148

### DIFF
--- a/eng/common/scripts/Invoke-ToolDescriptionEvaluator.ps1
+++ b/eng/common/scripts/Invoke-ToolDescriptionEvaluator.ps1
@@ -1,47 +1,50 @@
 <#
 .SYNOPSIS
-Evaluates tool and prompt descriptions and produces a Markdown results report using Tool Description Evaluator from Mcp repo.
+    Evaluates tool and prompt descriptions and produces a Markdown results report using Tool Description Evaluator from Mcp repo.
 
 .DESCRIPTION
-This script builds and runs the ToolDescriptionEvaluator (.NET) against a set of tool and prompt definitions.
-It restores and compiles the evaluator, executes it with the provided JSON inputs, and emits a `results.md`
-report. The script supports configuration via parameters and environment variables for the embedding model
-service used during evaluation.
+    This script builds and runs the ToolDescriptionEvaluator (.NET) against a set of tool and prompt definitions.
+    It restores and compiles the evaluator, executes it with the provided JSON inputs, and emits a `results.md`
+    report. The script supports configuration via parameters and environment variables for the embedding model
+    service used during evaluation.
+
+.LINK
+    https://github.com/microsoft/mcp/tree/main/eng/tools/ToolDescriptionEvaluator
 
 .PARAMETER EvaluatorPath
-The path to the evaluator project root (or its `src` directory) that will be restored, built, and executed.
+    The path to the evaluator project root (or its `src` directory) that will be restored, built, and executed.
 
 .PARAMETER ToolsFilePath
-The path to the JSON file containing tool definitions to be evaluated.
+    The path to the JSON file containing tool definitions to be evaluated.
 
 .PARAMETER PromptsFilePath
-The path to the JSON file containing prompt definitions to be evaluated.
+    The path to the JSON file containing prompt definitions to be evaluated.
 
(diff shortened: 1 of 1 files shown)


eng/common/scripts/Invoke-ToolDescriptionEvaluator.ps1
```diff
@@ -0,0 +1,99 @@
...
+.PARAMETER EvaluatorPath
+The path to the evaluator project root (or its `src` directory) that will be restored, built, and executed.
+
+.PARAMETER ToolsFilePath
+The path to the JSON file containing tool definitions to be evaluated.
+
+.PARAMETER PromptsFilePath
+The path to the JSON file containing prompt definitions to be evaluated.
+
+.PARAMETER outputFilePath
```
jeo02:
```suggestion
.PARAMETER OutputFilePath
```

eng/common/scripts/Invoke-ToolDescriptionEvaluator.ps1
```diff
@@ -0,0 +1,99 @@
...
+The target folder where the generated `results.md` will be moved after the evaluator runs.
+
+.PARAMETER AoaiEndpoint
+The full endpoint URL for the embedding model (e.g., Azure OpenAI embeddings) used by the evaluator.
+
+.PARAMETER TextEmbeddingApiKey
```
jeo02, line 29: Do we need to use the an API key for the embedding model or can we use RBAC to access the model.
If there's a model in the shared open ai playground resource for embedding RBAC is already setup and there wouldn't be a need for the key.

jeo02: I assume the tool description tool requires it though

smw-ms: yes the tool description eval tool requires these two params

eng/common/scripts/Invoke-ToolDescriptionEvaluator.ps1
smw-ms: fixed

eng/common/scripts/Invoke-ToolDescriptionEvaluator.ps1
jeo02: If we plan to do this from the pipeline we will have to store the key in a key vault

eng/common/scripts/Invoke-ToolDescriptionEvaluator.ps1
```diff
@@ -0,0 +1,99 @@
...
+
+# Build & run the evaluator
+Write-Host "Changing directory to evaluator: $EvaluatorPath"
+Push-Location $EvaluatorPath
+try { 
+    $env:AOAI_ENDPOINT = $AoaiEndpoint
+    $env:TEXT_EMBEDDING_API_KEY = $TextEmbeddingApiKey
+
+    Write-Host "Restoring packages..."
+    dotnet restore
```
jeo02: Why do we need to dotnet restore?

jeo02 (changes requested):
Looks good overall would just go through the copilot suggestions and see what's useful then I'll give it a once over and approve.

eng/common/scripts/Invoke-ToolDescriptionEvaluator.ps1
```diff
@@ -0,0 +1,99 @@
+<#
+.SYNOPSIS
+Evaluates tool and prompt descriptions and produces a Markdown results report using Tool Description Evaluator from Mcp repo.
```
benbp: I'd add some references to the actual azsdk tool in the synopsis and/or other places, otherwise without context this is pretty mysterious.

smw-ms: fixed

eng/common/scripts/Invoke-ToolDescriptionEvaluator.ps1
smw-ms: we don't really. removed

jeo02 (approved):

PR #13221, opened 2025-12-05
Add support for authenticating to an Azure DevOps feed outside of organization

conniey:
* Add support for authenticating to feeds in a different Azure DevOps organization
* Based on documentation from: https://eng.ms/docs/coreai/devdiv/one-engineering-system-1es/1es-docs/azure-artifacts/managed-identity-authentication-tasks#npmauthenticatev0

### DIFF
--- a/eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
+++ b/eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
@@ -6,6 +6,9 @@ parameters:
   - name: CustomCondition
     type: string
     default: succeeded()
+  - name: ServiceConnection
+    type: string
+    default: ''
 
 steps:
 - pwsh: |
@@ -21,8 +24,10 @@ steps:
     $content | Out-File '${{ parameters.npmrcPath }}'
   displayName: 'Create .npmrc'
   condition: ${{ parameters.CustomCondition }}
+
 - task: npmAuthenticate@0
   displayName: Authenticate .npmrc
   condition: ${{ parameters.CustomCondition }}
   inputs:
     workingFile: ${{ parameters.npmrcPath }}
+    azureDevOpsServiceConnection: ${{ parameters.ServiceConnection }}


weshaggard:
While I'm ok with the extensibility here I'm curious what other feed you are trying to connect to? In particular we are soon going to be forced to use only a single feed so we will not be able to connect to different ones.

eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
```diff
@@ -21,8 +24,10 @@ steps:
     $content | Out-File '${{ parameters.npmrcPath }}'
   displayName: 'Create .npmrc'
   condition: ${{ parameters.CustomCondition }}
+
 - task: npmAuthenticate@0
   displayName: Authenticate .npmrc
   condition: ${{ parameters.CustomCondition }}
   inputs:
     workingFile: ${{ parameters.npmrcPath }}
+    azureDevOpsServiceConnection: ${{ parameters.ServiceConnection }}
```
benbp, line 33: Does this still work if it's empty, i.e. should there be a conditional for the if the parameter is defined or not?

conniey:
> While I'm ok with the extensibility here I'm curious what other feed you are trying to connect to? In particular we are soon going to be forced to use only a single feed so we will not be able to connect to different ones.

1ES has an internal MCP registry that internal folks are allowed to consume.  In order to show up on the list of available MCP servers, we need to publish to that registry.

eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
conniey: The default value for the parameter is ''.  From the documentation, if we only need "need registries in this organization or collection, leave this blank".  
We have an instance where it is empty when publishing to our dev feed: https://github.com/microsoft/mcp/pull/1430/changes#diff-656f8890507cc25f7c89da3df58428cc8e5323d76f144b7c68aacf308b01e5dfL33

So I don't forget where the documentation is because I have so many tabs open...
- [npmAuthenticate@0](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/npm-authenticate-v0?view=azure-pipelines)
- [npmAuthenticate@0 custom](https://eng.ms/docs/coreai/devdiv/one-engineering-system-1es/1es-docs/azure-artifacts/managed-identity-authentication-tasks#npmauthenticatev0)

benbp (approved):

PR #13222, opened 2025-12-05
LogGroupStart should have local build output

hallipr:
The script call `Invoke-LoggedMsBuildCommand $command -GroupOutput` should log the command before logging its output.
When run locally, `LogGroupStart` logs nothing, so the command isn't logged.

### DIFF
--- a/eng/common/scripts/logging.ps1
+++ b/eng/common/scripts/logging.ps1
@@ -94,6 +94,9 @@ function LogGroupStart() {
   elseif (Test-SupportsGitHubLogging) {
     Write-Host "::group::$args"
   }
+  else {
+    Write-Host "> $args"
+  }
 }
 
 function LogGroupEnd() {


eng/common/scripts/logging.ps1
```diff
@@ -94,6 +94,9 @@ function LogGroupStart() {
   elseif (Test-SupportsGitHubLogging) {
     Write-Host "::group::$args"
   }
+  else {
+    Write-Host "> $args"
+  }
```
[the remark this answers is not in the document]
hallipr: In local runs, there's no reason to log the end of the group.  The header isn't repeated in the tail

eng/common/scripts/logging.ps1
```diff
@@ -94,6 +94,9 @@ function LogGroupStart() {
   elseif (Test-SupportsGitHubLogging) {
     Write-Host "::group::$args"
   }
+  else {
+    Write-Host "> $args"
```
weshaggard, line 98: If you add this then you should probably alter https://github.com/Azure/azure-sdk-tools/blob/b1f31f9dfdac2fde3bf57907faabc5cb1818b25c/eng/common/scripts/Helpers/CommandInvocation-Helpers.ps1#L35

hallipr: It already accounts for that:
```
    if($GroupOutput) {
        LogGroupStart $Command
    } else {
        Write-Host "> $Command"
    }
```

We either log it as a grouped header or just a simple `>` line.
If we logged it as a grouped header on a local build, it will now still come out as a simple `>` line

Issue #13004, opened 2025-11-19
[APIView] Lightweight App Permissions


tjprescott:
Today, APIView has permissions established through configuration options, mostly the "approvers" list, which is self-select.

However, we have UI components today that we would like to show/hide or enable/disable based on the role of the user, but we have no clear, formal means of designating user roles.

I'm proposing some kind of lightweight RBAC system whereby we can designate app permissions to users and/or groups of users and use that consistently to decide what functionality is available to users.

**User Roles**
In order of how permissive each role is
- Service Team
- SDK Team
- Deputy Architect - similar permissions to architect 
- Architect - elevated permissions when only an architect will do
- Admin - highest level permissions, only for APIView maintainers

Roles could be scoped to languages. For example, Johan would be an SDK Team member generally (for all languages) but the Architect for Python. So this would mean Johan could do whatever an SDK Team member could generally do in APIView for any language, but would have Archtect-level permissions for Python. However, we would not be able to use admin-level features for any language.

This would also resolve weird things like how we currently treat anyone in the "approvers" list as an architect, which is sloppy modeling.

tjprescott:
The first order of business here would be to propose how this would be structured in TypeSpec.

tjprescott:
@AlitzelMendez's proposal

---

## 1. Problem

Currently, APIView has a simple permissions model based on a configuration-level "approvers" list (`configuration["approvers"]`). This approach has several limitations:

1. **No formal role hierarchy** - Anyone in the approvers list gets the same elevated permissions
2. **No language scoping** - Users can't be designated as architects for specific languages
3. **Sloppy modeling** - The current system treats anyone in the approvers list as an architect
4. **Limited UI control** - No clear mechanism to show/hide features based on user roles
5. **Self-select nature** - The approvers list is self-administered

---

## 2. Proposed User Roles

Roles are divided into two categories based on their scope:

### Global Roles
These roles apply to all languages and cannot be language-scoped:

| Role | Description |
|------|-------------|
| **ServiceTeam** | Service team members who create and manage reviews |
| **SdkTeam** | SDK team members with additional privileges |
| **Admin** | Highest level permissions, for APIView maintainers only |

### Language-Scoped Roles
These roles **must** be assigned to a specific language:

| Role | Description |
|------|-------------|
| **DeputyArchitect** | Similar permissions to Architect, for designated backups |
| **Architect** | Elevated permissions for API review approval |

---

## 3. Permission Matrix

[...]

PR #13231, opened 2025-12-05
[APIView] Proposal Lightweight App Permissions

AlitzelMendez:
Proposal for: https://github.com/Azure/azure-sdk-tools/issues/13004

### DIFF
--- a/src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
+++ b/src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
@@ -213,18 +213,13 @@ export class PermissionsService {
         return this.hasGlobalRole(permissions, "Admin") ||
                this.hasLanguageRole(permissions, ["Architect", "DeputyArchitect"], language);
     }
-    
-    /** Check if user has elevated global permissions (SdkTeam or Admin) */
-    hasElevatedAccess(permissions: EffectivePermissions): boolean {
-        return this.hasGlobalRole(permissions, ["Admin"]);
-    }
 }
 ```
 
 **Usage examples:**
 ```typescript
 // Check for a single role
-this.permissionsService.hasGlobalRole(permissions, "SdkTeam");
+this.permissionsService.hasGlobalRole(permissions, "Admin");
 
 // Check for multiple roles at once
 this.permissionsService.hasGlobalRole(permissions, ["SdkTeam", "Admin"]);


src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,136 @@
...
+| **SdkTeam** | SDK team members with additional privileges | Not language-specific |
+| **DeputyArchitect** | Similar permissions to Architect, for designated backups | Per-language |
+| **Architect** | Elevated permissions for API review approval | Per-language |
+| **Admin** | Highest level permissions, for APIView maintainers only | Not language-specific |
+
+---
+
+## 3. Permission Matrix
+
+| Action | ServiceTeam | SdkTeam | DeputyArchitect | Architect | Admin |
```
tjprescott, line 43: Service Team and SDK Team seem to have the same permissions, so should they just be merged? Same with Deputy Architect and Architect, unless we create some distinction.

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,136 @@
...
+|--------|:-----------:|:-------:|:---------------:|:---------:|:-----:|
+| View reviews | ✓ | ✓ | ✓ | ✓ | ✓ |
+| Create reviews | ✓ | ✓ | ✓ | ✓ | ✓ |
+| Add comments | ✓ | ✓ | ✓ | ✓ | ✓ |
+| Delete own comments | ✓ | ✓ | ✓ | ✓ | ✓ |
+| Delete any comment | | | ✓ | ✓ | ✓ |
```
tjprescott, line 49: To be clear this means "Delete any AI comment". It should probably be language scoped as well.

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,136 @@
...
+| Manage user permissions | | | | | ✓ |
+| Access admin features | | | | | ✓ |
+
+*\* Language-scoped: Only for languages where user has this role*
+
+---
+
+## 4. Data Storage
+
+Create a new `Permissions` container in Cosmos DB:
```
tjprescott: Why not simply store these in the User Profile data? This info would be needed to render the appropriate UI and probably also *should* be used on the backend to make sure that calls are only made on behalf of authorized users.

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,136 @@
...
+    { "role": "Architect", "language": "Python" }
+  ],
+  "lastUpdatedOn": "2024-12-04T00:00:00Z",
+  "lastUpdatedBy": "admin"
+}
+```
+
+**Group Permission Document:**
+```json
+{
```
tjprescott: Are you thinking to do some kind of AD checks to automatically form these groups?  This could be nice since it would allow us to just rely on AD.

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+**Example**: User has direct `SdkTeam` (global) + group membership gives `Architect` (Python)
+- `HasRole(Architect, "Python")` → ✅ true
+- `HasRole(Architect, "Java")` → ❌ false  
+- `HasRole(SdkTeam, "Java")` → ✅ true (global applies)
+
+---
+
+## 6. API Summary Table
```
tjprescott: I think we should give thought to what helpers you'd want on the UI and the backend to check user permissions.

tools/apiview/parsers/apiview-treestyle-parser-schema/permissions.tsp
```diff
@@ -0,0 +1,70 @@
...
+    /** Service team members who create and manage reviews */
+    ServiceTeam,
+    /** SDK team members with additional privileges */
+    SdkTeam,
+    /** Deputy architect with architect-like permissions */
+    DeputyArchitect,
+    /** Architect with elevated API review permissions */
+    Architect,
+    /** Admin with full system permissions */
+    Admin
```
tjprescott, line 12: What happens if there are no Admins?

tools/apiview/parsers/apiview-treestyle-parser-schema/permissions.tsp
```diff
@@ -0,0 +1,70 @@
...
+
+/** Represents a role assignment that can be global or language-scoped */
+model RoleAssignment {
+    /** The role being assigned */
+    role: UserRole;
+    /** 
+     * The language this role applies to. 
+     * If null or empty, the role is global (applies to all languages).
+     */
+    language?: string;
```
tjprescott: So if you give someone the Architect role but leave this blank... is that valid as "architect for everything" or would that be an error?

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: that would be another good approach, and I am not against that, I was setting it up in a different place just to facilitate the monitoring part and not to mix the logic with the groups, but at the end on the day this information can be with the profile

tools/apiview/parsers/apiview-treestyle-parser-schema/permissions.tsp
AlitzelMendez: to give Admin permissions?

tools/apiview/parsers/apiview-treestyle-parser-schema/permissions.tsp
AlitzelMendez: it would fail as the scope for Architect and deputy architect is per language, let me add that info to the tsp

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: We talked about this offline and the conclusion was that even though permissions are the same, UI/UX is not going to be necessary the same

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: yes! this is langauges scoped

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: I added this section! :) let me know what you think!

tools/apiview/parsers/apiview-treestyle-parser-schema/permissions.tsp
AlitzelMendez: we discussed this offline

tools/apiview/parsers/apiview-treestyle-parser-schema/permissions.tsp
AlitzelMendez: This was a very interesting question that got me thinking, so I changed how this is defined so it actually enforces the definition of languages when needed. This way, we can trust the information in the database without making incorrect assumptions.

My approach is very strict, ensuring that roles can't be assigned incorrectly. However, we could make it a bit simpler or less strict if you prefer—please take a look! :)

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: no for now and we talked about this offline

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: discussed offline

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,277 @@
...
+{
+    // 1. Get all groups where user is a member
+    var groups = await _permissionRepo.GetGroupsForUserAsync(userId);
+    
+    // 2. Merge group-inherited, highest role wins per language
+    return MergePermissions(groups);
+}
+```
+
+**Example**: User has direct `SdkTeam` (global) + group membership gives `Architect` (Python)
```
tjprescott: Aren't all memberships via groups now? So the user would be in an SdkTeam group and a Python architects group?

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,277 @@
...
+```typescript
+@Injectable({ providedIn: 'root' })
+export class PermissionsService {
+    
+    /** Check if user has a global role */
+    hasGlobalRole(permissions: EffectivePermissions, role: GlobalRole): boolean {
+        return permissions.roles.some(r => 
+            r.kind === "global" && r.role === role
+        );
+    }
```
tjprescott: We might want `role` to accept an array of roles. For example if you wanted to say "This should be accessible to SDK Team and Admins" then you wouldn't want to have to call this twice, once for each role. 

Alternatively, if the global roles are clearly orderable (Admin > SDK Team > Service Team) then we could get away with the role being "at least" this value.

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,277 @@
...
+    
+    /** Check if user has a language-scoped role for a specific language */
+    hasLanguageRole(permissions: EffectivePermissions, role: LanguageScopedRole, language: string): boolean {
+        return permissions.roles.some(r => 
+            r.kind === "scoped" && r.role === role && r.language === language
+        );
+    }
```
tjprescott: Same here. If I want to check that someone is an architect or deputy architect for Python, I don't want to have to call this twice. If role was simply an array, that should suffice.

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,277 @@
...
+    
+    /** Check if user is Admin */
+    isAdmin(permissions: EffectivePermissions): boolean {
+        return this.hasGlobalRole(permissions, "Admin");
+    }
```
tjprescott: Yeeeah, but I would think it would be cleaner to simply call `hasGlobalRole(..., [SDKTeam, Admin])` instead of `hasGlobalRole(..., SDKTeam) || isAdmin(...)`

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,277 @@
...
+    
+    /** Check if user can approve for a specific language */
+    canApprove(permissions: EffectivePermissions, language: string): boolean {
+        return this.isAdmin(permissions) ||
+               this.hasLanguageRole(permissions, "Architect", language) ||
+               this.hasLanguageRole(permissions, "DeputyArchitect", language);
+    }
+}
```
tjprescott: This *might* change... languages might decide that only their architect can approve... dunno. but yeah, it would be basically like this.

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
```diff
@@ -0,0 +1,277 @@
...
+
+### UI Usage Example
+
+```typescript
+// In a component
+if (this.permissionsService.canApprove(this.userProfile.permissions, review.language)) {
+    this.showApproveButton = true;
+}
+
+if (this.permissionsService.isAdmin(this.userProfile.permissions)) {
```
tjprescott: I still prefer just making `hasGlobalRole` accept an array because otherwise, every other conditional is gonna need an `isAdmin` tagged on... OR... I guess we could just assign the admins the `SDKTeam` role as well 🤔

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: yes!! updated!

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: Great point, I went for the rout of role arrays, thinking that 1. SDK Team / Service team are going to be handle with the same permission level and when they will be different, they will be a different/specific UI

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: solved :)

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: I went for array :)

src/dotnet/APIView/docs/apiview-lightweight-permissions-proposal.md
AlitzelMendez: ok I changed this :)

tjprescott:
This looks great. I copied the proposal into issue #13004 so we can use it as the basis for implementation. Thank you @AlitzelMendez!

PR #13253, opened 2025-12-08
Add options metadata to list command output

### LINKED ISSUE #13139: Add options metadata to list command output
https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/ListCommandTool.cs#L52

smw-ms:
Issue: https://github.com/Azure/azure-sdk-tools/issues/13139

New output (azsdk list -o json) : 
[output.json](https://github.com/user-attachments/files/24043975/output.json)

New output (azsdk list):
[output.txt](https://github.com/user-attachments/files/24043984/output.txt)

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/ToolListResponse.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/ToolListResponse.cs
@@ -44,16 +44,16 @@ namespace Azure.Sdk.Tools.Cli.Models.Responses
             sb.AppendLine(
                 $"{("Name").PadRight(nameWidth)}{sep}" +
                 $"{("Command").PadRight(cmdWidth)}{sep}" +
-                $"{("Description").PadRight(descWidth)}{sep}" +
-                $"{("Options").PadRight(optsWidth)}"
+                $"{("Options").PadRight(optsWidth)}{sep}" +
+                $"{("Description").PadRight(descWidth)}"
             );
 
             // --- Underline ---
             sb.AppendLine(
                 $"{new string('-', nameWidth)}{sep}" +
                 $"{new string('-', cmdWidth)}{sep}" +
-                $"{new string('-', descWidth)}{sep}" +
-                $"{new string('-', optsWidth)}"
+                $"{new string('-', optsWidth)}{sep}" +
+                $"{new string('-', descWidth)}"
             );
 
             // --- Rows ---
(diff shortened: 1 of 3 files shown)


smw-ms:
Could update: https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/update-mcp-tools-docs.ps1 as well with new options metadata if that sounds like a good idea

benbp:
Looking at the new output, some of the objects have the command line options (when CLI is defined), and the MCP only ones have the MCP options. I know that it's meant to be a way to find the mcp->cli analog commands, but seems a bit confusing to have the option schema change. Should it just be empty if we don't have a CLI equivalent?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/ToolListResponse.cs
```diff
@@ -16,22 +16,81 @@ protected override string Format()
...
+                $"{("Description").PadRight(descWidth)}{sep}" +
+                $"{("Options").PadRight(optsWidth)}"
+            );
+
+            // --- Underline ---
+            sb.AppendLine(
+                $"{new string('-', nameWidth)}{sep}" +
+                $"{new string('-', cmdWidth)}{sep}" +
+                $"{new string('-', descWidth)}{sep}" +
+                $"{new string('-', optsWidth)}"
```
praveenkuttappan: Can you list the options next to command line? so tool name, command line, options and description

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/ListCommandTool.cs
```diff
@@ -66,7 +67,27 @@ private List<ToolInfo> GetToolNameAndDescription(ParseResult parseResult)
                 if (!tools.Any(t=> t.McpToolName == mcpTool.Name))
                 {
                     // Add new tool that does not have CLI command representation
-                    tools.Add(new ToolInfo(mcpTool.Name, "", mcpTool.Description ?? ""));
+                    var required = new HashSet<string>();
```
praveenkuttappan: Is this info required for MCP tools that does not have any command line?

smw-ms: it is not really required. I agree with @benbp https://github.com/Azure/azure-sdk-tools/pull/13253#issuecomment-3628861628

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/ToolListResponse.cs
smw-ms: done

praveenkuttappan (approved):

benbp:
Forgot to ask earlier, does the options include those inherited from parent comands? I noticed the global options aren't included (debug, output) which is good, but any options from 2nd level parents should probably show up.

smw-ms:
> Forgot to ask earlier, does the options include those inherited from parent comands? I noticed the global options aren't included (debug, output) which is good, but any options from 2nd level parents should probably show up.

Example: For `azsdk pkg samples generate` `--package-path` is from the 2nd level parent (pkg). It shows up in the tool list output because when the command was initialized,    `SharedOptions.PackagePath` was explicitly included as one of its option. I think that is how sub commands inherit from their parent in general, so it should be fine

PR #13258, opened 2025-12-08
Replaced Default Credentials with Token Credential

radhgupta:
## Summary
Removed hardcoded `DefaultAzureCredential` instantiation in `OpenAiAnswerService` and replaced it with dependency-injected `TokenCredential` to align with the existing authentication pattern used throughout the application.

## Changes
- Updated `OpenAiAnswerService` constructor to accept `TokenCredential` parameter
- Updated `AnswerFactory` to pass the injected credential to `OpenAiAnswerService`
- Registered `TokenCredential` as a singleton in the DI container in `Program.cs`

### DIFF
--- a/tools/issue-labeler/src/IssueLabelerService/Program.cs
+++ b/tools/issue-labeler/src/IssueLabelerService/Program.cs
@@ -23,7 +23,13 @@ var host = new HostBuilder()
         var isRunningInAzure = functionConfig["IsRunningInAzure"] == "true";
 
         // Use appropriate credential based on environment
-        TokenCredential credential = isRunningInAzure ? new ManagedIdentityCredential() : new DefaultAzureCredential();
+        TokenCredential credential = isRunningInAzure 
+            ? new ManagedIdentityCredential()
+            : new ChainedTokenCredential(
+                new AzureCliCredential(),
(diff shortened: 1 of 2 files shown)


tools/issue-labeler/src/IssueLabelerService/Program.cs
```diff
@@ -71,6 +71,7 @@
             return new SearchIndexClient(searchEndpoint, credential);
         });
 
+        services.AddSingleton<TokenCredential>(credential);
```
weshaggard, line 80: You also need to update the usage of DefaultAzureCredential above this function.

radhgupta: This line is registering the credential from line 26 into the DI container for dependency injection. In production (IsRunningInAzure = true), this will be `ManagedIdentityCredential`, not `DefaultAzureCredential`. 
The DefaultAzureCredential is only used locally for development convenience.  Do we still need to remove it??

weshaggard: Yes we should still be specific about the credential in all cases.

weshaggard (approved):

jeo02 (approved):

Issue #10239, opened 2025-04-08
[Python APIView] add blank line to class without methods


swathipil:
For example, CredentialUnavailableError doesn't have any child methods because it only inherits from the azure.core ClientAuthException, and inherited methods are not shown. https://github.com/Azure/azure-sdk-for-python/blob/8db67675b45d7792e51894ec6ddc01ed1858afba/sdk/identity/azure-identity/azure/identity/_exceptions.py#L10

![Image](https://github.com/user-attachments/assets/c45583b4-d2f8-4e18-aad1-273c94ee68c3)

This class should still have two blank lines as a child lines.

PR #13259, opened 2025-12-08
[Python APIView] clean up APIView to align closer to Python syntax

swathipil:
Moved out the token json -> md file logic to a script in eng/common/scripts as per discussion with Travis. (PR #14344)
fixes: #10239

In this PR:
- Adding `: ...` to function definitions to bring API closer to Python syntax.
- Fixing no blank lines after empty class bug.

### DIFF
--- a/packages/python-packages/apiview-stub-generator/CHANGELOG.md
+++ b/packages/python-packages/apiview-stub-generator/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Version 0.3.27 (Unreleased)
 Removed `ivar`/`cvar` keywords from token file in favor of the `ClassVar` type annotation to distinguish then.
 Fixed a bug where classes without children were not being post-fixed with blank lines.
+Added a `:...` after function definitions to align closer to Python syntax.
 
 ## Version 0.3.26 (2026-02-20)
 Fixed a bug where classes with no custom name attribute cause errors.


packages/python-packages/apiview-stub-generator/apistub/scripts/Export-APIViewMarkdown.ps1
```diff
@@ -0,0 +1,155 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+
+<#
+.SYNOPSIS
```
tjprescott: @weshaggard @scbedd this script is in `eng/common` but it being duplicated here. What options exist so that we would not need to do this?

weshaggard: We should use the one in eng/common here if it is an exact copy.

scbedd: Agreed. We're in the same repo so this won't have to be very stilted at least.

swathipil: removed

PR #13267, opened 2025-12-08
Updated artifactJson input to artifacts input for set-test-pipeline-version pipeline

raych1:
To allow iterating artifact array and filter on the value in the callers.

### DIFF
--- a/eng/common/scripts/SetTestPipelineVersion.ps1
+++ b/eng/common/scripts/SetTestPipelineVersion.ps1
@@ -6,6 +6,8 @@ param (
   [Parameter(mandatory = $false)]
   [string]$PackageNames = "",
   [Parameter(mandatory = $false)]
+  # ServiceDirectory is required when using PackageNames,
+  # or when Artifacts do not include their own ServiceDirectory property.
   [string]$ServiceDirectory,
   [Parameter(mandatory = $false)]
   [string]$TagSeparator = "_",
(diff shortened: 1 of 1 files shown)


eng/common/scripts/SetTestPipelineVersion.ps1
```diff
@@ -5,63 +5,77 @@ param (
...
+      if ([String]::IsNullOrWhiteSpace($packageName)) {
+        LogError "Artifact 'name' property is null or empty."
+        exit 1
+      }
+
+      # Check for ServiceDirectory property
+      $artifactServiceDirectory = $ServiceDirectory
+      if (Get-Member -InputObject $artifact -Name 'ServiceDirectory' -MemberType Properties) {
+        if (![String]::IsNullOrWhiteSpace($artifact.ServiceDirectory)) {
+          $artifactServiceDirectory = $artifact.ServiceDirectory
```
weshaggard, line 48: Should we add validation to ensure the service directory is the same for each artifact?

weshaggard: Nevermind that shouldn't be needed given we are processing them one at a time

raych1: We cannot do that because the patch release pipeline would include packages from different service directories. 

With this approach, I can make the patch release pipeline work for both template and template-v2 projects.

Issue #10732, opened 2025-05-23
API View - Last Updated column is not accurate


JoshLove-msft:
The Last Updated should show the last time anything was updated on the review - including a new revision being added. Currently it only refers to when an update to the review entity itself was made. In the screenshot below, a new revision was added for each of the Generator libraries the same day as the screenshot being taken:

![Image](https://github.com/user-attachments/assets/3e06dc08-64a4-4953-9322-1e3127c8aead)

PR #13280, opened 2025-12-09
Fix: Sync review LastUpdatedOn timestamp with revision updates

helen229:
Close: https://github.com/Azure/azure-sdk-tools/issues/10732

**Problem**

The "Last Updated" field in APIView was showing when the review entity was last modified, not when any revision activity occurred. This caused reviews to appear stale even when new revisions were added.

**Solution Implemented**

Added UpdateParentReviewLastUpdatedOnAsync() method that:
Reads the parent review when a revision is upserted
Compares timestamps: only updates if revision is newer than review
Handles missing reviews gracefully (catches NotFound exception)
Modified UpsertAPIRevisionAsync() to call the update method after saving revision

This fix only affects new changes to revisions; existing revisions won't be affected and resorted

Before:

image

image

After:

image

### DIFF
--- a/src/dotnet/APIView/APIViewUnitTests/APIViewUnitTests.csproj
+++ b/src/dotnet/APIView/APIViewUnitTests/APIViewUnitTests.csproj
@@ -19,7 +19,7 @@
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="6.10.0" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="8.0.0" />
-    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.40.0" />
+    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.47.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="8.0.0" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.13.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />


src/dotnet/APIView/APIViewWeb/Repositories/CosmosAPIRevisionsRepository.cs
```diff
@@ -19,10 +19,12 @@ namespace APIViewWeb
     public class CosmosAPIRevisionsRepository : ICosmosAPIRevisionsRepository
     {
         private readonly Container _apiRevisionContainer;
+        private readonly Container _reviewsContainer;
 
         public CosmosAPIRevisionsRepository(IConfiguration configuration, CosmosClient cosmosClient)
         {
             _apiRevisionContainer = cosmosClient.GetContainer(configuration["CosmosDBName"], "APIRevisions");
+            _reviewsContainer = cosmosClient.GetContainer(configuration["CosmosDBName"], "Reviews");
```
AlitzelMendez: There is a cosmosReviewsRepository class: https://github.com/Azure/azure-sdk-tools/blob/main/src/dotnet/APIView/APIViewWeb/Repositories/CosmosReviewRepository.cs 

All the updates should be happening there, so you should move the `UpdateParentReviewLastUpdatedOnAsync` method to that class and call it here, that will also reduce the duplicate implementation of  `public async Task<ReviewListItemModel> GetReviewAsync(string reviewId)
`

src/dotnet/APIView/APIViewUnitTests/CosmosAPIRevisionsRepositoryTests.cs
```diff
@@ -0,0 +1,199 @@
...
+        VerifyReviewRead(TestReviewId);
+        VerifyReviewUpdated(TestReviewId, review.LastUpdatedOn);
+    }
+
+    [Fact]
+    public async Task UpsertRevision_SkipsParentReview_WhenOlder()
+    {
+        // Arrange
+        var revision = CreateTestRevision();
+        var review = CreateTestReview(TestReviewId, DateTime.UtcNow); // Current time (newer)
```
AlitzelMendez: what is, DateTime.UtcNow) comparead too? it is preferible to have a specific date (you can have a last year date and a new year date, or any date for being specified) on test to avoid racing conditions that could cause a flaky test

helen229: I made two separate changes for unit tests and use hardcoded date for comparison logic.

Feature validated: Parent review's LastUpdatedOn updates only when revision is newer
CosmosAPIRevisionsRepositoryTests → Verifies the call is made
CosmosReviewRepositoryTests → Verifies the comparison logic works correctly

AlitzelMendez (approved):

PR #13304, opened 2025-12-10
Update copilot instructions to support both local and pipeline based SDK gen

praveenkuttappan:
Added prompt in SDK generation workflow so user can choose to generate SDK locally or use a pipeline to generate the SDK. Local SDK generation instruction document also has updates to include additional steps like validation, test and update metadata file.

### DIFF
--- a/eng/common/instructions/azsdk-tools/local-sdk-workflow.instructions.md
+++ b/eng/common/instructions/azsdk-tools/local-sdk-workflow.instructions.md
@@ -71,6 +71,7 @@ High level steps involved:
 
 - Check if the provided repository path exists and matches the selected SDK language repository.
 - If invalid → prompt user to re-enter a valid path.
+- Verify setup to ensure all required tools are installed for the selected language.
 
 ---
 


eng/common/instructions/azsdk-tools/local-sdk-workflow.instructions.md
raych1, line 107: Part B title needs to be updated which previously contained build step only. Or we might remove the `Part` layer.

eng/common/instructions/azsdk-tools/local-sdk-workflow.instructions.md
```diff
@@ -117,3 +123,29 @@ Help the user generate and build SDKs locally from TypeSpec API specifications u
...
+
+### Step 4: Run package tests
+
+**Actions**:
+
+- Run `azsdk_package_run_tests` MCP tool to run tests on the generated SDK package in the identified project directory.
+
+---
+
+### Step 5: Update change log, metadata and version
```
raych1, line 146: Do we want to split these tools in separate steps? Not sure which approach is preferred for the agent, but the point is the execution sequence of these tools needs to remain fixed.

eng/common/instructions/azsdk-tools/typespec-to-sdk.instructions.md
```diff
@@ -17,8 +17,28 @@ Pre-requisites:
...
 
 ## Step: Generate SDKs
-**Goal**: Generate SDKs
-**Message to user**: "SDK generation will take approximately 15-20 minutes. Currently, SDKs are generated using the Azure DevOps pipeline. SDK generation is supported only from a merged API spec or from an API spec pull request in the https://github.com/Azure/azure-rest-api-specs repository."
+
+As a first step, you must prompt the user to understand the intention of SDK generation. Based on the user input, you can either run SDK generation locally or use SDK generation pipeline.
+
+Generate SDK locally in cases below:
+- If the user wants to walk through each SDK generation step locally. This approach requires user to have the setup for each language to generate SDK locally. User can create a pull request for each language after completing all the steps in SDK generation successfully to get the generated SDK reviewed and merged.
+- If the user wants to troubleshoot a failed SDK generation by the pipeline.
```
raych1: Should we exclude this intension since current implementation of generation & build step are different between the pipeline and MCP tools? This might be valid when we integrate these CLIs to the pipeline.

eng/common/instructions/azsdk-tools/typespec-to-sdk.instructions.md
```diff
@@ -17,8 +17,28 @@ Pre-requisites:
 # SDK generation steps
 
 ## Step: Generate SDKs
-**Goal**: Generate SDKs
-**Message to user**: "SDK generation will take approximately 15-20 minutes. Currently, SDKs are generated using the Azure DevOps pipeline. SDK generation is supported only from a merged API spec or from an API spec pull request in the https://github.com/Azure/azure-rest-api-specs repository."
+
+As a first step, you must prompt the user to understand the intention of SDK generation. Based on the user input, you can either run SDK generation locally or use SDK generation pipeline.
```
raych1, line 21: Should we add detail prompts to the user to collect the user intent? For example asking questions like "do you want to customize the SDK code?" The reason is that the following instructions include similar guidance for evaluating the user intent.

eng/common/instructions/azsdk-tools/typespec-to-sdk.instructions.md
maririos: we could, and maybe instead give other reason like: if the user wants to add more tests, samples, make modifications to the end result?

maririos (approved):

eng/common/instructions/azsdk-tools/local-sdk-workflow.instructions.md
praveenkuttappan: I think it's better to keep it under one main step with substeps in correct order.

eng/common/instructions/azsdk-tools/typespec-to-sdk.instructions.md
praveenkuttappan: Local SDK gen is supported for multiple reasons.  Current instruction is good enough to help the agent to prompt the user. But I will file an issue to add your suggestion to include as part of eval change.

praveenkuttappan: Filed https://github.com/Azure/azure-sdk-tools/issues/13313

eng/common/instructions/azsdk-tools/typespec-to-sdk.instructions.md
praveenkuttappan: yes. I am adding more use cases like adding tests, samples and more customization.

eng/common/instructions/azsdk-tools/local-sdk-workflow.instructions.md
```diff
@@ -5,6 +5,12 @@ description: "Guide the user to generate and build SDKs locally for a TypeSpec b
 # Goal
 
 Help the user generate and build SDKs locally from TypeSpec API specifications using the `azure-sdk-mcp` tools.
+High level steps involved:
+1. Generate SDK locally
+2. Build / Compile SDK locally
+3. Run package checks
+4. Run package tests
+5. Update change log, metadata and version
```
ronniegeraghty, line 13:
```suggestion
High level steps involved:
1. Generate SDK locally
2. Build / Compile SDK locally
3. Run package tests
4. Update change log, metadata and version
5. Run package checks
```
You might need to switch these because come of the check, validate the updates to metadata and docs so they need to be done after those are updated.

eng/common/instructions/azsdk-tools/local-sdk-workflow.instructions.md
```diff
@@ -117,3 +123,29 @@ Help the user generate and build SDKs locally from TypeSpec API specifications u
 **Actions**:
 
 - Run `azsdk_package_build_code` MCP tool to compile the SDK in the identified project directory.
+
+---
+
+### Step 3: Run package validation
```
ronniegeraghty, line 130: Same comment as above, this step needs to come after updating metadata and docs since the checks validate those changes.

eng/common/instructions/azsdk-tools/typespec-to-sdk.instructions.md
```diff
@@ -17,8 +17,28 @@ Pre-requisites:
...
+
+Generate SDK using pipeline:
+- If the user wants to generate SDK and get a PR automatically created in Azure SDK language repositories. This approach can be followed if the user does not want to setup local environment for each language.
+
+### Generate SDK locally:
+
+**Condition**: If user chooses to generate SDK locally
+**Message to user**: "Generating SDKs locally requires you to have the development environment set up for each language. Prompt the user to create a pull request for each language after completing the following steps successfully: generation, validation, build, test and update of metadata, change log, and version."
+**Actions**:
+Follow the steps in #file:local-sdk-workflow.instructions.md to generate and build SDKs locally from TypeSpec project.
```
ronniegeraghty, line 36: Is it worth mentioning the environment validation tool here for the agent to know how to check this?

eng/common/instructions/azsdk-tools/local-sdk-workflow.instructions.md
praveenkuttappan: Main reason to leave the change log and metadata as the last step was to guide the user to run checks and static validations. User won't update change log when just generating SDK for testing spec changes.

raych1: Package checks need to be run again before creating the PR.

PR #13310, opened 2025-12-10
Revamp issue-labeler project

MaiLinhP:
#### Replaced all old ML labeler processes with streamlined `IssueLabelerMLPipeline`. Most of this pipeline is meant to be run locally for more control over model training processes. It includes: 
1. `Downloader`: Download applicable issues to local .tsv file. 
2. `Trainer`: Use downloaded issues (along with synthetic issues generated from `AzureSdkIssueGeneratorService` if available) to train Category and Service models. This will also provide per label statistics as well as a list of labels that need more data for training. This list can be passed back into `AzureSdkIssueGeneratorService` to generate issues with specific labels.
3. `Predictor`: Use the category and service model zip files produced from Trainer to predict issues. 
4. `Tester`: Test models on all labeled issues for # of matches, mismatches, and no predictions. Also provides statistics on how models perform.
5. `DevOpsModelRetrainer`: Process to be handed to DevOps team to periodically retrain models automatically. This download issues, train models along with synthetic data if any, then replace current models with new ones.

#### Unified config setup.
Previously, most function app configs use `{org}/{repo}:{key}` or `defaults:{key}`. However, as `ModelHolderFactory` is a singleton and didn't have per repo config at runtime, we were doing roundabout logics to configure model files (check old `LegacyLabeler` and `IModelHolderFactoryLite.cs` for details). Per repo config is now available to `ModelHolderFactory`, which simplifies many configs and cut out unnecessary name translation/extraction logic.

#### Simplified the flow from `AzureSdkIssueLabelerService` to `Labeler`. 
Previously, there were many layers - `ModelHolder, ModelHolderFactory, ModelHolderFactoryLite, Labeler, LabelerLite, LegacyLabeler`. Replaced `LegacyLabeler` with `MLLabeler`, simplified the rest down to `ModelHolder` and `ModelHolderFactory`.

Previously, `ModelHolder` was also unaware of which label type it is, and the ordering of service vs category models in `ModelHolderFactory` was dictated by order listed in `defaults:IssueModel.{repo}.BlobConfigNames`. This made code unreadable without constantly referencing app config. `ModelHolder` now has `modelType` which makes ordering irrelevant, simplifies call to config and making code much more readable.

#### Cleaned up codebase
 Deleted all unused references and refactored functions.

### DIFF
--- a/tools/issue-labeler/src/IssueLabelerMLPipeline/src/GitHubClient/GitHubApi.cs
+++ b/tools/issue-labeler/src/IssueLabelerMLPipeline/src/GitHubClient/GitHubApi.cs
@@ -4,7 +4,6 @@
 using GraphQL;
 using GraphQL.Client.Http;
 using GraphQL.Client.Serializer.SystemTextJson;
-using Octokit;
 using System.Collections.Concurrent;
 using System.Net.Http.Json;
 


tools/issue-labeler/src/IssueLabelerMLPipeline/src/DevOpsModelRetrainer/DevOpsModelRetrainer.cs
```diff
@@ -0,0 +1,245 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Azure.Identity;
+using Azure.Storage.Blobs;
+using Microsoft.Extensions.Configuration;
```
jeo02, line 6: Any reason this class is all top level rather than being a class?

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Downloader/Downloader.cs
```diff
@@ -0,0 +1,39 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
jeo02, line 2: Also all top level would it not be better to have it be a class?

tools/issue-labeler/src/IssueLabelerMLPipeline/src/GitHubClient/GitHubApi.cs
```diff
@@ -0,0 +1,570 @@
...
+    /// <param name="issuesLimit">The maximum number of issues to download.</param>
+    /// <param name="pageSize">The number of items per page in GitHub API requests.</param>
+    /// <param name="pageLimit">The maximum number of pages to retrieve.</param>
+    /// <param name="retries">An array of retry delays in seconds.</param>
+    /// <param name="excludedAuthors">An array of authors to exclude from the results.</param>
+    /// <param name="verbose">Emit verbose output into the action log.</param>
+    /// <returns>The downloaded issues as an async enumerable collection of tuples containing the issue and its predicate-matched label (when only one matcing label is found).</returns>
+    public static async IAsyncEnumerable<(Issue Issue, string CategoryLabel, string ServiceLabel)> DownloadIssues(
+        string githubToken,
+        string org, string repo,
```
jeo02, line 74: (nit)
```suggestion
        string org, 
        string repo,
```

tools/issue-labeler/src/IssueLabelerMLPipeline/src/GitHubClient/GitHubApi.cs
```diff
@@ -0,0 +1,570 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using GraphQL;
+using GraphQL.Client.Http;
+using GraphQL.Client.Serializer.SystemTextJson;
+using Octokit;
```
jeo02:
```suggestion
```
Using graphql no octokit needed :)

tools/issue-labeler/src/IssueLabelerMLPipeline/src/GitHubClient/GitHubClient.csproj
```diff
@@ -0,0 +1,22 @@
...
+  <PropertyGroup>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="GitHubJwt" />
+    <PackageReference Include="GraphQL.Client" />
+    <PackageReference Include="GraphQL.Client.Serializer.SystemTextJson" />
+    <PackageReference Include="Octokit" />
```
jeo02, line 12: won't need octokit here either

tools/issue-labeler/src/IssueLabelerMLPipeline/src/GitHubClient/QueryModel.cs
```diff
@@ -0,0 +1,97 @@
...
+{
+    public class RepositoryQuery<T>
+    {
+        public required RepositoryItems Repository { get; init; }
+
+        public class RepositoryItems
+        {
+            public required T Result { get; init; }
+        }
+    }
```
jeo02, line 14: Probably should just make a models folder and put these all in there as separate classes that way there much easier to find.

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Predictor/Models.cs
```diff
@@ -0,0 +1,55 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
jeo02, line 1: Class per file :) like the other comment

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Predictor/Predictor.cs
```diff
@@ -0,0 +1,193 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
jeo02, line 1: top level but seeing as so much of the code is structured by classes, I think we should keep consistent and make this a class as well. (Unless its top level for some reason I don't see)

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Tester/Tester.cs
```diff
@@ -0,0 +1,236 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
jeo02, line 1: top level

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Tester/Tester.cs
```diff
@@ -0,0 +1,236 @@
...
+            Score = score,
+            Label = labels.GetItemOrDefault(index).ToString()
+        })
+        .OrderByDescending(p => p.Score)
+        .FirstOrDefault(p => threshold is null || p.Score >= threshold);
+
+    return bestScore is not null ? (bestScore.Label, bestScore.Score) : ((string?)null, (float?)null);
+}
+
+class TestStats
```
jeo02, line 214: Put this in its own file

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Trainer/Trainer.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
jeo02, line 1: top level

jeo02 (changes requested):
Looks good

Wish I could have figured out GraphQL myself, but it was quite difficult to use. Do you encounter less rate limiting through GraphQL vs using the API through octokit? I imagine it should be quite an improvement. 

Mostly just nits on the styling. I have nothing against top-level statements (even though the constant comments made it seem so) but since everything else is in classes I think we should keep it consistent. 

Also saw a couple of classes that are declared all in one file. To find it easier it would be better to just declare them separately under Models folder.

The synthetic issues are really cool !!

tools/issue-labeler/Directory.Build.props
```diff
@@ -27,6 +27,12 @@
     <MicrosoftNETTestSdkVersion>16.11.0</MicrosoftNETTestSdkVersion>
   </PropertyGroup>
 
+  <PropertyGroup>
```
jsquire, line 30: Is this for local support of the Azure Function?   I'm wondering if we should move this to a local props file that we ignore so that CI doesn't get confused.

tools/issue-labeler/src/IssueLabeler.Shared/IModelHolderFactory.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace IssueLabeler.Shared
+{
+    public interface IModelHolderFactory
+    {
+        Task<IModelHolder[]> CreateModelHolders(string owner, string repo);
+        Task<IModelHolder> CreateModelHolder(string owner, string repo, string modelType);
+        Task<IPredictor> GetPredictor(string owner, string repo, string modelType);
+    }
+    public class ModelHolderFactory : IModelHolderFactory
```
jsquire, line 15: Please split into separate files.    In general, .NET has a strong one-class-per-file norm that we'll want to stick with unless there's a type that can't stand on its own.   In that case, we'd want to nest it in its parent.

tools/issue-labeler/src/IssueLabeler.Shared/IModelHolderFactory.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        private readonly ConcurrentDictionary<(string, string, string), IModelHolder> models = new ConcurrentDictionary<(string, string, string), IModelHolder>();
+        private readonly ILogger<ModelHolderFactory> logger;
+        private readonly IRepositoryConfigurationProvider configurationProvider;
+        private SemaphoreSlim sem = new SemaphoreSlim(1,1);
```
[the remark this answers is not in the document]
jsquire: +1.    This is a good suggestion.

tools/issue-labeler/src/IssueLabeler.Shared/IModelHolderFactory.cs
```diff
@@ -0,0 +1,139 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using Microsoft.Extensions.Logging;
```
jsquire, line 4:
```suggestion

using Microsoft.Extensions.Logging;
```

nit: We generally throw a blank between the header and statements as a convention.

tools/issue-labeler/src/IssueLabeler.Shared/IModelHolderFactory.cs
```diff
@@ -0,0 +1,139 @@
...
+{
+    public interface IModelHolderFactory
+    {
+        Task<IModelHolder[]> CreateModelHolders(string owner, string repo);
+        Task<IModelHolder> CreateModelHolder(string owner, string repo, string modelType);
+        Task<IPredictor> GetPredictor(string owner, string repo, string modelType);
+    }
+    public class ModelHolderFactory : IModelHolderFactory
+    {
+        private readonly ConcurrentDictionary<(string, string, string), IModelHolder> models = new ConcurrentDictionary<(string, string, string), IModelHolder>();
```
jsquire, line 17: Just a quick mention that if you wanted to avoid the "Look at mah Java!" feel, you can shorten up the declarations on either side.   Nothing wrong with the verbose form if that's your preference.

```csharp
 private readonly ConcurrentDictionary<(string, string, string), IModelHolder> models = new();
```

```csharp
 private readonly var models = new ConcurrentDictionary<(string, string, string), IModelHolder>();
```

tools/issue-labeler/src/IssueLabeler.Shared/IModelHolderFactory.cs
```diff
@@ -0,0 +1,139 @@
...
+        private readonly ILogger<ModelHolderFactory> logger;
+        private readonly IRepositoryConfigurationProvider configurationProvider;
+        private SemaphoreSlim sem = new SemaphoreSlim(1,1);
+
+        public ModelHolderFactory(
+            ILogger<ModelHolderFactory> logger,
+            IRepositoryConfigurationProvider configurationProvider)
+        {
+            this.configurationProvider = configurationProvider;
```
jsquire, line 26: As much as it pains me, our team (and Microsoft in general) is not a fan of `this.` unless it's required for differentiation.   Its one of the reasons why the `_` prefix or `ConstantForm` is used for class members.

tools/issue-labeler/src/IssueLabeler.Shared/IModelHolderFactory.cs
```diff
@@ -0,0 +1,139 @@
...
+    public interface IModelHolderFactory
+    {
+        Task<IModelHolder[]> CreateModelHolders(string owner, string repo);
+        Task<IModelHolder> CreateModelHolder(string owner, string repo, string modelType);
+        Task<IPredictor> GetPredictor(string owner, string repo, string modelType);
+    }
+    public class ModelHolderFactory : IModelHolderFactory
+    {
+        private readonly ConcurrentDictionary<(string, string, string), IModelHolder> models = new ConcurrentDictionary<(string, string, string), IModelHolder>();
+        private readonly ILogger<ModelHolderFactory> logger;
```
jsquire, line 18: Team standards say that class-level members should be prefixed with `_`  or, when `readonly` may be pascal-cased to indicate that they're essentially constant after construction.

My personal preference is the latter, so this would be `private readonly ILogger<ModelHolderFactory> Logger;`.

I'm not a fan of the prefix, personally, but we should conform since this is in an Azure SDK repository rather than being a sandbox/prototype.

tools/issue-labeler/src/IssueLabeler.Shared/IModelHolderFactory.cs
```diff
@@ -0,0 +1,139 @@
...
+                modelHolders[0] = serviceHolder;
+                modelHolders[1] = categoryHolder;
+
+                return modelHolders;
+            }
+
+            // Some models need to be initialized; acquire the semaphore and initialize.
+            try
+            {
+                if (!this.sem.Wait(0))
```
jsquire, line 46: Gotta kill the `this.` in references too.

tools/issue-labeler/src/IssueLabeler.Shared/IModelHolderFactory.cs
```diff
@@ -0,0 +1,139 @@
...
+                modelHolder = await InitFor(owner, repo, modelType);
+                this.models.GetOrAdd((owner, repo, modelType), modelHolder);
+            }
+
+            return modelHolder;
+        }
+
+        public async Task<IPredictor> GetPredictor(string owner, string repo, string modelType)
+        {
+            var modelHolder = await CreateModelHolder(owner, repo, modelType);
```
jsquire, line 109: This is an unusual pattern.   I'd expect to see `CreateModelHolder` throw rather than return `null`, unless there's some golden path scenario where we would want a `null`?

tools/issue-labeler/src/IssueLabeler.Shared/IModelHolderFactory.cs
```diff
@@ -0,0 +1,139 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using Microsoft.Extensions.Logging;
+using System.Collections.Concurrent;
+
+namespace IssueLabeler.Shared
+{
+    public interface IModelHolderFactory
+    {
+        Task<IModelHolder[]> CreateModelHolders(string owner, string repo);
```
jsquire, line 11: nit:  The general C# convention is that asynchronous methods end with `Async`.    This is slowly changing as asynchronous code becomes the norm rather than exception, but the suffix case is still the norm.

As long as we're consistent in the codebase, whatever form you use is fine here.   I wanted to mention it, as you'll see it in product code and especially in the public API for our libraries.

tools/issue-labeler/src/IssueLabeler.Shared/IRepositoryConfigurationProvider.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace IssueLabeler.Shared
+{
+    /// <summary>
```
jsquire, line 6: We should be consistent in whether or not we decorate with XML comments.   Since this an internal tool, we have a lot of flexibility here.   Use them always, use them never, just the contract interfaces/bases... all good.   

We have them here, but not in `IModelHolderFactory` which I'd expect to follow the same convention.

tools/issue-labeler/src/IssueLabeler.Shared/IssueGeneratorPayload.cs
```diff
@@ -0,0 +1,17 @@
+using System.ComponentModel;
+using Newtonsoft.Json;
+
+namespace IssueLabeler.Shared
+{
+    public class IssueGeneratorPayload
+    {
+        public string RepositoryName { get; set; }
+        public string OutputFilename { get; set; }
+        public string? numIssues { get; set; }
```
jsquire, line 10: One of these things is not like the other ones....

- This should be PascalCase and annotate if the name needs to change for serialization
- This should be `NumberOfIssues` or similar to match the voice of the other members.

tools/issue-labeler/src/IssueLabeler.Shared/Predictor.cs
```diff
@@ -0,0 +1,138 @@
...
+
+using Microsoft.Extensions.Logging;
+using Microsoft.ML;
+using Microsoft.ML.Data;
+
+namespace IssueLabeler.Shared
+{
+    public class Predictor : IPredictor
+    {
+        private static SemaphoreSlim sem = new SemaphoreSlim(1);
```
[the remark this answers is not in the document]
jsquire: +1; this is a good suggestion.

jsquire, line 12: Just making sure that the `new SemaphoreSlim(1)` was intended rather than the usual `(1, 1)`.   Generally, its a good idea to specify the max so that we don't accidentally cause an issue via unintended double `Release`

tools/issue-labeler/src/IssueLabeler.Shared/Predictor.cs
```diff
@@ -0,0 +1,138 @@
...
+            if (predEngine == null)
+            {
+                throw new InvalidOperationException("expected prediction engine loaded.");
+            }
+            GitHubIssuePrediction prediction;
+            bool acquired = false;
+
+            try
+            {
+                await sem.WaitAsync();
```
jsquire, line 54: Do we need to constrain predictions to one-at-a-time?   Seems limiting given that issues can come in concurrently.

tools/issue-labeler/src/IssueLabeler.Shared/Predictor.cs
```diff
@@ -0,0 +1,138 @@
...
+                }
+            }
+
+            VBuffer<ReadOnlyMemory<char>> slotNames = default;
+            predEngine.OutputSchema[nameof(GitHubIssuePrediction.Score)].GetSlotNames(ref slotNames);
+
+            float[] probabilities = prediction.Score;
+            var labelPredictions = GetBestThreePredictions(probabilities, slotNames);
+
+            float maxProbability = probabilities.Max();
```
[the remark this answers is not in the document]
jsquire: ooooh, burn!

tools/issue-labeler/src/IssueLabelerMLPipeline/Directory.Packages.props
```diff
@@ -0,0 +1,27 @@
+<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+
+  <ItemGroup>
```
jsquire, line 6: Definitely need to bump these.   Identity is out of date enough that it scares me.

tools/issue-labeler/src/IssueLabelerMLPipeline/NuGet.config
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
jsquire, line 1: We need to remove this.   Microsoft has a new policy that requires us to pass through a proxy for all NuGet pulls and this resets that.

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Common/DataFileUtils.cs
```diff
@@ -0,0 +1,79 @@
...
+            Directory.CreateDirectory(outputDir);
+        }
+    }
+
+    /// <summary>
+    /// Sanitizes the specified text by replacing certain characters to ensure it is collapsed onto a single line and compatible with tab-separated values.
+    /// </summary>
+    /// <param name="text">The text to sanitize.</param>
+    /// <returns>The sanitized text.</returns>
+    public static string SanitizeText(string text)
```
jsquire, line 28: I think we can add `.AsSpan()` on `text` and save a bunch of allocations.

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Common/LabelType.cs
```diff
@@ -0,0 +1,5 @@
+public enum LabelType
+{
+    Category = 1,
```
jsquire, line 3: Why are we explicitly numbering?   Generally, you don't want to do that unless you're actually using the numeric values.    We also have an issue here as `default(LabelType)`  ends up being an invalid because it gets initialized as `(LabelType)0`.    We'll want to add an `Unknown = 0` at least here.

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Common/LabelType.cs
```diff
@@ -0,0 +1,5 @@
...
+    Service = 2
+}
```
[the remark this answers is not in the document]
jsquire: +1 .  We do need that.

tools/issue-labeler/src/IssueLabelerMLPipeline/src/DevOpsModelRetrainer/Args.cs
```diff
@@ -0,0 +1,191 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+public struct Args
+{
+    public string? GitHubToken { get; set; }
```
jsquire, line 6: This looks like a file that we borrowed.   We'll want to update to our header.   I was going to suggest using `System.CommandLine` but if we inherited this and we know it works, better to not.

jsquire:
GitHub is being really, really laggy today.   I left some thoughts but didn't dig into the back half of the files yet.

tools/issue-labeler/src/IssueLabelerService/IssueGenerator/OpenAiIssueGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+        private async Task UploadToBlobAsync(string filePath, string repositoryName)
+        {
+            var containerClient = _blobClient.GetBlobContainerClient("generated-issues");
+            await containerClient.CreateIfNotExistsAsync();
+
+            var blobName = $"{repositoryName}/{Path.GetFileName(filePath)}";
+            var blobClient = containerClient.GetBlobClient(blobName);
+
+            using var fileStream = File.OpenRead(filePath);
+            await blobClient.UploadAsync(fileStream, overwrite: true);
```
radhgupta, line 212: I noticed there's some blob storage code that's getting repeated across a few files. We've got similar patterns in:

OpenAiIssueGenerator.cs (uploading generated issues)
DevOpsModelRetrainer.cs (uploading models + downloading synthetic data)
IssueTriageContentRetrieval.cs (uploading JSON content + soft delete config)


I was wondering if it we could create a service class for blob related operating and reuse it?

tools/issue-labeler/src/IssueLabelerService/IssueGenerator/OpenAiIssueGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+
+            _logger.LogInformation($"Uploaded TSV file to blob: {blobName}");
+        }
+
+        private string GetCategoryLabelsForPrompt(IEnumerable<Label> labels)
```
radhgupta, line 217: Seems like `GetCategoryLabelsForPrompt` and `GetServiceLabelsForPrompt` are almost similar can we make a single method??

tools/issue-labeler/src/IssueLabelerService/IssueGenerator/OpenAiIssueGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+
+            // Check if the blob exists
+            if (!await blobClient.ExistsAsync())
+            {
+                throw new FileNotFoundException($"Blob for repository '{repositoryName}' not found.");
+            }
+
+            // Download the blob content
+            var response = await blobClient.DownloadContentAsync();
+            var labelsJson = response.Value.Content.ToString();
```
radhgupta, line 190: For large label files, consider streaming the JSON directly instead of loading the entire content into memory.
```suggestion
     using var stream = await blobClient.OpenReadAsync();
    var labels = await JsonSerializer.DeserializeAsync<IEnumerable<Label>>(stream);

```

tools/issue-labeler/src/IssueLabelerService/IssueGenerator/OpenAiIssueGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+            KnowledgeAgentRetrievalRequest retrievalRequest = BuildRetrievalRequest(instructions, message);
+
+            var retrievalResult = await _retrievalClient.RetrieveAsync(retrievalRequest);
+
+            var contextBlock = GetContextBlock(retrievalResult);
+
+            var response = await _ragService.SendMessageQnaAsync(instructions, message, modelName, contextBlock);
+            var issues = JsonConvert.DeserializeObject<IEnumerable<IssueTriageContent>>(response);
+
+            if (issues == null)
```
radhgupta, line 78: I'm not sure, but is it possible that the AI response could be empty or not have a valid JSON structure? Currently we only check if issues is null after deserialization, but what if the AI returns an empty array, malformed JSON, or unexpected schema?

tools/issue-labeler/src/IssueLabelerService/IssueGenerator/OpenAiIssueGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+            }
+
+            // Write answer to TSV for model input, as well as more readable JSON file
+            // to local storage
+            var jsonFileName = $"{payload.OutputFilename}.json";
+            var fileName = $"{payload.OutputFilename}.tsv";
+            await using (StreamWriter writer = new StreamWriter(fileName))
+            {
+                await writer.WriteLineAsync(FormatIssueRecord("CategoryLabel", "ServiceLabel", "Title", "Body"));
+                foreach (var issue in issues)
```
radhgupta, line 91: for `var issues` the multiple enumeration isn't too severe since we're deserializing from a string, I think we could materialize the collection once with .ToList() to avoid it

tools/issue-labeler/src/IssueLabelerService/IssueGenerator/OpenAiIssueGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+            return string.Join(
+                "\n\n",
+                allSources.Select((obj, i) =>
+                    $"Title: {obj["title"]}\nDescription: {obj["content"]}\nTerms: {obj["terms"]}")
+            );
+        }
+        private async Task<IEnumerable<Label>> GetLabelsAsync(string repositoryName)
+        {
+            // Initialize BlobServiceClient
+            var containerClient = _blobClient.GetBlobContainerClient("labels");
```
radhgupta, line 177: Consider moving  "labels" to configuration or constants class for better maintainability

tools/issue-labeler/src/IssueLabelerService/IssueGenerator/OpenAiIssueGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+            {
+                throw new InvalidOperationException("Failed to deserialize labels from blob.");
+            }
+
+            return labels;
+        }
+
+        private async Task UploadToBlobAsync(string filePath, string repositoryName)
+        {
+            var containerClient = _blobClient.GetBlobContainerClient("generated-issues");
```
radhgupta, line 205: Consider moving container names to configuration or constants class for better maintainability

tools/issue-labeler/src/IssueLabelerService/IssueGenerator/OpenAiIssueGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+        {
+            var serviceLabels = labels
+                .Where(AzureSdkLabel.IsServiceLabel)
+                .Select(label => label.Name)
+                .ToList();
+
+            return string.Join(", ", serviceLabels);
+        }
+
+                public static string FormatTemplate(string template, Dictionary<string, string> replacements, ILogger logger)
```
radhgupta, line 236: I noticed there are static utility methods mixed with instance methods in OpenAiIssueGenerator. The static methods like FormatTemplate, FormatIssueRecord, and SanitizeText break the dependency injection pattern used elsewhere in the class.

Consider either:

- Moving these to a dedicated ITextProcessingService that can be injected
- Making them instance methods so they can use the existing _logger field

This would make the code more consistent with the DI pattern used throughout the rest of the class

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Common/DataFileUtils.cs
radhgupta, line 28: We have identical SanitizeText and FormatIssueRecord methods in both projects.  Consider moving these to the shared IssueLabeler.Shared project

tools/issue-labeler/src/IssueLabelerMLPipeline/src/DevOpsModelRetrainer/DevOpsModelRetrainer.cs
```diff
@@ -0,0 +1,245 @@
...
+        
+        // Ensure directory exists for nested blob paths
+        var directory = Path.GetDirectoryName(localFilePath);
+        if (!string.IsNullOrEmpty(directory))
+        {
+            Directory.CreateDirectory(directory);
+        }
+        
+        Console.WriteLine($"Downloading: {blobItem.Name}");
+        await blobClient.DownloadToAsync(localFilePath);
```
radhgupta, line 231: Sequential blob downloads will be slow for large datasets.Can we parallelize it??

tools/issue-labeler/src/IssueLabelerMLPipeline/src/DevOpsModelRetrainer/DevOpsModelRetrainer.cs
```diff
@@ -0,0 +1,245 @@
...
+        localFilePaths.Add(localFilePath);
```
radhgupta, line 232: How many synthetic files are we expecting here? Should we be concerned about the growing list causing memory issues? If repositories could have hundreds or thousands of synthetic files, this approach might consume significant memory. would it be worth considering if we need batching or streaming for large datasets.

tools/issue-labeler/src/IssueLabelerMLPipeline/src/Common/App.cs
```diff
@@ -0,0 +1,58 @@
...
+        var success = false;
+
+        try
+        {
+            await task;
+            success = true;
+        }
+        catch (AggregateException ex)
+        {
+            Console.WriteLine($"ERROR: Exception occurred: {ex.Message}");
```
radhgupta, line 52: I notice there are quite a few Console.WriteLine statements throughout. Since the rest of the codebase uses ILogger for structured logging, should we consider refactoring these to use a logger instead? This would give us better control over log levels, structured data, and consistency with the rest of the application.

tools/issue-labeler/src/IssueLabelerMLPipeline/src/DevOpsModelRetrainer/DevOpsModelRetrainer.cs
```diff
@@ -0,0 +1,245 @@
...
+        else
+        {
+            Console.WriteLine($"Dry run mode: Skip upload to {categoryModelBlobName} and {serviceModelBlobName}");
+        }
+
+        return true;
+    }
+    catch (Exception ex)
+    {
+        Console.WriteLine($"ERROR: Error retraining {typeName} models for {fullRepoName}: {ex.Message}");
```
radhgupta, line 183: Will it be worth adding retry logic like we have in GithubApi.cs??

tools/issue-labeler/src/IssueLabelerService/IssueGenerator/OpenAiIssueGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+        private readonly KnowledgeAgentRetrievalClient _retrievalClient;
+        private BlobServiceClient _blobClient;
+
+        public OpenAiIssueGenerator(ILogger<IssueGeneratorFactory> logger, RepositoryConfiguration config, TriageRag ragService, BlobServiceClient blobClient)
+        {
+            _config = config;
+            _ragService = ragService;
+            _logger = logger;
+            _blobClient = blobClient;
+            _retrievalClient = new KnowledgeAgentRetrievalClient(
```
radhgupta, line 32: I notice the KnowledgeAgentRetrievalClient is created in the constructor but never disposed. Since this likely implements IDisposable and holds network resources, it could cause connection leaks in long-running scenarios.

Should ww Make OpenAiIssueGenerator implement IDisposable and dispose the client,

tools/issue-labeler/src/IssueLabelerMLPipeline/src/DevOpsModelRetrainer/DevOpsModelRetrainer.cs
MaiLinhP: Right now we have one, and each synthetic data file does require a team member to proofread for accuracy, so I do not foresee us having more than say 10 quality files that can be used :)

PR #13325, opened 2025-12-11
Refactor API Review task conditions in YAML

weshaggard:
DevOps will try to resolve the service connections even if there is a runtime condition. So, we need to change this condition to be a template condition to fix pipelines in the other.

### DIFF
--- a/eng/common/pipelines/templates/steps/create-apireview.yml
+++ b/eng/common/pipelines/templates/steps/create-apireview.yml
@@ -37,7 +37,7 @@ steps:
       parameters:
         WorkingDirectory: ${{ parameters.SourceRootPath }}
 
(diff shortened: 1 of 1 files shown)


AlitzelMendez (approved):

eng/common/pipelines/templates/steps/create-apireview.yml
hallipr, line 36: Do we still need to set default branch?

weshaggard: We might not need it but that can be done as a follow-up I want to get this in ASAP to get all the public pipelines unblocked.

AlitzelMendez (approved):

eng/common/pipelines/templates/steps/create-apireview.yml
weshaggard: Looking at the full yaml there also may be other steps that indirectly rely on this DefaultBranch being set so we would need to do some deeper diving before blindly moving it.

PR #13330, opened 2025-12-12
Reorganize tool namespaces & tests

smw-ms:
- Restructured `azsdk-cli\Azure.Sdk.Tools.Cli.Tests\Tools` to match the folder structure of `azsdk-cli\Azure.Sdk.Tools.Cli\Tools`
- Aligned namespaces with their folder path

This PR isn’t critical; it’s just an organizational idea I wanted to share.


praveenkuttappan (approved):

benbp (approved):
Originally we had decided to just use the flat `*.Tools` namespace, and then later changed our minds to go back to the more conventional approach. It may be possible to remove the style rule that disallows namespaces not matching directory structure to prevent the drift we're seeing here: https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/.editorconfig#L4

Though I forget if there are other places where we still intentionally have the namespacing flat.

Issue #12059, opened 2025-09-11
Add automation to report on non-released SDKs


maririos:
A release plan includes the month the SDKs are aiming to released. After that day a notification email would be sent to the owner of the Release Planner to remind them to release the missing SDKs or file for an exclusion.
- Frequency: ??
- To: release planner owner, Justin, Josephine, ???
- Content: It should include:
  - Release plan link
  - Guidelines link: https://aka.ms/azsdk/dpcodegen
  - Missing SDKs
  - Exclusion process: https://eng.ms/docs/products/azure-developer-experience/onboard/request-exception

maririos:
@MrJustinB could you review the requirements and add answers for my `???` . Also, can you provide a template for the email that would be sent?

MrJustinB:
Send to service team and cc AzureSDKAPEX <azsdkapex@microsoft.com>

MrJustinB:
[tier 1 SDK exclusion template.docx](https://github.com/user-attachments/files/22351534/tier.1.SDK.exclusion.template.docx)

praveenkuttappan:
This change will use MCP CLI to get past due release plans.
https://github.com/Azure/azure-sdk-tools/issues/12611

MrJustinB:
[SDKs not released email template.docx](https://github.com/user-attachments/files/24598755/SDKs.not.released.email.template.docx)

PR #13333, opened 2025-12-12
Add automation to report on non-released SDKs

smw-ms:
Resolves Issue: https://github.com/Azure/azure-sdk-tools/issues/12059

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
@@ -1055,7 +1055,7 @@ namespace Azure.Sdk.Tools.Cli.Tools.ReleasePlan
             }
         }
 
-        private async Task NotifyOwnersOfOverdueReleasePlans(List<ReleasePlanDetails> releasePlans, string emailerUri)
+        private async Task NotifyOwnersOfOverdueReleasePlans(List<ReleasePlanWorkItem> releasePlans, string emailerUri)
         {
             const string subject = "Action Required: Azure SDKs Not Yet Published for Your Release Plan";
             
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/ReleasePlan/ReleasePlanToolTests.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Tools/ReleasePlan/ReleasePlanToolTests.cs
@@ -439,7 +439,7 @@ namespace Azure.Sdk.Tools.Cli.Tests.Tools.ReleasePlan
         public async Task Test_notification_includes_correct_missing_sdks()
         {
             var mockDevOps = new Mock<IDevOpsService>();
-            var plan = new ReleasePlanDetails
+            var plan = new ReleasePlanWorkItem
             {
                 WorkItemId = 200,
                 Owner = "Test Owner",
@@ -486,7 +486,7 @@ namespace Azure.Sdk.Tools.Cli.Tests.Tools.ReleasePlan
         public async Task Test_notification_excludes_approved_and_requested_languages()
         {
             var mockDevOps = new Mock<IDevOpsService>();
-            var plan = new ReleasePlanDetails
+            var plan = new ReleasePlanWorkItem
             {
                 WorkItemId = 201,
                 Owner = "Test Owner",
@@ -533,7 +533,7 @@ namespace Azure.Sdk.Tools.Cli.Tests.Tools.ReleasePlan
         public async Task Test_notification_excludes_go_for_dataplane()
         {
             var mockDevOps = new Mock<IDevOpsService>();
-            var plan = new ReleasePlanDetails
+            var plan = new ReleasePlanWorkItem
             {
                 WorkItemId = 202,
                 Owner = "Test Owner",
@@ -579,7 +579,7 @@ namespace Azure.Sdk.Tools.Cli.Tests.Tools.ReleasePlan
         public async Task Test_notification_includes_go_for_management_plane()
         {
             var mockDevOps = new Mock<IDevOpsService>();
-            var plan = new ReleasePlanDetails
+            var plan = new ReleasePlanWorkItem
             {
                 WorkItemId = 203,
                 Owner = "Test Owner",


eng/scripts/Report-Unreleased-Sdks.ps1
```diff
@@ -0,0 +1,102 @@
+<#
```
weshaggard: Should this be in the azsdk cli instead of a script?

benbp: If so, we can put it under the `eng` subcommand group. These commands are only visible when run in debug mode (via dotnet run).

eng/pipelines/report-unreleased-sdks.yml
```diff
@@ -0,0 +1,29 @@
+trigger: none
+pr: none 
+
+pool:
+  name: azsdk-pool
+  demands: ImageOverride -equals ubuntu-24.04
```
benbp: We should use the pre-defined image variables instead, it makes them easier to change en masse.

You can include the image variable template at the top level of indentation in this file.

```
variables:
  - template: /eng/pipelines/templates/variables/image.yml
```

Related example (though in a stage template):

https://github.com/Azure/azure-sdk-tools/blob/0865a952d2c3e25b0c92bb0232e78b593a551701/eng/pipelines/prune-acr-images.yml#L17-L26

In this instance, the pool declaration would need to look like below. There's a chance `demands` doesn't support variable evaluation, but hopefully it does.

```
pool:
  name: $(LINUXPOOL)
  demands: ImageOverride -equals $(LINUXVMIMAGE)
```

eng/pipelines/report-unreleased-sdks.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  - job: ReportUnreleasedSdks
+    steps:
+      - checkout: self
+
+      - task: PowerShell@2
+        displayName: 'Install Azure SDK MCP'
```
benbp, line 17: Maybe it's worth putting this task into `eng/common/pipelines/templates/steps/install-azure-sdk-tools-cli.yml`? I'll do this as a separate PR after this goes in so we don't have to do the eng/common sync process for these changes.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -1005,5 +1029,94 @@ public async Task<ReleasePlanListResponse> ListOverdueReleasePlans()
                 return new ReleasePlanListResponse { ResponseError = $"An error occurred while retrieving overdue release plans: {ex.Message}" };
             }
         }
+
+        private async Task NotifyOwnersOfOverdueReleasePlans(List<ReleasePlanDetails> releasePlans, string emailerUri)
+        {
+            const string subject = "Action Required: Missing Tier 1 language in your Azure SDK Release Plan";
```
praveenkuttappan: We are notifying users about their pending release(past due release) and not about missing SDK in release plan. right?

smw-ms: @MrJustinB  created an email template he wanted to use in the issue: https://github.com/Azure/azure-sdk-tools/issues/12059

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -1005,5 +1029,94 @@ public async Task<ReleasePlanListResponse> ListOverdueReleasePlans()
...
+                var missingSDKs = new List<string> { ".NET", "JavaScript", "Python", "Java", "Go" };
+                // Skip Go for Data Plane release plans
+                if (releasePlan.IsDataPlane)
+                {
+                    missingSDKs.Remove("Go");
+                }
+                // Remove languages that are already released
+                foreach (var info in releasePlan.SDKInfo)
+                {
+                    if (string.Equals(info.ReleaseStatus, "released", StringComparison.OrdinalIgnoreCase))
```
praveenkuttappan: We should also skip the languages which are marked as excluded. This information is available in language specific fields in the work item. We should ignore all languages with exclusion status requested/approved.

`ReleaseExclusionStatusForPython`
`ReleaseExclusionStatusForDotnet`
`ReleaseExclusionStatusForJava`
`ReleaseExclusionStatusForJavaScript`
`ReleaseExclusionStatusForGo`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
```diff
@@ -1005,5 +1029,94 @@ public async Task<ReleasePlanListResponse> ListOverdueReleasePlans()
...
+                var plane = releasePlan.IsManagementPlane ? "Management Plane" : "Data Plane";
+                var releasePlanLink = releasePlan.ReleasePlanLink;
+                
+                // Identify missing Tier 1 languages
+                var missingSDKs = new List<string> { ".NET", "JavaScript", "Python", "Java", "Go" };
+                // Skip Go for Data Plane release plans
+                if (releasePlan.IsDataPlane)
+                {
+                    missingSDKs.Remove("Go");
+                }
```
praveenkuttappan: I think it's better not to handle the list here. It will easily be forgotten that we have a list maintained here too. Instead, we can just rely on the SDKInfo  list in the release plan. 

releasePlan.SDKInfo.Where(s => !string.Equals(s.ReleaseStatus, "released", StringComparison.OrdinalIgnoreCase))

praveenkuttappan:
@maririos @MrJustinB   Current approach identifies the list of past due release plans and will send an email notification to release plan owner. It will send an email notification to users until they get the SDK as released or excluded. We can schedule this pipeline to run weekly so it's less aggressive but act as a reminder email weekly to users with past due release plans. Alternate solution is to add a field in work item to check if the notification was already sent or not, so we send the email only once per release plan. Please share your thoughts on this solution. I am in favor of repeated weekly notification which act as a reminder to complete the past due release plans.

MrJustinB:
@praveenkuttappan @maririos  I am ok with weekly, but I am wondering if I misunderstood the feature when I drafted the email template. This PR is to notify teams that have decided to have SDKs for all tier 1 languages BUT they have not released them yet? If so, I should update the template to match that scenario. However, if we're adding self-serve to release planner for data plane, aren't the SDKs released automatically on behalf of the service teams? If so, does that make this PR obsolete? If not, can you please help me understand the specific scenario with self-serve enabled?

benbp (approved):
pipeline pieces LGTM

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
praveenkuttappan: @MrJustinB This change detects the overdue release plans based on the release month selected in the release plan. So it is not only about the languages opted out in the release plans. It will detect the release plans that were supposed to be completed but not released for all required languages yet. I think we should update the email template to clarify the message.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
smw-ms: done

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/ReleasePlanTool.cs
smw-ms: if no sdk info is filled out in work item the release plan object will have an empty list for sdk info section
 https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs#L250

praveenkuttappan (approved):

PR #13334, opened 2025-12-12
Add 'APIView prod deployment' to internal service connections

weshaggard:


tjprescott:
@weshaggard out of curiosity, what does adding APIView to this script do?

weshaggard:
> @weshaggard out of curiosity, what does adding APIView to this script do?

This script is used to setup our pipelines and currently we added this new APIView service connection usage to our internal pipelines, but they don't permissions to that. This will add it to the list of service connections we grant permissions to our pipeline.

raych1 (approved):

PR #13377, opened 2025-12-17
[azsdk-cli, go] Some fixes to make the pkg validate tool a bit more useful

richardpark-msft:
- Search, recursively, for go.mod files. This helps in the 'azcore' case where we have some perf tests in a sub-folder.
- Use the repo configured golangci-lint configuration file when linting
- Added in a convenience method for PackageCheckResult that can handle multiple ProcessResults.
- Parsing the go.mod file to get the actual package name instead of guessing using file system paths
- Adding some convenience constructors to make sure we're always sending package name and the language.

Related to https://github.com/Azure/azure-sdk-tools/issues/11399

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.Checks.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.Checks.cs
@@ -63,29 +63,49 @@ public partial class GoLanguageService : LanguageService
 
     public override async Task<PackageCheckResponse> AnalyzeDependencies(string packagePath, bool fixCheckErrors = false, CancellationToken ct = default)
     {
+        var results = new List<ProcessResult>();
+
         try
         {
-            var goGetArgs = new List<string>(["get", "-u", "all"]);
-            var goModVersion = await GetGoModVersionAsync(Path.Join(packagePath, "go.mod"), ct);
+            // in some modules we have two go.mod files (or multiple). We want to update those as well.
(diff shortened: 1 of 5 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.Checks.cs
```diff
@@ -63,36 +63,57 @@ public async Task<PackageCheckResponse> CreateEmptyPackage(string packagePath, s
...
+                if (tidyResult.ExitCode != 0)
+                {
+                    break;
+                }
             }
 
-            // Now tidy, to cleanup any deps that aren't needed
-            var tidyResult = await processHelper.Run(new ProcessOptions(goUnix, ["mod", "tidy"], goWin, ["mod", "tidy"], workingDirectory: packagePath), ct);
-            return new PackageCheckResponse(tidyResult);
+            return new PackageCheckResponse(results);
```
praveenkuttappan: Can you please set language and package name(if applicable) in response?

richardpark-msft: I'll take a look!

richardpark-msft: Okay, I fixed this and added some new constructors to make it easier to pass in those two attributes.

There's a few other response types that probably need similar treatment, and then we need to move everyone over to them.

praveenkuttappan (approved):

PR #13385, opened 2025-12-17
Update missing telemetry in SDK generation, TypeSpec validation and add TypeSpec as language in TypeSpec tool responses

praveenkuttappan:
This PR adds missing telemetry data for few tools.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/SpecWorkFlowTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/SpecWorkFlowTool.cs
@@ -204,11 +204,17 @@ namespace Azure.Sdk.Tools.Cli.Tools.ReleasePlan
                     response.Status = "Failed";
                 }
                 response.SetLanguage(language);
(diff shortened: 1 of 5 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/SpecWorkFlowTool.cs
```diff
@@ -204,12 +204,18 @@ public async Task<ReleaseWorkflowResponse> RunGenerateSdkAsync(string typespecPr
                     response.Status = "Failed";
                 }
                 response.SetLanguage(language);
+                string typeSpecProjectPath = "";
```
maririos, line 207: it is now intentional to have the `TypeSpecProject` not set when the path is invalid, right?
If that is correct and we want to make that the default behavior, it would be good to add a test for that scenario

praveenkuttappan: Added test case

chrisradek (approved):

PR #13398, opened 2025-12-18
[azsdk-cli, go] Fixing a few more Go issues, and adding in a few more convenience funcs.

### LINKED ISSUE #11399: [Activity] Run Go specific validation

richardpark-msft:
- Fixing a bug in the Go changelog command that was passing the wrong value (path instead of module name!)
- Fixed an issue in GetPackageProperties() where we'd also try to parse stderr output, which is never a good idea.
- Added copyright header validation into the Lint() step.

Other small QoL stuff:
- Added a constructor to ProcessOptions that lets you use a win/unix command with the _same_ args, which is the common case for me.
- Added a "Stdout" property to ProcessResult which only extracts Stdout (unlike Output, which has both stderr and stdout)

Fixes #11399

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.Checks.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.Checks.cs
@@ -158,7 +158,7 @@ public partial class GoLanguageService : LanguageService
                 ["-Packages", packagePath]
             );
 
-            var results = await powershellHelper.Run(powershellOptions, ct);
+            result = await powershellHelper.Run(powershellOptions, ct);
             processResults.Add(result);
 
             return new PackageCheckResponse(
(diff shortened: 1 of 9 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.Checks.cs
```diff
@@ -180,7 +188,12 @@ public async Task<PackageCheckResponse> BuildProject(string packagePath, Cancell
         }
     }
 
-    public static async Task<string> GetSDKPackageName(string packagePath, CancellationToken cancellationToken = default)
+    /// <summary>
+    /// Gets the full go module name for a package path, by reading go.mod
+    /// </summary>
+    /// <param name="packagePath">A full path to a package folder (eg: [gitroot]/sdk/messaging/azservicebus)</param>
+    /// <returns>The full go module name (ex: github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus)</returns>
```
praveenkuttappan: Isn't the package name maintained in tooling and engsys for Go begins with `sdk/` so in this example, it is `sdk/messaging/azservicebus`.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.Checks.cs
```diff
@@ -194,22 +207,41 @@ public static async Task<string> GetSDKPackageName(string packagePath, Cancellat
...
     public override async Task<PackageCheckResponse> UpdateSnippets(string packagePath, bool fixCheckErrors = false, CancellationToken cancellationToken = default)
     {
         return await Task.FromResult(new PackageCheckResponse());
     }
 
     public override async Task<PackageCheckResponse> ValidateChangelog(string packagePath, bool fixCheckErrors = false, CancellationToken cancellationToken = default)
     {
-        var packageName = await GetSDKPackageName(packagePath, cancellationToken);
-        return await commonValidationHelpers.ValidateChangelog(packageName, packagePath, fixCheckErrors, cancellationToken);
+        var packageSubPath = await GetSubPath(packagePath, cancellationToken);
```
praveenkuttappan: Why is this using GetSubPath in this case when GetModeulName is used in other places? Can we use the same in all responses if this is not too relevant for Go workflow?

praveenkuttappan (approved):
Couple of questions about module name. Otherwise, it looks fine to me.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.Checks.cs
richardpark-msft: For Go, that's the actual package name. The sub-path is part of it, but the actual package name is the fully qualified bit, with the repo. 

I figure it's best to just use the full name - we've been bitten in the past when we used the last segment in APIView, for instance, and it just seems better to avoid any future issues by being unambiguous.

praveenkuttappan (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/GoLanguageService.Checks.cs
richardpark-msft: Changed over to use GetSubPath. The field is actually intended to be the ecosystem package name but Go is using a truncated version within all the systems, and this should value should match it.

PR #13402, opened 2025-12-19
More Eval Scenarios

### LINKED ISSUE #12307: Evaluation Scenarios
Add in more scenarios to cover more steps in the copilot instructions. 

This will be blocked until we create a mocked MCP since I don't want to spam create release plans. 

Also look into the different entry points to the azsdk MCP and test those as well as Scenarios.

### LINKED ISSUE #13455: Optional Tool Call
- Support optional calls to a tool during test.
- A good example would be verifying setup, allow it as optional that way we don't need to specify in the prompt that we should not run verify setup.

jeo02:
- Added a couple more eval scenarios
- Resolves #13455 allowing optional parameters so we don't need to rely on the prompt for not running x tool.
- Left the prompts as is since they do run quicker if less tools are called, but saying "don't run verify setup" for example is no longer needed. 
- Resolves #12307


tools/azsdk-cli/Azure.Sdk.Tools.Cli.Evaluations/Scenarios/AzureRestApiSpecs/Evaluate_GetModifiedTypespecProjects.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Azure.Sdk.Tools.Cli.Evaluations.Scenarios
+{
+    public partial class Scenario
+    {
+        [Test]
+        [Category(RepositoryCategories.AzureRestApiSpecs)]
+        public async Task Evaluate_GetModifiedTypespecProjects()
+        {
+            const string prompt = "List the TypeSpec projects modified in my repo. My setup has already been verified, do not run azsdk_verify_setup. Path to root: C:\\azure-rest-api-specs. Compare against main.";
```
benbp, line 15: Seems like we could use a way to always skip verify setup so these scenario prompts don't have to duplicate this instruction

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Evaluations/Helpers/ChatCompletion.cs
```diff
@@ -67,12 +72,52 @@ public async Task<ChatResponse> GetChatResponseWithExpectedResponseAsync(IEnumer
...
+            // No optional calls to filter. 
+            if(optionalCallIds.Count == 0)
+            {
+                return messages;
+            }
+
+            foreach (var message in messages)
+            {
+                var functionCalls = message.Contents.OfType<FunctionCallContent>();
+                var functionResults = message.Contents.OfType<FunctionResultContent>();
```
praveenkuttappan: Unused?

praveenkuttappan (approved):

praveenkuttappan (approved):

PR #13447, opened 2026-01-04
examples-automation, special handling for azure-resourcemanager at tsp

weidongxu-microsoft:
azure-resourcemanager lib need a special handling for getting an example from complete Swagger filename from an incomplete TypeSpec filename.

e.g. from `2025-05-01/listSecrets.json` to `specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2025-05-01/examples/listSecrets.json`


weidongxu-microsoft:
Here is another incentive to see if we'd move the samples back to its lib (instead of in azure-resourcemanager lib).

This PR does not need to be merged, if we decided to do above.

XiaofeiCao:
Created issue: 
- https://github.com/Azure/azure-sdk-for-java/issues/47621

PR #13453, opened 2026-01-05
[js-api-parser] add interface generator

KarishmaGhiya:
- adding interface generator

### DIFF
--- a/tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
+++ b/tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
@@ -50,9 +50,7 @@ function generate(item: ApiInterface, deprecated?: boolean): ReviewToken[] {
             deprecated,
           }),
         );
-        tokens.push(
-          createToken(TokenKind.Text, tp.constraintExcerpt.text.trim(), { deprecated }),
-        );
+        processExcerptTokens(tp.constraintExcerpt.spannedTokens, tokens, deprecated);
       }
 
       if (tp.defaultTypeExcerpt?.text.trim()) {
--- a/tools/apiview/parsers/js-api-parser/test/tokenGenerators/interface.spec.ts
+++ b/tools/apiview/parsers/js-api-parser/test/tokenGenerators/interface.spec.ts
@@ -66,7 +66,11 @@ function createDefaultExportExcerptTokens(interfaceName: string): ExcerptToken[]
 }
 
 // Helper to create a mock TypeParameter
-function createMockTypeParameter(name: string, constraintText?: string): TypeParameter {
+function createMockTypeParameter(
+  name: string,
+  constraintText?: string,
+  defaultTypeText?: string,
+): TypeParameter {
   const constraint = constraintText
     ? {
         text: constraintText,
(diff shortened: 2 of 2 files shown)


tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
```diff
@@ -0,0 +1,88 @@
...
+          createToken(TokenKind.Keyword, "extends", {
+            hasPrefixSpace: true,
+            hasSuffixSpace: true,
+            deprecated,
+          }),
+        );
+        processExcerptTokens(tp.constraintExcerpt.spannedTokens, tokens, deprecated);
+      }
+
+      if (index < typeParameters.length - 1) {
```
maorleger, line 67: Just food for thought, if we implemented a simple `intersperse` helper we could use that instead of doing the same check everywhere.

The idea would be:

`const typeParameterTokens = typeParameters.map(<logic to map to review tokens>).intersperse(createToken(TokenKind.Text, ",", { hasSuffixSpace: true, deprecated }))`

Or something along those lines. Not for this PR, but we could do a cleanup pass separately

The basic idea:

1. Build an array of ReviewToken
2. call `intersperse` getting a new array
3. tokens.push(result of [2])

jeremymeng:
Looks good!

jeremymeng (approved):

tools/apiview/parsers/js-api-parser/test/tokenGenerators/interface.spec.ts
```diff
@@ -0,0 +1,545 @@
+import { describe, expect, it } from "vitest";
```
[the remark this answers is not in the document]
KarishmaGhiya: updated

tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
KarishmaGhiya: Thank you @maorleger ! Tracking this in https://github.com/Azure/azure-sdk-tools/issues/13479

tools/apiview/parsers/js-api-parser/src/tokenGenerators/interfaces.ts
```diff
@@ -0,0 +1,90 @@
...
+            hasPrefixSpace: true,
+            hasSuffixSpace: true,
+            deprecated,
+          }),
+        );
+        tokens.push(
+          createToken(TokenKind.Text, tp.constraintExcerpt.text.trim(), { deprecated }),
+        );
+      }
+
```
maorleger, line 55: Somewhere in here we're missing the logic to show the "default" type for generic parameters

https://www.typescriptlang.org/docs/handbook/2/generics.html#generic-parameter-defaults

For example 

```ts
export interface PagedAsyncIterableIterator<TElement, TPage = TElement[], TPageSettings = PageSettings> {
```

Renders as 

```ts
export interface PagedAsyncIterableIterator<TElement, TPage, TPageSettings> {
```

Would you mind looking into that?

Looks like you can find it via `tp.defaultTypeExcerpt`

KarishmaGhiya: Yes processing of `tp.defaultTypeExcerpt` was missing entirely. Added it now. Can you review again?

PR #13468, opened 2026-01-06
Upgrade MCP library to 0.5.0

praveenkuttappan:
PR contains changes to upgrade MCP protocol lib and changes in instrumentation tool to support implementing Metadata getter (a requirement in new lib when inheriting delegate mcp tool class).

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
@@ -9,16 +9,28 @@ using static Azure.Sdk.Tools.Cli.Telemetry.TelemetryConstants;
 
 namespace Azure.Sdk.Tools.Cli.Tools.Core;
 
-public class InstrumentedTool(
-    ITelemetryService telemetryService,
-    ILogger logger,
-    McpServerTool innerTool
-) : DelegatingMcpServerTool(innerTool)
+public class InstrumentedTool : DelegatingMcpServerTool
 {
+    private readonly ITelemetryService telemetryService;
+    private readonly ILogger logger;
+    private readonly McpServerTool innerTool;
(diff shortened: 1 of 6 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
```diff
@@ -9,17 +9,29 @@
...
+    public InstrumentedTool(
+        ITelemetryService telemetryService,
+        ILogger logger,
+        McpServerTool innerTool
+    ) : base(innerTool)
+    {
+        this.telemetryService = telemetryService;
+        this.logger = logger;
+        this.innerTool = innerTool;
+    }
```
benbp, line 31: All this seems like an unnecessary change from primary constructor.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
```diff
@@ -9,17 +9,29 @@
...
+
+    public override IReadOnlyList<object> Metadata => innerTool.Metadata;
```
benbp, line 33: What is the intended use for `Metadata`?

praveenkuttappan: Metadata contains tool description and input schema etc.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/InstrumentedTool.cs
praveenkuttappan: It's mainly to avoid the warning when innerTool reference is used in InstrumentedTool class to get the metadata.

benbp (approved):

PR #13475, opened 2026-01-06
[azsdk-cli] Log to MCP clients over json-rpc instead of stdout

### LINKED ISSUE #11804: [azsdk-cli] Support MCP logging/notifications for ILogger
The MCP protocol supports logging. We should be able to use this now via `ILogger` so that debug messages and sub-process streaming can get sent back to the client.

https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging#protocol-messages

### LINKED ISSUE #13422: [Bug Bash] Too many warning messages which should ideally be info
### Prompts Used

- Verify setup for regeneration

### Language

_No response_

### Service or Resource Provider

_No response_

### Logs, Traces, or Screenshots

General logs for azure-sdk-mcp show [warning] messages which are ideally info
Image

### Feedback

Why do the mcp logs output have server stderr when they are not errors?

benbp:
resolves #11804
fixes #13422

This PR leverages the [MCP logging utility](https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging) and [MCP SDK logging](https://github.com/modelcontextprotocol/csharp-sdk/blob/main/docs/concepts/logging/logging.md) to route output from components back to the MCP client via json-rpc. Previously we were sending back all server log messages over stderr to avoid the misleading `failed to parse message` log prefix from MCP clients, but it was still confusing because all logs would show up as `[server stderr]`. With this update logs will show up in a normal format prefixed by the category name.

Aside from visual cleanup, the intent of this change is to also give some feedback to the user when there are long-running process operations going on. The internal process helper forwards stdout by default, so with this change that output now gets sent via MCP logging. In the future we could also consider building in first class support for the [MCP progress utility](https://github.com/modelcontextprotocol/csharp-sdk/blob/main/docs/concepts/progress/progress.md), but I think to start this will be sufficient.

Example mcp client logging output when running an external process:

> 2026-01-09 15:15:36.281 [info] Running command [pwsh -File /tmp/azsdk_example_c55940.ps1 foobar2] in /home/ben/ai/projects/mcp-progress/tools/azsdk-cli/Azure.Sdk.Tools.Cli
2026-01-09 15:15:36.287 [info] --------------------------------------------------------------------------------
2026-01-09 15:15:37.014 [info] [azsdk_example_c55940.ps1] 1: foobar2
2026-01-09 15:15:38.028 [info] [azsdk_example_c55940.ps1] 2: foobar2
2026-01-09 15:15:39.795 [error] [azsdk_example_c55940.ps1] [31;1mWrite-Error: [31;1mTest error message, no failure[0m
2026-01-09 15:15:40.796 [info] [azsdk_example_c55940.ps1] 3: foobar2
2026-01-09 15:15:40.833 [info] --------------------------------------------------------------------------------
2026-01-09 15:15:40.859 [info] Server (azsdk 0.5.12.0), Client (Visual Studio Code 1.106.3) method 'tools/call' request handler completed.


samvaity (approved):

samvaity:
@benbp I think this implementation can close https://github.com/Azure/azure-sdk-tools/issues/13422

benbp:
Added a couple updates:

- Removed the category name so the logger view is cleaner (users don't need to see the class FQN).
- Added a filter so process output logs don't get uploaded to azure monitor
- Refactored logger setup into its own extensions class so Program.cs is cleaner

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Extensions/LoggingExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                consoleLogOptions.LogToStandardErrorThreshold = LogLevel.Error;
+            });
+            builder.AddFilter<ConsoleLoggerProvider>((category, level) =>
+                level >= LogLevel.Error
+                && category != null
+                && (category.StartsWith("Microsoft.Hosting", StringComparison.Ordinal)
+                    || category.StartsWith("Microsoft.AspNetCore", StringComparison.Ordinal)
+                    || category.StartsWith("Microsoft.Extensions.Hosting", StringComparison.Ordinal)));
+        }
```
samvaity, line 39: New to C#, but we filter only these categories as they ensure errors, is that so?

benbp: Whoops I think my comment explaining this code block got lost in some refactoring. I'll restore it.

It's kind of a clunky approach. Some dependencies may use the logger instance provided by the host builder (i.e. asp.net) before the MCP server is initialized and can register it's logging provider. For most of these logs we want to swallow them so we don't spam the user with misleading `failed to parse message` logs from logger stdout. However, in the case of any fatal server startup errors, we wouldn't want to swallow those as it could make debugging difficult. So this is a little filter to make sure that we hide all non-error logs but show all asp.net server logs even if they can't be sent over the mcp protocol.

benbp: This whole logging setup is very auto-magical and was a confusing mess for me to try to understand and set up initially. I'm happy to walk you through it more if you're curious.

samvaity: [This](https://github.com/Azure/azure-sdk-tools/pull/13475/files#r2677580010) clarifies it, thanks.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/McpLogging.cs
```diff
@@ -0,0 +1,134 @@
...
+    private bool enabled = true;
+
+    public McpServer? Current
+    {
+        get => enabled ? current : null;
+    }
+
+    public bool IsEnabled => enabled;
+
+    public void TrySet(McpServer? server)
```
samvaity:
```suggestion
    public void SetIfEnabledAndUnset(McpServer? server)
```
TrySet sounds like it will return a boolean

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/McpLogging.cs
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using System.Collections.Concurrent;
+using ModelContextProtocol.Server;
+
+namespace Azure.Sdk.Tools.Cli.Helpers;
+
+// NOTE: the classes in this file are not thread-safe.
```
samvaity, line 7: ✔️

samvaity (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/McpLogging.cs
benbp: How about `Initialize`?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Extensions/LoggingExtensions.cs
benbp: Awesome. I copied most of that comment over to the filter location.

Issue #13318, opened 2025-12-11
[APIView] Latest Main tag not always accurate


swathipil:
The revision tagged "Latest Main" seems to always be applied to the latest generated revision from the nightly run, whether or not that revision was generated from "main" or another branch. This is causing some confusion. It would be clearer to relabel this "Latest Build" or something similar.

tjprescott:
Alternatively, would it make more sense to simply make "Latest Main" _actually_ be the latest from Main?

AlitzelMendez:
I don't think we can do that right now, with the information that we have as we are never storing the source branch of the revision

PR #13499, opened 2026-01-08
[APIView] Change Latest Main tag as it is misleading 

AlitzelMendez:
closes: https://github.com/Azure/azure-sdk-tools/issues/13318

The tag "Latest Main" is not assigned to the latest build from main, but to the latest automatic build, adding the "Main" part is really misleading


tjprescott (commented):
This is a simple change, but I'm not sure if the best fix is to just change the label to be more accurate or to fix the logic to ensure that "Latest Main" really IS the latest on main and not some weird branch.

AlitzelMendez:
> This is a simple change, but I'm not sure if the best fix is to just change the label to be more accurate or to fix the logic to ensure that "Latest Main" really IS the latest on main and not some weird branch.

we are not storign the branch information anywhere, so even if we fix it for the new builds is not going to be fixed for old ones, in some cases we have that information as tags that have tags saying "source build main" but is not on every language

Issue #13479, opened 2026-01-07
[js-api-parser] Implemented a simple `intersperse` helper


KarishmaGhiya:
Implemented a simple `intersperse` helper we could use that instead of doing the same check everywhere.

The idea would be:

`const typeParameterTokens = typeParameters.map(<logic to map to review tokens>).intersperse(createToken(TokenKind.Text, ",", { hasSuffixSpace: true, deprecated }))`

Or something along those lines. Not for this PR, but we could do a cleanup pass separately

The basic idea:

1. Build an array of ReviewToken
2. call `intersperse` getting a new array
3. tokens.push(result of [2])

_Originally posted by @maorleger in https://github.com/Azure/azure-sdk-tools/pull/13453#discussion_r2666143710_

PR #13516, opened 2026-01-09
[js-api-parser] Add Class generator

KarishmaGhiya:


KarishmaGhiya:
Do we want to support `export declare class <ClassName>` or convert it to `export class <ClassName>` @jeremymeng @maorleger

jeremymeng:
> Do we want to support `export declare class <ClassName>` or convert it to `export class <ClassName>` @jeremymeng @maorleger

It's okay to not special-case it and keep it. It shouldn't hurt Api reviewing.

maorleger:
> > Do we want to support `export declare class <ClassName>` or convert it to `export class <ClassName>` @jeremymeng @maorleger
> 
> It's okay to not special-case it and keep it. It shouldn't hurt Api reviewing.

Agree with Jeremy, no need to do any special work for it one way or another. Although, reading the code, I say optimize for less code in this case - like the snippet in https://github.com/Azure/azure-sdk-tools/pull/13516/changes#diff-c6b4c208211bc75ae63b5ce32bc49b97c3155d3914050c511346855e6fa8ddadR31-R34 can be removed entirely IMO

tools/apiview/parsers/js-api-parser/src/tokenGenerators/class.ts
```diff
@@ -0,0 +1,122 @@
...
+  const nameToken = createToken(TokenKind.TypeName, item.displayName, { deprecated });
+  nameToken.NavigateToId = item.canonicalReference.toString();
+  nameToken.NavigationDisplayName = item.displayName;
+  nameToken.RenderClasses = ["class"];
+  tokens.push(nameToken);
+
+  // Add type parameters
+  if (typeParameters?.length > 0) {
+    tokens.push(createToken(TokenKind.Text, "<", { deprecated }));
+    typeParameters.forEach((tp, index) => {
```
maorleger, line 47: Let's create a ticket to let copilot extract a helper function and centralize this typeParameter rendering logic. No need to do that now, just track it, so that we can just do that in one fell swoop once we're done. This can probably be done together with the #13479 work I think

tools/apiview/parsers/js-api-parser/src/tokenGenerators/class.ts
```diff
@@ -0,0 +1,122 @@
...
+  tokens.push(createToken(TokenKind.Keyword, "export", { hasSuffixSpace: true, deprecated }));
+
+  // Check for default export
+  const isDefaultExport = item.excerptTokens.some((t) => t.text.includes("export default"));
+  if (isDefaultExport) {
+    tokens.push(createToken(TokenKind.Keyword, "default", { hasSuffixSpace: true, deprecated }));
+  }
+
+  // Add declare keyword (common in .d.ts files)
+  const hasDeclare = item.excerptTokens.some((t) => t.text.includes("declare"));
```
maorleger: Let's remove it, `declare` adds no value in this context I think

tools/apiview/parsers/js-api-parser/src/tokenGenerators/class.ts
KarishmaGhiya: added in https://github.com/Azure/azure-sdk-tools/issues/13479

jeremymeng:
I noticed that doc comments are not added.  Can you make sure there's a ticket for adding it?

jeremymeng:
> I noticed that doc comments are not added. Can you make sure there's a ticket for adding it?

please discard. I didn't turn on showing of documentation.

tools/apiview/parsers/js-api-parser/src/tokenGenerators/class.ts
KarishmaGhiya: done

PR #13538, opened 2026-01-12
Refactor build logic to LanguageService

### LINKED ISSUE #12852: Extract SDK build code logic into a shared service
The customization code tool requires the ability to compile SDK code. To support this and improve reusability, it would be beneficial to extract the core build logic from the SDK build tool and make it available as a shared service that can be invoked by other tools.

A potential approach is to move the build logic into the corresponding language service. The current build logic depends on the `SpecGenSdkConfigHelper` class, and we should evaluate whether to migrate this dependency to the language service when work on this.

raych1:
Resolved https://github.com/Azure/azure-sdk-tools/issues/12852

**Key Changes**:
- Move `BuildAsync` method from `SdkBuildTool/CustomizedCodeUpdateTool` to `LanguageService` base class
- Updated tests

**Test Screenshot**:

image

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageService.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageService.cs
@@ -19,14 +19,9 @@ public sealed partial class JavaScriptLanguageService : LanguageService
         ICommonValidationHelpers commonValidationHelpers,
         IFileHelper fileHelper,
         ISpecGenSdkConfigHelper specGenSdkConfigHelper)
+        : base(processHelper, gitHelper, logger, commonValidationHelpers, fileHelper, specGenSdkConfigHelper)
     {
         this.npxHelper = npxHelper;
-        base.processHelper = processHelper;
-        base.gitHelper = gitHelper;
-        base.logger = logger;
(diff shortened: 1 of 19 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageService.cs
```diff
@@ -17,14 +17,16 @@ public JavaScriptLanguageService(
         IGitHelper gitHelper,        
         ILogger<LanguageService> logger,
         ICommonValidationHelpers commonValidationHelpers,
-        IFileHelper fileHelper)
+        IFileHelper fileHelper,
+        ISpecGenSdkConfigHelper specGenSdkConfigHelper)
     {
         this.npxHelper = npxHelper;
         base.processHelper = processHelper;
```
weshaggard: nit: It is probably worth having a constructor in the base class to set these base specific variables.

weshaggard (approved):

benbp (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/JavaScriptLanguageService.cs
raych1: Added the constructors to the base class.

PR #13556, opened 2026-01-13
Add login-to-github script and DevOps template

weshaggard:
- Switch to using a github app instead of our bot PAT
- You will need a token for each github org you want to gain access too
- You will need to grant the new github app permission to the repos you want to use this automation

### DIFF
--- a/eng/common/scripts/login-to-github.ps1
+++ b/eng/common/scripts/login-to-github.ps1
@@ -89,6 +89,14 @@ function New-GitHubAppJwt {
       --algorithm RS256 `
       --digest $Base64Value | ConvertFrom-Json
 
+  if ($LASTEXITCODE -ne 0) {
+    throw "Failed to sign JWT with Azure Key Vault. Error: $SignResult"
+  }
+
+  if (!$SignResultJson.signature) {
+    throw "Azure Key Vault response does not contain a signature. Response: $($SignResultJson | ConvertTo-Json -Compress)"
+  }
+
   $Signature = $SignResultJson.signature
   return "$UnsignedToken.$Signature"
 }
@@ -104,7 +112,7 @@ function Get-GitHubInstallationId {
     $headers = Get-Headers -Jwt $Jwt -ApiVersion $ApiVersion
 
     $uri = "$ApiBase/app/installations"
-    $resp = Invoke-RestMethod -Method Get -Headers $headers -Uri $uri -TimeoutSec 30
+    $resp = Invoke-RestMethod -Method Get -Headers $headers -Uri $uri -TimeoutSec 30 -MaximumRetryCount 3
 
     $resp | Foreach-Object { Write-Host "  $($_.id): $($_.account.login) [$($_.target_type)]" }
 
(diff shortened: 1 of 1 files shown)


eng/common/scripts/login-to-github.ps1
```diff
@@ -0,0 +1,172 @@
...
+  if ($null -ne $env:SYSTEM_TEAMPROJECTID) {
+    Write-Host "##vso[task.setvariable variable=$variableName;issecret=true]$installationToken"
+    Write-Host "Azure DevOps variable '$variableName' has been set (secret)."
+  }
+
+  try {
+    Write-Host "`n--- gh auth status ---"
+    $gh_token_value_before = $env:GH_TOKEN
+    $env:GH_TOKEN = $installationToken
+    & gh auth status
```
benbp, line 172: Need to set `$PSNativeCommandUseErrorActionPreference = $true` if you want this to hit the catch block warning.

weshaggard: Even adding that it doesn't seem to hit that catch block. Either way status is more informational than anything. I probably should remove the catch block isntead.

eng/common/scripts/login-to-github.ps1
```diff
@@ -0,0 +1,169 @@
...
+        [Parameter(Mandatory)][string] $Jwt,
+        [Parameter(Mandatory)][string] $ApiBase,
+        [Parameter(Mandatory)][string] $ApiVersion,
+        [Parameter(Mandatory)][string] $InstallationTokenOwner
+    )
+
+    $headers = Get-Headers -Jwt $Jwt -ApiVersion $ApiVersion
+
+    $uri = "$ApiBase/app/installations"
+    $resp = Invoke-RestMethod -Method Get -Headers $headers -Uri $uri -TimeoutSec 30
```
raych1: Should we add retry for GH API intermittent failures?

eng/common/scripts/login-to-github.ps1
benbp: It should, if fully logged out:

```
$ $ErrorActionPreference = 'Stop'
$ $PSNativeCommandUseErrorActionPreference = $true
$ gh auth status
You are not logged into any GitHub hosts. To log in, run: gh auth login
NativeCommandExitException: Program "gh" ended with non-zero exit code: 1.
$ try { gh auth status } catch { echo ERROR }
You are not logged into any GitHub hosts. To log in, run: gh auth login
ERROR
```

eng/common/scripts/login-to-github.ps1
benbp: If using `-MaximumRetryCount` should probably add `-RetryIntervalSec` greater than 1 to avoid potential secondary rate limits (though they may not apply to this API route).

weshaggard: I will add retry. RetryInternalSec defaults to 5 so I think we are good there. https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-restmethod?view=powershell-7.5#-retryintervalsec

raych1 (approved):

PR #13559, opened 2026-01-13
Update NuGet.Config to include Azure SDK feed

weshaggard:
Added Azure SDK feed and comment regarding package sources.

### DIFF
--- a/tools/azsdk-cli/README.md
+++ b/tools/azsdk-cli/README.md
@@ -1,4 +1,3 @@
-# For testing CI
 # Azure SDK CLI and MCP server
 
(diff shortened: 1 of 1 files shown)


tools/azsdk-cli/README.md
```diff
@@ -1,3 +1,4 @@
+# For testing CI
```
weshaggard:
```suggestion
```

Undo before merging

weshaggard:
I've tested the azsdk-cli pipeline which works but I've not tested all the other pipelines in this repo but I expect them to work so I'm going to merge this and if we see issues we can address them.

PR #13561, opened 2026-01-13
Enabled multiple runs on windows

### LINKED ISSUE #13255: azsdk MCP tool could not be run in two vscode process
Image

raych1:
Fixed #13255 

**Key Changes**:
- Run the `exe` from the temporary installed location and log warnings

**Test Screenshot**:

image

### DIFF
--- a/eng/common/mcp/azure-sdk-mcp.ps1
+++ b/eng/common/mcp/azure-sdk-mcp.ps1
@@ -146,14 +146,14 @@ try {
     $updateSucceeded = $true
 }
 catch {
-    if ($Run) {
-        # In MCP mode, warn and fall back to the existing installed version
+    if ($Run -and (Test-Path $exeDestination)) {
+        # In MCP mode and the executable exists, warn and fall back to the existing installed version
         log -warn "Could not update '$exeDestination': $($_.Exception.Message)"
         log -warn "Falling back to the currently installed version."
     }
     else {
-        # In update-only mode, exit with error
-        log -err "Could not update '$exeDestination': $($_.Exception.Message)"
+        # In update-only mode or the executable does not exist, exit with error
+        log -err "Could not install or update '$exeDestination': $($_.Exception.Message)"
         exit 1
     }
 }


eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -138,9 +138,25 @@ if (-not (Test-Path $toolInstallDirectory)) {
...
+# In that case, fall back to running from the temp directory.
+$exeToRun = $exeDestination
+try {
+    Copy-Item -Path $tempExe -Destination $exeDestination -Force
+}
+catch {
+    if ($_.Exception.Message -match 'being used by another process') {
+        log -warn "Could not update '$exeDestination' because it is currently running. Using temporary copy instead."
+        log -warn "The update will be applied the next time the tool is installed when no instance is running."
+        $exeToRun = $tempExe
```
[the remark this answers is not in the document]
raych1: The temp folder we run from is the freshly installed tool path and should be fine in terms of the concern of being updated.

eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -138,9 +138,27 @@ if (-not (Test-Path $toolInstallDirectory)) {
 }
 $exeName = Split-Path $tempExe -Leaf
 $exeDestination = Join-Path $toolInstallDirectory $exeName
-Copy-Item -Path $tempExe -Destination $exeDestination -Force
```
weshaggard, line 141: Why not just warn that we cannot update it because it is currently running and then run the existing install?

raych1: This is not about the update experience. The core issue is user cannot start two MCP server on windows. In practice, users would like to use MCP tools in multiple VSCode instances, having one opened with the spec repo and others opened with SDK repo.

With this approach, it doesn't impact the current already running instance. It also allows the second instance to be started successfully from the newly installed location instead of failing because of the existing instance.
CC: @benbp

benbp: > The core issue is user cannot start two MCP server on windows

What is this issue, can you elaborate? I don't see why we can't run the same exe twice for separate processes. Is there a port collision or something?

I agree that we should just warn and continue with the current version if there's a pre-existing server instance running.

benbp:
Also not sure why we aren't capturing the inner warning from your screenshot and wrapping the exception in json-rpc format. I'll look into it.

eng/common/mcp/azure-sdk-mcp.ps1
raych1: The issue is line 141 `copy-item` fails on windows as the `exe` is locked when it's running.

benbp: Yes, in that case I think we should do what Wes suggested, not copy to a new location.

raych1: My change doesn't copy it to a new location. Instead, it tries to run it from the installed location.

weshaggard: It should run the one from the destination directory. Why are we trying to copy blindly if it isn't an update scenario we shouldn't need to download or copy anything.

raych1: Currently when this script runs, it firstly installs the `azsdk` package to the `azsdk-install-$($guid)` sub-folder under the `temp` folder. Then copy the `azsdk.exe` to the destination folder. The reason is commented at https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/mcp/azure-sdk-mcp.ps1#L100 

The proposed change is to skip copying the newly installed `azsdk.exe` to the destination if it's locked by the running process. And run the `azsdk.exe` from the `azsdk-install-$($guid)` sub-folder where it is installed to.

benbp: I think that introduces a lot of complexity around knowing what version is installed, could leave around a lot of exe bloat in the install directory, and will also create confusion as we can still only have one `azsdk` executable resolve in `PATH` if run by the user for CLI mode. So they may think they have updated the CLI in PATH but actually they didn't.

We should only ever have one version installed in one place, it's fine if the update fails with a warning and we fallback to the current installed version. However, if the user runs in update only mode (no `-Run` parameter) then we should fail the script with an error.

raych1: I agreed the change should be scoped to the MCP mode only. Updated.

raych1: Image

benbp: Sorry I should have clarified better, I mean it for both mcp and cli. I do not think we should be doing temp installs at all.

eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -138,9 +138,42 @@ if (-not (Test-Path $toolInstallDirectory)) {
...
+# Try to copy the new version
+$updateSucceeded = $false
+try {
+    Copy-Item -Path $tempExe -Destination $exeDestination -Force
+    $updateSucceeded = $true
+}
+catch {
+    # Detect a file-in-use error in a locale-independent way.
+    # ERROR_SHARING_VIOLATION => 0x80070020 => -2147024864
+    if (($_.Exception -is [System.IO.IOException]) -and ($_.Exception.HResult -eq -2147024864)) {
```
weshaggard: Do we need to be this specific? I think if we fail for any reason then we should just log the and continue.

eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -138,9 +138,42 @@ if (-not (Test-Path $toolInstallDirectory)) {
...
-Copy-Item -Path $tempExe -Destination $exeDestination -Force
 
-log "Package $package is installed at $exeDestination"
+# Try to copy the new version
+$updateSucceeded = $false
+try {
+    Copy-Item -Path $tempExe -Destination $exeDestination -Force
+    $updateSucceeded = $true
+}
+catch {
```
weshaggard, line 148: Are we sure that Copy-Item throws in this context?

raych1: Yes, it is. The `catch` section code was executed.

Image

eng/common/mcp/azure-sdk-mcp.ps1
raych1: I agree that the specific error can be removed. But I think for the `update` mode, it shouldn't continue and should log error and exit. For `mcp` mode, it only logs the warning.

eng/common/mcp/azure-sdk-mcp.ps1
raych1: Based on [the comment](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/mcp/azure-sdk-mcp.ps1#L100), the temp install pattern was added to avoid dumping other release zip contents (besides the exe) into the user's `bin` directory. If we remove temp installs and extract directly to ~/bin, those extra files will end up there.

eng/common/mcp/azure-sdk-mcp.ps1
```diff
@@ -138,9 +138,42 @@ if (-not (Test-Path $toolInstallDirectory)) {
...
+    }
+}
+
+# Clean up temp directory
+if (Test-Path $tempInstallDirectory) {
+    Remove-Item -Path $tempInstallDirectory -Recurse -Force -ErrorAction SilentlyContinue
+}
+
+if ($updateSucceeded) {
+    log "Package $package is installed at $exeDestination"
```
benbp: Not sure why I originally called it package...


```suggestion
    log "Executable $package is installed at $exeDestination"
```

eng/common/mcp/azure-sdk-mcp.ps1
raych1: I removed the error filtering and have made the logic applied for all exceptions.

image

benbp: @weshaggard @raych1 I worry about just ignoring all errors and think it's better to check for the exception type like was being done here. Most people are not going to see these warnings logs and will just continue running on older versions without realizing if they have a real issue installing latest. At least checking for the active exe error reduces the surface area of this possibility.

weshaggard: Do we expect the copy to fail in many other cases? Also, if this isn't an update it won't matter and you will still need to restart in order to get a new version.

raych1: I would say the other type of exception might happen very rarely at least I've never seen the other exceptions. Users may go back to check the logs if finding gaps on the current instance.

benbp:
@raych1 I'm just realizing, is there a possibility here where the copy fails (e.g. directory permissions) but the exe has never been installed before, so we still try to run the program? Might need a test-path condition alongside the "are we updating" condition.

raych1:
> @raych1 I'm just realizing, is there a possibility here where the copy fails (e.g. directory permissions) but the exe has never been installed before, so we still try to run the program? Might need a test-path condition alongside the "are we updating" condition.

@benbp Good catch! I updated the condition to ensure the executable file exist.

PR #13562, opened 2026-01-13
Generalize LinkWorkItemAsChildAsync to CreateWorkItemRelationAsync

### LINKED ISSUE #13507: Link a work item to a given parent work item
- Very similar to this https://github.com/Azure/azure-sdk-tools/blob/be21153dc65cb65fe291036234290643b36c5b09/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs#L450
- Given the parents work item id and the child url we link the parent to the child which is what the method above does.

### LINKED ISSUE #13542: Creating a work item relation
- Similar to `LinkWorkItemAsChildAsync` but is less specific and can be used for any relationship. 
- https://github.com/jeo02/azure-sdk-tools/blob/b489a63ff653bc4c6870e10f738dc21de330d243/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs?plain=1#L450C1-L450C83
- Convert `LinkWorkItemAsChildAsync` into `CreateWorkItemRelation`
- Will likely need to recreate the cli command happening [here](https://github.com/Azure/azure-sdk-tools/blob/87d4a0801b6249d728175720a7a3c8bb3eda7a5d/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1#L438) with something of our own. Can be done in the same method. 
- Implementation of the CLI command can be found [here](https://github.com/Azure/azure-devops-cli-extension/blob/a4a2615e8fa2e5de3072ff0a49f3dd70c0f17fa0/azure-devops/azext_devops/dev/boards/relations.py#L24)

jeo02:
- Allows us to attach any relation between work items with either the url or id of the target.
- Will be needed to update the existing create work item to match functionality of the PowerShell script. 
- Equivalent PowerShell function can be found [here](https://github.com/Azure/azure-sdk-tools/blob/87d4a0801b6249d728175720a7a3c8bb3eda7a5d/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1#L394) and mostly copies over the logic from the `az board work-item relation add` that can be found [here](https://github.com/Azure/azure-devops-cli-extension/blob/a4a2615e8fa2e5de3072ff0a49f3dd70c0f17fa0/azure-devops/azext_devops/dev/boards/relations.py#L24).
- Resolves #13507
- Resolves #13542


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
```diff
@@ -447,31 +447,81 @@ private async Task<WorkItem> CreateWorkItemAsync(ReleasePlanDetails releasePlan,
             return workItem;
         }
```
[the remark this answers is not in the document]
jeo02: ill just make it a if, else if so that if both are provided we use id first.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
```diff
@@ -450,31 +447,80 @@ private async Task<WorkItem> CreateWorkItemAsync(ReleasePlanDetails releasePlan,
...
-                // Add work item as child of release plan work item
-                var jsonLinkDocument = new Microsoft.VisualStudio.Services.WebApi.Patch.Json.JsonPatchDocument
+                throw new Exception("To create work item relation, either Target ID or Target URL must be provided.");
+            }
+
+            var workItemClient = connection.GetWorkItemClient();
+
+            // Resolve relation type system name/reference
+            // ex: Child, Parent, Related, etc map to the appropriate name.
+            var relationTypes = await workItemClient.GetRelationTypesAsync();
```
weshaggard: Does this make an API call? If so we might consider caching it to avoid a lot of calls.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
```diff
@@ -450,31 +447,80 @@ private async Task<WorkItem> CreateWorkItemAsync(ReleasePlanDetails releasePlan,
...
+                    Path = "/relations/-",
+                    Value = new WorkItemRelation
+                    {
+                        Rel = relationTypeSystemName,
+                        Url = targetUrl
+                    }
+                });
             }
+
+            await workItemClient.UpdateWorkItemAsync(patchDocument, id);
```
weshaggard: Doesn't the update already return the new workitem?

jeo02: oh yea it does, didn't notice that.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
```diff
@@ -111,11 +111,16 @@ public interface IDevOpsService
...
     {
         private static readonly string RELEASE_PLANER_APP_TEST = "Release Planner App Test";
+        private Lazy<Task<List<WorkItemRelationType>>>? _cachedRelationTypes;
+
         private static readonly string[] SUPPORTED_SDK_LANGUAGES = { "Dotnet", "JavaScript", "Python", "Java", "Go" };
 
         [GeneratedRegex("\\|\\s(Beta|Stable|GA)\\s\\|\\s([\\S]+)\\s\\|\\s([\\S]+)\\s\\|")]
         private static partial Regex SdkReleaseDetailsRegex();
 
+        private Lazy<Task<List<WorkItemRelationType>>> GetCachedRelationTypes() =>
```
weshaggard: Why does this need to be a Lazy<Task>? Why not just a List<WorkItemRelationType>?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
```diff
@@ -450,31 +455,78 @@ private async Task<WorkItem> CreateWorkItemAsync(ReleasePlanDetails releasePlan,
...
-                // Add work item as child of release plan work item
-                var jsonLinkDocument = new Microsoft.VisualStudio.Services.WebApi.Patch.Json.JsonPatchDocument
+                throw new Exception("To create work item relation, either Target ID or Target URL must be provided.");
+            }
+
+            var workItemClient = connection.GetWorkItemClient();
+
+            // Resolve relation type system name/reference
+            // ex: Child, Parent, Related, etc map to the appropriate name.
+            var relationTypes = await GetCachedRelationTypes().Value;
```
weshaggard: I would probably hide this call inside of ResolveRelationTypeSystemName function given it is the only thing that uses it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
jeo02: True I was thinking thread safety initially but since each DevOpsService would have it's own instance of the cache its pointless.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
```diff
@@ -450,31 +455,79 @@ private async Task<WorkItem> CreateWorkItemAsync(ReleasePlanDetails releasePlan,
...
+                        Rel = relationTypeSystemName,
+                        Url = targetUrl
+                    }
+                });
+            }
+
+            return await workItemClient.UpdateWorkItemAsync(patchDocument, id);
+        }
+
+        private async Task<string> ResolveRelationTypeSystemName(string relationType)
```
weshaggard, line 519: This is probably fine for now but we can also consider making an enum for the relationTypes we want to support and build a static mapping for them to avoid the need to making this mapping.

weshaggard (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
jeo02: https://github.com/Azure/azure-sdk-tools/issues/13586

PR #13566, opened 2026-01-13
Add telemetry for CLI mode

benbp:
This adds some of the missing pieces to enable activity events for CLI command invocations to be uploaded to azure monitor.

In addition, the telemetry classes have been cleaned up and unified a bit.

Finally, I've removed logger uploads to azure monitor as I don't think we need them. @praveenkuttappan let me know if I'm missing anything about why we would want this data. I could easily add it back.


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Telemetry/TelemetryService.cs
```diff
@@ -91,11 +57,16 @@ public static void RegisterServerTelemetry(IServiceCollection services, bool deb
 
         await _isInitialized.Task;
 
-        var activity = Parent.StartActivity(activityId);
+        var activity = Parent.StartActivity(activityId, ActivityKind.Server);
```
praveenkuttappan: This will require a change in datax side to map app insights into new kusto DB and PowerBi will also require a change. right? Do we want to do that change?

benbp: Defaulting back to `Internal` and we can migrate both MCP and CLI telemetry when it's convenient.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/MCPToolBase.cs
```diff
@@ -46,37 +42,19 @@ public async Task<int> InstrumentedCommandHandler(Command command, ParseResult p
...
-                activity?.SetStatus(ActivityStatusCode.Ok);
-            }
-            else
-            {
-                activity?.SetStatus(ActivityStatusCode.Error);
-            }
-
-            if (activity != null)
-            {
-                AddCustomTelemetryFromResponse(activity, response);
```
praveenkuttappan, line 78: Custom properties were added to activity to tag the language, package name, typespec project sdk type etc to top level in telemetry. Deleting this line will stop tagging these properties in telemetry. right?

praveenkuttappan (approved):

PR #13585, opened 2026-01-14
[doc-warden] Don't traverse excluded directories

### LINKED ISSUE #13207: JS readme check timing out
specifically for JS I keep seeing timeouts on readme validate, unsure if its an env problem or not

timovv:
To fix #13207

### DIFF
--- a/packages/python-packages/doc-warden/warden/warden_common.py
+++ b/packages/python-packages/doc-warden/warden/warden_common.py
@@ -165,9 +165,13 @@ def walk_directory_for_pattern(
         os.path.normpath(pattern) for pattern in target_patterns
     ]
 
-    # if an omitted path is a directory with a ** we should not scan that path at all, since we don't care about ANY files in that 
-    # directory. This prevents us from going deep into heavily nested folders e.g node_modules.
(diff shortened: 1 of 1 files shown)


scbedd:
This looks eminently reasonable to me. Biggest impact is that we'll lose the list of omitted files that dumps at the end (which we don't look at anyway).

weshaggard:
I don't have an issue with the changes. 

The only thing I'm wondering is if we should simply delete this tool usages at this point. I believe I already stopped calling verify-readme(s).yml in the Java repo for similar reasons. I don't think there is much value add for this tool right now so just removing the usages of it in the JS repo might be a better move.

scbedd:
> I don't have an issue with the changes.
> 
> The only thing I'm wondering is if we should simply delete this tool usages at this point. I believe I already stopped calling verify-readme(s).yml in the Java repo for similar reasons. I don't think there is much value add for this tool right now so just removing the usages of it in the JS repo might be a better move.

I think with this fix, we could actually add it pain-freely back to the Java repo. As it stands I'm going to get this merged and a new version of doc warden released. I know we don't want to spend any effort on this tool but these are pretty simple changes.

Issue #13137, opened 2025-12-02
[APIView] get metadata API


swathipil:
It would be helpful to have a separate endpoint/CLI tool/something to get metadata for a specific review, such as:
- language
- PR/branch the review was generated from
- associated TypeSpec project if CrossLanguage info is already hooked up and readily available

This can be used to identify how and where API review feedback can be applied via a separate MCP tool.

praveenkuttappan:
@swathipil What's the use case of this? Are you looking for a CLI command or MCP tool? If it is an mcp tool: how does this fit in agentic flow?

swathipil:
@praveenkuttappan - I'm working on putting together an MCP tool that will:
1. Get comments from an APIView and github PR
2. Call agent to sort them to see which comments should be applied to TSP, handwritten API, and what are "just discussion". --> "If cannot be done in client.tsp, return list of comments + give instructions to post in the channel asking about how to do it."
3. Call the MCP tool from the azure-rest-api-specs repo so that it can grab the TSP comments and apply them to the appropriate project client.tsp files using the rules the "client customizations" guide [[here](https://github.com/Azure/azure-sdk-tools/blob/09eb5fa59f1af6b78e61ff13ddfbbd53559e5629/eng/common/knowledge/customizing-client-tsp.md)].
4. Commit and push the changes to the appropriate branch.
5. For handwritten API comments, pass them onto the CustomizedCodeTool to be addressed in the appropriate `azure-sdk-for-<language>` repo. etc.

I would expect this tool to be used during the API review process to help service teams address architect feedback. I believe the metadata I listed should be available on the server-side right now and will be helpful with the above process.

AlitzelMendez:
I had an offline conversation with Praveen about the issue, and this was the conclusion 

A generic CLI command that returns:  
- Package name  
- Language  
- Approval status  
- Approved user(s)  
- Package version  
- Linked TypeSpec project name (if available)  

**CLI support:**  
Three input types:  
1. Package name / language combo  
2. Link to review  
3. Link to revision  

My expectation is that for the package name / language combo, we will have an endpoint that returns the review. 
For the links to review or revision, we already have helpers, so we will only need the metadata endpoint.

tjprescott:
I think there is a get metadata endpoint that is internal only. We can consider making this internal/external or a dedicated external endpoint.

AlitzelMendez:
this endpoint is now in production, I will add this to the CLI :)

[...]

PR #13587, opened 2026-01-14
[APIView] Add review metadata API endpoints for external consumers

AlitzelMendez:
first part for: https://github.com/Azure/azure-sdk-tools/issues/13137, after endpoints get to production, I can start working on the cli part 

This PR adds new token-authenticated API endpoints to support external tooling that need to retrieve review and revision metadata from APIView

### New Endpoints

- **`GET /api/reviews/resolve`** - Resolves a review/revision from flexible inputs:
  - Package name + language (+ optional version)
  - APIView URL (with optional revision ID in query string)
  
- **`GET /api/reviews/metadata`** - Returns full review and revision metadata for a given revision ID

### Pattern
1. Call `/resolve` with whatever context you have (URL, package name, etc.)
2. Use the returned `revisionId` for all subsequent API calls

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/LeanControllers/ReviewsTokenAuthController.cs
+++ b/src/dotnet/APIView/APIViewWeb/LeanControllers/ReviewsTokenAuthController.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Threading.Tasks;
 using APIViewWeb.Helpers;
@@ -107,7 +108,7 @@ public class ReviewsTokenAuthController : ControllerBase
     /// <param name="revisionId">The API revision ID (from /resolve endpoint).</param>
     /// <returns>Full review and revision metadata.</returns>
     [HttpGet("metadata", Name = "GetReviewMetadata")]
-    public async Task<ActionResult<ReviewMetadata>> GetMetadata([FromQuery] string revisionId)
+    public async Task<ActionResult<ReviewMetadata>> GetMetadata([FromQuery, Required] string revisionId)
     {
         if (string.IsNullOrEmpty(revisionId))
         {


src/dotnet/APIView/APIViewWeb/Managers/ReviewSearch.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            _logger.LogWarning("Review not found for package: {PackageName}, language: {Language}", packageName,
+                language);
+            return null;
+        }
+
+        APIRevisionListItemModel revision;
+        if (!string.IsNullOrEmpty(version))
+        {
+            var revisions = await _apiRevisionsManager.GetAPIRevisionsAsync(review.Id, version);
```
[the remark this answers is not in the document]
AlitzelMendez: this is the default, adding it is redundant

src/dotnet/APIView/APIViewWeb/LeanModels/ReviewMetadata.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public string ReviewId { get; set; }
+    public string PackageName { get; set; }
+    public string Language { get; set; }
+    public bool IsApproved { get; set; }
+    public string CreatedBy { get; set; }
+    public DateTime CreatedOn { get; set; }
+    public DateTime LastUpdatedOn { get; set; }
+    public RevisionMetadata Revision { get; set; }
+}
```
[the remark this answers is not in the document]
AlitzelMendez: oh this is true!!

src/dotnet/APIView/APIViewWeb/LeanControllers/ReviewsTokenAuthController.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        try
+        {
+            APIRevisionListItemModel revision = await _apiRevisionsManager.GetAPIRevisionAsync(revisionId);
+            if (revision == null)
+            {
+                return NotFound($"Revision '{revisionId}' not found.");
+            }
+
+            ReviewListItemModel review = (await _reviewManager.GetReviewsAsync(new List<string>{revision.ReviewId})).FirstOrDefault();
```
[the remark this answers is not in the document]
AlitzelMendez: what a great idea using the collection expression syntax! way cleaner!!!

src/dotnet/APIView/APIViewWeb/Managers/RevisionResolver.cs
```diff
@@ -0,0 +1,162 @@
...
+            if (revision == null)
+            {
+                _logger.LogWarning("Revision not found for ID: {RevisionId}", revisionId);
+                return null;
+            }
+
+            return new RevisionResolveResult { ReviewId = revision.ReviewId, RevisionId = revision.Id };
+        }
+
+        ReviewListItemModel review = (await _reviewManager.GetReviewsAsync(new List<string>{reviewId})).FirstOrDefault();
```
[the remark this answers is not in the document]
AlitzelMendez: done!

src/dotnet/APIView/APIViewWeb/typespec/reviews.tsp
```diff
@@ -0,0 +1,105 @@
...
+     * @param packageName Package name (e.g., "Azure.Storage.Blobs"). Requires language.
+     * @param language Programming language (e.g., "C#", "Python"). Requires packageName.
+     * @param version Optional package version. If not specified, resolves to latest revision.
+     * @param link APIView URL (e.g., "https://spa.apiview.dev/review/{id}?activeApiRevisionId={revId}")
+     * @returns Resolved IDs for use in subsequent API calls.
+     */
+    @get
+    @route("/resolve")
+    op resolve(
+        @query packageName?: string,
```
tjprescott: `packageName` might not be the best name, since it could be something less specific. If someone sends "storage blobs" that should be an acceptable input. `packageDescription` or `packageQuery` might be better interpretted by an LLM.

AlitzelMendez: make sense! changed

src/dotnet/APIView/APIViewWeb/Managers/RevisionResolver.cs
```diff
@@ -43,10 +43,10 @@ public async Task<RevisionResolveResult> ResolveByPackageAsync(
 
         //TODO: Add package name check after this is done: https://github.com/Azure/azure-sdk-tools/issues/13557
```
tjprescott: For this, I would recommend just calling into a method that contains this TODO. This way the many endpoints that might need this LLM name resolution can be coded *now* to use that, and when it's hooked up you only have to change it in one place.

AlitzelMendez: my expectation is that this is the only method that will call that endpoint as the bottom line of the usage will be finding the     `ReviewId = revision.ReviewId, RevisionId = revision.Id` base on the packageQuery name

AlitzelMendez: but you know, yes! I can extract a section of the code!! :) will do it!

AlitzelMendez: done!!

AlitzelMendez: sorry, at the end undid the changes , BUT I did a renaming of this methods and restructure a little to be clear that this is the only place to resolve packageQuery :)

src/dotnet/APIView/APIViewWeb/LeanControllers/ReviewsTokenAuthController.cs
```diff
@@ -0,0 +1,179 @@
...
+    /// <summary>
+    /// Get full metadata for a revision.
+    /// Requires a revisionId (obtained from the /resolve endpoint).
+    /// </summary>
+    /// <param name="revisionId">The API revision ID (from /resolve endpoint).</param>
+    /// <returns>Full review and revision metadata.</returns>
+    [HttpGet("metadata", Name = "GetReviewMetadata")]
+    public async Task<ActionResult<ReviewMetadata>> GetMetadata([FromQuery] string revisionId)
+    {
+        if (string.IsNullOrEmpty(revisionId))
```
praveenkuttappan, line 113: Whys is revision ID mandatory? Can we take review Id or package name and return metadata from latest automatic revision when revision ID is empty?

src/dotnet/APIView/APIViewWeb/LeanModels/ReviewMetadata.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+
+namespace APIViewWeb.LeanModels;
+
+public class ReviewMetadata
+{
+    public string ReviewId { get; set; }
```
praveenkuttappan, line 7: Can we reuse existing models for review and revision?

src/dotnet/APIView/APIViewWeb/Managers/ResolvePackage.cs
```diff
@@ -0,0 +1,212 @@
...
+using APIViewWeb.LeanModels;
+using APIViewWeb.Managers.Interfaces;
+using APIViewWeb.Models;
+using APIViewWeb.Services;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace APIViewWeb.Managers;
+
+public class ResolvePackage : IResolvePackage
```
praveenkuttappan: This class name slightly confusing or not conveying what it does. It's not resolving the package. It finds the API review for a package. right? Can these methods be in ReviewSearch/Find helper?

src/dotnet/APIView/APIViewWeb/LeanControllers/ReviewsTokenAuthController.cs
AlitzelMendez: The design I have in mind for this is that all the endpoints we are going to be creating for the CLI/MCP will only accept the revisionId.

Why? I don't want to overcomplicate every endpoint by saying that it can receive:
- url
- packageQuery + language + version (optional)
- revisionId
- review + version (optional)

and then checking the scenarios before resolving what to do, having every endpoint doing these checks.

This is the endpoint and people are free to use it directly with the revisionId. For the CLI/MCP implementation, we will receive all those possible combinations:
- url
- packageQuery + language + version (optional)
- revisionId

and then resolve accordingly 
- If it is a URL, we can either call the resolver or resolve it in the CLI code as it is simple
- If it is any combination of (packageQuery + language + version (optional)), we will make a call to our endpoint that resolves that and then call the endpoint that returns the data
- If it is revisionId, call the endpoint directly

The downside is that this could lead to having two calls, but it also allows us to have a clear definition for every endpoint and not explaining all the possible parameters which becomes hard to communicate to the user.

For the MCP case, the expectation is that it remembers the revisionId in the future... but if this is not achievable, we can always get the revisionId.

src/dotnet/APIView/APIViewWeb/LeanModels/ReviewMetadata.cs
AlitzelMendez: That would mean returning the whole review and revision and adding the `PullRequestRepository`. I don't think we want to return all that information, considering that there is duplicated data between Review and Revision, and the review/revision objects contain much more data than just the metadata (such as history, codeFileName, codeFileId, etc.). I think exposing all our data would require careful consideration. is there any reason do you prefer that approach?

src/dotnet/APIView/APIViewWeb/Managers/ResolvePackage.cs
AlitzelMendez: sure, changed!

PR #13593, opened 2026-01-15
Match Create Work Item in CLI and Powershell

### LINKED ISSUE #13539: Update Existing Create Work Item to match functionality of PS
Currently there is an existing `CreateWorkItemAsync` method that provides overlapping functionality. However, it is missing some pieces. 
The following fields either come in as extra parameters or inside of the release plan object.
- Iteration -> System.IterationPath
- Area -> System.AreaPath
- AssignedTo -> System.AssignedTo
- Fields -> Dictionary of field to value mapping. 
- Tag -> Sytem.Tags

The following should just be extra parameters as they will be used in another method called `CreateWorkItemRelation`.
- ParentId
- RelatedId

I will remove `OuputCommand` parameter since it seems to be for printing output which is not necessary anymore.

jeo02:
Currently there is an existing CreateWorkItemAsync method that provides overlapping functionality. However, it is missing some extra fields passed in and it has an option to create work item relations. Resolves #13539
- Add in the create relations with optional parameters:
ParentId
RelatedId
- Add in the following fields
Language -> Custom.Language
Package -> Custom.Package
GroupId -> Custom.GroupId
PackageDisplayName -> Custom.PackageDisplayName
PackageType -> Custom.PackageType
PackageTypeNewLibrary -> Custom.PackageTypeNewLibrary
PackageVersionMajorMinor -> Custom.PackageVersionMajorMinor
ServiceName -> Custom.ServiceName
PackageRepoPath -> Custom.PackageRepoPath
EpicType -> Custom.EpicType
Title -> System.Title
- Made a new type to hold all the new values with the ones from `ReleasePlannerDetails` as the name would start feeling misleading if I kept adding there. 
- Also placed all the `JsonPatchDocument` logic inside of `GetPatchDocument`, a bit easier to see everything being sent to create work item that way. 

Found Iteration -> System.IterationPath, Area -> System.AreaPath, and parameters i omitted from the PowerShell to not be needed. If they are needed, then we can add them back.


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/AzureDevOps/ApiSpecWorkItem.cs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation.
```
benbp, line 1: I think it would be good to move all these devops models into `Models/AzureDevOps/`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/AzureDevOps/EpicWorkItem.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+using System.Reflection;
+using System.Text;
+using Microsoft.VisualStudio.Services.WebApi.Patch.Json;
+
+namespace Azure.Sdk.Tools.Cli.Models.AzureDevOps
+{
+    public class EpicWorkItem : WorkItemBase
```
praveenkuttappan: Let's rename this to Service work item. Both service and Product is of work item type Epic.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/AzureDevOps/WorkItemBase.cs
```diff
@@ -0,0 +1,78 @@
...
+        public string Tag { get; set; } = string.Empty;
+
+        [FieldName("System.Title")]
+        public string Title { get; set; } = string.Empty;
+
+        public string Description { get; set; } = string.Empty;
+
+        public string Status { get; set; } = string.Empty;
+
+        [FieldName("Custom.PrimaryPM")]
```
praveenkuttappan, line 32: This sound like a field only for Service and Product. Do you see this field being set for package work item?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Attributes/FieldNameAttribute.cs
```diff
@@ -0,0 +1,9 @@
+namespace Azure.Sdk.Tools.Cli.Models
+{
+    [AttributeUsage(AttributeTargets.Property)]
+    public class FieldNameAttribute : Attribute
```
praveenkuttappan, line 4: Let's put this under attribute namespace

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/AzureDevOps/WorkItemBase.cs
jeo02: https://github.com/Azure/azure-sdk-tools/blob/cb50209803cedb891e94a8fe0e515c24e72d6e48/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs#L242

I had gotten it from here but can remove the attribute if that's not something we wish to set. Since it is in the original release plan details I'd assume it's also used for release/api spec

praveenkuttappan (approved):

PR #13610, opened 2026-01-16
[python] update python instructions for environment set up

### LINKED ISSUE #13246: [Bugbash] python script is not called by mcp when generating changelog and updating version
The mcp doesn't call python script as configured in [python sdk repo](https://github.com/Azure/azure-sdk-for-python/blob/48964961551073cde02742378cb5a9a2140b6781/eng/swagger_to_sdk_config.json#L26-L37) to generate changelog and update version.

Image

Here is generated changelog file which obviously does not call python script:

Image

msyyc:
fix https://github.com/Azure/azure-sdk-tools/issues/13246

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
@@ -101,23 +101,23 @@
             {
                 "requirement": "azpysdk",
                 "check": ["azpysdk", "--help"],
-                "instructions": ["Navigate to the Python SDK repository root directory", "Ensure your virtual environment is activated", "python -m pip install eng/tools/azure-sdk-tools[build]", "Check that it was installed with azpysdk --help"],
+                "instructions": ["Navigate to the Python SDK repository root directory", "Ensure your virtual environment named '.venv' is created and activated", "python -m pip install eng/tools/azure-sdk-tools[build]", "Check that it was installed with azpysdk --help"],
                 "reason": "`azpysdk` is used for validating Python SDKs."
             },
             {
                 "requirement": "sdk_generator",
                 "check": ["sdk_generator", "--help"],
-                "instructions": ["Navigate to the Python SDK repository root directory", "Ensure your virtual environment is activated", "python -m pip install eng/tools/azure-sdk-tools[sdk_generator]", "Check that it was installed with sdk_generator --help"]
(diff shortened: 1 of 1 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -101,23 +101,23 @@
             {
                 "requirement": "azpysdk",
                 "check": ["azpysdk", "--help"],
-                "instructions": ["Navigate to the Python SDK repository root directory", "Ensure your virtual environment is activated", "python -m pip install eng/tools/azure-sdk-tools[build]", "Check that it was installed with azpysdk --help"],
+                "instructions": ["Navigate to the Python SDK repository root directory", "Ensure your virtual environment named '.venv' is created and activated", "python -m pip install eng/tools/azure-sdk-tools[build]", "Check that it was installed with azpysdk --help"],
```
raych1, line 104: Is `.venv` named virtual environment used across tools in the python repo? Just want to make sure it's consistent across tools and won't break other cases.

JennyPng: I don't think a specific virtual env name is required by tools, as devs can arbitrarily name their virtual environment , cc @l0lawrence 

So the problem is changelog_generate runs without invoking from the user's venv, where the requirements should be installed?

note that the venv used by the MCP server can be set via environment variable `AZSDKTOOLS_PYTHON_VENV_PATH` and you can run commands using that venv's Python executable

An example on how this is used is in https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/PythonLanguageService.Checks.cs#L30 where the IPythonHelper is used to resolve the venv executable

msyyc: `note that the venv used by the MCP server can be set via environment variable AZSDKTOOLS_PYTHON_VENV_PATH and you can run commands using that venv's Python executable`

@JennyPng Could you show a command to set `environment variable AZSDKTOOLS_PYTHON_VENV_PATH` so that mcp tool could be confirmed invoked within that virtual environment. I tried `export AZSDKTOOLS_PYTHON_VENV_PATH=XXX` but mcp tool is not called as expected in target virtual environment.

JennyPng: for linux/WSL, I believe this worked for me https://stackoverflow.com/questions/65258771/in-ubuntu-wsl-how-can-you-store-permanent-environment-variables

add the env var to the bash_profile and run source on it, and then restart the terminal

PR #13642, opened 2026-01-20
[Inner Loop Customizations] feedback classifier

### LINKED ISSUE #13497: [Client Customization] Integrate the APIView feedback entry point into the CustomizedCodeUpdateTool
### 🧩 Name of the Activity

Add integration for the APIView feedback in the CustomizedCodeUpdateTool

### 🎯 Goal of this Activity

Automate the process for updating client.tsp/code based on APIView/architect feedback to speed up the review process time for service teams.

### 🌟 Champion Scenarios

Context:
- User passes in APIView URL

Results:
- TypeSpec customizations: client.tsp is updated to address any comments that can be addressed via TypeSpec decorators
   -- Success: All comments that require no further changes (like discussion comments/TypeSpec decorators that appropriately address a comment/no linting/build/etc. failures + custom code conflicts from decorator updates are all resolved)
   -- Failure: All comments that cannot be addressed via TypeSpec decorators + require code customizations/build and linting errors will be returned

### 🔧 How is this Activity Done Today?

Currently, service teams manually look through all APIView comments and update the client.tsp. They then validate by compiling the tsp file + generating the SDK.

### 🛠️ Tools Needed to Support This Activity

- APIView get-metadata tool/endpoint/etc.: https://github.com/Azure/azure-sdk-tools/issues/13137
- https://github.com/orgs/Azure/projects/865/views/4?pane=issue&itemId=3790920938&issue=Azure%7Cazure-sdk-tools%7C13498

### LINKED ISSUE #13498: [Client Customizations] feedback/error classifier
Add an MCP service/tool to be integrated into the CustomizedCodeUpdateTool, which feedback/comments/errors/breaking changes/etc., and classifies them into:
* Success: No changes needed (discussion comments)
* Failure: Max iterations exceeded OR stalled progress detected
* TypeSpec Customizations: Issue fixable with client.tsp TSP decorators

Subtask:
* Add tests and ensure classifier produces deterministic output for all languages with:
  - build errors
  - linting errors
  - breaking changes
  - architect feedback
  - PR comments
  - discussion/questions

swathipil:
fixes: #13498
fixes: #13497

Issues created for follow-up tasks:
- #14067
- #14068


.github/skills/customization-classifier/tests/README.md
```diff
@@ -0,0 +1,123 @@
...
+```bash
+pytest test_classifier.py -v
+```
+
+### Run specific test
+```bash
+pytest test_classifier.py -v -k "basic_rename"
+```
+
+## Test Cases
```
swathipil: remove section

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -118,4 +119,100 @@ public static AIFunction CreateListFilesTool(
             "ListFiles",
             description);
     }
+
+    /// <summary>
```
swathipil, line 151: TODO: Is this necessary? Remove and evaluate results.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/FetchWebpageTool.cs
```diff
@@ -0,0 +1,90 @@
+using System.ComponentModel;
```
swathipil: remove file

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Microagents/Tools/GrepSearchTool.cs
```diff
@@ -0,0 +1,119 @@
+using System.ComponentModel;
```
swathipil: remove file

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,293 @@
...
+        string? serviceName = null,
+        string? codeCustomizationDocUrl = null,
+        CancellationToken ct = default)
+    {
+        if (customizableItems.Count == 0)
+        {
+            return true;
+        }
+
+        // Stage 1: Batch classify all items in a single LLM call
```
swathipil: // TODO: Test accuracy/speed of single item vs. multiple items vs. all items for larger batches to determine "ideal" batch size

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
swathipil: leaving in b/c grep search will be more optimal for checking pattterns in build log errors against package files

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -120,150 +180,640 @@ public override async Task<CommandResponse> HandleCommand(ParseResult parseResul
...
+
+                // Step 2: First classification (empty context - determines what needs TypeSpec work)
+                logger.LogInformation("=== First Classification ===");
+                await classifier.ClassifyAsync(customizable, globalContext.ToString(), language, null, codeCustomizationDocUrl, ct);
+                MoveItemsByStatus(customizable, success, failure);
+
+                // TODO: TESTING - Output results and return early
+                OutputResults(customizable, success, failure, globalContext.ToString());
+                return BuildResponse(success, failure, customizable);
+
```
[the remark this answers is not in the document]
swathipil: will be updated later when other sections are added

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -120,150 +180,640 @@ public override async Task<CommandResponse> HandleCommand(ParseResult parseResul
...
     }
 
-    [McpServerTool(Name = CustomizedCodeUpdateToolName), Description("Update customized TypeSpec-generated client code")]
-    public Task<CustomizedCodeUpdateResponse> UpdateAsync(string commitSha, string packagePath, CancellationToken ct = default)
-        => RunUpdateAsync(commitSha, packagePath, ct);
+    [Description("Classify feedback and update customized TypeSpec-generated client code")]
+    public Task<CommandResponse> ClassifyAndUpdate(
+        string commitSha,
+        string packagePath,
+        string tspProjectPath,
```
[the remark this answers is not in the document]
swathipil: removed

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
swathipil: | Batch Size (40 comments available total)   | Accuracy | Duration | Items/sec | LLM Calls |
|--------------|----------|----------|-----------|-----------|
| 1            | 82.5%    | 714.7s   | 0.06      | 40        |
| 5            | 82.5%    | 216.5s   | 0.18      | 8         |
| 100 (all)    | 100%     | 235.9s   | 0.17      | 1         |

## Key Findings

- **Large batches = best accuracy** (100% vs 82.5%)
- **Large batches = best speed** (≈3× faster than batch size 1)
- Interestingly, smaller batches miss the same patterns — `@access` decorator items fail without sufficient context

- Updated batch size to >40.
- If the @access decorator items fail in practice, the prompt/ref doc may need to be updated.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/PlainTextFeedbackItem.cs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Azure.Sdk.Tools.Cli.Models;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Sdk.Tools.Cli.Helpers.ClientCustomization;
```
samvaity: These look like models to me than helpers?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -118,4 +119,81 @@ public static AIFunction CreateListFilesTool(
...
+                                    break;
+                                }
+                            }
+                        }
+                        if (matches.Count >= maxResults)
+                        {
+                            break;
+                        }
+                    }
+                    catch { /* Skip unreadable files */ }
```
samvaity: Consider adding a counter for skipped files or at least logging so that we log if an exception occurs.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -118,4 +119,81 @@ public static AIFunction CreateListFilesTool(
...
             description);
     }
+
+    /// <summary>
+    /// Creates a GrepSearch tool that searches for patterns in files within a base directory.
+    /// </summary>
+    /// <param name="baseDir">The base directory for relative path resolution.</param>
+    /// <param name="description">Optional custom description for the tool.</param>
+    /// <returns>An AIFunction that searches files for patterns.</returns>
+    public static AIFunction CreateGrepSearchTool(
```
samvaity, line 157: Good functional capability, I think this will help in other works too.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -118,4 +119,81 @@ public static AIFunction CreateListFilesTool(
...
+        string baseDir,
+        string description = "Search for patterns in files within the project")
+    {
+        return AIFunctionFactory.Create(
```
samvaity, line 161: We should add some directory exclusions
  // Directories to skip during recursive search
    HashSet<string> ignoredDirs = new(StringComparer.OrdinalIgnoreCase)
    {
        ".git", "node_modules", "bin", "obj", "dist", ".vs", ".idea",
        "packages", "TestResults", "__pycache__", ".tox", "target"
    };

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -118,4 +119,81 @@ public static AIFunction CreateListFilesTool(
...
+                if (!ToolHelpers.TryGetSafeFullPath(baseDir, path, out var searchPath))
+                {
+                    throw new ArgumentException("The provided path is invalid or outside the allowed base directory.");
+                }
+                if (!Path.Exists(searchPath))
+                {
+                    throw new ArgumentException($"Path {path} does not exist", nameof(path));
+                }
+
+                var regex = isRegex ? new Regex(pattern, RegexOptions.IgnoreCase | RegexOptions.Compiled) : null;
```
samvaity:
```suggestion
                 // Regex with timeout to prevent ReDoS from LLM-generated patterns
            Regex? regex = null;
            if (isRegex)
            {
                try
                {
                    regex = new Regex(pattern, RegexOptions.IgnoreCase, TimeSpan.FromMilliseconds(500));
                }
                catch (ArgumentException ex)
                {
                    throw new ArgumentException($"Invalid regex pattern: {ex.Message}", nameof(pattern));
                }
            }
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -118,4 +119,81 @@ public static AIFunction CreateListFilesTool(
...
+                var files = File.Exists(searchPath)
+                    ? [searchPath]
+                    : Directory.GetFiles(searchPath, "*", SearchOption.AllDirectories).Take(2000).ToArray();
```
samvaity: 🤔 @m-redding C# wise do you know if there is a better way to do this?
is it `Directory.EnumerateFiles` ?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -118,4 +119,81 @@ public static AIFunction CreateListFilesTool(
...
+                        foreach (var line in File.ReadLines(file))
+                        {
+                            lineNum++;
+                            if (isRegex ? regex!.IsMatch(line) : line.Contains(pattern, StringComparison.OrdinalIgnoreCase))
+                            {
+                                matches.Add(new
+                                {
+                                    FilePath = Path.GetRelativePath(baseDir, file),
+                                    LineNumber = lineNum,
+                                    Content = line.Trim()
```
samvaity: I think this will remove meaningful indentation in code files. Is there a specific need to do this?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/PlainTextFeedbackItem.cs
```diff
@@ -0,0 +1,43 @@
...
+            Context = string.Empty
+        };
+
+        var batch = new FeedbackBatch
+        {
+            Items = new List<FeedbackItem> { feedbackItem }
+        };
+
+        return Task.FromResult(batch);
+    }
```
samvaity:
```suggestion
     ct.ThrowIfCancellationRequested();
    _logger.LogInformation("Processing plain text feedback");

    return Task.FromResult(new FeedbackBatch
    {
        Items = [new FeedbackItem
        {
            Text = _plainText,
            Context = string.Empty
        }]
    });
```

samvaity: actually, is the FeedbackBatch a list of just one item here, don't understand?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/FeedbackClassificationResponse.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        if (!string.IsNullOrEmpty(Message))
+        {
+            sb.AppendLine(Message);
+        }
+
+        return sb.ToString();
+    }
+
+    public class ItemClassificationDetails
```
samvaity, line 57: nit: need docstrings

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/FeedbackClassificationResponse.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        public string ItemId { get; set; } = string.Empty;
```
samvaity, line 62: how does this itemId help?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/FeedbackData.cs
```diff
@@ -0,0 +1,86 @@
...
+    
+    /// <summary>
+    /// Item has been successfully resolved
+    /// </summary>
+    SUCCESS,
+    
+    /// <summary>
+    /// Item cannot be resolved and requires manual intervention
+    /// </summary>
+    FAILURE
```
samvaity: nit: requires_manual_intervention

failure could suggest something else?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Prompts/Templates/FeedbackClassificationTemplate.cs
```diff
@@ -0,0 +1,220 @@
...
+
+/// <summary>
+/// Template for classifying SDK feedback and routing them to the appropriate phase.
+/// Supports batch classification of multiple feedback items in a single LLM call,
+/// with strictly formatted ID-keyed output for robust parsing.
+/// </summary>
+public class FeedbackClassificationTemplate : BasePromptTemplate
+{
+    public override string TemplateId => "feedback-classification";
+    public override string Version => "2.0.0";
```
samvaity:
```suggestion
    public override string Version => "1.0.0";
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Prompts/Templates/FeedbackClassificationTemplate.cs
```diff
@@ -0,0 +1,220 @@
...
+    public override string Description => "Classify SDK feedback items in batch and route to appropriate phase";
+
+    private readonly string? _serviceName;
+    private readonly string? _language;
```
samvaity, line 21: should these be required instead of optional?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Prompts/Templates/FeedbackClassificationTemplate.cs
```diff
@@ -0,0 +1,220 @@
...
+        sb.AppendLine($"""
+        **Current Context:**
+        - Service: {_serviceName ?? "N/A"}
+        - Language: {_language ?? "N/A"}
+
+        **Task:**
+        Classify ALL of the feedback items listed below. For each item, determine the appropriate classification: **TSP_APPLICABLE**, **SUCCESS**, or **FAILURE**.
+        - If the feedback is non-actionable (discussion, informational, "keep as is", or about build/generation succeeding), classify as **SUCCESS**.
+        - If the feedback is actionable AND TypeSpec client customization decorators can address it (based on the reference documentation below), classify as **TSP_APPLICABLE**.
+        - If the feedback is actionable but NO TypeSpec decorators can address it (requires code-level changes), classify as **FAILURE**.
```
samvaity: here failure == requires code-level changes whereas above failure == requires manual intervention.
We should confirm.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /// <summary>
+    /// Classifies feedback items in chunked batch LLM calls. Mutates items in place.
+    /// </summary>
+    public async Task<FeedbackClassificationResponse> ClassifyItemsAsync(
+        List<FeedbackItem> items, 
+        string globalContext, 
+        string? language = null,
+        string? serviceName = null,
```
samvaity, line 70: these should not be optional?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,271 @@
...
+            matchedIds.Add(id);
+            ApplyClassification(item, classification, reason);
+        }
+
+        // Handle any items that weren't in the response
+        foreach (var item in items.Where(i => !matchedIds.Contains(i.Id)))
+        {
+            _logger.LogWarning("Item {Id} was not found in batch classification result. Defaulting to FAILURE.", item.Id);
+            item.Status = FeedbackStatus.FAILURE;
+            item.Context += "\nClassification failed: item missing from batch LLM response";
```
samvaity: consider StringBuilder for better performance.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,271 @@
...
+    /// Reason: explanation
+    /// </summary>
+    private void ParseBatchResult(List<FeedbackItem> items, string result)
+    {
+        var itemLookup = items.ToDictionary(i => i.Id, i => i);
+        var matchedIds = new HashSet<string>();
+
+        // Match blocks like: [some-guid-id]\nClassification: ...\nReason: ...
+        var blockPattern = new Regex(
+            @"\[(?<id>[^\]]+)\]\s*\n\s*Classification:\s*(?<classification>\S+)\s*\n\s*Reason:\s*(?<reason>.+)",
```
samvaity: compiled for every call. Consider making it a static compiled field

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,271 @@
...
+                $"Could not determine spec repo root from TypeSpec project path: {tspProjectPath}",
+                nameof(tspProjectPath));
+        }
+        return repoRoot;
+    }
+
+    /// <summary>
+    /// Loads the TypeSpec client customization guide from eng/common/knowledge/customizing-client-tsp.md.
+    /// </summary>
+    private async Task<string> LoadTspCustomizationGuideAsync(CancellationToken ct)
```
samvaity: Is this loaded every call? Since this is static content we should avoid reloading.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,271 @@
...
+        var result = await _agentRunner.RunAsync(new CopilotAgent<string>
+        {
+            Instructions = prompt,
+            MaxIterations = 25,
+            Tools = specTools
+        }, ct);
+
+        _logger.LogInformation("=== Batch Classification Result ===");
+        _logger.LogInformation("{Result}", result);
+        _logger.LogInformation("=== End Batch Result ===");
```
samvaity: should this be debug?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
m-redding: yeah Directory.EnumerateFiles would be a better way to do this because you can start enumerating over the array before it's done creating the entire collection, whereas with  GetFiles you are waiting to get the entire list. For big searches up to 2000 files like this it would be more efficient

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -118,4 +119,81 @@ public static AIFunction CreateListFilesTool(
...
+                            continue;
+                        }
+
+                        int lineNum = 0;
+                        foreach (var line in File.ReadLines(file))
+                        {
+                            lineNum++;
+                            if (isRegex ? regex!.IsMatch(line) : line.Contains(pattern, StringComparison.OrdinalIgnoreCase))
+                            {
+                                matches.Add(new
```
m-redding: nit you could use a record here to be a little more readable but also fine as is

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -118,4 +119,81 @@ public static AIFunction CreateListFilesTool(
...
+                    throw new ArgumentException($"Path {path} does not exist", nameof(path));
+                }
+
+                var regex = isRegex ? new Regex(pattern, RegexOptions.IgnoreCase | RegexOptions.Compiled) : null;
+                var files = File.Exists(searchPath)
+                    ? [searchPath]
+                    : Directory.GetFiles(searchPath, "*", SearchOption.AllDirectories).Take(2000).ToArray();
+
+                var matches = new List<object>();
+                foreach (var file in files)
```
m-redding: Could consider adding some parallelization here to speed this up

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/IFeedbackItem.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License.
+
+using Azure.Sdk.Tools.Cli.Models;
+
+namespace Azure.Sdk.Tools.Cli.Helpers.ClientCustomization;
+
+/// <summary>
+/// Represents a source of feedback for SDK customization (APIView comments, build errors, etc.)
+/// </summary>
+public interface IFeedbackItem
```
m-redding: Could the preprocessing just be done directly instead of abstracting them out? Since the usage isn't polymorphic and I don't see this used in dependency injection. Also, the implementation of preprocess is very different for the two types of feedback

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
m-redding: Or maybe even adding a "filetype" option to only look at .cs files for example

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,271 @@
...
+        var tspApplicableCount = 0;
+        var classifications = new List<FeedbackClassificationResponse.ItemClassificationDetails>();
+
+        foreach (var item in items)
+        {
+            var classification = item.Status switch
+            {
+                FeedbackStatus.SUCCESS => "SUCCESS",
+                FeedbackStatus.FAILURE => "FAILURE",
+                _ => "TSP_APPLICABLE"
```
m-redding, line 113: curious - why make tsp applicable the default instead of failure?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -268,6 +288,82 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string commitSha
...
+            }
+
+            var classifier = new FeedbackClassifierService(
+                _agentRunner,
+                loggerFactory,
+                typeSpecHelper,
+                tspProjectPath
+            );
+
+            FeedbackBatch feedbackBatch;
```
m-redding: Could this just be a list of feedback items instead of needing to define a type for the batch?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -268,6 +288,82 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string commitSha
...
+            if (!string.IsNullOrWhiteSpace(apiViewUrl))
+            {
+                feedbackBatch = await new APIViewFeedbackItem(apiViewUrl, feedbackService, loggerFactory.CreateLogger<APIViewFeedbackItem>()).PreprocessAsync(ct);
```
m-redding: similar to my other comments but could the work to collect the list of feedback just be done directly here rather than creating an abstraction?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/PlainTextFeedbackItem.cs
swathipil: For now, this is just one item. In the future, we could take a text that has multiple items that we parse into a batch. The main reason for making this a batch right now is to align it with the APIView feedback output, where each item is a comment and the batch is all comments in the revision.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/PlainTextFeedbackItem.cs
swathipil: Models seem to be objects that define data shape. These items also define behavior/have logic in PreprocessAsync, so I thought they should be here.

swathipil: Updating the name so it's clearer that the function of this is to do the preprocessing, rather than act as an "Item".

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Prompts/Templates/FeedbackClassificationTemplate.cs
swathipil: Good catch! Language should be required already and I forgot to update this, but should we also add an extra flag to pass in serviceName if it can't be retrieved from the input?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Prompts/Templates/FeedbackClassificationTemplate.cs
swathipil: OOB: I've categorized code-level changes under "REQUIRES_MANUAL_INTERVENTION" for now, since we haven't incorporated the guidance yet. In the separate PR to add the Code customizations applicable classification, I'll move these under that classification

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/FeedbackClassificationResponse.cs
swathipil: Useful when doing batch classification/customizations. This will be used by the to map the id in the tsp customization result to the corresponding feedback so we can add it to the item's context. It also helps map the classifier's response (which classifies in batches and will include id) to the corresponding feedback item.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/APIViewFeedbackSource.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        _logger.LogInformation("Converted {Count} comments to feedback items", feedbackItems.Count);
+
+        return new FeedbackBatch
+        {
+            Items = feedbackItems,
+            Language = metadata.Language
+        };
+    }
+}
```
m-redding: nit missing a couple newlines

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/PlainTextFeedbackSource.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    public PlainTextFeedbackSource(
+        string plainText,
+        ILogger<PlainTextFeedbackSource> logger)
+    {
+        _plainText = plainText;
+        _logger = logger;
+    }
+
+    public Task<FeedbackBatch> CreateBatchAsync(CancellationToken ct = default)
```
m-redding: This doesn't need to be async

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -208,4 +229,86 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string packagePa
...
+            );
+
+            FeedbackBatch feedbackBatch;
+            if (!string.IsNullOrWhiteSpace(apiViewUrl))
+            {
+                feedbackBatch = await new APIViewFeedbackSource(apiViewUrl, feedbackService, loggerFactory.CreateLogger<APIViewFeedbackSource>()).CreateBatchAsync(ct);
+            }
+            else if (!string.IsNullOrWhiteSpace(plainTextFeedback))
+            {
+                feedbackBatch = await new PlainTextFeedbackSource(plainTextFeedback, loggerFactory.CreateLogger<PlainTextFeedbackSource>()).CreateBatchAsync(ct);
```
m-redding:
```suggestion
                feedbackBatch = new FeedbackBatch
                  {
                      Items = [new FeedbackItem
                      {
                          Text = plainTextFeedback,
                          Context = string.Empty
                      }]
                  };
```
Could this just be done directly? Or are there plans to eventually expand what is done with the plain text?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/APIViewFeedbackSource.cs
```diff
@@ -0,0 +1,59 @@
...
+        string apiViewUrl,
+        IAPIViewFeedbackService feedbackService,
+        ILogger<APIViewFeedbackSource> logger)
+    {
+        _apiViewUrl = apiViewUrl;
+        _feedbackService = feedbackService;
+        _logger = logger;
+    }
+
+    public async Task<FeedbackBatch> CreateBatchAsync(CancellationToken ct = default)
```
m-redding: curious - could this go on APIViewFeedbackService ?

m-redding:
```suggestion
    public async Task<(string, List<FeedbackItem>)> CreateBatchAsync(CancellationToken ct = default)
```
just a thought, it might simplify it a bit since later on they're split back out anyway, but also fine as is

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
samvaity: +1

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -124,16 +132,200 @@ public static AIFunction CreateListFilesTool(
...
+    {
+        try
+        {
+            return enumerate();
+        }
+        catch
+        {
+            return [];
+        }
+    }
```
m-redding:
```suggestion
            IEnumerable<string> subdirectories;
            try
            {
                subdirectories = Directory.EnumerateDirectories(current);
            }
            catch
            {
                continue;
            }
            
            foreach (var subdirectory in subdirectories)
            {
                if (IsIgnoredDirectory(subdirectory))
                {
                    continue;
                }
            
                pending.Push(subdirectory);
            }
        }
    }
```
Could something like this work?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -124,16 +132,200 @@ public static AIFunction CreateListFilesTool(
...
+
+    private static bool IsIgnoredDirectory(string directoryPath)
```
m-redding: nit but since this is only used in one place could it be in line?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,282 @@
...
+
+        var referenceDocContent = await LoadTspCustomizationGuideAsync(ct);
+
+        for (int i = 0; i < chunks.Count; i++)
+        {
+            var chunk = chunks[i];
+            _logger.LogInformation("Processing batch {BatchNum}/{TotalBatches} ({ItemCount} items)", 
+                i + 1, chunks.Count, chunk.Count);
+            await BatchClassifyAsync(chunk, globalContext, language, serviceName, referenceDocContent, ct);
+        }
```
m-redding:
```suggestion
        foreach (var chunk in items.Chunk(_batchSize))
        {
            await BatchClassifyAsync(chunk.ToList(), globalContext, language, serviceName, referenceDocContent, ct);
        }
```
could potentially simplify this if you'd be okay cutting the logs

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,282 @@
...
+        if (items.Count == 0)
+        {
+            return new FeedbackClassificationResponse { Message = "No items to classify.", Classifications = [] };
+        }
+
+        var chunks = items
+            .Select((item, index) => new { item, index })
+            .GroupBy(x => x.index / _batchSize)
+            .Select(g => g.Select(x => x.item).ToList())
+            .ToList();
```
m-redding: Could avoid this ToList enumeration by foreaching on chunks without calling ToList and just removing the log below

m-redding (commented):
just a few suggestions/nits looks pretty good to me

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -124,16 +132,200 @@ public static AIFunction CreateListFilesTool(
...
+                if (IsIgnoredDirectory(subdirectory))
+                {
+                    continue;
+                }
+
+                pending.Push(subdirectory);
+            }
+        }
+    }
+
```
samvaity, line 330: nit: should these be helpers, to provide traversal or do they handle specifics?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/PlainTextFeedbackSource.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using Azure.Sdk.Tools.Cli.Models;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Sdk.Tools.Cli.Helpers.ClientCustomization;
+
+/// <summary>
+/// Feedback source from plain text (e.g., build logs, error messages, user input)
+/// </summary>
+public class PlainTextFeedbackSource
```
samvaity: Why can't we just have one Feedback source and have variations of it if needed? This is basically the entry point correct.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,282 @@
...
+        RegexOptions.Multiline | RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+    private readonly ICopilotAgentRunner _agentRunner;
+    private readonly ILogger<FeedbackClassifierService> _logger;
+    private readonly string _tspProjectPath;
+    private readonly string _specRepoBasePath;
+    private readonly ITypeSpecHelper _typeSpecHelper;
+    private readonly int _batchSize;
+
+    public const int DefaultBatchSize = 50;
```
samvaity, line 50: what does this stand for?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,282 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Applies a classification string and reason to a single feedback item.
+    /// </summary>
+    private void ApplyClassification(FeedbackItem item, string classification, string reason)
+    {
+        FeedbackStatus status;
+        if (classification.Contains("SUCCESS", StringComparison.OrdinalIgnoreCase))
```
samvaity: Seems like this is very easy to misread, what is LLM returns "not success" in that case we are still classifying it as success in that scenario?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,282 @@
...
+            item.AppendContext($"Classification: {classification}\nReason: {reason}", leadingNewLines: 2);
+        }
+        
+        _logger.LogInformation("Item {Id} classified as {Status}: {Reason}", item.Id, status, reason);
+    }
+
+    /// <summary>
+    /// Resolves the spec repo base path by walking up from tspProjectPath.
+    /// </summary>
+    private string GetSpecRepoBasePath(string tspProjectPath)
```
samvaity: Can we use this https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TypeSpecHelper.cs#L104 if and if not consider moving to helpers ?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,282 @@
...
+    /// Applies a classification string and reason to a single feedback item.
+    /// </summary>
+    private void ApplyClassification(FeedbackItem item, string classification, string reason)
+    {
+        FeedbackStatus status;
+        if (classification.Contains("SUCCESS", StringComparison.OrdinalIgnoreCase))
+        {
+            status = FeedbackStatus.SUCCESS;
+        }
+        else if (classification.Contains("REQUIRES_MANUAL_INTERVENTION", StringComparison.OrdinalIgnoreCase))
```
samvaity: also, could we use switch?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
swathipil: I wanted to introduce the batch to keep track of shared data (like language) -- but I can remove it since the assumption is that all feedback will be for the same language rn + there's no other shared data.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -208,4 +229,86 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string packagePa
...
+                plainTextFeedback = await File.ReadAllTextAsync(plainTextFeedbackFile, ct);
+                logger.LogInformation("Read {length} characters from feedback file: {file}", plainTextFeedback.Length, plainTextFeedbackFile);
+            }
+
+            if (string.IsNullOrEmpty(tspProjectPath) || !Directory.Exists(tspProjectPath))
+            {
+                throw new DirectoryNotFoundException($"TypeSpec project path does not exist: {tspProjectPath}");
+            }
+
+            var classifier = new FeedbackClassifierService(
```
samvaity: Should this be instantiated with the tooling?

samvaity: Actually, why is this a service, can it be a helper?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -208,4 +229,86 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string packagePa
...
+            {
+                feedbackBatch = await new APIViewFeedbackSource(apiViewUrl, feedbackService, loggerFactory.CreateLogger<APIViewFeedbackSource>()).CreateBatchAsync(ct);
+            }
+            else if (!string.IsNullOrWhiteSpace(plainTextFeedback))
+            {
+                feedbackBatch = await new PlainTextFeedbackSource(plainTextFeedback, loggerFactory.CreateLogger<PlainTextFeedbackSource>()).CreateBatchAsync(ct);
+            }
+            else
+            {
+                throw new ArgumentException("Either --apiview-url or --plain-text-feedback (or --plain-text-feedback-file) must be provided.");
```
samvaity: Ideally this should be handled in the classifier service and the error must be bubbled up and not handled in the customization tooling.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
swathipil: No strong reason - I can change it to failure. I believe my thinking was that Success/Failure are stopping conditions for the classifier loop, and I wanted to indicate that these items would be going through a classification loop by default. Curious why default to Failure? So that in the case of actual failure, these should all be handled manually?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -208,4 +229,86 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string packagePa
...
+            var resolvedLanguage = language ?? feedbackBatch.Language;
+            if (string.IsNullOrEmpty(resolvedLanguage))
+            {
+                throw new ArgumentException("Language is required but could not be determined from the feedback source. Please specify --language (e.g. python, java, dotnet, go, javascript).");
+            }
+
+            if (string.IsNullOrEmpty(serviceName))
+            {
+                throw new ArgumentException("Service name is required. Please specify --service-name.");
+            }
```
samvaity: If these are truly required move them to tool-level parameter definition. But should remove throwing exception as they are not in context for the tooling.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -208,4 +229,86 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string packagePa
             AppliedPatches = patches
         };
     }
+
+    /// <summary>
+    /// Classifies feedback items from various sources (APIView, build errors, etc.) as TSP_APPLICABLE
+    /// (can be resolved with TSP customizations), SUCCESS (no action needed), or REQUIRES_MANUAL_INTERVENTION (requires manual code customizations).
+    /// </summary>
+    private async Task<FeedbackClassificationResponse> Classify(
```
samvaity, line 227: What is the integration point for this method or is it to be covered in a follow up PR?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
swathipil: I gave it strict instructions to only return one of the 3 specified classifications, so it shouldn't be returning "NOT SUCCESS" and I haven't seen it doing that. Will use .Equals to make sure instead of .Contains though!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
swathipil: There are plans to eventually take and parse different types of plain text input (in my own head 😄), but it makes sense to remove this implementation now and add that in later if we need.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
samvaity: yes, if we are sure about the LLM returns, it would be resilient to use "SUCCESS" => FeedbackStatus.SUCCESS pattern

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/APIViewFeedbackSource.cs
swathipil: Updated this to just return the list of items, and moved GetLanguage into a separate function as per notes above.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
swathipil: sg, we can add logs back in later if needed.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/APIViewFeedbackSource.cs
swathipil: should work, and will clean up extra files

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/ClientCustomization/PlainTextFeedbackSource.cs
swathipil: yes that's correct. I've actually removed both of these source files and am manually creating the plain text feedback items as per Maddy's suggestion (and moved the APIView feedback item processing to the existing APIViewFeedbackService).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
swathipil: IgnoredDirectories has specifics, so I think it would belong here. We can probably extract this if needed though and add the ignored directories as a param.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
swathipil: These is the # of feedback items we send in to the classifier at a time to be classified. With 1 at a time, it takes an excessive amount of time. It seems larger batch size is better, both b/c it processes a little faster and not too large that there are issues with token limits. Also, APIView feedback is often ~50 or below on the higher end.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
swathipil: yes we can, replacing

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
swathipil: I think we should leave these optional for now. Will remove the exceptions.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
swathipil: Summarizing the OOB discussion with Maddy and Sameeksha: Helper classes are more static and the classifier acts more as a service since it depends on some injected dependencies. To align it with "service" expectations:

- the classifier should probably have an interface for DI registration.
- It should be injected as a singleton in [azure-sdk-tools/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs at main · Azur…](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs) 
- the tool should accept an IFeedbackClassifierService in its constructor
- The service would need to be updated to be stateless - so tspProjectPath and batchSize would need to be moved to be parameters on the individual methods rather than part of the ctor

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/FeedbackClassifierService.cs
```diff
@@ -0,0 +1,282 @@
...
+
+namespace Azure.Sdk.Tools.Cli.Services;
+
+/// <summary>
+/// Classifies feedback items using LLM-powered classification.
+/// Uses chunked batch LLM calls to classify items as TSP_APPLICABLE, SUCCESS, or REQUIRES_MANUAL_INTERVENTION.
+/// </summary>
+public class FeedbackClassifierService
+{
+    private static readonly ConcurrentDictionary<string, string> TspCustomizationGuideCache = new(StringComparer.OrdinalIgnoreCase);
```
m-redding, line 40: does this need to be static?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
swathipil: Yes, to be used in a future PR where we can incorporate the classifier -> worker A -> worker B loop.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Services/FeedbackClassifierServiceTests.cs
```diff
@@ -0,0 +1,509 @@
...
+        Assert.That(item2.Status, Is.EqualTo(FeedbackStatus.REQUIRES_MANUAL_INTERVENTION), "Missing items should default to REQUIRES_MANUAL_INTERVENTION");
+        Assert.That(item2.Context, Does.Contain("missing from batch LLM response"));
+    }
+
+    #endregion
+
+    #region Caching Tests
+
+    [Test]
+    public async Task LoadTspCustomizationGuideAsync_CachesContentAfterFirstRead()
```
m-redding, line 339: I kinda feel like this test isn't needed? or maybe it could be replaced with something more like "testing classification doesn't fail if the file is deleted after the first call" that doesn't require the reflection?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Prompts/Templates/FeedbackClassificationTemplate.cs
samvaity: For classifying correctly, we would need language to be required. Not sure about serviceName needs.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -208,4 +219,56 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string packagePa
...
+        string tspProjectPath,
+        string? apiViewUrl,
+        string? plainTextFeedback,
+        string? plainTextFeedbackFile,
```
samvaity, line 231: do we need both?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -208,4 +219,56 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string packagePa
...
+        string? language,
+        string? serviceName,
+        CancellationToken ct)
+    {
+        try
+        {
+            // Read feedback from file if provided
+            if (!string.IsNullOrWhiteSpace(plainTextFeedbackFile))
```
samvaity, line 239: Ideally this is just input and not feedback item correct? Find it confusing as to why we are reading feedback in classify items.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -208,4 +219,56 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string packagePa
...
+            List<FeedbackItem> feedbackItems = [];
+            if (!string.IsNullOrWhiteSpace(apiViewUrl))
+            {
+                feedbackItems = await feedbackService.GetFeedbackItemsAsync(apiViewUrl, ct);
+                language ??= await feedbackService.GetLanguageAsync(apiViewUrl, ct);
+            }
+            else if (!string.IsNullOrWhiteSpace(plainTextFeedback))
+            {
+                feedbackItems = [new FeedbackItem { Text = plainTextFeedback, Context = string.Empty }];
+            }
```
samvaity, line 264: This all feels like it belongs to feedback service gathering/collecting feedback from different sources. I would think `Classify` just receives this as input, is there  a better way to structure this?

samvaity (approved):
Few suggestions, lgtm otherwise.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
```diff
@@ -208,4 +219,56 @@ private async Task<CustomizedCodeUpdateResponse> RunUpdateAsync(string packagePa
...
+                if (!File.Exists(plainTextFeedbackFile))
+                {
+                    throw new FileNotFoundException($"Plain text feedback file does not exist: {plainTextFeedbackFile}");
+                }
+
+                plainTextFeedback = await File.ReadAllTextAsync(plainTextFeedbackFile, ct);
+                logger.LogInformation("Read {length} characters from feedback file: {file}", plainTextFeedback.Length, plainTextFeedbackFile);
+            }
+
+            if (string.IsNullOrEmpty(tspProjectPath) || !Directory.Exists(tspProjectPath))
```
m-redding, line 250: nit should check args at the beginning

m-redding (approved):
looks good to me too! just left a couple nits

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
swathipil: not needed but was good for testing, can cut it down to just the text, and add the file later if needed.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/CustomizedCodeUpdateTool.cs
swathipil: Will remove plainTextFeedbackFile and only keep the string plainText option.

PR #13648, opened 2026-01-20
[AZ SDK CLI] Update package type for .NET

### LINKED ISSUE #12839: Package Info is missing SDK type for  .NET

https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotnetLanguageService.cs#L44

Currently SDK type is set in CI and Engsys tools using the following code, and we should use the same logic to set the properties in MCP package info so values match.
https://github.com/Azure/azure-sdk-for-net/blob/main/eng/scripts/Language-Settings.ps1#L61C25-L61C32

m-redding:
Resolves https://github.com/Azure/azure-sdk-tools/issues/12839


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotnetLanguageService.cs
```diff
@@ -87,69 +102,98 @@ public override async Task<PackageInfo> GetPackageInfo(string packagePath, Cance
...
-                }
+                return (null, null, null);
+            }
+
+            // Parse the identity string:  'pkgPath' 'serviceDir' 'pkgName' 'pkgVersion' 'sdkType' 'isNewSdk' 'dllFolder' 'AotCompatOptOut'
+            var parts = identity.Split(new[] { "' '" }, StringSplitOptions. RemoveEmptyEntries)
+                .Select(p => p.Trim('\'', ' '))
+                .ToArray();
+
+            if (parts.Length >= 5)
```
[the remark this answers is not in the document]
m-redding: We only need the 5th one right now

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/DotnetLanguageService.cs
```diff
@@ -87,69 +102,97 @@ public override async Task<PackageInfo> GetPackageInfo(string packagePath, Cance
...
         }
         catch (Exception ex)
         {
             logger.LogError(ex, "Error reading .NET package info from {packagePath}", packagePath);
-            return (null, null);
+            return (null, null, null);
+        }
+    }
+
+    private (string? Name, string? Version, string? SdkType) FallbackToXmlParsing(string csprojPath)
```
raych1: Do we need this fall back to read 'packageId' and 'version' from the project file? The point is that the `sdkType` won't be set in this case and the downstream tools replying on this would not function. Should we fail fast if the `sdkType` isn't fetched successfully from the `msbuild` call?

m-redding: This is just preserving the current behavior; I left it because I liked that it was very flexible, for example if you didn't have .NET downloaded, or wanted to run in a different repo for some reason, the xml parsing would still work. But it's definitely not strictly necessary since with .NET downloaded and running in azure-sdk-for-net the target will always work, so I can take it out.

m-redding: I also followed the existing pattern here and in the other language checks - instead of failing, just returning null. I'm assuming it's up to the downstream tools to fail if they don't have what they need

raych1 (approved):

PR #13660, opened 2026-01-20
[AZ SDK CLI] fix tsp client instructions

m-redding:
I am not an expert in this area but the instructions for this requirement don't match the actual requirement of the tool. It seems the tool needs a global install in order to pass, but the instructions are for a local install?? Not totally sure but I spent awhile debugging with copilot and this is what finally fixed my issue so that verify setup successfully passed

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
@@ -9,7 +9,7 @@
             {
                 "requirement": "tsp-client >= 0.24.0",
                 "check": ["tsp-client", "--version"],
-                "instructions": ["cd eng/common/tsp-client", "npm ci"]
+                "instructions": ["npm install -g @azure-tools/typespec-client-generator-cli@latest"]
             },
             {
                 "requirement": "tsp >= 1.0.0",
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
@@ -9,7 +9,7 @@
             {
                 "requirement": "tsp-client >= 0.24.0",
                 "check": ["tsp-client", "--version"],
-                "instructions": ["npm install -g @azure-tools/typespec-client-generator-cli@latest"]
+                "instructions": ["npm install -g @azure-tools/typespec-client-generator-cli@0.31.0"]
             },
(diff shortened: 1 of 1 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -9,7 +9,7 @@
             {
                 "requirement": "tsp-client >= 0.24.0",
                 "check": ["tsp-client", "--version"],
-                "instructions": ["cd eng/common/tsp-client", "npm ci"]
+                "instructions": ["npm install -g @azure-tools/typespec-client-generator-cli@0.31.0"]
```
m-redding, line 12: my previous comment and copilot's comment got hidden by the update- summarizing:

The check and instructions are not compatible - these instructions don't work to make the check pass when I was working in a new dev box. This updated instruction was the only thing that finally worked for me to be able to get the check to pass.

copilot commented saying that this instruction doesn't match what's happening in the CI, in that case, maybe the check is incorrect? IDK exactly what's wrong but I can say for sure that the current check/instructions do not work

raych1: The instruction is correct in terms of installing 'tsp-client' from the centralized `package-lock.json`, which is used across repos and pipelines. However, we haven't changed our `tsp-client` usage to use it from the installed location. I created a [task](https://github.com/Azure/azure-sdk-tools/issues/13667) to update the usage.

raych1 (changes requested):
@m-redding thank you for bringing this up. The instruction is correct and we need to change the `tsp-client` usage. See another [task](https://github.com/Azure/azure-sdk-tools/issues/13667) for details.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
m-redding: I tried this on a new dev box and tsp-client --version doesn't work at all when you follow the instructions here, even beyond just what we're doing in the tooling. I read the issue and since I don't have a lot of context in this area I'm not totally sure if that would fix the issue, but either the check or instructions in this file needs to be fixed because it doesn't work

raych1: Correct. The check is supposed to be something like `npm exec --prefix {$TspClientDir} --no -- tsp-client --version`.  CC: @JennyPng

PR #13662, opened 2026-01-21
Update work item parent method in DevOps service 

### LINKED ISSUE #13659: Update Work Item Parent
- Can be found 
https://github.com/Azure/azure-sdk-tools/blob/6e6038ab75feb6b557b0ed360b94ba5413e33ff9/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1#L346
- Need to first remove Item Relation first. Create method `RemoveWorkItemRelationAsync`
- Then `CreateItemRelationAsync` can create the relation I need.

jeo02:
- Resolves #13659
- Added `UpdateWorkItemParentAsync` which updates the parent which the child work item is pointing to. 
- Added 'RemoveWorkItemRelationAsync` to remove the previous parent of the child work item. Can also be useful when removing relations elsewhere. 
- Renamed `ProductWorkItem` to `PackageWorkItem` because I realized that in the powershell the fields were actually for creating a Package Work Item. Product and Service have identical fields so might rename `ServiceWorkItem`
- Powershell equivalent can be found here:
https://github.com/Azure/azure-sdk-tools/blob/be21153dc65cb65fe291036234290643b36c5b09/eng/common/scripts/Helpers/DevOps-WorkItem-Helpers.ps1#L346

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/AzureDevOps/PackageWorkItem.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/AzureDevOps/PackageWorkItem.cs
@@ -4,7 +4,7 @@ using Azure.Sdk.Tools.Cli.Attributes;
 
 namespace Azure.Sdk.Tools.Cli.Models.AzureDevOps
 {
-    public class ProductWorkItem : WorkItemBase
+    public class PackageWorkItem : WorkItemBase
     {
         [FieldName("Custom.Language")]
         public string Language { get; set; } = string.Empty;
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
@@ -1384,7 +1384,7 @@ namespace Azure.Sdk.Tools.Cli.Services
 
         public async Task UpdateWorkItemParentAsync(WorkItemBase child, WorkItemBase parent)
         {
-            if (child.WorkItemId == parent.WorkItemId)
+            if (child.ParentId == parent.WorkItemId)
             {
                 return; // already the parent
             }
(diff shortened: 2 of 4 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/AzureDevOps/PackageWorkItem.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Azure.Sdk.Tools.Cli.Models.AzureDevOps
 {
-    public class ProductWorkItem : WorkItemBase
+    public class PackageWorkItem : WorkItemBase
```
praveenkuttappan, line 7: Can you rename service work item to EpicWorkItem and use for both service and Product?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/DevOpsService.cs
```diff
@@ -1339,5 +1377,19 @@ public async Task<WorkItem> UpdateWorkItemAsync(int workItemId, Dictionary<strin
...
+        private async Task UpdateWorkItemParentAsync(WorkItemBase child, WorkItemBase parent)
+        {
+            if (child.ParentId == parent.WorkItemId)
+            {
+                return; // already the parent
+            }
+
+            // Remove existing parent link
+            // Child must have existing parent link or remove work item relation will fail.
+            await RemoveWorkItemRelationAsync(child.WorkItemId, "Parent", child.ParentId);
```
praveenkuttappan: Do we want to run removal when parentId is 0? A new api spec work item won't have any parent. We link the parent after it's created.

jeo02: yes makes sense, I also added ParentId to default to 0 if not set

praveenkuttappan (approved):

PR #13663, opened 2026-01-21
Refactor VerifySetup to get requirements from code

### LINKED ISSUE #13342: [Bug Bash] azure-rest-api-specs - agent really wants to install typespec/tcgc globally
### Prompts Used

_with tspconfig.yaml in context_
> Generate the TypeScript SDK for me for this TypeSpec project

_Assistant_: Ran azsdk_verify_setup, showed required steps and asked if I wanted to do local or pipeline generation

> Let's work through the required setup steps first

### Language

JS

### Service or Resource Provider

AI DocumentIntelligence

### Logs, Traces, or Screenshots
Expand details for full logs
<details>
<section>Full logs</section>
Image

Image

Image

Image

Image

Image

</details>

### azsdk_verify_setup
__Input__:
```json
{
  "langs": [
    "JavaScript"
  ],
  "packagePath": "/home/cradek/bug-bash/azure-rest-api-specs"
}
```
__Output__:
```json
{
    "results": [
        {
            "requirement": "tsp-client \u003E= 0.24.0",
            "instructions": [
                "cd eng/common/tsp-client",
                "npm ci"
            ],
            "requirementStatusDetails": "Command tsp-client --version failed to execute.\n\tException: An error occurred trying to start process \u0027tsp-client\u0027 with working directory \u0027/home/cradek/bug-bash/azure-rest-api-specs\u0027. No such file or directory"
        },
        {
            "requirement": "tsp \u003E= 1.0.0",
            "instructions": [
                "npm install -g @azure-tools/typespec-client-generator-cli@latest @typespec/compiler@latest"
            ],
            "requirementStatusDetails": "Command tsp --version failed to execute.\n\tException: An error occurred trying to start process \u0027tsp\u0027 with working directory \u0027/home/cradek/bug-bash/azure-rest-api-specs\u0027. No such file or directory"
        },
        {
            "requirement": "python",
            "instructions": [
                "Install Python from https://www.python.org/downloads/"
            ],
            "requirementStatusDetails": "Command python --version failed to execute.\n\tException: An error occurred trying to start process \u0027python\u0027 with working directory \u0027/home/cradek/bug-bash/azure-rest-api-specs\u0027. No such file or directory"
        },
        {
            "requirement": "pnpm",
            "instructions": [
                "npm install -g pnpm",
                "Run pnpm install in the repo root."
            ],
            "requirementStatusDetails": "Command pnpm --version failed to execute.\n\tException: An error occurred trying to start process \u0027pnpm\u0027 with working directory \u0027/home/cradek/bug-bash/azure-rest-api-specs\u0027. No such file or directory"
        }
    ],
    "operation_status": "Succeeded"
}
```

### Feedback
__NOTE__: This sequence was ran in the azure-rest-api-specs repo.

The agent should have first ran `npm ci` from the root of the repo to install the repo's dependencies. This would have included the TypeSpec compiler and tsp-client.

Instead, the agent tried to install `tsp-client` in `eng/common/tsp-client` (not what we want), and then tcgc/tsp globally:

Image

When I asked the agent if the repo had any steps to get started, it correctly ran `npm ci` from the repo root. However I was still asked one more time to install tcgc/tsp globally when that shouldn't be needed.

### LINKED ISSUE #13554: Refactor VerifySetup Requirements to be in code instead of json
move reqs to code for more flexibility

supports real needed use case of having different instructions per requirement based on the repo

JennyPng:
Closes https://github.com/Azure/azure-sdk-tools/issues/13554 
Closes https://github.com/Azure/azure-sdk-tools/issues/13342 by returning alternative instructions for the tsp/tsp-client reqs for the rest-api-spec repo

Refactors verify setup to retrieve requirements from code instead of JSON config for more flexibility in scenario2, and to enable cleaner conditional instruction returning based on OSplatform/language/repo

-  Creates Requirement abstract class, and subclasses for every setup requirement
- Creates RequirementContext class, which contains any context info that may inform the behavior of a requirement check
- decouples env setup from LanguageService by primarily encapsulating 'should check this requirement' information within a Requirement itself
- enables custom check logic beyond just running a command and checking exit code, by having an overridable `RunCheckAsync` function in `Requirement`
- reduces noise by returning only platform-specific instructions rather than relying on the LLM to determine the platform and format response accordingly
- updates unit test mocks accordingly 

- removes the old json and related models, etc.

## instructions for installing tsp in azure-rest-api-specs (on ubuntu)

image

## checking python reqs on ubuntu

(note, this is in the azure-sdk-tools repo, so the instructions for tsp are different than above)

image


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Requirement.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// Determines whether this requirement should be checked in the given context.
+    /// </summary>
+    /// <param name="ctx">The current environment context.</param>
+    /// <returns>True if the requirement should be checked, false otherwise.</returns>
+    public virtual bool ShouldCheck(RequirementContext ctx) => true;
+
+    /// <summary>
+    /// Optional minimum version (e.g., "22.16.0"). Null if no specific version required.
+    /// </summary>
+    public virtual string? MinVersion => null;
```
l0lawrence, line 38: are there any requirements that require an exact version, would we accidentally install too high of a version?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/CoreRequirements.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        public override string? Reason => "Python is required for all repos because it's used in a common Verify-Readme Powershell script.";
+
+        public override IReadOnlyList<string> GetInstructions(RequirementContext ctx)
+        {
+            if (ctx.IsLinux) 
+            {
+                return [
+                    "sudo apt install python3 python3-pip python3-venv",
+                    "sudo apt install python-is-python3"
```
l0lawrence, line 182: nice :)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Requirement.cs
JennyPng: nothing is currently defined with an exact version https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json

which is why i didn't add a MaxVersion, but it would be easy to add in the future if needed

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/AllRequirements.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Azure.Sdk.Tools.Cli.Models;
+
+namespace Azure.Sdk.Tools.Cli.Services.SetupRequirements;
```
l0lawrence, line 6: Are these Services - should they go under helper/SetupRequirements or some other folder maybe Models? If it is a service I believe it should get registered in the ServiceRegistration.cs

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
```diff
@@ -1,124 +0,0 @@
-{
-    "categories": {
-        "core": [
-            {
```
samvaity, line 4: I am curious, what is the reason to move away from the json and rather use model classes instead?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/SetupRequirements/GoRequirements.cs
```diff
@@ -0,0 +1,81 @@
...
+        public override string[] CheckCommand => ["go", "version"];
+
+        public override bool ShouldCheck(RequirementContext ctx) 
+            => ctx.Languages.Contains(SdkLanguage.Go);
+
+        public override IReadOnlyList<string> GetInstructions(RequirementContext ctx)
+        {
+            if (ctx.IsLinux) 
+            {
+                return ["sudo snap install go"];
```
richardpark-msft:
```suggestion
                return ["sudo snap install go --classic"];
```

--classic is required, since the Go app needs access to the actual filesystem, not just a sandbox.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/GoRequirements.cs
```diff
@@ -0,0 +1,81 @@
...
+            }
+           
+            return ["Download and install the latest Go version of the compiler from https://go.dev/dl/"];
+        }
+    }
+
+    public class GoImportsRequirement : Requirement
+    {
+        public override string Name => "goimports";
+        public override string[] CheckCommand => ["pwsh", "-Command", "Get-Command", "goimports"];
```
richardpark-msft: On Windows, and Linux, you can just run 'goimports' since it's provided as an exe. Or do we have to i nclude this boilerplate?

```suggestion
        public override string[] CheckCommand => ["goimports"];
```

JennyPng: I think I needed this because goimports doesn't support the `version` flag, and `help` doesn't exit with 0 which the tool treats as a failure 😢

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/AllRequirements.cs
JennyPng: I wasn't super sure, but copilot says since these classes are "stateful" they don't really fit under helpers or models 😄 i'll add it to serviceregistration

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Configuration/RequirementsV1.json
JennyPng: I was mainly motivated by a bug bash issue which basically said tsp-client needs a different instruction in azure-rest-api-specs repo vs. in other repos

I could make that work in JSON but it would be annoying to add more and more fields and have specialized parsing logic😄

I think the model classes makes it much more flexible for conditional instructions, and there will likely be more conditional behavior needed for scenario 2 when we have auto-install mode, to handle cross-platform differences

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/AllRequirements.cs
benbp: I wouldn't say there's a clear place. I'd lean towards helpers but given how things are set up today, maybe `Services/Languages/Requirements/*.cs` alongside `Services/Languages/Samples/`?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Requirement.cs
l0lawrence: sg to me :)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/GoRequirements.cs
benbp: Maybe instead of executing the check command in the verify setup tool it should just call a method on the base class like `RunCheckCommand()`. Then subsequently the child classes can either define a `CheckCommand[]` property OR a `CheckCommandFn()` override. That way if any special handling is necessary like parsing help text we don't have to create inline powershell commands. Ideally we reduce the dependency on specific shells being installed where possible, so we can eventually burn down the powershell requirement entirely for most scenarios.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/SetupRequirements/GoRequirements.cs
benbp: Linux distro differences are going to be a thorny area to manage setup instructions for. `snap` is an ubuntu distro specific install method that won't work elsewhere. Additionally snap may have issues on older WSL installs that don't have systemd enabled (like mine). I think for go it's probably better to just use the direct download method?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/CoreRequirements.cs
```diff
@@ -0,0 +1,157 @@
...
+
+    public class NodeRequirement : Requirement
+    {
+        public override string Name => "Node.js";
+        public override string? MinVersion => "22.16.0";
+        public override string[] CheckCommand => ["node", "--version"];
+
+        public override IReadOnlyList<string> GetInstructions(RequirementContext ctx)
+        {
+            if (ctx.IsLinux)
```
benbp, line 33: Don't want to scope creep this PR specifically, but for `IsLinux` we probably need to dig deeper into distro specific differences. For example, if the user is on debian or ubuntu (`lsb-release -a`) then we can return `apt install` instructions, but otherwise this is not going to work. I don't know how many users we would realistically have using other distros, but might be worth baking in slightly more intelligence to the `IsLinux` check.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/SetupRequirements/GoRequirements.cs
JennyPng: So I noticed an issue on WSL specifically where depending how you install things, the MCP server can't find the installation in agent mode, maybe due to some PATH or shell differences

but it seems to reliably find things installed with apt or snap, which is why we wanted to just suggest using that install method

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/GoRequirements.cs
richardpark-msft: > I think I needed this because goimports doesn't support the `version` flag, and `help` doesn't exit with 0 which the tool treats as a failure 😢

Oh, okay, this is fine. Good to add a comment on there, I can see someone else (not me, now!) optimizing that at a future point without realizing.

JennyPng: @benbp I see, I can look into that, I thought since Powershell is needed to start the mcp server in the first place, it would be fine to utilize it for checks 🤔

JennyPng:
oops i did something wrong with a merge 😶

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/SetupRequirements/GoRequirements.cs
richardpark-msft: @benbp, just based on what's easiest and "default", I'm guessing most people using WSL are going to be using Ubuntu, since it's one of the easiest ones to install from the store. And if you're using a non-Ubuntu WSL then you're going to be able to translate that snap command to your equivalent.

richardpark-msft: > So I noticed an issue on WSL specifically where depending how you install things, the MCP server can't find the installation in agent mode, maybe due to some PATH or shell differences
> 
> but it seems to reliably find things installed with apt or snap, which is why we wanted to just suggest using that install method

With all of that said before about using snap, we should NOT use 'apt'. The go version that apt installed tends to be rather old.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/GoRequirements.cs
benbp: It's not technically required for mcp it's just the installer script, and we also have the CLI invocation scenario.

Regardless I think it's better to not offload any processing logic to embedded scripts if we can avoid it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/SetupRequirements/GoRequirements.cs
benbp: Yeah I think the point about the install method putting it somewhere that's already in PATH is a big one, we want to avoid issues like that.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/AllRequirements.cs
JennyPng: Got it, I moved the requirements folder under `services/languages`

I don't _think_ I need to register in ServiceRegistration since these are static classes and not dependency injected?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/CoreRequirements.cs
JennyPng: I'll file an issue for this to address in the future

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/AllRequirements.cs
benbp: An alternative you could consider is injecting a non-static requirements class and adding a `RunCheck` method to the base requirements class. That way testing has better separation of components. In the verify setup tool tests you would only have to mock `requirements.RunCheck` instead of adding a mock for each process command. Then you could have a more dedicated `RequirementsTests.cs` that tests each implementation with the more specific process helper mock.

However then the requirements class becomes less of a data class so I can see an argument for the current structure. Though if we end up consuming requirements from more places than the verify setup tool then we're going to have different code paths for invoking requirements that could drift. In that future scenario `RunCheck` being in the base might make more sense.

I don't have a super strong opinion on this so I'd be fine if you want to leave it as is.

JennyPng: I see, I'll make a note of this in an issue 😄 I think I'll leave as is so I can start working on the auto-install logic for scenario 2

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/CoreRequirements.cs
```diff
@@ -0,0 +1,176 @@
...
+            }
+            return ["Download Node.js from https://nodejs.org"];            
+        }
+    }
+
+    public class TspClientRequirement : Requirement
+    {
+        public override string Name => "tsp-client";
+        public override string? MinVersion => "0.24.0";
+        public override string[] CheckCommand => ["tsp-client", "--version"];
```
raych1: The `checkCommand` needs to be different for spec repo vs sdk repos.

For sdk repos, it should be something like `npm exec --prefix 'eng/common/tsp-client' --no -- tsp-client --version`.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/RequirementContext.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    /// <summary>
+    /// True if running on Linux.
+    /// </summary>
+    public bool IsLinux => Platform == OSPlatform.Linux;
+
+    /// <summary>
+    /// The name of the repository (folder name).
+    /// </summary>
+    public string RepoName => Path.GetFileName(RepoRoot);
```
raych1: This is not reliable because the cloned folder name can be customized. We should use `gitHelper.GetRepoNameAsync` instead.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/RequirementContext.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// True if running on Linux.
+    /// </summary>
+    public bool IsLinux => Platform == OSPlatform.Linux;
+
+    /// <summary>
+    /// Determines if the current repository is the azure-rest-api-specs repository.
+    /// </summary>
+    public bool IsSpecsRepo()
+    {
+        return RepoName.Equals("azure-rest-api-specs", StringComparison.OrdinalIgnoreCase);
```
raych1: I just found that we have `typeSpecHelper.IsRepoPathForSpecRepoAsync` method which takes a path then check if it comes from spec repo. We may use it since it supports private spec repo (`azure-rest-api-specs-pr`) too. Alternatively, we should make `RepoName` checking work for private spec repo.

JennyPng: Ok, I'll just add to the check in RequirementContext as I don't think it's necessary to inject typespechelper just for this?

raych1 (approved):

PR #13669, opened 2026-01-21
Removed LibGit2Sharp dependency and use git CLI

### LINKED ISSUE #13630: Support git worktree working directories
On the secondary git worktree workspace, the `.git` is a file instead of a folder. 
The content of the `.git` file might be like "gitdir: /home/user/src/azure-sdk-for-rust/.git/worktrees/azure-sdk-for-rust2", which points to the real location of the repo metadata.

Current [code](@Azure/azure-sdk-tools/files/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs) only processes the case when `.git` is a folder. We will need to handle the other case when it's a file.

raych1:
Resolved #13630 

**Key Changes**:
- Removed the dependency on `LibGit2Sharp` library
- Updated the `gitHelper.cs` to use git command directly including:
  - get repo root
  - get branch name
  - get remote url
  - get merge base commit
- Extended test coverage


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
```diff
@@ -23,28 +23,85 @@ public class GitHelper(IGitHubService gitHubService, ILogger<GitHelper> logger)
         private readonly ILogger<GitHelper> logger = logger;
         private readonly IGitHubService gitHubService = gitHubService;
 
+        /// <summary>
```
benbp: If we're going to start shelling out for `git` then it should be implemented as a [CommandHelper](https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/CommandHelpers.cs). Also see the [Options](https://github.com/Azure/azure-sdk-tools/tree/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/Options) classes for how to add new command helper implementations.

This way we get standardized logging, output streaming, error handling and response types.

benbp: We could still hide the command helper calls behind this `GitHelper` though for most cases (avoids a big refactor and keeps the extra processing logic in this class).

raych1: Do you mean follow `CommandHelper` pattern to add `GitCommandHelper`, `GitOptions` then use `GitCommandHelper` in the `GitHelper`?

raych1: Updated the code.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
```diff
@@ -23,28 +23,85 @@ public class GitHelper(IGitHubService gitHubService, ILogger<GitHelper> logger)
...
-            using (var repo = new Repository(repoRoot))
+            try
             {
-                // Get the current branch
-                Branch currentBranch = repo.Head;
-                var targetBranch = repo.Branches[targetBranchName];
-
-                // Find the merge base (common ancestor) (git merge-base main HEAD)
-                var mergeBaseCommit = repo.ObjectDatabase.FindMergeBase(currentBranch.Tip, targetBranch.Tip);
+                var mergeBaseSha = RunGitCommand(repoRoot, $"merge-base HEAD {targetBranchName}");
```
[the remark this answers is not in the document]
raych1: It doesn't go thru the shell interpreter.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
raych1: @benbp could you take another look when getting a chance?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
```diff
@@ -18,33 +17,65 @@ public interface IGitHelper
...
+        private readonly IGitCommandHelper gitCommandHelper = gitCommandHelper;
+
+        /// <summary>
+        /// Runs a git command and returns the trimmed stdout output.
+        /// </summary>
+        /// <param name="workingDirectory">The directory to run the git command in</param>
+        /// <param name="arguments">The git command arguments (without 'git' prefix), space-separated</param>
+        /// <returns>The trimmed stdout output from the git command</returns>
+        /// <exception cref="InvalidOperationException">Thrown when git command fails</exception>
+        private string RunGitCommand(string workingDirectory, string arguments)
```
benbp: Can this helper method be removed entirely in favor of just `new GitOptions(); gitCommandHelper.Run()` in the other `GitHelper` methods? If the argument parsing is needed that could just be moved to the options constructor.

Though it will cause more files to be touched I don't think we should call `GetAwaiter`, everything should just change to be an `async` method.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/TestHelpers/GitTestHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// <summary>
+    /// Runs a git command in the specified directory.
+    /// </summary>
+    /// <param name="workingDirectory">The directory to run the command in</param>
+    /// <param name="arguments">The git command arguments (without 'git' prefix)</param>
+    /// <exception cref="InvalidOperationException">Thrown when the git command fails</exception>
+    private static void RunGit(string workingDirectory, string arguments)
+    {
+        var startInfo = new ProcessStartInfo
```
benbp: We should use the command helper for this

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/TestHelpers/GitTestHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Diagnostics;
+
+namespace Azure.Sdk.Tools.Cli.Tests.TestHelpers;
+
+/// <summary>
+/// Utility methods for git operations in tests.
+/// Uses git CLI directly instead of LibGit2Sharp for better compatibility
+/// with worktrees and to avoid native library dependencies.
+/// </summary>
+public static class GitTestHelper
```
benbp, line 14: Is the test helper necessary vs. just using the command helper directly in the tests?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
benbp: @raych1 added a couple code organization/simplification comments otherwise looks good.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
raych1: I hesitated to change this to async since it would touch quite a few files, and the git commands typically complete very quickly. That said, I’m happy to update it if you feel strongly about it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/TestHelpers/GitTestHelper.cs
raych1: It would be better to be kept as separate helper file given that multiple test files reference it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
raych1: Made the call async and invoke gitCommandHelper inline in other methods.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/TestHelpers/GitTestHelper.cs
raych1: Updated. Please take another look.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
benbp: Yeah originally I was thinking it would be nice to avoid the refactor, but I think we should just pull off the bandaid and do it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
```diff
@@ -1,125 +1,119 @@
...
-                    mergeBaseCommit?.Sha);
-                return mergeBaseCommit?.Sha ?? "";
+                    "Git merge base - Current branch: {currentBranch}, Merge base SHA: {mergeBaseCommitSha}",
+                    currentBranch,
+                    mergeBaseSha);
+                return mergeBaseSha;
+            }
+            catch (Exception ex)
+            {
+                logger.LogWarning(
```
benbp: I don't think we should swallow exceptions here

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
```diff
@@ -1,125 +1,119 @@
...
-            using var repo = new Repository(repoRoot);
-            var branchName = repo.Head.FriendlyName;
-            return branchName;
+            var repoRoot = await DiscoverRepoRootAsync(pathInRepo, ct);
+            var options = new GitOptions("rev-parse --abbrev-ref HEAD", repoRoot);
+            var result = await gitCommandHelper.Run(options, ct);
+            
+            if (result.ExitCode != 0)
+            {
+                throw new InvalidOperationException($"Git command failed: {result.Output.Trim()}");
```
benbp:
```suggestion
                throw new InvalidOperationException($"Command '{options.ShortName}' failed: {result.Output.Trim()}");
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
```diff
@@ -1,125 +1,119 @@
...
+            var repoRoot = await DiscoverRepoRootAsync(pathInRepo, ct);
+            var options = new GitOptions("remote get-url origin", repoRoot);
+            var result = await gitCommandHelper.Run(options, ct);
+            
+            if (result.ExitCode != 0)
             {
-                // Convert SSH URL to HTTPS URL
-                // git@github.com:Azure/azure-rest-api-specs.git -> https://github.com/Azure/azure-rest-api-specs.git
-                return gitUrl.Replace("git@github.com:", "https://github.com/");
+                throw new InvalidOperationException($"Git command failed: {result.Output.Trim()}");
```
benbp:
```suggestion
                throw new InvalidOperationException($"Command '{options.ShortName}' failed: {result.Output.Trim()}");
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -200,8 +200,8 @@ private async Task<DefaultCommandResponse> RunCheck(SetupRequirements.Requiremen
                 return reqsToCheck;
             }
 
-            languageSvs.Add(GetLanguageService(packagePath));
-        }
+            languageSvs.Add(await GetLanguageServiceAsync(packagePath, ct));
```
benbp, line 203: Potential perf impact here. Now that we're shelling out it may be worth revisiting caching by path at some point in the future if we see noticeable delays (unlikely).

benbp:
A couple notes otherwise LGTM

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
raych1: Agreed, it should be fine as the `rev-parse` command is extremely fast within milliseconds level. We'll revisit the cache if perf degradation is noticeable.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/GitHelper.cs
raych1: I thought to keep the previous contract returning empty string. But it's fine to throw the exception since the caller also handles the exception case. It's consistent with other methods by throwing exception now.

raych1:
> A couple notes otherwise LGTM

@benbp, thanks for the review! Could you please approve if the change looks good now?

benbp (approved):

PR #13682, opened 2026-01-22
Provide tool call references via server instructions

samvaity:
## How it works:
image
image

closes: #https://github.com/Azure/azure-sdk-tools/issues/13691


benbp:
This seems like a pretty simple way to solve some tool selection problems.

If we go with this approach, we should extend the CI checks we have to make sure that tool references in plain text correspond to tools that actually exist in the source code. Currently we do this for the `eng/common/instructions` files only:

https://github.com/Azure/azure-sdk-tools/blob/main/.github/workflows/instructions.yml

samvaity:
> This seems like a pretty simple way to solve some tool selection problems.
> 
> If we go with this approach, we should extend the CI checks we have to make sure that tool references in plain text correspond to tools that actually exist in the source code. Currently we do this for the `eng/common/instructions` files only:
> 
> https://github.com/Azure/azure-sdk-tools/blob/main/.github/workflows/instructions.yml

Yeah, will do if needed. Currently, updated the server instructions to use generic descriptions rather than specific tool names, so CI validation isn't needed yet. If we add specific tool references later, we'll extend the workflow

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Resources/azsdk-rules.txt
```diff
@@ -0,0 +1,8 @@
+Azure SDK Tools MCP server usage rules:
+- When working with Azure SDK repositories, TypeSpec API specifications, SDK code generation, SDK releases, or SDK workflows, use azsdk tools.
+- When generating SDKs from TypeSpec, ask user preference for local vs pipeline-based generation before proceeding.
```
raych1, line 3: Could you share some test result? How does it work along with instructions in the spec repo and sdk repo?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Resources/azsdk-rules.txt
```diff
@@ -0,0 +1,8 @@
...
+- When analyzing Azure Pipeline failures or CI/CD issues, use azsdk pipeline analysis tools.
+- When working with TypeSpec projects, use azsdk TypeSpec tools.
+- When managing pull requests in Azure SDK repositories, use azsdk GitHub tools.
+- When working with SDK release workflows or release plans, use azsdk release tools.
```
raych1, line 7: I am curious where does the tool category come from. Such as: 'azsdk release tools', 'azsdk pipeline analysis tools'. Will it be used to match with the description of the tools?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Program.cs
```diff
@@ -85,11 +87,38 @@ public static WebApplicationBuilder CreateAppBuilder(string[] args, string outpu
...
+            });
             builder.Services
                 .AddMcpServer()
                 .WithStdioServerTransport();
         }
 
         return builder;
     }
+
+    private static string LoadServerInstructions()
```
benbp, line 103: I wonder if it's better to define the instructions in a here string in a static class given that we're adding almost the same lines of code implementing this method. I don't have a strong opinion either way.

PR #13684, opened 2026-01-22
[azsdk-cli] Sanitize paths in telemetry

benbp:
We don't want paths containing usernames to show up in our telemetry data, for example @chrisradek:

image

These can show up anywhere: process output, exceptions, custom tags we set in tools, etc.

Right now this takes a brute force approach and analyzes every activity tag for a path-like pattern and redacts it. We keep the child segments of paths after well-known directory names like repository so that we can still do package usage correlation (e.g. `/home/ben/azure-sdk-for-python/sdk/foo` becomes `[PATH REDACTED]/azure-sdk-for-python/sdk/foo`. Additionally other components like the git helper can add to the dir name allowlist, for example when we call `DiscoverRepoRoot` that will update that root dir so we don't exclude workspaces where the name doesn't match the github remote.

I anticipate this sanitization infra will also be extended to process additional types of patterns in the future.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Helpers/TelemetryPathSanitizerTests.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Helpers/TelemetryPathSanitizerTests.cs
@@ -5,16 +5,25 @@ namespace Azure.Sdk.Tools.Cli.Tests.Helpers;
 
 public class TelemetryPathSanitizerTests
 {
-    [TestCase("specification/service", "[PATH REDACTED]/specification/service")]
-    [TestCase("azure-rest-api-specs/specification/service", "[PATH REDACTED]/azure-rest-api-specs/specification/service")]
+    [TestCase("azure-rest-api-specs/specification/service", "azure-rest-api-specs/specification/service")]
     [TestCase("/azure-rest-api-specs/specification/service", "[PATH REDACTED]/azure-rest-api-specs/specification/service")]
     [TestCase("/Users/ben/specification/service", "[PATH REDACTED]/specification/service")]
-    [TestCase(@"C:\Users\ben\sdk\azure-sdk-for-net\sdk\storage", @"[PATH REDACTED]\azure-sdk-for-net\sdk\storage")]
-    [TestCase("C:/Users/ben/sdk/azure-sdk-for-net/sdk/storage", "[PATH REDACTED]/azure-sdk-for-net/sdk/storage")]
-    [TestCase("/home/ben/sdk/azure-sdk-for-net/sdk/storage", "[PATH REDACTED]/azure-sdk-for-net/sdk/storage")]
-    [TestCase("/home/ben/sdk/azure-sdk-for-go/sdk/resourcemanager/advisor/armadvisor", "[PATH REDACTED]/azure-sdk-for-go/sdk/resourcemanager/advisor/armadvisor")]
+    [TestCase(@"C:\Users\ben\sdk\azure-sdk-for-net\sdk\storage", @"[PATH REDACTED]\sdk\azure-sdk-for-net\sdk\storage")]
+    [TestCase("C:/Users/ben/repos/azure-sdk-for-net/sdk/storage", "[PATH REDACTED]/azure-sdk-for-net/sdk/storage")]
+    [TestCase("/home/ben/repos/azure-sdk-for-net/sdk/storage", "[PATH REDACTED]/azure-sdk-for-net/sdk/storage")]
     [TestCase("~/specification/service", "[PATH REDACTED]/specification/service")]
     public void Sanitize_PreservesAllowlistedSegments(string input, string expected)
+    {
+        var sanitized = TelemetryPathSanitizer.Sanitize(input);
+
+        Assert.That(sanitized, Is.EqualTo(expected));
+    }
+
+    [TestCase("sdk/resourcemanager/advisor/armadvisor", "sdk/resourcemanager/advisor/armadvisor")]
+    [TestCase("github.com/Azure/azure-sdk-for-go/sdk/azcore", "github.com/Azure/azure-sdk-for-go/sdk/azcore")]
+    [TestCase("@azure/storage-blob", "@azure/storage-blob")]
+    [TestCase("@azure-rest/core-client", "@azure-rest/core-client")]
+    public void Sanitize_PreservesAllowlistedSegmentsForPackageNamingConventions(string input, string expected)
     {
         var sanitized = TelemetryPathSanitizer.Sanitize(input);
 
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Helpers/TelemetryPathSanitizerTests.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Helpers/TelemetryPathSanitizerTests.cs
@@ -9,6 +9,8 @@ public class TelemetryPathSanitizerTests
     [TestCase("azure-rest-api-specs/specification/service", "[PATH REDACTED]/azure-rest-api-specs/specification/service")]
     [TestCase("/Users/ben/specification/service", "[PATH REDACTED]/specification/service")]
     [TestCase(@"C:\Users\ben\azure-sdk-for-net\sdk\storage", @"[PATH REDACTED]\azure-sdk-for-net\sdk\storage")]
+    [TestCase("C:/Users/ben/azure-sdk-for-net/sdk/storage", "[PATH REDACTED]/azure-sdk-for-net/sdk/storage")]
+    [TestCase("~/specification/service", "[PATH REDACTED]/specification/service")]
     public void Sanitize_PreservesAllowlistedSegments(string input, string expected)
     {
         var sanitized = TelemetryPathSanitizer.Sanitize(input);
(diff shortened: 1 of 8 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/MCPToolBase.cs
```diff
@@ -67,7 +67,7 @@ public async Task<int> InstrumentedCommandHandler(Command command, ParseResult p
         }
         catch (Exception ex)
         {
-            activity?.AddException(ex);
+            TelemetryExceptionHelper.AddSanitizedException(activity, ex);
```
raych1, line 70: Just to confirm, this sanitizer only affects values sent to telemetry and does not alter logs in the console, right?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Telemetry/TelemetryPathSanitizer.cs
```diff
@@ -0,0 +1,355 @@
...
+using System.Collections.Concurrent;
+using System.Text;
+
+namespace Azure.Sdk.Tools.Cli.Telemetry;
+
+// Paths in telemetry can contain usernames and other identifying information.
+// We want to redact this info but keep the child segments of paths that are
+// relevant for correlating per-package usage of this tool, based on known
+// directory prefixes or names that correspond to our repositories.
+public static class TelemetryPathSanitizer
```
raych1, line 13: Other than paths, do we have a need to redact other values, like mac address or device id? I see these are currently collected as part of telemetry.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/MCPToolBase.cs
benbp: Yep, just the telemetry pipeline, doesn't touch logs

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Telemetry/TelemetryPathSanitizer.cs
benbp: Those values are already anonymized, for example we only upload a max address hash not the actual value

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Helpers/TelemetryPathSanitizerTests.cs
```diff
@@ -0,0 +1,143 @@
...
+public class TelemetryPathSanitizerTests
+{
+    [TestCase("specification/service", "[PATH REDACTED]/specification/service")]
+    [TestCase("azure-rest-api-specs/specification/service", "[PATH REDACTED]/azure-rest-api-specs/specification/service")]
+    [TestCase("/azure-rest-api-specs/specification/service", "[PATH REDACTED]/azure-rest-api-specs/specification/service")]
+    [TestCase("/Users/ben/specification/service", "[PATH REDACTED]/specification/service")]
+    [TestCase(@"C:\Users\ben\sdk\azure-sdk-for-net\sdk\storage", @"[PATH REDACTED]\azure-sdk-for-net\sdk\storage")]
+    [TestCase("C:/Users/ben/sdk/azure-sdk-for-net/sdk/storage", "[PATH REDACTED]/azure-sdk-for-net/sdk/storage")]
+    [TestCase("/home/ben/sdk/azure-sdk-for-net/sdk/storage", "[PATH REDACTED]/azure-sdk-for-net/sdk/storage")]
+    [TestCase("~/specification/service", "[PATH REDACTED]/specification/service")]
```
praveenkuttappan, line 14: Can you please add a test case for Go package name? Go package name is relative directory name begins with "sdk/<>"

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Telemetry/TelemetryPathSanitizer.cs
```diff
@@ -0,0 +1,352 @@
...
+{
+    public const string Redacted = "[PATH REDACTED]";
+
+    private static readonly string[] AllowlistedSegments =
+    [
+        "azure-rest-api-specs",
+        "specification",
```
praveenkuttappan, line 20: is it worth to add `sdk`?

benbp: I was thinking that `sdk` for service directories is likely to be a child path of a repo path, so it felt unnecessary (even `specification` could probably be removed by this logic). And I don't think it's important that we catch a parent path like `/home/ben/sdk/azure-sdk-for-net` (`[PATH REDACTED]/sdk/azure-sdk-for-net`), because I'm worried we could see stuff like `C:/sdk/benbp/azure-sdk-for-net`.

tools/azsdk-cli/Azure.Sdk.Tools.Cli.Tests/Helpers/TelemetryPathSanitizerTests.cs
```diff
@@ -0,0 +1,143 @@
+using Azure.Sdk.Tools.Cli.Telemetry;
+using NUnit.Framework;
+
+namespace Azure.Sdk.Tools.Cli.Tests.Helpers;
+
+public class TelemetryPathSanitizerTests
+{
+    [TestCase("specification/service", "[PATH REDACTED]/specification/service")]
```
richardpark-msft: I'm curious, we seem to have a lot of code to redact basically just a few cases? Is this not just a regexp that handles these cases?

* `C:\Users\<user>`
* `/home/<user>`**

benbp: A regex would be a lot simpler but I'm worried the allowed string list might grow and what should be a fast processing step will end up adding noticeable delays. We only do all this processing after we finish the command and before we return to the user, so anything over 200ms is noticeable (combined with the time it takes to upload). For example, @praveenkuttappan already pointed out 4 more allowed strings we need to add so we don't redact go and js/ts package names, and currently people may add very large output values from sub-process runs to their response objects.

It was easy enough to generate the parsing code and tests and it's not critical path so I'm not super worried about the code maintenance aspect of it enough compared against (potential) perf issues. The regex approach also has the potential to become pretty thorny and hard to read (processing relative paths, processing paths that don't end in whitespace, etc.).

richardpark-msft: > A regex would be a lot simpler but I'm worried the allowed string list might grow and what should be a fast processing step will end up adding noticeable delays. 

If that was our concern, I'd definitely want some measurements. Regexps seem like a well optimized area of text processing libraries. Also, the text that we're sanitizing over isn't really that big. It might be we're optimizing a bit early.

This is your call, as you're the maintainer of this telemetry piece. :)

benbp: There are a few places in the code where we copy process output into the response, and the process output could have the potential to be very large. We're trying to think of ways to solve that problem (don't upload those types of responses to telemetry) which would simplify potential optimization needs.

I agree this reeks of premature optimization. I'll mock up a regex alternative and profile them both.

richardpark-msft: BTW, given that we're trying to clean up our telemetry, I'm okay if you want to check this in and do the regexp update (if needed) in another PR.

benbp: Evaluating a simple exception, a 1mb string with 20 paths and a 10mb string with 20 paths, but using the tokenized sanitization logic (vs. regex replace grouping). Will test with groupings next.

```
      TelemetryPathSanitizer mode=Parser elapsedMs=0.628 length=23652
      TelemetryPathSanitizer mode=Parser elapsedMs=38.873 length=1048600
      TelemetryPathSanitizer mode=Parser elapsedMs=295.333 length=10485805
      TelemetryPathSanitizer mode=Regex elapsedMs=0.869 length=23652
      TelemetryPathSanitizer mode=Regex elapsedMs=92.388 length=1048600
      TelemetryPathSanitizer mode=Regex elapsedMs=999.731 length=10485805
```

benbp: Using regex matching more completely ends up in a nightmare regex (and 50ms vs 300ms in the 1mb case). Without matching the regex is simple but we only reduce about 30 lines of code, as the actual token replacement logic remains. So I'm not sure the replacement solution buys us any worthwhile reduction in complexity.

benbp: The cases aren't as simple as a home directory pattern, as I've seen people have PII later down in the path (see the image in the PR description, for example) and similarly the path may or may not contain an `azure-sdk*` patterned directory name (e.g. worktrees), so we need to be able to query against both the hardcoded allowlist and the dynamic repo root cache populated by the git helper.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Telemetry/TelemetryPathSanitizer.cs
```diff
@@ -0,0 +1,356 @@
...
+        // 'sdk' and 'github.com' must be allowed so we don't redact golang package names
+        // which look like 'sdk/resourcemanager/advisor/armadvisor'
+        // or 'github.com/Azure/azure-sdk-for-go/sdk/azcore'
+        "sdk",
+        "github.com",
+        // '@azure*' must be allowed so we don't fully redact js/ts package names
```
raych1, line 26: Should we include `azure`? Do you consider moving this allowlist to a config file?

benbp: I switched the `azure-sdk-*` prefix to `azure*`. I think if the list grows a bit more we can move it to a config but ideally it stays small.

raych1 (approved):

PR #13686, opened 2026-01-22
[Chatbot][Frontend] Add redirect-channel text and reorganize config code

JialinHuang803:
This PR includes:

**Bot Static Messages Updates:**

* Simplify contact card text.
* Add conditional footer with channel redirect.

**Azure Blob Storage Integration:**

* Added a new `BlobClientManager` class to `src/config/blobClient.ts` that manages a singleton Azure Blob Storage client, providing methods to download blobs and retrieve last modified times.
* Refactored `ChannelConfigManager` in `src/config/channel.ts` to use `BlobClientManager` for loading and watching the `channel.yaml` configuration from Blob Storage, removing legacy Blob client initialization logic. 
* Added a new `TenantConfigManager` class to `src/config/tenant.ts` that loads the `tenant.yaml` configuration from Blob Storage for tenant-to-channel mapping.

**Configuration Initialization and Local Development:**

* Introduced a `ConfigFacade` to centralize configuration initialization, ensuring all config managers are initialized before use in `src/app/app.ts`.
* Enhanced local development experience by supporting local settings and a local RAG tenant in `.env.testtool` and channel config logic.

**Environment and Infrastructure Updates:**

* Updated all environment files (`.env.dev`, `.env.preprod`, `.env.prod`, `.env.testtool`) to add the `AZURE_BLOB_STORAGE_URL` variable.
* Modified infrastructure templates (`azure.bicep`, `azure.parameters.json`) to accept and propagate the Blob Storage URL as a parameter/environment variable.

**Build & Tooling:**

* Fixed the test tool environment variable file path in the `package.json` scripts.

### DIFF
--- a/tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/channel.ts
+++ b/tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/channel.ts
@@ -193,7 +193,6 @@ class ChannelConfigManager {
         error: error.message,
         blob: config.channelConfigBlobName,
       });
-      throw new Error(`Failed to load channel configuration: ${error.message}`);
     }
   }
 
--- a/tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
+++ b/tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
@@ -160,7 +160,6 @@ class TenantConfigManager {
         error: error.message,
         blob: config.tenantConfigBlobName,
       });
-      throw new Error(`Failed to load tenant configuration: ${error.message}`);
     }
   }
 


tools/sdk-ai-bots/azure-sdk-qa-bot/src/cards/components/contact.ts
```diff
@@ -20,7 +20,12 @@ export function createContactCard() {
     body: [
       {
         type: 'TextBlock',
-        text: '🤖The responses are generated by AI and may contain outdated information. Please verify any important details independently. 🔁For follow-up questions, @mentioning the bot is recommended. 📝Please note that the bot is unable to reply to edited messages.',
+        text: '🤖 AI-generated response. Please verify before taking action.',
+        wrap: true,
+      },
+      {
+        type: 'TextBlock',
+        text: '📝 Please note that the bot is unable to reply to edited messages.',
```
lirenhe: Can we make it more clearly, e.g.  

The chatbot only responds to the original message. If a message is edited, it will not reply again.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/cards/components/contact.ts
```diff
@@ -20,7 +20,12 @@ export function createContactCard() {
     body: [
       {
         type: 'TextBlock',
-        text: '🤖The responses are generated by AI and may contain outdated information. Please verify any important details independently. 🔁For follow-up questions, @mentioning the bot is recommended. 📝Please note that the bot is unable to reply to edited messages.',
+        text: '🤖 AI-generated response. Please verify before taking action.',
```
lirenhe: Let's make the message more clearly:  'This response is generated by AI. Please review it before taking action.'

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/blobClient.ts
```diff
@@ -0,0 +1,121 @@
+import { BlobServiceClient, ContainerClient } from '@azure/storage-blob';
+import { logger } from '../logging/logger.js';
+import config from './config.js';
+import { getAzureCredential } from '../common/shared.js';
+
+// Azure Blob Storage configuration
+const CONTAINER_NAME = 'bot-configs';
```
lirenhe: Can we put the config into env file?

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/channel.ts
```diff
@@ -21,7 +20,6 @@ export interface ChannelConfig {
 }
 
 // Azure Blob Storage configuration
-const CONTAINER_NAME = 'bot-configs';
 const BLOB_NAME = 'channel.yaml';
 const FALLBACK_RAG_ENDPOINT = 'https://azuresdkbot-dqh7g6btekbfa3hh.eastasia-01.azurewebsites.net';
```
lirenhe: It is in existing code, but better not hard code those values in src.

JialinHuang803: Moved to env.

tools/sdk-ai-bots/azure-sdk-qa-bot/env/.env.preprod
```diff
@@ -14,10 +14,15 @@ AZURE_RESOURCE_GROUP_NAME=azure-sdk-qa-bot-test
...
 AZURE_TABLE_NAME_FOR_CONVERSATION='TeamsChannelConversationsDev'
 
+## Azure Blob Storage
+AZURE_BLOB_STORAGE_URL='https://azuresdkqabotstorage.blob.core.windows.net/'
+BLOB_CONTAINER_NAME='bot-configs'
+CHANNEL_CONFIG_BLOB_NAME='channel.yaml'
+TENANT_CONFIG_BLOB_NAME='tenant.yaml'
 
 ## Email Alerts
 ALERT_EMAIL_ADDRESSES='chunyu@microsoft.com,jiaqzhang@microsoft.com'
```
lirenhe: Could you add yourself into the alert email address?

tools/sdk-ai-bots/azure-sdk-qa-bot/env/.env.dev
```diff
@@ -29,9 +34,12 @@ STORAGE_RESOURCE_GROUP_NAME='azure-sdk-qa-bot-dev'
 ## RAG Service scope
 RAG_SERVICE_SCOPE='api://azure-sdk-qa-bot-dev/.default'
 
+## RAG Service fallback
+FALLBACK_RAG_ENDPOINT='https://azuresdkbot-dqh7g6btekbfa3hh.eastasia-01.azurewebsites.net'
```
chunyu3: Does a fallback RAG service exist?
Even if we configure a fallback endpoint, we should not fall back to the production environment. Doing so would forward requests from the development environment to production, which could introduce risk and potentially impact the production environment. We should avoid putting the production environment at risk.

tools/sdk-ai-bots/azure-sdk-qa-bot/env/.env.dev
```diff
@@ -29,9 +34,12 @@ STORAGE_RESOURCE_GROUP_NAME='azure-sdk-qa-bot-dev'
...
+FALLBACK_RAG_TENANT='azure_sdk_qa_bot'
 
 ## Docker
-DOCKER_IMAGE_TAG="dev-0.0.19"
+DOCKER_IMAGE_TAG="dev-0.0.20"
```
chunyu3, line 35: Better to update the Image_tag in other environments to refer to exact tag deployed

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/blobClient.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+    }
+
+    return undefined;
+  }
+
+  /**
+   * Get blob last modified time
+   */
+  public async getBlobLastModified(blobName: string): Promise<Date | undefined> {
```
chunyu3: For more clearness
```suggestion
  public async getBlobLastModifiedTime(blobName: string): Promise<Date | undefined> {
```

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/channel.ts
```diff
@@ -94,25 +50,35 @@ class ChannelConfigManager {
...
     // Find matching channel
-    const channel = config.channels.find((ch) => ch.id === channelId);
+    const channel = channelConfig.channels.find((ch) => ch.id === channelId);
     if (channel) {
       return {
         name: channel.name,
         id: channel.id,
-        tenant: channel.tenant || config.default.tenant,
-        endpoint: channel.endpoint || config.default.endpoint,
+        tenant: channel.tenant,
```
chunyu3: What if `channel.tenant` is undefined?

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/channel.ts
```diff
@@ -209,34 +164,19 @@ class ChannelConfigManager {
...
       if (!fileContent) {
-        logger.warn(`Failed to download channel config blob after ${this.maxRetry} attempts, still use old config.`);
-        return;
+        // If we already have a config, keep using it
+        if (this.config) {
+          logger.warn('Failed to download channel config blob, still use old config.');
+          return;
+        }
+        // First initialization - no config to fall back to
+        throw new Error('Failed to download channel config blob and no existing config available');
```
chunyu3: `Throw exception` will crash all the front-end qa bot application. And we may not aware that the application is done and restart it.
The more suitable is to Logger a error and return the teams post message to indicate that the bot meet some problem and cannot answer.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/channel.ts
```diff
@@ -247,12 +187,19 @@ class ChannelConfigManager {
       this.config = parsedConfig;
 
       // Get blob properties for last modified time
-      const properties = await blobClient.getProperties();
-      this.lastModified = properties.lastModified || new Date();
+      try {
+        const lastModified = await this.blobClientManager.getBlobLastModified(config.channelConfigBlobName);
+        this.lastModified = lastModified || new Date();
```
chunyu3: better to use ?? to exclude only undefined/null
```suggestion
        this.lastModified = lastModified ?? new Date();
```

tools/sdk-ai-bots/azure-sdk-qa-bot/src/turn/ThinkingHandler.ts
```diff
@@ -75,17 +85,49 @@ export class ThinkingHandler {
...
+    if (routeTenant) {
+      try {
+        // Get channel info (name and url) from tenant ID
+        const tenant = await this.tenantConfigManager.getTenant(routeTenant);
+        if (!tenant) {
+          logger.warn(`Tenant not found for route_tenant: ${routeTenant}`, { meta: this.meta });
+        } else {
+          const displayName = tenant.channel_name || routeTenant;
+          const channelLink = tenant.channel_link ? `[${displayName}](${tenant.channel_link})` : `${displayName}`;
+          const redirectText = `💬 Not resolved? Try posting in the 👉 ${channelLink} channel where our domain experts can provide a deeper dive.`;
```
chunyu3: Not clear whether chat bot will automatically redirect the post to another channel or ask user to re-post himself

tools/sdk-ai-bots/azure-sdk-qa-bot/env/.env.dev
JialinHuang803: The `DOCKER_IMAGE_TAG` in other .env files were also updated to the latest deployed tag. For preprod and prod, it's the value that I will deploy after checkin.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/channel.ts
JialinHuang803: The existing code ensures that `channel.tenant` and `channel.endpoint` are not empty because: 
1. the properties of the interface ChannelItem are all required.
```
export interface ChannelItem {
  name: string;
  id: string;
  tenant: string;
  endpoint: string;
}
```
2. in the `ensureConfig()` method, if any value is undefined, an error will be thrown:
```
for (const channel of this.config.channels) {
  if (!channel.id || !channel.name || !channel.tenant || !channel.endpoint) {
    throw new Error(`Channel configuration is incomplete: ${JSON.stringify(channel)}`);
  }
}
```
They may be redundant, but I don't plan to update it in this PR.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/channel.ts
JialinHuang803: The error occurs during initialization, not when users send messages in the channel. If the bot can't load channel config on first start, it's fundamentally broken. And it may make more sense to keep tracking and ensure that there is no error when deploying the bot.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/turn/ThinkingHandler.ts
JialinHuang803: The users should re-post themselves~

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/channel.ts
JialinHuang803: As discussed offline, we will throw the error in the `initialize()` method.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/channel.ts
JialinHuang803: I've removed the validation code and added back the `config.default.xxx` logic.

tools/sdk-ai-bots/azure-sdk-qa-bot/env/.env.dev
JialinHuang803: As discussed offline, we will remove the fallback constant values.

tools/sdk-ai-bots/azure-sdk-qa-bot/env/.env.dev
```diff
@@ -14,13 +14,18 @@ AZURE_RESOURCE_GROUP_NAME=azure-sdk-qa-bot-dev
 RESOURCE_SUFFIX=azsdkqabotdev
 
 
-## Azure Table
+## Azure Table Storage
 AZURE_STORAGE_URL='https://azuresdkqabotdevstorage.table.core.windows.net/'
 AZURE_TABLE_NAME_FOR_CONVERSATION='TeamsChannelConversationsDev'
 
+## Azure Blob Storage
+AZURE_BLOB_STORAGE_URL='https://azuresdkqabotdevstorage.blob.core.windows.net/'
```
chunyu3: If we define a separate environment variable for bot_storage_url, the blob and table storage may belong to different storage accounts.
Accessing these storage resources requires the bot to be granted appropriate permissions, which are configured during deployment by executing the Bicep templates.
https://github.com/Azure/azure-sdk-tools/blob/485cb0b32491ee9e1e0f66d44d16b9ce90209ecd/tools/sdk-ai-bots/azure-sdk-qa-bot/infra/storagePermissions.bicep#L14
When a different storage account is used for the blob, the required permissions may not be granted automatically. We may need to require permissions during deployment.
Since, in our environment, both the table and blob resources reside in the same storage account,  we just need to define the constraint in configure file to avoid two-permission-granted-flow. It is preferable to define a single environment variable, STORAGE_ACCOUNT_NAME, and derive both azure_table_storage_url and azure_blob_storage_url from that storage account.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
```diff
@@ -0,0 +1,232 @@
...
+    if (!this.config) {
+      throw new Error('Failed to initialize tenant configuration: config could not be loaded from blob storage');
+    }
+    this.startWatching();
+  }
+
+  /**
+   * Get the current tenant configuration
+   */
+  public async getConfig(): Promise<TenantConfig> {
```
chunyu3: This function may not be defined as async func.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
```diff
@@ -0,0 +1,232 @@
...
+    if (!this.config) {
+      throw new Error('Tenant configuration is not loaded');
```
chunyu3: Still not good to throw error

tools/sdk-ai-bots/azure-sdk-qa-bot/env/.env.dev
JialinHuang803: Updated: Remove the Table URL and Blob URL environment variables and get the URLs from the Storage account name.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
```diff
@@ -0,0 +1,232 @@
...
+
+      // Get blob properties for last modified time
+      try {
+        this.lastModified = (await this.blobClientManager.getBlobLastModifiedTime(config.tenantConfigBlobName)) ?? new Date();
+      } catch (metadataError) {
+        logger.warn('Failed to get blob last modified time, using current time', {
+          blob: config.tenantConfigBlobName,
+          error: metadataError.message,
+        });
+        this.lastModified = new Date();
```
chunyu3: Not good to set a faked modifiedTime.
Consider a rare case:
After you read the blob content, the blob is changed. And then you fail to retrieve the modified time, and set the modified time the current time which is later than the actually modified time. In the next loop, the system always determines that the in‑memory config is newer than the one in storage and therefore decides not to reload it. Consequently, the configuration stored in blob storage is never loaded again.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
```diff
@@ -0,0 +1,232 @@
...
+      });
+      throw new Error(`Failed to load tenant configuration: ${error.message}`);
+    }
+  }
+
+  /**
+   * Promise-based delay function using async/await
+   */
+  private async delay(ms: number): Promise<void> {
+    await setTimeout(ms);
```
chunyu3: Just curious,
why do we use async setTimeout and as a result make the delay an async func?

tools/sdk-ai-bots/azure-sdk-qa-bot/src/turn/ThinkingHandler.ts
chunyu3: I mean that the `redirectText` does not clear to info user to re-post themselves. Better to refine the sentence.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
JialinHuang803: This code is identical to channel.ts, which has been in place since the beginning. Here is the explanation by Copilot:
The Promise-based `setTimeout` clean async/await flow in the watch loop:
```
private async watchLoop(): Promise<void> {
  while (this.isWatching) {
    await this.checkAndReload();
    await setTimeout(this.watchInterval);  // Clean linear flow
  }
}
```
Without it, you'd need nested callbacks or promisify it yourself:
```
// Ugly callback version
private watchLoop(): void {
  this.checkAndReload().then(() => {
    setTimeout(() => {
      if (this.isWatching) {
        this.watchLoop();  // Recursive call
      }
    }, this.watchInterval);
  });
}
```
The `delay()` wrapper method is unnecessary and I can simplify it by just calling `await setTimeout(this.watchInterval);` directly.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
JialinHuang803: Goot point. Will remove the `new Date()` for both channel and tenant configs.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/turn/ThinkingHandler.ts
JialinHuang803: Perhaps the message 'Try posting ...' already informs users to post in the target channel. Do we have any recommendations for improving this?

JialinHuang803: I can change the 'Try posting ...' to 'Please re-post ...' which may make it clearer.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
JialinHuang803: Removed the 'async' for all relevant methods.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
JialinHuang803: Updated.

tools/sdk-ai-bots/azure-sdk-qa-bot/src/config/tenant.ts
```diff
@@ -0,0 +1,221 @@
...
+  private async checkAndReload(): Promise<void> {
+    if (this.loadPromise) {
+      await this.loadPromise;
+      return;
+    }
+
+    try {
+      const currentModified = await this.blobClientManager.getBlobLastModifiedTime(config.tenantConfigBlobName);
+
+      if (!this.lastModified || (currentModified && currentModified > this.lastModified)) {
```
chunyu3, line 83: maybe change condition as following to include currentModified not retrieved case
```suggestion
      if (!this.lastModified || !currentModified || currentModified > this.lastModified) {
```

JialinHuang803: I think it's fine that we just simply skip this check and try again next cycle (in 5 seconds) if `currentModified` is null since we seldom change the config.

Issue #13599, opened 2026-01-15
Update SDK details MCP failure due to invalid package name.


praveenkuttappan:
Update SDK details MCP tool validates the package name and fails the request if package name is not as per the expected format.

WE should check if we can update the copilot instructions so it passes the correct package name in the first attempt.

Error message: `Unsupported package name(s) detected: Go -\\u003E armedgemarketplace`
Tool name: `azsdk_update_sdk_details_in_release_plan`

tools args: ```{"releasePlanWorkItemId":30207,"sdkDetails":"[  {    "Language": "JavaScript",    "PackageName": "@azure/arm-edgemarketplace"  },  {    "Language": "Java",    "PackageName": "azure-resourcemanager-edgemarketplace"  },  {    "Language": "Python",    "PackageName": "azure-mgmt-edgemarketplace"  },  {    "Language": "Go",    "PackageName": "armedgemarketplace"  }]"}```

PR #13693, opened 2026-01-23
Add instruction to normalize package name

smw-ms:
Resolves: https://github.com/Azure/azure-sdk-tools/issues/13599

Example Response 1:
example1

Example Response 2:
example2

### DIFF
--- a/eng/common/instructions/azsdk-tools/sdk-details-in-release-plan.instructions.md
+++ b/eng/common/instructions/azsdk-tools/sdk-details-in-release-plan.instructions.md
@@ -36,9 +36,16 @@ description: 'Identify languages configured in the TypeSpec project and add it t
 **Success Criteria**: Release plan exists or user informed to create one.
 
 # Step 3: Update Release Plan with SDK Information
-> **(MANDATORY - DO NOT SKIP) ALWAYS validate all package names against the format rules before calling any update tool, even if the user provides SDK details directly. Auto-correct and inform the user of invalid package names.**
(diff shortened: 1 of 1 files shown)


eng/common/instructions/azsdk-tools/sdk-details-in-release-plan.instructions.md
```diff
@@ -8,7 +8,7 @@ description: 'Identify languages configured in the TypeSpec project and add it t
    - For Java and Python, use `emitter-output-dir` for package name if it exists. Otherwise use `package-dir` to get the package name as fallback approach.
    - For .NET, use namespace property to get package name.
    - For JavaScript, use `packagedetails:name` property to get package name.
-   - For Go, use module name and remove `github.com/Azure/azure-sdk-for-go/` to get package name
+   - For Go, use module name and remove `github.com/Azure/azure-sdk-for-go/` to get package name. If the Go package name doesn't start with `sdk/`, automatically normalize it to the format `sdk/resourcemanager/<service>/<packagename>`.
```
praveenkuttappan: This will be confusing for data plane because `resourcemanager` is present only for management plane. Let's just say that Go package name starts with `sdk/`

smw-ms: i updated the code. but the only thing about that is if you stick sdk/ in front of anything it will let the tool go through without any problems

smw-ms: we could check that the go package name is actually a valid path in the repo as well inside the tool?

PR #13698, opened 2026-01-23
Fix empty owners issue

weshaggard:
We have valid reasons to empty the owners list for particular paths so the parser needs to handle that.


weshaggard:
@danieljurek are you aware of any reason we should update to make this change?

Issue #13722, opened 2026-01-26
[APIView] Missing Package Version in JavaScript Reviews Causes Copilot Review Detection Failure


AlitzelMendez:
In some cases, specifically for JavaScript reviews, the `packageVersion` field is missing or empty in the review data. This causes the Copilot review detection logic to fail when trying to determine if a review has been reviewed by Copilot.

For example: [@azure/storage-blob](https://spa.apiview.dev/review/481422771b7d4e2cb37cc95fc00faba2?activeApiRevisionId=acfa9e00b3894a0aa594a19ac99260ce&diffApiRevisionId=f7073d29844b4dcf996ba2891d87e006&diffStyle=trees)

## Scenario

When checking if a review version has been reviewed by Copilot, the system relies on the `packageVersion` to identify the specific revision:

```cshpar
getIsReviewVersionReviewedByCopilot(this.review.id, packageVersion)
```

## Solution

Added a simple check to exempt reviews without package versions from the Copilot review requirement

The reason is that a version is crucial to know if the review is even required by copilot, we need to know if is a preview version or no

PR #13724, opened 2026-01-26
[APIView] Fix Copilot Review Require Detection when version not available

AlitzelMendez:
closes: https://github.com/Azure/azure-sdk-tools/issues/13722


AlitzelMendez:
> It seems to me if version isn't provided, there's no way we should be able to approve it.

that is a good point, I talked with Marian and she agrees,  I will follow up with the person that had the complaint, to clarify this, I put copilot as a requirement again for JS

tjprescott:
If we disable the button and it gives a reason, we should distinguish between "Approval is disabled because a Copilot Review is required" and "Approval is disabled because this API has no version"

PR #13750, opened 2026-01-27
Mark package as released using azsdk cli command

praveenkuttappan:
Mark a package as released using azsdk cli command from release pipeline. This will ensure package release status is updated in telemetry whenever a package is released.

### DIFF
--- a/eng/common/pipelines/templates/steps/mark-release-completion.yml
+++ b/eng/common/pipelines/templates/steps/mark-release-completion.yml
@@ -6,7 +6,6 @@ parameters:
 steps:
   - template: /eng/common/pipelines/templates/steps/install-azsdk-cli.yml
     parameters:
-      InstallDirectory: $(Agent.TempDirectory)
       SourceRootPath: ${{ parameters.SourceRootPath }}
 
   - task: AzureCLI@2


eng/common/scripts/Mark-ReleasePlanCompletion.ps1
```diff
@@ -30,35 +40,21 @@ function Process-Package([string]$packageInfoPath)
 {
```
weshaggard, line 40: @benbp are you thinking about a generalized wrapper to handle a set of package info files or perhaps adding that logic into the cli itself?

@praveenkuttappan while this might work for now I do feel like it would like to eliminate this script and just call a cli command inline so that not every caller needs to have a wrapper script.

eng/common/pipelines/templates/steps/mark-release-completion.yml
```diff
@@ -4,13 +4,25 @@ parameters:
   SourceRootPath: $(Build.SourcesDirectory)
 
 steps:
+  - task: Powershell@2
+    displayName: 'Install Azure SDK MCP'
+    continueOnError: true
+    condition: and(succeeded(), ne(variables['Skip.MarkReleaseCompletion'], 'true'))
```
weshaggard: We will like want this took for other steps as well so we probably should make a separate template and not base it on this skip variable but perhaps another variable that indicates that it is already installed or not.

eng/common/pipelines/templates/steps/mark-release-completion.yml
```diff
@@ -4,13 +4,23 @@ parameters:
   SourceRootPath: $(Build.SourcesDirectory)
 
 steps:
+  - task: Powershell@2
```
benbp: Eventually we should move this into a more common pre-req step in pipelines so we don't end up installing this a bunch of times

eng/common/scripts/Mark-ReleasePlanCompletion.ps1
benbp: Agree if/when we can remove the outer script that is ideal

eng/common/scripts/Mark-ReleasePlanCompletion.ps1
```diff
@@ -30,35 +33,16 @@ function Process-Package([string]$packageInfoPath)
...
-        Write-Host "Multiple release plans found for package name: $PackageName with work item IDs: $concatenatedIds. Using the first release plan to update release status."
-        $activeReleasePlan = $workItems[0]
-    }
-    # Update release status
-    Write-Host "Release plan work item ID: $($activeReleasePlan["id"])"
-    Write-Host "Marking release completion for package, name: $PackageName version: $PackageVersion"
-    Update-ReleaseStatusInReleasePlan $activeReleasePlan.id "Released" $PackageVersion
-    Write-Host "Successfully marked release completion for package, name: $PackageName version: $PackageVersion."
+    Write-Host "Marking release completion for package, name: $PackageName"
+    $releaseInfo = &$AzsdkExePath release-plan update-release-status -p $PackageName -l $LanguageDisplayName -s "Released"
```
benbp:
```suggestion
    $releaseInfo = & $AzsdkExePath release-plan update-release-status -p $PackageName -l $LanguageDisplayName -s "Released"
```

benbp: more readable

benbp: Also I think we should use the full option names instead of the shorthand for readability in CI yaml

eng/common/pipelines/templates/steps/mark-release-completion.yml
```diff
@@ -4,13 +4,18 @@ parameters:
...
+
   - task: AzureCLI@2
     inputs:
       azureSubscription: opensource-api-connection
       scriptType: pscore
       scriptLocation: scriptPath
       scriptPath: ${{ parameters.SourceRootPath }}/eng/common/scripts/Mark-ReleasePlanCompletion.ps1
-      arguments: -PackageInfoFilePath '${{ parameters.ConfigFileDir }}/${{ parameters.PackageArtifactName }}.json'
+      arguments: >
+        -PackageInfoFilePath '${{ parameters.ConfigFileDir }}/${{ parameters.PackageArtifactName }}.json' -AzsdkExePath "$(Agent.TempDirectory)/azsdk$(if ($IsWindows) { '.exe' } else { '' })"
```
weshaggard: Is the windows logic to add the .exe needed? I think windows should work with out.

weshaggard: Image

praveenkuttappan: Thanks. Fixed.

eng/common/scripts/Mark-ReleasePlanCompletion.ps1
benbp: @weshaggard I'd prefer to have the looping/batching within the cli, yes

eng/common/pipelines/templates/steps/install-azsdk-cli.yml
```diff
@@ -0,0 +1,13 @@
...
+  InstallDirectory: $(Agent.TempDirectory)
+
+steps:
+  - task: Powershell@2
+    displayName: 'Install Azure SDK Tools CLI'
+    inputs:
+      targetType: 'inline'
+      script: >
+        ./eng/common/mcp/azure-sdk-mcp.ps1
+        -InstallDirectory ${{ parameters.InstallDirectory }}
```
benbp: We could have this template set the install directory as a pipeline variable, then other locations could just reference `$(AZSDK)`

benbp (approved):

praveenkuttappan:
Dependency: This PR can be merged only after https://github.com/Azure/azure-sdk-tools/pull/13796 is merged and released.

eng/common/pipelines/templates/steps/mark-release-completion.yml
```diff
@@ -4,13 +4,19 @@ parameters:
   SourceRootPath: $(Build.SourcesDirectory)
 
 steps:
+  - template: /eng/common/pipelines/templates/steps/install-azsdk-cli.yml
+    parameters:
+      InstallDirectory: $(Agent.TempDirectory)
```
weshaggard: I would exclude this if it is the default value.

praveenkuttappan: Done.

eng/common/pipelines/templates/steps/mark-release-completion.yml
praveenkuttappan: This is still required. Pipeline failed with command not found without exe on .NET release pipeline.

Issue #13744, opened 2026-01-27
[azsdk] Investigate microagents vs github copilot sdk


chrisradek:
Now that https://github.com/github/copilot-sdk is public, investigate migrating from microagents to the GitHub Copilot SDK.

Some more context:

The TypeSpec customization microagent struggled with the current microagent approach due to the complexity of the requests causing the processing time on the LLM side to exceed our timeouts. Increasing client timeouts and switching to the streaming API helped somewhat, but was still seeing excessively long times for the agent to complete work (median of 5.3 minutes over 29 runs with some taking over 30 minutes)

richardpark-msft:
Switching over also alleviates _us_ from having to own/fund an OpenAI instance, which would be awesome as well. It also lets the user pick the model, rather than us being locked into a specific one. 

There's a lot of benefits, even if it offers equivalent performance.

chrisradek:
Posting some numbers from experiments.

## The setup
I created an agent in azsdk to apply TypeSpec customizations to a project based on a request.
The target typespec project was the ai Document Intelligence service, with the following prompt:
> For Python SDK, add operation groups to DocumentIntelligenceAdministrationClient to organize the administration operations into logical subclients:
> 1. A 'Models' operation group for document model operations (buildModel, getModel, listModels, deleteModel, composeModel, copyModelTo, etc.)
> 2. A 'Classifiers' operation group for classifier operations (buildClassifier, getClassifier, listClassifiers, deleteClassifier, etc.)
> The goal is to improve the Python SDK ergonomics by grouping related operations together.

I tested several different agent options:
- microagent: standard microagents we have today
- microagent with streaming API: microagents using the OpenAI Streaming API
- GitHub Copilot SDK
  - Opus 4.5
  - Sonnet 4.5
  - GPT-5.2-Codex
 
For each of these, I ran the agent ~30 times.

## Results

### Microagents (GPT-5)

The majority of these requests failed due to repeated 408 (client timeout) errors. The client timeout for the OpenAI client is set to 100 seconds, and after the first 1-2 requests, GPT-5 was taking longer than that to return a full response.

### Consolidated results

| Metric | Microagents w/streaming (GPT-5) | Copilot SDK (Sonnet 4.5) | Copilot SDK (Opus 4.5) | Copilot SDK (GPT-5.2-Codex) |
|--------|---------------------|--------------------------|------------------------|----------------------------|
[...]

PR #13777, opened 2026-01-29
[azsdk] Add TypeSpec Customization service with copilot agent

### LINKED ISSUE #13553: Create Phase A microagent for TypeSpec client customizations
## Summary

This issue tracks creating the microagent used for applying TypeSpec customizations in the `azsdk_package_customize_code` tool.

The microagent takes the following inputs:
- TypeSpec project path
- Context containing client customization requests, such as SDK generation/build failures and free form requests.

The microagent is responsible for:
- identifying TypeSpec customizations that can be made given the provided information
- Creating/modifying `client.tsp` in the TypeSpec project
- Verifying that `client.tsp` compiles
- Reporting the list of changes made to `client.tsp` to address the input context

## Acceptance Criteria
Microagent is able to:
- identify which parts of the context can be solved with client.tsp customizations
- Make in-place edits to `client.tsp` without corrupting the file
- Make in-place edits to `client.tsp` without altering existing, unrelated customizations
- Reports concise summary of changes

## Extra considerations

We have limited tools available to microagents today. We can read and edit entire files, but don't have search capabilities. This may cause context window issues if the microagent attempts to read entire TypeSpec projects. With testing, we may need to identify ways to handle cases where the context window limits are approached. Optimizations may be follow-up issues, but we should avoid outright failing in this scenario.

chrisradek:
Fixes #13553 

This PR adds support for Copilot SDK agents. See https://github.com/Azure/azure-sdk-tools/issues/13744 for more details on comparing microagents to copilot sdk agents.

This also adds the TypeSpec Customization service which uses the copilot agents.

There are 2 issues that need to be sorted out and may block migrating existing microagents to copilot SDK:
1. The copilot SDK requires the copilot CLI to be installed on the host machine. The copilot SDK is planning to spend some time cleaning up the packaging story (https://github.com/github/copilot-sdk/issues/248) but in the meantime we'll either need to ask users to install the copilot CLI, or perhaps even install it via the azsdk powershell install script.
2. The copilot SDK doesn't automatically work with GitHub Copilot Coding Agent (web site). To work in this context, we would need to have a PAT that we can then pass to the azsdk as an environment variable.


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/TypeSpec/TypeSpecCustomizationService.cs
```diff
@@ -0,0 +1,164 @@
...
+        // Create the tools using shared tool factories
+        var tools = CreateTools(typespecProjectPath);
+
+        // Create and run the copilot agent
+        var agent = new CopilotAgent<TypeSpecCustomizationResult>
+        {
+            Instructions = instructions,
+            Tools = tools,
+            MaxIterations = maxIterations,
+            Model = "claude-opus-4.5"
```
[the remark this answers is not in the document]
chrisradek: This was intentionally chosen due to the faster average speed to complete customization tasks compared to sonnet.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
```diff
@@ -85,6 +88,24 @@ public static void RegisterCommonServices(IServiceCollection services, OutputHel
...
+                var logger = sp.GetService<ILogger<CopilotClient>>();
+                return new CopilotClient(new CopilotClientOptions
+                {
+                    UseStdio = true,
+                    AutoStart = true,
+                    Logger = logger
+                });
+            });
+            services.AddSingleton<ICopilotClientWrapper, CopilotClientWrapper>();
+            services.AddScoped<ICopilotAgentRunner, CopilotAgentRunner>();
```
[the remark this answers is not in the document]
chrisradek: The client is a singleton so that a single underlying copilot CLI process can be shared across sessions.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Azure.Sdk.Tools.Cli.csproj
```diff
@@ -24,6 +25,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.AI.Agents.Persistent" Version="1.2.0-beta.5" />
+    <PackageReference Include="GitHub.Copilot.SDK" Version="0.1.18" />
```
deyaaeldeen:
```suggestion
    <PackageReference Include="GitHub.Copilot.SDK" Version="0.1.20" />
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
```diff
@@ -85,6 +88,24 @@ public static void RegisterCommonServices(IServiceCollection services, OutputHel
             services.AddScoped<IAzureAgentServiceFactory, AzureAgentServiceFactory>();
             services.AddScoped<ICommonValidationHelpers, CommonValidationHelpers>();
 
+            // Copilot SDK services for new agents (CopilotAgent<T> pattern)
+            services.AddScoped<CopilotTokenUsageHelper>();
+            services.AddSingleton<CopilotClient>(sp =>
+            {
+                var logger = sp.GetService<ILogger<CopilotClient>>();
+                return new CopilotClient(new CopilotClientOptions
```
deyaaeldeen, line 97: should we verify it is authenticated first? `GetAuthStatusAsync` returns a `res.IsAuthenticated`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/Tools/FileTools.cs
```diff
@@ -0,0 +1,121 @@
...
+                var entries = Directory.GetFileSystemEntries(path, searchPattern, searchOption)
+                    .Select(f => Path.GetRelativePath(baseDir, f))
+                    .ToArray();
+
+                return entries;
+            },
+            "ListFiles",
+            description);
+    }
+}
```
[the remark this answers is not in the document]
chrisradek: Done

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
chrisradek: Yeah - turns out this is an important step. At least at the moment no error is thrown if you start a session and can't authenticate: copilot just hangs forever because it never gets into an idle state.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotAgentRunner.cs
```diff
@@ -0,0 +1,215 @@
...
+public class CopilotAgentRunner(
+    ICopilotClientWrapper client,
+    CopilotTokenUsageHelper tokenUsageHelper,
+    ILogger<CopilotAgentRunner> logger) : ICopilotAgentRunner
+{
+    /// <summary>
+    /// Ensures the GitHub Copilot CLI is installed and authenticated.
+    /// Throws a descriptive exception if not available or not authenticated.
+    /// </summary>
+    private async Task EnsureAuthenticatedAsync(CancellationToken ct)
```
benbp, line 19: Nit: the `Async` suffixes are unnecessary.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotAgentRunner.cs
```diff
@@ -0,0 +1,215 @@
...
+    {
+        CopilotAuthStatus authStatus;
+        try
+        {
+            authStatus = await client.GetAuthStatusAsync(ct);
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
```
benbp, line 26: Might be better UX to first check if it's installed and then check auth so we can serve more precise error messages.

benbp: Or have some sort of `IsInstalled` cached property

chrisradek: That's sort of what I'm doing with this one call. If it throws - then I assume copilot cli isn't installed. I could try to see if the command is available instead though.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotAgentRunner.cs
chrisradek: My newness to c# is showing 😅

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotAgentRunner.cs
```diff
@@ -0,0 +1,185 @@
...
+            // SendAsync returns the message ID but doesn't wait for processing
+            // We need to wait for SessionIdleEvent to know when the agent is done
+            await session.SendAsync(new MessageOptions { Prompt = prompt }, ct);
+            
+            // Wait for the session to become idle (all tool calls completed)
+            // Use cancellation-aware wait with timeout to prevent indefinite hangs
+            logger.LogDebug("Waiting for session to become idle...");
+            using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(ct);
+            linkedCts.CancelAfter(agent.IdleTimeout);
+            await sessionIdleTcs.Task.WaitAsync(linkedCts.Token);
```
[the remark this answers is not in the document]
benbp: I think copilot has a good point with this review comment. We do a similar thing in the process helper.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotAgentRunner.cs
```diff
@@ -0,0 +1,215 @@
...
+            if (agent.ValidateResult != null)
+            {
+                var validation = await agent.ValidateResult(capturedResult);
+                if (!validation.Success)
+                {
+                    var reason = validation.Reason is string str
+                        ? str
+                        : JsonSerializer.Serialize(validation.Reason);
+                    logger.LogWarning("Agent result failed validation: {Reason}. Retrying.", reason);
+                    prompt = $"The result you provided did not pass validation: {reason}. Please try again.";
```
benbp: maybe the biggest nit I've ever suggested

```suggestion
                    prompt = $"The result you provided did not pass validation: {reason}. Try again.";
```

chrisradek: It does save at least 1 token.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/TypeSpecCustomizationResult.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
```
benbp: Should this go in a `Models` sub-directory+namespace like `TypeSpecCustomizationServiceResult.cs` below?

benbp: Also these two files appear identical

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Prompts/Templates/TypeSpecCustomizationTemplate.cs
```diff
@@ -0,0 +1,125 @@
...
+        this.referenceDocPath = referenceDocPath;
+    }
+
+    /// <summary>
+    /// Builds the complete TypeSpec customization prompt using the configured parameters.
+    /// </summary>
+    /// <returns>Complete structured prompt for TypeSpec client.tsp customization</returns>
+    public override string BuildPrompt()
+    {
+        var referenceDocContent = File.ReadAllText(referenceDocPath);
```
benbp: nit: use the async method and pass down the ct that's in the calling method.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/TypeSpecCustomizationResult.cs
chrisradek: > Also these two files appear identical

Ah you're right - they used to report failures differently but now they are the same. I'll remove this one

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/TypeSpec/TypeSpecCustomizationService.cs
benbp, line 120: I wonder if we want to maintain some sort of centralized model config that will make it easier to rev these when they get updated, vs. having to search/replace.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotTokenUsageHelper.cs
```diff
@@ -0,0 +1,97 @@
...
+/// Tracks token usage for Copilot SDK agents.
+/// 
+/// The Copilot SDK reports inputTokens as the cumulative context size (not incremental).
+/// Each turn's inputTokens = system prompt + all previous messages + current turn.
+/// outputTokens IS incremental per turn, so we sum those.
+/// cacheReadTokens shows how much of inputTokens came from cache (not reprocessed).
+/// 
+/// To avoid overcounting, we track the previous cumulative inputTokens and compute deltas.
+/// </summary>
+public class CopilotTokenUsageHelper(IRawOutputHelper outputHelper)
```
benbp: This seems like a rip of `TokenUsageHelper`, can we not just extend that, or inherit? Seems like it's possibly just the incremental token counting that's different?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/TypeSpec/TypeSpecCustomizationService.cs
```diff
@@ -0,0 +1,164 @@
...
+        };
+
+        logger.LogInformation("Running TypeSpecCustomization copilot agent with {ToolCount} tools and max {MaxIterations} iterations...",
+            tools.Count, maxIterations);
+
+        var result = await copilotAgentRunner.RunAsync(agent, ct);
+
+        tokenUsageHelper.LogUsage();
```
benbp: Should this be a part of `RunAsync()` so it always logs and people's don't have to remember to do this?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/TypeSpec/TypeSpecCustomizationService.cs
```diff
@@ -0,0 +1,164 @@
...
+    /// in eng/common/knowledge/ under the repository root.
+    /// </summary>
+    private async Task<string?> FindReferenceDocAsync(string typespecProjectPath, CancellationToken ct)
+    {
+        var repoRoot = await gitHelper.DiscoverRepoRootAsync(typespecProjectPath, ct);
+        if (string.IsNullOrEmpty(repoRoot))
+        {
+            return null;
+        }
+        var candidate = Path.Combine(repoRoot, "eng", "common", "knowledge", "customizing-client-tsp.md");
```
benbp, line 157: Should we add a unit test or something for this to make sure this file exists? Would fail in scenarios where we run an agent cli to do testing sandboxed to `tools/azsdk-cli` though :/

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
```diff
@@ -85,6 +88,24 @@ public static void RegisterCommonServices(IServiceCollection services, OutputHel
...
+                {
+                    UseStdio = true,
```
benbp, line 99: Not sure how this will interplay with our requirement for stdio to be captured and translated to json-rpc in mcp mode

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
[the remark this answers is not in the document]
benbp: probably worth a code comment above

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotTokenUsageHelper.cs
chrisradek: Yeah, incremental token counting is the biggest difference. I can look extending it.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
chrisradek: Hmmm, I tested this with a CLI command, but not in an MCP tool. If it works when invoked via MCP, can I assume this is ok? Or is there more to watch out for?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/TypeSpec/TypeSpecCustomizationService.cs
chrisradek: Sure, I can add one. The live test exercises this path, but it won't run in CI unless copilot is installed and authenticated.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/ServiceRegistrations.cs
chrisradek: I tested running the agent in an MCP tool and it didn't cause any problems. The final result was passed back from the MCP tool as expected

benbp: Oh never mind I misunderstood what that field was doing. It's not pushing to stdout it's just using stdio as the protocol

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotAgentRunner.cs
benbp: It's somewhat conventional naming when there are libraries that support async and sync methods, so it sneaks in here and there, but I prefer to avoid it if not required.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotAgentRunner.cs
benbp: Ah I understand, so auth status won't throw when unauthenticated.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotTokenUsageHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+/// Tracks token usage for Copilot SDK agents.
+/// 
+/// The Copilot SDK reports inputTokens as the cumulative context size (not incremental).
+/// Each turn's inputTokens = system prompt + all previous messages + current turn.
+/// outputTokens IS incremental per turn, so we sum those.
+/// cacheReadTokens shows how much of inputTokens came from cache (not reprocessed).
+/// 
+/// To avoid overcounting, we track the previous cumulative inputTokens and compute deltas.
+/// </summary>
+public class CopilotTokenUsageHelper(IRawOutputHelper outputHelper) : TokenUsageHelper(outputHelper)
```
benbp: I was thinking about this more and remembered/realized that TokenUsageHelper is really meant to be a singleton state bag per request (hence `AddScoped()`). So if we inherit from it, but set the same telemetry tags, that means we could end up losing token usage info in scenarios where both the `TokenUsageHelper` instance AND the `CopilotTokenUsageHelper` instance are updated (not common right now, more defensive programming). Last writer to telemetry would win in that case. What we really want is for all components using tokens to update into the same token usage helper.

One idea is to move the copilot helper `Add()` method into `TokenUsageHelper` as `AddCumulative()` or `AddForCopilotCli()` and then calling that from the copilot agent runner.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotAgentRunner.cs
```diff
@@ -0,0 +1,225 @@
...
+            using var timeoutCts = new CancellationTokenSource(agent.IdleTimeout);
+            using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(ct, timeoutCts.Token);
+            try
+            {
+                await sessionIdleTcs.Task.WaitAsync(linkedCts.Token);
+            }
+            catch (OperationCanceledException) when (timeoutCts.IsCancellationRequested)
+            {
+                throw new TimeoutException(
+                    $"Agent session idle timeout of {agent.IdleTimeout} was exceeded while waiting for the agent to complete.");
```
benbp:
```suggestion
                    $"Agent session idle timeout of {agent.IdleTimeout.Minutes}m was exceeded while waiting for the agent to complete.");
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/CopilotAgents/CopilotTokenUsageHelper.cs
chrisradek: @benbp I've merged the 2 into TokenUsageHelper and got rid of the Copilot-specific one.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/TokenUsageHelper.cs
```diff
@@ -5,17 +5,45 @@ namespace Azure.Sdk.Tools.Cli.Helpers;
...
+    /// 
+    /// Note: Agents should use either <see cref="Add"/> or <see cref="AddCumulative"/>, not both.
+    /// Mixing the two will produce incorrect prompt token counts.
+    /// </summary>
+    /// <param name="model">The model used for this turn.</param>
+    /// <param name="inputTokens">Cumulative context size (not incremental).</param>
+    /// <param name="outputTokens">Output tokens for this turn (incremental — added to total).</param>
+    public void AddCumulative(string model, double inputTokens, double outputTokens)
+    {
+        Add(model, inputTokens - promptTokens, outputTokens);
```
benbp: Is this new calculation correct? I had been trying to understand the original implementation and it seemed like it was trying to calculate and substract cached tokens from the running total, whereas this new method doesn't seem correct.

chrisradek:
Created https://github.com/Azure/azure-sdk-tools/issues/13919 as a follow-up for TokenUsageHelper sessions.

benbp (approved):

PR #13796, opened 2026-01-30
Disable response file handling for command line

praveenkuttappan:
azsdk command line fails with "Response file not found" when command line param value contains "@"

for e.g. Following command to release @azure/template package fails.
Command: `.\azsdk package release --[ackage-name "@azure/template" --language "JavaScript"`
Error: `Error reading response file 'azure/template': Could not find a part of the path 'C:\git\azure\template'`

Fix is to disable response file handling when parsing cli command.


maririos (approved):
Are there tests we can add to validate this behavior change?

praveenkuttappan:
> Are there tests we can add to validate this behavior change?

@maririos  I have added a test with a config to disable response file handling. Actual change in Command line static method and the test replicate the same parsing logic.

PR #13818, opened 2026-01-31
[tsp-client] remove autorest dep

catalinaperalta:
Autorest is reaching eol, this pr removes the dependency from tsp-client.


tools/tsp-client/src/commands.ts
```diff
@@ -599,51 +598,42 @@ export async function convertCommand(argv: any): Promise<void> {
...
+      try {
+        await unlink(joinPaths(rootUrl, "resources.json"));
+      } catch (err) {
+        Logger.error(`Error occurred while attempting to delete resources.json: ${err}`);
+        process.exit(1);
+      }
+    }
+  } catch (err) {
+    Logger.error(
+      `An error occurred during convert command. Verify that you have autorest installed and accessible. Error: ${err}`,
```
catalinaperalta: This is just a catch all for any error and point out that this command relies on autorest

PR #13831, opened 2026-02-02
Support version and release date update in the CHANGELOG.md

### LINKED ISSUE #13830: Support version and release date update in the CHANGELOG.md
It should find the specified entry in the `CHANGELOG.md` based on the input version. Then update the `release date` accordingly.

raych1:
Resolved https://github.com/Azure/azure-sdk-tools/issues/13830

**Key Changes:**
- Added `IChangelogHelper` and `ChangelogHelper` for parsing and updating CHANGELOG.md files
- Implemented common logic in `LanguageService.UpdateVersionAsync` to:
  - Validate changelog exists and has an entry for the target version
  - Update release date in changelog
- Added base method `UpdatePackageVersionInFilesAsync`


weshaggard:
With this change will we be able to remove our changelog PS scripts?

raych1:
> With this change will we be able to remove our changelog PS scripts?

This PR only migrates some functions of the changelog PS scripts which are needed in the version update scenario. If the goal is to remove the PS scripts, I will create a separate task for that.

weshaggard:
Yes please file another issue to finish that as it would be good to get all the changelog support in the cli

raych1:
> Yes please file another issue to finish that as it would be good to get all the changelog support in the cli

I've created https://github.com/Azure/azure-sdk-tools/issues/13855.

weshaggard (approved):

PR #13859, opened 2026-02-04
[Teams Chatbot] Refine static Q&A knowledge

JiaqiZhang-Dev:
Resolved: https://github.com/Azure/azure-sdk-pr/issues/2496
Guideline: https://github.com/Azure/azure-sdk-tools/blob/7dfc8ca0d8d63b6a910dd79cdebd98757495fa8a/tools/sdk-ai-bots/docs/KNOWLEDGE_REFINEMENT_GUIDE.md

- Stage 1: Refine the existing typespec q&a into knowledge. ETA: 3.13

### DIFF
--- a/tools/sdk-ai-bots/docs/KNOWLEDGE_REFINEMENT_GUIDANCE.md
+++ b/tools/sdk-ai-bots/docs/KNOWLEDGE_REFINEMENT_GUIDANCE.md
@@ -86,7 +86,7 @@ For ongoing Q&A collected from weekly badcase reviews:
 
 ### Data Source
 
-https://ms.portal.azure.com/#view/Microsoft_Azure_Storage/ContainerMenuBlade/~/overview/storageAccountId/%2Fsubscriptions%2Fa18897a6-7e44-457d-9260-f2854c0aca42%2FresourceGroups%2Fazure-sdk-qa-bot%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2Fazuresdkqabotstorage/path/evaluation-dataset/etag/%220x8DDEF6EC73F0C18%22/defaultId//publicAccessVal/None
+The `evaluation-dataset` folder of the production environment storage resource.
 
 ### Criteria for Good Candidates
 
@@ -202,7 +202,7 @@ Before adding new knowledge, follow this verification process:
 
 If the bot cannot answer correctly, search for existing knowledge:
 
-- AI Search Portal: [azure-sdk-knowledge index](https://ms.portal.azure.com/#view/Microsoft_Azure_Search/Index.ReactView/id/%2Fsubscriptions%2Fa18897a6-7e44-457d-9260-f2854c0aca42%2FresourceGroups%2Fazure-sdk-qa-bot-dev%2Fproviders%2FMicrosoft.Search%2FsearchServices%2Fazuresdkqabot-dev-search%23azure-sdk-knowledge/location/West%20US%202/sku/standard)
+- Open the production environment Azure AI Search resource
 - Search for keywords related to the question
 - Check if similar content already exists
 


jianingwang123:
I think the bot should be able to answer this case based on existing knowledge (especially removing and renaming the old property and adding a new property with the new default). maybe need to track.
```
# 13.Default value starting from a specific API version?

## question 
Hello, we currently have a property (StatelessServiceProperties,minInstancePercentage) that does not currently have a default value in our spec, but in practice is treated as if it is 0. We would like to treat the default prior to 2025-06-01 as 0, then as a different value from 2025-06-01 onward in our service.
1. We haven't changed the default for an existing property before. Are there concerns about this intended behavior? 
2. In Typespec, is it possible to add a default value in our spec from a specific api version (say 2025-06-01) onward?

## answer
I think you need to run this change by the breaking change board, as a change in the default may be breaking, depending on the details.  Also, the most important thing is to make sure that the API description accurately reflects service behavior - if the default has always been in place, for example, it may be better to just change the default and go through the breaking change process.  Yes, it is possible to do this in TypeSpec, but involves removing and renaming the old property and adding a new property with the new default, like this

/** Employee properties */
model EmployeeProperties {
  /** Age of employee */
  @removed(Versions.`2024-11-01-preview`)
  @renamedFrom(Versions.`2024-11-01-preview`, "age")
  formerAge?: int32;

  /** Age of employee */
  @added(Versions.`2024-11-01-preview`)
  age?: int32 = 21;
}

```
bot response:
```
1) Concerns with changing the default behavior
- Yes. Changing a default value is considered a breaking change risk for existing clients. The Azure breaking-change policy explicitly calls out “Change default value” as a change that can break old clients because they “don’t expect new default behavior”.
- The Azure versioning guidelines also say you must not change the behavior of an API version that is already available to customers (preview or GA). So if 2025-06-01 is a new api-version, the safe approach is:
  - Keep the old behavior (default effectively 0) for all older api-versions.
  - Introduce the new default only in the new api-version (2025-06-01).

2) Can TypeSpec express a default value that applies only from a specific api-version onward?
- TypeSpec supports schema defaults using `=` on a property (this produces an OpenAPI `default`), but defaults are intended to represent server-side defaults and should not be treated as client-side defaults.
- The provided guidance does not describe any versioning decorator that can make a property’s `= <default>` vary by api-version. In other words, there isn’t documented support here for “default is 0 before 2025-06-01, and X starting 2025-06-01” as a versioned schema-default.
- If your goal is SDK/client-side defaults, there is a legacy decorator `@Azure.ClientGenerator.Core.Legacy.clientDefaultValue(...)`, but it is explicitly discouraged for new services and is intended only to mitigate SDK breaking changes for brownfield services.

Given the above, the compliant path is to implement the behavior change only in the new api-version (2025-06-01) at the service level, and (optionally) document the server-side default in the spec. If you want a concrete TypeSpec pattern for documenting different defaults per version, please share the relevant snippet of your version enum and the `StatelessServiceProperties` model so I can suggest an approach that stays within the supported versioning patterns in the guidance.
```

JiaqiZhang-Dev:
Onwner| File | Count | Progress
-- | -- | -- | --
Kyle Zhang|typespec_2025_05_22.md | 25 | https://github.com/Azure/azure-sdk-tools/pull/13859/commits/aabe38c6a57c332b4efd29ae90683f09cdbbbb12
Kyle Zhang| typespec_2025_06_26.md | 33 | [7db9a89](https://github.com/Azure/azure-sdk-tools/pull/13859/commits/7db9a89c7836f9d9ed55b397e08660ec848e06d3)
Kyle Zhang| typespec_2025_07_03.md | 13 | [7db9a89](https://github.com/Azure/azure-sdk-tools/pull/13859/commits/7db9a89c7836f9d9ed55b397e08660ec848e06d3)
Kyle Zhang| typespec_2025_09_26.md | 2 | [7db9a89](https://github.com/Azure/azure-sdk-tools/pull/13859/commits/7db9a89c7836f9d9ed55b397e08660ec848e06d3)
Jianing Wang|typespec_2025_05_22.md | 25 | https://github.com/Azure/azure-sdk-tools/pull/13859/changes/de15846f4df3974f5d248bc4b9637d913a762f1a
Jianing Wang | typespec_2025_07_10.md | 3| N/A (no items to refine)
Jianing Wang| typespec_2025_07_17.md | 6| https://github.com/Azure/azure-sdk-tools/pull/13859/commits/73abd654b294f9bfdfba588c0f89f505053b4132
Jianing Wang| typespec_2025_07_24.md | 7| https://github.com/Azure/azure-sdk-tools/pull/13859/commits/73abd654b294f9bfdfba588c0f89f505053b4132
Jianing Wang| typespec_2025_07_31.md | 4| https://github.com/Azure/azure-sdk-tools/pull/13859/commits/73abd654b294f9bfdfba588c0f89f505053b4132
Jianing Wang | typespec_2025_08_07.md | 3| https://github.com/Azure/azure-sdk-tools/pull/13859/commits/73abd654b294f9bfdfba588c0f89f505053b4132
Jianing Wang| typespec_2025_08_14.md | 6| N/A (no items to refine)
Jianing Wang | typespec_2025_09_12.md | 3| N/A (no items to refine)
Jianing Wang | typespec_2025_09_19.md | 3| N/A (no items to refine)
Jianing Wang| typespec_2025_10_10.md | 9| https://github.com/Azure/azure-sdk-tools/pull/13859/commits/73abd654b294f9bfdfba588c0f89f505053b4132
Jianing Wang | typespec_2025_10_17.md | 3| N/A (no items to refine)
Total |   | 145

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/management-plane/arm-templates.md
```diff
@@ -0,0 +1,166 @@
...
+model MyResourceProperties {
+  metadata?: KeyValuePair[];
+}
+```
+
+This guidance applies to ARM (management plane) only.
+
+## Use @useFinalStateVia to control LRO final-state resolution in ARM PUT operations
+
+`final-state-via: "original-uri"` means the original PUT request URI is polled to retrieve the final resource state. ARM defaults to `Azure-AsyncOperation` when that header is present, even for PUT operations.
```
chunyu3: maybe we need to re-arrange the content:

Final-state can be retrieved from `"original-uri"`, `"azure-async-operation"`, `"location"`. ARM defaults to `Azure-AsyncOperation` when that header is present, even for PUT operations.

Use `@useFinalStateVia` can override the default. e.g. Use `@useFinalStateVia` to override to `original-uri`:
```typespec
@armResourceOperations
interface Employees {
  @useFinalStateVia("original-uri")
  create is ArmResourceCreateOrReplaceAsync<Employee>;
}
```

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/management-plane/arm-templates.md
```diff
@@ -0,0 +1,166 @@
...
+interface Employees {
+  move is ArmResourceActionAsync<Employee, MoveRequest, MoveResponse>;
+}
+```
+
+The future direction is to represent the final result only in `x-ms-long-running-operation-options` via `final-state-schema`, but not all language emitters support this yet.
+
+## Adding final-state-schema for a single POST action LRO
+
+Set `FinalResult` on the LRO header type for that specific operation to emit `final-state-schema` scoped to only that operation, without requiring `emit-lro-options: "all"` globally.
```
chunyu3: `final-state-schema` is a thermology of swagger. curious whether user will know about it and ask question to add this.

And  `LroHeaders = ArmAsyncOperationHeader<FinalResult = MyActionResponse>' will not emit `final-state-schema`. please confirm this

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/management-plane/arm-templates.md
```diff
@@ -0,0 +1,166 @@
...
+```typespec
+...ResourceNameParameter<
+  Resource = VolumeGroup,
+  SegmentName = "volumegroups",  // consistent lowercase
+>;
+```
+
+Favor camelCase for fixed segments (e.g., `/resourceGroups`). TypeSpec templates such as `ArmResourceListByParent` automatically use the segment from the resource model.
+
+## Same model cannot have both SubscriptionLocationResource and ResourceGroupLocationResource scopes
```
chunyu3, line 65: Is this knowledge about 'multiple-path'?
if so, maybe we should convert to the general scenario.

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/management-plane/arm-templates.md
```diff
@@ -0,0 +1,166 @@
...
+**The core reason** is semantic, not tooling limitation. ResourceModelWithAllowedPropertySet exists as a legacy compatibility helper used in limited examples; it is not treated as an actual ARM resource type in TypeSpec. As a result, it cannot represent a true resource boundary and is intentionally rejected by ARM resource validation. Carrying this pattern forward would preserve Swagger-era modeling quirks and make the spec harder to evolve.
+**The correct approach** is to model the resource as a proper ARM resource type, typically TrackedResource, and reuse the legacy “allowed property set” via composition rather than inheritance. Resource-specific fields should live in a separate properties model and be spread or referenced, keeping the resource definition accurate and future-proof. This approach aligns with TypeSpec ARM guidelines, avoids invalid base types, and produces clearer, more maintainable specifications and SDKs.
+
+## Customizing Parent Resource Key Names for Child Operations
```
chunyu3: update key names is to update the resource segment part in the path. right?
Maybe we need to describe this here. User may not know to update a key name, but he may find out that the segment name is not correct, and need to update.

jianingwang123: > update key names is to update the resource segment part in the path. right? Maybe we need to describe this here. User may not know to update a key name, but he may find out that the segment name is not correct, and need to update.

I’ve updated this case and added a scenario to make it clearer—could you please re-review it?

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/management-plane/arm-templates.md
skywing918: No, from ElasticSan **volumeGroups** vs **volumegroups** case‑sensitivity inconsistency issue

jianingwang123:
Several cases need to track: 
doc exists, but the bot's answers are incorrect:
1. Looking for examples on running TypeSpec in ADO Pipelines
2. Swagger LintDiff Suppression via TypeSpec
3. What decorator should be used in this case?

bot’s answers are partly correct and partly hallucinating：
1. Is SDK generation required for the private preview?

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/management-plane/arm-templates.md
```diff
@@ -0,0 +1,239 @@
...
+}
+```
+
+## Async ARM resource action with no immediate body but final result on success
+
+Use `ArmResourceActionAsync<Resource, Request, Response>` where `Response` is the type returned in the `Succeeded` state. This currently emits both a `202 Accepted` (no body) and a `200 OK` with the response body, which is required by current lintdiff rules.
+
+```typespec
+@armResourceOperations
+interface Employees {
```
JiaqiZhang-Dev: @skywing918 could you help add the `MoveResponse` defination for this case?

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/validation.md
```diff
@@ -0,0 +1,187 @@
...
+namespace MyService.DataPlane {
+  enum Versions {
+    @useDependency(Azure.Core.Versions.v1_0_Preview_2)
+    @useDependency(MyService.Shared.Versions.v1)
+    v2024_01_01: "2024-01-01",
+  }
+}
+```
+
+The error "Namespace '' is referencing types from versioned namespace 'Azure.Core' but didn't specify which versions" means a namespace is using versioned types without declaring `@useDependency`.
```
JiaqiZhang-Dev: @skywing918 Please remove L57 and L71, and add these two notes:
1. To prevent ARM library dependencies in the data plane, use aliases or common types for modeling items like resource IDs.
2. Apply `@useDependency` when referencing versioned namespaces such as `Azure.Core` or `Azure.ResourceManager`.

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/validation.md
```diff
@@ -0,0 +1,187 @@
...
+
+This error typically indicates an outdated globally installed TypeSpec compiler. Update to the latest version:
+
+```bash
+npm install -g @typespec/compiler@latest
+```
+
+If working in the `azure-rest-api-specs` repository, follow the repo-specific setup documentation rather than installing globally. Ensure you are using Node.js 20+ and npm 7+.
+
+## Augmented decorators on resource name do not apply to LegacyOperations path parameters
```
JiaqiZhang-Dev: @skywing918 This case should be moved to `decorators.md`

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/validation.md
```diff
@@ -0,0 +1,187 @@
...
+
+The error "Namespace '' is referencing types from versioned namespace 'Azure.Core' but didn't specify which versions" means a namespace is using versioned types without declaring `@useDependency`.
+
+## visibility-sealed error when armResourceInternal is applied twice
+
+`TrackedResource<T>` already applies `@Azure.ResourceManager.Private.armResourceInternal` internally. Adding it again explicitly on a model that uses `is TrackedResource<T>` causes "Visibility of property 'name' is sealed and cannot be changed."
+
+**Fix**: Remove the redundant `@armResourceInternal` decorator. With `model is`, the decorator is already inherited from `TrackedResource<T>`. It was only needed when using `model extends` in older TypeSpec versions.
+
+## readme.md is required for SDK generation (Avocado failure)
```
JiaqiZhang-Dev: @skywing918 This file should only include tsv related issues, other spec checks should be put into `api-spec-review/spec-validation.md` file

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/validation.md
```diff
@@ -0,0 +1,187 @@
...
+
+Even for brand-new TypeSpec-based APIs, Avocado validates that a `readme.md` exists somewhere in the spec directory to configure SDK generation. TypeSpec compilation succeeds without it, but the SDK pipeline needs it to determine which API versions to generate and where the OpenAPI files are.
+
+The location and factoring of `readme.md` can vary by service. You need at least one in your service directory structure.
+
+## PR bot adds WaitForARMFeedback label even when CI checks fail
```
JiaqiZhang-Dev: @skywing918 please move this into `api-spec-review/spec-review-process.md` file

tools/sdk-ai-bots/azure-sdk-qa-bot-knowledge-sync/knowledge/typespec/validation.md
```diff
@@ -0,0 +1,187 @@
...
+When using tools like `oav generate-examples` to auto‑generate Swagger examples for a TypeSpec‑based spec, the generated examples may **not satisfy constraints defined in the spec**, such as string patterns or required formats. This is expected behavior: the example generator primarily produces a **structural placeholder**, not fully valid values.
+
+Common validation failures indicate that:
+
+- Example strings do not match required patterns
+- Example values do not satisfy required formats
+
+In most cases, the validation error message explicitly instructs you to replace the placeholder with a compliant value. The intended workflow is to **manually update the auto‑generated example values** so they conform to the constraints. Full placeholder materialization is not automated today and has not been prioritized, so manual adjustment is required to pass CI validation.
+
+## Swagger LintDiff Fails When Swagger Files Are Not Referenced by README
```
JiaqiZhang-Dev: @jianingwang123 please move this case into `api-spec-review/spec-validation.md` file

tools/sdk-ai-bots/docs/KNOWLEDGE_REFINEMENT_GUIDANCE.md
```diff
@@ -0,0 +1,347 @@
...
+- [ ] Evaluate each candidate using the criteria in Step 1
+- [ ] For suitable candidates, transform following Step 2 guidelines
+- [ ] Organize into appropriate category following Step 3
+- [ ] Submit PR with the new knowledge entries
+
+## Step 1: Evaluate Raw Q&A for Knowledge Potential
+
+### Data Source
+
+https://ms.portal.azure.com/#view/Microsoft_Azure_Storage/ContainerMenuBlade/~/overview/storageAccountId/%2Fsubscriptions%2Fa18897a6-7e44-457d-9260-f2854c0aca42%2FresourceGroups%2Fazure-sdk-qa-bot%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2Fazuresdkqabotstorage/path/evaluation-dataset/etag/%220x8DDEF6EC73F0C18%22/defaultId//publicAccessVal/None
```
lirenhe: If a document is public (or broadly shared), it must not contain real Azure resource URLs.
Can we use this way?


You can find the endpoint URL in:
Azure Portal → Your Resource → Overview → Endpoint

another approach is to use place holder e.g.

 https://{resource-name}.azurewebsites.net

tools/sdk-ai-bots/docs/KNOWLEDGE_REFINEMENT_GUIDANCE.md
```diff
@@ -0,0 +1,347 @@
...
+1. Remove `static_typespec_qa` from the knowledge sources(already done)
+2. Run the backend service locally(reference tools\sdk-ai-bots\azure-sdk-qa-bot-backend\README.md)
+3. Ask the bot the original question from the raw Q&A
+4. If the bot gives a correct answer → **No need to add** (knowledge already exists in other sources)
+
+**2. Search Existing Knowledge in AI Search**
+
+If the bot cannot answer correctly, search for existing knowledge:
+
+- AI Search Portal: [azure-sdk-knowledge index](https://ms.portal.azure.com/#view/Microsoft_Azure_Search/Index.ReactView/id/%2Fsubscriptions%2Fa18897a6-7e44-457d-9260-f2854c0aca42%2FresourceGroups%2Fazure-sdk-qa-bot-dev%2Fproviders%2FMicrosoft.Search%2FsearchServices%2Fazuresdkqabot-dev-search%23azure-sdk-knowledge/location/West%20US%202/sku/standard)
```
lirenhe: same above

tools/sdk-ai-bots/docs/KNOWLEDGE_REFINEMENT_GUIDANCE.md
JiaqiZhang-Dev: changed to 'The `evaluation-dataset` folder of the production environment storage resource.'

tools/sdk-ai-bots/docs/KNOWLEDGE_REFINEMENT_GUIDANCE.md
JiaqiZhang-Dev: changed to `Open the production environment Azure AI Search resource`

JiaqiZhang-Dev:
> When we will sync all those knowledge to storage to replace existing static-qa? Shall we update sync-config to regually sync the knowledge?

Once this PR merged, I'll sync those knowledge to storage by updating sync-config

PR #13860, opened 2026-02-04
[Customizations] add CLI tool to address apiview feedback via coding agent

### LINKED ISSUE #13768: [Client Customizations] add CLI command `azsdk review create-issue...` to address comments
Related to #13497. Will enable us to start automating the process to address APIView feedback.

Task: Create a CLI command that will fetch APIView Feedback and create a GH issue to address the feedback.

Deliverable:
* CLI command `azsdk review create-issue --apiview-url <url> ...`

Scope:
- Will require agent to merge all discussion comments into a single actionable comment (e.g. Needs more discussion, Keep as is, Update ...)
-  Does NOT require classifier or microagents described in #13497.

swathipil:
fixes: #13768


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedCommandGroups.cs
```diff
@@ -47,6 +47,11 @@ public static class SharedCommandGroups
             Description: "Sample operations for SDK packages"
         );
 
+        public static readonly CommandGroup Review = new(
+            Verb: "review",
```
praveenkuttappan: I believe there is already a group for apiview

swathipil: We wanted to avoid putting this under apiview since this isn't a functionality of the APIView service. Updating this to `tsp delegate-apiview-feedback`.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,591 @@
...
+using OpenAI.Chat;
+using Azure.Sdk.Tools.Cli.Services;
+using Azure.Sdk.Tools.Cli.Services.APIView;
+
+namespace Azure.Sdk.Tools.Cli.Helpers;
+
+/// <summary>
+/// Represents complete metadata for an APIView review
+/// </summary>
+public class ReviewMetadata
```
praveenkuttappan: Can you move these classes into models namespace?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,591 @@
...
+    public string RevisionId { get; set; } = string.Empty;
+    
+    [JsonPropertyName("pullRequestNo")]
+    public int? PullRequestNo { get; set; }
+    
+    [JsonPropertyName("pullRequestRepository")]
+    public string? PullRequestRepository { get; set; }
+    
+    [JsonPropertyName("revisionLabel")]
+    public string? RevisionLabel { get; set; }
```
praveenkuttappan: is label really required to resolve feedback? What's the use of these fields?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: yes!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: Yes, it's used for getting the tsp commit SHA that the PR should be working off of.

Flow: RevisionLabel may contain a SourceBranch (in the language repo) when the APIView is generated from a branch rather than a PR. In this case, we look in the language repo folder for tsp-location.yaml and the commit SHA listed there.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,591 @@
...
+}
+
+/// <summary>
+/// Represents a consolidated comment from a discussion thread
+/// </summary>
+// TODO: Add ThreadUrl property for direct links to APIView discussion threads
+public class ConsolidatedComment
+{
+    public string ThreadId { get; set; } = string.Empty;
```
praveenkuttappan: A line will have a list of comments. Some of them from reviewers and replies from package owners. Is there any need to tag the role of commenter when consolidating comment threads?
FYI @tjprescott @AlitzelMendez

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: Here's an example of what an issue looks like: https://github.com/swathipil/azure-rest-api-specs/issues/27

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,591 @@
...
+    public string Comment { get; set; } = string.Empty;
+}
+
+/// <summary>
+/// Helper interface for APIView feedback customizations operations
+/// </summary>
+public interface IAPIViewFeedbackHelper
+{
+    Task<List<ConsolidatedComment>> GetConsolidatedComments(string apiViewUrl);
+    Task<ReviewMetadata> GetMetadata(string apiViewUrl);
```
praveenkuttappan: This looks like an incorrect place to put this Get Metadata. This function should be in apiview service class. Helper class is to provide a helper function and interaction with external service should be in service class.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,591 @@
...
+    Task<(string? commitSha, string? tspProjectPath, string? targetRepo)> DetectShaAndTspPath(ReviewMetadata metadata);
+}
+
+/// <summary>
+/// Represents a raw comment from APIView API
+/// </summary>
+internal class APIViewComment
```
praveenkuttappan: This should also be moved to model namespace. I believe there is already a namespace for apiview models.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: As of right now, there doesn't seem to be a need to identify comenter vs. reviewer. I assume when this tool is run, commenters and reviewers will already have agreed about the "next step" that needs to be taken, so it should be clear what the resolution should be.

If we find this isn't the case, we can always update the prompt to add reviewer vs. commenter role.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,591 @@
...
+}}";
+
+        var modelName = Environment.GetEnvironmentVariable("AZURE_OPENAI_MODEL_DEPLOYMENT") 
+            ?? Environment.GetEnvironmentVariable("OPENAI_MODEL") 
+            ?? "gpt-4o";
+
+        var chatClient = _openAIClient.GetChatClient(modelName);
+        var messages = new[] { new UserChatMessage(prompt) };
+        var response = await chatClient.CompleteChatAsync(messages);
+        var result = response.Value.Content[0].Text;
```
praveenkuttappan: We should add error handling and exception handling here.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,591 @@
...
+Respond in JSON format:
+{{
+  ""comment"": ""direct actionable statement""
+}}";
+
+        var modelName = Environment.GetEnvironmentVariable("AZURE_OPENAI_MODEL_DEPLOYMENT") 
+            ?? Environment.GetEnvironmentVariable("OPENAI_MODEL") 
+            ?? "gpt-4o";
+
+        var chatClient = _openAIClient.GetChatClient(modelName);
```
praveenkuttappan: How do we handle the scenario when openAIClient is not configured? 
FYI @chrisradek This is a good candidate to use Copilot SDK.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,591 @@
...
+        var response = await chatClient.CompleteChatAsync(messages);
+        var result = response.Value.Content[0].Text;
+
+        // Strip markdown code blocks if present (OpenAI sometimes wraps JSON in ```json ... ```)
+        var jsonText = result.Trim();
+        if (jsonText.StartsWith("```"))
+        {
+            // Remove markdown code blocks
+            var lines = jsonText.Split('\n');
+            jsonText = string.Join('\n', lines.Skip(1).Take(lines.Length - 2));
```
praveenkuttappan: There is an assumption of line length. Is it better to simply replace ````json or use a REGEX to parse the json from string?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,591 @@
...
+            Comment = jsonResponse?["comment"]?.ToString() ?? string.Empty
+        };
+    }
+
+    /// <summary>
+    /// Gets complete metadata (language, packageName, PR info, etc.) for an APIView URL
+    /// </summary>
+    public async Task<ReviewMetadata> GetMetadata(string apiViewUrl)
+    {
+        var (revisionId, reviewId) = ExtractIdsFromUrl(apiViewUrl);
```
praveenkuttappan: reviewId is not used. Is it required to be sent in request? If not, do you need to parse it?

swathipil: no, but it is returned by default from the ExtractIdsFromUrl method in the APIViewService. This will be moved over there as well.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: Still keeping this here for deserializing into a ReviewMetadata object, but the http call is moved over the the APIViewService in its own GetMetadata method and aligns with the return value type in the other methods there.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/APIViewFeedbackService.cs
```diff
@@ -0,0 +1,591 @@
...
+            throw new ArgumentException($"Could not extract IDs from URL: {apiViewUrl}");
+        }
+        
+        return (revisionId, reviewId);
+    }
+
+    /// <summary>
+    /// Detects commit SHA and TypeSpec project path from review metadata
+    /// </summary>
+    public async Task<(string? commitSha, string? tspProjectPath, string? targetRepo)> DetectShaAndTspPath(ReviewMetadata metadata)
```
praveenkuttappan, line 366: This method relies on info about PR and branch. Can we discuss to understand it better?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: It should always be valid b/c we're using the OpenAIClient from ServiceRegistrations.cs, which uses the `https://openai-shared.openai.azure.com/` endpoint if one is not already set.

swathipil: If there is an error, it'll be caught in DelegateAPIViewFeedbackTool.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: replacing with regex

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/IFeedbackItem.cs
```diff
@@ -0,0 +1,123 @@
...
+    
+    /// <summary>
+    /// Individual feedback items for per-item classification
+    /// </summary>
+    public List<FeedbackItem> FeedbackItems { get; set; } = new();
+    
+    /// <summary>
+    /// Target SDK language (e.g., python, csharp, java)
+    /// </summary>
+    public string? Language { get; set; }
```
praveenkuttappan: Can you use SDKLanguage enum here? There is already an enum that's used everywhere else in the cli tools.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Commands/SharedCommandGroups.cs
```diff
@@ -46,7 +46,6 @@ public static class SharedCommandGroups
             Verb: "samples",
             Description: "Sample operations for SDK packages"
         );
-
```
AlitzelMendez, line 49: nit: let's keep this line

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        try
+        {
+            _logger.LogInformation("Fetching APIView feedback from {Url}", apiViewUrl);
+
+            // Get consolidated comments and metadata
+            var comments = await _helper.GetConsolidatedComments(apiViewUrl!);
+            var metadata = await _helper.GetMetadata(apiViewUrl!);
+
+            if (comments == null || comments.Count == 0)
```
AlitzelMendez: this can be simplified to `comments.Count == 0`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
```diff
@@ -0,0 +1,230 @@
...
+using Azure.Sdk.Tools.Cli.Tools.Core;
+
+namespace Azure.Sdk.Tools.Cli.Tools.TypeSpec;
+
+[Description("Delegate APIView feedback to GitHub Copilot coding agent for TypeSpec client customizations")]
+public class DelegateAPIViewFeedbackTool : MCPTool
+{
+    private const string ToolName = "azsdk_tsp_delegate_apiview_feedback";
+    
+    private readonly IAPIViewFeedbackHelper _helper;
```
AlitzelMendez: I don't think helper is an ideal name considering everything is done by the class, I think it should be renamed to `IAPIViewFeedbackService` or something along those lines

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
```diff
@@ -0,0 +1,230 @@
...
+        var apiViewUrl = parseResult.GetValue(_apiViewUrlArg);
+        var repoOverride = parseResult.GetValue(_repoOption);
+        var dryRun = parseResult.GetValue(_dryRunOption);
+
+        try
+        {
+            _logger.LogInformation("Fetching APIView feedback from {Url}", apiViewUrl);
+
+            // Get consolidated comments and metadata
+            var comments = await _helper.GetConsolidatedComments(apiViewUrl!);
```
AlitzelMendez: for what I see `apiViewUrl` is essential for any of this to work, so I think that instead of doing `apiViewUrl!` we should check for it 

```csharp 
var apiViewUrl = parseResult.GetValue(_apiViewUrlArg);
if (string.IsNullOrEmpty(apiViewUrl))
{
    return new DefaultCommandResponse { Message = "APIView URL is required" };
}
```

and remove the use of `!`

AlitzelMendez: I think that even as it is right now if the value is null this will crash anyway, so let's do the check before going forward

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
```diff
@@ -0,0 +1,230 @@
...
+            var metadata = await _helper.GetMetadata(apiViewUrl!);
```
AlitzelMendez: any reason to pass the url instead of the revisionId? like in this class to the 
```csharp
        var (revisionId, reviewId) = APIViewReviewTool.ExtractIdsFromUrl(apiViewUrl);
```
and then have
```
   public async Task<ReviewMetadata> GetMetadata(string revisionId)
```

I think this would setup the environment if later on you want to expand accepting other inputs, but just curious

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
```diff
@@ -0,0 +1,230 @@
...
+            _logger.LogInformation("Found {Count} actionable comment(s) for {Package} ({Language})",
+                comments.Count, metadata.PackageName, metadata.Language);
+
+            // Detect commit SHA, TypeSpec project path, and target repo
+            _logger.LogInformation("Detecting commit SHA and TypeSpec project path");
+            var (commitSha, tspProjectPath, detectedRepo) = await _helper.DetectShaAndTspPath(metadata);
+
+            // Use override repo if specified, otherwise use detected repo, fallback to default
+            var targetRepo = !string.IsNullOrEmpty(repoOverride) ? repoOverride : detectedRepo ?? "Azure/azure-rest-api-specs";
+            var repoParts = targetRepo.Split('/');
```
AlitzelMendez: nit, you can do a tuple deconstruction here `var (owner, repoName) = (targetRepo.Split('/')[0], targetRepo.Split('/')[1]);`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
```diff
@@ -0,0 +1,230 @@
...
+            _logger.LogError(ex, "Failed to address APIView feedback");
+            return new DefaultCommandResponse
+            {
+                Message = $"Error: {ex.Message}"
+            };
+        }
+    }
+
+    [Description("Delegate APIView feedback to GitHub Copilot coding agent for TypeSpec client customizations")]
+    public async Task<DefaultCommandResponse> DelegateAPIViewFeedbackAsync(
```
AlitzelMendez, line 72: can you refactor and extract shared logic? this is almost a duplication of a previous one, if we want to edit or anything always be needing a modification in both places 

more like

```csharp 
 public override async Task<CommandResponse> HandleCommand(ParseResult parseResult, CancellationToken ct)
    {
        var apiViewUrl = parseResult.GetValue(_apiViewUrlArg);
        var repoOverride = parseResult.GetValue(_repoOption);
        var dryRun = parseResult.GetValue(_dryRunOption);

        return await DelegateAPIViewFeedbackCoreAsync(apiViewUrl, repoOverride, dryRun, ct);
    }

    [Description("Delegate APIView feedback to GitHub Copilot coding agent for TypeSpec client customizations")]
    public async Task<DefaultCommandResponse> DelegateAPIViewFeedbackAsync(
        string apiViewUrl,
        string? repo = null,
        bool dryRun = false,
        CancellationToken ct = default)
    {
        return await DelegateAPIViewFeedbackCoreAsync(apiViewUrl, repo, dryRun, ct);
    }
```

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Prompts/Templates/APIViewFeedbackIssueTemplate.cs
```diff
@@ -0,0 +1,168 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text;
```
AlitzelMendez, line 4: nit: remove the ones that are not being used

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,540 @@
...
+    Task<(string? commitSha, string? tspProjectPath, string? targetRepo)> DetectShaAndTspPath(ReviewMetadata metadata);
+}
+
+/// <summary>
+/// Helper class for APIView feedback customizations operations
+/// </summary>
+public class APIViewFeedbackHelper : IAPIViewFeedbackHelper
+{
+    private readonly IAPIViewService _apiViewService;
+    private readonly IAPIViewHttpService _apiViewHttpService;
```
AlitzelMendez: nit: not used

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,540 @@
...
+        _apiViewHttpService = apiViewHttpService;
+        _openAIClient = openAIClient;
+        _gitHubService = gitHubService;
+        _logger = logger;
+    }
+
+    /// <summary>
+    /// Gets consolidated comments from an APIView URL by filtering, grouping, and consolidating comments
+    /// </summary>
+    public async Task<List<ConsolidatedComment>> GetConsolidatedComments(string apiViewUrl)
```
AlitzelMendez: I think ideally you should only receive the revisionId, this is for a bigger vision that I have on apiview, but also if you need the review for something else this is fine

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
```diff
@@ -0,0 +1,540 @@
...
+        {
+            comments = JsonSerializer.Deserialize<List<APIViewComment>>(commentsJson);
+        }
+        catch (JsonException ex)
+        {
+            _logger.LogError(ex, "Failed to deserialize comments JSON for revision {RevisionId}", revisionId);
+            throw new InvalidOperationException($"Failed to parse APIView response for revision {revisionId}", ex);
+        }
+        
+        if (!comments.Any())
```
AlitzelMendez: nit: usually is better to use count like `comments?.Count == 0`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
swathipil: I was thinking about the most common user scenario, and I think people would find it most convenient to copy the apiview url rather than find the revisionId.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: added

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
swathipil: makes sense! will avoid duplication

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
swathipil: kept all code in just the Delegate method and had HandleCommand call that!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
tjprescott: I mean, whether they will have "agreed" on the path forward is very speculative. If you run this comment midway through a review, that's almost certain not to be the case. But that's more a question of process than engineering I think.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
praveenkuttappan: @AlitzelMendez Can APIView take commit SHA ID as a param in metadata for PR API reviews and return it in response. I would like not to use RevisionLabel to pass commit sha when metadata is returned. This will break if pipeline started pushing different info as label.
@swathipil Metadata can contain a branch name and commit SHA. Another option is to ask coding agent to use GH cli to checkout the pr. `gh pr checkout <pr number>` unless you have some tasks in the MCP tool.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            return new DefaultCommandResponse { Message = "APIView URL is required" };
+        }
+
+        try
+        {
+            _logger.LogInformation("Fetching APIView feedback from {Url}", apiViewUrl);
+
+            // Extract revision ID from URL
+            var (revisionId, reviewId) = APIViewReviewTool.ExtractIdsFromUrl(apiViewUrl);
```
praveenkuttappan, line 88: What happens if user provides an invalid URL? revision ID and review ID will be null in that case. right? Can you return a clear error message to provide valid APIView url in that case?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: rename ParseReviewMetadata

praveenkuttappan: Rename method to parse apireview metdata.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/APIViewFeedbackService.cs
swathipil: OOB: Split up the getTSPLocation/SHA method into the 3 separate scenarios. And add more comments to clarify about assumptions. Leave a TODO for the RevisionLabel logic that it should be updated to directly call the Branch once that has been added to the metadata.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: Discussed w/ Praveen: Splitting up the code and adding more comments here to provide more clarity.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/TypeSpec/DelegateAPIViewFeedbackTool.cs
swathipil: The ExtractIdsFromUrl method will throw an "Invalid APIView URL" error that will be logged and returned to the user in the response.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/APIViewFeedbackService.cs
swathipil: updated!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: updated!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/APIView/APIViewService.cs
```diff
@@ -44,4 +46,30 @@ public APIViewService(
...
+
+        if (result == null)
+        {
+            _logger.LogWarning("No metadata found for revision {RevisionId}", revisionId);
+        }
+
+        return result;
+    }
+
+    public async Task<string?> Resolve(string reviewId, string revisionId)
```
AlitzelMendez: if we are adding resolve in here, I think it should look close to https://github.com/Azure/azure-sdk-tools/blob/9945eba74aeb1d8a819b2d5830baf3c308a375df/src/dotnet/APIView/APIViewWeb/LeanControllers/ReviewsTokenAuthController.cs#L65 as that is the intention of resolve, being able to pass 
    /// - Package name + language (+ optional version)
    /// - APIView URL (review or revision link)

so let's at least have `Resolve(string link)` because having review and revisionId means there was some processing already also we are hardcoding `apiview.dev`

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
AlitzelMendez: I returned the sourceBranch on the metadata (still not in prod) did this calculation of revisionLabel and added another parameter on the endpoint to actually have the source branch, so ideally let's use the sourceBranch returned on metadata

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/APIView/APIViewService.cs
swathipil: updated!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/APIView/APIViewService.cs
```diff
@@ -44,4 +49,40 @@ public APIViewService(
...
+
+        if (result == null)
+        {
+            _logger.LogWarning("Failed to resolve URL {Url}", url);
+            return null;
+        }
+
+        try
+        {
+            var resolveResponse = JsonSerializer.Deserialize<ResolvePackageResponse>(result);
```
AlitzelMendez: Do we need to serialize here?

At this point, I’d prefer not to serialize at the response point. We’re making frequent changes, and every time we add something to the response, we’d need to update this to take advantage of it. That would be fine on its own, but since this spans two different projects, it increases the time it takes for everything to reach production—first with the APIView release, then update this code and then when this gets released.

AlitzelMendez: also no big deal, but I prefer if you serialize in where you are getting the response

swathipil: updated!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: @tjprescott - Added this to an issue for follow-up tasks for this tool: https://github.com/Azure/azure-sdk-tools/issues/13910

Along with a task to investigate using `delegate` to directly create PRs and removing the issue creation step.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/APIViewFeedbackService.cs
```diff
@@ -0,0 +1,672 @@
...
+                return await GetTspShaFromSpecsRepoPR(prOwner, prRepoName, prNumber, prRepo);
+            }
+
+            // Scenario 2: If PR info is provided and PR is in language repo, tsp-location.yaml should be parsed to get specs repo and SHA
+            return await GetTspShaFromLanguageRepoPR(prOwner, prRepoName, prNumber, prRepo, metadata.PackageName, metadata.Language);
+        }
+
+        // Scenario 3: If PR info is not provided, then APIView was generated from source branch in the language SDK repo and tsp-location.yaml should be parsed from there
+        // TODO: RevisionLabel currently provides SourceBranch info. Should be updated to use SourceBranch field once it's available: https://github.com/Azure/azure-sdk-tools/issues/13661
+        if (!string.IsNullOrEmpty(revision.RevisionLabel))
```
praveenkuttappan, line 407: Shouldn't we use `main` as default branch is branch info is not available for review created from language repo? Or do you expect  `main` as value from apiview in default case?

praveenkuttappan (approved):
Looks fine to me to test this feature. There are few changes to be done in areas where APIView metadata is processed but it's all TODO in separate PR.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/APIViewFeedbackService.cs
swathipil: Yes, main should be passed as default value from apiview if the PR info is not available.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/APIViewFeedbackHelper.cs
swathipil: @AlitzelMendez - Added this to the follow-up issue!

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Prompts/Templates/APIViewFeedbackIssueTemplate.cs
```diff
@@ -0,0 +1,165 @@
...
+    {
+        _packageName = packageName;
+        _language = language;
+        _apiViewUrl = apiViewUrl;
+        _comments = comments;
+        _commitSha = commitSha;
+        _tspProjectPath = tspProjectPath;
+    }
+
+    private static string GetCodeCustomizationDocUrl(string language) => language.ToLowerInvariant() switch
```
maririos, line 50: We should move this to a config or getting the data from a source instead of hardcoding it. If not already, could you open a GH issue for this?

maririos: or add aka.ms links

PR #13869, opened 2026-02-04
[AVC] Update Metrics

### LINKED ISSUE #13847: [AVC] Update "neutral" meaning in metrics
Today, we treat neutral as "undefined" and thus our metrics are skewed by the large number of neutral comments vs. the small amount of up or down-voted comments.

Instead, we want to pilot a reporting change. There will be five  buckets:
1. Good (upvoted)
2. Bad (downvoted or deleted)
3. Neutral but Resolved (treat as good)
4. Neutral and Unresolved and Unapproved (continue to treat as undefined)
5. Neutral but Unresolved and Approved (treat as bad)

The rationale for 3 is that, regardless of whether the resolution was a change or an explanation, the neutral comment sparked useful investigation. The rationale for 4 is that we can't infer anything--the architect may not have even seen the comment! The rationale for 5 is that if the architect approved the revision with an unresolved copilot comment, they must not have care about it being addressed, so we infer "bad".

tjprescott:
Closes #13847.

We now track with more granularity comment quality, inferrring "good" and "bad" from neutral comments based on other signals. Add the ability to exclude languages from the report and generate PNG charts from the command. The definition of "active review" has also been updated.


packages/python-packages/apiview-copilot/scratch/charts/adoption.png
AlitzelMendez, line 1: do we need to include the images on the pr? or can we add them to `.gitignore`?

packages/python-packages/apiview-copilot/src/_apiview.py
```diff
@@ -9,6 +9,46 @@
 from src._credential import get_credential
 from src._utils import get_language_pretty_name, to_iso8601
 
+
+def get_version_type(version: Optional[str]) -> str:
```
AlitzelMendez, line 14: I think there is a gap for 0.x.x versions, as those are not GA

AlitzelMendez:
can you share how the result/metric looks after running the tool?
is this tool to be integrated with power bi? or is a different effort?

packages/python-packages/apiview-copilot/scratch/charts/adoption.png
tjprescott: Ah, no they should be in gitignore!

tjprescott:
@AlitzelMendez  here's an example of a created chart:
image

I will work on wiring these up to PowerBI, but the chart option allows me to easily make a chart on demand in the meantime. The schema was made to remain consistent with what we have.

packages/python-packages/apiview-copilot/scratch/charts/adoption.png
AlitzelMendez: I see that you added the gitignore but probably as this was pushed before that the files are still in this pull request :)

AlitzelMendez (approved):

PR #13879, opened 2026-02-05
Exclude serialization methods from API view

srnagar:
This pull request updates the logic in the `JavaASTAnalyser` to exclude certain serialization-related methods from the public API surface. Specifically, it ensures that `fromJson` and `toJson` methods in classes implementing `JsonSerializable` are not treated as part of the public API, as they are considered implementation details.

### Public API filtering improvements

* Updated the method filtering logic in `visitConstructorsOrMethods` to skip `fromJson` and `toJson` methods for types that implement `JsonSerializable`, preventing these internal serialization methods from being included in the public API contract.
* Added a new helper method `isFromOrToJsonMethod` to encapsulate the logic for identifying `fromJson` and `toJson` methods in `JsonSerializable` implementations.

image

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -91,7 +91,7 @@
 /tools/apiview/parsers/csharp-api-parser/ @azure/azsdknet @jsquire
 /tools/apiview/parsers/js-api-parser/     @jeremymeng @maorleger
 /src/swift/                               @tjprescott
-/src/java/apiview-java-processor/         @JonathanGiles
+/src/java/apiview-java-processor/         @JonathanGiles @alzimmermsft @srnagar
 /src/go/                                  @jhendrixMSFT @chlowell @RickWinter
 
 ###########


src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -1060,6 +1063,24 @@ private void visitConstructorsOrMethods(final TypeDeclaration<?> typeDeclaration
             });
     }
 
+    private boolean isFromOrToJsonMethod(TypeDeclaration<?> typeDeclaration, CallableDeclaration<?> callableDeclaration) {
+        if (typeDeclaration instanceof ClassOrInterfaceDeclaration) {
+            ClassOrInterfaceDeclaration classOrInterfaceDeclaration = (ClassOrInterfaceDeclaration) typeDeclaration;
+            return classOrInterfaceDeclaration.getImplementedTypes() != null && classOrInterfaceDeclaration.getImplementedTypes()
+                    .stream()
+                    .anyMatch(implementedType -> {
+                        if (implementedType.getNameAsString().equals("JsonSerializable")) {
```
alzimmermsft: Should add the same for `XmlSerializable` with `toXml` and `fromXml`.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -1060,6 +1063,24 @@ private void visitConstructorsOrMethods(final TypeDeclaration<?> typeDeclaration
...
+                            if (callableDeclaration.getNameAsString().equals("toJson") ||
```
alzimmermsft: Should add an error message if the type implements `JsonSerializable` (and when added `XmlSerializable`) and doesn't have the to and from methods publicly available.

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
```diff
@@ -1060,6 +1079,42 @@ private void visitConstructorsOrMethods(final TypeDeclaration<?> typeDeclaration
...
+     * <li>The class implements JsonSerializable and method name is toJson or fromJson</li>
+     * <li>The class implements XmlSerializable and method name is toXml or fromXml</li>
+     * </ul>
+     *
+     * @param typeDeclaration     The type declaration to check interface implementations
+     * @param callableDeclaration The method declaration to check the name
+     * @return true if the method is a serialization method, false otherwise
+     */
+    private boolean isSerializationMethod(TypeDeclaration<?> typeDeclaration, CallableDeclaration<?> callableDeclaration) {
+        if (typeDeclaration instanceof ClassOrInterfaceDeclaration) {
```
alzimmermsft: Should early out if the CallableDeclaration fails

`callableDeclaration.isMethodDeclaration()`

Side nit: I prefer the `isX` and `asX` methods in JavaParser as they're a bit cleaner

srnagar: This part of the code only gets executed if it's a method declaration

src/java/apiview-java-processor/src/main/java/com/azure/tools/apiview/processor/analysers/JavaASTAnalyser.java
srnagar: As discussed offline, we'll make this a checkstyle rule instead - https://github.com/Azure/azure-sdk-for-java/issues/47915

Issue #10319, opened 2025-04-15
Comment notification emails do not link to the comment


christothes:
E-mail notifications about comments contain a link to the comment, but they do not resolve to the correct view context.

Example:
---------------
DominikMe commented on this review.

In [Azure.Communication.PhoneNumbers.PhoneNumbersBrowseResult](https://spa.apiview.dev/review/e465f73b881e41a5875a9d614a96d3ad?activeApiRevisionId=fdf629cdb3a4425abc6110641e6db30b#Azure.Communication.PhoneNumbers.PhoneNumbersBrowseResult):

Is the API pageable? Then we should return a Pageable type and not a class with a list in it.
--------------

The link should be this:

https://spa.apiview.dev/review/e465f73b881e41a5875a9d614a96d3ad?activeApiRevisionId=061a45ba52e94961b49727067254f90a&nId=Azure.Communication.PhoneNumbers.PhoneNumbersBrowseRequest.PhoneNumberType

PR #13887, opened 2026-02-06
[APIView] Fix comment notification emails not resolving to comment

smw-ms:
Resolves Issue: https://github.com/Azure/azure-sdk-tools/issues/10319

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
+++ b/src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
@@ -36,6 +36,7 @@ namespace APIViewWeb.Managers
         private readonly IEmailTemplateService _emailTemplateService;
         private readonly IHttpClientFactory _httpClientFactory;
         private readonly ILogger<NotificationManager> _logger;
+        private readonly IEnumerable<LanguageService> _languageServices;
 
         private const string ENDPOINT_SETTING = "APIVIew-Host-Url";
 
@@ -46,7 +47,8 @@ namespace APIViewWeb.Managers
             TelemetryClient telemetryClient,
             IEmailTemplateService emailTemplateService,
             IHttpClientFactory httpClientFactory,
-            ILogger<NotificationManager> logger)
+            ILogger<NotificationManager> logger,
+            IEnumerable<LanguageService> languageServices)
         {
             _configuration = configuration;
             _apiviewEndpoint = configuration.GetValue<string>(ENDPOINT_SETTING);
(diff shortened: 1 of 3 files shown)


src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -182,13 +185,19 @@ private string GetApproverReviewHtmlContent<T>(UserProfileModel user, T model) w
...
+                configuration: _configuration,
+                languageServices: _languageServices,
+                elementId: comment.ElementId);
             var commentText = comment.CommentText;
             var poster = comment.CreatedBy;
             var userLink = new Uri($"{_apiviewEndpoint}/Assemblies/Profile/{poster}");
             var sb = new StringBuilder();
             sb.Append($"<a href='{userLink.ToString()}'>{poster}</a>");
-            sb.Append($" mentioned you in <a href='{reviewLink.ToString()}'><b>{reviewName}</b></a>");
+            sb.Append($" mentioned you in <a href='{reviewLink}'><b>{reviewName}</b></a>");
```
tjprescott, line 200: My understanding is that we don't have mention functionality, so is the language here correct?

smw-ms:
> One comment. Also, can you provide a screenshot of what the new email looks like?
image

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
tjprescott: Interesting. @smw-ms your screenshot doesn't match this code. Is this dead code?

smw-ms: I wouldn't expect it to since I didn't tag/mention anyone in my comment. From what I seen in the code the tag/mention functionality is basically already setup. just need to add some code that parses the comment for any tagged users.

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
smw-ms, line 209: @tjprescott this is the method that was used in my test email

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
tjprescott: No no, we don't want to re-enable that (at least, not yet).  My question is, there is where the diff is, so I don't see how a change _here_ would be producing a change in the email if that method your email called into didn't change.

tjprescott: Ah nevermind I see. The fix is in ResolveReviewUrl and both of these methods were updated to use that.

tjprescott (approved):

PR #13896, opened 2026-02-06
Enable VerifySetup to auto-install supported missing requirements

### LINKED ISSUE #13897: Implement auto-install for verifysetup

JennyPng:
closes https://github.com/Azure/azure-sdk-tools/issues/13897

- enables auto-install
- in CLI mode, if `install` is passed, will run interactive mode and get permission for each installable requirement
- in agent mode, agent must pass in list of exact failed requirement names that the user agrees to install
- also enables requirements to specify one layer of dependency, so that if a prerequisite already failed, the check/auto-install will immediately return failure instead of attempting.

## core changes 

Updates Requirement model to have
- `isAutoInstallable` bool - defaults false 
- `NotAutoInstallableReason` - can be set to explain to the user why a requirement is not auto-installable, and will be returned by the tool for a req if `isAutoInstallable` is false
- `GetInstallCommands` - returns executable install commands, should be overridden when a requirement is auto-installable
- `GetInstructions` - this returns purely readable instructions for when we're not auto-installing - it should use `GetInstallCommands` for requirements that are auto-installable
- `RunInstallAsync` - executes install commands
- `RunCheckAsync` - runs command that checks if a requirement is present
- `DependsOn` - list of requirement names that are dependencies

Updates RequirementContext with
- `FailedRequirements` - used to check per requirement if a dependency failed already

Updates VerifySetupTool to
- MCP: accept a `requirementsToInstall` list parameter - agent mode expects list of exact reqs requested to be installed
- CLI: accepts subcommand `install` for interactive install mode, can pass in -y to allow all installs w/o prompting
- try to install missing requirements - after a successful install, will recheck it to verify that it was really installed

Python Requirement Handling
- resolves a venv before installing - uses venv specified by env var if set, falls back on looking for `.venv` in the repo root, and finally falls back on creating `.venv` at the repo root
- CONCERN: several tools are installed via `pip install eng/tools/azure-sdk-tools` which must be run from the root of the Python repo, so the auto-install here only works if the detected reporoot is python - not sure if there's a way around this

## auto-installable reqs:
- tsp
- tsp-client
- git longpaths
- pnpm
- goimports
- go generator
- all python specific reqs except python itself (with venv handling)

## reqs we will not auto-install:
everything else - languages, runtimes, tools bundled with languages, system-level installs

### Agent Mode (with claude sonnet 4)

image

image

#### NO --auto-install
image

#### WITH JUST --auto-install 
image

image


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/CoreRequirements.cs
```diff
@@ -42,66 +43,80 @@ public class TspClientRequirement : Requirement
...
         public override string? MinVersion => "0.24.0";
+        public override bool IsAutoInstallable => true;
+
+        public override string[][]? GetInstallCommands(RequirementContext ctx)
+        {
+            var workingDir = ctx.IsSpecsRepo()
+                ? ctx.RepoRoot
+                : Path.Combine(ctx.RepoRoot, "eng", "common", "tsp-client");
+
+            return [["npm", "ci", "--prefix", workingDir]];
```
l0lawrence, line 55: so node would need to be installed first -- but the user has to install it since its system level? So if they wanted to autoinstall they would get an error about missing npm, install node, refresh copilot, and say auto-install?

l0lawrence: I guess this goes along with your first open question :)

JennyPng: yes currently that is correct - so it might make sense for this to be smart and not run if it knows the Node check failed - I will just try implementing this

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/JavaScriptRequirements.cs
```diff
@@ -18,6 +18,11 @@ public class PnpmRequirement : Requirement
     {
         public override string Name => "pnpm";
         public override string[] CheckCommand => ["pnpm", "--version"];
+        public override IReadOnlyList<string> DependsOn => ["Node.js"];
+        public override bool IsAutoInstallable => true;
+
+        public override string[][]? GetInstallCommands(RequirementContext ctx)
+            => [["npm", "install", "-g", "pnpm"]];
```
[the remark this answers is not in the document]
JennyPng: Uncertain about this case 

Currently, the auto-install commands for `pnpm` only install pnpm, but the manual pnpm instructions originally were:
```
"npm install -g pnpm",
"Navigate to the root of the JavaScript SDK repository.",
"pnpm install"
```

it would be difficult to represent the additional instructions in the auto-install commands, since being able to auto-execute the `pnpm install` requires navigating to the specific JS repo or already running the mcp tool from the JS repo. 

One possible solution is to add a `PostInstallInstructions` field that informs of manual action required, but not sure if this is the best approach

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/PythonRequirements.cs
```diff
@@ -34,6 +53,122 @@ public override string[] CheckCommand
...
+        var repoVenvPath = Path.Combine(ctx.RepoRoot, ".venv");
+        var repoVenvPythonExe = Path.Combine(repoVenvPath, binDir, exeName);
+        if (Directory.Exists(repoVenvPath) && File.Exists(repoVenvPythonExe))
+        {
+            return (repoVenvPythonExe, null);
+        }
+
+        // 3. Create a new .venv at repo root
+        var createResult = await RunCommandAsync(
+            processHelper, ["python", "-m", "venv", repoVenvPath], ctx, ct, InstallTimeout);
```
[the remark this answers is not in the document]
JennyPng: out of scope / we don't support python2

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/VerifySetupResponse.cs
```diff
@@ -23,13 +23,38 @@ protected override string Format()
...
+                {
+                    sb.AppendLine($"        - Auto-install failed: {result.AutoInstallError}");
+                    sb.AppendLine($"        - Instructions: {string.Join(", ", result.Instructions)}");
+                }
+                else
+                {
+                    sb.AppendLine($"        - Instructions: {string.Join(", ", result.Instructions)}");
+                    if (result.IsAutoInstallable)
+                    {
+                        sb.AppendLine($"        - Tip: Re-run with --auto-install to install this automatically");
```
benbp: This should probably be captured into the `NextSteps` property as well so that JSON mode shows the suggestion (e.g. if we end up directing CLI agents to use json mode and/or for MCP responses)

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/Requirement.cs
```diff
@@ -72,14 +124,114 @@ public virtual async Task<RequirementCheckOutput> RunCheckAsync(
...
+    /// Runs the auto-installation for this requirement. Override for custom install logic
+    /// (e.g., venv handling, working-directory changes).
+    /// The default implementation executes each command from <see cref="GetInstallCommands"/> sequentially,
+    /// short-circuiting on the first failure.
+    /// </summary>
+    /// <param name="processHelper">Process execution helper.</param>
+    /// <param name="ctx">The current environment context.</param>
+    /// <param name="ct">Cancellation token.</param>
+    /// <returns>The result of the install operation.</returns>
+    public virtual async Task<RequirementCheckOutput> RunInstallAsync(
```
benbp: I think we had a longer discussion about where this should go last time and settled on here. With all this logic being added I would say this does not belong in `Models` anymore. Perhaps `Services/Languages/SetupRequirements/Requirement.cs`?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/VerifySetupResponse.cs
JennyPng: addressed 👍

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/Requirement.cs
JennyPng: addressed

eng/common/instructions/azsdk-tools/verify-setup.instructions.md
```diff
@@ -3,19 +3,26 @@ description: 'Verify Setup'
...
-Your goal is to identify the project repo root, and pass in the `packagePath` to the Verify Setup tool. For a language repo, pass in the language of the repo. 
+Your goal is to identify the project repo root, and pass in the `packagePath` to the Verify Setup tool. For a language repo, pass in the language of the repo.
+
+## Instructions
+1. Check what's missing by calling verify setup with just language and package path parameters. The tool will respond with missing requirements, and if any are installable by the tool itself.
+2. Ask the user if they want help installing any missing requirements that the tool can install. On approval, **use the tool again WITH THE EXACT PARAMETERS (langs=<language(s)>, packagePath=<path>, requirementsToInstall=["req", "req2", ...])**
 
 ## Examples
 - in `azure-sdk-for-js`, run `azsdk_verify_setup` with `(langs=javascript, packagePath=<path>/azure-sdk-for-js)`.
+- to install, run `azsdk_verify_setup` with `(langs=javascript, packagePath=<path>/azure-sdk-for-js), requirementsToInstall=['pnpm', 'tsp'])`.
```
benbp, line 16: I was thinking we should always refer to both the MCP tool name and the cli command so both mcp clients and cli agents can use the instructions files? Not sure what the best language is to help the client understand which mode it is operating in (or maybe it knows).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/SetupRequirements/PythonRequirements.cs
```diff
@@ -11,29 +11,227 @@ namespace Azure.Sdk.Tools.Cli.Services.SetupRequirements;
...
+        return (repoVenvPythonExe, null);
+    }
+
+    /// <summary>
+    /// Handles venv resolution and pip install for Python tool requirements.
+    /// 1. Resolves the venv Python executable (env var → existing .venv → create new).
+    /// 2. Resolves the pip install target (validating azure-sdk-for-python repo for relative paths).
+    /// 3. Runs pip install for the requirement's PipInstallTarget.
+    /// </summary>
+    public override async Task<RequirementCheckOutput> RunInstallAsync(
```
benbp: nit: we don't need the `Async` naming in suffixes

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -57,22 +56,123 @@ public VerifySetupTool(IProcessHelper processHelper, ILogger<VerifySetupTool> lo
         AllowMultipleArgumentsPerToken = true
     };
 
+    private readonly Option<List<string>> autoInstallParam = new("--auto-install", "-i")
```
benbp: The naming of `--auto-install` and `--allow-upgrade` for CLI mode seems a bit confusing to me. Is my understanding correct here?

- If auto-install has values passed to it, we will install the specified tools with no prompting
- If auto-install has no values passed, we will detect missing tools and prompt for installation
- If allow-upgrade is also passed with empty auto-install, we will detect missing tools and install with no prompting

Auto implies to me as a user that it will be automatic, with no prompting. Also having a flag that operates as either an array of values OR a boolean is non-intuitive and I think CLI users and possibly MCP clients will have trouble using it.

Some possible suggestions/alternatives, let me know your thoughts:

**CLI idea 1**
- `azsdk setup install --tools <foo> <bar>` installs specified tools automatically
- `azsdk setup install` Detects missing tools and prompts for install
- `azsdk setup install --yes|-y` Detects and does not prompt, mirrors `apt install --yes` convention

**CLI idea 2**
- `azsdk setup --install-tools <foo> <bar>` installs specified tools automatically
- `azsdk setup --install-missing` Detects missing tools and prompts for install
- `azsdk setup --install-missing --yes|-y` Detects and does not prompt, mirrors `apt install --yes` convention

benbp: I think the MCP mode makes sense since it will always force the agent to pick tools to install.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Models/Responses/CommandResponse.cs
```diff
@@ -68,7 +68,7 @@ public override string ToString()
     {
         var value = Format();
 
-        List<string> messages = [];
+        List<string> messages = string.IsNullOrWhiteSpace(value) ? [] : [value];
```
benbp: This should be removed. We intentionally drop any non-error / next steps properties from the output response when those properties are present. This forces callers to properly set the `ResponseError` property in failure cases vs. just always setting child class properties. In the past this was not adhered to very well, and it can hide issues like `OperationStatus` being incorrect or our telemetry queries not aggregating error responses.

benbp: Ah, I see the issue here and why this was added. If we populate `NextSteps` for a happy path case as I suggested, that swallows the `Format()` output. So maybe the logic needs to be extended a bit to account for that (if errors, don't include `Format()`, otherwise do even if `NextSteps` is populated).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
JennyPng: Yes your understanding is correct

Idea 1 makes sense to me, I'll make that change

eng/common/instructions/azsdk-tools/verify-setup.instructions.md
JennyPng: I wonder if mentioning the CLI in these instructions will bias the agent towards using the CLI tool instead of the MCP tool? a CLI agent could use the -h flag to gain more context on the tool right 🤔 I find the main value of this instruction is to enforce correct parameter usage

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/LanguageMcpMultiCommandTool.cs
```diff
@@ -0,0 +1,38 @@
...
+        protected IGitHelper gitHelper => _languageHelper.GitHelper;
+
+        public LanguageMcpMultiCommandTool(IEnumerable<LanguageService> languageServices, IGitHelper gitHelper, ILogger<LanguageMcpMultiCommandTool> logger)
+        {
+            _languageHelper = new LanguageToolHelper(languageServices, gitHelper);
+            this.logger = logger;
+        }
+
+#pragma warning disable MCP003 // Tool methods must return Response types, built-in value types, or string
+        public Task<LanguageService> GetLanguageServiceAsync(string packagePath, CancellationToken ct = default)
```
benbp: @praveenkuttappan I've been thinking we need to switch this back into a resolver class. It feels like the wrong place to have this logic as part of the tool class inheritance chain. But it will be an annoyingly large refactor so I'm wondering whether to just let it stay as is. But you can see how messy it's starting to get now that we need the logic in other places (and it's also annoying to have to instantiate a tool class in test when you just need to resolve a language service).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/LanguageToolHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+/// </summary>
+internal sealed class LanguageToolHelper(
+    IEnumerable<LanguageService> languageServices,
+    IGitHelper gitHelper)
+{
+    public IEnumerable<LanguageService> LanguageServices { get; } = languageServices;
+    public IGitHelper GitHelper { get; } = gitHelper;
+
+#pragma warning disable MCP003 // Tool methods must return Response types, built-in value types, or string
+    public async Task<LanguageService> GetLanguageServiceAsync(string packagePath, CancellationToken ct = default)
```
benbp: I think trying to be DRY here is making things messier and more complicated than they need to be, as we end up with more code than the original logic we're trying to avoid duplicating. I think we should just copy the logic for now across `LanguageMcpTool` and `LanguageMcpMultiCommandTool` and I will refactor it out into a resolver class later.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Verify/VerifySetupTool.cs
```diff
@@ -57,22 +56,130 @@ public VerifySetupTool(IProcessHelper processHelper, ILogger<VerifySetupTool> lo
...
+        setupCommand.Options.Add(SharedOptions.PackagePath);
+
+        var installCommand = new Command(InstallCommandName, "Install missing environment requirements")
         {
-            languagesParam,
-            SharedOptions.PackagePath
+            toolsParam,
+            yesParam
         };
+        setupCommand.Subcommands.Add(installCommand);
```
benbp: While possible with `System.CommandLine`, doing it this way breaks our pseudo-convention of "one class per command hierarchy level" and causes things like the list command tool to skip for `azsdk verify setup` (https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/Core/ListCommandTool.cs#L38).

I think what needs to be done instead is a `VerifySetupInstallTool.cs` class with its own 

```
    public override CommandGroup[] CommandHierarchy { get; set; } = [
        SharedCommandGroups.Verify,
        SharedCommandGroups.Setup,
    ];
```

Also, kind of outside the scope of this PR, but do we have plans for other `verify` type actions? We're getting pretty deep now into the command tree without any other branches (e.g. no `verify foobar`). I wonder if considering `azsdk setup verify [all|tooling]`, `azsdk setup install` etc. might make more semantic sense?

JennyPng: I don't know if we will have other verify actions 🤔 I think the azsdk setup ... suggestion makes sense, but I won't address it in this PR

benbp (approved):
LGTM, changelog conflict will need updating

benbp:
We had some issues with eng/common syncing/applying a patch due to an accidental merge weeks ago into the cpp repo. I'm going to force merge this then manually merge these sync PRs:

https://github.com/Azure/azure-sdk-for-go/pull/26076
https://github.com/Azure/azure-sdk-for-python/pull/45433
https://github.com/Azure/azure-sdk-for-js/pull/37399
https://github.com/Azure/azure-sdk-for-java/pull/48162
https://github.com/Azure/azure-sdk-for-net/pull/56600
https://github.com/Azure/azure-sdk-for-rust/pull/3830
https://github.com/Azure/azure-sdk-for-cpp/pull/6977
https://github.com/Azure/azure-sdk-for-c/pull/3378
https://github.com/Azure/azure-sdk-for-android/pull/2131
https://github.com/Azure/azure-sdk-for-ios/pull/2366
https://github.com/Azure/azure-sdk/pull/9678
https://github.com/Azure/azure-rest-api-specs/pull/40865

PR #13899, opened 2026-02-06
Lazy initialize AzureSdkKnowledgeBaseService

benbp:
Currently the `AzureSdkKnowledgeBaseService` does all its setup in the constructor, which includes info logging and other operations. As we force instantiate all tool classes on CLI startup in `CommandRunner.cs`, this causes the `TypeSpecAuthoringTool` to load `AzureSdkKnowledgeBaseService`. The effect was that we info logging distracting and irrelevant logs for any command, in addition to spending time doing configuration and authentication.

This PR fixes this issue by moving all config in the service to a lazy `Initialize()` method that only gets called when the service is used. Additionally the http client log level threshold is changed to `Warning` to prevent noisy request logs in typespec authoring requests.

Example problematic output:

```
⇉ ⇉ ⇉ dotnet run -- example hello-world foo
info: Azure.Sdk.Tools.Cli.Services.AzureSdkKnowledgeBaseService[0]
      Azure knowledge base service endpoint has not been configured. You can set the environment variable AZURE_SDK_KB_ENDPOINT.
info: Azure.Sdk.Tools.Cli.Services.AzureSdkKnowledgeBaseService[0]
      Using default endpoint: https://azuresdkqabot-dev-serve-authoring-epgbcvbpa3adcvcu.westus2-01.azurewebsites.net
info: Azure.Sdk.Tools.Cli.Services.AzureSdkKnowledgeBaseService[0]
      Azure knowledge base service endpoint and client id are configured. Initializing authentication.
info: Azure.Sdk.Tools.Cli.Tools.Example.HelloWorldTool[0]
      Echoing message: foobar
RESPONDING TO 'foobar' with SUCCESS: 0
Duration: 1ms

```

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/AzureSdkKnowledgeBaseService.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/AzureSdkKnowledgeBaseService.cs
@@ -15,13 +15,16 @@ namespace Azure.Sdk.Tools.Cli.Services
     /// </summary>
     public class AzureSdkKnowledgeBaseService : IAzureSdkKnowledgeBaseService
     {
+        private IPublicClientApplication? _msalApp;
+
         private readonly HttpClient _httpClient;
         private readonly ILogger<AzureSdkKnowledgeBaseService> _logger;
         private readonly AzureSdkKnowledgeBaseOptions _options;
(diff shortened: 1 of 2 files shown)


benbp:
FYI @chunyu3

praveenkuttappan (approved):

PR #13947, opened 2026-02-11
[azsdk-cli] Add azsdk upgrade command and notify users when version is stale

benbp:
New functionality in this PR:

- `azsdk` checks for new versions proactively and notifies the user when one is found.
  ```
  A new version of azsdk is available (0.5.14 -> 0.5.16)
  Release notes: https://github.com/Azure/azure-sdk-tools/releases/tag/azsdk_0.5.16
  Run 'azsdk upgrade' or invoke the `#azsdk_upgrade` mcp tool to get it
  ```
- Add an `azsdk upgrade` command and `#azsdk_upgrade` mcp tool that performs a two-step, in place upgrade
  - Because of file locking restrictions preventing in-place upgrade of a running process, `upgrade` downloads the new executable to a temp location and spawns a hidden command that waits for the original process to exit. After exit it copies itself to the original location (`/home/bin/azsdk`).
  - In MCP mode we also shut down the server (to remove the process lock) after returning a response to the client saying that a server restart is required.
- Adds an update cache to `~/.azsdk/update-config.json` that tracks the last time we checked for a new version and the last time we notified the user. For CLI mode we don't always want to check for updates as that will add noticeable latency to every call. Similarly we don't want to spam the user on every invocation about a new version. The TTL for a new version check is 1 day and to notify is 3 days.

The core logic is in `UpgradeService.cs` which is called from `UpgradeTool.cs`


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Upgrade/UpgradeService.cs
```diff
@@ -0,0 +1,735 @@
...
+
+    private static readonly JsonSerializerOptions jsonOptions = new()
+    {
+        WriteIndented = true
+    };
+
+    public string GetCurrentVersion()
+    {
+        var assembly = Assembly.GetExecutingAssembly();
+        var version = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion
```
weshaggard, line 57: Do we not already do this somewhere in the tool? Also I hope we would never end up with 0.0.0 here.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Upgrade/UpgradeService.cs
```diff
@@ -0,0 +1,735 @@
...
+    }
+
+    /*
+      Two-step upgrade flow:
+      The running executable cannot be replaced while it's running (file lock).
+
+      To work around this, we use a two-step process:
+        1. The original azsdk executable downloads the new version to a temp directory
+        2. The original azsdk executable spawns the NEW exe with --complete-upgrade <original-path>
+        3. The original azsdk executable exits immediately (releasing the file lock)
```
weshaggard, line 275: Do we have to worry about multiple instances of the exe running and only want of them requested the upgrade but others remain running?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Upgrade/UpgradeService.cs
benbp: We might, I'll check, or at least move out to some sort of constant (possibly compile time generated).

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Upgrade/UpgradeService.cs
```diff
@@ -0,0 +1,735 @@
...
+        }
+        catch
+        {
+            return null;
+        }
+    }
+
+    private string GetCurrentExecutablePath()
+    {
+        var homeDir = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
```
weshaggard, line 559: Is this really the way to get the path? Why can't you get via Asssembly.GetExecutingAssembly()?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Upgrade/UpgradeService.cs
benbp: Yeah in this case we would fail the upgrade silently. I'm not sure how to work around this. Possibly while performing the main `upgrade` command/tool we could inspect other running azsdk processes and if they exist, fail?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Upgrade/UpgradeService.cs
weshaggard: Or Enviroment.ProcessPath?

benbp: It gets annoying if the command is getting tested via `dotnet run`, so for debug mode we just use the default install path. In other cases we inspect the process path below and only default if it's null.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Upgrade/UpgradeService.cs
```diff
@@ -0,0 +1,735 @@
...
+            var pollInterval = TimeSpan.FromMilliseconds(100);
+            var elapsed = TimeSpan.Zero;
+
+            logger.LogInformation("Waiting for target executable to be replaceable: {Path}", targetPath);
+
+            while (elapsed < maxWait)
+            {
+                ct.ThrowIfCancellationRequested();
+
+                if (TryReplaceExecutable(currentExePath, targetPath))
```
weshaggard, line 306: Should we add some kill commands for azsdk process?

weshaggard (approved):
I don't enjoy the complexity but I don't have any real alternatives.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Upgrade/UpgradeService.cs
benbp: Neither case is ideal, but I think it's preferable to fail silently. If someone has a long running task (e.g. tsp generation) going on with another instance of the server/cli it would be bad to interrupt that mid-stream.

benbp:
@weshaggard I agree it's a decent amount of complexity, but on the bright side it's hidden from the user and should _hopefully_ perform in a robust way.

PR #13956, opened 2026-02-11
[APIView] Remove NotifySubscribersOnNewRevisionAsync functionality

smw-ms:
In order to align with new notification system an email notification is no longer sent on new api revision upload


tjprescott:
@smw-ms I think rather than keep this open we should close it and just revisit it when more of the right pieces have fallen into place. It will likely be part of that larger PR.

PR #13961, opened 2026-02-11
[AVC] Java improvements

### LINKED ISSUE #13889: [AVC] Consistently low value Java comments from AVC
Metrics show that comments in Java remain abnormally lower rated compared to other languages that have a high implicit or explicit quality score.

- Resolve all filterable comments
- Review feedback attached to any downvoted comments
- Review any relevant memories to see if they are being processed correctly

tjprescott:
Closes #13889

This PR:

- adds scenarios to the Java filter to screen out known accepted usage patterns
- enhances various AVC commands
- adds `avc db link/unlink` commands

### DIFF
--- a/packages/python-packages/apiview-copilot/tests/resolve_language_test.py
+++ b/packages/python-packages/apiview-copilot/tests/resolve_language_test.py
@@ -1,0 +1,147 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+# pylint: disable=missing-class-docstring,missing-function-docstring
+
+"""
+Tests for resolve_language and resolve_language_to_canonical in cli.py.
+"""
+
+import pytest
+from cli import _LANGUAGE_ALIAS_TABLE, resolve_language, resolve_language_to_canonical
+
+
+@pytest.fixture(autouse=True)
+def _clear_alias_table():
+    """Clear the cached alias table before each test for isolation."""
+    _LANGUAGE_ALIAS_TABLE.clear()
+    yield
+    _LANGUAGE_ALIAS_TABLE.clear()
+
+
+# ---------- Canonical names resolve correctly ----------
+
+
+@pytest.mark.parametrize(
+    "canonical, expected_pretty",
+    [
+        ("android", "Android"),
+        ("clang", "Clang"),
+        ("cpp", "C++"),
+        ("dotnet", "C#"),
+        ("golang", "Go"),
+        ("ios", "Swift"),
+        ("java", "Java"),
+        ("python", "Python"),
+        ("rust", "Rust"),
+        ("typescript", "JavaScript"),
+    ],
+)
+def test_canonical_names(canonical, expected_pretty):
+    assert resolve_language(canonical) == (canonical, expected_pretty)
+
+
+# ---------- Pretty names resolve correctly ----------
+
+
+@pytest.mark.parametrize(
+    "pretty, expected_canonical",
+    [
+        ("Android", "android"),
+        ("Clang", "clang"),
+        ("C++", "cpp"),
+        ("C#", "dotnet"),
+        ("Go", "golang"),
+        ("Swift", "ios"),
+        ("Java", "java"),
+        ("Python", "python"),
+        ("Rust", "rust"),
+        ("JavaScript", "typescript"),
+    ],
+)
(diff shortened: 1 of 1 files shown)


packages/python-packages/apiview-copilot/metadata/java/filter.yaml
```diff
@@ -1,6 +1,9 @@
...
-  3. DO NOT suggest enforcing consistency in naming conventions for enums
-  4. DO NOT comment on enum naming conventions
-  5. DO NOT complain about hardcoding dependency versions in the POM.
\ No newline at end of file
+  3. DO NOT suggest changes to anything annotated with the @Generated decorator.
+  4. DO NOT suggest enforcing consistency in naming conventions for enums
+  5. DO NOT comment on enum naming conventions
+  6. DO NOT complain about hardcoding dependency versions in the POM.
+  7. DO NOT comment on methods or overrides lacking an implementation.
+  8. DO NOT recommend annotating parameters to express nullability expectations.
```
tjprescott: @srnagar please review this file. The rest you probably don't care about 😄

packages/python-packages/apiview-copilot/metadata/java/filter.yaml
```diff
@@ -1,6 +1,9 @@
 exceptions: |
   1. DO NOT make comments that don't actually identify a problem
   2. DO NOT comment on lack of documentation comments
-  3. DO NOT suggest enforcing consistency in naming conventions for enums
-  4. DO NOT comment on enum naming conventions
-  5. DO NOT complain about hardcoding dependency versions in the POM.
\ No newline at end of file
+  3. DO NOT suggest changes to anything annotated with the @Generated decorator.
```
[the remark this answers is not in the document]
srnagar: This is a bit too broad. We still want suggestions on generated methods - for e.g. if there is a casing issue, a typo or too many params in a method, we want those comments on method annotated with `@Generated`. The fix will be either in the main spec or in customization (client.tsp).

packages/python-packages/apiview-copilot/metadata/java/filter.yaml
srnagar: 5 - we expect comments on enum naming conventions. So, we should remove 5. The others look good. Are there any examples that required us to add 5?

tjprescott: I don't recall, but we can remove it.

packages/python-packages/apiview-copilot/cli.py
```diff
@@ -654,10 +730,13 @@ def handle_agent_thread_resolution(comments_path: str, remote: bool = False):
...
     code = data.get("code", None)
-    if language not in SUPPORTED_LANGUAGES:
+    # Normalize language to the lowercase APIView format expected by SUPPORTED_LANGUAGES
+    normalized_language = normalize_language(language).lower() if language else language
+    apiview_language = next((l for l in SUPPORTED_LANGUAGES if l.lower() == normalized_language), None)
+    if not apiview_language:
         print(f"Unsupported language `{language}`")
         return
-    pretty_language = get_language_pretty_name(language)
+    pretty_language = get_language_pretty_name(apiview_language)
```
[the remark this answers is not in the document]
tjprescott: Fixed

packages/python-packages/apiview-copilot/src/_apiview.py
```diff
@@ -16,6 +16,21 @@
 from src._credential import get_credential
 from src._utils import get_language_pretty_name, to_iso8601
 
+# Maps AICommentFeedbackReason enum values to human-readable messages.
+# Mirrors the C# ApiView structure (CommentFeedbackRequest.cs).
+FEEDBACK_REASON_MESSAGES = {
```
AlitzelMendez, line 21: If we are defining the readable message here, do we also need to have it in the C# code?
Can C# just return the enums, or do we need to construct the message in C# so Copilot can act on it?

I’m asking because on the C# side we aren’t using the reason message—we only use it on the client side to render the message in the UI. So if we’re already defining it here, I’m not sure if adding the definition in C# is unnecessary and could create another place we’d need to update in the future.

tjprescott: We do use it. I had this mirror the logic in the recent C# PR. This command is intended to replicate what that logic does.

AlitzelMendez: we use it to send it to copilot, so copilot is the one actually using it, and the message is for copilots interpretation, but I get that otherwiser we will need another separate endpoint that processes the reasons

PR #13962, opened 2026-02-11
Add Rust support to azsdk MCP

heaths:


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Languages/Samples/RustSampleLanguageContext.cs
```diff
@@ -0,0 +1,101 @@
...
+        if (!paths.Any())
+        {
+            throw new ArgumentException("At least one path must be provided", nameof(paths));
+        }
+
+        var allPaths = paths.ToList();
+        var packagePath = paths.First();
+
+        workspacePath = await gitHelper.DiscoverRepoRootAsync(packagePath, ct);
+        allPaths.Add(workspacePath);
```
deyaaeldeen, line 95: This makes me nervous, what is the intent here? If we want to load certain files in the top-level, I think we should add them explicitly, otherwise, we would be trying to load the entire repo and I am not sure if the files of interest will make it or not.

RickWinter: The intent is for the model to leverage the source code in the repo. Heath's experiments have shown this leads to much better results when coding Rust.

Is there an example or data that shows this design results in files of interest not being picked up?

heaths: Sorry, I meant to add `Cargo.toml` to that path. I did some refactoring and guess I forgot to migrate that line.

LarryOsterman (approved):

RickWinter (approved):

heaths:
No need to approve. This is a draft. It's far from incomplete. I hope to pick it back up next week. As you know, higher priority things have been popping up recently.

heaths:
Still needed but higher priority items remain.

PR #13963, opened 2026-02-12
update typespec project path in release status telemetry

praveenkuttappan:
Update TypeSpec project path found in the release plan in response so that it will be available in telemetry.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/PackageReleaseStatusTool.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/PackageReleaseStatusTool.cs
@@ -128,6 +128,7 @@ namespace Azure.Sdk.Tools.Cli.Tools.ReleasePlan
                 }
 
                 response.ReleasePlanId = releasePlan.ReleasePlanId;
+                response.TypeSpecProject = releasePlan.APISpecProjectPath;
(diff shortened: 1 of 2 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/PackageReleaseStatusTool.cs
```diff
@@ -128,6 +128,7 @@ public async Task<ReleaseStatusUpdateResponse> UpdatePackageReleaseStatus(string
                 }
 
                 response.ReleasePlanId = releasePlan.ReleasePlanId;
+                response.TypeSpecProject = releasePlan.APISpecProjectPath;
```
benbp, line 131: It would be more defensive if `ReleaseStatusUpdateResponse` threw an argument exception for any required fields in its constructor, and then only got instantiated at the end (or in the error returns above). That way we're less vulnerable to these types of bugs.

benbp (approved):

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Tools/ReleasePlan/PackageReleaseStatusTool.cs
praveenkuttappan: It's possible to have blank spec path in release plan because we started asking and storing spec path in release planner recently.

Issue #13965, opened 2026-02-12
[js] Exclude Alpha Versions from Beta Changelog Comparison


skywing918:
Alpha versions already exist on npm. When generating a new beta release, the current changelog comparison logic incorrectly involves alpha versions during version comparison. Beta releases should not compare against alpha versions, as beta is the first preview stage. We should adjust the logic to exclude alpha versions from beta changelog generation.

PR #13966, opened 2026-02-12
 Previous published version isn’t a beta, should not comparison.

skywing918:
Alpha versions already exist on npm. When generating a new beta release, should be first beta
image

image

if beta version already exist on npm, should compared with beta version
image

image

fixes https://github.com/Azure/azure-sdk-tools/issues/13965


kazrael2119:
LGTM

PR #13982, opened 2026-02-13
Type alias generator

KarishmaGhiya:


tools/apiview/parsers/js-api-parser/src/tokenGenerators/typeAlias.ts
```diff
@@ -0,0 +1,111 @@
...
+          }),
+        );
+        processExcerptTokens(tp.defaultTypeExcerpt.spannedTokens, tokens, deprecated);
+      }
+
+      if (index < typeParameters.length - 1) {
+        tokens.push(createToken(TokenKind.Punctuation, ",", { hasSuffixSpace: true, deprecated }));
+      }
+    });
+    tokens.push(createToken(TokenKind.Punctuation, ">", { deprecated }));
```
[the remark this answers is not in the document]
KarishmaGhiya: Done

tools/apiview/parsers/js-api-parser/src/tokenGenerators/variable.ts
```diff
@@ -0,0 +1,59 @@
...
+  tokens.push(createToken(TokenKind.Keyword, "export", { hasSuffixSpace: true, deprecated }));
+
+  // Check for default export
+  const isDefaultExport = item.excerptTokens.some((t) => t.text.includes("export default"));
+  if (isDefaultExport) {
+    tokens.push(createToken(TokenKind.Keyword, "default", { hasSuffixSpace: true, deprecated }));
+  }
+
+  // Add const keyword
+  tokens.push(createToken(TokenKind.Keyword, "const", { hasSuffixSpace: true, deprecated }));
```
jeremymeng, line 29: is `let` variable possible?  We probably shouldn't have this case though in our API

tools/apiview/parsers/js-api-parser/src/tokenGenerators/variable.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (typeText) {
+    children = parseTypeText(typeText, tokens, deprecated);
+    if (!children) {
+      tokens.push(createToken(TokenKind.Punctuation, ";", { deprecated }));
+    }
+  } else {
+    tokens.push(createToken(TokenKind.Punctuation, ";", { deprecated }));
+  }
+
+  return { tokens, children };
```
jeremymeng, line 59: is value missing? for example the value in `export const MAX_SIZE = 1000000;`

KarishmaGhiya: I can check and get back

KarishmaGhiya: done

tools/apiview/parsers/js-api-parser/src/tokenGenerators/variable.ts
KarishmaGhiya: The @microsoft/api-extractor-model's ApiVariable class doesn't expose a variableDeclarationKind property to distinguish between const/let/var, so the parser currently has no way to determine the original keyword. In practice for Azure SDK packages, all exported variables are expected to be const per SDK guidelines, making let an edge case that isn't supported.

PR #13996, opened 2026-02-13
Spec: Live and Recorded Tests

timovv:
Spec for core testing deliverables that are a part of Scenario 2.

### DIFF
--- a/tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
+++ b/tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
@@ -1,0 +1,182 @@
+# Spec: Live and Recorded Tests
+
+## Table of Contents
+
+- [Definitions](#definitions)
+- [Background / Problem Statement](#background--problem-statement)
+- [Goals and Exceptions/Limitations](#goals-and-exceptionslimitations)
+- [Design Proposal](#design-proposal)
+- [Alternatives Considered](#alternatives-considered)
+- [Success Criteria](#success-criteria)
+- [Agent Prompts](#agent-prompts)
+- [CLI Commands](#cli-commands)
+- [Testing Strategy](#testing-strategy)
+
+---
+
+## Definitions
+
+- **Test Mode**: The mode in which tests are executed - Record, Playback, or Live.
+- **Record Mode**: Tests run against real Azure resources and record HTTP interactions for later playback.
+- **Playback Mode**: Tests run against previously recorded HTTP interactions without hitting real Azure resources.
+- **Live Mode**: Tests run against real Azure resources without recording.
+- **Test Resources**: Azure resources required to run live or recorded tests (e.g., storage accounts, key vaults).
+- **Test Proxy**: The tool that intercepts and records/plays back HTTP traffic during test execution.
+
+---
+
+## Background / Problem Statement
+
+### Current State
+
+As part of the scenario 1 work, we support running tests in **playback mode only** in all supported languages.
+
+### Why This Matters
+
+Being able to run tests against live resources, and in record mode, is necessary step for brand new packages. Before tests can be run in playback mode, tests must be recorded and written.
+
+---
+
+## Goals and Exceptions/Limitations
+
+### Goals
+
+- [ ] Support for running tests in record, playback, and live mode
+- [ ] Test resource deployment and cleanup
+- [ ] Stretch goal: Bicep authoring support. To be discussed in a future spec.
+- [ ] Stretch goal: Test authoring support. To be discussed in a future spec.
+
+---
+
+## Design Proposal
+
+### Running tests in record and playback mode
+
+The `azsdk_package_run_tests` tool will be enhanced to support live and recorded tests. This will be done via a new `TestMode` enum with values `Record`, `Live`, and `Playback` passed to the `LanguageService` `RunAllTests` method. Language owners will be responsible for handling the enum value and using it to set the correct test mode. If no test mode was specified by the user, the tool will select `Playback` as the default value.
+
+In live and record modes, a dictionary of environment variables will also be provided which should be passed through to the language's test runner. These environment variables will typically come from the New-TestResources.ps1 script.
+
+```cs
+public enum TestMode
+{
+    Record,
+    Live,
+    Playback
+}
+
+public abstract class LanguageService
+{
+    // Update to RunAllTests method to be implemented by each language
+    public abstract Task<TestRunResponse> RunAllTests(string packagePath, TestMode testMode, IDictionary<string, string>? liveTestEnvironment, CancellationToken ct = default);
+}
+```
+
+### Test resource deployment and cleanup
+
+Deployment of test resources involves many different variables depending on the user's needs. We do not know which subscription, tenant, etc. the user might have available or want to use. For example, devs on the SDK team might use the Playground subscription to deploy test resources, but devs on service teams may have their own subscriptions they might want to use. Or, maybe the user wants to deploy to TME. Given the array of possibilities here, a __skill__ will be developed for deploying test resources in the context of running live and recorded tests.
+
+#### Skill outline
+
+The skill will guide the agent through these steps:
+
+1. Ensure that the user has an active Azure PowerShell context (`Get-AzContext`), and, if not, guide them through logging in via Azure PowerShell (`Connect-AzureAccount`).
+2. Run the `eng/common/TestResources/New-TestResources.ps1` PowerShell script with the necessary parameters (service directory, resource group name, deployment region, any additional parameters provided by the user e.g. enableHsm for Key Vault...)
+3. Save the environment variables output by the script to a well-known location to be passed the run tests tool
+4. Invoke the test tool in the selected test mode with the test environment
+5. Clean up test resources when done using the `eng/common/TestResources/Remove-TestResources.ps1` script.
+
+The skill will also include troubleshooting steps. For example, for when test resource deployment fails, it will have information about potential causes and fixes (e.g. auth issues).
+
+---
+
+## Alternatives Considered
+
+### Alternative: New tool instead of skill
+
+**Description:**
+Create a new tool with a rigid set of steps for deploying test resources.
+
+**Why not chosen:**
+A skill is more flexible and should do a better job at 'guiding' the user along the right path, instead of having a tool with a rigid set of steps that might not all apply to whatever the user is trying to do.
+
+---
+
+## Success Criteria
+
+This feature/tool is complete when:
+
+- [ ] All five languages successfully execute tests in all three modes
+- [ ] Test resources can be deployed and cleaned up via the skill
+
+---
+
+## Agent Prompts
+
+### Run Tests in Playback Mode
+
+**Prompt:**
+
+```text
+Run the tests for this package
+```
+
+**Expected Agent Activity:**
+
+Playback mode should be the default behavior.
+
+1. Agent should call the run test tool in playback mode and report the results.
+
+### Run Tests in Record Mode
+
+**Prompt:**
+
+```text
+Run the tests in record mode using the live test deployment specified in #path/to/.env
+```
+
+**Expected Agent Activity:**
+
+1. Agent should call the run test tool in record mode directly with the provided environment variables instead of attempting to deploy test resources.
+
+### Deploy Test Resources and Run Live Tests
+
+**Prompt:**
+
+```text
+Deploy test resources and run live tests
+```
+
+**Expected Agent Activity:**
+
+1. Agent should follow steps in the test resource deployment skill to deploy the test resources and run the tests.
+
+---
+
+## CLI Commands
+
+### Run Tests
+
+**Command:**
+
+```bash
+azsdk package test run --mode <playback|record|live> [--package-path <path>] [--test-environment <path-to-.env>]
+```
+
+**Options:**
+
+- `-m, --mode <mode>`: Test mode - playback, record, or live (default: playback)
+- `-p, --package-path <path>`: Path to the package to test (optional)
+- `--test-environment <path>`: Path to a .env file containing deployment environment variables (optional)
+
+**Expected Output:**
+
+```
+[[ a log of the test run; will vary by language. ]]
+```
+
+---
+
+## Testing Strategy
+
+- Unit and integration tests to ensure that the test tool passes through the correct parameters to the language implementations
+- Evaluation of the skill description and tool calls made to ensure the skill is working as expected
--- a/tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
+++ b/tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
@@ -56,6 +56,8 @@ The `azsdk_package_run_tests` tool will be enhanced to support live and recorded
 
 In live and record modes, a dictionary of environment variables will also be provided which should be passed through to the language's test runner. These environment variables will typically come from the New-TestResources.ps1 script.
 
+When tests are run in record mode and all tests pass, the tool will automatically push the recorded test assets to the assets repo. This avoids the need for the user or a separate skill to handle the assets push step.
+
 ```cs
 public enum TestMode
 {
@@ -137,6 +139,7 @@ Run the tests in record mode using the live test deployment specified in #path/t
 **Expected Agent Activity:**
 
 1. Agent should call the run test tool in record mode directly with the provided environment variables instead of attempting to deploy test resources.
+2. If all tests pass, the tool will push the recorded test assets to the assets repo automatically.
 
 ### Deploy Test Resources and Run Live Tests
 
@@ -168,10 +171,26 @@ azsdk package test run --mode <playback|record|live> [--package-path <path>] [--
 - `-p, --package-path <path>`: Path to the package to test (optional)
 - `--test-environment <path>`: Path to a .env file containing deployment environment variables (optional)
 
-**Expected Output:**
+**Expected Output (success):**
 
+```text
+<test runner output; will vary by language>
+
+"Succeeded"
+Package: <package-name>
+Language: <language>
+Version: <version>
+Package Type: <SdkType>
 ```
-[[ a log of the test run; will vary by language. ]]
+
+**Expected Output (failure):**
+
+```text
+<test runner output; will vary by language>
+
+[ERROR] Test run failed with a non-zero exit code
+[NEXT STEPS]
+Inspect the error message and attempt to resolve it
 ```
 
 ---
--- a/tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
+++ b/tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
@@ -33,7 +33,7 @@ As part of the scenario 1 work, we support running tests in **playback mode only
 
 ### Why This Matters
 
-Being able to run tests against live resources, and in record mode, is necessary step for brand new packages. Before tests can be run in playback mode, tests must be recorded and written.
+Being able to run tests against live resources, and in record mode, is a necessary step for brand new packages. Before tests can be run in playback mode, tests must be recorded and written.
 
 ---
 
@@ -81,11 +81,11 @@ Deployment of test resources involves many different variables depending on the 
 
 The skill will guide the agent through these steps:
 
-1. Ensure that the user has an active Azure PowerShell context (`Get-AzContext`), and, if not, guide them through logging in via Azure PowerShell (`Connect-AzureAccount`).
+1. Ensure that the user has an active Azure PowerShell context (`Get-AzContext`), and, if not, guide them through logging in via Azure PowerShell (`Connect-AzAccount`).
 2. Run the `eng/common/TestResources/New-TestResources.ps1` PowerShell script with the necessary parameters (service directory, resource group name, deployment region, any additional parameters provided by the user e.g. enableHsm for Key Vault...)
-3. Save the environment variables output by the script to a well-known location to be passed the run tests tool
+3. Save the environment variables output by the script to a well-known location to be passed to the run tests tool
 4. Invoke the test tool in the selected test mode with the test environment
-5. Clean up test resources when done using the `eng/common/TestResources/Remove-TestResources.ps1` script.
+5. Ask the user whether they want to clean up test resources. If yes, clean up using the `eng/common/TestResources/Remove-TestResources.ps1` script. If no, leave resources in place for subsequent test runs.
 
 The skill will also include troubleshooting steps. For example, for when test resource deployment fails, it will have information about potential causes and fixes (e.g. auth issues).
 
@@ -162,7 +162,7 @@ Deploy test resources and run live tests
 **Command:**
 
 ```bash
-azsdk package test run --mode <playback|record|live> [--package-path <path>] [--test-environment <path-to-.env>]
+azsdk package test run --mode <playback|record|live> [--package-path <path>] [--test-environment <path-to-.env>] [--timeout <seconds>]
 ```
 
 **Options:**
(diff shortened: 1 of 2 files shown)


tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,182 @@
...
+
+---
+
+## Design Proposal
+
+### Running tests in record and playback mode
+
+The `azsdk_package_run_tests` tool will be enhanced to support live and recorded tests. This will be done via a new `TestMode` enum with values `Record`, `Live`, and `Playback` passed to the `LanguageService` `RunAllTests` method. Language owners will be responsible for handling the enum value and using it to set the correct test mode. If no test mode was specified by the user, the tool will select `Playback` as the default value.
+
+In live and record modes, a dictionary of environment variables will also be provided which should be passed through to the language's test runner. These environment variables will typically come from the New-TestResources.ps1 script.
```
l0lawrence, line 57: as in the tool will provide them to the user? like set ENVAR 1,2,3 please and re-run?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,182 @@
...
+**Prompt:**
+
+```text
+Deploy test resources and run live tests
+```
+
+**Expected Agent Activity:**
+
+1. Agent should follow steps in the test resource deployment skill to deploy the test resources and run the tests.
+
```
l0lawrence, line 155: what are we doing with test failures/how will we report that (for example there could be a lot of failures for run live) do we list out all the tests that failed or just the # of tests that failed etc.

l0lawrence: maybe a more detailed output of the tool could be helpful

l0lawrence: and on this note for record -- do we need a skill to guide the user to update the assets repo, or will we handle that in the tool itself?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: We should be able to set environment variables with the ProcessHelper so that the environment variables are available to the test subprocess. The user shouldn't have to do anything at that point.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: It's no different to running the tests in playback where the tool just returns the output of the test run. That should have all the necessary information, and then it's up to the agent to summarize the output and decide what it wants to do with it. I did want to put some effort into actually parsing the output and giving a summary at some point but I don't think it's strictly necessary right now.

I can update the tool output below with the actual shape that gets reported, it's  just not really changing from scenario 1.

Pushing to the assets repo is a great point and honestly I forgot that was a thing that has to happen. Let me think about it. It would be nice to put it in the skill if we can.

timovv: OK I thought about it and I don't see any issue with just integrating it into the tool like "if the tests pass then the tool is responsible for pushing the tests to the assets repo"

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
samvaity: So, these will be read from the environment if they are set else the tool will notify the user to set them, is that right?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,201 @@
...
+#### Skill outline
+
+The skill will guide the agent through these steps:
+
+1. Ensure that the user has an active Azure PowerShell context (`Get-AzContext`), and, if not, guide them through logging in via Azure PowerShell (`Connect-AzureAccount`).
+2. Run the `eng/common/TestResources/New-TestResources.ps1` PowerShell script with the necessary parameters (service directory, resource group name, deployment region, any additional parameters provided by the user e.g. enableHsm for Key Vault...)
+3. Save the environment variables output by the script to a well-known location to be passed the run tests tool
+4. Invoke the test tool in the selected test mode with the test environment
+5. Clean up test resources when done using the `eng/common/TestResources/Remove-TestResources.ps1` script.
```
samvaity: Will it be the skill that guides when to do the cleanup ?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,201 @@
...
+Playback mode should be the default behavior.
+
+1. Agent should call the run test tool in playback mode and report the results.
+
+### Run Tests in Record Mode
+
+**Prompt:**
+
+```text
+Run the tests in record mode using the live test deployment specified in #path/to/.env
```
samvaity, line 136: if the path to env is not specified, will the default behavior be to deploy test resources?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
samvaity: could we add a failure scenario case in examples?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,201 @@
...
+- [ ] Support for running tests in record, playback, and live mode
+- [ ] Test resource deployment and cleanup
+- [ ] Stretch goal: Bicep authoring support. To be discussed in a future spec.
+- [ ] Stretch goal: Test authoring support. To be discussed in a future spec.
+
+---
+
+## Design Proposal
+
+### Running tests in record and playback mode
```
raych1, line 53: Does the test tool start test proxy when it's called in `record` mode? Should we add `test proxy` to the setup requirement?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,201 @@
...
+```bash
+azsdk package test run --mode <playback|record|live> [--package-path <path>] [--test-environment <path-to-.env>]
+```
+
+**Options:**
+
+- `-m, --mode <mode>`: Test mode - playback, record, or live (default: playback)
+- `-p, --package-path <path>`: Path to the package to test (optional)
+- `--test-environment <path>`: Path to a .env file containing deployment environment variables (optional)
+
```
raych1, line 174: Do we consider adding a timeout option since live tests might run much longer?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,201 @@
...
+### Why This Matters
+
+Being able to run tests against live resources, and in record mode, is necessary step for brand new packages. Before tests can be run in playback mode, tests must be recorded and written.
+
+---
+
+## Goals and Exceptions/Limitations
+
+### Goals
```
raych1, line 42: Seems a `test-filter` feature was mentioned before, is it still part of the goals of `scenario 2`?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Yes. The thought here is that the skill manages the deployment and therefore also should manage the cleanup.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Yes, it would be similar to the live mode example below. I didn't do every single combination here in the interest of brevity.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Can you be more specific around what failure scenario you would like to see? Since it's a skill, the exact output in a failure scenario is not going to be determinstic.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Good point, we should add one. I would advocate for each language setting a reasonable (increased) default timeout for live tests, but it should also be possible to override.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: So, I can only speak for JavaScript here with certainty, but I _believe_ this is the case for other languages as well: the language's test runner itself handles installing and starting the test proxy if it isn't running already. Assuming that's true for everyone, we wouldn't need it as a setup requirement. What's more, considering we haven't run into this issue yet with playback tests, which we already support, and which do require the test proxy, I think we should be covered here.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: This spec focuses on how we handle live and recorded tests. While filtering is in scope for scenario 2, I would consider it separate to what's in this doc.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: If the environment variables are present in the environment they would cascade down to the test runner, so that would work. But since the environment variables needed vary a lot based on service and language, I don't think we'd have a reliable way of detecting whether the variables have been set or not. That said, the tests would presumably fail if the env vars were missing, hopefully with a useful otuput message. I think that will probably be enough.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,201 @@
...
+
+1. Agent should call the run test tool in record mode directly with the provided environment variables instead of attempting to deploy test resources.
+2. If all tests pass, the tool will push the recorded test assets to the assets repo automatically.
+
+### Deploy Test Resources and Run Live Tests
+
+**Prompt:**
+
+```text
+Deploy test resources and run live tests
```
ronniegeraghty, line 149: Assuming if you just asked it to run live tests without explicitly stating to do the deploy, it still will deploy the test resources.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
m-redding: This is how it works in the .NET Azure test framework:
- if the standard variables set by the new test resources script couldn't be found, we ask the user if we can try to bootstrap test resources using new test resources
- If we bootstrap and the needed environment variables are still not there, it fails with "Unable to find environment variable {prefixedName} or {name} required by test. Make sure the test environment was initialized using the eng/common/TestResources/New-TestResources.ps1 script."

So in .NET we wouldn't need any work here, I think it would honestly be easier for the user to not even have to pass any environment variables at this step, but as long as the variables are optional then that seems fine to me

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,201 @@
...
+---
+
+## Agent Prompts
+
+### Run Tests in Playback Mode
+
+**Prompt:**
+
+```text
+Run the tests for this package
```
ronniegeraghty, line 122: Are all tests for a package always able to be recorded. 
My thought is if a user asks for test to be run, I assume they expect the entire test suite for the package to be run, and if some tests are live only tests, they wouldn't be run. 

If I'm wrong and no tests are live only, then not an issue. If I'm correct, we should think about how we run the full suite of playback tests and live only tests.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,201 @@
...
+
+Deployment of test resources involves many different variables depending on the user's needs. We do not know which subscription, tenant, etc. the user might have available or want to use. For example, devs on the SDK team might use the Playground subscription to deploy test resources, but devs on service teams may have their own subscriptions they might want to use. Or, maybe the user wants to deploy to TME. Given the array of possibilities here, a __skill__ will be developed for deploying test resources in the context of running live and recorded tests.
+
+#### Skill outline
+
+The skill will guide the agent through these steps:
+
+1. Ensure that the user has an active Azure PowerShell context (`Get-AzContext`), and, if not, guide them through logging in via Azure PowerShell (`Connect-AzureAccount`).
+2. Run the `eng/common/TestResources/New-TestResources.ps1` PowerShell script with the necessary parameters (service directory, resource group name, deployment region, any additional parameters provided by the user e.g. enableHsm for Key Vault...)
+3. Save the environment variables output by the script to a well-known location to be passed the run tests tool
```
m-redding: Like I was saying in my other comment I think this step can be skipped for .NET

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,201 @@
...
+public abstract class LanguageService
+{
+    // Update to RunAllTests method to be implemented by each language
+    public abstract Task<TestRunResponse> RunAllTests(string packagePath, TestMode testMode, IDictionary<string, string>? liveTestEnvironment, CancellationToken ct = default);
+}
+```
+
+### Test resource deployment and cleanup
+
+Deployment of test resources involves many different variables depending on the user's needs. We do not know which subscription, tenant, etc. the user might have available or want to use. For example, devs on the SDK team might use the Playground subscription to deploy test resources, but devs on service teams may have their own subscriptions they might want to use. Or, maybe the user wants to deploy to TME. Given the array of possibilities here, a __skill__ will be developed for deploying test resources in the context of running live and recorded tests.
```
ronniegeraghty, line 78: Is it common for service teams to run SDK tests on their own subscriptions and not just use one provided by our team?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
m-redding: I think this makes sense but I wonder if it could be nice to allow longer living resources if someone is doing multiple iterations?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
m-redding: Honestly I've had good experiences with just letting the LLM figure out how to address test failures, so I think it makes sense to have this be a skill, and leave the agent to handle the tests when they fail

m-redding (approved):

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: I don't think people outside of our team use the Playground subscription at least. I don't know if there are other subscriptions we own that they can use.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: That should be fine. It sounds like we'll need to tweak the skill slightly for different languages.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Yeah, good point. Maybe we could have the skill ask the user if they want the resources cleaned up or not instead of doing it proactively?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: The standard behavior (at least in JS...) is that the default mode is playback and then if there are any live-only tests they are skipped unless the test mode is set to live. I think this is fine personally but maybe I'm just used to it.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Yeah, or rather it would figure out whether test resources need to be deployed or not, and deploy them if needed.

l0lawrence (approved):
lgtm - skill sounds good for test failures :)

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
raych1: That makes sense, and would like to see the feature to be supported in scenario 2.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Updated to ask so that we support this

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Added a `--timeout` option to the spec. Thanks for the feedback

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Yeah, they're optional -- so if your language doesn't need them I think it should be fine

raych1 (approved):

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+**Expected Output (failure):**
+
+```text
+<test runner output; will vary by language>
+
+[ERROR] Test run failed with a non-zero exit code
+[NEXT STEPS]
+Inspect the error message and attempt to resolve it
```
samvaity, line 194: Do we not give details about what failed in the test ?

samvaity: Since this is a skill, we should/would have details.

samvaity (approved):

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: The details will show up in the

> `<test runner output; will vary by language>`

for the LLM to interpret. This is what we have for playback tests today and it seems to work well enough.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Test resource deployment and cleanup
+
+Deployment of test resources involves many different variables depending on the user's needs. We do not know which subscription, tenant, etc. the user might have available or want to use. For example, devs on the SDK team might use the Playground subscription to deploy test resources, but devs on service teams may have their own subscriptions they might want to use. Or, maybe the user wants to deploy to TME. Given the array of possibilities here, a __skill__ will be developed for deploying test resources in the context of running live and recorded tests.
+
+#### Skill outline
+
+The skill will guide the agent through these steps:
+
+1. Ensure that the user has an active Azure PowerShell context (`Get-AzContext`), and, if not, guide them through logging in via Azure PowerShell (`Connect-AzAccount`).
```
benbp, line 84: FYI at least for TME login, `Connect-AzAccount` _cannot_ be run with `-UseDeviceAuthentication` due to attestation policies from our tenant admin. So this command always has to be run in an interactive terminal that has the ability to launch a browser (i.e. probably not an agent session). Users on WSL also have to have the `wslu` package installed for `wslview` which can launch the windows browser process.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,202 @@
...
+2. Run the `eng/common/TestResources/New-TestResources.ps1` PowerShell script with the necessary parameters (service directory, resource group name, deployment region, any additional parameters provided by the user e.g. enableHsm for Key Vault...)
+3. Save the environment variables output by the script to a well-known location to be passed to the run tests tool
+4. Invoke the test tool in the selected test mode with the test environment
+5. Ask the user whether they want to clean up test resources. If yes, clean up using the `eng/common/TestResources/Remove-TestResources.ps1` script. If no, leave resources in place for subsequent test runs.
+
+The skill will also include troubleshooting steps. For example, for when test resource deployment fails, it will have information about potential causes and fixes (e.g. auth issues).
```
benbp, line 90: We can source some of these steps from various docs in our wiki.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
benbp: We also have some classes in `azsdk` that can handle parsing `.trx` file data to pass more structured output to an agent. We could consider having the test runner always outputting these test result files and then using our purpose built test analysis tools/commands. This is more useful as a scenario when there are hundreds of failing tests vs. just a few.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
```diff
@@ -0,0 +1,202 @@
...
+---
+
+## CLI Commands
+
+### Run Tests
+
+**Command:**
+
+```bash
+azsdk package test run --mode <playback|record|live> [--package-path <path>] [--test-environment <path-to-.env>] [--timeout <seconds>]
```
benbp, line 165: This could quickly going to become super complex if we want to support stuff like per-language test filters and other language tooling variability. Is the idea that we just have a simple happy path test command for recording that intentionally excludes complexity (since you aren't doing filtering anyway)?

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: The problem here is that not every language outputs `.trx` files outside of the pipeline context. We use a pipeline step to convert e.g. [here](https://github.com/Azure/azure-sdk-for-js/blob/7a3ec04510e38457ce7b22cb137d16d31a216c08/eng/pipelines/templates/steps/test.yml#L78-L83) which is not available locally. I was exploring making the test runner output more structured at one point (return an object instead of a response type with the raw output) but we haven't had a strong need for that feature yet -- the LLM seems to handle just reading the console output pretty well.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Yeah, I know we're papering over a lot of complexity here. The goal here is essentially to make the existing command that we have for playback tests work for live and recorded tests, and no more. So I think that "happy path" is a good description.

FWIW, re filtering (but maybe for a later discussion): I was sort of hoping that we could just make the filter a transparent pass-through for each language's the test runner to handle in its own way, at least as a first cut. Trying to unify that sounds like it would be a painful task indeed.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
timovv: Interestingly I had no issues with Connect-AzAccount launching a browser from a VS Code Copilot terminal. Granted, it was a while ago when I last tried this. And the CLI will likely be a different story.

benbp: Yeah vscode can do it fine, it's just a wsl specific issue.

tools/azsdk-cli/docs/specs/4-live-and-recorded-tests.md
benbp: yeah or maybe something like `azsdk package test run <args> -- <passthrough args` which is a typical pattern in other CLIs

PR #14000, opened 2026-02-13
[azsdk] integrate TypeSpec customizations into CodeCustomiztionUpdateTool

chrisradek:
This PR adds the TypeSpec customization service to the CodeCustomizationUpdateTool. It assumes that the azure-rest-api-specs repo and the azure-sdk-for-\* repos are both available locally, so it has been updated to take the typespec project path, and no longer requires a commit sha.

This also updates the sdk regen/build to happen after the typespec customizations are applied.

There is one part of the integration for scenario 2 - https://github.com/Azure/azure-sdk-tools/pull/13642 includes integrating the classifier into the CodeCustomizationUpdateTool, which will be responsible for determining if the tool is done with its work, or if changes can be handled by the tool.


m-redding:
replaced by https://github.com/Azure/azure-sdk-tools/pull/14190

Issue #13960, opened 2026-02-11
[APIView] Adding a reviewer sends duplicate notification emails to all existing reviewers


smw-ms:
**Root Cause**
In [`APIRevisionsController.cs` (lines 200-201)](https://github.com/Azure/azure-sdk-tools/blob/main/src/dotnet/APIView/APIViewWeb/LeanControllers/APIRevisionsController.cs#L200-L201) and [`Review.cshtml.cs` (lines 325-326)](https://github.com/Azure/azure-sdk-tools/blob/main/src/dotnet/APIView/APIViewWeb/Pages/Assemblies/Review.cshtml.cs#L325-L326), the full `reviewers` HashSet is passed to `NotifyApproversOfReview`, which then emails everyone in that list

**Expected Behavior**
Only newly-added reviewers should receive an email notification.

Issue #13974, opened 2026-02-12
[APIView] Convert APIView notification emails to a centralized template-based system


smw-ms:
Existing infrastructure (namespace review) : https://github.com/smw-ms/azure-sdk-tools/blob/71ea336969af3b492afc9bc59279b40b8317c5d6/src/dotnet/APIView/APIViewWeb/Services/EmailTemplateService.cs#L12

APIView notification emails are currently built using ad‑hoc string concatenation. 
Templates would define the email structure and content, with code supplying only the required data.

PR #14005, opened 2026-02-13
[APIView] Fix duplicate reviewer request emails for already-assigned

smw-ms:
Resolves: https://github.com/Azure/azure-sdk-tools/issues/13960 & https://github.com/Azure/azure-sdk-tools/issues/13974

+ Added test email context
+ Renamed `SendEmailsAsync` to be more clear/distinguishable

### DIFF
--- a/src/dotnet/APIView/APIViewUnitTests/EmailTemplatePreviewGeneratorTests.cs
+++ b/src/dotnet/APIView/APIViewUnitTests/EmailTemplatePreviewGeneratorTests.cs
@@ -22,7 +22,7 @@ namespace APIViewUnitTests
     // artifacts/bin/APIViewUnitTests/email-previews for local visual inspection.
     public class EmailTemplatePreviewGeneratorTests
     {
-        [Fact]
+        [Fact(Skip = "Local visual inspection test. Enable explicitly for local runs when preview files are needed.")]
         public async Task GenerateEmailPreviews()
         {
             string repoRoot = FindRepositoryRoot();
--- a/src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
+++ b/src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
@@ -90,10 +90,11 @@ namespace APIViewWeb.Managers
 
         public async Task NotifyUserOnCommentTagAsync(CommentItemModel comment)
         {
+            var review = await _reviewRepository.GetReviewAsync(comment.ReviewId);
+
             foreach (string username in comment.TaggedUsers)
             {
                 if(string.IsNullOrEmpty(username)) continue;
-                var review = await _reviewRepository.GetReviewAsync(comment.ReviewId);
                 var user = await _userProfileRepository.TryGetUserProfileAsync(username);
                 var reviewUrl = ManagerHelpers.ResolveReviewUrl(
                     reviewId: review.Id,
@@ -105,7 +106,7 @@ namespace APIViewWeb.Managers
                 var content = await _emailTemplateService.RenderAsync(
                     EmailTemplateKey.CommentTag,
                     CommentTagEmailModel.Create(_apiviewEndpoint, comment, review, reviewUrl));
-                await SendUserEmailsAsync(review.PackageName, "Comment Mention", user, content);
+                await SendEmailAsync(user?.Email, BuildEmailSubject("Comment Mention", review.PackageName), content);
             } 
         }
 
@@ -124,8 +125,8 @@ namespace APIViewWeb.Managers
                 var reviewerProfile = await _userProfileRepository.TryGetUserProfileAsync(reviewer);
                 var content = await _emailTemplateService.RenderAsync(
                     EmailTemplateKey.ReviewerAssigned,
-                    ReviewerAssignedEmailModel.Create(_apiviewEndpoint, userProfile.UserName, apiRevision.Id, apiRevision.PackageName));
-                await SendUserEmailsAsync(apiRevision.PackageName, "Review Requested", reviewerProfile, content);
+                    ReviewerAssignedEmailModel.Create(_apiviewEndpoint, userProfile.UserName, apiRevision.ReviewId, apiRevision.PackageName));
+                await SendEmailAsync(reviewerProfile?.Email, BuildEmailSubject("Review Requested", apiRevision.PackageName), content);
             }
         }
 
@@ -136,6 +137,7 @@ namespace APIViewWeb.Managers
                 NewRevisionEmailModel.Create(_apiviewEndpoint, review, revision));
             await SendSubscriberEmailsAsync(review, user, htmlContent, null, "New Revision Uploaded");
         }
+        
         /// <summary>
         /// Toggle Subscription to a Review
         /// </summary>
@@ -207,11 +209,6 @@ namespace APIViewWeb.Managers
             return $"[APIView] {eventName} for {packageName}";
         }
 
-        private async Task SendUserEmailsAsync(string packageName, string eventName, UserProfileModel user, string htmlContent)
-        {
-            // SendEmailAsync already handles email validation
-            await SendEmailAsync(user.Email, BuildEmailSubject(eventName, packageName), htmlContent);
-        }
         private async Task SendSubscriberEmailsAsync(ReviewListItemModel review, ClaimsPrincipal user, string htmlContent, ISet<string> notifiedUsers, string eventName)
         {
             var initiatingUserEmail = GetUserEmail(user);
@@ -251,12 +248,6 @@ namespace APIViewWeb.Managers
             // Send single email to all subscribers
             var emailToList = string.Join("; ", subscribers);
             await SendEmailAsync(emailToList, BuildEmailSubject(eventName, review.PackageName), htmlContent);
-        }
-
-        private static string GetUserName(ClaimsPrincipal user)
-        {
-            var name = user.FindFirstValue(ClaimConstants.Name);
-            return string.IsNullOrEmpty(name) ? user.FindFirstValue(ClaimConstants.Login) : name;
         }
 
         private async Task<string> GetEmailAddress(string username)
@@ -301,7 +292,7 @@ namespace APIViewWeb.Managers
             return emailAddresses;
         }
 
-        public async Task NotifyApproversOnNamespaceReviewRequestAsync(ClaimsPrincipal user, ReviewListItemModel review, IEnumerable<ReviewListItemModel> languageReviews = null, string notes = "")
+        public async Task NotifyNamespaceReviewRequestRecipientsAsync(ClaimsPrincipal user, ReviewListItemModel review, IEnumerable<ReviewListItemModel> languageReviews = null, string notes = "")
         {
             try
             {
@@ -343,6 +334,13 @@ namespace APIViewWeb.Managers
             if (string.IsNullOrEmpty(_emailSenderServiceUrl))
             {
                 _logger.LogTrace("Email sender service URL is not configured. Email will not be sent to {EmailToList} with subject: {Subject}", emailToList, subject);
+                return;
+            }
+
+            if (string.IsNullOrWhiteSpace(emailToList))
+            {
+                _logger.LogTrace("Email recipient is empty. Email will not be sent for subject: {Subject}", subject);
+                _telemetryClient.TrackTrace($"Email recipient is empty. Email will not be sent for subject: {subject}");
                 return;
             }
 
--- a/src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
+++ b/src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
@@ -432,8 +432,9 @@ namespace APIViewWeb.Managers
             // Update the reviews identified by review IDs with namespace approval fields
             await MarkAssociatedReviewsForNamespaceReview(relatedReviews, userId, requestedOn, reviewGroupId);
 
-            // Send email notifications to preferred approvers with the actual language review data
-            await _notificationManager.NotifyApproversOnNamespaceReviewRequestAsync(user, typeSpecReview, sdkLanguageReviews);            return typeSpecReview;
+            // Send email notifications to language approvers (resolved via permissions) with the actual language review data
+            await _notificationManager.NotifyNamespaceReviewRequestRecipientsAsync(user, typeSpecReview, sdkLanguageReviews);
+            return typeSpecReview;
         }
 
         /// <summary>
--- a/src/dotnet/APIView/APIViewUnitTests/NamespaceApprovalTests.cs
+++ b/src/dotnet/APIView/APIViewUnitTests/NamespaceApprovalTests.cs
@@ -177,7 +177,7 @@ namespace APIViewUnitTests
                 .Callback<ReviewListItemModel>(review => updatedReviews.Add(review))
                 .Returns(Task.CompletedTask);
 
-            _mockNotificationManager.Setup(n => n.NotifyApproversOnNamespaceReviewRequestAsync(
+            _mockNotificationManager.Setup(n => n.NotifyNamespaceReviewRequestRecipientsAsync(
                     It.IsAny<ClaimsPrincipal>(), It.IsAny<ReviewListItemModel>(), It.IsAny<IEnumerable<ReviewListItemModel>>(), It.IsAny<string>()))
                 .Returns(Task.CompletedTask);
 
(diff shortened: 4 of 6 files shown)


src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -237,7 +251,7 @@ private async Task SendUserEmailsAsync<T>(T model, UserProfileModel user, string
             // SendEmailAsync already handles email validation
             await SendEmailAsync(user.Email, $"Notification from APIView - {model.PackageName}", htmlContent);
         }
-        private async Task SendEmailsAsync(ReviewListItemModel review, ClaimsPrincipal user, string htmlContent, ISet<string> notifiedUsers)
+        private async Task SendSubscriberNotificationsAsync(ReviewListItemModel review, ClaimsPrincipal user, string htmlContent, ISet<string> notifiedUsers)
```
tjprescott: Since this sends an email, it seems `SendSubscriberEmailAsync` is more self-documenting. A notification in APIView should be the little notification bell dot.

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -389,6 +402,25 @@ private async Task SendEmailAsync(string emailToList, string subject, string con
...
+        private static string AppendTestModeFooter(string content, string intendedRecipients, bool isTestMode)
+        {
+            if (!isTestMode)
+            {
+                return content;
+            }
+
+            var encodedRecipients = WebUtility.HtmlEncode(intendedRecipients);
+            return content + $"<br><br><hr><div style='font-size: 12px; color: #666;'><b>Test mode:</b> Intended recipients: {encodedRecipients}</div>";
+        }
```
tjprescott, line 383: Since you are touching this email, would it be possible to refactor this to use the template-based email service instead of constructing it in code? Also, why does it say "Test mode:"?

src/dotnet/APIView/APIViewWeb/Pages/Assemblies/Review.cshtml.cs
```diff
@@ -322,8 +322,8 @@ public async Task<IActionResult> OnPostToggleAPIRevisionApprovalAsync(string id,
         /// <returns></returns>
         public async Task<ActionResult> OnPostRequestReviewersAsync(string id, string apiRevisionId, HashSet<string> reviewers)
         {
-            await _apiRevisionsManager.AssignReviewersToAPIRevisionAsync(User, apiRevisionId, reviewers);
             await _notificationManager.NotifyApproversOfReview(User, apiRevisionId, reviewers);
+            await _apiRevisionsManager.AssignReviewersToAPIRevisionAsync(User, apiRevisionId, reviewers);
```
tjprescott, line 333: Why the switch in order here? It seems that if something about the DB update failed, you shouldn't email anyone. With this order, it's going to send an email and if the DB update fails, well, the email has already been sent.

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -84,9 +84,23 @@ public async Task NotifyUserOnCommentTag(CommentItemModel comment)
...
+
+            var existingReviewers = new HashSet<string>(
+                apiRevision.AssignedReviewers.Select(assignment => assignment.AssingedTo),
+                StringComparer.OrdinalIgnoreCase);
+
+            var newReviewers = reviewers
+                .Where(reviewer => !existingReviewers.Contains(reviewer))
+                .ToList();
+
+            foreach (var reviewer in newReviewers)
```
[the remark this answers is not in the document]
tjprescott: In practice, this is very unlikely to occur. You might see if it's a quick fix in which case I'd say go ahead and do it, but if it's a lot of work, I would say not worth it at this time.

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
smw-ms: It says "Test mode" because when I'm testing the email feature they all get redirected to the "apiview-email-test-address", but in order to test it effectively I need more context like who the email would really get sent to. This test footer does not appear in the production version because in that case "apiview-email-test-address" would be empty.

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
smw-ms: done

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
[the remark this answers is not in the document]
smw-ms: yeah it's not really a quick fix

src/dotnet/APIView/APIViewWeb/Pages/Assemblies/Review.cshtml.cs
smw-ms: changed. i moved the logic to filter new reviewers to be outside the notification manager. originally, i had swapped the call to check the current DB state first

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
smw-ms: done. refactored emails  to use the template-based email service

src/dotnet/APIView/ClientSPA/src/app/_components/review-page-options/review-page-options.component.html
```diff
@@ -116,7 +116,8 @@ <h6 class="mb-0 fw-semibold">Reviewers</h6>
                         Nothing to show
                     </div>
                     <div *ngFor="let approver of filteredApprovers"
-                        class="d-flex align-items-center px-3 py-2 reviewer-item cursor-pointer"
+                        class="d-flex align-items-center px-3 py-2 reviewer-item"
+                        style="cursor: pointer;"
```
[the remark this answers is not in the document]
smw-ms: the original way didn't work

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -78,28 +89,43 @@ public async Task NotifyUserOnCommentTag(CommentItemModel comment)
...
             {
                 var reviewerProfile = await _userProfileRepository.TryGetUserProfileAsync(reviewer);
-                await SendUserEmailsAsync(apiRevision, reviewerProfile,
-                    GetApproverReviewHtmlContent(userProfile, apiRevision));
+                var content = await _emailTemplateService.GetApproverReviewEmailAsync(
+                    userProfile.UserName,
+                    apiRevision.Id,
+                    apiRevision.PackageName);
+                await SendUserEmailsAsync(apiRevision, reviewerProfile, content);
             }
```
tjprescott: @AlitzelMendez now that we no longer have "Approvers" (or at least, to whatever extent we do have them, they are non-functional) does this require any changes or updates?

src/dotnet/APIView/APIViewWeb/Services/Interfaces/IEmailTemplateService.cs
```diff
@@ -0,0 +1,29 @@
...
+            string typeSpecUrl,
+            IEnumerable<ReviewListItemModel> languageReviews);
+
+        Task<string> GetApproverReviewEmailAsync(string requesterUserName, string reviewId, string reviewName);
+
+        Task<string> GetCommentTagEmailAsync(CommentItemModel comment, ReviewListItemModel review, string reviewUrl);
+
+        Task<string> GetSubscriberCommentEmailAsync(CommentItemModel comment, string elementUrl = null);
+
+        Task<string> GetNewRevisionEmailAsync(ReviewListItemModel review, APIRevisionListItemModel revision);
```
tjprescott: So, I'm wondering why we have these bespoke methods. Maybe this is "the C# way" (I'm not very knowledgeable about C#), but I would expect something like:

`GetEmailAsync(string templateName, Dictionary params)`

This is what I'm used to for templating languages. You point to some template file and have placeholders than are filled in by the params. Right now it feels like we are still programmatically building the emails. The fact that so many of the email templates are single lines strengthens that feeling.

Can we include example actual example emails in here somewhere? This is part of the goal of coalescing on the email UX. If I get an email like this:

Image

I'm gonna immediately archive it, and maybe make a filter rule if I get a lot.

Our emails should have a consistent look and feel, probably with a standard header and footer. A templating system should help achieve those goals.

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -68,7 +68,18 @@ public NotificationManager(IConfiguration configuration,
         public async Task NotifySubscribersOnComment(ClaimsPrincipal user, CommentItemModel comment)
```
tjprescott: If the `NotificationManager` basically means "emails" and we are using the email template system, do we still even need NotificationManager? Can we get rid of it?

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
AlitzelMendez: I was looking at the code, and I think the original implementation (not Summer changes) makes this naming super confusing.

The method is
```csharp
NotifyApproversOfReview(ClaimsPrincipal user, string apiRevisionId, HashSet<string> reviewers)
````

so
1. You can actually send this email no anyone as it has the parameters of reviewews
2. All the places where it is called a list of reviewers is shared so it is being used to notify reviewers rather than approvers 

so per se is not related with approvers at all, so no changes needed, it would be nice to update the name of the method in case we actually want an "Approvers" method

**Comment by smw-ms** [comment]
@tjprescott here is what the body of the emails would look like
image
image
image
image
image
image

**Example in Actual Outlook Context**
image

File: src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs (hunk 8 above)
Comment (smw-ms) [reply src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs]: i'd keep notificationManager. the template system just renders email content. notificationmanager still handles the notification workflow (who to notify, when, etc)

File: src/dotnet/APIView/APIViewUnitTests/EmailTemplatePreviewGeneratorTests.cs [hunk 9]
```diff
@@ -0,0 +1,236 @@
...
+    public class EmailTemplatePreviewGeneratorTests
+    {
+        [Fact]
+        public async Task GenerateEmailPreviews()
+        {
+            string repoRoot = FindRepositoryRoot();
+            string apiViewWebProjectPath = Path.Combine(repoRoot, "src", "dotnet", "APIView", "APIViewWeb");
+            string outputPath = Path.Combine(repoRoot, "artifacts", "bin", "APIViewUnitTests", "email-previews");
+
+            Directory.CreateDirectory(outputPath);
```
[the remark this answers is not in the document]
smw-ms: done

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -69,41 +69,72 @@ public NotificationManager(IConfiguration configuration,
...
+            await SendSubscriberEmailsAsync(review, user, content, comment.TaggedUsers, "New Comment");
         }
 
-        public async Task NotifyUserOnCommentTag(CommentItemModel comment)
+        public async Task NotifyUserOnCommentTagAsync(CommentItemModel comment)
         {
             foreach (string username in comment.TaggedUsers)
             {
                 if(string.IsNullOrEmpty(username)) continue;
                 var review = await _reviewRepository.GetReviewAsync(comment.ReviewId);
```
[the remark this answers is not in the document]
smw-ms: done

src/dotnet/APIView/APIViewWeb/Managers/ReviewManager.cs
```diff
@@ -432,8 +432,8 @@ public async Task<ReviewListItemModel> RequestNamespaceReviewAsync(ClaimsPrincip
             // Update the reviews identified by review IDs with namespace approval fields
             await MarkAssociatedReviewsForNamespaceReview(relatedReviews, userId, requestedOn, reviewGroupId);
 
-            // Send email notifications to language approvers with the actual language review data
-            await _notificationManager.NotifyApproversOnNamespaceReviewRequest(user, typeSpecReview, sdkLanguageReviews);            return typeSpecReview;
+            // Send email notifications to preferred approvers with the actual language review data
+            await _notificationManager.NotifyApproversOnNamespaceReviewRequestAsync(user, typeSpecReview, sdkLanguageReviews);            return typeSpecReview;
```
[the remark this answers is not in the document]
smw-ms: done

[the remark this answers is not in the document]
smw-ms: done

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -69,41 +69,72 @@ public NotificationManager(IConfiguration configuration,
...
             var apiRevision = await _apiRevisionRepository.GetAPIRevisionAsync(apiRevisionId);
+
             foreach (var reviewer in reviewers)
             {
                 var reviewerProfile = await _userProfileRepository.TryGetUserProfileAsync(reviewer);
-                await SendUserEmailsAsync(apiRevision, reviewerProfile,
-                    GetApproverReviewHtmlContent(userProfile, apiRevision));
+                var content = await _emailTemplateService.RenderAsync(
+                    EmailTemplateKey.ReviewerAssigned,
+                    ReviewerAssignedEmailModel.Create(_apiviewEndpoint, userProfile.UserName, apiRevision.Id, apiRevision.PackageName));
```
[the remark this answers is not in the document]
smw-ms: fixed

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -171,77 +202,17 @@ public async Task UnsubscribeAsync(ReviewListItemModel review, ClaimsPrincipal u
...
-            return sb.ToString();
+            return $"[APIView] {eventName} for {packageName}";
         }
 
-        private static string GetContentHeading(CommentItemModel comment, bool includeHtml) =>
-            $"{(includeHtml ? $"<b>{comment.CreatedBy}</b>" : $"{comment.CreatedBy}")} commented on this review.";
-
-        private async Task SendUserEmailsAsync<T>(T model, UserProfileModel user, string htmlContent) where T : BaseListitemModel 
+        private async Task SendUserEmailsAsync(string packageName, string eventName, UserProfileModel user, string htmlContent)
         {
```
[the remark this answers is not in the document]
smw-ms: done

tjprescott (commented):
I'm going to wait for Ali's approval since there's a lot of C# changes, but these new emails look really great! Thanks @smw-ms!!

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
```diff
@@ -69,42 +69,75 @@ public NotificationManager(IConfiguration configuration,
...
+                    elementId: comment.ElementId);
+                var content = await _emailTemplateService.RenderAsync(
+                    EmailTemplateKey.CommentTag,
+                    CommentTagEmailModel.Create(_apiviewEndpoint, comment, review, reviewUrl));
+                await SendEmailAsync(user?.Email, BuildEmailSubject("Comment Mention", review.PackageName), content);
             } 
         }
 
-        public async Task NotifyApproversOfReview(ClaimsPrincipal user, string apiRevisionId, HashSet<string> reviewers)
+        public async Task NotifyAssignedReviewersAsync(ClaimsPrincipal user, string apiRevisionId, HashSet<string> reviewers)
```
AlitzelMendez, line 113: is there are known reasons for us sending individual emails in this case?

src/dotnet/APIView/APIViewWeb/Pages/Assemblies/Review.cshtml.cs
```diff
@@ -325,8 +325,18 @@ public async Task<IActionResult> OnPostToggleAPIRevisionApprovalAsync(string id,
         /// <returns></returns>
         public async Task<ActionResult> OnPostRequestReviewersAsync(string id, string apiRevisionId, HashSet<string> reviewers)
```
AlitzelMendez, line 326: when is this email sent? I want to understand why we also need the `newlyAddedReviewers` in here :)

smw-ms: it's sent when you add someone as a reviewer. there was a bug sending an email to every reviewer already selected every time you would add new reviewer, which is now fixed

src/dotnet/APIView/APIViewWeb/Managers/NotificationManager.cs
smw-ms: reviewers can be added at different times. we don't want existing reviewers to get duplicate emails

PR #14007, opened 2026-02-14
[APIView] Diagnostic Updates

tjprescott:
This PR

- Adds correlation IDs to diagnostics which are backfilled if needed to associate multiple instances of the same diagnostic.
- Disables feedback and delete options on diagnostics.
- Makes the update button spin when applied so the UI does not appear to "not update".
- Fixed a bug with the severity dropdown in bulk update.

### DIFF
--- a/src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
+++ b/src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
@@ -9,6 +9,7 @@ using System.Net.Http.Headers;
 using System.Security.Claims;
 using System.Security.Cryptography;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
@@ -445,8 +446,18 @@ namespace APIViewWeb.Managers
         /// <returns></returns>
         public async Task SoftDeleteCommentAsync(ClaimsPrincipal user, CommentItemModel comment)
         {
-            await AssertOwnerAsync(user, comment);
-            var changeUpdate = ChangeHistoryHelpers.UpdateBinaryChangeAction(comment.ChangeHistory, CommentChangeAction.Deleted, user.GetGitHubLogin());
+            string userName = user.GetGitHubLogin();
+
+            if (comment.CommentSource == CommentSource.Diagnostic)
+            {
+                throw new AuthorizationFailedException();
+            }
+            else
+            {
+                await AssertOwnerAsync(user, comment);
+            }
+
+            var changeUpdate = ChangeHistoryHelpers.UpdateBinaryChangeAction(comment.ChangeHistory, CommentChangeAction.Deleted, userName);
             comment.ChangeHistory = changeUpdate.ChangeHistory;
             comment.IsDeleted = changeUpdate.ChangeStatus;
             await _commentsRepository.UpsertCommentAsync(comment);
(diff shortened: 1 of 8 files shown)


src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -445,8 +446,18 @@ public async Task SoftDeleteCommentAsync(ClaimsPrincipal user, string reviewId,
         /// <returns></returns>
         public async Task SoftDeleteCommentAsync(ClaimsPrincipal user, CommentItemModel comment)
         {
-            await AssertOwnerAsync(user, comment);
-            var changeUpdate = ChangeHistoryHelpers.UpdateBinaryChangeAction(comment.ChangeHistory, CommentChangeAction.Deleted, user.GetGitHubLogin());
+            string userName = user.GetGitHubLogin();
+
+            if (comment.CommentSource == CommentSource.Diagnostic)
```
AlitzelMendez, line 451: does this mean that we don't want to do sofdeletes for diagnostics?
 
why is the reason behind this change?

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -844,9 +855,31 @@ public async Task<List<CommentItemModel>> SyncDiagnosticCommentsAsync(
...
+                // related-comment grouping can work for older diagnostic comments.
+                var correlationByCommentId = new Dictionary<string, string>();
+                foreach (var diagnostic in diagnostics)
+                {
+                    string diagnosticHash = GenerateDiagnosticHash(diagnostic.TargetId, diagnostic.Text);
+                    string commentId = $"diag-{apiRevisionId}-{diagnosticHash}";
+                    correlationByCommentId[commentId] = GetDiagnosticCorrelationId(diagnostic);
+                }
+
+                foreach (var existingComment in existingDiagnosticCommentsForRevision)
```
AlitzelMendez, line 872: considering that there are places with 2k comments, can we have a way to don't do this one by one?

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -844,9 +855,31 @@ public async Task<List<CommentItemModel>> SyncDiagnosticCommentsAsync(
...
                     .ToList();
+
+                // Even when diagnostics are unchanged, backfill/update correlation ids so
+                // related-comment grouping can work for older diagnostic comments.
+                var correlationByCommentId = new Dictionary<string, string>();
+                foreach (var diagnostic in diagnostics)
+                {
+                    string diagnosticHash = GenerateDiagnosticHash(diagnostic.TargetId, diagnostic.Text);
+                    string commentId = $"diag-{apiRevisionId}-{diagnosticHash}";
+                    correlationByCommentId[commentId] = GetDiagnosticCorrelationId(diagnostic);
```
AlitzelMendez, line 869: when an under which circumstances we think this value is going to be updated?
why it need to be reviewed every time even diagnostics doesn't change? if diagnostics doesnt change no new relationship should be needed 

if this is just for the migration the number should be a small set, just the ones that wer already opened, we can create a script to calculate those

src/dotnet/APIView/ClientSPA/src/app/_components/shared/comment-thread/comment-thread.component.html
```diff
@@ -109,13 +109,13 @@
                             </a>
                         </ng-template>
                     </p-menu>
-                    <button [attr.data-btn-id]="comment.id" (click)="toggleUpVoteAction($event)" class="p-panel-header-icon p-link">
+                    <button *ngIf="!isDiagnostic(comment)" [attr.data-btn-id]="comment.id" (click)="toggleUpVoteAction($event)" class="p-panel-header-icon p-link">
```
AlitzelMendez, line 112: this kind of real-time checks are one of the things that are causing a huge issue on the performance, can we manage this in a different way?

src/dotnet/APIView/ClientSPA/src/app/_components/conversations/conversations.component.ts
```diff
@@ -273,7 +276,14 @@ export class ConversationsComponent implements OnChanges {
...
     if (this.comments.some(c => c.id === commentUpdates.commentId!)) {
-      this.comments.find(c => c.id === commentUpdates.commentId!)!.commentText = commentUpdates.commentText!;
+      const comment = this.comments.find(c => c.id === commentUpdates.commentId!)!;
+      if (commentUpdates.commentText !== undefined) {
+        comment.commentText = commentUpdates.commentText;
+      }
+      if (commentUpdates.severity !== undefined) {
+        comment.severity = commentUpdates.severity;
+      }
+      this.createCommentThreads();
```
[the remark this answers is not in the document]
AlitzelMendez: could you check this scenarios please? just test that everything keeps working

AlitzelMendez (changes requested):
could you please add E2E test/UI tests for the fixes that you are doing? 

like: repro the severity error and just have a test that does the bulk work (I think there are some test on the area already)

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
```diff
@@ -990,6 +1031,29 @@ private static string BuildDiagnosticCommentText(string diagnosticText, string h
...
+        private static string GetDiagnosticCorrelationId(CodeDiagnostic diagnostic)
+        {
+            if (!string.IsNullOrWhiteSpace(diagnostic?.DiagnosticId))
+            {
+                return diagnostic.DiagnosticId.Trim();
+            }
+
+            if (!string.IsNullOrWhiteSpace(diagnostic?.Text))
+            {
+                var match = Regex.Match(diagnostic.Text, @"\[([^\[\]]+)\]\s*$");
```
AlitzelMendez, line 1047: on this case, it would be better if we compile the regex just once, performance wise, like `private static readonly Regex TrailingBracketRegex = new(@"\[([^\[\]]+)\]\s*$", RegexOptions.Compiled);`

src/dotnet/APIView/APIViewWeb/Managers/CommentsManager.cs
tjprescott: That's right. The only way to "clear" a diagnostic should be to resolve the underlying issue. This has always been the case.

tjprescott:
@copilot please summarize the changes in this PR by what features it changes, bugs it fixes, etc. Please include the feedback from others. Rather than rebase this, I'm going to recreate it from scratch with this feedback.

PR #14036, opened 2026-02-18
Add WorkloadIdentity in azsdk cli auth chain for GH actions

praveenkuttappan:
GitHub agent workflow cannot use AzureCliCredential chain because gh-aw runs in a sandbox and it does not have access to $Home/.azure to get authenticated token on the Github runner machine. This PR adds workload identity credential into the chain when running on GitHub action or pipeline so CLI command running on the gh-aw action can use workload identity auth.

### DIFF
--- a/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Azure/AzureService.cs
+++ b/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Azure/AzureService.cs
@@ -18,13 +18,26 @@ public class AzureService : IAzureService
         // that we need to optimize for a cached credential.
         if (IsRunningInPipeline())
         {
-            return new ChainedTokenCredential(                
-                new WorkloadIdentityCredential(new WorkloadIdentityCredentialOptions { TenantId = tenantId }),
+            string? azureSubscriptionTenant = Environment.GetEnvironmentVariable("AZURESUBSCRIPTION_TENANT_ID");
+            string? azureSubscriptionClient = Environment.GetEnvironmentVariable("AZURESUBSCRIPTION_CLIENT_ID");
+            string? azureServiceConnection = Environment.GetEnvironmentVariable("AZURESUBSCRIPTION_SERVICE_CONNECTION_ID");
+            string? accessToken = Environment.GetEnvironmentVariable("SYSTEM_ACCESSTOKEN");
+
+            if (IsGitHubAction() || string.IsNullOrEmpty(azureSubscriptionTenant) || string.IsNullOrEmpty(azureSubscriptionClient) || string.IsNullOrEmpty(azureServiceConnection) || string.IsNullOrEmpty(accessToken))
+            {
+                return new ChainedTokenCredential(
+                    new WorkloadIdentityCredential(new WorkloadIdentityCredentialOptions { TenantId = tenantId }),
+                    new AzureCliCredential(new AzureCliCredentialOptions { TenantId = tenantId })
+                );
+            }
+
+            // Use AzurePipelineCredential in chain only when env values are present
+            return new ChainedTokenCredential(                        
                 // Environment variables for Azure pipeline credentials are created by Azure pipeline tasks AzureCLI@2 and AzurePowerShell@5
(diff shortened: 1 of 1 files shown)


tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Azure/AzureService.cs
```diff
@@ -18,7 +18,11 @@ public TokenCredential GetCredential(string? tenantId = null)
         // that we need to optimize for a cached credential.
         if (IsRunningInPipeline())
         {
-            return new AzureCliCredential(new AzureCliCredentialOptions { TenantId = tenantId });
+            return new ChainedTokenCredential(
+                new AzureCliCredential(new AzureCliCredentialOptions { TenantId = tenantId }),
+                new WorkloadIdentityCredential(new WorkloadIdentityCredentialOptions { TenantId = tenantId })
```
weshaggard: Will WorkloadIdentityCredential work for both github and ADO pipelines? Or will we also need the AzurePipelinesCredential as well?

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Azure/AzureService.cs
```diff
@@ -18,7 +18,11 @@ public TokenCredential GetCredential(string? tenantId = null)
...
+            );
+
```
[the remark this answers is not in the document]
weshaggard: Probably worth changing the order

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Azure/AzureService.cs
praveenkuttappan: Updated

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Azure/AzureService.cs
```diff
@@ -18,7 +18,13 @@ public TokenCredential GetCredential(string? tenantId = null)
         // that we need to optimize for a cached credential.
         if (IsRunningInPipeline())
         {
-            return new AzureCliCredential(new AzureCliCredentialOptions { TenantId = tenantId });
+            return new ChainedTokenCredential(                
+                new WorkloadIdentityCredential(new WorkloadIdentityCredentialOptions { TenantId = tenantId }),
+                // Environment variables for Azure pipeline credentials are created by Azure pipeline tasks AzureCLI@2 and AzurePowerShell@5
+                new AzurePipelinesCredential(Environment.GetEnvironmentVariable("AZURESUBSCRIPTION_TENANT_ID"), Environment.GetEnvironmentVariable("AZURESUBSCRIPTION_TENANT_ID"), Environment.GetEnvironmentVariable("AZURESUBSCRIPTION_SERVICE_CONNECTION_ID"), Environment.GetEnvironmentVariable("SYSTEM_ACCESSTOKEN")),
```
weshaggard: Do you know what happens if those env variables aren't present? I just want to be sure it behaves correctly and doesn't throw.

praveenkuttappan: Fixed to use AzurePipelineCredential only when env variables are present.

tools/azsdk-cli/Azure.Sdk.Tools.Cli/Services/Azure/AzureService.cs
```diff
@@ -18,7 +18,26 @@ public TokenCredential GetCredential(string? tenantId = null)
...
+            if (IsGitHubAction() || string.IsNullOrEmpty(azureSubscriptionTenant) || string.IsNullOrEmpty(azureSubscriptionClient) || string.IsNullOrEmpty(azureServiceConnection) || string.IsNullOrEmpty(accessToken))
+            {
+                return new ChainedTokenCredential(
+                    new WorkloadIdentityCredential(new WorkloadIdentityCredentialOptions { TenantId = tenantId }),
+                    new AzureCliCredential(new AzureCliCredentialOptions { TenantId = tenantId })
+                );
+            }
+
+            // Use AzurePipelineCredential in chain only when env values are present
+            return new ChainedTokenCredential(
```
benbp, line 35: Isn't it enough to just return this chained credential instead of checking all the env vars above? It should just skip past the azp cred in a gh actions context to find the workload identity one.

praveenkuttappan: Azp throws exception if param values are null. https://github.com/Azure/azure-sdk-for-net/blob/2c6db27865c453db6d43f552230424164bd818dc/sdk/identity/Azure.Identity/src/Credentials/AzurePipelinesCredential.cs#L48

